PR #147, opened 2014-10-29
Integrate new wallet address manager package

davecgh:
This pull request converts the wallet to use the new secure hierarchical deterministic wallet address manager package.

The following is an overview of modified functionality:

- The wallet must now be created before starting the executable
- A new flag `--create` has been added to create the new wallet using wizard style question and answer prompts
- Starting the process without an existing wallet will now display a message to run it with `--create`
- Providing the `--create` flag with an existing wallet will simply show an error and return
- When an existing legacy keystore exists, the `--create` flag will prompt  you for the old private passphrase and automatic convert all old addresses to the new wallet being created

Github seems to have incorrectly closed the previous pull request #138.


davecgh:
This has been updated to include all of the new paths and rebased accordingly.

jrick:
Because accounts are created with 0 addresses, but manager.LastExternalAddress returns the *next* address without error (which is not yet watched or managed), the getacctounaddress RPC is currently broken.  To fix, we will need Num{Internal,External}Addresses methods to the manager and check for 0 external addresses in the GetAccountAddress handler.

davecgh:
Another option is to create the first address on the internal and external branches when an account is created.

davecgh:
This has been updated for the btcscript -> btcd/txscript move and rebased accordingly.

jrick:
I've pushed a intwaddrmgr branch to this repo.  The keystore->waddrmgr replacement and integration here is in good shape but there are several other components (like txstore -> wtxmgr) that we want to complete before pushing this to master.  PRs which depend on intwaddr should be rebased to this branch and submitted as a intwaddrmgr PR (not a PR to master).

#154
#155
#183
#185

PR #155, opened 2014-12-17
Account name

tuxcanfly:

Updated `waddrmgr` to support account names and integrated with wallet RPC calls.


waddrmgr/db.go
```diff
@@ -432,6 +506,74 @@ func serializeBIP0044AccountRow(encryptedPubKey,
 	return rawData
 }
 
+// fetchAllAccounts loads information about all accounts from the database.
+// The returned value is a slice of account numbers which can be used to load
+// the respective account rows.
+func fetchAllAccounts(tx walletdb.Tx) ([]uint32, error) {
+	accounts := []uint32{}
```
tuxcanfly: I'm now using a 'meta' bucket to store metadata like no of accounts, no of addrs etc and using it to create sized slices.

Moved `bucket := ...` to the first line.

waddrmgr/db.go
```diff
@@ -432,6 +506,74 @@ func serializeBIP0044AccountRow(encryptedPubKey,
...
+}
+
+// fetchAccountName retreives the account name given an account number from
+// the database.
+func fetchAccountName(tx walletdb.Tx, account uint32) (string, error) {
+	bucket := tx.RootBucket().Bucket(acctNameIdxBucketName)
+
+	offset := uint32(0)
+	var acctName string
+	bucket.ForEach(func(k, v []byte) error {
```
tuxcanfly: I've added a new bucket to store in the index mapping account name to it's id.

waddrmgr/db.go
```diff
@@ -458,6 +600,49 @@ func fetchAccountInfo(tx walletdb.Tx, account uint32) (interface{}, error) {
...
+// putAddrAccountIndex stores the given key to the address account index of the database.
+func putAddrAccountIndex(tx walletdb.Tx, account uint32, addrHash []byte) error {
+	bucket, err := tx.RootBucket().Bucket(addrAcctIdxBucketName).
+		CreateBucketIfNotExists(uint32ToBytes(account))
+	if err != nil {
+		return nil
+	}
+
+	// Write a null value keyed by the address hash
+	err = bucket.Put(addrHash, []byte{0})
```
tuxcanfly: Added a var called `nullVal` and using it as a global.

waddrmgr/db.go
```diff
@@ -862,6 +1022,62 @@ func existsAddress(tx walletdb.Tx, addressID []byte) bool {
 	return bucket.Get(addrHash[:]) != nil
 }
 
+// fetchAccountAddresses loads information about addresses of an account from the database.
+// The returned value is a slice address rows for each specific address type.
+// The caller should use type assertions to ascertain the types.
+func fetchAccountAddresses(tx walletdb.Tx, account uint32) ([]interface{}, error) {
+	var addrs []interface{}
```
tuxcanfly: Yup, doing this now.

waddrmgr/manager.go
```diff
@@ -1470,6 +1485,188 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
 	return acctInfo.lastInternalAddr, nil
 }
 
+// NewAccount creates and returns a new account stored in the manager based
+// on the given account name.  If an account with the same name already exists,
+// ErrDuplicateAccount will be returned.  Since creating a new account requires
+// access to the cointype keys (from which extended account keys are derived),
+// it requires the manager to be unlocked.
+func (m *Manager) NewAccount(name string) (uint32, error) {
```
tuxcanfly: Yup, adding the mutex lock and checks as described. We were calling `LookupAccount` with the mutex lock held so I've added a unexpected version `lookupAccount` to make it usable.

waddrmgr/manager.go
```diff
@@ -1470,6 +1485,188 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
...
+		return putAccountInfo(tx, account, acctPubEnc,
+			acctPrivEnc, 0, 0, name)
+	})
+	return account, err
+}
+
+// RenameAccount renames an account stored in the manager based on the
+// given account number with the given name.  If an account with the same name
+// already exists, ErrDuplicateAccount will be returned.
+func (m *Manager) RenameAccount(account uint32, name string) error {
```
tuxcanfly: Checked and added mutex lock to newly exported funcs.

waddrmgr/manager.go
```diff
@@ -1470,6 +1485,188 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
...
+		if err != nil {
+			str := "failed to  encrypt public key for account"
+			return managerError(ErrCrypto, str, err)
+		}
+		acctPrivEnc, err := m.cryptoKeyPriv.Encrypt([]byte(acctKeyPriv.String()))
+		if err != nil {
+			str := "failed to encrypt private key for account"
+			return managerError(ErrCrypto, str, err)
+		}
+		// We have the encrypted account extended keys, so save them to the database
```
tuxcanfly: Re-formatted both to wrap properly.

waddrmgr/manager.go
```diff
@@ -1470,6 +1485,188 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
...
+			return managerError(ErrLocked, str, err)
+		}
+		coinTypeKeyPriv, err := hdkeychain.NewKeyFromString(string(serializedKeyPriv))
+		if err != nil {
+			str := fmt.Sprintf("failed to create cointype extended private key")
+			return managerError(ErrKeyChain, str, err)
+		}
+
+		// Derive the account key using the cointype key
+		acctKeyPriv, err := deriveAccountKey(coinTypeKeyPriv, account)
```
tuxcanfly: Oops... added the check now.

waddrmgr/manager.go
```diff
@@ -1470,6 +1485,188 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
...
+
+	// Fetch latest account, and create a new account in the same transaction
+	err = m.namespace.Update(func(tx walletdb.Tx) error {
+		var err error
+		// Fetch the latest account number to generate the next account number
+		account, err = fetchLastAccount(tx)
+		if err != nil {
+			return err
+		}
+		// TODO: needs to be atomic?
```
tuxcanfly: OK, removed the TODO.

waddrmgr/manager.go
```diff
@@ -1657,6 +1861,19 @@ func checkBranchKeys(acctKey *hdkeychain.ExtendedKey) error {
 	return err
 }
 
+// ValidateAccountName validates the given account name and returns an error, if any.
+func ValidateAccountName(name string) error {
```
tuxcanfly: Done.

waddrmgr/db.go
```diff
@@ -862,14 +1225,108 @@ func existsAddress(tx walletdb.Tx, addressID []byte) bool {
 	return bucket.Get(addrHash[:]) != nil
 }
 
+// fetchAddrAccount returns the account to which the given address belongs to.
+// It looks up the account using the addracctidx index which maps the address
+// hash to it's corresponding account id.
```
davecgh: it's -> its

waddrmgr/db.go
```diff
@@ -862,14 +1225,108 @@ func existsAddress(tx walletdb.Tx, addressID []byte) bool {
...
+		str := "account not found"
+		return 0, managerError(ErrAccountNotFound, str, nil)
+	}
+	return binary.LittleEndian.Uint32(val), nil
+}
+
+// fetchAccountAddresses loads information about addresses of an account from the database.
+// The returned value is a slice address rows for each specific address type.
+// The caller should use type assertions to ascertain the types.
+func fetchAccountAddresses(tx walletdb.Tx, account uint32) ([]interface{}, error) {
```
davecgh: I'd rather see this function make use of `fetchAddress` rather than repeating the same code which would require updating it in two places when a new address type is added.  I understand that `fetchAddress` expects the address whereas the account index stores the hash of the address (which is absolutely correct to do).  I'd suggest simply refactoring the common code into a separate function that both call with the hash of the address so that `fetchAddress` can take the hash of the passed address, while this function looks up the already hashed address and can pass it directly.

waddrmgr/db.go
```diff
@@ -862,14 +1225,108 @@ func existsAddress(tx walletdb.Tx, addressID []byte) bool {
...
+
+	accountBucket := tx.RootBucket().Bucket(addrAcctIdxBucketName).
+		Bucket(uint32ToBytes(account))
+	// if index bucket is missing the account, there hasn't been any address
+	// entries yet
+	if accountBucket == nil {
+		return []interface{}{}, nil
+	}
+
+	numAccountAddrs, err := fetchNumAccountAddrs(tx, account)
```
davecgh: Ignoring the error doesn't seem like proper behavior here.  The called function will return zero if the bucket has not yet been initialized, so any error that occurs here is unexpected and should be handled appropriately.

waddrmgr/db.go
```diff
@@ -862,14 +1225,108 @@ func existsAddress(tx walletdb.Tx, addressID []byte) bool {
...
+// The caller should use type assertions to ascertain the types.
+func fetchAccountAddresses(tx walletdb.Tx, account uint32) ([]interface{}, error) {
+	bucket := tx.RootBucket().Bucket(addrBucketName)
+
+	accountBucket := tx.RootBucket().Bucket(addrAcctIdxBucketName).
+		Bucket(uint32ToBytes(account))
+	// if index bucket is missing the account, there hasn't been any address
+	// entries yet
+	if accountBucket == nil {
+		return []interface{}{}, nil
```
davecgh: This can be `return nil, nil` to be consistent with the rest of the codebase.  Any calling code doing either `len(addrs)` or `for _, addr := range addrs` will work properly.

waddrmgr/db.go
```diff
@@ -862,14 +1225,108 @@ func existsAddress(tx walletdb.Tx, addressID []byte) bool {
...
+			str := fmt.Sprintf("unsupported address type '%d'",
+				row.addrType)
+			return managerError(ErrDatabase, str, nil)
+		}
+		if err != nil {
+			return err
+		}
+
+		// Handle index out of range due to invalid metadata
+		if i > len(addrs)-1 {
```
davecgh: This check should probably be done just after skipping buckets since there is no reason to load and deserialize an address row that there is no space for.

waddrmgr/db.go
```diff
@@ -292,6 +376,64 @@ func fetchWatchingOnly(tx walletdb.Tx) (bool, error) {
 	return buf[0] != 0, nil
 }
 
+// fetchNumAllAddresses loads the metadata - number of all addrs - from the database.
+func fetchNumAllAddresses(tx walletdb.Tx) (uint32, error) {
+	bucket := tx.RootBucket().Bucket(metaBucketName)
+
+	val := bucket.Get(numAllAddrsName)
+	// Return 0 if the metadata is uninitialized
+	if val == nil {
```
davecgh: Along the same lines as mentioned elsewhere, I would much rather this information be initialized at creation time with an error if it's missing here.  That is much more deterministic and doesn't hide errors as easily.

waddrmgr/db.go
```diff
@@ -292,6 +376,64 @@ func fetchWatchingOnly(tx walletdb.Tx) (bool, error) {
...
+func fetchNumAccountAddrs(tx walletdb.Tx, account uint32) (uint32, error) {
+	bucket := tx.RootBucket().Bucket(metaBucketName).Bucket(uint32ToBytes(account))
+	// Return 0 if bucket is uninitialized
+	if bucket == nil {
+		return 0, nil
+	}
+
+	val := bucket.Get(numAcctAddrsName)
+	// Return 0 if the metadata is uninitialized
+	if val == nil {
```
davecgh: Same deal here. This data should be initialized when the bucket is created and be an error if it's missing here.

waddrmgr/db.go
```diff
@@ -292,6 +376,64 @@ func fetchWatchingOnly(tx walletdb.Tx) (bool, error) {
...
+	return n, nil
+}
+
+// fetchNumAccounts loads the metadata - number of all accounts - from the database.
+func fetchNumAccounts(tx walletdb.Tx) (uint32, error) {
+	bucket := tx.RootBucket().Bucket(metaBucketName)
+
+	val := bucket.Get(numAccountsName)
+	// Return 0 if the metadata is uninitialized
+	if val == nil {
```
davecgh: Same here.

waddrmgr/db.go
```diff
@@ -432,6 +588,81 @@ func serializeBIP0044AccountRow(encryptedPubKey,
...
 
+// fetchAllAccounts loads information about all accounts from the database.
+// The returned value is a slice of account numbers which can be used to load
+// the respective account rows.
+func fetchAllAccounts(tx walletdb.Tx) ([]uint32, error) {
+	bucket := tx.RootBucket().Bucket(acctBucketName)
+
+	numAccounts, err := fetchNumAccounts(tx)
+	if err != nil {
+		return []uint32{}, err
```
davecgh: `return nil, err`

waddrmgr/db.go
```diff
@@ -432,6 +588,81 @@ func serializeBIP0044AccountRow(encryptedPubKey,
 	return rawData
 }
 
+// fetchAllAccounts loads information about all accounts from the database.
+// The returned value is a slice of account numbers which can be used to load
+// the respective account rows.
+func fetchAllAccounts(tx walletdb.Tx) ([]uint32, error) {
```
davecgh: Please add a TODO here to deal with the fact that we will likely want to switch over to an iterator over a slice in the future.  There can be up to 2^31-1 accounts, and while it's unlikely to be a problem anytime in the near future, clearly attempting to allocate and return a slice that large won't work.

waddrmgr/db.go
```diff
@@ -432,6 +588,81 @@ func serializeBIP0044AccountRow(encryptedPubKey,
...
+}
+
+// fetchAccountName retreives the account name given an account number from
+// the database.
+func fetchAccountName(tx walletdb.Tx, account uint32) (string, error) {
+	bucket := tx.RootBucket().Bucket(acctIdIdxBucketName)
+
+	val := bucket.Get(uint32ToBytes(account))
+	if val == nil {
+		str := "account name not found"
```
davecgh: I think this one is "account not found"

waddrmgr/db.go
```diff
@@ -432,6 +588,81 @@ func serializeBIP0044AccountRow(encryptedPubKey,
...
+}
+
+// fetchAccountByName retreives the account number given an account name
+// from the database.
+func fetchAccountByName(tx walletdb.Tx, name string) (uint32, error) {
+	bucket := tx.RootBucket().Bucket(acctNameIdxBucketName)
+
+	val := bucket.Get(stringToBytes(name))
+	if val == nil {
+		str := "account not found"
```
davecgh: While this one is "account name not found".

waddrmgr/db.go
```diff
@@ -484,36 +790,86 @@ func putAccountInfo(tx walletdb.Tx, account uint32, encryptedPubKey,
...
+	}
+	return nil
 }
 
-// fetchNumAccounts loads the number of accounts that have been created from
-// the database.
-func fetchNumAccounts(tx walletdb.Tx) (uint32, error) {
-	bucket := tx.RootBucket().Bucket(acctBucketName)
+// putNumAllAddrs stores the provided metadata - number of new addrs - to the database.
+func putNumAllAddrs(tx walletdb.Tx, count uint32) error {
```
davecgh: The behavior of this put should not be different from all the rest.  In particular, the existing put functions take the specific value to put and so should this one.  The problem with the current approach of taking a count and incrementing it is this means, it's impossible to delete addresses, which needs to be supported.  It will be common to allow watch-only addresses to be deleted, for example.

waddrmgr/db.go
```diff
@@ -484,36 +790,86 @@ func putAccountInfo(tx walletdb.Tx, account uint32, encryptedPubKey,
...
+	if err != nil {
+		str := fmt.Sprintf("failed to update metadata '%s'", numAllAddrsName)
+		return managerError(ErrDatabase, str, err)
+	}
+	return nil
+}
 
-	return binary.LittleEndian.Uint32(val), nil
+// putNumAccountAddrs stores the provided metadata - number of new account addrs - to the database.
+func putNumAccountAddrs(tx walletdb.Tx, account uint32, count uint32) error {
```
davecgh: The same is true for this function as `putNumAllAddrs`.  The function should take the value to write instead of a count to increment by.

waddrmgr/db.go
```diff
@@ -484,36 +790,86 @@ func putAccountInfo(tx walletdb.Tx, account uint32, encryptedPubKey,
...
+	}
+	return nil
 }
 
-// putNumAccounts stores the number of accounts that have been created to the
-// database.
-func putNumAccounts(tx walletdb.Tx, numAccounts uint32) error {
-	bucket := tx.RootBucket().Bucket(acctBucketName)
+// putNumAccounts stores the provided metadata - number of new accounts - to the database.
+func putNumAccounts(tx walletdb.Tx, count uint32) error {
```
davecgh: Same deal with the count versus value to write here.

waddrmgr/db.go
```diff
@@ -862,14 +1225,108 @@ func existsAddress(tx walletdb.Tx, addressID []byte) bool {
...
+		return nil, maybeConvertDbError(err)
+	}
+
+	return addrs, nil
+}
+
 // fetchAllAddresses loads information about all addresses from the database.
 // The returned value is a slice of address rows for each specific address type.
 // The caller should use type assertions to ascertain the types.
 func fetchAllAddresses(tx walletdb.Tx) ([]interface{}, error) {
```
davecgh: Please add a TODO here for an iterator in the same vain as `fetchAllAccounts`.   I know you didn't write this function, but I'd like for it to be called out as something we'll need to address.

It's even more important here than with `fetchAllAccounts` since you could potentially have 2^62-1 - 2^32-2 addresses.

waddrmgr/db.go
```diff
@@ -292,6 +376,64 @@ func fetchWatchingOnly(tx walletdb.Tx) (bool, error) {
 	return buf[0] != 0, nil
 }
 
+// fetchNumAllAddresses loads the metadata - number of all addrs - from the database.
+func fetchNumAllAddresses(tx walletdb.Tx) (uint32, error) {
```
davecgh: A uint32 is not sufficient to hold the potential number of addresses.  There can by 2^31 - 1 non-hardened addresses per account and up to 2^31 - 2 (one is reserved for imported addresses) accounts.  Thus, the max potential number of addresses is 2^62 - 2^32 - 2^31 + 2.  So a uint64 will cover it.

waddrmgr/db.go
```diff
@@ -1222,6 +1685,24 @@ func upgradeManager(namespace walletdb.Namespace) error {
...
+			str := "failed to create account id index bucket"
+			return managerError(ErrDatabase, str, err)
+		}
+
+		_, err = rootBucket.CreateBucketIfNotExists(metaBucketName)
+		if err != nil {
+			str := "failed to create meta bucket"
+			return managerError(ErrDatabase, str, err)
+		}
+
```
davecgh: As mentioned elsewhere, when the meta bucket is created, it think be initialized with all counters fields at zero and the other places can then error if the field isn't there.

waddrmgr/manager.go
```diff
@@ -1138,6 +1152,25 @@ func (m *Manager) Lock() error {
...
+// LookupAccount loads account number stored in the manager for the given
+// account name
+func (m *Manager) LookupAccount(name string) (uint32, error) {
+	m.mtx.RLock()
+	defer m.mtx.RUnlock()
+
+	return m.lookupAccount(name)
+}
+
+func (m *Manager) lookupAccount(name string) (uint32, error) {
```
davecgh: Please define functions before they're called to stay consistent. Also, there is no comment.

waddrmgr/manager.go
```diff
@@ -1470,6 +1503,238 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
...
+	}
+
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	if m.locked {
+		return 0, managerError(ErrLocked, errLocked, nil)
+	}
+
+	var account uint32
```
davecgh: These var declarations can be moved closer to the `Update` func that uses them since there is no need for them if the account validation bits fail.

waddrmgr/manager.go
```diff
@@ -1470,6 +1503,238 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
...
+	defer m.mtx.Unlock()
+
+	var acctName string
+	err := m.namespace.View(func(tx walletdb.Tx) error {
+		var err error
+		acctName, err = fetchAccountName(tx, account)
+		return err
+	})
+	if err != nil {
+		return acctName, err
```
davecgh: Please use explicit zero value returns since `acctName` could contain anything at this point.

```Go
return "", err
```

waddrmgr/manager.go
```diff
@@ -1470,6 +1503,238 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
...
+	defer m.mtx.Unlock()
+
+	var accounts []uint32
+	err := m.namespace.View(func(tx walletdb.Tx) error {
+		var err error
+		accounts, err = fetchAllAccounts(tx)
+		return err
+	})
+	if err != nil {
+		return accounts, err
```
davecgh:
```Go
return nil, err
```

waddrmgr/manager.go
```diff
@@ -1593,26 +1858,17 @@ func newManager(namespace walletdb.Namespace, net *btcnet.Params,
...
-//
-// In particular this is the hierarchical deterministic extended key path:
-//   m/44'/<coin type>'/<account>'
-func deriveAccountKey(masterNode *hdkeychain.ExtendedKey, coinType uint32,
-	account uint32) (*hdkeychain.ExtendedKey, error) {
-
+// deriveCoinTypeKey derives the cointype key which can be used to derive the
+// extended key for an account according to the hierarchy described by BIP0044
+// given the coin type key.
+func deriveCoinTypeKey(masterNode *hdkeychain.ExtendedKey,
```
davecgh:
```Go
//
+// In particular this is the hierarchical deterministic extended key path:
+// m/44'/<coin type>'
```

waddrmgr/manager.go
```diff
@@ -1593,26 +1858,17 @@ func newManager(namespace walletdb.Namespace, net *btcnet.Params,
...
-// hierarchy described by BIP0044 given the master node.
-//
-// In particular this is the hierarchical deterministic extended key path:
-//   m/44'/<coin type>'/<account>'
-func deriveAccountKey(masterNode *hdkeychain.ExtendedKey, coinType uint32,
-	account uint32) (*hdkeychain.ExtendedKey, error) {
-
+// deriveCoinTypeKey derives the cointype key which can be used to derive the
+// extended key for an account according to the hierarchy described by BIP0044
+// given the coin type key.
```
davecgh:
```Go
//
// In particular this is the hierarchical deterministic extended key path:
// m/44'/<coin type>'
```

davecgh:
The encrypted private coin type key is not being removed in `deletePrivateKeys`.

waddrmgr/db.go
```diff
@@ -148,14 +189,13 @@ var (
 	cryptoPubKeyName    = []byte("cpub")
 	cryptoScriptKeyName = []byte("cscript")
 	watchingOnlyName    = []byte("watchonly")
+	coinTypePrivKeyName = []byte("ctpriv")
```
davecgh: I'd prefer these to be before `watchingOnlyName` to keep the key names grouped.

davecgh:
The coinType extended keys need to be zeroed in the Create function immediately after their final use.  The other keys are cleared when the manager is locked, but the coin type is not.

waddrmgr/manager.go
```diff
@@ -1470,6 +1503,238 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
...
+			return err
+		}
+
+		// Decrypt the cointype key
+		serializedKeyPriv, err := m.cryptoKeyPriv.Decrypt(coinTypePrivEnc)
+		if err != nil {
+			str := fmt.Sprintf("failed to decrypt cointype serialized private key")
+			return managerError(ErrLocked, str, err)
+		}
+		coinTypeKeyPriv, err := hdkeychain.NewKeyFromString(string(serializedKeyPriv))
```
davecgh: The plain text coin type key bytes need to be zeroed as soon as they are used.
```Go
zero(serializedKeyPriv)
```

NOTE: This is intentionally before the error check so it happens regardless.

waddrmgr/manager.go
davecgh: The coin type extended key bytes need to be zeroed as soon as they are used.
```Go
coinTypeKeyPriv.Zero()
```

NOTE: This is intentionally before the error check so it happens regardless.

waddrmgr/manager_test.go
```diff
@@ -1117,6 +1117,202 @@ func testChangePassphrase(tc *testContext) bool {
 	return true
 }
 
+// testNewAccount tests the new account creation func of the address manager works
+// as expected.
+func testNewAccount(tc *testContext) bool {
+	// Cannot create new accounts in watching-only mode
+	if tc.watchingOnly {
+		return true
```
davecgh: This needs to verify that attempting to create a new account when in watching only mode returns `ErrWatchingOnly`.

waddrmgr/manager_test.go
```diff
@@ -1117,6 +1117,202 @@ func testChangePassphrase(tc *testContext) bool {
...
+	}
+	// Unlock the wallet to decrypt cointype keys required
```
davecgh: Before unlocking the wallet, this needs to verify that attempting to create a new account when the wallet is locked returns `waddrmgr.ErrLocked`.

waddrmgr/manager_test.go
```diff
@@ -1117,6 +1117,202 @@ func testChangePassphrase(tc *testContext) bool {
...
+		// Existing wallet manager will have 2 accounts
+		expectedAccounts = append(expectedAccounts, 2)
+	}
+	accounts, err := tc.manager.AllAccounts()
+	if err != nil {
+		tc.t.Errorf("AllAccounts: unexpected error: %v", err)
+		return false
+	}
+	if len(expectedAccounts) != len(accounts) {
+		if len(accounts) != len(expectedAccounts) {
```
davecgh: Unnecessary double if here.

waddrmgr/manager_test.go
```diff
@@ -1117,6 +1117,202 @@ func testChangePassphrase(tc *testContext) bool {
...
+		"15HNivzKhsLaMs1qRdQN1ifoJYUnJ2xW9z",
+		"13NhXy2nCLMwNug1TZ6uwaWnxp3uTqdDQq",
+	}
+	addrs, err := tc.manager.ActiveAccountAddresses(tc.account)
+	if err != nil {
+		tc.t.Errorf("ActiveAccountAddresses: unexpected error: %v", err)
+		return false
+	}
+	if len(expectedAddrs) != len(addrs) {
+		if len(addrs) != len(expectedAddrs) {
```
davecgh: Unnecessary double if here.

waddrmgr/db.go
```diff
@@ -862,18 +1212,118 @@ func existsAddress(tx walletdb.Tx, addressID []byte) bool {
...
+		return nil, maybeConvertDbError(err)
+	}
+
+	return addrs, nil
+}
+
 // fetchAllAddresses loads information about all addresses from the database.
 // The returned value is a slice of address rows for each specific address type.
 // The caller should use type assertions to ascertain the types.
+// TODO(tuxcanfly): Switch over to an iterator to support the maximum of 2^32 - 2^31 + 2 addrs
```
davecgh: Missed the most significant 2^62 term in the max number of addresses.  It should be `2^62 - 2^32 - 2^31 + 2`.

waddrmgr/manager.go
tuxcanfly: Zeroing the `coinTypeKeyPriv` now.

waddrmgr/db.go
tuxcanfly: Added a func `fetchAddressByHash` which can be re-used from `fetchAddress` as well `fetchAccountAddresses`, `fetchAllAddresses` One problem here is we are expecting `addressID` in `deserializeAddressRow` but we don't have it when querying by `addrHash`. Using the `addrHash` as a placeholder for now.

waddrmgr/manager.go
tuxcanfly: Zeroing the `serializedKeyPriv` now.

tuxcanfly:
Now deleting cointype keys in `deletePrivateKey` and zeroing them in `Create` as well as `NewAccount`

waddrmgr/db.go
```diff
@@ -717,27 +1049,34 @@ func serializeScriptAddress(encryptedHash, encryptedScript []byte) []byte {
...
 // address type.  The caller should use type assertions to ascertain the type.
 func fetchAddress(tx walletdb.Tx, addressID []byte) (interface{}, error) {
+	addrHash := fastsha256.Sum256(addressID)
+	return fetchAddressByHash(tx, addrHash[:])
+}
+
+// fetchAddressByHash loads address information for the provided address hash from
+// the database.  The returned value is one of the address rows for the specific
+// address type.  The caller should use type assertions to ascertain the type.
+func fetchAddressByHash(tx walletdb.Tx, addrHash []byte) (interface{}, error) {
```
davecgh: Define before `fetchAddress`, please.

waddrmgr/db.go
davecgh: Fair point.  It's only being used for the error message though.  How about just removing that parameter from the function calls and error messages, and then, at the two call sites for `fetchAddress`:

```Go
if err != nil {
	if merr, ok := err.(*ManagerError); ok {
		desc := fmt.Sprintf("failed to fetch address '%s': %v",
			address.ScriptAddress(), merr.Description)
		merr.Description = desc
	}
}
```

Then add a comment on the errors in the functions that they are later prefixed with the address that caused the failure.

rpcserver.go
```diff
@@ -165,6 +165,38 @@ func (c *websocketClient) send(b []byte) error {
...
+
+// isManagerDuplicateAddressError returns whether or not the passed error is due to a
+// duplicate item being provided to the address manager.
+func isManagerDuplicateAddressError(err error) bool {
+	merr, ok := err.(waddrmgr.ManagerError)
+	if !ok {
+		return false
+	}
+
+	return merr.ErrorCode == waddrmgr.ErrDuplicateAddress
```
jrick: Please use the same `ok && ...` idiom here to match the other `isFooError` functions.

rpcserver.go
```diff
@@ -972,16 +991,12 @@ func (s *rpcServer) PostClientRPC(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Create the response and error from the request.  Three special cases
```
jrick: Two

wallet.go
```diff
@@ -156,19 +174,53 @@ func (w *Wallet) updateNotificationLock() {
 	w.notificationLock = noopLocker{}
 }
 
+// GetCreditAccount returns the first account that can be associated
+// with the given credit.
+// If no account is found, ErrAccountNotFound is returned.
+func (w *Wallet) GetCreditAccount(c txstore.Credit) (uint32, error) {
```
jrick: Don't use Get in the method name for getters; it's not idiomatic Go (same thing below, for GetDebitAccount).

Also, would it make sense to return both the account number and name here, or is that an expensive lookup and isn't needed in all places?

tuxcanfly: Thanks, I'll rename that.

The name is not required in all places and requires a db hit so it might not be worth it to include here.

rpcserver.go
```diff
@@ -1333,6 +1360,7 @@ var rpcHandlers = map[string]requestHandler{
 	// Reference implementation wallet methods (implemented)
 	"addmultisigaddress":     AddMultiSigAddress,
 	"createmultisig":         CreateMultiSig,
+	"createnewaccount":       CreateNewAccount,
```
jrick: Move this (and renameaccount) to to the extensions below; there is no equivalent bitcoind request.

rpcserver.go
```diff
@@ -1684,16 +1705,20 @@ func GetAddressesByAccount(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd)
...
+	var balance btcutil.Amount
+	var err error
+	if cmd.Account == nil || *cmd.Account == "*" {
+		balance, err = w.CalculateBalance(cmd.MinConf)
+	} else {
+		account, err := w.Manager.LookupAccount(*cmd.Account)
+		if err != nil {
+			return nil, err
+		}
+		balance, err = w.CalculateAccountBalance(account, cmd.MinConf)
```
jrick: err is shadowed here.  The check after the if block will never fail if this else branch is taken.

rpcserver.go
```diff
@@ -1738,13 +1765,17 @@ func GetAccount(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interface{
...
+	// Fetch the associated account
+	account, err := w.Manager.AddrAccount(addr)
 	if err != nil {
 		return nil, btcjson.ErrInvalidAddressOrKey
 	}
 
-	return "", nil
+	acctName, err := w.Manager.AccountName(account)
+	if err != nil {
+		return nil, btcjson.ErrInvalidAddressOrKey
```
jrick: Wrong error

rpcserver.go
```diff
@@ -1827,18 +1856,57 @@ func KeypoolRefill(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interfa
 	return nil, nil
 }
 
-// GetNewAddress handlesa getnewaddress request by returning a new
-// address for an account.  If the account does not exist or the keypool
-// ran out with a locked wallet, an appropiate error is returned.
+// CreateNewAccount handles a createnewaccount request by creating and
+// returning a new account. If the last account has no transaction history
```
jrick: The returned result is always nil (or JSON null).  If it's not already, can we make the default name for new accounts the string of the account number, and return that?

rpcserver.go
```diff
@@ -1991,8 +2072,14 @@ func GetTransaction(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interf
 			}
 		}
 
+		var acctName string
+		account, err := w.GetCreditAccount(cred)
+		if err == nil {
+			acctName, err = w.Manager.AccountName(account)
```
jrick: unchecked error

rpcserver.go
```diff
@@ -2008,13 +2095,24 @@ func GetTransaction(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interf
...
+	for _, account := range accounts {
+		acctName, err := w.Manager.AccountName(account)
+		if err != nil {
+			return nil, err
+		}
+		bal, err := w.CalculateAccountBalance(account, cmd.MinConf)
+		if err != nil {
+			return nil, err
+		}
+		accountBalances[acctName] = bal.ToUnit(btcutil.AmountBTC)
```
jrick: You can use the new Amount.ToBTC method here (perhaps a good idea to just search and replace it everywhere).

rpcserver.go
```diff
@@ -2038,34 +2136,26 @@ func ListLockUnspent(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (inter
...
-
-	for _, record := range w.TxStore.Records() {
-		for _, credit := range record.Credits() {
-			if !credit.Confirmed(cmd.MinConf, bs.Height) {
-				// Not enough confirmations, skip the current block.
-				continue
-			}
-			amount += credit.Amount()
-			confirmations = credit.Confirmations(bs.Height)
+	ret := []btcjson.ListReceivedByAccountResult{}
```
jrick: make() this with len=0 and cap=len(accounts)

rpcserver.go
```diff
@@ -2124,14 +2220,25 @@ func ListReceivedByAddress(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd)
...
 				addrStr := address.EncodeAddress()
 				addrData, ok := allAddrData[addrStr]
+				account, err := w.Manager.AddrAccount(address)
+				if err != nil {
+					return nil, btcjson.ErrInvalidAddressOrKey
+				}
+
+				acctName, err := w.Manager.AccountName(account)
+				if err != nil {
+					return nil, btcjson.ErrInvalidAddressOrKey
```
jrick: Wrong error

rpcserver.go
```diff
@@ -2212,12 +2319,21 @@ func ListSinceBlock(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interf
...
+	var transactions []btcjson.ListTransactionsResult
+	var err error
+	if cmd.Account == nil || *cmd.Account == "*" {
+		transactions, err = w.ListTransactions(cmd.From, cmd.Count)
+	} else {
+		account, err := w.Manager.LookupAccount(*cmd.Account)
+		if err != nil {
+			return nil, err
+		}
+		transactions, err = w.ListAccountTransactions(account, cmd.From, cmd.Count)
```
jrick: err is shadowed

rpcserver.go
```diff
@@ -2257,12 +2373,21 @@ func ListAddressTransactions(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd
...
+	var transactions []btcjson.ListTransactionsResult
+	var err error
+	if cmd.Account == nil || *cmd.Account == "*" {
+		transactions, err = w.ListAllTransactions()
+	} else {
+		account, err := w.Manager.LookupAccount(*cmd.Account)
+		if err != nil {
+			return nil, err
+		}
+		transactions, err = w.ListAllAccountTransactions(account)
```
jrick: shadowed

rpcserver.go
```diff
@@ -2434,7 +2559,8 @@ func SendToAddress(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interfa
 		cmd.Address: btcutil.Amount(cmd.Amount),
 	}
 
-	return sendPairs(w, chainSvr, cmd, pairs, 1)
+	// sendtoaddress always debits the default account, this matches bitcoind
```
jrick: This isn't incorrect, but it behaves differently.  We group utxos by account, but bitcoind does not.  Debiting from an account may spend *any* available input, and simply decrements the balance associated with that account.

That said, I'm ok with sendtoaddress only spending from the default account's utxos, as it's the least surprising behavior, and other RPCs exist for spending out of other accounts.

rpcserver.go
```diff
@@ -2853,31 +2910,64 @@ func ValidateAddress(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (inter
...
-
-			class := sa.ScriptClass()
-			// script type
-			result.Script = class.String()
-			if class == btcscript.MultiSigTy {
-				result.SigsRequired = int32(sa.RequiredSigs())
-			}
+	ainfo, err := w.Manager.Address(addr)
+	if managerErr, ok := err.(waddrmgr.ManagerError); ok {
+		if managerErr.ErrorCode == waddrmgr.ErrAddressNotFound {
```
jrick: Drop this into the `if err != nil` below and add a `isAddressNotFoundError` method to match the other error code checks.

rpcserver.go
```diff
@@ -2853,31 +2910,64 @@ func ValidateAddress(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (inter
...
+	if err != nil {
+		return nil, err
+	}
+
+	// The address lookup was successful which means there is further
+	// information about it available and it is "mine".
+	result.IsMine = true
+	acctName, err := w.Manager.AccountName(ainfo.Account())
+	if err != nil {
+		return result, err
```
jrick: JSON-RPC responses must have either a result or an error, not both.  There should a check further up in the server code that only includes the error in the response, but do return nil for the result here as well.

rpcserver.go
```diff
@@ -2853,31 +2910,64 @@ func ValidateAddress(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (inter
...
+	}
+	result.Account = acctName
+
+	switch ma := ainfo.(type) {
```
jrick: Please verify the types this switch is checking against are correct.  I think they may be pointers.

Oh what I'd do for variant types...

rpcserver.go
tuxcanfly: I meant to provide rationale for passing `waddrmgr.DefaultAccountNum` in `sendPairs`. I've updated to say "spends from the default account" .

waddrmgr/db.go
tuxcanfly: Added modified err in callers for both `fetchAddress` and `fetchAddressByHash` and updated comments accordingly.

tuxcanfly:
Thanks @davecgh @jrick  I've addressed all the items reviewed so far. I've rebased and squashed all the commits into two final commits. The major remaining things are

* default account names - as discussed we could make the new name optional and default it to the string of the account number ... also @davecgh mentioned having an alias for default account name e.g "default"  to help identify it more easily in RPC results ... on the same note, imported addresses account could be named "imported"

* deviating from BIP044 to make new account creation more easier. as per [https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery](BIP 0044):

    This algorithm is successful because software should disallow creation of new accounts if previous     one has no transaction history, as described in chapter "Account" above

as discussed, we plan to deviate from this and allow new accounts regardless of tx history for the first few X accounts (where X=100) Accordingly, we need to handle this when discovering accounts. 

Let me know in case I missed anything. I have a backup of the [previous review branch](https://github.com/tuxcanfly/btcwallet/tree/account-name-review) in case I've missed anything.

jcvernaleo:
@tuxcanfly 

The master branch of btcwallet doesn't compile any more due to a recent change in btcjson from the lastest PR.  @davecgh said you had the corresponding fix to that on this branch.  I know this branch is big and is a work in progress, but would it be possible to break that small change out so we can get btcwallet master compiling again?

jcvernaleo:
@tuxcanfly Ignore my last comment.  Looks like you did something else here anyway so I committed a fix so btcwallet compiles again.  Sorry for the noise.

tuxcanfly:
@davecgh I've updated the PR for the changes described above, i.e

* use 'default' and 'imported' account names
* allow a limited number of empty accounts

There are a few things we need to consider:

* Do we want to keep an alias for blank name to 'default' to maintain compat with existing RPC calls?
* Imported account will be listed along with the usual accounts in all account related RPC calls, hope that's OK

davecgh:
@tuxcanfly:  Nice.

Yes I think we should treat '' as 'default' for compatibility, however, the results would still list the account as 'default' even if '' is specified.

I also think showing the imported account makes sense.  It'll make it clear that it's a separate account and therefore has the same type of semantics as any other account, with the exception of being able to generate chained addresses and be recovered from the seed.

tuxcanfly:
Thanks, makes sense, will work on that.

tuxcanfly:
@davecgh Implemented both items as described above.

davecgh:
Thanks.  I'll review everything shortly.

waddrmgr/db.go
```diff
@@ -1324,3 +1632,49 @@ func upgradeManager(namespace walletdb.Namespace) error {
...
+// The following buckets were introduced in version 3 to support account names:
+// * acctNameIdxBucketName
+// * acctIdIdxBucketName
+// * metaBucketName
+func upgradeToVersion3(namespace walletdb.Namespace) error {
+	err := namespace.Update(func(tx walletdb.Tx) error {
+		currentMgrVersion := uint32(3)
+		rootBucket := tx.RootBucket()
+
+		_, err := rootBucket.CreateBucketIfNotExists(acctNameIdxBucketName)
```
davecgh: `CreateBucket`

waddrmgr/db.go
```diff
@@ -1324,3 +1632,49 @@ func upgradeManager(namespace walletdb.Namespace) error {
 
 	return nil
 }
+
+// upgradeToVersion3 upgrades the database from version 1 to version 3
```
davecgh: from version 2 to version 3.

waddrmgr/db.go
```diff
@@ -1312,6 +1610,16 @@ func upgradeManager(namespace walletdb.Namespace) error {
 	//	version = 3
 	// }
 
+	if version < 3 {
+		// Upgrade from version 1 to 3.
```
davecgh: from version 2 to 3.

davecgh:
This needs to depend on #207 because it will bump the version to version 3 and then #207 wouldn't work since the manager would already be a higher version.

davecgh:
This is looking pretty code, but I don't see any code to handle storing the coin type keys during an upgrade.  I see it is stored during `Create` with the `putCoinTypeKeys` function, but upgraded managers won't work.

I realize this is tricky because the master seed is not available, so you can't derive the coin type keys in order to store them.  I think this is a case where what we discussed in chat comes up.  The caller needs to provide additional information for the upgrade to be able to complete.  In this case, the master seed.

jrick:
I believe this will break some of the transaction creation RPCs that allow using `*` as the account name to spend any credit from any account.  We will either need a magic value for the account integer (but iirc `^uint32(0)` is already reserved for imported addresses) or some other flag to ignore the account credit checks when selecting inputs.

wallet.go
```diff
@@ -171,6 +176,40 @@ func (w *Wallet) updateNotificationLock() {
...
+// If no account is found, ErrAccountNotFound is returned.
+func (w *Wallet) DebitAccount(d txstore.Debits) (uint32, error) {
+	msgTx := d.Tx().MsgTx()
+	for _, txIn := range msgTx.TxIn {
+		op := txIn.PreviousOutPoint
+		record, ok := w.TxRecord(&op.Hash)
+		if !ok {
+			return 0, btcjson.ErrNoTxInfo
+		}
+		for _, c := range record.Credits() {
```
jrick: This is incorrect.  It should look up the account of credit that was spent, not the first credit created by the transaction (which may be sent to anther account's address).

tuxcanfly:
@davecgh added upgrade path including prompting for seed and passphrases. Two minor remaining issues:

* we need `pubPassPhrase` in `upgradeManager` but directly making it a param seems odd. I'm prompting for both priv and pub passphrases currently.

* there's a small bug due to which an extraneous account seems to be inserted during upgrade, trying to debug this.

I've moved a few funcs common to setup and upgrade to pkg `cmd` to make the re-usable.

Please take a look and let me know if you think this is on the right track.

davecgh:
I'd rather the new package be named something else.  We typically use cmd to denote utilities and such.

waddrmgr/db.go
```diff
@@ -1400,3 +1738,140 @@ func upgradeManager(namespace walletdb.Namespace) error {
...
+// * acctIdIdxBucketName
+// * metaBucketName
+func upgradeToVersion3(namespace walletdb.Namespace) error {
+	err := namespace.Update(func(tx walletdb.Tx) error {
+		currentMgrVersion := uint32(3)
+		rootBucket := tx.RootBucket()
+
+		reader := bufio.NewReader(os.Stdin)
+
+		seed, err := cmd.PromptExistingConsoleSeed(reader)
```
davecgh: The upgrade function must not prompt the user directly. This information needs to be provided by the caller through some kind of callback mechanism.

Keep in mind that the waddrmgr is intended to be useful for more than just wallet, and ultimately will be used in a GUI too.

waddrmgr/db.go
```diff
@@ -1400,3 +1738,140 @@ func upgradeManager(namespace walletdb.Namespace) error {
...
+		if err != nil {
+			return err
+		}
+
+		privPassPhrase, err := cmd.PromptConsolePass(reader, "Enter the private passphrase of your wallet", false)
```
davecgh: The same is true here.  None of this can prompt the user directly.

tuxcanfly:
Tried using callbacks passed down from wallet to waddrmgr to prompt for the seed, priv passphrase when required. Not sure if that's better than the suggestion discussed on IRC but found it simpler to implement it this way.

waddrmgr/db.go
```diff
@@ -38,6 +39,8 @@ var (
 	latestMgrVersion uint32 = LatestMgrVersion
 )
 
+type obtainFunc func() ([]byte, error)
```
davecgh: This should be exported since it's used in the Confg struct. It'll need a comment too once exported.

waddrmgr/db.go
```diff
@@ -1334,14 +1663,12 @@ func upgradeToVersion2(namespace walletdb.Namespace) error {
 
 // upgradeManager upgrades the data in the provided manager namespace to newer
 // versions as neeeded.
-func upgradeManager(namespace walletdb.Namespace) error {
-	// Get the current version.
+func upgradeManager(namespace walletdb.Namespace, pubPassPhrase []byte, obtainSeed, obtainPrivPassPhrase obtainFunc) error {
```
davecgh: As noted below, I suggest changing the signature to accept the entire config struct so it won't have to be changed in the future if any new information is needed.

waddrmgr/manager.go
```diff
@@ -1858,7 +2170,7 @@ func Open(namespace walletdb.Namespace, pubPassphrase []byte, chainParams *chain
 	}
 
 	// Upgrade the manager to the latest version as needed.
-	if err := upgradeManager(namespace); err != nil {
+	if err := upgradeManager(namespace, pubPassphrase, config.ObtainSeed, config.ObtainPrivatePass); err != nil {
```
davecgh: This would panic if a nil config (which is valid) is passed.  I'd move the section below which sets the default config above this line and change the `upgradeManager` signature to take the entire config struct instead of the individual entries.

waddrmgr/manager.go
```diff
@@ -82,11 +88,23 @@ const (
...
+
 // Options is used to hold the optional parameters passed to Create or Load.
 type Options struct {
-	ScryptN int
-	ScryptR int
-	ScryptP int
+	ScryptN           int
+	ScryptR           int
+	ScryptP           int
+	ObtainSeed        obtainFunc
```
davecgh: Please add a comment for this field's purpose.

```Go
	// ObtainSeed is a callback function that is potentially invoked during
	// upgrades.  It is intended to be used to request the wallet seed
	// from the user (or any other mechanism the caller deems fit).
```

waddrmgr/manager.go
```diff
@@ -82,11 +88,23 @@ const (
...
+	ObtainPrivatePass obtainFunc
```
davecgh: Please add a comment for this field's purpose.

```Go
	// ObtainPrivatePass is a callback function that is potentially invoked
	// during upgrades.  It is intended to be used to request the wallet
	// private passphrase from the user (or any other mechanism the caller
	// deems fit).
```

jrick:
I think that the account names should not be set in the RPC handlers for gettransaction and listtransactions.  These fields are deprecated in recent bitcoind wallet versions (which makes sense, since account balances there can change without an on-chain transaction) and we have plans for a proper ledger RPC to report credits and debits across accounts that is intentionally incompatible with bitcoind.

tuxcanfly:
@jrick Agreed, that would simplify some hoops I had to jump through to make the RPC results compat with core.

wallet.go
```diff
@@ -34,10 +36,12 @@ import (
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcutil/hdkeychain"
 	"github.com/btcsuite/btcwallet/chain"
 	"github.com/btcsuite/btcwallet/txstore"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/walletdb"
+	"golang.org/x/crypto/ssh/terminal"
```
jrick: Use "github.com/btcsuite/golangcrypto/ssh/terminal".

rpcserver.go
```diff
@@ -1044,19 +1060,29 @@ func (b blockDisconnected) notificationCmds(w *Wallet) []btcjson.Cmd {
...
-	n := btcws.NewTxNtfn("", &ltr)
+	n := btcws.NewTxNtfn(acctName, &ltr)
 	return []btcjson.Cmd{n}
 }
 
 func (d txDebit) notificationCmds(w *Wallet) []btcjson.Cmd {
 	blk := w.Manager.SyncedTo()
-	ltrs, err := txstore.Debits(d).ToJSON("", blk.Height, activeNet.Params)
+	acctName := waddrmgr.DefaultAccountName
+	if debitAccount, err := w.DebitAccount(txstore.Debits(d)); err == nil {
```
jrick: I would also just use "" here (this is an old api and will eventually be replaced) instead of looking up the account.  This is the last call to DebitAccount, so with this gone, you can remove that func as well.

wallet.go
```diff
@@ -722,6 +808,22 @@ func (w *Wallet) AddressUsed(addr waddrmgr.ManagedAddress) bool {
 	return addr.Used()
 }
 
+// AccountUsed returns whether there are any recorded transactions spending to
+// a given account. It returns true if atleast one address in the account was used
+// and false if no address in the account was used.
+func (w *Wallet) AccountUsed(account uint32) (bool, error) {
+	addrs, err := w.Manager.ActiveAccountAddresses(account)
```
jrick: This is an assumption because currently ActiveAccountAddresses returns *all* addresses from the account, but may not in the future.  If it is changed (when single use addrs are added) to only return unused addresses, the used account may be marked unused.

wallet.go
```diff
@@ -735,19 +837,52 @@ func (w *Wallet) CalculateBalance(confirms int) (btcutil.Amount, error) {
...
+// CalculateAccountBalance sums the amounts of all unspent transaction
+// outputs to the given account of a wallet and returns the balance.
+func (w *Wallet) CalculateAccountBalance(account uint32, confirms int) (btcutil.Amount, error) {
+	var bal btcutil.Amount
+
+	// Get current block.  The block height used for calculating
+	// the number of tx confirmations.
+	blk := w.Manager.SyncedTo()
+
+	for _, r := range w.TxStore.Records() {
```
jrick: This should only be checking unspent outputs, not every record.

wallet.go
```diff
@@ -1179,6 +1323,22 @@ func (w *Wallet) ResendUnminedTxs() {
 	}
 }
 
+// SortedActiveAccountAddresses returns a slice of addresses of an account in a wallet.
+func (w *Wallet) SortedActiveAccountAddresses(account uint32) ([]string, error) {
```
jrick: I would just remove this.  There's no reason to need addresses to be sorted.  I think this was previously added to mimic an implementation detail in bitcoind (a `std::map` is usually implemented as an RB tree, so iterating over k/v pairs has a defined order).  Instead, I would change `GetAddressesByAccount` to return a JSON array of all addresses (even those not marked active) in any order.

wallet.go
```diff
@@ -1262,6 +1420,37 @@ func (w *Wallet) TotalReceived(confirms int) (btcutil.Amount, error) {
 	return amount, nil
 }
 
+// TotalReceivedForAccount iterates through a wallet's transaction history,
+// returning the total amount of bitcoins received for a single wallet
+// account.
+func (w *Wallet) TotalReceivedForAccount(account uint32, confirms int) (btcutil.Amount, uint64, error) {
```
jrick: The previous function which did this and didn't take the account (GetAddressesByAccount) is no longer used anywhere and should be removed.

tuxcanfly: We need this for `getreceivedbyaccount`, `listreceivedbyaccount`

jrick: Right, we need `TotalReceivedForAccount`, but `TotalReceived` is no longer used and is dead code.

rpcserver.go
```diff
@@ -2204,12 +2294,7 @@ func ListSinceBlock(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interf
 func ListTransactions(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interface{}, error) {
 	cmd := icmd.(*btcjson.ListTransactionsCmd)
 
-	var account string
-	if cmd.Account != nil {
-		account = *cmd.Account
-	}
-
-	err := checkAccountName(account)
+	err := checkAccountName(*cmd.Account)
```
jrick: This is not safe.  The field is optional and this may be nil.

tuxcanfly:
@jrick Thanks, updated.

davecgh:
Alright, this looks good to go!  Please squash it down to a single commit and rebase to the latest master.

tuxcanfly:
Done.

tuxcanfly:
Great! Thanks for the patient reviews @davecgh @jrick

PR #178, opened 2015-01-30
Implement votingpool's startwithdrawal

gsalgado:
http://opentransactions.org/wiki/index.php/Startwithdrawal is basically what
this implements.  The interesting bits are in withdrawal.go and
input_selection.go. The former has the implementation of http://goo.gl/iUrQ3s
and the latter implements http://goo.gl/znOECm.

There are a lot of mechanical changes (e.g. votingPool -> pool in variables,
vp -> p in method receivers) and moving of some tests (e.g.  serialization) to
pool_wb_test.go

It also includes a couple functions to construct sigscripts using the raw
signatures returned by Pool.StartWithdrawal(). These are going to be used in
the implementation of
http://opentransactions.org/wiki/index.php/Updatewithdrawal but right now
they're used in tests to make sure the signatures returned by startwithdrawal
are correct.

Some of our test files have a _wb_ infix, which stands for whitebox. These are
in the same package (votingpool) as the actual code they test, so that they
can access the internal bits. We decided to use this approach because
there are too many internal bits (including data types) we wanted to test
thoroughly and it'd be annoying to use internal_test.go to expose them all for
tests.


davecgh:
We haven't forgot this.  The plan is to get it reviewed after #204 and #213 are merged.  I expect that to be within the next few days.

gsalgado:
Cool! This branch rebases cleanly on top of them, but I'll wait for them to be merged before I do so

votingpool/db.go
```diff
@@ -22,12 +22,13 @@ import (
 	"fmt"
 
 	"github.com/btcsuite/btcwallet/snacl"
-	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/walletdb"
 )
 
+// TODO: Rename votingPoolID to poolID everywhere in this file.
```
davecgh: Looks like this hasn't been done yet.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+	bucket, err := usedAddrs.CreateBucketIfNotExists(getUsedAddrBucketID(seriesID, branch))
+	if err != nil {
+		return newError(ErrDatabase, "failed to store used address hash", err)
+	}
+	return bucket.Put(uint32ToBytes(uint32(index)), encryptedHash)
+}
+
+// getUsedAddrHash returns the addr hash with the given index from the used
+// addresses bucket of the given pool, series and branch.
+func getUsedAddrHash(
```
davecgh: nitpick: The rest of function signatures throughout the code base use as much of the line as possible before wrapping to the next.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+	tx walletdb.Tx, poolID []byte, seriesID uint32, branch Branch, index Index) []byte {
+
+	usedAddrs := tx.RootBucket().Bucket(poolID).Bucket(usedAddrsBucketName)
```
davecgh: I haven't reviewed far enough to see if the pool id existence is checker before calling this function, but this will cause a panic when the id doesn't exist.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
 	privKeysEncrypted [][]byte
 }
 
+// getUsedAddrBucketID returns the used addresses bucket ID for the given series
+// and branch. It has the form seriesID:branch.
+func getUsedAddrBucketID(seriesID uint32, branch Branch) []byte {
+	return bytes.Join(
```
davecgh: nitpick: The rest of function code base uses as much of the line as possible before wrapping to the next.  There are a few more of these, but I won't comment on every one.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+		[][]byte{uint32ToBytes(seriesID), uint32ToBytes(uint32(branch))}, []byte(":"))
```
davecgh: This is creating a fair amount of garbage (2 byte slices for the seriesID and branch, a bytes slice for the colon, and a slice of slices for the fields).  It would obviously depend on how often this is called to determine whether it's worth it, but it would be more efficient to do something like:

```Go
var bucketID [9]
binary.LittleEndian.PutUint32(bucketID[0:4], seriesID)
bucketID[4] = ':'
binary.LittleEndian.PutUint32(bucketID[5:9], uint32(branch))
return bucketID[:]
```

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+}
+
+// putUsedAddrHash adds an entry (key==index, value==encryptedHash) to the used
+// addresses bucket of the given pool, series and branch.
+func putUsedAddrHash(tx walletdb.Tx, poolID []byte, seriesID uint32, branch Branch,
+	index Index, encryptedHash []byte) error {
+
+	usedAddrs := tx.RootBucket().Bucket(poolID).Bucket(usedAddrsBucketName)
```
davecgh: Will panic if called with an invalid pool ID.  As long as the caller is checking it first, this is fine.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+		return nil
+	}
+	return bucket.Get(uint32ToBytes(uint32(index)))
+}
+
+// getMaxUsedIdx returns the highest used index from the used addresses bucket
+// of the given pool, series and branch.
+func getMaxUsedIdx(tx walletdb.Tx, poolID []byte, seriesID uint32, branch Branch) (Index, error) {
+	maxIdx := Index(0)
+	usedAddrs := tx.RootBucket().Bucket(poolID).Bucket(usedAddrsBucketName)
```
davecgh: Panic with invalid pool ID.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+	bucket := usedAddrs.Bucket(getUsedAddrBucketID(seriesID, branch))
+	if bucket == nil {
+		return maxIdx, nil
+	}
+	err := bucket.ForEach(
```
davecgh: This will have severe scalability issues searching all of the keys each time.  Imagine when you start to have a few million addresses.

What do you think about keeping a key in a metadata bucket which is updated as necessary? I think that would be the most efficient.

Alternatively, if you don't want to keep a field updated, perhaps a large gap linear forward search + binary backwards search?  For example, check for 1000000, 2000000, 3000000, .... until it doesn't exist, and then use a binary search to find the max.  With this type of scheme, having say 50 million addresses would result in a worst case scenario of testing for 50 + log_2(50,000,000) = 70 keys instead of 50 million.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+		func(k, v []byte) error {
+			idx := Index(bytesToUint32(k))
```
davecgh: I highly recommend this be made a `var idx [4]byte` outside of the loop and uses `binary.LittleEndian.PutUint32(idx[:], k)` inside the loop.  Otherwise, this is generating huge amounts of garbage, and, as you know, large amount of GC in Go causes significant slowdown.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+func putPool(tx walletdb.Tx, poolID []byte) error {
+	poolBucket, err := tx.RootBucket().CreateBucket(poolID)
+	if err != nil {
+		return newError(ErrDatabase, fmt.Sprintf("cannot create pool %v", poolID), err)
+	}
+	_, err = poolBucket.CreateBucket(seriesBucketName)
 	if err != nil {
-		str := fmt.Sprintf("cannot create voting pool %v", votingPoolID)
-		return managerError(waddrmgr.ErrDatabase, str, err)
+		return newError(
```
davecgh: Inconsistent line wrapping.

votingpool/db.go
```diff
@@ -56,13 +59,79 @@ type dbSeriesRow struct {
...
+			ErrDatabase, fmt.Sprintf("cannot create series bucket for pool %v", poolID), err)
+	}
+	_, err = poolBucket.CreateBucket(usedAddrsBucketName)
+	if err != nil {
+		return newError(
```
davecgh: Inconsistent line wrapping.

votingpool/db.go
```diff
@@ -70,7 +139,7 @@ func putPool(tx walletdb.Tx, votingPoolID []byte) error {
 // loadAllSeries returns a map of all the series stored inside a voting pool
 // bucket, keyed by id.
 func loadAllSeries(tx walletdb.Tx, votingPoolID []byte) (map[uint32]*dbSeriesRow, error) {
-	bucket := tx.RootBucket().Bucket(votingPoolID)
+	bucket := tx.RootBucket().Bucket(votingPoolID).Bucket(seriesBucketName)
```
davecgh: Panic on invalid pool ID.

votingpool/db.go
gsalgado: Yeah, I refrained from doing so in this branch because it's not really related to the changes here. Happy to do so here if you won't mind the extra diff pollution when reviewing.

votingpool/db.go
gsalgado: I did that here because I like to see all arguments in a single line, but when that's not possible I'll indeed use as much as possible of the first line before wrapping. I don't feel strong about it, though, so I'm happy to consistently use as much as possible of the first line if you think it's worth it

votingpool/db.go
gsalgado: It should be called only a couple times for a getdepositaddress RPC call, but it may be called lots of times for a startwithdrawal, so I think it's worth changing it as you suggest

votingpool/db.go
gsalgado: These functions are only called from Pool methods, and a Pool is obtained by calling either pool.Load or pool.Create, both of which ensure there's a bucket for that pool. I think that's probably enough?

votingpool/db.go
gsalgado: This one is also called from Pool methods

votingpool/db.go
gsalgado: Yeah, I knew this would not scale but decided not to worry about it for now since it's used only when processing pool withdrawals, and those are not supposed to happen too frequently as the notaries (OT servers) will batch user withdrawal requests and go through a consensus round to agree on the withdrawal parameters to pass to the wallet. If it's ok with you I'd prefer to add a FIXME comment here for now, mentioning the two alternatives you suggest on how to improve it. I'll run some benchmarks to see how it performs with lots of addresses, to give us an idea of how bad the current implementation performs

gsalgado: So, I ran some benchmarks and these are some numbers I got: ~27ms for 500K entries, ~60ms for 1M entries, and ~280ms for 5M entries.

davecgh: A FIXME comment sounds fine.

votingpool/db.go
davecgh: Yep.  I mentioned it in another comment that I haven't reviewed all the call sites yet.  It sounds like it's covered then.

votingpool/db.go
```diff
@@ -70,7 +139,7 @@ func putPool(tx walletdb.Tx, votingPoolID []byte) error {
 // loadAllSeries returns a map of all the series stored inside a voting pool
 // bucket, keyed by id.
 func loadAllSeries(tx walletdb.Tx, votingPoolID []byte) (map[uint32]*dbSeriesRow, error) {
```
davecgh: I know this already exists, but just adding this here for discussion.  The same thing is being done over in `waddrmgr` where there is a load all type of function.  Unfortunately, this approach will have scalability issues.

We are considering adding a `Cursor` to walletdb potentially along with an `AddressIterator` to waddrmgr.  Do you have any suggestions and or issues in regards to this?  I think it would definitely apply here as I imagine the number of series can get huge over time.

votingpool/db.go
davecgh: Certainly faster than i thought it would be, but I guess it's not too surprising given the fact that bolt is all zero copy.

davecgh:
I still have a lot of review left to do on the code, but I noticed this:

$ golint
pool.go:361:1: exported method Pool.ActivateSeries should have comment or be unexported

votingpool/db.go
davecgh: After discussing this some more, see #215 for details on the planned direction to resole this in waddrmgr.  I suspect a similar approach here wold work well.  We're also adding a `Cursor` to walletdb, so depending on how the keys are actually stored and the need to do scans, there might be further optimization potential there.  Given the current code though, I suspect an iterator-style callback will work well here too.

votingpool/db.go
```diff
@@ -116,17 +185,18 @@ func putSeriesRow(tx walletdb.Tx, votingPoolID []byte, ID uint32, row *dbSeriesR
...
 		str := fmt.Sprintf("cannot create bucket %v", votingPoolID)
-		return managerError(waddrmgr.ErrDatabase, str, err)
+		return newError(ErrDatabase, str, err)
 	}
+	bucket = bucket.Bucket(seriesBucketName)
 	serialized, err := serializeSeriesRow(row)
 	if err != nil {
 		str := fmt.Sprintf("cannot serialize series %v", row)
-		return managerError(waddrmgr.ErrSeriesStorage, str, err)
+		return newError(ErrSeriesStorage, str, err)
```
davecgh: I noticed that some of the `ErrSeriesStorage` errors have an underlying error and some don't.  This seems a bit inconsistent with the rest of the code where `ErrDatabase` always has an underlying error and that others don't.

I suppose this is fine as long as it's documented, but it is a departure from the rest of the code.

votingpool/error.go
```diff
@@ -0,0 +1,211 @@
...
+
+// ErrorCode identifies a kind of error
+type ErrorCode int
+
+const (
+	// ErrInputSelection indicates an error in the input selection
+	// algorithm.
+	ErrInputSelection ErrorCode = iota
+
+	// ErrWithdrawalProcessing indicates an internal error when processing a withdrawal
```
davecgh: exceeds 80

votingpool/error.go
```diff
@@ -0,0 +1,211 @@
...
+	// request.
+	ErrWithdrawalProcessing
+
+	// ErrUnknownPubKey indicates a pubkey that does not belong to a given series.
```
davecgh: exceeds 80

votingpool/error.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// ErrSeriesVersion indicates that we've been asked to deal with
+	// a series whose version is unsupported
+	ErrSeriesVersion
+
+	// ErrSeriesNotExists indicates that an attempt has been made to access
+	// a series that does not exist.
+	ErrSeriesNotExists
+
+	// ErrSeriesAlreadyExists indicates that an attempt has been made to create
```
davecgh: exceeds 80

votingpool/error.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// ErrPoolAlreadyExists indicates that an attempt has been made to
+	// create a voting pool that already exists.
+	ErrPoolAlreadyExists
+
+	// ErrPoolNotExists indicates that an attempt has been made to access
+	// a voting pool that does not exist.
+	ErrPoolNotExists
+
+	// ErrScriptCreation indicates that the creation of a deposit script failed.
```
davecgh: exceeds 80

votingpool/error.go
```diff
@@ -0,0 +1,211 @@
...
+	ErrRawSigning
+
+	// ErrPreconditionNotMet indicates a programming error since a
+	// preconditon has not been met.
+	ErrPreconditionNotMet
+
+	// ErrTxSigning indicates an error when signing a transaction.
+	ErrTxSigning
+
+	// ErrSeriesIDNotSequential indicates an attempt to create a series with an
```
davecgh: exceeds 80

davecgh:
I think `error_wb_test.go` should just be named `error_test.go` since the errors apply to the entire package.

votingpool/error_wb_test.go
```diff
@@ -0,0 +1,76 @@
...
+		{ErrInvalidBranch, "ErrInvalidBranch"},
+		{ErrInvalidValue, "ErrInvalidValue"},
+		{ErrDatabase, "ErrDatabase"},
+		{ErrKeyChain, "ErrKeyChain"},
+		{ErrCrypto, "ErrCrypto"},
+		{ErrRawSigning, "ErrRawSigning"},
+		{ErrPreconditionNotMet, "ErrPreconditionNotMet"},
+		{ErrTxSigning, "ErrTxSigning"},
+		{ErrInvalidScriptHash, "ErrInvalidScriptHash"},
+		{ErrWithdrawFromUnusedAddr, "ErrWithdrawFromUnusedAddr"},
```
davecgh: {0xffff, "Unknown ErrorCode (65535)"},

votingpool/error_wb_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	if int(lastErr) != len(tests) {
```
davecgh: With the aforementioned addition, this would need to be `len(tests) - 1` to adjust.

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+	pubPassphrase  = []byte("pubPassphrase")
+	privPassphrase = []byte("privPassphrase")
+)
+
+func Example_depositAddress() {
+	// This example demonstrates how to create a voting pool with one
+	// series and get a deposit address for that series.
+
+	// Create the address manager and votingpool DB namespace.
+	mgr, vpNamespace, tearDownFunc, err := exampleCreateMgrAndDBNamespace()
```
davecgh: There probably needs to be an example which does what this function is doing along with a comment stating as much since the called function won't show up in the example documentation.

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
-	// This example demonstrates how to create a voting pool, create a
-	// series, get a deposit address from a series and lastly how to
-	// replace a series.
+var (
+	pubPassphrase  = []byte("pubPassphrase")
+	privPassphrase = []byte("privPassphrase")
+)
+
+func Example_depositAddress() {
+	// This example demonstrates how to create a voting pool with one
```
davecgh: The comment needs to be just above the function so it shows up properly in the generated godoc.

```Go
// This example ...
// ... for that series.
func Example_depositAddress() {
```

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+		return
+	}
+	fmt.Println("Generated deposit address:", addr.EncodeAddress())
+
+	// Output:
+	// Generated deposit address: 3QTzpc9d3tTbNLJLB7xwt87nWM38boAhAw
+}
+
+func Example_empowerSeries() {
+	// This example demonstrates how to empower a series by loading the private
```
davecgh: Same here on the comment.

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+		fmt.Printf("Cannot empower series: %v\n", err)
+		return
+	}
+
+	// Output:
+	//
+}
+
+func Example_startWithdrawal() {
+	// This example demonstrates how to empower a series by loading the private
```
davecgh: Same here on the comment.

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+
+	// Create the address manager and votingpool DB namespace.
+	mgr, vpNamespace, tearDownFunc, err := exampleCreateMgrAndDBNamespace()
+	if err != nil {
+		return
+	}
+	defer tearDownFunc()
 
+	// Create a pool and a series.
+	pool, seriesID, err := exampleCreatePoolAndSeries(mgr, vpNamespace)
```
davecgh: Probably worthwhile to call out that the called function here is doing the same thing demonstrated in the example for DepositAddress since the person reading the documentation can't see what this function is doing (aside from inferring it from the name).

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+}
+
+func Example_empowerSeries() {
+	// This example demonstrates how to empower a series by loading the private
+	// key for one of the series' public keys.
+
+	// Create the address manager and votingpool DB namespace.
+	mgr, vpNamespace, tearDownFunc, err := exampleCreateMgrAndDBNamespace()
+	if err != nil {
+		return
```
davecgh: I think you need to print the error here.  Otherwise any errors will result in no output, which is what the example output expects meaning it would pass the tests as if the example worked even though it didn't.

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+	}
+	defer tearDownFunc()
 
+	// Create a pool and a series.
+	pool, seriesID, err := exampleCreatePoolAndSeries(mgr, vpNamespace)
+	if err != nil {
+		return
```
davecgh: Same deal with printing the error here.

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+}
+
+func Example_startWithdrawal() {
+	// This example demonstrates how to empower a series by loading the private
+	// key for one of the series' public keys.
+
+	// Create the address manager and votingpool DB namespace.
+	mgr, vpNamespace, tearDownFunc, err := exampleCreateMgrAndDBNamespace()
+	if err != nil {
+		return
```
davecgh: And here.

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+	}
+	defer tearDownFunc()
+
+	// Create a pool and a series.
+	pool, seriesID, err := exampleCreatePoolAndSeries(mgr, vpNamespace)
+	if err != nil {
+		return
```
davecgh: Print error

votingpool/example_test.go
```diff
@@ -17,109 +17,262 @@
...
+		fmt.Println(err)
+		return
+	}
+	lastSeriesID := seriesID
+	dustThreshold := btcutil.Amount(1e4)
+	currentBlock := int32(19432)
+	roundID := uint32(0)
+	txstore, tearDownFunc, err := exampleCreateTxStore()
+	if err != nil {
+		return
```
davecgh: Print error

votingpool/db.go
gsalgado: Actually here it shouldn't be a problem because the plan is to have only one series per week. Having said that, I think we should get rid of loadAllSeries completely because we have nothing that operates on the full set of series. Right now loadAllSeries() populates .seriesLookup which is then used by Pool.GetSeries() but we could just make it load the desired series directly from the DB, and get rid of both loadAllSeries() and seriesLookup.

gsalgado: The only downside I can think of is that GetSeries() would then require the manager to be unlocked. I've added a FIXME here too, anyway

votingpool/db.go
gsalgado: Yeah, that's because we're returning ErrSeriesStorage when we get a bolt error (in which case we use that as the underlying error) but also when we fail to serialize the series data (in which case we have no underlying error). Maybe we should use it only for the latter (and possibly rename it to ErrSeriesSerialization), and use ErrDatabase for the former?

gsalgado:
It was called error_wb_test.go because it was a written as a whitebox test so that it could access the private lastErr variable. I've renamed it and exported that variable via internal_test.go, though, so now it's called error_test.go and is on the votingpool_test package.

votingpool/example_test.go
gsalgado: Yeah, good point. I've added an example for the Create function (ExampleCreate) and changed the comment here to refer to that.

votingpool/example_test.go
gsalgado: All example* functions did that, but it's probably better if they just return the error and the callsites print them -- makes for more legible examples. I'll change them

votingpool/factory_test.go
```diff
@@ -0,0 +1,384 @@
...
+	for _, amt := range amts {
+		msgtx.AddTxOut(wire.NewTxOut(amt, pkScript))
+	}
+	return msgtx
+}
+
+func TstNewDepositScript(t *testing.T, p *Pool, seriesID uint32, branch Branch, idx Index) []byte {
+	script, err := p.DepositScript(seriesID, branch, idx)
+	if err != nil {
+		t.Fatalf("Failed to create depositscript for series %d, branch %d, index %d: %v",
```
davecgh: Should deposit script have a space?

votingpool/factory_test.go
```diff
@@ -0,0 +1,384 @@
...
+	}
+	return credits
+}
+
+// TstCreatePool creates a Pool on a fresh walletdb and returns it. It also
+// returns the pool's waddrmgr.Manager (which uses the same walletdb, but with a
+// different namespace) as a convenience, and a teardown function that closes
+// the Manager and removes the directory used to store the database.
+func TstCreatePool(t *testing.T) (tearDownFunc func(), mgr *waddrmgr.Manager, pool *Pool) {
+	// XXX: This should be moved somewhere else eventually as not all of our
```
davecgh: Reminder about this.

gsalgado: I can't think of a decent alternative to having the t.Parallel() call here, though. I'd rather avoid repeating it on every test, but having it here is not ideal either. Any suggestions?

votingpool/db.go
davecgh: I'm ok with either as long as it's documented.  I do have a very slight preference for the latter.

votingpool/factory_test.go
```diff
@@ -0,0 +1,384 @@
...
+	return OutputRequest{
+		PkScript:    pkScript,
+		Address:     addr,
+		Amount:      amount,
+		Server:      "server",
+		Transaction: transaction,
+	}
+}
+
+func TstNewWithdrawalOutput(
```
davecgh: Inconsistent line wrapping.

votingpool/factory_test.go
```diff
@@ -0,0 +1,384 @@
...
+	r OutputRequest, status outputStatus, outpoints []OutBailmentOutpoint) *WithdrawalOutput {
+	output := &WithdrawalOutput{
+		request:   r,
+		status:    status,
+		outpoints: outpoints,
+	}
+	return output
+}
+
+func TstNewWithdrawalAddress(
```
davecgh: Same.

votingpool/input_selection.go
```diff
@@ -0,0 +1,286 @@
...
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcwallet/txstore"
+)
+
+const eligibleInputMinConfirmations = 100
+
+// Credit is an abstraction over txstore.Credit used in the construction of
+// voting pool withdrawal transactions.
+type Credit interface {
+	fmt.Stringer
```
davecgh: Does the Credit interface actually require a stringer?  It can implement the Stringer interface without actually requiring it in the Credit interface.

jrick: Methods can't be defined on interface types, you have to compose them.

votingpool/db.go
gsalgado: Ok, I went with the latter.

votingpool/input_selection.go
davecgh: Yes, but that's not what I'm saying here.  I see no reason for the `Credit` interface to _require_ a stringer implementation.  The `type credit struct` which implements the `Credit` interface can also implement the `Stringer` interface without the `Credit` interface _requiring_ it.

gsalgado: Yeah, I can't think of a reason why the Credit interface should require a Stringer implementation. Removed

votingpool/input_selection.go
```diff
@@ -0,0 +1,286 @@
...
+// Less returns true if the element at positions i is smaller than the
+// element at position j. The 'smaller-than' relation is defined to be
+// the lexicographic ordering defined on the tuple (SeriesID, Index,
+// Branch, TxSha, OutputIndex).
+func (c byAddress) Less(i, j int) bool {
+	if c[i].Address().seriesID < c[j].Address().seriesID {
+		return true
+	}
+
+	if c[i].Address().seriesID == c[j].Address().seriesID &&
```
davecgh: Super minor thing, but I wonder if rather than rechecking each of these conditions multiple times, if it would be better to do something like:

```Go
if c[i].Address().seriesID < c[j].Address().seriesID {
	return true
}
if c[i].Address().seriesID > c[j].Address().seriesID {
	return false
}

// The seriesID are equal, so compare index.
if c[i].Address().index < c[j].Address().index {
	return true
}
if c[i].Address().index > c[j].Address().index {
	return false
}

// The seriesID and index are equal, so compare branch.
if c[i].Address().branch < c[j].Address().branch {
	return true
}
if c[i].Address().branch > c[j].Address().branch {
	return false
}

// The seriesID, index, and branch are equal, so compare hash.
txidComparison := bytes.Compare(c[i].TxSha().Bytes(), c[j].TxSha().Bytes())
if txidComparison < 0 {
	return true
}
if txidComparison > 0 {
	return false
}

// The seriesID, index, branch, and hash are equal, so compare index.
if c[i].OutputIndex() < c[j].OutputIndex() {
	return true
}

return false
```

davecgh: Also, using locals for the .Address() would avoid a lot of bounds checking.

Something like:
```Go
iAddress := c[i].Address()
jAddress := c[j].Address()

if iAddress.seriesID < jAddress.seriesID {
	return true
}
.....
```

votingpool/input_selection.go
```diff
@@ -0,0 +1,286 @@
...
+		txidComparison == 0 &&
+		c[i].OutputIndex() < c[j].OutputIndex() {
+		return true
+	}
+	return false
+}
+
+// getEligibleInputs returns eligible inputs with addresses between startAddress
+// and the last used address of lastSeriesID.
+func (p *Pool) getEligibleInputs(
```
davecgh: Inconsistent argument formatting

votingpool/input_selection.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return inputs, nil
+}
+
+// nextAddr returns the next WithdrawalAddress according to the input selection
+// rules: http://opentransactions.org/wiki/index.php/Input_Selection_Algorithm_(voting_pools)
+// It returns nil if the new address' seriesID is >= stopSeriesID.
+func nextAddr(p *Pool, seriesID uint32, branch Branch, index Index, stopSeriesID uint32) (
+	*WithdrawalAddress, error) {
+	branch++
```
davecgh: Probably doesn't matter much sense it's a local, but shouldn't this mutation happen after the attempt to get the series?  That is to say, if the series doesn't exist, there is no reason to increment the branch.

votingpool/input_selection.go
```diff
@@ -0,0 +1,286 @@
...
+		}
+		branch = 0
+	}
+
+	if seriesID >= stopSeriesID {
+		return nil, nil
+	}
+
+	addr, err := p.WithdrawalAddress(seriesID, branch, index)
+	if err != nil && err.(Error).ErrorCode == ErrWithdrawFromUnusedAddr {
```
davecgh: As long as WithdrawalAddress guarantees a non-nil error will always be a `votingpool.Error` this is ok, but it will panic should any other be returned.

votingpool/input_selection.go
```diff
@@ -0,0 +1,286 @@
...
+	}
+	return addr, err
+}
+
+// highestUsedSeriesIndex returns the highest index among all of this Pool's
+// used addresses for the given seriesID. It returns 0 if there are no used
+// addresses with the given seriesID.
+func (p *Pool) highestUsedSeriesIndex(seriesID uint32) (Index, error) {
+	maxIdx := Index(0)
+	series := p.GetSeries(seriesID)
```
davecgh: I've noticed this throughout the code, but curious is there is a specific reason, such as a name collision, for why it's named `GetSeries` instead of `Series` since the former is against Effective Go style guidelines?

votingpool/input_selection.go
```diff
@@ -0,0 +1,286 @@
...
+			maxIdx = idx
+		}
+	}
+	return maxIdx, nil
+}
+
+// groupCreditsByAddr converts a slice of credits to a map from the string
+// representation of an encoded address to the unspent outputs associated with
+// that address.
+func groupCreditsByAddr(
```
davecgh: Inconsistent argument formatting.

votingpool/input_selection_wb_test.go
```diff
@@ -0,0 +1,403 @@
...
+	c2 := TstCreateInputs(t, store2, pkScript, []int64{int64(dustThreshold)})[0]
+	// the calculation of if it has been confirmed does this:
+	// chainheigt - bh + 1 >= target, which is quite weird, but the
+	// reason why I need to put 902 as *that* makes 1000 - 902 +1 = 99 >=
+	// 100 false
+	c2.BlockHeight = int32(902)
+	if pool.isCreditEligible(c2, eligibleInputMinConfirmations, chainHeight, dustThreshold) {
+		t.Errorf("Input is eligible and it should not be.")
+	}
+
```
davecgh: Extra newline.

votingpool/input_selection_wb_test.go
```diff
@@ -0,0 +1,403 @@
...
+func (c *TstFakeCredit) Address() WithdrawalAddress { return c.addr }
+func (c *TstFakeCredit) Amount() btcutil.Amount     { return c.amount }
+func (c *TstFakeCredit) TxOut() *wire.TxOut         { return nil }
+func (c *TstFakeCredit) OutPoint() *wire.OutPoint {
+	return &wire.OutPoint{Hash: *c.txSha, Index: c.outputIndex}
+}
+
+func TstNewFakeCredit(t *testing.T, pool *Pool, series uint32, index Index, branch Branch, txSha []byte, outputIdx int) *TstFakeCredit {
+	var hash wire.ShaHash
+	copy(hash[:], txSha)
```
davecgh: Probably not really a big deal since it's unlikely to change anytime soon, but this is assuming the structure of the ShaHash type.  The more correct way to do this would be to call `hash.SetBytes(txsha)` which can throw an error if the format of the bytes were to ever change.

votingpool/input_selection_wb_test.go
```diff
@@ -0,0 +1,403 @@
...
+	var pkScripts [][]byte
+	for idx := startIdx; idx <= stopIdx; idx++ {
+		for branch := startBranch; branch <= stopBranch; branch++ {
+			pkScripts = append(pkScripts, TstCreatePkScript(t, pool, seriesID, branch, idx))
+		}
+	}
+	return pkScripts
+}
+
+func checkWithdrawalAddressMatches(
```
davecgh: Inconsistent argument formatting.

votingpool/pool.go
```diff
@@ -32,6 +32,15 @@ const (
 	minSeriesPubKeys = 3
 )
 
+// CurrentVersion is the version used for newly created Series.
+var CurrentVersion uint32 = 1
```
davecgh: This should probably be an untyped constant.

votingpool/pool.go
```diff
@@ -55,6 +64,37 @@ type Pool struct {
 	namespace    walletdb.Namespace
 }
 
+// PoolAddress represents a voting pool P2SH address, generated by
+// deriving public HD keys from the series' master keys using the given
+// branch/index and constructing a M-of-N multi-sig script.
+type PoolAddress interface {
+	fmt.Stringer
```
davecgh: Same deal with the required `Stringer` here as for the `Series` interface.  I don't think it should be required by the `PoolAddress` interface.

votingpool/pool.go
```diff
@@ -121,11 +161,11 @@ func LoadAndGetDepositScript(namespace walletdb.Namespace, m *waddrmgr.Manager,
 func LoadAndCreateSeries(namespace walletdb.Namespace, m *waddrmgr.Manager, version uint32,
 	poolID string, seriesID, reqSigs uint32, rawPubKeys []string) error {
 	pid := []byte(poolID)
-	vp, err := Load(namespace, m, pid)
+	p, err := Load(namespace, m, pid)
 	if err != nil {
-		managerErr := err.(waddrmgr.ManagerError)
-		if managerErr.ErrorCode == waddrmgr.ErrVotingPoolNotExists {
-			vp, err = Create(namespace, m, pid)
+		vpErr := err.(Error)
```
davecgh: As previously mentioned, this will panic if the error from Load is not a `votinpool.Error`.  Need to be sure that it only ever returns that.

votingpool/pool.go
```diff
@@ -162,25 +202,32 @@ func LoadAndEmpowerSeries(namespace walletdb.Namespace, m *waddrmgr.Manager,
 
 // GetSeries returns the series with the given ID, or nil if it doesn't
 // exist.
-func (vp *Pool) GetSeries(seriesID uint32) *SeriesData {
-	series, exists := vp.seriesLookup[seriesID]
+func (p *Pool) GetSeries(seriesID uint32) *SeriesData {
```
davecgh: As pointed out in a previous comment, this should be `Series` instead of `GetSeries` per the Effective Go guidelines.

votingpool/pool.go
```diff
@@ -290,47 +339,75 @@ func (vp *Pool) putSeries(version, seriesID, reqSigs uint32, inRawPubKeys []stri
 //
 // - rawPubKeys has to contain three or more public keys;
 // - reqSigs has to be less or equal than the number of public keys in rawPubKeys.
-func (vp *Pool) CreateSeries(version, seriesID, reqSigs uint32, rawPubKeys []string) error {
-	if series := vp.GetSeries(seriesID); series != nil {
+// FIXME: The minimum allowed value for seriesID should be 1:
```
davecgh: This should probably be done before it goes in.  It seems pretty straight forward.

votingpool/input_selection.go
gsalgado: Yeah, I like your version more. Used it almost verbatim, except that I replaced the last if block and the last return with a "return c[i].OutputIndex() < c[j].OutputIndex()"

votingpool/input_selection.go
gsalgado: Yeah, no reason to do that before checking that the series exists.

votingpool/input_selection.go
gsalgado: Yes, all public methods/functions in the votingpool package should return only errors of that type.

votingpool/input_selection.go
gsalgado: I thought that applied only to getters for struct fields, where no arguments are needed. I'll change it to Series()

votingpool/input_selection_wb_test.go
gsalgado: Ok, I've changed it but that required me to change TestCreditSortingByAddress because it was passing txSha values too small to this function.

votingpool/pool.go
gsalgado: Yep, it should return only errors of that type

votingpool/pool.go
gsalgado: Yeah, it is straightforward but it breaks lots of tests that use seriesID==0, and I didn't want to add even more unrelated changes to this PR, but now it's done :)

votingpool/input_selection.go
davecgh: Cool.  I like the change on the last if block too.

davecgh:
Thanks for the quick changes.  I'm a little over halfway done on the review and still reviewing.  Taking a while since I'm working it in between other things.

votingpool/factory_test.go
davecgh: Short of using the new test setup function available in Go 1.4 (which we can't do here since we still support Go 1.3), the only choice is to put it in each Test (or in some common function that each test calls).

votingpool/pool.go
```diff
@@ -387,11 +467,15 @@ func validateAndDecryptKeys(rawPubKeys, rawPrivKeys [][]byte, vp *Pool) (pubKeys
 // seriesLookup map with them. If there are any private extended keys for
 // a series, it will also ensure they have a matching extended public key
 // in that series.
-func (vp *Pool) LoadAllSeries() error {
+//
+// This method must be called with the Pool's manager unlocked.
+// FIXME: We should be able to get rid of this (and loadAllSeries/seriesLookup)
```
davecgh: Reminder about this unless you are you planning to leave it for later.

davecgh:
I noticed that this changes the branch and index parameters over to use the specific types `Branch` and `Index`, respectively.  Are there future plans to expose methods on them?

I'm asking because I generally prefer to avoid creating types just for the sake of doing so since it hides the underlying type making it more difficult to reason about.  It also tends to lead to a bunch of casts in the code.

Of course, when the type provides methods which expose additional functionality, it makes sense, but I don't see these particular types doing anything here except acting a stand-ins for uint32.

votingpool/pool.go
```diff
@@ -502,68 +569,138 @@ func (vp *Pool) DepositScript(seriesID, branch, index uint32) ([]byte, error) {
...
 	return script, nil
 }
 
+// ChangeAddress returns a new votingpool address for the given seriesID and
+// index, on the 0th branch (which is reserved for change addresses). The series
+// with the given ID must be active.
+func (p *Pool) ChangeAddress(seriesID uint32, index Index) (*ChangeAddress, error) {
+	series := p.Series(seriesID)
+	if series == nil {
+		return nil, newError(
```
davecgh: Argument formatting.

votingpool/pool.go
```diff
@@ -502,68 +569,138 @@ func (vp *Pool) DepositScript(seriesID, branch, index uint32) ([]byte, error) {
...
+}
+
+// WithdrawalAddress queries the address manager for the P2SH address
+// of the redeem script generated with the given series/branch/index and uses
+// that to populate the returned WithdrawalAddress. This is done because we
+// should only withdraw from previously used addresses but also because when
+// processing withdrawals we may iterate over a huge number of addresses and
+// it'd be too expensive to re-generate the redeem script for all of them.
+// This method must be called with the manager unlocked.
+func (p *Pool) WithdrawalAddress(
```
davecgh: Argument formatting.

votingpool/pool.go
```diff
@@ -502,68 +569,138 @@ func (vp *Pool) DepositScript(seriesID, branch, index uint32) ([]byte, error) {
...
+	seriesID uint32, branch Branch, index Index) (*WithdrawalAddress, error) {
+	// TODO: Ensure the given series is hot.
```
davecgh: Can this be done for this PR, or is there missing infrastructure?

votingpool/pool.go
```diff
@@ -502,68 +569,138 @@ func (vp *Pool) DepositScript(seriesID, branch, index uint32) ([]byte, error) {
...
+		return nil, err
+	}
+	pAddr, err := p.poolAddress(seriesID, branch, index, script)
+	if err != nil {
+		return nil, err
+	}
+	return &WithdrawalAddress{poolAddress: pAddr}, nil
+}
+
+func (p *Pool) poolAddress(
```
davecgh: args

votingpool/pool.go
```diff
@@ -581,18 +718,153 @@ func (vp *Pool) EmpowerSeries(seriesID uint32, rawPrivKey string) error {
...
+	return maxIdx, err
+}
+
+// String returns a string encoding of the underlying bitcoin payment address.
+func (a *poolAddress) String() string {
+	return a.addr.EncodeAddress()
+}
+
+func (a *poolAddress) addrIdentifier() string {
+	return fmt.Sprintf(
```
davecgh: Inconsistent args formatting.

votingpool/pool.go
```diff
@@ -604,8 +876,12 @@ func (s *SeriesData) IsEmpowered() bool {
...
-// XXX(lars): We should probably make our own votingpoolError function.
-func managerError(c waddrmgr.ErrorCode, desc string, err error) waddrmgr.ManagerError {
-	return waddrmgr.ManagerError{ErrorCode: c, Description: desc, Err: err}
+func (s *SeriesData) getPrivKeyFor(pubKey *hdkeychain.ExtendedKey) (*hdkeychain.ExtendedKey, error) {
+	for i, key := range s.publicKeys {
+		if key.String() == pubKey.String() {
+			return s.privateKeys[i], nil
+		}
+	}
+	return nil, newError(
```
davecgh: Inconsistent args formatting.

votingpool/pool_test.go
```diff
@@ -393,237 +271,118 @@ func TestCreateSeries(t *testing.T) {
...
+	vp.TstRunWithManagerUnlocked(t, mgr, func() {
+		if err := pool.EmpowerSeries(seriesID, vp.TstPrivKeys[1]); err != nil {
+			t.Fatalf("Failed to empower series: %v", err)
+		}
+	})
 
-	if err := pool.EmpowerSeries(seriesID, privKey1); err != nil {
-		t.Fatalf("Failed to empower series: %v", err)
-	}
+	err := pool.ReplaceSeries(
```
davecgh: Arg formatting.

votingpool/pool_test.go
```diff
@@ -636,18 +395,17 @@ var replaceSeriesTestData = []replaceSeriesTestEntry{
 	{
 		testID: 0,
 		orig: seriesRaw{
-			id:      0,
+			id:      1,
 			version: 1,
 			reqSigs: 2,
-			pubKeys: votingpool.CanonicalKeyOrder(
```
davecgh: Arg formatting.

votingpool/pool_test.go
```diff
@@ -636,18 +395,17 @@ var replaceSeriesTestData = []replaceSeriesTestEntry{
...
-				[]string{pubKey0, pubKey1, pubKey2, pubKey4}),
+			pubKeys: vp.CanonicalKeyOrder(
```
davecgh: Arg formatting.

votingpool/pool_test.go
```diff
@@ -1148,238 +882,63 @@ func TestDecryptExtendedKeyCannotCreateResultKey(t *testing.T) {
...
-		}
-
-		_, err = votingpool.SerializeSeries(
-			test.version, active, test.reqSigs, encryptedPubs, encryptedPrivs)
-
-		checkManagerError(t, fmt.Sprintf("Test #%d", testNum), err, test.err)
-	}
+	// When the series is not active, we should get an error.
+	pubKeys = vp.TstPubKeys[3:6]
+	vp.TstCreateSeries(
```
davecgh: args

votingpool/pool_test.go
```diff
@@ -1148,238 +882,63 @@ func TestDecryptExtendedKeyCannotCreateResultKey(t *testing.T) {
...
+	// When the requested address is not present in the set of used addresses
+	// for that Pool, we should get an error.
+	_, err := pool.WithdrawalAddress(1, 2, 3)
+	vp.TstCheckError(t, "", err, vp.ErrWithdrawFromUnusedAddr)
 }
 
-func TestDeserializationErrors(t *testing.T) {
-	tearDown, _, _ := setUp(t)
-	defer tearDown()
+func checkPoolAddress(
```
davecgh: Arg formatting.

votingpool/pool_wb_test.go
```diff
@@ -0,0 +1,462 @@
...
+		if err != nil {
+			t.Fatalf("Test #%d - Error encrypting privkeys: %v", testNum, err)
+		}
+
+		row := &dbSeriesRow{
+			version:           test.version,
+			active:            test.active,
+			reqSigs:           test.reqSigs,
+			pubKeysEncrypted:  encryptedPubs,
+			privKeysEncrypted: encryptedPrivs}
```
davecgh: Nit: The closing `}` should be on the next line for consistency.

votingpool/pool_wb_test.go
```diff
@@ -0,0 +1,462 @@
...
+	tearDown, _, _ := TstCreatePool(t)
+	defer tearDown()
+
+	tests := []struct {
+		serialized []byte
+		err        ErrorCode
+	}{
+		{
+			serialized: make([]byte, 1000000),
+			// Too many bytes (over waddrmgr.seriesMaxSerial).
```
davecgh: This comment is no longer accurate.

votingpool/pool_wb_test.go
```diff
@@ -0,0 +1,462 @@
...
+			err: ErrSeriesSerialization,
+		},
+		{
+			serialized: make([]byte, 10),
+			// Not enough bytes (under waddrmgr.seriesMinSerial).
```
davecgh: Same here.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcwallet/txstore"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/fastsha256"
+)
+
+// Maximum tx size (in bytes). This should be the same as bitcoind's
+// MAX_STANDARD_TX_SIZE.
+const txMaxSize = 100000
```
davecgh: I've mentioned it a few times, and it's nothing for you to worry about for this PR, but this is another example of why btcd should provide a policy package with these constants exported.  Cases like this are likely to be forgotten when/if the max standard tx size changes.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+
+// feeIncrement is the minimum transation fee (0.00001 BTC, measured in satoshis)
+// added to transactions requiring a fee.
+const feeIncrement = 1e3
```
davecgh: Same here.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+func (s byOutBailmentID) Len() int      { return len(s) }
+func (s byOutBailmentID) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
+func (s byOutBailmentID) Less(i, j int) bool {
+	return bytes.Compare(s[i].outBailmentIDHash(), s[j].outBailmentIDHash()) < 0
+}
+
+func (s outputStatus) String() string {
+	strings := map[outputStatus]string{
+		statusSuccess: "success",
+		statusPartial: "partial-",
```
davecgh: Is this `-` here intentional?  Doesn't seem consistent with success/split.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+
+// Fees returns the total amount of network fees included in all transactions
+// generated as part of a withdrawal.
+func (s *WithdrawalStatus) Fees() btcutil.Amount {
+	return s.fees
+}
+
+// NextInputAddr returns the votingpool address that should be used as the
+// startAddress of subsequent withdrawals.
+func (s *WithdrawalStatus) NextInputAddr() WithdrawalAddress {
```
davecgh: Shouldn't this function be calculating/setting the next one as well so subsequent calls will continue getting the next one?

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+	return s.nextInputAddr
+}
+
+// NextChangeAddr returns the votingpool address that should be used as the
+// changeStart of subsequent withdrawals.
+func (s *WithdrawalStatus) NextChangeAddr() ChangeAddress {
```
davecgh: Same deal here.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+	return OutBailmentID(fmt.Sprintf("%s:%d", r.Server, r.Transaction))
+}
+
+// outBailmentIDHash returns a byte slice which is used when sorting
+// OutputRequests.
+func (r OutputRequest) outBailmentIDHash() []byte {
+	if r.cachedHash != nil {
+		return r.cachedHash
+	}
+	str := fmt.Sprintf("%s%d", r.Server, r.Transaction)
```
davecgh: If this function is called frequently, I'd suggest doing:

```Go
str := r.Server + strconv.Itoa(int(r.Transaction))
```

The reason is that `fmt.Sprintf` uses reflection and is therefore something like 50-100x slower than using `strconv` in cases like this.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+type withdrawal struct {
+	roundID         uint32
+	status          *WithdrawalStatus
+	transactions    []*withdrawalTx
+	pendingRequests []OutputRequest
+	eligibleInputs  []Credit
+	current         *withdrawalTx
+	// newWithdrawalTx is a member of the structure so it can be replaced for
+	// testing purposes.
+	newWithdrawalTx func() *withdrawalTx
```
davecgh: Does this actually need to be unique for every withdrawal?  I'd think this should be package level instead of keeping an additional member for every withdrawal instance.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+// withdrawalTx represents a transaction constructed by the withdrawal process.
+type withdrawalTx struct {
+	inputs  []Credit
+	outputs []*withdrawalTxOut
+	fee     btcutil.Amount
+
+	// calculateFee calculates the expected network fees for this transaction.
+	// We use a func() field instead of a method so that it can be replaced in
+	// tests.
+	calculateFee func() btcutil.Amount
```
davecgh: Same question here regarding package level.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+
+	// isTooBig is a member of the structure so it can be replaced for testing
+	// purposes.
+	isTooBig func() bool
```
davecgh: Same question here regarding package level.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+
+	// calculateSize is a member of the structure so it can be replaced for testing
+	// purposes.
+	calculateSize func() int
```
davecgh: Same question here regarding package level.

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+// inputs.
+func (w *withdrawal) popInput() Credit {
+	input := w.eligibleInputs[0]
+	w.eligibleInputs = w.eligibleInputs[1:]
+	return input
+}
+
+// pushInput adds a new input to the top of the stack of eligible inputs.
+func (w *withdrawal) pushInput(input Credit) {
+	w.eligibleInputs = append([]Credit{input}, w.eligibleInputs...)
```
davecgh: I wonder if you should reverse the stack semantics here.  This will cause a lot of extra work and garbage because it will create a a new single element slice and append the rest of the stack to it, thereby potentially requiring several grows.  Each time a slice has to grow, it has to copy all elements.

So like:

```Go
w.eligibleInputs = append(w.eligibleInputs, input)
```

And then for `popInput`:

```Go
input := w.eligibleInputs[len(w.eligibleInputs)-1]
w.eligibleInputs = w.eligibleInputs[:len(w.eligibleInputs)-2]
return input
```

votingpool/withdrawal.go
```diff
@@ -0,0 +1,939 @@
...
+		n := nBytesToSerialize(redeemScriptLen)
+		sigScriptLen := 1 + (74 * int(addr.series().reqSigs)) + redeemScriptLen + 1 + n
+		txin.SignatureScript = bytes.Repeat([]byte{1}, sigScriptLen)
+	}
+	return msgtx.SerializeSize()
+}
+
+// nBytesToSerialize returns the number of bytes needed to serialize the given
+// value.
+func nBytesToSerialize(value int) int {
```
davecgh: I'd use `wire.VarIntSerializeSize` instead of a custom function.

votingpool/withdrawal_test.go
```diff
@@ -0,0 +1,133 @@
...
+	changeStart := vp.TstNewChangeAddress(t, pool, def.SeriesID, 0)
+
+	startAddr := vp.TstNewWithdrawalAddress(t, pool, def.SeriesID, 0, 0)
+	lastSeriesID := def.SeriesID
+	dustThreshold := btcutil.Amount(1e4)
+	currentBlock := int32(vp.TstInputsBlock + vp.TstEligibleInputMinConfirmations + 1)
+	var status *vp.WithdrawalStatus
+	var err error
+	vp.TstRunWithManagerUnlocked(t, mgr, func() {
+		status, err = pool.StartWithdrawal(
```
davecgh: Inconsistent arg formatting.

votingpool/withdrawal_wb_test.go
```diff
@@ -0,0 +1,1165 @@
...
+	// At this point we should have two finalized transactions.
+	if len(w.transactions) != 2 {
+		t.Fatalf("Wrong number of finalized transactions; got %d, want 2", len(w.transactions))
+	}
+
+	// First tx should have one output with amount of 1, the first input from
+	// the list of eligible inputs, and no change output.
+	firstTx := w.transactions[0]
+	req1 := requests[0]
+	checkTxOutputs(
```
davecgh: Inconsistent arg formatting.

votingpool/withdrawal_wb_test.go
```diff
@@ -0,0 +1,1165 @@
...
+		if input != inputs[i] {
+			t.Fatalf("Unexpected input; got %s, want %s", input, inputs[i])
+		}
+	}
+}
+
+// signTxAndValidate will construct the signature script for each input of the given
+// transaction (using the given raw signatures and the pkScripts from credits) and execute
+// those scripts to validate them.
+func signTxAndValidate(
```
davecgh: Inconsistent arg formatting.

davecgh:
@gsalgado I'm done with the review (phew that was a lot of code!).  Once everything is addresses, I'll get it merged.

votingpool/pool.go
gsalgado: This one was just added when addressing one of your comments here, so yes I'd rather leave it for later.

gsalgado:
The new Branch/Index types were indeed added just to make it harder to accidentally use one as the other, yes.

votingpool/pool.go
gsalgado: It's missing infrastructure, yes

votingpool/withdrawal.go
gsalgado: Yeah, it's defined that way in the spec because in a subsequent step these will be update to include the ID of the last series that need to be thawed in order to completely fulfil the request

votingpool/withdrawal.go
gsalgado: The next input addr is calculated once we finish processing the requests in a given startwithdrawal call. That information is used only in the response sent to the clients as they'll use it in subsequent startwithdrawal calls -- internally we have no need for it.

votingpool/withdrawal.go
gsalgado: That applies for the next change addr too

votingpool/withdrawal.go
gsalgado: Even though it's cached I made the change as it doesn't reduce readability IMO. Also good to know about fmt.Sprintf

votingpool/withdrawal.go
gsalgado: There was no real need for that, and after moving that and the others into package-level functions it turned out we no longer need to replace newWithdrawalTx() anymore as the tests can replace just the specific functions they care about (calculateFee, calculateSize or isTooBig)

votingpool/withdrawal.go
gsalgado: Would you be ok with saving this for later (leaving a TODO comment for now)? I ask because this would require changing getElibibleInputs(), the byAddress sorting and some test helpers, and most of their tests will break when I make those changes.

gsalgado:
Thanks a lot for the review, @davecgh! I've done most of the changes you suggested, except for a few ones where I either left a question or some other suggestion. I'll squash them all once everything is addressed.

votingpool/withdrawal.go
davecgh: Sure.  I'm ok with a TODO here since the package isn't actually being integrated yet.

votingpool/pool.go
davecgh: Looks like this one is still outstanding.

votingpool/withdrawal.go
davecgh: Alright, thanks for clarifying.

votingpool/factory_test.go
gsalgado: You mean TestMain()? I think that wouldn't work because in that function we don't have access to the testing.T instance. I've changed the comment from an XXX to a regular one since we don't seem to have any better alternatives at this point

gsalgado:
I think I've addressed all your comments now, @davecgh. Just let me know if you think it's ready to be squashed and I'll do it.

davecgh:
@gsalgado:  Looks good now.  Everything is addressed or has a TODO for it.

Let's squash it and I'll merge it.

gsalgado:
Awesome, all squashed and ready to go!

PR #181, opened 2015-01-31
Optimize unlocks for already unlocked addrmgrs.

jrick:
This is performed by saving the SHA512(salt+passphrase) of the
waddrmgr.Manager private passphrase each time the manager is unlocked.
If another call to Unlock is performed before the next Lock, the hash
is applied to the new input.  If it matches, we know the passphrase is
(likely) equal, so return early and continue using the crypto keys
already in memory.  If it does not match, we know for certain the
passphrase is incorrect and the manager is locked.

### DIFF
--- a/waddrmgr/manager.go
+++ b/waddrmgr/manager.go
@@ -17,6 +17,8 @@
 package waddrmgr
 
 import (
+	"crypto/rand"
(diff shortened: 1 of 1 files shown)


davecgh:
OK

PR #182, opened 2015-01-31
waddrmgr: Avoid unnecessarily rederiving keys on unlock.

jrick:
The slice of keys which must have their private extended keys derived
on unlock was never being removed from and all of these keys were
being rederived unnecessarily on every unlock.  Fix this by re-slicing
the deriveOnUnlock slice to remove the just derived key if the
derivation was successful.


davecgh:
Nice catch!  OK

PR #183, opened 2015-02-02
Mark addresses used

tuxcanfly:
Added a database flag on Address to mark used addresses and prevent address re-use.


davecgh:
Thanks.  I took a very quick look and noticed that the serialization formation of addresses has been changed, but there is no code to upgrade the existing addresses which will be in the old format.

tuxcanfly:
OK, didn't consider that since the integration branch is not on master, but I'll add it in case anyone is already using the branch.

tuxcanfly:
Looks like build failed due to travis:

`An error occurred while compiling the build script: Travis::Shell::Generator::TaintedOutput` 

A forced rebuild should set this back to green.

tuxcanfly:
Closing in favor of #196 which is based on intwaddrmgr instead of master

PR #184, opened 2015-02-04
Consolidate and optimize zero functions.

jrick:
Removes an XXX from the votingpool package.

I didn't touch the zero functions in the keystore package as it is being deprecated and would like to avoid any unnecessary merge conflicts from moving that to the legacy directory.

Zeroing is also performed in hdkeychain, so do we want this here, or somewhere else in our libraries so hdkeychain can also take advantage?  Alternatively, when the btcutil repo is split, we can move hdkeychain to btcwallet, and then update it to use this package.

Benchmarks from a xeon e3 (Bytes is the byte slice zeroing func provided by this package, Bytea* are the fixed-size funcs):

```
BenchmarkXor32  30000000                52.1 ns/op
BenchmarkXor64  20000000                91.5 ns/op
BenchmarkRange32        50000000                31.8 ns/op
BenchmarkRange64        30000000                49.5 ns/op
BenchmarkBytes32        200000000               10.1 ns/op
BenchmarkBytes64        100000000               15.4 ns/op
BenchmarkBytea32        1000000000               2.24 ns/op
BenchmarkBytea64        300000000                4.46 ns/op
```

### DIFF
--- a/internal/zero/benchmark_test.go
+++ b/internal/zero/benchmark_test.go
@@ -1,0 +1,83 @@
+package zero_test
+
+import (
+	"testing"
+
+	. "github.com/btcsuite/btcwallet/internal/zero"
+)
+
+var (
+	bytes32 = make([]byte, 32) // typical key size
+	bytes64 = make([]byte, 64) // passphrase hash size
+	bytea32 = new([32]byte)
+	bytea64 = new([64]byte)
+)
+
+// xor is the "slow" byte zeroing implementation which this package
+// originally replaced.  If this function benchmarks faster than the
+// functions exported by this package in a future Go version (perhaps
+// by calling runtime.memclr), replace the "optimized" versions with
+// this.
+func xor(b []byte) {
+	for i := range b {
+		b[i] ^= b[i]
+	}
+}
+
+// zrange is an alternative zero implementation that, while currently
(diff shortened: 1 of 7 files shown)


jrick:
Updated branch with tests.

jrick:
Added funcs to zero 32 and 64-byte arrays.  Also moved the zero arrays for Bytes and BigInt inside the function (so they live on the stack and are cleared every time the func is entered).  This shows slower benchmarks due to initializing the same memory over and over again, but should be much faster than a cache miss, and these still benchmark faster than the simple loops which set each byte individually.

davecgh:
Think this package should have a README.md to be consistent with the others like walletdb/waddrmgr/votingpool/etc?

jrick:
Depends. Do we expect others to use this package, or should it be an `internal` package?

davecgh:
It can probably be an internal package.  I like having a README.md for documentation purposes when people are learning how all the code fits together, but if you don't think it's worth it, that's fine.

jrick:
```Go
+// Package zero contains functions to clear data from byte slices and
+// multi-precision integers.
+package zero
```

Good enough for me :)

jrick:
Rebased+squashed, and made zero an internal package.

Issue #130, opened 2014-09-17
Is it possible to add a "watch only" address over RPC or WS?


lanephillips:
That is, can I add an address without a private key and still have btcwallet track the balance for that address?

jrick:
See #204.

hjfreyer:
To extend the request: it'd be great if you could import a public key for a BIP44 account, and have the app manage the whole subtree. The use case being monitoring and sending funds to a wallet in cold storage while still maintaining the privacy benefits of one-time addresses.

PR #185, opened 2015-02-04

tuxcanfly:

Depends on `import-address` branch of `btcjson` and `btcd`

Refs #130


tuxcanfly:
Closing in favor of #192

PR #186, opened 2015-02-05
Integrate address manager (package waddrmgr).

jrick:
This PR is being used to track the progress of merging the wallet DB changes to the master branch.  Currently, this branch is *not* ready for merging, but we would like Travis to perform builds of this branch regardless.

Changes to this PR should be submitted as pull requests to the intwaddrmgr branch.

Replaces #147.

Edit:

We have changed our decision and plan to merge the new address manager and wallet conversion code, even without wtxmgr being complete at the moment.  A description of the new features and changes copied from the PR:

- The wallet must now be created before starting the executable
- A new flag `--create` has been added to create the new wallet using wizard style question and answer prompts
- Starting the process without an existing wallet will now display a message to run it with `--create`
- Providing the `--create` flag with an existing wallet will simply show an error and return
- When an existing legacy keystore exists, the `--create` flag will prompt  you for the old private passphrase and automatic convert all old addresses to the new wallet being created


jrick:
I'm considering using merge, not rebase, to keep this branch in sync with changes applied to master.  This is so pull requests which build off of these changes will always only show the new commits introduced by the PR and not all of the previous commits from this intwaddrmgr branch.

A final rebase will be used before applying this branch to master to flatten the history and make the merge a fast-forward merge.

Thoughts?  I'd like to make working on top of this code as painless as possible, and transplanting series of commits from every PR for every intwaddrmgr rebase just feels like busy work to me.

davecgh:
Sounds fine.

jrick:
Pushed two additional bug fixes to this integration branch.  While they are not directly related to the integration, they require the new address manager's `SyncedTo` method to return the most recent block that the wallet is marked in sync with (this is not the same as btcd's most recently notified block).  These commits (1a6694d and 7a3b70) should not be squashed when the integration is finished.

jrick:
Rebased to latest master.  All the incremental merge commits have been removed.

jrick:
Removed the two additional fixes that were applied to this branch.  They will be tracked by new PRs after this is merged.

PR #189, opened 2015-02-06
Address several issues pointed out by lint and vet.

jrick:
This brings the entire tree closer to but not 100% goclean.sh clean.


davecgh:
These are all straight forward changes.  Looks good.

PR #192, opened 2015-02-06

tuxcanfly:

Depends on `import-address` branch of `btcjson` and `btcd`

Refs #130


jrick:
Needs a merge for travis, it's trying to pull btcec from the old location.

tuxcanfly:
OK, rebased and pushed.

rpcserver.go
```diff
@@ -1813,6 +1815,51 @@ func GetUnconfirmedBalance(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd)
...
+	addr, err := btcutil.DecodeAddress(cmd.Address, activeNet.Params)
+	if err != nil {
+		return nil, btcjson.ErrInvalidAddressOrKey
+	}
+
+	_, err = w.ImportAddress(addr, nil, cmd.Rescan)
+	if err != nil {
+		return nil, err
+	}
+	return nil, nil
```
jrick: These last 4 lines can be just:

```
return nil, err
```

jrick:
The importpubkey handler *only* imports compressed pubkeys.  Ideally it should handle all serialization formats, or worst case, we document that it only supports compressed keys.

rpcserver.go
```diff
@@ -1813,6 +1815,51 @@ func GetUnconfirmedBalance(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd)
...
 
+// ImportAddress handles an importaddress request by parsing
+// a encoded address and importing it as a watch-only address to the
+// imported address account.
+func ImportAddress(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interface{}, error) {
+	cmd := icmd.(*btcjson.ImportAddressCmd)
+
+	addr, err := btcutil.DecodeAddress(cmd.Address, activeNet.Params)
+	if err != nil {
+		return nil, btcjson.ErrInvalidAddressOrKey
```
jrick: You can use the error code from this var, but please add extra context in the error message with the error string.

rpcserver.go
```diff
@@ -1813,6 +1815,51 @@ func GetUnconfirmedBalance(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd)
...
+// ImportPubKey handles an importpubkey request by parsing
+// a raw public key and importing it as a watch-only address to the
+// imported address account.
+func ImportPubKey(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interface{}, error) {
+	cmd := icmd.(*btcjson.ImportPubKeyCmd)
+
+	// Import the public key, handling any errors.
+	pubKeyBytes, err := hex.DecodeString(cmd.PubKey)
+	if err != nil {
+		return nil, btcjson.ErrInvalidAddressOrKey
```
jrick: Error needs additional context.

rpcserver.go
```diff
@@ -1813,6 +1815,51 @@ func GetUnconfirmedBalance(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd)
...
+	}
+	pubKey, err := btcec.ParsePubKey(pubKeyBytes, btcec.S256())
+	if err != nil {
+		return nil, btcjson.ErrInvalidAddressOrKey
```
jrick: Error needs additional context.

rpcserver.go
```diff
@@ -1813,6 +1815,51 @@ func GetUnconfirmedBalance(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd)
...
+	}
+	_, err = w.ImportPubKey(pubKey, nil, cmd.Rescan)
+	switch {
+	case isManagerDuplicateError(err):
+		// Do not return duplicate key errors to the client.
+		return nil, nil
+	case isManagerLockedError(err):
```
jrick: There are no private keys to encrypt and the public crypto keys are already in memory.  Why would a manager locked error be expected here?

waddrmgr/address.go
```diff
@@ -110,64 +110,37 @@ type ManagedScriptAddress interface {
 	Script() ([]byte, error)
 }
 
+// baseAddress represents a address. It has only the most basic data common to
+// both managed and un-managed addresses.
+type baseAddress struct {
```
jrick: Rename it to publicAddress? This name reeks of OO. :)

waddrmgr/db.go
```diff
@@ -69,6 +70,24 @@ const (
...
+func byteAsBool(b byte) (bool, error) {
+	switch b {
+	case falseByte:
+		return false, nil
+	case trueByte:
+		return true, nil
+	default:
+		return false, errors.New("invalid byte representation of bool")
+	}
+}
```
jrick: I'd be ok with simplifying this to just returning `b != 0` and removing the error.  This was more of a concern of single bit flips when writing to a flat file, but the wallet db is checksummed.

waddrmgr/db.go
```diff
@@ -112,9 +132,16 @@ type dbChainAddressRow struct {
 	index  uint32
 }
 
-// dbImportedAddressRow houses additional information stored about an imported
+// dbImportedP2PKHAddressRow houses additional information stored about an imported
+// P2PKH address in the database.
+type dbImportedP2PKHAddressRow struct {
```
jrick: I think we should allow importing watching P2SH addresses as well, without the associated script to redeem the output.  Rename this to dbImportedHash160AddressRow?

waddrmgr/db.go
```diff
@@ -112,9 +132,16 @@ type dbChainAddressRow struct {
...
+	dbAddressRow
+	encryptedPKHash []byte
```
jrick: encryptedHash160

waddrmgr/db.go
```diff
@@ -554,19 +588,24 @@ func deserializeAddressRow(addressID, serializedAddress []byte) (*dbAddressRow,
...
 	binary.LittleEndian.PutUint32(buf[1:5], row.account)
 	binary.LittleEndian.PutUint64(buf[5:13], row.addTime)
 	buf[13] = byte(row.syncStatus)
-	binary.LittleEndian.PutUint32(buf[14:18], uint32(rdlen))
-	copy(buf[18:18+rdlen], row.rawData)
+	if row.watchingOnly {
+		buf[14] = trueByte
+	} else {
+		buf[14] = falseByte
+	}
```
jrick: A small helper func to choose the correct byte here would be a little nicer to read.

waddrmgr/db.go
```diff
@@ -606,14 +645,63 @@ func serializeChainedAddress(branch, index uint32) []byte {
...
+	retRow.encryptedPKHash = make([]byte, pkHashLen)
+	copy(retRow.encryptedPKHash, row.rawData[offset:offset+pkHashLen])
+	offset += pkHashLen
+
+	return &retRow, nil
+}
+
+// serializeImportedP2PKHAddress returns the serialization of the raw data field for
+// an imported P2PKH address.
+func serializeImportedP2PKHAddress(encryptedPKHash []byte) []byte {
```
jrick: Because this function only deals with P2PKH/Hash160-sized data, we can avoid the need to encode the length.  Instead, I would use `ripemd160.Size` throughout the serialization and deserialization code to refer the size of the hash.

waddrmgr/db.go
```diff
@@ -623,39 +711,49 @@ func deserializeImportedAddress(addressID []byte, row *dbAddressRow) (*dbImporte
 		return nil, managerError(ErrDatabase, str, nil)
 	}
 
-	retRow := dbImportedAddressRow{
+	retRow := dbImportedPubKeyAddressRow{
 		dbAddressRow: *row,
 	}
 
-	pubLen := binary.LittleEndian.Uint32(row.rawData[0:4])
+	var offset uint32
```
jrick: I'd avoid the use of the offset variable here since all of the offsets are constant (the pubLen below can be removed, see above comment).  No need to use a var to keep track of exactly where the next byte needs to be read from.

waddrmgr/db.go
```diff
@@ -623,39 +711,49 @@ func deserializeImportedAddress(addressID []byte, row *dbAddressRow) (*dbImporte
...
+func serializeImportedPubKeyAddress(encryptedPKHash, encryptedPubKey, encryptedPrivKey []byte) []byte {
 	// The serialized imported address raw data format is:
 	//   <encpubkeylen><encpubkey><encprivkeylen><encprivkey>
 	//
-	// 4 bytes encrypted pubkey len + encrypted pubkey + 4 bytes encrypted
-	// privkey len + encrypted privkey
+	// 4 bytes encrypted pubkey len + encrypted pubkey +
+	// 4 bytes encrypted privkey len + encrypted privkey
 	pubLen := uint32(len(encryptedPubKey))
 	privLen := uint32(len(encryptedPrivKey))
```
jrick: I thought we were not importing addresses with private keys.

waddrmgr/db.go
```diff
@@ -623,39 +711,49 @@ func deserializeImportedAddress(addressID []byte, row *dbAddressRow) (*dbImporte
...
+	var offset uint32
+	binary.LittleEndian.PutUint32(rawData[offset:offset+4], pubLen)
+	offset += 4
+	copy(rawData[offset:offset+pubLen], encryptedPubKey)
+	offset += pubLen
+
 	binary.LittleEndian.PutUint32(rawData[offset:offset+4], privLen)
 	offset += 4
 	copy(rawData[offset:offset+privLen], encryptedPrivKey)
+	offset += privLen
```
jrick: offset is dead (unused after this), don't increment it here.

waddrmgr/db.go
```diff
@@ -623,39 +711,49 @@ func deserializeImportedAddress(addressID []byte, row *dbAddressRow) (*dbImporte
...
+	// 4 bytes encrypted pubkey len + encrypted pubkey +
+	// 4 bytes encrypted privkey len + encrypted privkey
 	pubLen := uint32(len(encryptedPubKey))
 	privLen := uint32(len(encryptedPrivKey))
 	rawData := make([]byte, 8+pubLen+privLen)
-	binary.LittleEndian.PutUint32(rawData[0:4], pubLen)
-	copy(rawData[4:4+pubLen], encryptedPubKey)
-	offset := 4 + pubLen
+
+	var offset uint32
```
jrick: I would revert this code to the old way it calculated offset, to me it reads clearer.  Shorter and less state to consider.

waddrmgr/db.go
```diff
@@ -735,7 +833,11 @@ func fetchAddress(tx walletdb.Tx, addressID []byte) (interface{}, error) {
 	case adtChain:
 		return deserializeChainedAddress(addressID, row)
 	case adtImport:
-		return deserializeImportedAddress(addressID, row)
+		if row.watchingOnly {
+			return deserializeImportedP2PKHAddress(addressID, row)
+		} else {
+			return deserializeImportedPubKeyAddress(addressID, row)
+		}
```
jrick: Both P2PKH (really, any Hash160 address) and pubkeys can be watching only.

waddrmgr/db.go
```diff
@@ -887,7 +1009,7 @@ func fetchAllAddresses(tx walletdb.Tx) ([]interface{}, error) {
 		case adtChain:
 			addrRow, err = deserializeChainedAddress(k, row)
 		case adtImport:
-			addrRow, err = deserializeImportedAddress(k, row)
+			addrRow, err = deserializeImportedPubKeyAddress(k, row)
```
jrick: This should return pubkeys as well.

jrick:
I may be misremembering, but I thought the plan was to not allow importing private keys (and saving imported addresses with those key) due to issues with recovery, etc.

waddrmgr/manager.go
```diff
@@ -539,9 +539,25 @@ func (m *Manager) chainAddressRowToManaged(row *dbChainAddressRow) (ManagedAddre
...
+	pkHash, err := m.cryptoKeyPub.Decrypt(row.encryptedPKHash)
+	if err != nil {
+		str := "failed to decrypt pkhash for imported address"
+		return nil, managerError(ErrCrypto, str, err)
+	}
+	ma, err := newAddress(m, row.account, pkHash)
+	if err != nil {
+		return nil, err
+	}
+	return ma, nil
```
jrick: Simplify this to:

```
return newAddress(m, row.account, pkHash)
```

davecgh:
We do allow importing private keys.  There are certain circumstances such as vanity addresses and legacy keys which need to be supported.  It's true they can't be required automatically so they should generally be discouraged, but I do think they need to be supported (they already are supported in the waddrmgr except it can't import watching-only).

jrick:
Sure, and I agree it's good to keep it if we can.  I just remember hearing that discussed but can't remember who I talked to...

tuxcanfly:
Addressed the items reviewed above, chiefly:

* rename P2PKH to Hash160 in manager funcs and fields
* {de}serialize pubkey address: Use fixed sizes instead of offsets
* Add new addrtype (adtWatching) for imported watching only addrs

Since we have the new addrtype we can probably remove the `watchingOnly` flag on `dbAddressRow` but I think it might be useful for e.g. to filter all watching-only address irrespective of type. 

Regarding naming of `dbImportedPubKeyAddressRow`, I agree it includes private key information but I'm using the same convention as wallet Manager where we have `ManagedPubKeyAddress` which includes private key too. We can rename both for consistency if required but I think PubKey address conveys that the address is public key based and may or may not have the private key.

PR #194, opened 2015-02-10
waddrmgr: Update documentation for walletdb.

davecgh:
This pull request updates the documentation which discusses creating and opening the manager to properly mention the wallet database namespace as well as another typo.

### DIFF
--- a/waddrmgr/doc.go
+++ b/waddrmgr/doc.go
@@ -73,18 +73,17 @@ for this purpose.
 Creating a New Address Manager
 
 A new address manager is created via the Create function.  This function accepts
-the path to a database file to create, passphrases, network, and perhaps most
-importantly, a cryptographically random seed which is used to generate the
-master node of the hierarchical deterministic keychain which allows all
-addresses and private keys to be recovered with only the seed.  The GenerateSeed
-function in the hdkeychain package can be used as a convenient way to create a
-random seed for use with this function.  The address manager is locked
-immediately upon being created.
+a wallet database namespace, passphrases, network, and perhaps most importantly,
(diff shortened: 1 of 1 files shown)


jrick:
Looks ok.

While reading the diff I also noticed a reference to the type `waddrmgr.ManagerError`.  We can probably drop the package prefix for that.

davecgh:
I suppose it could be.  It's written from the viewpoint of the caller though and they'll need the package prefix.  It's probably fine either way.  I'll change it over if you think it should be.

jrick:
That's how the standard library package documentation refers to its own types (but code examples do still include the package names for obvious reasons).

davecgh:
Fair point.  I changed it over.

waddrmgr/doc.go
```diff
@@ -73,18 +73,17 @@ for this purpose.
...
-the path to a database file to create, passphrases, network, and perhaps most
-importantly, a cryptographically random seed which is used to generate the
-master node of the hierarchical deterministic keychain which allows all
-addresses and private keys to be recovered with only the seed.  The GenerateSeed
-function in the hdkeychain package can be used as a convenient way to create a
-random seed for use with this function.  The address manager is locked
-immediately upon being created.
+a wallet database namespace, passphrases, network, and perhaps most importantly,
+a cryptographically random seed which is used to generate the master node of the
+hierarchical deterministic keychain which allows all addresses and private keys
```
jrick: while we're in the doco fixing mood: "all derived addresses"?

davecgh:
I'll do another pull request for that change real quick.  Thought that was it.

PR #201, opened 2015-02-26
Use txscript.StandardVerifyFlags

dajohi:
By using txscript.StandardVerifyFlags when creating and validating
transactions, we can ensure the transactions successfully created
will be accepted by btcd.

### DIFF
--- a/createtx.go
+++ b/createtx.go
@@ -415,14 +415,10 @@ func signMsgTx(msgtx *wire.MsgTx, prevOutputs []txstore.Credit, store *keystore.
 }
 
 func validateMsgTx(msgtx *wire.MsgTx, prevOutputs []txstore.Credit) error {
-	flags := txscript.ScriptVerifyDERSignatures | txscript.ScriptStrictMultiSig
-	bip16 := time.Now().After(txscript.Bip16Activation)
-	if bip16 {
-		flags |= txscript.ScriptBip16
-	}
 	for i, txin := range msgtx.TxIn {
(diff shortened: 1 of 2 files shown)


createtx.go
```diff
@@ -415,11 +415,7 @@ func signMsgTx(msgtx *wire.MsgTx, prevOutputs []txstore.Credit, store *keystore.
 }
 
 func validateMsgTx(msgtx *wire.MsgTx, prevOutputs []txstore.Credit) error {
-	flags := txscript.ScriptVerifyDERSignatures | txscript.ScriptStrictMultiSig
-	bip16 := time.Now().After(txscript.Bip16Activation)
-	if bip16 {
-		flags |= txscript.ScriptBip16
-	}
+	flags := txscript.StandardVerifyFlags
```
jrick: Declare this as const

rpcserver.go
```diff
@@ -2841,8 +2841,7 @@ func SignRawTransaction(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (in
 
 		// Either it was already signed or we just signed it.
 		// Find out if it is completely satisfied or still needs more.
-		flags := txscript.ScriptBip16 | txscript.ScriptVerifyDERSignatures |
-			txscript.ScriptStrictMultiSig
+		flags := txscript.StandardVerifyFlags
```
jrick: const (or just pass in in directly, it was a var before so they could be modified for bip16)

jrick:
ok

PR #202, opened 2015-03-02
Rescan after btcd reconnect.

jrick:
Otherwise, if the websocket connection to btcd is lost and
reestablished, there is no sync to the current best block, and btcd
will not notify wallet of new relevant transactions.


davecgh:
OK after squash.

PR #203, opened 2015-03-02
Implement several btcd RPCs in wallet as well.

jrick:
Wallet should handle these so that wallet clients don't end up
creating requests to btcd about the latest processed block, which is
not the same as wallet's most recently processed block.

By providing wallet clients with this info, we avoid a race where the
client thinks wallet has processed N blocks, but in fact is still
synced to N-1 (and perhaps currently processing transactions from
block N).  This can cause unexpected results for many of the
bitcoind-compatible RPC APIs due to their reliance on number of
confirmations, rather than using absolute block heights.


davecgh:
OK

PR #204, opened 2015-03-02

tuxcanfly:

Refs #192


waddrmgr/db.go
```diff
@@ -549,14 +581,14 @@ func putNumAccounts(tx walletdb.Tx, numAccounts uint32) error {
 // the common parts.
 func deserializeAddressRow(addressID, serializedAddress []byte) (*dbAddressRow, error) {
 	// The serialized address format is:
-	//   <addrType><account><addedTime><syncStatus><rawdata>
+	//   <addrType><account><addedTime><syncStatus><watchingOnly><rawdata>
```
davecgh: Why do we need a watching only flag as well as an `adtWatching` address type?  It would be ideal to avoid changing this format because, if you do, you'll have to write upgrade code and modify every existing address in the database which I think we really don't want to do.

davecgh: After thinking about it some more, I think having the extra field is more flexible.  However, at the moment there is some redundancy with the `adtWatching` type that doesn't quite feel right.

If we look at it from the perspective of what a user will need to do for watching-only purposes:

* Import BIP0032 public key chains
* Import "loose" public keys (those which are not a part of BIP0032 keychain)
* Import the hash160 associated with a pay-to-script-hash

Considering that, the first case is an `adtChain` with a watching-only flag, the second is an `adtImport` with a watching-only flag, and the third is an `adtScript` with a watching-only flag.

waddrmgr/db.go
```diff
@@ -70,11 +71,31 @@ type addressType uint8
...
+func boolAsByte(b bool) byte {
+	if b {
+		return trueByte
+	}
+	return falseByte
+}
+
+// byteAsBool converts a byte to a bool.
+func byteAsBool(b byte) bool {
+	return b != 0
```
davecgh: Not that I find 0 hard to read here, but just above you've used constants for `trueByte` and `falseByte`, so shouldn't this be consistent with `b != falseByte`?

waddrmgr/db.go
```diff
@@ -770,9 +841,11 @@ func fetchAddress(tx walletdb.Tx, addressID []byte) (interface{}, error) {
 	case adtChain:
 		return deserializeChainedAddress(addressID, row)
 	case adtImport:
-		return deserializeImportedAddress(addressID, row)
+		return deserializeImportedPubKeyAddress(addressID, row)
 	case adtScript:
 		return deserializeScriptAddress(addressID, row)
+	case adtWatching:
+		return deserializeImportedHash160AddressRow(addressID, row)
```
davecgh: Given the above comments, this line will end up going away with the removal of the `adtWatching` type.

Of note is there is already code to deal with watching-only address conversion in `deletePrivateKeys` which is used when converting the entire manager to watching-only, so, outside of the addition of the watching-only flag to each address, there shouldn't be any need to modify the raw data serialization formats at all.

waddrmgr/db.go
tuxcanfly: Updated to only use the `watchingOnly` flag and removed `adtWatching` type. Importing "loose" public keys and P2SH hashes works using the `importpubkey` and `importaddress` commands. Now working on importing public key chains. 

Now working on importing BIP0032 public key chains.

waddrmgr/address.go
```diff
@@ -88,22 +88,31 @@ type ManagedScriptAddress interface {
...
 // the private key associated with the public key.
 type managedAddress struct {
-	manager          *Manager
-	account          uint32
-	address          *btcutil.AddressPubKeyHash
-	imported         bool
-	internal         bool
-	compressed       bool
-	used             bool
+	publicAddress
```
davecgh: What is the purpose of splitting this out?

waddrmgr/address.go
```diff
@@ -263,6 +272,21 @@ func (a *managedAddress) ExportPrivKey() (*btcutil.WIF, error) {
 	return btcutil.NewWIF(pk, a.manager.chainParams, a.compressed)
 }
 
+// newAddress returns a new address based on the passed account and address
+// pubkeyhash.
+func newAddress(m *Manager, account uint32, addressID []byte) (*publicAddress, error) {
```
davecgh: If the `publicAddress` bits aren't separated out (see previous comment), this can simply be:

```Go
newManagedPublicAddress(m *Manager, account uint32, addressID []byte) (*managedAddress, error)
```

davecgh:
Since the format of existing addresses in the db is being changed here, the db version needs to be bumped and it needs upgrade code to handle the new serialization.  You should be able to do it without requiring the private passphrase by specifically targetting the new field offset.

waddrmgr/address.go
tuxcanfly: I thought it might be logical to separate the structs for loose vs keychain addrs for future e.g if we wanted to add a method specific to loose addrs.

waddrmgr/address.go
tuxcanfly: Thanks, I've reverted to `managedAddress` and updated the method name.

tuxcanfly:
Yeah, we will need a upgrade to initialize the `watchingOnly` field, I guess it can be inferred by checking that `pubkey` is not nil.

tuxcanfly:
@davecgh Added upgrade path, addressed items reviewed above. Tested that we support importing pubkeys, p2pkh and p2sh addresses now.

waddrmgr/db.go
```diff
@@ -902,35 +927,48 @@ func deserializeImportedAddress(row *dbAddressRow) (*dbImportedAddressRow, error
 		dbAddressRow: *row,
 	}
 
-	pubLen := binary.LittleEndian.Uint32(row.rawData[0:4])
+	pkHashLen := binary.LittleEndian.Uint32(row.rawData[0:4])
```
davecgh: Please stick with the existing offset approach.  There is less chance of error when updating the code with that approach.

waddrmgr/db.go
```diff
@@ -1865,3 +1816,165 @@ func upgradeToVersion3(namespace walletdb.Namespace, seed, privPassPhrase, pubPa
...
+
+// upgradeToVersion4 upgrades the database from version 3 to version 4
+// The following fields were added to the serialized address row:
+// * 'watchingOnly' - boolean indicating whether the address is watching-only
+// This is required to flag each imported address as watching-only
+// * 'encryptedPubKeyHash' - encrypted pubkey hash of imported p2pkh address
+// This is required to import addresses without the associated public key.
+func upgradeToVersion4(namespace walletdb.Namespace) error {
+	err := namespace.Update(func(tx walletdb.Tx) error {
+		currentMgrVersion := uint32(4)
```
davecgh: No need for this local.  Just set it directly in the call to `putManagerVersion` below.

waddrmgr/db.go
```diff
@@ -1865,3 +1816,165 @@ func upgradeToVersion3(namespace walletdb.Namespace, seed, privPassPhrase, pubPa
...
+		bucket := tx.RootBucket().Bucket(addrBucketName)
+
+		err := bucket.ForEach(func(k, v []byte) error {
+			// Skip buckets.
+			if v == nil {
+				return nil
+			}
+
+			// Insert watchingOnly bool at byte 14
+			// Since importing watching-only addrs was added in this version we
```
davecgh: Minor nit: I've made it a point to avoid personal pronouns in the address manager.  Please change this to:

```Go
			// Since importing watching-only addrs was not added
			// until this version, mark all existing addresses as
			// not being watching-only.
```

waddrmgr/db.go
```diff
@@ -1361,7 +1402,7 @@ func deletePrivateKeys(tx walletdb.Tx) error {
 			// Reserialize the imported address without the private
 			// key and store it.
 			row.rawData = serializeImportedAddress(
-				irow.encryptedPubKey, nil)
+				irow.encryptedPubKeyHash, irow.encryptedPubKey, nil)
```
davecgh: Please respect column 80.

waddrmgr/db.go
```diff
@@ -902,35 +927,48 @@ func deserializeImportedAddress(row *dbAddressRow) (*dbImportedAddressRow, error
...
 	pubLen := uint32(len(encryptedPubKey))
 	privLen := uint32(len(encryptedPrivKey))
-	rawData := make([]byte, 8+pubLen+privLen)
-	binary.LittleEndian.PutUint32(rawData[0:4], pubLen)
-	copy(rawData[4:4+pubLen], encryptedPubKey)
-	offset := 4 + pubLen
-	binary.LittleEndian.PutUint32(rawData[offset:offset+4], privLen)
-	offset += 4
-	copy(rawData[offset:offset+privLen], encryptedPrivKey)
+	rawData := make([]byte, 12+pkHashLen+pubLen+privLen)
```
davecgh: As above, please stick with the existing offset approach.

waddrmgr/db.go
```diff
@@ -1865,3 +1816,165 @@ func upgradeToVersion3(namespace walletdb.Namespace, seed, privPassPhrase, pubPa
...
+		err := bucket.ForEach(func(k, v []byte) error {
+			// Skip buckets.
+			if v == nil {
+				return nil
+			}
+
+			// Insert watchingOnly bool at byte 14
+			// Since importing watching-only addrs was added in this version we
+			// can safely initialize this to falseByte
+			x := make([]byte, len(v)+1)
```
davecgh: Please choose a better name than x.

waddrmgr/db.go
```diff
@@ -1865,3 +1816,165 @@ func upgradeToVersion3(namespace walletdb.Namespace, seed, privPassPhrase, pubPa
...
+			// Update imported private key addresses for encryptedPubKeyHash field
+			switch addrType {
+			case adtImport:
+				// Insert nil for encrypted pkhash in rawData
+				rdlen := binary.LittleEndian.Uint32(x[15:19])
+				binary.LittleEndian.PutUint32(x[15:19], uint32(rdlen+4))
+				rawData := make([]byte, rdlen+4)
+				copy(rawData, x[19:])
+				binary.LittleEndian.PutUint32(x[19:23], 0)
+				copy(x[23:], rawData)
```
davecgh: The will cut off the end of the raw data.  `x` here will only be `len(v)+1` which doesn't include the extra 4 bytes, so when `rawData` is copied into `x[23:]` it will end up being 4 bytes too long.

wallet.go
```diff
@@ -1122,6 +1124,102 @@ func (w *Wallet) DumpWIFPrivateKey(addr btcutil.Address) (string, error) {
 	return wif.String(), nil
 }
 
+// ImportAddress imports a watch-only address to the wallet.
+func (w *Wallet) ImportAddress(addr btcutil.Address, bs *waddrmgr.BlockStamp,
+	rescan bool) (string, error) {
+
+	// The starting block for the key is the genesis block unless otherwise
```
davecgh: This needs a check/error if it's not for the current network the manager is using.

Something like:

```Go
if !addr.IsForNet(activeNet.Params) {
	// error...
}
```

wallet.go
```diff
@@ -1122,6 +1124,102 @@ func (w *Wallet) DumpWIFPrivateKey(addr btcutil.Address) (string, error) {
...
+	// specified.
+	if bs == nil {
+		bs = &waddrmgr.BlockStamp{
+			Hash:   *activeNet.Params.GenesisHash,
+			Height: 0,
+		}
+	}
+
+	// Check whether addr is P2SH or P2PKH and import accordingly
```
davecgh: Rather than doing this, I'd suggest just doing a type switch on the address and moving this entire logic to the manager's `ImportAddress` function which should work with any address type.  The necessary decoding has already been done by the time it's an address and base58 decoding is quite slow.

Also, there needs to be a default case which returns an unsupported address error.

Something like:

```Go
switch addr := addr.(type) {
case *btcutil.AddressPubKeyHash:
	// Handle p2pkh addresses...

case *btcutil.AddressPubKey:
	// Handle p2pk addresses...

case *btcutil.AddressScriptHash:
	// Handle p2sh addresses...

default:
	// Unsupported address type error....
}
```

wallet.go
```diff
@@ -1122,6 +1124,102 @@ func (w *Wallet) DumpWIFPrivateKey(addr btcutil.Address) (string, error) {
...
+		return "", err
+	}
+	switch {
+	case chaincfg.IsPubKeyHashAddrID(netID):
+		_, err := w.Manager.ImportAddress(addr, bs)
+		if err != nil {
+			return "", err
+		}
+	case chaincfg.IsScriptHashAddrID(netID):
+		_, err := w.Manager.ImportScript(addr.ScriptAddress(), bs)
```
davecgh: This is incorrect.  ImportScript requires the actual script, not just the p2sh.  Please see my previous message about using a type switch in the manager's `ImportAddress` function which will prevent this type of mistake.

davecgh: I'd also like to note that the tests must not be thorough enough to not catch this.

waddrmgr/manager.go
```diff
@@ -963,6 +973,147 @@ func (m *Manager) existsAddress(addressID []byte) (bool, error) {
...
+
+// ImportAddress imports a raw pkhash as a watch-only address into the address manager.
+//
+// All imported addresses will be part of the account defined by the
+// ImportedAddrAccount constant.
+func (m *Manager) ImportAddress(addr btcutil.Address, bs *BlockStamp) (ManagedAddress, error) {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	pubKeyHash := addr.ScriptAddress()
```
davecgh: As mentioned below, this function should be looking at the type of the address with a type assertion and handling it appropriately for all supported types.  See my comment in the Wallet's `ImportAddress` function for more details.

wallet.go
tuxcanfly: Sorry, looks like I overlooked the test for importing script addrs. Let me add that.

wallet.go
tuxcanfly: OK, that sounds better, let me get that.

waddrmgr/db.go
tuxcanfly: Thanks, missed that. Will add a test to prevent this.

wallet.go
davecgh: This check should probably also go into the manager's `ImportAddress` function along with the code below since the manager needs to reject it.

tuxcanfly:
Addressed all the review items above except the upgrade re-serialization, will be done with that in a bit.

tuxcanfly:
@davecgh Done with the re-serialization bit. Added docs to make it readable. Please review.

waddrmgr/db.go
```diff
@@ -821,9 +842,11 @@ func deserializeAddressRow(serializedAddress []byte) (*dbAddressRow, error) {
 	row.account = binary.LittleEndian.Uint32(serializedAddress[1:5])
 	row.addTime = binary.LittleEndian.Uint64(serializedAddress[5:13])
 	row.syncStatus = syncStatus(serializedAddress[13])
-	rdlen := binary.LittleEndian.Uint32(serializedAddress[14:18])
+	watchingOnly := byteAsBool(serializedAddress[14])
+	row.watchingOnly = watchingOnly
```
davecgh: No need for the local here.

```Go
row.watchingOnly = byteAsBool(serializedAddress[14])
```

waddrmgr/db.go
```diff
@@ -1168,6 +1200,23 @@ func putScriptAddress(tx walletdb.Tx, addressID []byte, account uint32,
 	return nil
 }
 
+// putImportedAddress stores the provided imported address information to the
+// database.
+func putImportedAddress(tx walletdb.Tx, addressID []byte, account uint32,
+	status syncStatus, watchingOnly bool, encryptedPubKeyHash, encryptedPubKey, encryptedPrivKey []byte) error {
```
davecgh: Everything after `encryptedPubKeyHash` should be on the next line.

waddrmgr/db.go
```diff
@@ -1168,6 +1200,23 @@ func putScriptAddress(tx walletdb.Tx, addressID []byte, account uint32,
...
+
+	rawData := serializeImportedAddress(encryptedPubKeyHash, encryptedPubKey, encryptedPrivKey)
```
davecgh: As should `encryptedPrivKey` here.

waddrmgr/db.go
```diff
@@ -1361,7 +1410,9 @@ func deletePrivateKeys(tx walletdb.Tx) error {
 			// Reserialize the imported address without the private
 			// key and store it.
 			row.rawData = serializeImportedAddress(
-				irow.encryptedPubKey, nil)
+				irow.encryptedPubKeyHash,
```
davecgh:
```Go
			row.rawData = serializeImportedAddress(
				irow.encryptedPubKeyHash, irow.encryptedPubKey,
				nil)
```

waddrmgr/db.go
tuxcanfly: Yup, let me setup vim to highlight col 80 overflow.

waddrmgr/db.go
```diff
@@ -1865,3 +1824,166 @@ func upgradeToVersion3(namespace walletdb.Namespace, seed, privPassPhrase, pubPa
...
+			// Skip buckets.
+			if v == nil {
+				return nil
+			}
+
+			// Since importing watching-only addrs was not added
+			// until this version, mark all existing addresses as
+			// not being watching-only.
+			serializedAddress := make([]byte, len(v)+1)
+			copy(serializedAddress[:13], v[:13])
```
davecgh: This is an off-by-one.  It should be:
```Go
copy(serializedAddress[:14], v[:14])
```

I'm really concerned the tests are not catching these.

tuxcanfly: Sorry, missed that. Let me add a test case for this upgrade.

waddrmgr/address.go
```diff
@@ -195,28 +199,48 @@ func (a *managedAddress) Used() bool {
...
+
 // PubKey returns the public key associated with the address.
 //
 // This is part of the ManagedPubKeyAddress interface implementation.
-func (a *managedAddress) PubKey() *btcec.PublicKey {
-	return a.pubKey
+func (a *managedAddress) PubKey() (*btcec.PublicKey, error) {
+	// Make sure public key is available for this address.
+	if a.pubKey == nil {
+		return nil, managerError(ErrNoPubKey, errNoPubKey, nil)
```
davecgh: This is probably better as `ErrWatchingOnly`?

waddrmgr/address.go
```diff
@@ -225,7 +249,7 @@ func (a *managedAddress) ExportPubKey() string {
 // This is part of the ManagedPubKeyAddress interface implementation.
 func (a *managedAddress) PrivKey() (*btcec.PrivateKey, error) {
 	// No private keys are available for a watching-only address manager.
-	if a.manager.watchingOnly {
+	if a.manager.watchingOnly || a.watchingOnly {
```
davecgh: I suspect this should now just be `if a.watchingOnly {`.  All addreses under a watching-only address manager must now be marked watching-only too, so this could hide a bug by checking both conditions.

waddrmgr/address.go
```diff
@@ -460,12 +502,19 @@ func (a *scriptAddress) Used() bool {
...
+	return a.watchingOnly
+}
+
 // Script returns the script associated with the address.
 //
 // This implements the ScriptAddress interface.
 func (a *scriptAddress) Script() ([]byte, error) {
 	// No script is available for a watching-only address manager.
-	if a.manager.watchingOnly {
+	if a.manager.watchingOnly || a.watchingOnly {
```
davecgh: Same deal here regarding the check for `if a.watchingOnly {`.

rpcserver.go
```diff
@@ -2890,7 +2943,11 @@ func ValidateAddress(w *Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (inter
 	switch ma := ainfo.(type) {
 	case waddrmgr.ManagedPubKeyAddress:
 		result.IsCompressed = ma.Compressed()
-		result.PubKey = ma.ExportPubKey()
+		pubKey, err := ma.ExportPubKey()
+		if err != nil {
+			return nil, ErrAddressNotInWallet
```
davecgh: I'm pretty sure this should just be a break instead of an error return.  Just because the wallet doesn't have it, does not mean it's an invalid address.  See the code below for errors on `ma.Script()`.

Also note that the call to `ma.ExportPubKey` needs to be before setting `result.IsCompressed` because that information is unknown if the pubkey isnt available.

waddrmgr/manager.go
```diff
@@ -1065,16 +1220,15 @@ func (m *Manager) ImportPrivateKey(wif *btcutil.WIF, bs *BlockStamp) (ManagedPub
 	var managedAddr *managedAddress
 	if !m.watchingOnly {
 		managedAddr, err = newManagedAddress(m, ImportedAddrAccount,
-			wif.PrivKey, wif.CompressPubKey)
+			wif.PrivKey, wif.CompressPubKey, true)
 	} else {
 		pubKey := (*btcec.PublicKey)(&wif.PrivKey.PublicKey)
 		managedAddr, err = newManagedAddressWithoutPrivKey(m,
-			ImportedAddrAccount, pubKey, wif.CompressPubKey)
+			ImportedAddrAccount, pubKey, wif.CompressPubKey, m.watchingOnly, true)
```
davecgh: 80

waddrmgr/manager_test.go
```diff
@@ -121,6 +138,12 @@ func testManagedPubKeyAddress(tc *testContext, prefix string, gotAddr waddrmgr.M
 	// Since this is only available when the manager is unlocked, also check
 	// for the expected error when the manager is locked.
 	gotPrivKey, err := gotAddr.PrivKey()
+	if wantAddr.watchingOnly {
```
davecgh: I'd combine this in the switch case below like so:

```Go
case tc.watchingOnly, wantAddr.watchingOnly:
``` 
Can avoid the extra check and get the better error prefix on a failure.

waddrmgr/manager_test.go
```diff
@@ -191,11 +214,18 @@ func testManagedPubKeyAddress(tc *testContext, prefix string, gotAddr waddrmgr.M
...
 // are checked to ensure they return the correct error.
 func testManagedScriptAddress(tc *testContext, prefix string, gotAddr waddrmgr.ManagedScriptAddress, wantAddr *expectedAddr) bool {
-	// Ensure script is the expected value for the managed address.
 	// Ensure script is the expected value for the managed address.  Since
 	// this is only available when the manager is unlocked, also check for
 	// the expected error when the manager is locked.
 	gotScript, err := gotAddr.Script()
+	// If the script is watching-only, it should return the expected error in
+	// all cases
+	if wantAddr.watchingOnly {
```
davecgh: Same here.

```Go
case tc.watchingOnly, wantAddr.watchingOnly:
```

waddrmgr/address.go
tuxcanfly: Yeah, it's `ErrWatchingFromADistanceOnly` :)

waddrmgr/address.go
tuxcanfly: Yup, we should prefer single source of truth. Let me update that.

waddrmgr/manager_test.go
tuxcanfly: Neat. Updated both places.

tuxcanfly:
Needs tests with the version upgrades test setup in #237

rpcserver.go
```diff
@@ -1832,6 +1838,53 @@ func GetUnconfirmedBalance(w *wallet.Wallet, chainSvr *chain.Client, icmd btcjso
...
+// ImportAddress handles an importaddress request by parsing
+// a encoded address and importing it as a watch-only address to the
+// imported address account.
+func ImportAddress(w *wallet.Wallet, chainSvr *chain.Client, icmd btcjson.Cmd) (interface{}, error) {
+	cmd := icmd.(*btcjson.ImportAddressCmd)
+
+	addr, err := btcutil.DecodeAddress(cmd.Address, activeNet.Params)
+	if err != nil {
+		return nil, btcjson.Error{
+			Code:    btcjson.ErrInvalidAddressOrKey.Code,
```
jrick: Check the struct definitions at the top of this file.  This can be simplified by simply doing:

```Go
return nil, InvalidAddressOrKeyError{err}
```

rpcserver.go
```diff
@@ -1832,6 +1838,53 @@ func GetUnconfirmedBalance(w *wallet.Wallet, chainSvr *chain.Client, icmd btcjso
...
+	if !addr.IsForNet(activeNet.Params) {
+		return nil, btcjson.ErrInvalidAddressOrKey
+	}
+
+	_, err = w.ImportAddress(addr, nil, cmd.Rescan)
+	return nil, err
+}
+
+// ImportPubKey handles an importpubkey request by parsing
+// a raw public key and importing it as a watch-only address to the
```
jrick: It's not really an address, it's a pubkey (which is a different kind of pkScript).

Also, this only seems to import *compressed* pubkeys, even if the original pubkey in the request is uncompressed.  We need to consider both cases.

rpcserver.go
```diff
@@ -1891,7 +1944,11 @@ func CreateNewAccount(w *wallet.Wallet, chainSvr *chain.Client, icmd btcjson.Cmd
 
 	_, err = w.Manager.NewAccount(cmd.Account)
 	if isManagerLockedError(err) {
-		return nil, btcjson.ErrWalletUnlockNeeded
+		return nil, btcjson.Error{
+			Code: btcjson.ErrWalletUnlockNeeded.Code,
+			Message: "Creating an account requires the wallet to be unlocked. " +
+				"Enter the wallet passphrase with walletpassphrase to unlock",
+		}
```
jrick: I think all this code is redundant.  Needs a rebase over latest master?

waddrmgr/address.go
```diff
@@ -98,6 +101,7 @@ type managedAddress struct {
 	internal         bool
 	compressed       bool
 	used             bool
+	watchingOnly     bool
 	pubKey           *btcec.PublicKey
 	privKeyEncrypted []byte
```
jrick: Why do we need a watchingOnly boolean here when the privKeyEncrypted field could be nilled instead?

waddrmgr/address.go
```diff
@@ -263,10 +287,26 @@ func (a *managedAddress) ExportPrivKey() (*btcutil.WIF, error) {
 	return btcutil.NewWIF(pk, a.manager.chainParams, a.compressed)
 }
 
+// newManagedPublicAddress returns a new address based on the passed account
+// and public key hash.
+func newManagedPublicAddress(m *Manager, account uint32, pkHash []byte) (*managedAddress, error) {
```
jrick: All addresses are public :)

Maybe replace Public here with WatchingOnly?

waddrmgr/address.go
```diff
@@ -366,6 +407,7 @@ type scriptAddress struct {
 	scriptCT        []byte
 	scriptMutex     sync.Mutex
 	used            bool
+	watchingOnly    bool
```
jrick: Same deal here as with the p2pkh address.  We could define a watching-only p2sh address as one with a nil scriptEncrypted field.

waddrmgr/db.go
```diff
@@ -109,12 +128,13 @@ type dbBIP0044AccountRow struct {
...
-	addTime    uint64
-	syncStatus syncStatus
-	used       bool
-	rawData    []byte // Varies based on address type field.
+	addrType     addressType
+	account      uint32
+	addTime      uint64
+	syncStatus   syncStatus
+	used         bool
+	watchingOnly bool
```
jrick: I think this is unnecessary too.  We could use 0 as the encrypted privkey length in the raw data.

waddrmgr/db.go
```diff
@@ -696,12 +717,14 @@ func deleteAccountIDIndex(tx walletdb.Tx, account uint32) error {
...
-		str := fmt.Sprintf("failed to store account name index key %s", name)
-		return managerError(ErrDatabase, str, err)
+	// Skip if we're working with the initial db version
+	if bucket != nil {
+		// Write the account number keyed by the account name.
+		err := bucket.Put(stringToBytes(name), uint32ToBytes(account))
+		if err != nil {
+			str := fmt.Sprintf("failed to store account name index key %s", name)
+			return managerError(ErrDatabase, str, err)
+		}
```
jrick: Looks unrelated.  Needs rebase?

waddrmgr/db.go
```diff
@@ -1641,227 +1665,292 @@ func createManagerNS(namespace walletdb.Namespace) error {
 // upgradeToVersion2 upgrades the database from version 1 to version 2
 // 'usedAddrBucketName' a bucket for storing addrs flagged as marked is
 // initialized and it will be updated on the next rescan.
-func upgradeToVersion2(namespace walletdb.Namespace) error {
-	err := namespace.Update(func(tx walletdb.Tx) error {
-		currentMgrVersion := uint32(2)
-		rootBucket := tx.RootBucket()
+func upgradeToVersion2(tx walletdb.Tx) error {
+	rootBucket := tx.RootBucket()
```
jrick: Why is previous update code being modified? fixed with a rebase?

waddrmgr/manager.go
```diff
@@ -963,6 +972,152 @@ func (m *Manager) existsAddress(addressID []byte) (bool, error) {
...
+		return nil, managerError(ErrWrongNet, str, nil)
+	}
+
+	addressID := addr.ScriptAddress()
+	alreadyExists, err := m.existsAddress(addressID)
+	if err != nil {
+		return nil, err
+	}
+	if alreadyExists {
+		str := fmt.Sprintf("address for address ID %x already exists",
```
jrick: This is a far less useful error message than just printing the address string.

waddrmgr/manager.go
```diff
@@ -963,6 +972,152 @@ func (m *Manager) existsAddress(addressID []byte) (bool, error) {
 	return exists, nil
 }
 
+// ImportAddress imports the given address as a watch-only address into the address manager.
+//
+// All imported addresses will be part of the account defined by the
+// ImportedAddrAccount constant.
+func (m *Manager) ImportAddress(addr btcutil.Address, bs *BlockStamp) (ManagedAddress, error) {
```
jrick: Needs some kind of comment about what bs is for.

waddrmgr/manager.go
```diff
@@ -963,6 +972,152 @@ func (m *Manager) existsAddress(addressID []byte) (bool, error) {
...
+			if updateStartBlock {
+				return putStartBlock(tx, bs)
+			}
+
+			return nil
+		})
+	case *btcutil.AddressPubKey:
+		// Encrypt public key.
+		pubKey := addr.PubKey()
+		serializedPubKey := pubKey.SerializeCompressed()
```
jrick: Again, can't assume it's always compressed.  Need to handle both cases.

waddrmgr/address.go
tuxcanfly: OK, I'll try to replace it with nilled privKeyEncrypted. Also, as per the IRC discussion, I think if we really want to store `watchingOnly` bool, it could be a part of the flag field.

waddrmgr/manager.go
tuxcanfly: Thanks, let me update that. I think we can check whether the passed pubkey was compressed/uncompressed based on length and store in accordingly.

waddrmgr/address.go
tuxcanfly: Hmm, how about `newManagedPKHashAddress`? WatchingOnly could mean addresses imported using `importpubkey` i.e. pubkey based addresses. This func is specifically for P2PKH addresses imported using just the pkhash.

jrick: PKHash or P2PKH are both fine here

tuxcanfly: Thanks, keeping consistent with `newScriptAddress` I think I'll go with `newPKHashAddress`. It's not really 'managed' so I've removed that.

waddrmgr/db.go
tuxcanfly: Thanks, yeah fixed both issues with rebase. Just moved the definition of `upgradeManager` to the end, after all `upgradeToVersion` funcs.

waddrmgr/db.go
tuxcanfly: OK, updated to use `WatchingOnly()` to return `len(a.privKeyEncrypted) == 0` so we're doing this now.

waddrmgr/address.go
tuxcanfly: Thanks, doing this now.

rpcserver.go
tuxcanfly: Thanks, updated.

rpcserver.go
tuxcanfly: Yeah I guess it was, rebased now so should be fixed.

rpcserver.go
tuxcanfly: OK, updated the doc. Also, we're writing compressed/uncompressed based on the input pubkey now.

jrick:
This needs a rebase over latest master.  Could you also squash all of the fixes (all commits after 43f8801) into a single commit?

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -154,6 +154,16 @@ var helpDescsEnUS = map[string]string{
 	"importprivkey-label":     "Unused (must be unset or 'imported')",
 	"importprivkey-rescan":    "Rescan the blockchain (since the genesis block) for outputs controlled by the imported key",
 
+	// ImportPubKeyCmd help.
+	"importpubkey--synopsis": "Imports a raw public key to the 'imported' account.",
+	"importpubkey-pubkey":    "The raw public key. Maybe compressed or uncompressed.",
```
jrick: May be

internal/rpchelp/methods.go
```diff
@@ -50,6 +50,8 @@ var Methods = []struct {
 	{"gettransaction", []interface{}{(*btcjson.GetTransactionResult)(nil)}},
 	{"help", append(returnsString, returnsString[0])},
 	{"importprivkey", nil},
+	{"importpubkey", nil},
+	{"importaddress", nil},
```
jrick: importaddress should come before importprivkey (here and everywhere else in the commit) so the methods remain alphabetical.

rpcserver.go
```diff
@@ -1341,6 +1341,8 @@ var rpcHandlers = map[string]struct {
 	// here because it hasn't been update to use the reference
 	// implemenation's API.
 	"getunconfirmedbalance":   {handler: GetUnconfirmedBalance},
+	"importpubkey":            {handler: ImportPubKey},
+	"importaddress":           {handler: ImportAddress},
```
jrick: alphabetical

waddrmgr/address.go
```diff
@@ -62,11 +65,11 @@ type ManagedPubKeyAddress interface {
 	ManagedAddress
 
 	// PubKey returns the public key associated with the address.
-	PubKey() *btcec.PublicKey
+	PubKey() (*btcec.PublicKey, error)
```
jrick: I think we should change the signature not to introduce an error (for the not-exists case) but to return the public key and a boolean for whether the pubkey is compressed or not.  Since watching addresses can be imported without the pubkey, not having the pubkey exist is now a totally valid state, and the caller must determine whether this is an error or not.

For the not-exists case, we would either need another (a second) bool return, or return a nil pubkey.  I'm leaning towards the additional return to avoid situations where the caller forgets to check for the nil case, since the Go type system isn't strong enough to enforce this.

waddrmgr/address.go
```diff
@@ -62,11 +65,11 @@ type ManagedPubKeyAddress interface {
...
 
 	// ExportPubKey returns the public key associated with the address
 	// serialized as a hex encoded string.
-	ExportPubKey() string
+	ExportPubKey() (string, error)
```
jrick: This may not be needed anymore with the PubKey method changes, since the caller would be able to create the exported key in compressed or uncompressed serialized formats themselves.

waddrmgr/address.go
```diff
@@ -113,6 +116,10 @@ var _ ManagedPubKeyAddress = (*managedAddress)(nil)
 // used by the caller without worrying about it being zeroed during an address
 // lock.
 func (a *managedAddress) unlock(key EncryptorDecryptor) ([]byte, error) {
+	if a.WatchingOnly() {
+		return nil, managerError(ErrWatchingOnly, errWatchingOnly, nil)
```
jrick: Are there any locations where managed addresses are unlocked before being checked to be watching-only?  If so, returning the error up the stack would be incorrect.

tuxcanfly: Don't think so, I see `unlock` being called from `scriptAddress.Script` and `managedAddress.PrivKey` both of which start with the check `if a.WatchingOnly() return nil, managerError(ErrWatchingOnly, errWatchingOnly, nil)`

waddrmgr/address.go
tuxcanfly: OK, we already have a db field called `compressed` on `managedAddress` which we are using for determining whether the pubkey is compressed or not. If we change the signature we might want to remove that field to avoid ambiguity.

Also, the `PubKey` signature is consistent with `PrivKey` which returns `ErrWatchingOnly` when private key is not available.

rpcserver.go
```diff
@@ -1820,6 +1826,51 @@ func GetUnconfirmedBalance(w *wallet.Wallet, chainSvr *chain.Client, icmd interf
...
+func ImportPubKey(w *wallet.Wallet, chainSvr *chain.Client, icmd interface{}) (interface{}, error) {
+	cmd := icmd.(*btcjson.ImportPubKeyCmd)
+
+	pubKeyBytes, err := decodeHexStr(cmd.PubKey)
+	if err != nil {
+		return nil, err
+	}
+
+	var compressed bool
+	switch n := len(pubKeyBytes); n {
```
jrick: I think this whole switch can be removed.  The correct length is checked by `ParsePubKey` below.  All you would need to pass to `ImportPubKey` is the bool `len(pubKeyBytes) == btcec.PubKeyBytesLenCompressed`.

I kind of think that the btcec type should save whether the pubkey was compressed or not, but oh well.

rpcserver.go
```diff
@@ -1820,6 +1826,51 @@ func GetUnconfirmedBalance(w *wallet.Wallet, chainSvr *chain.Client, icmd interf
...
+	case btcec.PubKeyBytesLenUncompressed:
+		compressed = false
+	case btcec.PubKeyBytesLenCompressed:
+		compressed = true
+	default:
+		e := errors.New("Invalid serialized pubkey length")
```
jrick: This error will be going away with the rest of the switch statement, but fwiw Go errors should not be capitalized (golint will complain).

The rpc server should probably capitalize all of them at a higher level though, to match the style used by core.

rpcserver.go
```diff
@@ -3112,8 +3163,13 @@ func ValidateAddress(w *wallet.Wallet, chainSvr *chain.Client, icmd interface{})
 
 	switch ma := ainfo.(type) {
 	case waddrmgr.ManagedPubKeyAddress:
+		// Make sure the public key is available, break out if it's not.
+		pubKey, err := ma.ExportPubKey()
+		if err != nil {
+			break
```
jrick: If we are not changing the signature of ExportPubKey (or PubKey), then we must explicitly check that the error is the one we expect, and return all others.  This is why I had mentioned returning an optional pubkey, but not erroring (because code like this ends up being written and real errors may be thrown away).

waddrmgr/db.go
jrick: Reminder to take care of this.  I don't particularly care for offset variables, but to be consistent with the rest of the code, it should create the offset var first, and then slice between `[offset:offset+fieldlen]`.

davecgh: They are superior to 4+4+8+len(thing)+x+y+len(z.foo).

jrick: With variable-sized data I agree

waddrmgr/db.go
```diff
@@ -902,10 +904,17 @@ func deserializeImportedAddress(row *dbAddressRow) (*dbImportedAddressRow, error
 		dbAddressRow: *row,
 	}
 
-	pubLen := binary.LittleEndian.Uint32(row.rawData[0:4])
+	pkHashLen := binary.LittleEndian.Uint32(row.rawData[0:4])
+	retRow.encryptedPubKeyHash = make([]byte, pkHashLen)
+	copy(retRow.encryptedPubKeyHash, row.rawData[4:4+pkHashLen])
```
jrick: For the record, this is not correctly bounds-checked, and may panic on a bad value.  These programming errors are all throughout this package though; we'll fix them up later.

rpcserver.go
tuxcanfly: Thanks, removed the swtich and simplified this.

rpcserver.go
tuxcanfly: Thanks, noted.

tuxcanfly:
OK, updated `PubKey` to return `compressed` as well as `ok`. Refactored related methods accordingly.

jrick:
The changed methods need better documentation to describe what all the return values are.  Either the interface definition can use named returns to document the values, or comments should be added to describe what each bool means.

tuxcanfly:
Thanks, using named returns in the interface and also updated docs about the return values.

csmcanarney:
What's the status on this? The corresponding PRs btcsuite/btcd#279 and btcsuite/btcjson#41 were merged a long time ago.

jrick:
There are several conflicts preventing this from being merged.  There will be even more conflicts if dcrwallet/#315 is backported to btcwallet.

We're also moving away from the idea of importing keys into a HD wallet, both to simplify our code and to make the entire wallet recoverable from just the seed.  Would need a discussion about if we still even want this feature.

PR #205, opened 2015-03-02
Use the correct synced block when calculating confs.

jrick:
This was previously using the most recently notified (by the chain
package) block, but transaction processing from this block may not be
finished yet.  Using this block's height to calculate the number of
confirmations is therefore incorrect, and can result in every RPC
handler missing transactions or returning transactions from the wrong
block.


davecgh:
OK

PR #207, opened 2015-03-02
Mark addresses used 

tuxcanfly:
Added a database flag on Address to mark used addresses and prevent address re-use.

Refs #196

### DIFF
--- a/waddrmgr/db.go
+++ b/waddrmgr/db.go
@@ -29,7 +29,7 @@ import (
 
 const (
 	// LatestMgrVersion is the most recent manager version.
-	LatestMgrVersion = 1
+	LatestMgrVersion = 2
 )
 
 var (
@@ -107,6 +107,7 @@ type dbAddressRow struct {
 	account    uint32
 	addTime    uint64
 	syncStatus syncStatus
+	used       bool
 	rawData    []byte // Varies based on address type field.
 }
 
@@ -162,6 +163,9 @@ var (
 
 	// Account related key names (account bucket).
 	acctNumAcctsName = []byte("numaccts")
+
+	// Used addresses (used bucket)
+	usedAddrBucketName = []byte("usedaddrs")
 )
 
 // uint32ToBytes converts a 32 bit unsigned integer into a 4-byte slice in
@@ -732,6 +736,17 @@ func serializeScriptAddress(encryptedHash, encryptedScript []byte) []byte {
 	return rawData
 }
 
+// fetchAddressUsed returns true if the provided address hash was flagged as used.
+func fetchAddressUsed(tx walletdb.Tx, addrHash [32]byte) bool {
+	bucket := tx.RootBucket().Bucket(usedAddrBucketName)
+
+	val := bucket.Get(addrHash[:])
+	if val != nil {
+		return true
+	}
+	return false
+}
+
 // fetchAddress loads address information for the provided address id from
 // the database.  The returned value is one of the address rows for the specific
 // address type.  The caller should use type assertions to ascertain the type.
@@ -749,6 +764,7 @@ func fetchAddress(tx walletdb.Tx, addressID []byte) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
+	row.used = fetchAddressUsed(tx, addrHash)
 
 	switch row.addrType {
 	case adtChain:
@@ -761,6 +777,23 @@ func fetchAddress(tx walletdb.Tx, addressID []byte) (interface{}, error) {
 
 	str := fmt.Sprintf("unsupported address type '%d'", row.addrType)
 	return nil, managerError(ErrDatabase, str, nil)
+}
+
+// markAddressUsed flags the provided address id as used in the database.
+func markAddressUsed(tx walletdb.Tx, addressID []byte) error {
+	bucket := tx.RootBucket().Bucket(usedAddrBucketName)
+
+	addrHash := fastsha256.Sum256(addressID)
+	val := bucket.Get(addrHash[:])
+	if val != nil {
+		return nil
+	}
+	err := bucket.Put(addrHash[:], []byte{0})
+	if err != nil {
+		str := fmt.Sprintf("failed to mark address used %x", addressID)
+		return managerError(ErrDatabase, str, err)
+	}
+	return nil
 }
 
 // putAddress stores the provided address information to the database.  This
@@ -1243,6 +1276,13 @@ func createManagerNS(namespace walletdb.Namespace) error {
 			return managerError(ErrDatabase, str, err)
 		}
 
+		// usedAddrBucketName bucket was added after manager version 1 release
+		_, err = rootBucket.CreateBucket(usedAddrBucketName)
+		if err != nil {
+			str := "failed to create used addresses bucket"
+			return managerError(ErrDatabase, str, err)
+		}
+
 		if err := putManagerVersion(tx, latestMgrVersion); err != nil {
 			return err
 		}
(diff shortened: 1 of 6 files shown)


davecgh:
I haven't looked very closely yet, but please define functions before they're used to be consistent with the rest of the code.

waddrmgr/db.go
```diff
@@ -1228,6 +1262,12 @@ func upgradeManager(namespace walletdb.Namespace) error {
 			return managerError(ErrDatabase, str, err)
 		}
 
+		_, err = rootBucket.CreateBucketIfNotExists(usedAddrBucketName)
```
davecgh: Please add a comment here that this bucket was added in version 2.

waddrmgr/db.go
```diff
@@ -744,6 +749,35 @@ func fetchAddress(tx walletdb.Tx, addressID []byte) (interface{}, error) {
...
+	err := bucket.Put(addrHash[:], []byte{0})
+	if err != nil {
+		str := fmt.Sprintf("failed to mark address used %x", addressID)
+		return managerError(ErrDatabase, str, err)
+	}
+	return nil
+}
+
+// fetchAddrUsed returns true if the provided address id was flagged as used.
+func fetchAddrUsed(tx walletdb.Tx, addressID []byte) bool {
```
davecgh: All of the other function names include 'Address'.  Let's name this `fetchAddressUsed` to be consistent.

waddrmgr/db.go
```diff
@@ -744,6 +749,35 @@ func fetchAddress(tx walletdb.Tx, addressID []byte) (interface{}, error) {
...
+	bucket := tx.RootBucket().Bucket(usedAddrBucketName)
+
+	addrHash := fastsha256.Sum256(addressID)
```
davecgh: Since this function is only called internally from `fetchAddress` and it already has the hash, I'd suggest changing the signature to `addrHash []byte` so we can avoiding needlessly calculating the hash again.

davecgh:
While I see there is code to make sure invoking `Used` on the address gives the expected result, I don't see any tests which actually mark an address used and expect it to be true.  Please add a test for this.

It should be done in the same fashion as the others such that it is marked and tested, the manager is closed and reopened, and it is tested again to ensure it is still marked used.  Note that all of that is already handled via the TestManager function, so you'd just want to add a `testMarkedUsed` function that is invoked from `testManagerAPI`, probably just before the `testChangePassphrase` call.  You can examine the `create` flag in the test context to determine whether the addresses should already be marked used or not.

waddrmgr/db.go
```diff
@@ -1268,8 +1308,41 @@ func upgradeManager(namespace walletdb.Namespace) error {
...
+
+// upgradeVersion1to2 upgrades the database from version 1 to version 2
+// 'usedAddrBucketName' a bucket for storing addrs flagged as marked is initialized
+// and it will be updated on the next rescan
+func upgradeVersion1to2(namespace walletdb.Namespace) error {
+	err := namespace.Update(func(tx walletdb.Tx) error {
+		rootBucket := tx.RootBucket()
+		mainBucket := tx.RootBucket().Bucket(mainBucketName)
+
+		_, err := rootBucket.CreateBucketIfNotExists(usedAddrBucketName)
```
davecgh: This is already done above, so I don't see any reason to do it again.

tuxcanfly: OK, I thought we might need this for migrating existing db but I guess `upgradeManager` takes care of that. If we don't need this, should we keep the existing version number since we don't need any upgrades?

waddrmgr/db.go
```diff
@@ -1268,8 +1308,41 @@ func upgradeManager(namespace walletdb.Namespace) error {
...
+		if err != nil {
+			str := "failed to create used addresses bucket"
+			return managerError(ErrDatabase, str, err)
+		}
+
+		var version uint32
+		var buf [4]byte
+		version = LatestMgrVersion
```
davecgh: This is wrong.  The function is explicitly to update version 1 to 2.  If you use the constant here, when it gets bumped to version 3, it would mess up the version 1 -> 2 -> 3 upgrade paths.

waddrmgr/db.go
```diff
@@ -1268,8 +1308,41 @@ func upgradeManager(namespace walletdb.Namespace) error {
 
 	// Upgrade the manager as needed.
 	if version < LatestMgrVersion {
-		// No upgrades yet.
+		// Upgrade addresses used flag
+		upgradeVersion1to2(namespace)
```
davecgh: this if check was meant as a placeholder.  It should look like:

```Go
if version < 2 {
	// Do the version 1 to 2 upgrade
	version = 2
}
if version < 3 {
	// Do the version 2 to 3 upgrade
	version = 3
}
....
```

waddrmgr/db.go
```diff
@@ -730,6 +745,8 @@ func fetchAddress(tx walletdb.Tx, addressID []byte) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
+	used := fetchAddressUsed(tx, addrHash)
```
davecgh: I don't think we need the extra local here:

`row.used = fetchAddressUsed(tx, addrHash)`

davecgh:
This will need to updated for #209.  Without handling that upcoming change, the database could be left without the bucket being created (even if this one went in first).

tuxcanfly:
@davecgh Found an issue when adding tests. The `used` flag is only loaded once on the first fetch, so if a `ManagedAddress` was marked used after it's in the cache, it will have stale data. I guess one way is to always query the db on `Used`. Or maybe delete the cache key on marking used. Will check and push a fix for this. Thanks.

davecgh:
@tuxcanfly:  I suspected there might be something like that.  Glad adding the tests found it.

waddrmgr/db.go
```diff
@@ -1266,10 +1303,102 @@ func upgradeManager(namespace walletdb.Namespace) error {
...
+	//		return err
+	//	}
+	//
+	//	// The manager is now at version 3.
+	//	version = 3
+	// }
+
+	if version < 2 {
+		// The manager is now at version 2.
+		version = 2
```
davecgh: I'd move this after the call to `upgradeToVersion2` as the sample shows.  No point in changing it only to change it back imo.

Also, the comment about there currently being no upgrades should be modified since it's no longer true.

waddrmgr/db.go
```diff
@@ -1266,10 +1303,102 @@ func upgradeManager(namespace walletdb.Namespace) error {
...
+		return managerError(ErrUpgrade, str, nil)
+	}
+
+	return nil
+}
+
+// upgradeToVersion2 upgrades the database from version 1 to version 2
+// 'usedAddrBucketName' a bucket for storing addrs flagged as marked is
+// initialized and it will be updated on the next rescan.
+func upgradeToVersion2(namespace walletdb.Namespace, currentMgrVersion uint32) error {
```
davecgh: The function's sole purpose to upgrade to version 2, so it shouldn't take a version.  I notice the code below is setting it to 2 manually (as expected).

waddrmgr/db.go
```diff
@@ -1243,6 +1276,13 @@ func createManagerNS(namespace walletdb.Namespace) error {
 			return managerError(ErrDatabase, str, err)
 		}
 
+		// usedAddrBucketName bucket was added after manager version 1 release
+		_, err = rootBucket.CreateBucketIfNotExists(usedAddrBucketName)
```
davecgh: CreateBucket

waddrmgr/db.go
```diff
@@ -1324,3 +1374,29 @@ func upgradeManager(namespace walletdb.Namespace) error {
 
 	return nil
 }
+
+// upgradeToVersion2 upgrades the database from version 1 to version 2
+// 'usedAddrBucketName' a bucket for storing addrs flagged as marked is
+// initialized and it will be updated on the next rescan.
+func upgradeToVersion2(namespace walletdb.Namespace) error {
```
davecgh: Please declare this before it's used (i.e before `upgradeManager`)

waddrmgr/manager_test.go
```diff
@@ -1016,6 +1017,32 @@ func testImportScript(tc *testContext) bool {
...
+func testMarkUsed(tc *testContext) bool {
+	expectedAddr1 := expectedAddr{
+		addressHash: hexToBytes("2ef94abb9ee8f785d087c3ec8d6ee467e92d0d0a"),
+		used:        true,
+	}
+	prefix := "MarkUsed"
+	chainParams := tc.manager.ChainParams()
+	addrHash := expectedAddr1.addressHash
+	addr, err := btcutil.NewAddressPubKeyHash(addrHash, chainParams)
+
```
davecgh: Please add a test here to make sure the address is not marked used before it's marked as used below.  This will also ensure that the cached address is properly updated.

davecgh:
Alright, latest changes look good.  I think the only thing left on this one is the requested additional tests.

tuxcanfly:
@davecgh Updated the test to make cover the initial unmarked case.

davecgh:
OK

PR #208, opened 2015-03-02
Update README for recent changes.

jrick:


README.md
```diff
@@ -8,28 +8,40 @@ btcwallet is a daemon handling bitcoin wallet functionality for a
...
+Unencrypted private keys are not supported and are never written to
+disk.  btcwallet uses the
+`m/44'/<coin type>'/<account>'/<branch>/<address index>`
+HD path for all derived addresses, as described by
+[BIP0044](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki).
+
+Due to the sensitive nature of public data in a BIP0032 wallet,
+btcwallet provides the option of encrypting not just private keys, but
+public data as well.  This is intended to thwart privacy risks where a
+wallet file is compromised without exposing all current and future
```
davecgh: Might want to reword this slightly to better indicate what it means.  As it is, it reads like the coins are at risk, which isn't the case.

jrick: suggestions?

davecgh: Perhaps something like:

This is intended to thwart privacy risks where a wallet file is compromised without exposing all current and future addresses *(public keys)* managed by the wallet.  While access to this information would not allow an attacker to spend or steal coins, it does mean they could track all transactions involving your addresses and therefore know your exact balance.

davecgh:
OK

PR #209, opened 2015-03-03
waddrmgr: Make create/update logic more explicit.

davecgh:
This pull request makes the creation and updating of the address manager more explicit so it's easier to upgrade in the future.

In particular, rather than treating the initial creation as an upgrade by relying on creating the initial buckets on the fly on each load, the code now explicitly provides distinct create and upgrade paths that are invoked from the `Create` and `Open` functions, respectively.

It also adds some commented out sample code to illustrate how upgrades should be done and a check to ensure bumping the version number without writing upgrade code results in a new error, `ErrUpgrade`, being returned.

Finally, a test has been added for the new functionality.

### DIFF
--- a/waddrmgr/internal_test.go
+++ b/waddrmgr/internal_test.go
@@ -33,6 +33,10 @@ import (
 // when tests are run.
 var TstMaxRecentHashes = maxRecentHashes
 
+// TstLatestMgrVersion makes the unexported latestMgrVersion variable available
+// for change when the tests are run.
+var TstLatestMgrVersion = &latestMgrVersion
+
 // Replace the Manager.newSecretKey function with the given one and calls
 // the callback function. Afterwards the original newSecretKey
 // function will be restored.
--- a/waddrmgr/manager_test.go
+++ b/waddrmgr/manager_test.go
@@ -1508,6 +1508,16 @@ func TestManager(t *testing.T) {
 	})
 	mgr.Close()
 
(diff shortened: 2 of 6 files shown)


waddrmgr/internal_test.go
```diff
@@ -33,6 +33,10 @@ import (
 // when tests are run.
 var TstMaxRecentHashes = maxRecentHashes
 
+// TstLatestMgrVersion makes the unexported latestMgrVersion variable availabe
+// for change the tests are run.
```
jrick: available is misspelled and I think you meant to say "when the tests are run".

waddrmgr/manager_test.go
```diff
@@ -1508,6 +1508,16 @@ func TestManager(t *testing.T) {
 	})
 	mgr.Close()
 
+	// Ensure the expected error is returned if the latest manager version
+	// constant is bumped without writing code to actually do the upgrade.
+	*waddrmgr.TstLatestMgrVersion++
+	_, err = waddrmgr.Open(mgrNamespace, pubPassphrase,
+		&chaincfg.MainNetParams, fastScrypt)
+	if !checkManagerError(t, "Upgade needed", err, waddrmgr.ErrUpgrade) {
```
jrick: Upgrade is misspelled

jrick:
Other than the typos this looks ok.

PR #211, opened 2015-03-04
Add new --createtemp flag to btcwallet

cjepson:
The createtemp flag is used to create a temporary simnet wallet for use with btcsim.
The flag only works if the datadir and network to use it on are specified.
After starting btcwallet with the flag, the wallet immediately opens and is functional.

### DIFF
--- a/config.go
+++ b/config.go
@@ -58,6 +58,7 @@ var (
 type config struct {
 	ShowVersion      bool     `short:"V" long:"version" description:"Display version information and exit"`
 	Create           bool     `long:"create" description:"Create the wallet if it does not exist"`
+	CreateTemp       bool     `long:"createtemp" description:"Create a temporary simulation wallet (pass=password) in the data directory indicated; must call with --datadir"`
 	CAFile           string   `long:"cafile" description:"File containing root certificates to authenticate a TLS connections with btcd"`
 	RPCConnect       string   `short:"c" long:"rpcconnect" description:"Hostname/IP and port of btcd RPC server to connect to (default localhost:18334, mainnet: localhost:8334, simnet: localhost:18556)"`
 	DebugLevel       string   `short:"d" long:"debuglevel" description:"Logging level {trace, debug, info, warn, error, critical}"`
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -366,6 +367,22 @@ func loadConfig() (*config, []string, error) {
...
+	// data directory.
+	if cfg.DataDir == defaultDataDir && cfg.CreateTemp {
+		fmt.Fprintln(os.Stderr, "Tried to create a temporary simulation "+
+			"wallet, but failed to specify data directory!")
+		os.Exit(0)
+	}
+
+	// Exit if you try to use a simulation wallet on anything other than
+	// simnet or testnet3.
+	if (!cfg.SimNet || cfg.MainNet) && cfg.CreateTemp {
```
davecgh: I think this just be `if !cfg.SimNet && cfg.CreateTemp {`.  The code above has already verified that only a single network is selected.  Also, the comment above should just say 'other than simnet.'

config.go
```diff
@@ -393,7 +410,7 @@ func loadConfig() (*config, []string, error) {
 		// Created successfully, so exit now with success.
 		os.Exit(0)
 
-	} else if !fileExists(dbPath) {
+	} else if !fileExists(dbPath) && cfg.CreateTemp == false {
```
davecgh: I'd reorganize this to simplify the conditions by checking the temp check first.  Also, no need to use `== false/true` as `if cfg.CreateTemp` or `if !cfg.CreateTemp` is the idiomatic approach.

```Go
if cfg.CreateTemp {
	// Do the temp creation bits
} else if cfg.Create {
	// Do the normal creation bits
} else if !fileExists(dbPath) {
	// Error bits.
}
```

walletsetup.go
```diff
@@ -438,3 +438,49 @@ func createWallet(cfg *config) error {
...
+	// Create the wallet database backed by bolt db.
+	db, err := walletdb.Create("bdb", dbPath)
+	if err != nil {
+		return err
+	}
+
+	// Create the address manager.
+	waddrmgrNamespace, err := db.Namespace(waddrmgrNamespaceKey)
+	if err != nil {
+		return err
```
davecgh: Database would not be properly shutdown here.  Use `defer db.Close()` just after the closing brace on `walletdb.Create`.

config.go
```diff
@@ -405,6 +422,31 @@ func loadConfig() (*config, []string, error) {
 		}
 		fmt.Fprintln(os.Stderr, err)
 		return nil, nil, err
+
+	} else if cfg.CreateTemp == true {
+		tempWalletExists := false
+
+		// Error if the wallet already exists.
+		if fileExists(dbPath) {
+			err := fmt.Errorf("The wallet already exists. Loading this " +
```
davecgh: This isn't an error.  Just call `fmt.Fprintf(os.Stderr, "This wallet .....\n")`

config.go
```diff
@@ -405,6 +422,31 @@ func loadConfig() (*config, []string, error) {
...
+				"wallet instead.")
+			fmt.Fprintln(os.Stderr, err)
+			tempWalletExists = true
+		}
+
+		if !tempWalletExists {
+			// Ensure the data directory for the network exists.
+			if err := checkCreateDir(netDir); err != nil {
```
davecgh: Is this actually needed?  I believe you've added code to ensure this temp wallet is required to be somewhere besides the normal data directory.

cjepson: If you don't call err := checkCreateDir(netDir), if the datadir is missing the wallet will never initialize creation (createSimulateWallet throws an error).

The fileExists calling is just a method for if someone screws up and tries to create a wallet in a directory that already has one; it's loading it as a default behaviour. It could also fail by default, however cfg.Create above would then have inconsistent behaviour.

config.go
```diff
@@ -366,10 +367,58 @@ func loadConfig() (*config, []string, error) {
...
+		err := fmt.Errorf("The flags --create and --createtemp can not " +
+			"be specified together. Use --help for more information.")
+		fmt.Fprintln(os.Stderr, err)
+		return nil, nil, err
+	}
+
+	if cfg.CreateTemp {
+		tempWalletExists := false
+
+		// Error if the wallet already exists.
```
davecgh: It looks like this comment is incorrect.

cjepson: Yeah, I guess it's not really an error (or even needed) as the string below clearly shows it continuing.

davecgh:
Looks good now.  Please squash it into a single commit and rebase to latest master and I'll get it merged.

PR #212, opened 2015-03-04
Set chain synced on empty rescan

tuxcanfly:
When an empty rescan is requested, the chain was not being marked as synced and this was preventing syncing afterwards because `w.ChainSynced()` would return `false`


jrick:
As discussed on IRC, this needs to remove the early return completely, so the address manager can be marked synced with the block height/hash from the rescanfinished notification.  This is extra work, but btcd can be optimized to send a rescanfinished and return early if there's nothing to rescan.

davecgh:
Why not just update the manager with the latest height instead of doing an RPC round trip?  Wallet should already be keeping track of this.

davecgh:
This has been rebased and merged as commit 62c7167504e5b4e47ea0b57f62a2d4c76254bae0.

PR #213, opened 2015-03-04
Refactoring to create a 'wallet' package

manan19:


jrick:
Please don't add a global wallet instance to the main package.  You can end up with races between reading and setting the pointer.  Instead, use the wallet pointers that are passed down to all of the handlers, after the wallet has been loaded and added to the server with `SetWallet`.

btcwallet.go
```diff
@@ -25,11 +25,13 @@ import (
 	"runtime"
 
 	"github.com/btcsuite/btcwallet/chain"
+	"github.com/mirrorx/btcwallet/wallet"
```
jrick: btcsuite

config.go
```diff
@@ -24,24 +24,23 @@ import (
...
 
 const (
 	defaultCAFilename        = "btcd.cert"
 	defaultConfigFilename    = "btcwallet.conf"
-	defaultBtcNet            = wire.TestNet3
 	defaultLogLevel          = "info"
 	defaultLogDirname        = "logs"
 	defaultLogFilename       = "btcwallet.log"
 	defaultDisallowFree      = false
+	defaultUseAddrIndex      = false
```
jrick: This should be in a different PR

config.go
```diff
@@ -49,6 +48,7 @@ var (
 	btcwalletHomeDir   = btcutil.AppDataDir("btcwallet", false)
 	btcdHomedirCAFile  = filepath.Join(btcdHomeDir, "rpc.cert")
 	defaultConfigFile  = filepath.Join(btcwalletHomeDir, defaultConfigFilename)
+	defaultPubPassphrase = "public"
```
jrick: This should probably be exported from another package.

Also, this file is not `go fmt`-formatted.

manan19:
This removes the global wallet ref
https://github.com/btcsuite/btcwallet/commit/87c18979ef4b4e2ca39c535f6086a2820f53c7ee

config.go
```diff
@@ -263,7 +255,7 @@ func loadConfig() (*config, []string, error) {
 	// Pre-parse the command line options to see if an alternative config
 	// file or the version flag was specified.
 	preCfg := cfg
```
jrick: This has changed semantics now because cfg is a pointer, not a `config`, so it doesn't make a shallow copy of the struct.  Is this intentional?

log.go
```diff
@@ -42,33 +34,19 @@ const (
...
 	log        = btclog.Disabled
+	walletLog    = btclog.Disabled
 	txstLog    = btclog.Disabled
 	chainLog   = btclog.Disabled
 )
 
 // subsystemLoggers maps each subsystem identifier to its associated logger.
 var subsystemLoggers = map[string]btclog.Logger{
-	"BTCW": log,
+	"BTCWD": log,
```
jrick: Try to keep all logging subsystems 4 chars long

wallet/config.go
```diff
@@ -0,0 +1,38 @@
...
+type Config struct {
+	DataDir          string
+	WalletPass       string
+	DisallowFree     bool
+	UseAddrIndex     bool
+	ChainParams		 *chaincfg.Params
+	WalletDbName     string
+}
+
+func (w *Wallet) GetConfig() (*Config) {
```
jrick: I would prefer if the config was just an exported member of the wallet, so it could be accessed directly without a getter.

wallet/wallet.go
```diff
@@ -1041,7 +1047,7 @@ func (w *Wallet) ImportPrivateKey(wif *btcutil.WIF, bs *waddrmgr.BlockStamp,
 
 // ExportWatchingWallet returns a watching-only version of the wallet serialized
 // in a map.
-func (w *Wallet) ExportWatchingWallet() (map[string]string, error) {
+func (w *Wallet) ExportWatchingWallet(walletPass string) (map[string]string, error) {
```
jrick: The public passphrase should be reachable via the config, through the wallet receiver.  No need to pass it as a parameter.

config.go
```diff
@@ -366,8 +358,17 @@ func loadConfig() (*config, []string, error) {
 		return nil, nil, err
 	}
 
+	walletConfig = &wallet.Config{
+		DataDir:				cfg.DataDir,
+		WalletPass:				cfg.WalletPass,
+		DisallowFree:			cfg.DisallowFree,
+		UseAddrIndex:			cfg.UseAddrIndex,
+		ChainParams:			activeNet.Params,
+		WalletDbName:			walletDbName,
```
jrick: The db filename shouldn't be a config option.  Instead, move the constant from the old config.go to the new package.

log.go
manan19: addressed

wallet/wallet.go
manan19: addressed

config.go
manan19: changed it back to how it was originally

config.go
manan19: Could you please elaborate? Do you want defaultPubPassphrase only to be exported from another package? 

go fmt - addressed.

config.go
manan19: removed it.

btcwallet.go
manan19: I'll change this at the very end - right before it's ready to merge.

wallet/config.go
manan19: 74fccec47b (export Config instead of GetConfig())

wallet/wallet.go
```diff
@@ -1305,13 +1311,13 @@ func (w *Wallet) TxRecord(txSha *wire.ShaHash) (r *txstore.TxRecord, ok bool) {
 	return nil, false
 }
 
-// openWallet opens a wallet from disk.
-func openWallet() (*Wallet, error) {
-	netdir := networkDir(cfg.DataDir, activeNet.Params)
-	dbPath := filepath.Join(netdir, walletDbName)
+// OpenWallet opens a wallet from disk.
+func OpenWallet(config *Config) (*Wallet, error) {
```
davecgh: This can probably be renamed to Open.  So callers would do `w, err := wallet.Open(cfg)`.

wallet/config.go
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+package wallet
+
+import (
+	"github.com/btcsuite/btcd/chaincfg"
+)
+
+const (
+	walletDbWatchingOnlyName = "wowallet.db"
```
davecgh: I think this should be caller defined (part of the Config) like the normal database name.

wallet/createtx.go
```diff
@@ -164,11 +165,13 @@ func createTx(
 	bs *waddrmgr.BlockStamp,
 	feeIncrement btcutil.Amount,
 	mgr *waddrmgr.Manager,
-	changeAddress func(*waddrmgr.BlockStamp) (btcutil.Address, error)) (
+	changeAddress func(*waddrmgr.BlockStamp) (btcutil.Address, error),
```
davecgh: Not sure what happened here with this function having one parameter per line, but this doesn't match the style of the rest of the code.  I realized you just followed suit, but now that I'm reviewing this and see it, if you would, please go ahead and fix the other params too so it's consistent with the rest of the code base.

wallet/disksync.go
```diff
@@ -23,7 +23,7 @@ import (
 
 // checkCreateDir checks that the path exists and is a directory.
 // If path does not exist, it is created.
-func checkCreateDir(path string) error {
+func CheckCreateDir(path string) error {
```
davecgh: I not sure if this should be exported from the package.  It's basic operations that I'm not really sure make much sense to be exposed from this package.

wallet/wallet.go
```diff
@@ -60,7 +65,7 @@ const (
 
 // networkDir returns the directory name of a network directory to hold wallet
 // files.
-func networkDir(dataDir string, chainParams *chaincfg.Params) string {
+func NetworkDir(dataDir string, chainParams *chaincfg.Params) string {
```
davecgh: I'd think I'd prefer the call is responsible for locating the wallet database.

davecgh:
In general, I'd like to see everything dealing with the creation of the wallet database and where it lives completely moved out of the new package.

It would be much cleaner to leave it up to the caller where to locate the database file, when to create a new one, etc.  The wallet package, imo, should just take a `walletdb.Db` that has already been opened.

wallet/wallet.go
```diff
@@ -1305,13 +1311,13 @@ func (w *Wallet) TxRecord(txSha *wire.ShaHash) (r *txstore.TxRecord, ok bool) {
 	return nil, false
 }
 
-// openWallet opens a wallet from disk.
-func openWallet() (*Wallet, error) {
-	netdir := networkDir(cfg.DataDir, activeNet.Params)
-	dbPath := filepath.Join(netdir, walletDbName)
+// OpenWallet opens a wallet from disk.
+func OpenWallet(config *Config) (*Wallet, error) {
+	netdir := NetworkDir(config.DataDir, config.ChainParams)
```
davecgh: As mentioned elsewhere, I'd prefer all of this be up to the caller and this function takes an already opened `walletdb.DB` to work with.

jrick:
I don't think passing just an opened DB to the wallet package is enough at the moment, since txstore is outside the db.

But we could pass both in, and then when wtxmgr goes live, introduce a breaking API change which removes the txstore option.

davecgh:
Agreed for btcwallet.  By making the caller responsible for the db, that is a choice we can make for btcwallet.  On the other hand, perhaps GUIs might want to support multiple wallets in the same directory, one wallet in multiple directories, or some combination thereof.  Making the caller responsible for such things offers the flexibility to do all of these things.  Having the wallet responsible for it like it is now doesn't.

config.go
jrick: It doesn't belong in the main package.  I would add it to waddrmgr and export it so the wallet package can use it.

wallet/createtx.go
manan19: a6202e166e (go fmt. reformat function declaration.)

wallet/wallet.go
manan19: b3d77f5fab (rename to remove api redundancy)

wallet/disksync.go
manan19: fd1ab5c0a1 (basic util should not be exported from package)

manan19:
Moved out wallet files creation as well as ownership outside the wallet package. The wallet database and txStore need to be passed in as parameters to wallet.Open

davecgh:
Status Update: We're currently working to get #155 and #204 merged to master.  This will need to be rebased and updated to handle those changes once they're merged.  After that we'll give this another round of review.

davecgh:
Status Update: #155 has been merged to master.  We're working on #204 now.  Once that is one, we'll work on this one.

wallet/README.md
```diff
@@ -0,0 +1,31 @@
+wallet
+========
```
davecgh: Too many `=`.  The other READMEs make these match.

wallet/README.md
```diff
@@ -0,0 +1,31 @@
+wallet
+========
+
+[![Build Status](https://travis-ci.org/btcsuite/btcwallet.png?branch=master)]
+(https://travis-ci.org/btcsuite/btcwallet)
+
+## Feature Overview
+
+
```
davecgh: Let's flesh this section out.

config.go
```diff
@@ -24,24 +24,33 @@ import (
...
+
+	// defaultPubPassphrase is the default public wallet passphrase which is
+	// used when the user indicates they do not want additional protection
+	// provided by having all public data in the wallet encrypted by a
+	// passphrase only known to them.
+	defaultPubPassphrase = "public"
+
+	// maxEmptyAccounts is the number of accounts to scan even if they have no
+	// transaction history. This is a deviation from BIP044 to make account
+	// creation more easier by allowing a limited number of empty accounts.
```
davecgh: more easier -> easier

wallet/README.md
jrick: The main feature of this package is the integration of all the other wallet components (waddrmgr, txstore, and chain) and the Wallet type is 'runnable', so it will be continuously updating itself against changes notified by the remote btcd instance.

wallet/disksync.go
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright (c) 2013, 2014 Conformal Systems LLC <info@conformal.com>
```
davecgh: New file -> 2015

wallet/wallet.go
```diff
@@ -1437,60 +1369,15 @@ func (w *Wallet) TxRecord(txSha *wire.ShaHash) (r *txstore.TxRecord, ok bool) {
...
-		// Special case: if the address manager was successfully read
-		// (mgr != nil) but the transaction store was not, create a
-		// new txstore and write it out to disk.  Write an unsynced
-		// manager back to disk so on future opens, the empty txstore
-		// is not considered fully synced.
-		if mgr == nil {
-			return nil, err
-		}
+// Returns wallet db being used by a wallet
+func (w *Wallet) GetDb() walletdb.DB {
```
davecgh: Please name this `.Db()` per [Effective Go Guidelines](http://golang.org/doc/effective_go.html#Getters).

walletsetup.go
```diff
@@ -484,3 +557,88 @@ func createSimulationWallet(cfg *config) error {
 	fmt.Println("The wallet has been created successfully.")
 	return nil
 }
+
+func openDb(directory string, dbname string) (
+	*walletdb.DB, error) {
```
davecgh: This should be on the line before it.

walletsetup.go
```diff
@@ -484,3 +557,88 @@ func createSimulationWallet(cfg *config) error {
 	fmt.Println("The wallet has been created successfully.")
 	return nil
 }
+
+func openDb(directory string, dbname string) (
```
davecgh: Please comment the intent, caveats, etc of this function.

walletsetup.go
```diff
@@ -484,3 +557,88 @@ func createSimulationWallet(cfg *config) error {
...
+
+	// Open the database using the boltdb backend.
+	db, err := walletdb.Open("bdb", dbPath)
+	if err != nil {
+		return nil, err
+	}
+	return &db, nil
+}
+
+func openWaddrmgr(db *walletdb.DB, namespaceKey []byte, pass string,
```
davecgh: Please comment the intent, caveats, etc of this function.

walletsetup.go
```diff
@@ -484,3 +557,88 @@ func createSimulationWallet(cfg *config) error {
...
+		ObtainSeed:        promptSeed,
+		ObtainPrivatePass: promptPrivPassPhrase,
+	}
+	// Open address manager and transaction store.
+	//	var txs *txstore.Store
+	return waddrmgr.Open(namespace, []byte(pass),
+		chainParams, config)
+}
+
+func openWallet() (*wallet.Wallet, error) {
```
davecgh: Please comment the intent, caveats, etc of this function.

log.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/btcsuite/btcwallet/chain"
 	"github.com/btcsuite/btcwallet/txstore"
 	"github.com/btcsuite/seelog"
+	"github.com/mirrorx/btcwallet/wallet"
```
davecgh: btcsuite

walletsetup.go
```diff
@@ -26,15 +26,88 @@ import (
...
+	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	"github.com/btcsuite/btcutil/hdkeychain"
 	"github.com/btcsuite/btcwallet/legacy/keystore"
+	"github.com/btcsuite/btcwallet/txstore"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/walletdb"
 	_ "github.com/btcsuite/btcwallet/walletdb/bdb"
 	"github.com/btcsuite/golangcrypto/ssh/terminal"
+	"github.com/mirrorx/btcwallet/wallet"
```
davecgh: btcsuite

rpcserver.go
```diff
@@ -48,6 +48,7 @@ import (
 	"github.com/btcsuite/btcwallet/txstore"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/websocket"
+	"github.com/mirrorx/btcwallet/wallet"
```
davecgh: btcsuite

davecgh:
Also, please create a `doc.go` file along the lines of https://github.com/btcsuite/btcwallet/blob/master/waddrmgr/doc.go.

davecgh:
All changes reviewed.  As discussed in IRC, please add TODOs in `doc.go` and `README.md` for the documentation to be completed by #224, and squash/rebase this to the latest master.  Then I'll be able to OK it and get it merged.

manan19:
@davecgh 
- Added 2 TODO's in wallet/doc.go and wallet/README.md. 
- Rebased, squashed into one descriptive commit and force pushed.

Travis build seems good, ready to be merged in.

Issue #215, opened 2015-03-24
waddrmgr: Optimize by using for each style functions for data lists


davecgh:
The `Manager` provided by the waddrmgr package currently has various functions which return a slice of all associated data.  In particular, `ActiveAccountAddresses`, `AllAccountAddresses`, `AllAccounts`, and `AllActiveAddresses`.

While it is certainly convenient this way, these should be switched over to iterator style functions to enable better scalability.  As it stands, imagine if you have millions of active addresses.  Creating a slice of all of these takes a huge amount of memory and a non-trivial amount of time to generate.  In many cases, the caller might be skipping several of them anyways.  Accepting an iterator-style callback function to call with each entry, on the other hand, would allow iterating them one at a time therefore eliminate the need for huge memory allocations.

As an example, consider `AllActiveAddresses`.  The current signature is: `func (m *Manager) AllActiveAddresses() ([]btcutil.Address, error)`.  This should be changed to `func (m *Manager) ForEachActiveAddress(fn func(btcutil.Address) error) error`  This will allow the caller to in turn do:

```Go
err := m.ForEachActiveAddress(func(addr btcutil.Address) error {
	// Do something with each addr, possibly returning an error to stop
	// iteration and return the error from AllActiveAddresses.
	return nil
})
if err != nil {
	// Handle it.
}
```

In regards to implementation, to provide better efficiency for these, I would suggest the database should maintain appropriate indices for each of these in separate buckets.  Then, it is a simple matter of doing a `bucket.ForEach` on the appropriate bucket, creating the necessary type from the underlying raw data, and invoking the caller-provided function with the type instance.

jrick:
Fixed by fbf744bc5e0cde6a0febf8772b32af320f6454a1.

PR #216, opened 2015-03-26
waddrmgr: optimize using each style functions

tuxcanfly:
refs #215 

WIP: Trying out the approach by replacing `AllAccounts` with a `each` style func first.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -370,7 +370,11 @@ func (w *Wallet) SetChainSynced(synced bool) {
 // outputs.  This is primarely intended to provide the parameters for a
 // rescan request.
 func (w *Wallet) activeData() ([]btcutil.Address, []wtxmgr.Credit, error) {
-	addrs, err := w.Manager.AllActiveAddresses()
+	var addrs []btcutil.Address
+	err := w.Manager.ForEachActiveAddress(func(addr btcutil.Address) error {
+		addrs = append(addrs, addr)
+		return nil
+	})
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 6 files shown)


davecgh:
This looks like the discussed approach.  However, unless we plan to add the ability to allow inactive accounts, the name should probably be `ForEachAccount` instead of `ForEachActiveAccount`.

tuxcanfly:
Cool, let me correct the func names.

davecgh:
Looks good.  I know there are more coming, so I won't merge it until everything is done.

tuxcanfly:
Thanks, will be working on the remaining funcs.

tuxcanfly:
Updated the remaining funcs. Working on optimizing storage for active/inactive addresses as discussed.

tuxcanfly:
Refactored to store all active addresses in a bucket and use it as an index while iterating over active addresses.

Added upgrade path for  migrating from used bucket to active bucket. Running some checks, will be up for a review in a bit.

tuxcanfly:
To avoid confusion, we should stick to differentiating active/inactive addresses. I've replaced all references to `used` with `inactive`.

tuxcanfly:
OK, ready for review. @davecgh

tuxcanfly:
Also renamed a few methods for consistency e.g `AddrAccount` to `AddressAccount`

tuxcanfly:
Depends on #238 in the way we will treat used addresses and this PR optimizes in addition to each-style functions.

tuxcanfly:
As discussed on IRC, an address can expire after a duration and become inactive while still remaining unused. Therefore, instead of assuming that addr not present in active bucket is used, we should have an explicit db flag on address row using a flag field which determines whether the address is used or not. Similarly, the address should be marked expired after a certain configurable duration.

tuxcanfly:
Limiting this PR to `ForEach` impleementations, created #254 for implementing active/inactive addresses, which I'll be working on.

wallet/wallet.go
```diff
@@ -693,20 +697,14 @@ func (w *Wallet) ChangePassphrase(old, new []byte) error {
...
-			return true, nil
-		}
-	}
-	return false, nil
+	var used bool
+	var err error
+	_ = w.Manager.ForEachAccountAddress(account,
+		func(maddr waddrmgr.ManagedAddress) error {
+			used, err = maddr.Used()
+			return nil
```
jrick: This throws away the error and re-assigns the possibly true used value instead of breaking out of the loop and returning true to the caller.

jrick:
ActiveAccountAddresses wasn't switched over to the new style and I'd like to see some more details in each function comments about how it executes the passed function on each item, breaking early is an error is returned.  Just saying "it iterates over X" isn't descriptive enough.

tuxcanfly:
Moved `ActiveAccountAddresses` to `ForEachActiveAccountAddress` and updated the docs to be more descriptive. Fixed `AccountUsed` to break out and return true on first used addr.

jrick:
All of the calls to maybeConvertDbError will mangle the break error.  I think that errors from this package should be converted, but errors from a caller's callback should not be.

tuxcanfly:
Thanks, as discussed, I've added a global err `Break` to signal a early exit from the callback. Please review.

jrick:
ok

tuxcanfly:
Thanks

PR #217, opened 2015-03-31
Add wtxmgr package

jrick:
This change implements a scalable wallet transaction database with integrated spend tracking, based on walletdb.

Integration is being tracked in another PR: #234

This is not the final database change related to transactions.  In the future, wtxmgr will be an internal package used by waddrmgr to store transactions and manage output spend tracking per account.  However, in the interest of improving the current code on master in a timely manner, wtxmgr is currently being given its own namespace.  This matches with how the old transaction store was managed independent of addresses and accounts.

Known issues are marked with `TODO` comments.  I believe all blockers have been implemented or fixed at this point, and the only remaining TODOs deal with performance issues which can be improved later (such as creating more new byte slices than necessary or parsing more details than needed).

Checklist for merge:

- [x] Add equivalent of `txstore.Store.FindPreviousCredits` (needed for votingpool to fetch PkScripts)
- [x] ~~Make upgrade logic match that of newer waddrmgr~~ just nuke the bucket when this moves to waddrmgr
- [x] Increase test coverage
- [x] Add testable, runnable examples


jrick:
I'm currently adding APIs to query for all recorded details for a single transaction.  This is done under a single view.  At the moment the API looks like this:

```Go
type TxDetails struct {
        TxRecord
        Block   BlockMeta
        Credits []Credit
        Debits  []Debit
}

// TxDetails returns all saved details regarding a transaction.  In case of a
// hash collision, the most recent transaction with a matching hash is returned.
func (s *Store) TxDetails(txHash *wire.ShaHash) (*TxDetails, error) {
        var details *TxDetails
        err := scopedUpdate(s.namespace, func(ns walletdb.Bucket) error {
                var err error
                details, err = s.txDetails(ns, txHash)
                return err
        })
        return details, err
}

func (s *Store) txDetails(ns walletdb.Bucket, txHash *wire.ShaHash) (*TxDetails, error) {
        // First, check whether there exists an unmined transaction with this
        // hash.  Use it if found.
        v := existsRawUnmined(ns, txHash[:])
        if v != nil {
                return s.unminedTxDetails(ns, txHash, v)
        }

        // Otherwise, if there exists a mined transaction with this matching
        // hash, skip over to the newest and begin fetching all details.
        k, v := latestTxRecord(txHash)
        if v == nil {
                // not found
                return nil, nil
        }

        // Read k/v, lookup all matching credits, debits.

        return nil, nil
}

func (s *Store) unminedTxDetails(ns walletdb.Bucket, txHash *wire.ShaHash, v []byte) (*TxDetails, error) {
        // ...
}
```

This will only fetch details for one transaction at a time and is not suitable for fetching entire ranges of transactions, but perhaps some of this code can be reused for that.

jrick:
Since the Credit type already implemented includes details which are duplicates of the TxRecord, I'm going to add another type:

```Go
type CreditRecord struct {
        Index  uint32
        Spent  bool
        Change bool
}

type DebitRecord struct {
        Amount btcutil.Amount
        Index  uint32
}

type TxDetails struct {
        TxRecord
        Block   BlockMeta
        Credits []CreditRecord
        Debits  []DebitRecorg
}
```

Further details about the transaction inputs and outputs can then be accessed by indexing the `wire.MsgTx` slices.

jrick:
API added to lookup rich details regarding a single transaction.  I'll update the integration branch to use this for the `gettransaction` RPC.

Longer term, we'll want to move ToJSON out of this package and use these types instead (but not the new method, since we know the block or if it's unmined).  In fact, it may make sense to do that now, considering ToJSON is known broken when passed an unconfirmed transaction.

jrick:
All of the remaining parts in the integration branch deal with the RPC server returning details about transaction history over ranges of transactions.  I am thinking of reusing the `TxDetails` type for an API to handle these situations as well, but instead of requesting one transaction at a time, a `[]TxDetails` will be returned, one block at a time, over some specified range of blocks.

I have something like this in mind:

```Go
func (s *Store) RangeTransactions(begin, end int32, func([]TxDetails)) error

err := w.TxStore.RangeTransactions(0, 300000, func(details []TxDetails) {
        // ...
})
```

This will allow iteration over transactions one block ta a time.  The height range is inclusive, and the special height -1 may be used as a high bound to include unmined transactinos.  The db view will be held for the entire duration of the call to RangeTransactions.  If necessary, I may also pass a context type to the function so it can return early without error.

One downside here is `TxDetails` includes a `BlockMeta`, which will be the same for every returned detail at a time.  This is inefficient and cumbersome to use, so maybe I'll split out the `BlockMeta` from `TxDetails` and also return a `BlockMeta` for the `Store.TxDetails` method.  Another downside is that the view may be held to for a long time as almost every detail for every  transaction in this range is returned and then processed by the caller.  While multiple views can run concurrently, writes will block, and this will stop all syncing until all the views return.  However, the alternative of returning details from both before and after a change is probably worse.

davecgh:
I'll take a look at this tomorrow.

jrick:
@davecgh informed me that, with bolt, a single Update may begin while Views are active.  This is because pages are COW.  So a long running view isn't as much of a problem as I explained above.

jrick:
APIs settled (for now :)) and integration branch is compiling.

davecgh:
Probably not for the PR, but we did discuss moving the whole legacy folder under internal so it's no longer externally accessible.

walletdb/bdb/db.go
```diff
@@ -167,6 +167,76 @@ func (b *bucket) Delete(key []byte) error {
 	return convertErr((*bolt.Bucket)(b).Delete(key))
 }
 
+// Cursor returns a new cursor, allowing for iteration over the bucket's
+// key/value pairs and nested buckets in forward or backward order.
+//
+// This function is part of the walletdb.Bucket interface implementation.
+func (b *bucket) Cursor() walletdb.Cursor {
```
davecgh: Can we break all of this Cursor functionality into a separate PR?

wtxmgr/db.go
```diff
@@ -0,0 +1,1311 @@
...
+		str := fmt.Sprintf("balance: short read (expected 8 bytes, "+
+			"read %v)", len(v))
+		return 0, storeError(ErrData, str, nil)
+	}
+	return btcutil.Amount(byteOrder.Uint64(v)), nil
+}
+
+func putMinedBalance(ns walletdb.Bucket, amt btcutil.Amount) error {
+	v := make([]byte, 8)
+	byteOrder.PutUint64(v, uint64(amt))
```
davecgh: In the grand scheme, this doesn't matter much, but the reason for BigEndian makes sense when doing cursor scans on keys, but shouldn't the value themselves be LittleEndian since there is an extremely high probability this will be running on a LE host arch.

jrick: Maybe.  The only reason I'd see this mattering is if you wanted to optimize the call to `binary.LittleEndian.PutUintXX` on little endian hosts, but it doesn't do that at the moment (at least in the source code, I haven't checked compiler output).

walletdb/bdb/db.go
jrick: yep I'll split it out, let's get that merged first.

davecgh:
One thing I noticed in the db code is that it's using a ton of magic numbers.  They are commented so it's fairly easy to follow, but I suspect it could lead to being a pain to upgrade if a new field gets added and such so that, say the 44 in the block records needs to become 48.  I'd make a few constants like `const blkRecTxOffset = 44`.

jrick:
I considered that approach as well but then I have to keep a thousand different names in my head, instead of just consulting a single comment and calculating offsets in my head (which I don't find hard, perhaps others disagree).

wtxmgr/db.go
```diff
@@ -0,0 +1,1311 @@
...
+	return blockIterator{c: c, seek: seek}
+}
+
+// Works just like makeBlockIterator but will initially position the cursor at
+// the last k/v pair.  Use this with blockIterator.prev.
+func makeReverseBlockIterator(ns walletdb.Bucket) blockIterator {
+	seek := make([]byte, 4)
+	byteOrder.PutUint32(seek, ^uint32(0))
+	c := ns.Bucket(bucketBlocks).Cursor()
+	return blockIterator{c: c}
```
davecgh: `return blockIterator{c: c, seek:seek}`

davecgh:
Since the code is making assumptions about the size of the hash everywhere, I think it might be worthwhile to add an init panic if it's not.  Something like:

```Go
func init() {
	if wire.HashSize != 32 {
		panic("hash size is not 32 bytes as expected.")
	}
}
```

jrick:
I prefer static assertions for these cases, but sure, an init check is fine.

davecgh:
Agreed a static assertion would be better.  This should do the trick:

```Go
var _ [32]byte = wire.ShaHash{}
```

wtxmgr/db.go
```diff
@@ -0,0 +1,1311 @@
...
+	byteOrder.PutUint32(k[32:36], uint32(block.Height))
+	copy(k[36:68], block.Hash[:])
+	return k
+}
+
+func valueTxRecord(rec *TxRecord) ([]byte, error) {
+	var v []byte
+	if rec.SerializedTx == nil {
+		txSize := rec.MsgTx.SerializeSize()
+		v = make([]byte, 8, 8+txSize)
```
davecgh: :+1: Nice way here to avoid an extra copy.

wtxmgr/db.go
```diff
@@ -0,0 +1,1311 @@
...
+		str := "failed to delete unmined input"
+		return storeError(ErrDatabase, str, err)
+	}
+	return nil
+}
+
+// upgradeManager opens the tx store using the specified namespace or creates
+// and initializes it if it does not already exist.  It also provides
+// facilities to upgrade the data in the namespace to newer versions.
+func upgradeManager(namespace walletdb.Namespace) error {
```
davecgh: Everywhere else refers to it as the txstore.  It seems that `upgradeStore` would be a more natural name for this.

davecgh: Also, it looks like this was based on the old waddrmgr upgrade code.  It was changed a bit to explicitly separate the open/create functionality.  It now creates the buckets at creation time instead of lazily checking/creating them on open each time.  That approach also makes the upgrade process easier to handle since new buckets are created specifically in a certain version when doing upgrades and well be created by default for all newly managers at the current version.

wtxmgr/db.go
```diff
@@ -0,0 +1,1311 @@
...
+		rbErr := tx.Rollback()
+		if rbErr != nil {
+			str := "rollback failed"
+			return storeError(ErrDatabase, str, rbErr)
+		}
+		return err
+	}
+	err = tx.Commit()
+	if err != nil {
+		str := "commit failed"
```
davecgh: We need to check how bolt handles a commit failure.  I think it might require a call to `.Rollback` in this scenario to release the page references associated with it, but I'm not 100% sure on that.

davecgh:
I've finished reviewing the database code and it looks really solid overall.  As mentioned before, I think it's a little heavy on the magic numbers, but the comments regarding the indices are good enough to follow what's going on.

From an implementation perspective, it provides really nice properties for highly scalable access.  In particular, I like the how the transaction, credits, and debit record keys share the same prefix so prefix scans can be done to efficiently find all relevant records across buckets.

wtxmgr/doc.go
```diff
@@ -0,0 +1,97 @@
...
+// the backend for storing the serialized transaction objects in buckets.
+//
+// Transaction outputs which are spendable by wallet keys are called credits
+// (because they credit to a wallet's total spendable balance) and are modeled
+// using the Credit structure.  Transaction inputs which spend
+// previously-inserted credits are called debits (because they debit from the
+// wallet's spendable balance) and are modeled using the Debit structure.
+//
+// Besides just saving transactions, bidirectional spend tracking is also
+// performed on each credit and debit.  Unlike packages such as btcdb, which
```
davecgh: This is true now, but, as you know, btcdb is currently being rewritten where it doesn't do any spend tracking on its own.  Might want to avoid mentioning it here since this will no longer be true in a relatively short while.

wtxmgr/doc.go
```diff
@@ -0,0 +1,97 @@
...
+//
+// Finally, this package records transaction insertion history (such as the date
+// a transaction was first received) and is able to create the JSON reply
+// structure for RPC calls such as listtransactions for any saved transaction.
+//
+// To use the transaction store, a transaction must be first inserted with
+// InsertTx.  After an insert, credits and debits may be attached to the
+// returned transaction record using the AddCredit and AddDebits methods.
+//
+// Example use:
```
davecgh: I'd suggest creating runnable example via an `example_test.go` file (see [btcjson's examples](https://github.com/btcsuite/btcd/blob/master/btcjson/v2/btcjson/example_test.go) for reference).  Examples in doc.go almost always end up out-of-date and non-functional over time.  By creating them as runnable examples, they show up under the examples category on godoc and also have the benefit of being executed when go test runs along with comparing their output to the expected output.

davecgh:
I'd personally prefer the JSON bits to live outside of the txstore.  That is specific to the RPC server.

davecgh:
This will need a `README.md` before you merge it as well.

jrick:
The JSON stuff will move.  It's here now because that's how txstore did it.  The code in the PR currently needed it in this package so additional details could be queried out of the db as necessary, but in my current tree I've switched this to be a method of `TxDetails` and it no longer needs the Store.  Since fields in the `TxDetails` are all exported, I plan on moving this totally out of this package and into rpcserver.go, or perhaps the `wallet` package if it's needed there (that package also includes too much stuff specific to the rpc server).

wtxmgr/doc.go
jrick: This entire file was lifted from the old implementation and needs to be rewritten.

davecgh:
I noticed the same thing about wallet yesterday will looking through it again as well (too much RPC server stuff)

wtxmgr/db.go
jrick: https://github.com/boltdb/bolt/blob/6d043164a90129f792615ee964f7b0956c2d38b7/db.go#L481-513

If the commit fails (and the tx's db pointer is not nil, I'm not sure on the exact circumstances where that is true) there is no call to rollback.

jrick: https://github.com/boltdb/bolt/blob/6d043164a90129f792615ee964f7b0956c2d38b7/tx.go#L131

Rollbacks are performed by Commit on all error paths.

jrick:
@gsalgado What's the plan for txstore and wtxmgr in votingpool?

gsalgado:
I have to admit I was hoping I'd be able to migrate from using txstore to
wtxmgr by simply doing a find and replace.

jrick:
@gsalgado I'll take a quick look through the votingpool package and see if all of the necessary functionality remains.  While the internal logic is quite similar to txstore, the external API has been reorganized slightly, so while it won't be a find/replace change, I don't expect it to be too bad.

jrick:
@gsalgado Many of the tests seem to be using txstore just for creating credits.  The `wtxmgr.Credit` structure now contains only exported fields, so they can be mocked directly instead of creating them from the store.  See how I modified `wallet/createtx_test.go` on my integration branch for an example.

jrick:
@gsalgado I think the only obvious omission at this point is an equivalent to `FindPreviousCredits`.

gsalgado:
I'll have a look at mocking wtxmgr.Credit in the votingpool tests. Also, one of my pending PRs (#221) uses TxRecord.AddDebits() which seems to be gone; is that now taken care of by InsertTx()?

jrick:
@gsalgado yes, it's automatic now.

jrick:
WRT test coverage, almost all non-error paths in the core store functionality (which after splitting queries, lives entirely in tx.go and unconfirmed.go) are covered.  Some edge cases remain: coinbase transactions and handling blocks with multiple recorded transactions.

jrick:
Happy with the test coverage now.  Testable examples still need to be added but those shouldn't be depended upon for test coverage, so I don't expect it to increase anymore at this point.

jrick:
Ready for another round of review.  Everything since the last review is squashed into a single commit.

jrick:
Rebased over hashing API changes.

wtxmgr/db.go
```diff
@@ -0,0 +1,1401 @@
...
+//   it := makeCreditIterator(ns, prefix)
+//   for it.next() {
+//           // Use it.elem
+//           // If necessary, read additional details from it.ck, it.cv
+//   }
+//   if it.err != nil {
+//           // Handle error
+//   }
+//
+// The elem's Spent field is not set to true if the credits is spent by an
```
davecgh: credits -> credit

wtxmgr/db.go
```diff
@@ -0,0 +1,1401 @@
...
+// unmined transaction.  To check for this case:
+//
+//   k := canonicalOutPoint(&txHash, it.elem.Index)
+//   it.elem.Spent |= existsRawUnminedInput(ns, k) != nil
```
davecgh: |= should just be =

wtxmgr/db.go
```diff
@@ -0,0 +1,1402 @@
...
+		byteOrder.PutUint32(v, LatestVersion)
+		err := ns.Put(rootVersion, v)
+		if err != nil {
+			str := "failed to store latest database version"
+			return storeError(ErrDatabase, str, err)
+		}
+
+		// Save the creation date of the store if it does not already
+		// exist.
+		v = ns.Get(rootCreateDate)
```
davecgh: No need to check for this.  The code above already proved the namespace is empty.

wtxmgr/db.go
```diff
@@ -0,0 +1,1402 @@
...
+			byteOrder.PutUint64(v, uint64(time.Now().Unix()))
+			err := ns.Put(rootCreateDate, v)
+			if err != nil {
+				str := "failed to store database creation time"
+				return storeError(ErrDatabase, str, err)
+			}
+		}
+
+		// Write a zero balance if it does not already exist.
+		v = ns.Get(rootMinedBalance)
```
davecgh: Same here.

wtxmgr/db.go
```diff
@@ -0,0 +1,1395 @@
...
+//   b:  The primary bucket being operated on
+//   k:  A single bucket key
+//   v:  A single bucket value
+//   c:  A bucket cursor
+//   ck: The current cursor key
+//   cv: The current cursor value
+//
+// Functions use the naming scheme `Op[Raw]Type[Field]`, which performs the
+// operation `Op` on the type `Type`, optionally dealing with raw keys and
+// values if `Raw` is used.  Fetch and extract operations may only need read
```
davecgh: may only need read -> may only need **to** read

wtxmgr/query.go
```diff
@@ -0,0 +1,462 @@
...
+		if v == nil {
+			return nil
+		}
+		details, err = s.txDetails(ns, txHash, k, v)
+		return err
+	})
+	return details, err
+}
+
+func (s *Store) txDetails(ns walletdb.Bucket, txHash *wire.ShaHash, recKey, recVal []byte) (*TxDetails, error) {
```
davecgh: This would be better defined before it's called by `TxDetails` and `UniqueTxDetails`.

wtxmgr/query.go
```diff
@@ -0,0 +1,462 @@
...
+		return nil, credIter.err
+	}
+
+	debIter := makeDebitIterator(ns, recKey)
+	for debIter.next() {
+		if int(debIter.elem.Index) >= len(details.MsgTx.TxIn) {
+			str := "saved debit index exceeds number of inputs"
+			return nil, storeError(ErrData, str, nil)
+		}
+
```
davecgh: Minor Nit: This one has a blank line before appending while the credits above doesn't.

wtxmgr/query.go
```diff
@@ -0,0 +1,462 @@
...
+		details.Debits = append(details.Debits, debIter.elem)
+	}
+	return &details, debIter.err
+}
+
+func (s *Store) unminedTxDetails(ns walletdb.Bucket, txHash *wire.ShaHash, v []byte) (*TxDetails, error) {
```
davecgh: This would also be better defined before it's called by `TxDetails` and `UniqueTxDetails`.

Also, the code in `TxDetails` and `UniqueTxDetails` invokes this function before `txDetails`, so it should ideally be defined before `txDetails` as well.

wtxmgr/query.go
```diff
@@ -0,0 +1,462 @@
...
+
+		brk, err := s.rangeBlockTransactions(ns, begin, end, f)
+		if err == nil && !brk && !addedUnmined && end < 0 {
+			_, err = s.rangeUnminedTransactions(ns, f)
+		}
+		return err
+	})
+}
+
+func (s *Store) rangeUnminedTransactions(ns walletdb.Bucket, f func([]TxDetails) (bool, error)) (brk bool, err error) {
```
davecgh: Same deal as above.  Better defined before `RangeTransactions` which uses it.

wtxmgr/query.go
```diff
@@ -0,0 +1,462 @@
...
+		details = append(details, *detail)
+		return nil
+	})
+	if err == nil && len(details) > 0 {
+		return f(details)
+	}
+	return false, err
+}
+
+func (s *Store) rangeBlockTransactions(ns walletdb.Bucket, begin, end int32, f func([]TxDetails) (bool, error)) (brk bool, err error) {
```
davecgh: Same deal as above.  Better defined before `RangeTransactions` which uses it.

davecgh: No need for the named returns here.

wtxmgr/query.go
davecgh: No need for the named returns here.

wtxmgr/unconfirmed.go
```diff
@@ -0,0 +1,172 @@
...
+	for _, input := range rec.MsgTx.TxIn {
+		prevOut := &input.PreviousOutPoint
+		k := canonicalOutPoint(&prevOut.Hash, prevOut.Index)
+		err = putRawUnminedInput(ns, k, rec.Hash[:])
+		if err != nil {
+			return err
+		}
+	}
+
+	// TODO: increment credit amount for each credit (but those are unknown
```
davecgh: Reminder.  There are several other TODOs, but they're only optimizations.  This is logic related.

wtxmgr/unconfirmed.go
```diff
@@ -0,0 +1,172 @@
...
+	})
+	return txs, err
+}
+
+func (s *Store) unminedTxs(ns walletdb.Bucket) ([]*wire.MsgTx, error) {
+	var unmined []*wire.MsgTx
+	err := ns.Bucket(bucketUnmined).ForEach(func(k, v []byte) error {
+		// TODO: Parsing transactions from the db may be a little
+		// expensive.  It's possible the caller only wants the
+		// serialized tranasctions.
```
davecgh: tranasctions -> transactions

wtxmgr/unconfirmed.go
```diff
@@ -0,0 +1,172 @@
...
+	var txs []*wire.MsgTx
+	err := scopedView(s.namespace, func(ns walletdb.Bucket) error {
+		var err error
+		txs, err = s.unminedTxs(ns)
+		return err
+	})
+	return txs, err
+}
+
+func (s *Store) unminedTxs(ns walletdb.Bucket) ([]*wire.MsgTx, error) {
```
davecgh: Define before use please.

wtxmgr/unconfirmed.go
```diff
@@ -0,0 +1,172 @@
...
+	var unmined []*wire.MsgTx
+	err := ns.Bucket(bucketUnmined).ForEach(func(k, v []byte) error {
+		// TODO: Parsing transactions from the db may be a little
```
davecgh: Perhaps provide a separate func `UnminedTxsRaw() ([][]byte, error) {`?

wtxmgr/unconfirmed.go
jrick: It's not, although that isn't very clear from just this comment.  It's referring to this TODO from AddCredits:

```
// TODO(jrick): This should not be necessary.  Instead, pass the indexes
// that are known to contain credits when a transaction or merkleblock is
// inserted into the store.
```

Ideally credits would be marked during the transaction insert, not after and in a different DB transaction.  It just doesn't work this way yet (do we want to finish that for this pr?)

wtxmgr/query.go
jrick: Both put a newline after a bounds check...

wtxmgr/unconfirmed.go
jrick: There aren't any callers that would be able to use this.  `btcrpcclient` forces transactions to be passed as a `*wire.MsgTx`.

wtxmgr/query.go
jrick: I personally find putting internal implementations at the top of files to be more unreadable (what is this?? why do we need this?!?), but eh, everyone has their preferred style.

wtxmgr/query.go
jrick: The named returns act as a form of documentation (`brk` being short for `break`, because that is a keyword).  When true, it breaks out of the range.

Are they necessary? no.  Do they hurt? no.  Would a comment be better? maybe.  But if you read the file in order you already have seen exactly *how* the function is being used :)

jrick: I'll make it a comment though, to avoid the pointless debate.

wtxmgr/query.go
davecgh: The vast majority of the code base does it that way, so I'm just aiming for consistency.  Any pros and cons aside, consistency trumps personal preferences.

wtxmgr/query.go
davecgh: Thanks.

For what it's worth, I had no idea that brk was intended to mean break when reading it until I read the implementation.  Had I been looking at that in godoc (which I often do locally with the internal flag enabled so I can see all unexported functions), I would not have known what it was for since there is no comment.

As you said, no reason to debate it, but I did want to point out that it did not have the desired effect.

wtxmgr/unconfirmed.go
davecgh: Alright.  Nah, let's get this merged.  There are a lot of other blocked PRs waiting on it, and we can do these minor optimizations later.

davecgh:
This looks ready to me.  OK.

wtxmgr/db.go
```diff
@@ -0,0 +1,1395 @@
...
+
+// appendRawBlockRecord returns a new block record value with a transaction
+// hash appended to the end and an incremented number of transactions.
+func appendRawBlockRecord(v []byte, txHash *wire.ShaHash) ([]byte, error) {
+	if len(v) < 44 {
+		str := fmt.Sprintf("%s: short read (expected %d bytes, read %d)",
+			bucketBlocks, 44, len(v))
+		return nil, storeError(ErrData, str, nil)
+	}
+	newv := append(v[:len(v):len(v)], txHash[:]...)
```
tuxcanfly: Can't this be simplified to `append(v, ...`?

davecgh: It would not be guaranteed to be new in that case.

http://play.golang.org/p/U9Xe_WHNmF

tuxcanfly: Thanks, got it.

jrick: This is necessary because otherwise you may be writing to invalid memory.  The `v` parameter in this case will often be the value bytes directly returned from bolt.  And since slices returned from bolt have caps greater than their lengths, by not limiting the cap during an append, you would end up writing to who-knows-where.  Best case scenario here would be a segfault.  Worst case would be silently continuing while corrupting the DB.

tuxcanfly: Thanks, yeah I realize now the third index is for limiting the cap.

PR #221, opened 2015-04-01
votingpool: API to store withdrawal txs in the txstore

gsalgado:
This is not yet used but will be needed for the implementation of the updatewithdrawal API, which will be coming soon.

~~Depends on #234~~


jrick:
Code and tests look ok.  It will need a rebase after intwtxmgr lands.

gsalgado:
Rebased on top of master now that intwtxmgr landed

PR #230, opened 2015-04-03
Stop mocking global func() variables in votingpool tests

gsalgado:
Doing that may cause erratic test failures when we run them in parallel, so
move the functions the tests need to mock as struct fields that are not
shared across tests.


jrick:
ok

davecgh:
Can you rebase this and deal withe conflict?  It doesn't merge cleanly at the moment.

gsalgado:
Sure, conflict solved and rebased

davecgh:
Thanks.  Merged.

PR #233, opened 2015-04-06
Use btcjson v2 types with the updated btcrpcclient.

jrick:
btcrpcclient commit 9ca93b30ad11ec34348d2d788c58019571bf9524 switched
the client to btcjson v2 and broke the wallet build.  This change is
the minimum amount of work to make wallet work again.

The rpc server still uses btcjson v1, so there will be 2 versions of
btcjson linked in the resulting binary.  This is not optimal but not
compiling is worse.

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -21,7 +21,7 @@ import (
 	"sync"
 	"time"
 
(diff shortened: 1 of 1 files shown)


davecgh:
OK

PR #234, opened 2015-04-06
Integrate wtxmgr

jrick:
~~Depends on #217.~~ merged

This contains the changes to integrate wtxmgr with the rest of the wallet code and removes the legacy txstore package.


jrick:
I'm going to add an extra utility that can be used to drop the wtxmgr namespace.  This will allow us to force rescans.  I'm hesitant to make this a feature of the rpc server, since this step really should never be necessary (but good idea to add it just in case).

wtxmgr/query.go
```diff
@@ -0,0 +1,462 @@
...
+	return &details, nil
+}
+
+// RangeTransactions runs the function f on all transaction details between
+// blocks on the best chain over the height range [begin,end].  The special
+// height -1 may be used to also include unmined transactions.  If the end
+// height comes before the begin height, blocks are iterated in reverse order
+// and unmined transactions (if any) are processed first.
+//
+// The function f may return an error which, if non-nil, is propigated to the
```
tuxcanfly: typo: propigated

wallet/chainntfns.go
```diff
@@ -110,69 +107,91 @@ func (w *Wallet) disconnectBlock(bs waddrmgr.BlockStamp) {
...
+					return err
+				}
+				log.Debugf("Marked address %v used", addr)
 				continue
 			}
-			_, err := txr.AddCredit(uint32(txOutIdx), false)
-			if err != nil {
+
+			// Missing addresses are skipped.  Other errors should
+			// be propigated.
```
tuxcanfly: same typo here :)

jrick:
Rebased over hashing API changes.

davecgh:
Since the new utility isn't using the same flags package as wallet and everything, it's inconsistent.  I think it's important from a usability aspect for everything in the suite to be consistent.  Can you change the util over to use go-flags?

docs/force_rescans.md
```diff
@@ -0,0 +1,81 @@
...
+During the rescan, relevant transactions from previously unseen blocks are added
+to the wallet database and spend tracking is updated accordingly.  After the
+rescan at startup finishes, a wallet is marked in sync with the chain.
+
+When wallet is started without any transaction history, a rescan is performed
+for all blocks since the creation date of the wallet's first address.  There are
+two situations when this holds true:
+
+1. The wallet is newly created or was recreated from the seed
+
```
davecgh: I believe markdown formats lists better if there is no newline between the list items.

docs/force_rescans.md
```diff
@@ -0,0 +1,81 @@
...
+btcwallet will not drop transaction history by itself, as this is something that
+should not be necessary under normal wallet operation.  However, a tool,
+`dropwtxmgr`, is provided in the `cmd/dropwtxmgr` directory which may be used to
+drop the wallet transaction manager (wtxmgr) history from a wallet database.
+The tool may already be installed in your PATH, but if not, installing it is easy:
+
+```
+$ cd $GOPATH/src/github.com/btcsuite/btcwallet/cmd/dropwtxmgr
+$ go get
```
davecgh: `go install`?  I know go get will install it too, but it also tries to do more.

davecgh:
Guide about dropping transactions is a very nice addition.

docs/force_rescans.md
jrick: [rendered](https://github.com/jrick/btcwallet/blob/jrick_intwtxmgr/docs/force_rescans.md)

docs/force_rescans.md
jrick: not without `-u`.

jrick:
I can convert the flags over.  I just used the stdlib because this is a dev tool that won't see much use, and the entire flags package API fits in my head.

log.go
```diff
@@ -52,7 +52,7 @@ var (
 var subsystemLoggers = map[string]btclog.Logger{
 	"BTCW": log,
 	"WLLT": walletLog,
-	"TXST": txstLog,
+	"TXST": txmgrLog,
```
davecgh: Not a big deal, but probably should make the prefix match the name.  If it's going to be called tx manager now, ideally it would be referred to that throughout.  How about `"TMGR"` or `"TXMR"`?

walletsetup.go
```diff
@@ -558,9 +559,30 @@ func createSimulationWallet(cfg *config) error {
...
+	} else {
+		if !fi.IsDir() {
+			return fmt.Errorf("path '%s' is not a directory", path)
+		}
+	}
+
+	return nil
+}
+
 // openDb opens and returns a *walletdb.DB (boltdb here) given the
```
davecgh: `*walletdb.DB` -> `walletdb.DB `

jrick: I didn't write this :)

And the comment is correct, it actually does return the interface by pointer (because this is how the db is passed to the wallet config).

log.go
jrick: Sure, I like TMGR.

walletsetup.go
davecgh: Unless my diff is wrong, you (correctly) changed the signature to not return a pointer to the interface.

```diff
-func openDb(directory string, dbname string) (*walletdb.DB, error) {
+func openDb(directory string, dbname string) (walletdb.DB, error) {
```

jrick: Oops, you are correct.  I was looking at code on master when I saw it was still the pointer (and it appeared in the diff view here that my changed stopped just before that line).

davecgh:
Everything else reads correctly.  I'm going to spend some time testing it, but I'm done with the code review.

davecgh:
Doing some testing and noticed this doesn't remove the old `tx.bin` on update.

jrick:
No harm in leaving it around, and it can always be manually deleted.  I'd rather avoid checking for an ancient file each and every startup for the rest of forever to clean up something that maybe a handful of people might have.

davecgh:
Alright, I've tested this pretty well.  I've also confirmed the latest commit fixed the issue reported in IRC.  I also tested the `dropwtxmgr` tool and it works.  As mentioned on IRC, the tool isn't really consistent with the flags and such, and that could be improved, but I don't see that as a reason to hold up this PR.

OK.

PR #235, opened 2015-04-07
export utility functions from wallet package

manan19:
This allows developers using the wallet package to display the transaction fees prior to sending the transaction

### DIFF
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -62,11 +62,15 @@ const (
 	txOutEstimate = 8 + 1 + pkScriptEstimate
 )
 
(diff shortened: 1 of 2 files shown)


jrick:
These functions are very specific to the assumptions in the const comments, and do not hold for all transactinos (or even all transactions that this wallet may end up signing).  They also use some types (such as `txstore.Credit`/`wtxmgr.Credit` that are specific to wallet's internal operations and may end up in an internal package sometime in the future.

I think these are small enough that you could copy them directly if you need them, modifying them to use whatever assumptions they use depending on the kinds of transactions you're creating.

PR #237, opened 2015-04-09
Upgrade path tests

tuxcanfly:
Added setup for testing version upgrades as well as tests for versions 2 and 3.

Also, refactored `Create` to remove some duplicated code by creating the initial db version and applying upgrades.

### DIFF
--- a/waddrmgr/db.go
+++ b/waddrmgr/db.go
@@ -695,12 +695,14 @@ func deleteAccountIDIndex(tx walletdb.Tx, account uint32) error {
 // putAccountNameIndex stores the given key to the account name index of the database.
 func putAccountNameIndex(tx walletdb.Tx, account uint32, name string) error {
 	bucket := tx.RootBucket().Bucket(acctNameIdxBucketName)
-
-	// Write the account number keyed by the account name.
-	err := bucket.Put(stringToBytes(name), uint32ToBytes(account))
-	if err != nil {
-		str := fmt.Sprintf("failed to store account name index key %s", name)
-		return managerError(ErrDatabase, str, err)
+	// Skip if we're working with the initial db version
+	if bucket != nil {
+		// Write the account number keyed by the account name.
+		err := bucket.Put(stringToBytes(name), uint32ToBytes(account))
+		if err != nil {
+			str := fmt.Sprintf("failed to store account name index key %s", name)
+			return managerError(ErrDatabase, str, err)
+		}
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


waddrmgr/db.go
```diff
@@ -695,12 +695,14 @@ func deleteAccountIDIndex(tx walletdb.Tx, account uint32) error {
 // putAccountNameIndex stores the given key to the account name index of the database.
 func putAccountNameIndex(tx walletdb.Tx, account uint32, name string) error {
 	bucket := tx.RootBucket().Bucket(acctNameIdxBucketName)
-
-	// Write the account number keyed by the account name.
-	err := bucket.Put(stringToBytes(name), uint32ToBytes(account))
-	if err != nil {
-		str := fmt.Sprintf("failed to store account name index key %s", name)
-		return managerError(ErrDatabase, str, err)
+	// Skip if we're working with the initial db version
```
jrick: Why would this function ever be run if the DB was not upgraded to the latest version?

waddrmgr/db.go
```diff
@@ -708,12 +710,14 @@ func putAccountNameIndex(tx walletdb.Tx, account uint32, name string) error {
 // putAccountIDIndex stores the given key to the account id index of the database.
 func putAccountIDIndex(tx walletdb.Tx, account uint32, name string) error {
 	bucket := tx.RootBucket().Bucket(acctIDIdxBucketName)
-
-	// Write the account number keyed by the account id.
-	err := bucket.Put(uint32ToBytes(account), stringToBytes(name))
-	if err != nil {
-		str := fmt.Sprintf("failed to store account id index key %s", name)
-		return managerError(ErrDatabase, str, err)
+	// Skip if we're working with the initial db version
```
jrick: See comment on `putAccountNameIndex`.

jrick:
I don't think this is the proper way of going about db upgrade tests.  It complicates the rest of the db code (as it must now not only work with the latest version, but handle earlier versions as well) and doesn't test the actual upgrade path that a real db will be updated from.  Instead, if we cared enough for these tests, I would recreate old db versions (possibly dumping them compressed to a data resources directory), load each one, test that the open failed due to the wrong version, and then explicitly perform the upgrade (using the same code paths as a real db), and then check that everything is as you expected.

Better yet, we should make each upgrade func perform some sanity checks after the upgrade is thought to be complete, and error (and rollback) if something is not correct.  This way the db is never erroneously upgraded (or rather, that there is a much lower chance of this happening).

tuxcanfly:
Thanks, yeah @davecgh mentioned the same thing i.e. having sanity tests after each upgrade. I'll close this issue but I'll make a note of the suggested approach to come back to it later.

Issue #231, opened 2015-04-03
Proposal: remove listaddresstransactions RPC


jrick:
Opening a discussion to remove this extension RPC.  I don't find the result format or the intended behavior useful, and creating result objects for it does not scale well.

The original issue (#27) mentioned that it may be useful for showing transactions which are relevant to some address, or a group of addresses.  However, the result format it uses (it borrows from listtransactions) is nearly useless since it doesn't show transactions (but some parsed details about the transaction output, and other details which can't be correctly determined from jut the transaction like the fee.  Oh, and why do individual outputs show their own fee??).

Besides the rant about the stupid result format, we have accounts for address grouping now, and querying for transactions should be done by the account level (or perhaps later, by tag) so these can tracked as wallet is syncing.  listaddresstransactions doesn't work this way, but instead must iterate over the wallet's entire transaction history from every account to find any relevant transactions.  This obviously scales very poorly.

I'd like to hear other's opinions of removing the extension.

arnuschky:
Querying the transactions for a subset of addresses is certainly useful for us, see #220. Of course, the result should then also include these transactions. 

Might be most powerful if combined with some sort of tagging feature. Accounts are very limiting.

PR #238, opened 2015-04-10
CurrentAddress - mark first address used and return new address

tuxcanfly:
Fixed `Last(In/Ex)ternalAddress` to make sure `CurrentAddress` returns a new address when the first and subsequent addresses are flagged as used.

### DIFF
--- a/waddrmgr/db.go
+++ b/waddrmgr/db.go
@@ -113,7 +113,6 @@ type dbAddressRow struct {
 	account    uint32
 	addTime    uint64
 	syncStatus syncStatus
-	used       bool
 	rawData    []byte // Varies based on address type field.
 }
 
@@ -987,17 +986,6 @@ func serializeScriptAddress(encryptedHash, encryptedScript []byte) []byte {
 	return rawData
 }
 
-// fetchAddressUsed returns true if the provided address hash was flagged as used.
-func fetchAddressUsed(tx walletdb.Tx, addrHash []byte) bool {
-	bucket := tx.RootBucket().Bucket(usedAddrBucketName)
-
-	val := bucket.Get(addrHash[:])
-	if val != nil {
-		return true
-	}
-	return false
-}
-
 // fetchAddressByHash loads address information for the provided address hash
 // from the database.  The returned value is one of the address rows for the
 // specific address type.  The caller should use type assertions to ascertain
@@ -1016,7 +1004,6 @@ func fetchAddressByHash(tx walletdb.Tx, addrHash []byte) (interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
-	row.used = fetchAddressUsed(tx, addrHash[:])
 
 	switch row.addrType {
 	case adtChain:
@@ -1029,6 +1016,14 @@ func fetchAddressByHash(tx walletdb.Tx, addrHash []byte) (interface{}, error) {
 
 	str := fmt.Sprintf("unsupported address type '%d'", row.addrType)
 	return nil, managerError(ErrDatabase, str, nil)
+}
+
+// fetchAddressUsed returns true if the provided address id was flagged as used.
+func fetchAddressUsed(tx walletdb.Tx, addressID []byte) bool {
+	bucket := tx.RootBucket().Bucket(usedAddrBucketName)
+
+	addrHash := fastsha256.Sum256(addressID)
+	return bucket.Get(addrHash[:]) != nil
 }
 
 // markAddressUsed flags the provided address id as used in the database.
--- a/waddrmgr/manager.go
+++ b/waddrmgr/manager.go
@@ -374,7 +374,7 @@ func (m *Manager) Close() error {
 // The passed derivedKey is zeroed after the new address is created.
 //
 // This function MUST be called with the manager lock held for writes.
-func (m *Manager) keyToManaged(derivedKey *hdkeychain.ExtendedKey, account, branch, index uint32, used bool) (ManagedAddress, error) {
+func (m *Manager) keyToManaged(derivedKey *hdkeychain.ExtendedKey, account, branch, index uint32) (ManagedAddress, error) {
 	// Create a new managed address based on the public or private key
 	// depending on whether the passed key is private.  Also, zero the
 	// key after creating the managed address from it.
@@ -397,7 +397,6 @@ func (m *Manager) keyToManaged(derivedKey *hdkeychain.ExtendedKey, account, bran
 	if branch == internalBranch {
 		ma.internal = true
 	}
-	ma.used = used
 
 	return ma, nil
 }
@@ -512,7 +511,7 @@ func (m *Manager) loadAccountInfo(account uint32) (*accountInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	lastExtAddr, err := m.keyToManaged(lastExtKey, account, branch, index, false)
+	lastExtAddr, err := m.keyToManaged(lastExtKey, account, branch, index)
 	if err != nil {
 		return nil, err
 	}
@@ -527,7 +526,7 @@ func (m *Manager) loadAccountInfo(account uint32) (*accountInfo, error) {
 	if err != nil {
 		return nil, err
 	}
-	lastIntAddr, err := m.keyToManaged(lastIntKey, account, branch, index, false)
+	lastIntAddr, err := m.keyToManaged(lastIntKey, account, branch, index)
 	if err != nil {
 		return nil, err
 	}
@@ -563,7 +562,7 @@ func (m *Manager) chainAddressRowToManaged(row *dbChainAddressRow) (ManagedAddre
 		return nil, err
 	}
 
-	return m.keyToManaged(addressKey, row.account, row.branch, row.index, row.used)
+	return m.keyToManaged(addressKey, row.account, row.branch, row.index)
 }
 
 // importedAddressRowToManaged returns a new managed address based on imported
@@ -590,7 +589,6 @@ func (m *Manager) importedAddressRowToManaged(row *dbImportedAddressRow) (Manage
 	}
 	ma.privKeyEncrypted = row.encryptedPrivKey
 	ma.imported = true
-	ma.used = row.used
 
 	return ma, nil
 }
@@ -605,7 +603,7 @@ func (m *Manager) scriptAddressRowToManaged(row *dbScriptAddressRow) (ManagedAdd
 		return nil, managerError(ErrCrypto, str, err)
 	}
 
-	return newScriptAddress(m, row.account, scriptHash, row.encryptedScript, row.used)
+	return newScriptAddress(m, row.account, scriptHash, row.encryptedScript)
 }
 
 // rowInterfaceToManaged returns a new managed address based on the given
@@ -1173,7 +1171,7 @@ func (m *Manager) ImportScript(script []byte, bs *BlockStamp) (ManagedScriptAddr
 	// since it will be cleared on lock and the script the caller passed
 	// should not be cleared out from under the caller.
 	scriptAddr, err := newScriptAddress(m, ImportedAddrAccount, scriptHash,
-		encryptedScript, false)
+		encryptedScript)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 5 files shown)


manan19:
This solves part of the problem I was facing.

I think the first generated external address should be also be added to first rescan job. 

Right now, despite calling .CurrentAddress, the first address isn't included in the rescan job.
Since it's not included in that list, it won't be marked as used, even though it has a balance on it. .CurrentAddress will keep returning the same address.

This issue might need to be fixed in the wallet package.

Also, IMO .NewAddress should be changed to .newAddress (un-exported)

jrick:
What about erroring on last in/external address if none have been created yet?  It should never have been creating new addresses to begin with.

manan19:
@jrick That will work too. It means that .CurrentAddress can implement the error handling for .LastExternalAddress and call .NewAddress.

jrick:
The rescan with 0 addresses is actually intentional because the response that btcwallet gets after the finished rescan is the block that the wallet is marked in sync with.

However, btcd should detect the 0 address 0 outpoint rescan and not actually load blocks and compare addresses against empty maps in that case.

waddrmgr/manager.go
```diff
@@ -1580,7 +1580,11 @@ func (m *Manager) LastExternalAddress(account uint32) (ManagedAddress, error) {
 		return nil, err
 	}
 
-	return acctInfo.lastExternalAddr, nil
+	if acctInfo.nextExternalIndex > 0 {
+		return acctInfo.lastExternalAddr, nil
+	}
+
+	return nil, managerError(ErrAddressNotFound, "address not found", nil)
```
jrick: Since this isn't looking up a particular address by string, hash160, etc., can you change the error description string to something like "no previous external address"?

Also, need a similar error for LastInternalAddress below.

jrick:
Please update the PR title and description, since this problem is being fixed by a different method.

tuxcanfly:
Yup, I agree about not creating addresses before they are required.

Updated `Last(In/Ex)ternalAddress` to err on first  address. Also had to update `loadAccountInfo` to query whether the addr is used and clear the cache after it's flagged.

waddrmgr/db.go
```diff
@@ -1031,6 +1019,18 @@ func fetchAddressByHash(tx walletdb.Tx, addrHash []byte) (interface{}, error) {
 	return nil, managerError(ErrDatabase, str, nil)
 }
 
+// fetchAddressUsed returns true if the provided address id was flagged as used.
+func fetchAddressUsed(tx walletdb.Tx, addressID []byte) bool {
```
jrick: As discussed on IRC, I'd like to see all usage of the phrase "address ID" replaced with something like "Hash160", since it is the ripemd160 after sha256 of the pubkey, or the script.  This is ok for now, but please fix this in another pr.

waddrmgr/db.go
```diff
@@ -1031,6 +1019,18 @@ func fetchAddressByHash(tx walletdb.Tx, addrHash []byte) (interface{}, error) {
...
+	bucket := tx.RootBucket().Bucket(usedAddrBucketName)
+
+	addrHash := fastsha256.Sum256(addressID)
+	val := bucket.Get(addrHash[:])
+	if val != nil {
```
jrick:
```Go
return bucket.Get(addrHash[:]) != nil
```

waddrmgr/manager.go
```diff
@@ -1360,6 +1360,16 @@ func (m *Manager) Unlock(passphrase []byte) error {
 	return nil
 }
 
+// fetchUsed returns true if the provided address id was flagged used.
+func (m *Manager) fetchUsed(addressID []byte) bool {
+	var used bool
+	_ = m.namespace.View(func(tx walletdb.Tx) error {
```
jrick: Shouldn't ignore the error here.

waddrmgr/manager.go
```diff
@@ -686,12 +686,12 @@ func (m *Manager) Address(address btcutil.Address) (ManagedAddress, error) {
 	return m.loadAndCacheAddress(address)
 }
 
-// AddrAccount returns the account to which the given address belongs.
-func (m *Manager) AddrAccount(address btcutil.Address) (uint32, error) {
+// AddrAccount returns the account to which the given address id belongs.
+func (m *Manager) AddrAccount(addressID []byte) (uint32, error) {
```
jrick: Why was this changed from taking the address to the hash160?

manan19:
The way .CurrentAddress is implemented it would generate only 1 .NewAddress and stop. I don't think it will work well, when a wallet is being restored from a seed.

1st address, 0 balance, 2 transactions
change address, 0.1 balance, 1 transaction
2nd ext address, 0.2 balance, 1 transaction

.CurrentAddress in this scenario would be expected to return the 3rd ext address. It should recursively generate a new address until it comes across an address with 0 balance and 0 transactions. 

(I've assumed change addresses are generated from a different branch and hence the numbers)

jrick:
CurrentAddress only exists to implement the currentaddress rpc.  It's ok if it's not terribly useful for recovering from a seed, since it wasn't designed for that use case.

waddrmgr/db.go
tuxcanfly: OK. I'm going to keep any new vars in this PR as addressID so I can find and replace all of them in the other PR.

waddrmgr/manager.go
tuxcanfly: Ok, the only err possible here is `ErrBucketNotFound` which shouldn't happen (since the bucket is created at setup) but some other walletdb backend might return another err.

jrick: Or the db is unreachable, or mmap fails, or.... :)

waddrmgr/manager.go
```diff
@@ -1360,15 +1363,31 @@ func (m *Manager) Unlock(passphrase []byte) error {
...
+	err := m.namespace.View(func(tx walletdb.Tx) error {
+		used = fetchAddressUsed(tx, addressID)
+		return nil
+	})
+	return used, err
+}
+
+// MarkUsed updates the used flag for the provided address.
+func (m *Manager) MarkUsed(address btcutil.Address) error {
+	addressID := address.ScriptAddress()
```
tuxcanfly: @jrick I think this was the external API call you mentioned which takes `addressID`

waddrmgr/manager.go
tuxcanfly: I see, will keep that in mind.

waddrmgr/manager.go
jrick: as long as it's a different type, finding and unbreaking callers is easy.

waddrmgr/manager.go
```diff
@@ -1360,15 +1363,31 @@ func (m *Manager) Unlock(passphrase []byte) error {
...
 	if err != nil {
 		return maybeConvertDbError(err)
 	}
-	// 'used' flag has become stale so remove key from cache
+	// Clear caches which might have stale entries for used addresses
+	account, err := m.AddrAccount(address)
+	if err != nil {
+		return err
+	}
+	delete(m.acctInfo, account)
```
tuxcanfly: @jrick This is a bit of a hack since `acctInfo` caches the derived last (ex/in) addresses. I guess a better solution would be to stop deriving addrs and always load them from the db.

davecgh: I'd advise against this.  It's actually faster to rederive them than it is to load and deserialize them.

tuxcanfly:
@jrick @davecgh I think the larger problem is that we're storing the mutable `used` in an otherwise immutable address and this leads to issues like the above where we need to be careful about clearing the caches. I just started looking at wtxmgr but if we could index transactions by address, it might be easier to check that we have no transactions belonging to the address and hence determine it's used or not.

waddrmgr/manager.go
tuxcanfly: OK, I'll leave it to derive the addrs for now.

jrick:
I think the used flag is still correct.  If a transaction is sent to some address, and the address is marked "used", then if that transaction is removed from wtxmgr (perhaps it was double spent), the address shouldn't become unused.  Once used, the address should never become unused again.

tuxcanfly:
Yeah, just checked wtxmgr and looks like it is ranging over all transactions.

OK, let me test this out and prep for review. Thanks.

jrick:
Where is it ranging transactions?  Only place I can remember it doing that and filtering by address is `Wallet.ListAddressTransactions`, which I want to remove (along with the RPC).  See #231

jrick:
Well, let's keep things simple.  Until waddrmgr and wtxmgr are properly combined (this will happen after the current wtxmgr integration is merged), we'll keep the transaction iterating to detect used-ness.  Once combined, we'll introduce this flag again.

tuxcanfly:
Yeah, I was checking if `ListAddressTransactions`  was filtering by address or if it was using an index.

waddrmgr/manager_test.go
```diff
@@ -1028,28 +1028,32 @@ func testMarkUsed(tc *testContext) bool {
 	addrHash := expectedAddr1.addressHash
 	addr, err := btcutil.NewAddressPubKeyHash(addrHash, chainParams)
 
+	maddr, err := tc.manager.Address(addr)
+	if err != nil {
+		tc.t.Errorf("%s: unexpected error: %v", prefix, err)
```
jrick: Can't continue with the rest of this test, either return here of make this fatal

waddrmgr/manager_test.go
```diff
@@ -1028,28 +1028,32 @@ func testMarkUsed(tc *testContext) bool {
...
+	}
 	if tc.create {
 		// Test that initially the address is not flagged as used
-		maddr, err := tc.manager.Address(addr)
+		used, err := maddr.Used()
 		if err != nil {
 			tc.t.Errorf("%s: unexpected error: %v", prefix, err)
```
jrick: return or fatal

waddrmgr/manager_test.go
```diff
@@ -1028,28 +1028,32 @@ func testMarkUsed(tc *testContext) bool {
...
+		if used != false {
 			tc.t.Errorf("%v: unexpected used flag -- got "+
-				"%v, want %v", prefix, maddr.Used(), expectedAddr1.used)
+				"%v, want %v", prefix, used, expectedAddr1.used)
 		}
 	}
-	err = tc.manager.MarkUsed(addrHash)
+	err = tc.manager.MarkUsed(addr)
 	if err != nil {
 		tc.t.Errorf("%s: unexpected error: %v", prefix, err)
```
jrick: return or fatal

waddrmgr/manager_test.go
```diff
@@ -1028,28 +1028,32 @@ func testMarkUsed(tc *testContext) bool {
...
 	}
-	maddr, err := tc.manager.Address(addr)
+	used, err := maddr.Used()
 	if err != nil {
 		tc.t.Errorf("%s: unexpected error: %v", prefix, err)
```
jrick: return or fatal

wallet/chainntfns.go
```diff
@@ -148,6 +148,9 @@ func (w *Wallet) addReceivedTx(tx *btcutil.Tx, block *txstore.Block) error {
 			if err != nil {
 				return err
 			}
+			if err := w.markAddrsUsed(txr); err != nil {
```
jrick: This belongs in a different pr (or wait for intwtxmgr, which doesn't have this bug)

wallet/wallet.go
```diff
@@ -720,7 +719,9 @@ func (w *Wallet) diskWriter() {
 // there are any transactions with outputs to this address in the blockchain or
 // the btcd mempool.
 func (w *Wallet) AddressUsed(addr waddrmgr.ManagedAddress) bool {
-	return addr.Used()
+	// TODO: handle err
```
jrick: Can you add this?

In fact, just remove this function.  The caller already has the managed address, so they can check this themselves.

waddrmgr/manager_test.go
tuxcanfly: Added `return false` to all unexpected errs.

wallet/wallet.go
tuxcanfly: OK, removed.

wallet/chainntfns.go
tuxcanfly: OK, reverted the fix.

jrick:
All of the waddrmgr changes look ok now but we need handling in Wallet.CurrentAddress so that if the error is because no addresses exist yet, it handles it by creating the first external address.

tuxcanfly:
Updated `CurrentAddress` to handle `ErrAddressNotFound` and generate a new address.

jrick:
Thanks, going to test this out but the diff looks ok now.  Can you squash?

tuxcanfly:
Sure, done.

jrick:
rebase on master too

tuxcanfly:
OK, rebased.

PR #239, opened 2015-04-11
Improve account related test coverage

tuxcanfly:
Added test coverage for account related manager funcs.

### DIFF
--- a/waddrmgr/common_test.go
+++ b/waddrmgr/common_test.go
@@ -53,6 +53,119 @@ var (
 
 	// waddrmgrNamespaceKey is the namespace key for the waddrmgr package.
 	waddrmgrNamespaceKey = []byte("waddrmgrNamespace")
+
+	// expectedAddrs is the list of all expected addresses generated from the
+	// seed.
+	expectedAddrs = []expectedAddr{
+		{
+			address:     "14wtcepMNiEazuN7YosWY8bwD9tcCtxXRB",
+			addressHash: hexToBytes("2b49ecd0cf72006173e6e95acf416b6735b5f889"),
+			internal:    false,
+			compressed:  true,
+			imported:    false,
+			pubKey:      hexToBytes("02d8f88468c5a2e8e1815faf555f59cbd1979e3dbdf823f80c271b6fb70d2d519b"),
+			privKey:     hexToBytes("c27d6581b92785834b381fa697c4b0ffc4574b495743722e0acb7601b1b68b99"),
+			privKeyWIF:  "L3jmpy54Pc7MLXTN2mL8Xas7BJziwKaUGmgnXXzgGbVRdiAniXZk",
+		},
+		{
+			address:     "1N3D8jy2aQuUsKBsDgZ6ZPTVR9VhHgJYpE",
+			addressHash: hexToBytes("e6c59a1542138d1bf08f45cd18899557cf56b356"),
+			internal:    false,
+			compressed:  true,
+			imported:    false,
+			pubKey:      hexToBytes("02b9c175b908624f8a8eaac227d0e8c77c0eec327b8c512ad1b8b7a4b5b676971f"),
+			privKey:     hexToBytes("18f3b191019e83878a81557abebb2afda199e31d22e150d8bf4df4561671be6c"),
+			privKeyWIF:  "Kx4DNid19W8sjNFN3uPqQE7UYnCqyEp7unCvdkf2LrVUFpnDtwpB",
+		},
+		{
+			address:     "1VTfwD4iHre2bMrR9qGiJMwoiZGQZ8e6s",
+			addressHash: hexToBytes("0561e9373986965b647a57a09718e9c050215cfe"),
+			internal:    false,
+			compressed:  true,
+			imported:    false,
+			pubKey:      hexToBytes("0329faddf1254d490d6add49e2b08cf52b561038c72baec0edb3cfacff71ff1021"),
+			privKey:     hexToBytes("ccb8f6305b73136b363644b647f6efc0fd27b6b7d9c11c7e560662ed38db7b34"),
+			privKeyWIF:  "L45fWF6Yd736fDohuB97vwRRLdQQJr3ZGvbokk9ubiT7aNrg7tTn",
+		},
+		{
+			address:     "13TdEj4ehUuYFiSaB47eLVBwM2XhAhrK2J",
+			addressHash: hexToBytes("1af950be02584ca230b7078cec0cfd38dd71b468"),
+			internal:    false,
+			compressed:  true,
+			imported:    false,
+			pubKey:      hexToBytes("03d738324e2f0ce42e46975d7f8c7117c1670e3d7912b0291aea452add99674774"),
+			privKey:     hexToBytes("d6bc8ff768814fede2adcdb74826bd846924341b3862e3b6e31cdc084e992940"),
+			privKeyWIF:  "L4R8XyxYQyPSpTwj8w96tM86a6j3QA9jbRPj3RA7DVTVWk71ndeP",
+		},
+		{
+			address:     "1LTjSghkBecT59VjEKke331HxVdqcFwUDa",
+			addressHash: hexToBytes("d578a267a7174c6ba7f76b0ab2397ce0ba0c5c3c"),
+			internal:    false,
+			compressed:  true,
+			imported:    false,
+			pubKey:      hexToBytes("03a917acd5cd5b6f544b43f1921a35677e4d5320e5d2add2056039b4b44fdf905e"),
+			privKey:     hexToBytes("8563ade061110e03aee50695ffc5cb1c06c8310bde0a3674257c853c966968c0"),
+			privKeyWIF:  "L1h16Hunxomww4FrpyQP2iFmWNgG7U1u3awp6Vd3s2uGf7v5VU8c",
+		},
+		{
+			address:     "15HNivzKhsLaMs1qRdQN1ifoJYUnJ2xW9z",
+			addressHash: hexToBytes("2ef94abb9ee8f785d087c3ec8d6ee467e92d0d0a"),
+			internal:    true,
+			compressed:  true,
+			imported:    false,
+			pubKey:      hexToBytes("020a1290b997c0a234a95213962e7edcb761c7360f0230f698a1a3e71c37047bb0"),
+			privKey:     hexToBytes("fe4f855fcf059ec6ddf7b25f63b19aa49c771d1fcb9850b68ae3d65e20657a60"),
+			privKeyWIF:  "L5k4HivqXvohxBMpuwD38iUgi6uewffwZny91ZNYfM39RXH2x3QR",
+		},
+		{
+			address:     "1LJpGrAP1vWHuvfHqmUutQqFVYca2qwxhy",
+			addressHash: hexToBytes("d3c8ec46891f599bfeaa4c25918bfb3d46ea334c"),
+			internal:    true,
+			compressed:  true,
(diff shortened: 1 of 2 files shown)


waddrmgr/manager_test.go
```diff
@@ -1019,38 +1019,64 @@ func testImportScript(tc *testContext) bool {
 
 // testMarkUsed ensures used addresses are flagged as such.
 func testMarkUsed(tc *testContext) bool {
-	expectedAddr1 := expectedAddr{
-		addressHash: hexToBytes("2ef94abb9ee8f785d087c3ec8d6ee467e92d0d0a"),
-		used:        true,
+	tests := []struct {
+		name       string
+		addrType   btcutil.Address
```
jrick: Is this field needed? doesn't look used and the address in the p2sh address test is actually a p2pkh addr.

waddrmgr/manager_test.go
```diff
@@ -1019,38 +1019,64 @@ func testImportScript(tc *testContext) bool {
...
+		},
 	}
+
 	prefix := "MarkUsed"
 	chainParams := tc.manager.ChainParams()
-	addrHash := expectedAddr1.addressHash
-	addr, err := btcutil.NewAddressPubKeyHash(addrHash, chainParams)
+	for i, test := range tests {
+		addrHash := test.expected.addressHash
+		addr, err := btcutil.NewAddressPubKeyHash(addrHash, chainParams)
```
jrick: This makes a p2pkh address from the "script address" test.

waddrmgr/manager_test.go
```diff
@@ -1019,38 +1019,64 @@ func testImportScript(tc *testContext) bool {
...
+				addressHash: hexToBytes("2ef94abb9ee8f785d087c3ec8d6ee467e92d0d0a"),
+				used:        true,
+			},
+		},
+		{
+			name:     "script address",
+			addrType: &btcutil.AddressPubKeyHash{},
+			in:       "3MbyWAu9UaoBewR3cArF1nwf4aQgVwzrA5",
+			expected: expectedAddr{
+				addressHash: hexToBytes("da6e6a632d96dc5530d7b3c9f3017725d023093e"),
```
jrick: This is input to the test, not the expected results.

waddrmgr/manager_test.go
```diff
@@ -1019,38 +1019,64 @@ func testImportScript(tc *testContext) bool {
...
+				used:        true,
```
jrick: Unnecessary for this test, it always first tests that it is unused, marks used, and then checks if it's marked used.

waddrmgr/manager_test.go
```diff
@@ -1019,38 +1019,64 @@ func testImportScript(tc *testContext) bool {
...
-			tc.t.Errorf("%s: unexpected error: %v", prefix, err)
+			tc.t.Errorf("%s #%d: unexpected error: %v", prefix, i, err)
 		}
-		if maddr.Used() != false {
-			tc.t.Errorf("%v: unexpected used flag -- got "+
-				"%v, want %v", prefix, maddr.Used(), expectedAddr1.used)
+		if maddr.Used() != test.expected.used {
+			tc.t.Errorf("%s #%d: unexpected used flag -- got "+
+				"%v, want %v", prefix, i, maddr.Used(), test.expected.used)
 		}
```
jrick: I would consider closing the db here, and reopening, and check that it is still marked as used.  Perhaps do the same shutdown dance before testing for initially unsued as well.

waddrmgr/manager_test.go
```diff
@@ -1222,12 +1248,12 @@ func testNewAccount(tc *testContext) bool {
...
 
-// testLookupAccount tests the basic account lookup func of the address manager works
-// as expected.
+// testLookupAccount tests the basic account lookup func of the address manager
+// works as expected.
 func testLookupAccount(tc *testContext) bool {
 	// Lookup accounts created earlier in testNewAccount
 	expectedAccounts := map[string]uint32{
-		waddrmgr.DefaultAccountName:      waddrmgr.DefaultAccountNum,
+		"":                               waddrmgr.DefaultAccountNum,
```
jrick: Shouldn't this remain the same?

waddrmgr/manager_test.go
```diff
@@ -1255,6 +1281,26 @@ func testLookupAccount(tc *testContext) bool {
 	if !checkManagerError(tc.t, testName, err, wantErrCode) {
 		return false
 	}
+
+	// Test address account lookup
+	var expectedAccount uint32
+	addressHash := hexToBytes("2ef94abb9ee8f785d087c3ec8d6ee467e92d0d0a")
```
jrick: Since the seed is part of the global test binary, I'd like to see all of these hardcoded expected addresses to either also be globals, or passed around as part of the test context.

waddrmgr/manager_test.go
tuxcanfly: Thanks, removed the unnecessary fields and made the tests concise.

waddrmgr/manager_test.go
tuxcanfly: Yeah, the test case was being affected by the `testRenameAccount`. Fixing it to test both cases independently.

waddrmgr/manager_test.go
tuxcanfly: Thanks, globals seems fine, will update the test cases.

waddrmgr/manager_test.go
```diff
@@ -1019,38 +1019,66 @@ func testImportScript(tc *testContext) bool {
...
+		var addr btcutil.Address
+		var err error
+		switch test.addrType.(type) {
+		case *btcutil.AddressPubKeyHash:
+			addr, err = btcutil.NewAddressPubKeyHash(addrHash, chainParams)
+		case *btcutil.AddressScriptHash:
+			addr, err = btcutil.NewAddressScriptHashFromHash(addrHash, chainParams)
+		}
+		if err != nil {
+			tc.t.Errorf("%s #%d: NewAddress unexpected error: %v", prefix, i, err)
```
jrick: Should be fatal, or put a continue here so it doesn't keep trying to use the invalid address.

waddrmgr/manager_test.go
tuxcanfly: I'm using this field now, to switch between `managedAddress` and `scriptAddress` so that both paths are covered in the test cases.

waddrmgr/manager_test.go
tuxcanfly: OK, added `continue`.

waddrmgr/common_test.go
```diff
@@ -53,6 +53,149 @@ var (
 
 	// waddrmgrNamespaceKey is the namespace key for the waddrmgr package.
 	waddrmgrNamespaceKey = []byte("waddrmgrNamespace")
+
+	// 	expectedExternalAddrs is the list of expected external addresses
```
jrick: Please remove tab characters between the // and the comment text.

waddrmgr/common_test.go
```diff
@@ -53,6 +53,149 @@ var (
...
+	// 	generated from the seed
+	expectedExternalAddrs = []expectedAddr{
```
jrick: As a suggestion on how to organize this data, I would combine all of the relevant tables into a single slice, and make subslices for groups of data.  So something like:

```
var expectedAddrs []expectedAddr{
    // ...
}

var expectedExternalAddrs = expectedAddrs[0:5]

var expectedInternalAddrs = expectedAddrs[5:10]
```
And then also remove the string slices of expected addrs, since those are not necessary (the address strings are already in the `expectedAddr` structs.

waddrmgr/manager_test.go
jrick: The type itself is used, but not the data.  I would consider using constants here instead:

func testMarkUsed(tc *testContext) bool {
    type addrType byte
    const (
        addrPubKeyHash addrType = iota
        addrScriptHash
    )
    tests := []struct {
        // ...
        typ addrType
        // ...
    }

    // ...

        switch test.typ {
        case addrPubKeyHash:
            // ...
        case addrScriptHash:
            // ...
        default:
            panic("unreachable")
        }

    // ...
}
```

Comment (tuxcanfly) [inline waddrmgr/manager_test.go]: Thanks, updated to use `addrType` const

File: waddrmgr/manager_test.go (hunk 5 above)
Comment (tuxcanfly) [inline waddrmgr/manager_test.go]: I think we're already doing this in `testWatchingOnly` by calling `testManagerAPI` after converting and opening the watch-only manager.

File: waddrmgr/common_test.go (hunk 9 above)
Comment (tuxcanfly) [inline waddrmgr/common_test.go]: Sorry, slipped past my editor, removed.

File: waddrmgr/common_test.go (hunk 10 above)
Comment (tuxcanfly) [inline waddrmgr/common_test.go]: Thanks, I've organized them as suggested.

However, I had to remove a test case because addresses in the db are keyed by hash so they are not sort-able. I think the address key should include a prefix of (account + index) to make them  sortable. We already have a ticket for re-doing address keys in #242 so I'll work on it there.

Comment (jrick) [inline waddrmgr/common_test.go]: Make a map of expected values, remove them as they're found, error if any are left.

Comment (tuxcanfly) [inline waddrmgr/common_test.go]: Thanks, restored the test case with the suggested approach.

File: waddrmgr/manager_test.go [hunk 11]
```diff
@@ -60,6 +60,15 @@ type testContext struct {
 	watchingOnly bool
 }
 
+// addrType is the type of address being tested
+type addrType byte
+
+const (
+	addrPubKeyHash addrType = iota
+	addrPubKey     addrType = iota
```
jrick: Unneeded type and iota here

waddrmgr/manager_test.go
```diff
@@ -1019,47 +920,77 @@ func testImportScript(tc *testContext) bool {
...
 // testMarkUsed ensures used addresses are flagged as such.
 func testMarkUsed(tc *testContext) bool {
-	expectedAddr1 := expectedAddr{
-		addressHash: hexToBytes("2ef94abb9ee8f785d087c3ec8d6ee467e92d0d0a"),
-		used:        true,
+	tests := []struct {
+		name       string
+		typ        addrType
+		in         []byte
+		blockstamp waddrmgr.BlockStamp
```
jrick: unused field

waddrmgr/manager_test.go
```diff
@@ -1019,47 +920,77 @@ func testImportScript(tc *testContext) bool {
 
 // testMarkUsed ensures used addresses are flagged as such.
 func testMarkUsed(tc *testContext) bool {
-	expectedAddr1 := expectedAddr{
-		addressHash: hexToBytes("2ef94abb9ee8f785d087c3ec8d6ee467e92d0d0a"),
-		used:        true,
+	tests := []struct {
+		name       string
+		typ        addrType
+		in         []byte
```
jrick: Not a huge deal but we could make this a [20]byte since everything should be a RIPEMD160 hash.

waddrmgr/manager_test.go
tuxcanfly: Removed.

waddrmgr/manager_test.go
tuxcanfly: Thanks, removed it.

jrick:
This looks ok now.  Can you squash and rebase?

tuxcanfly:
@jrick Thanks, rebased now.

davecgh:
OK

davecgh:
Need rebase to latest master.

tuxcanfly:
OK, squashed and rebased to latest master.

PR #246, opened 2015-04-16
Only use the empty string as the default account name.

### LINKED ISSUE #245: Default account is renamable
This behavior strikes me as wrong:

```
$ btcctl --simnet --wallet renameaccount default default2    
$ btcctl --simnet --wallet listaccounts
{  
  "default2": 9.99997,  
  "imported": 0  
}  
$ btcctl --simnet --wallet renameaccount default2 ''
-4: account with the same name already exists  
$ btcctl --simnet --wallet renameaccount default2 default
-4: reserved account name  
```

Three problems:

1. The default account is accessible by two names, the empty string and "default".
2. The "default" pseudonym can be renamed to something else, without actually changing the "real" account name, the empty string.
3. Once the default pseudonym has been modified, it cannot be changed back to either "default" (it's a reserved name) or "" (the account still exists).

All three of these problems can be removed by simply always using "" as the identifier of the default account.  Sure, there may be users who don't understand what "" is and why it can't be renamed, but that is a problem that should be solved by a wallet client (cli, gui, etc.)

jrick:
Disallow renaming accounts with reserved names.

Fixes #245.

### DIFF
--- a/waddrmgr/db.go
+++ b/waddrmgr/db.go
@@ -30,7 +30,7 @@ import (
 
 const (
 	// LatestMgrVersion is the most recent manager version.
-	LatestMgrVersion = 3
+	LatestMgrVersion = 4
 )
 
 var (
@@ -1661,7 +1661,7 @@ func upgradeToVersion2(namespace walletdb.Namespace) error {
 
 // upgradeManager upgrades the data in the provided manager namespace to newer
 // versions as neeeded.
-func upgradeManager(namespace walletdb.Namespace, pubPassPhrase []byte, config *Options) error {
+func upgradeManager(namespace walletdb.Namespace, pubPassPhrase []byte, chainParams *chaincfg.Params, config *Options) error {
 	var version uint32
 	err := namespace.View(func(tx walletdb.Tx) error {
 		var err error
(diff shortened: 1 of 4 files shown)


tuxcanfly:
Just a minor issue - if the user had renamed `default -> default2` the entry mapping `default2` is still present in the name index. I've pushed a fix for this in my branch:

https://github.com/tuxcanfly/btcwallet/commit/89f2b70149319e58a67dd57fd561fd41276b7e06

Thanks.

jrick:
Looks good, I'll have to test it out but I like the idea.  However I think the explicit "default" alias removal above can be removed as well (it is a non-zero length alias to the default account).

tuxcanfly:
Cool, yeah that can be removed.

jrick:
ForEach is implemented as a cursor, and items from the bucket *cannot* be removed during cursor iteration, except by Cursor.Delete.  I'll make it a cursor and push to my branch.

jrick:
Needs my commit that adds cursors to walletdb from wtxmgr work, so I'll cherrypick that as well.

tuxcanfly:
Thanks, doesn't look like we're doing that atm, will keep it in mind.

davecgh:
Looks like the cursor bits somehow made it into this PR.

davecgh:
This needs more discussion in my opinion.  I don't want to be stuck with "" as the name for the account that is generated by default and I will most likely be using.

jrick:
Sure I agree but that is a client problem.  The user only sees this when using btcctl because btcctl exposes a very raw interface.  But a nicer client can show "" as "Default Account" or something nice.

Besides, this keeps compatibility just a little bit closer to bitcoin core wallet (as frivolous as that may be at this point, but I don't see the point in deviating needlessly).

davecgh:
As far as the review goes, the code looks correct and I've tested that it does what it intends.

jrick:
Closing this; see discussion in #245 for how we plan on solving this.  I'll work up another patch later.

PR #248, opened 2015-04-17
Rollback txstore on disconnected blocks.

jrick:
Seems that this code got lost over time, and the only place where
reorgs were handled were those that happened while wallet was
disconnected.


davecgh:
OK

Issue #227, opened 2015-04-03
[rpc] This needs to be converted to use v2 of the btcjson package.


davecgh:
As the title indicates, the main daemon, and any sub-packages which make use of btcjson, need to be converted to use v2 of the package.  Version one of the package is deprecated and will be going away soon.

See btcsuite/btcd#314 for the pull request which converted btcd to use the package.  The changes here will be substantially similar.

PR #252, opened 2015-04-20
Convert RPC server to btcjson v2

jrick:
~~Note: This PR is rebased over #234.~~ merged

This PR converts the RPC server from the old btcjson package to the new v2 API.

The client connection to btcd is already using btcjson v2, since it is a dependency for btcrpcclient.  After this change, wallet will no longer depend on two versions of the btcjson package.

- [x] ~~Help registrations still need to be added~~ done


jrick:
This now includes help texts and usage, and is ready for review.

davecgh:
Mind adding closes #227 to the commit message, so it'll auto close when this gets merged?

jrick:
will do

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"infowalletresult-connections":     "The number of connected peers",
+	"infowalletresult-proxy":           "The proxy used by the server",
+	"infowalletresult-difficulty":      "The current target difficulty",
+	"infowalletresult-testnet":         "Whether or not server is using testnet",
+	"infowalletresult-relayfee":        "The minimum relay fee for non-free transactions in BTC/KB",
+	"infowalletresult-errors":          "Any current errors",
+	"infowalletresult-paytxfee":        "The increment used each time more fee is required for an authored transaction",
+	"infowalletresult-balance":         "The balance of all accounts calculated with one block confirmation",
+	"infowalletresult-walletversion":   "The version of the address manager database",
+	"infowalletresult-unlocked_until":  "Unset",
```
davecgh: Unset?

jrick: There are several fields that we don't set (and leave zeroed) because they either don't make sense for our architecture, or we don't track it.

davecgh: Right.  How about something like `Reserved` or `Not applicable`?   Unset didn't make sense to me.

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"getreceivedbyaddress-minconf":   "Minimum number of block confirmations required before an output's value is included in the total",
+	"getreceivedbyaddress--result0":  "The total received amount valued in bitcoin",
+
+	// GetTransactionCmd help.
+	"gettransaction--synopsis":        "Returns a JSON object with details regarding a transaction relevant to this wallet.",
+	"gettransaction-txid":             "Hash of the transaction to query",
+	"gettransaction-includewatchonly": "Also consider transactions involving watched addresses",
+
+	// HelpCmd help.
+	"help--synopsis":   "Returns one line summaries for every server method or detailed usage of a single request.",
```
davecgh: The help command can do individual commands too.  This is what btcd says:

`Returns a list of all commands or help for a specified command`

Probably should make then consistent.

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"help-command":     "Method to query detailed usage for",
+	"help--condition0": "command=''",
```
davecgh: For comparison to btcd:

```
	"help--synopsis":   "Returns a list of all commands or help for a specified command.",
	"help-command":     "The command to retrieve help for",
	"help--condition0": "no command provided",
	"help--condition1": "command specified",
	"help--result0":    "List of commands",
	"help--result1":    "Help for specified command",
```

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"help--condition1": "command!=''",
+	"help--result0":    "One line usage for every server request",
+	"help--result1":    "Detailed usage of a single request",
+
+	// GetTransactionResult help.
+	"gettransactionresult-amount":          "The total amount this transaction credits to the wallet, valued in bitcoin",
+	"gettransactionresult-fee":             "The total input value minus the total output value, or 0 if 'txid' is not a sent transaction",
+	"gettransactionresult-confirmations":   "The number of block confirmations of the transaction",
+	"gettransactionresult-blockhash":       "The hash of the block this transaction is mined in, or the empty string if unmined",
+	"gettransactionresult-blockindex":      "Unset",
```
davecgh: Another instance of Unset.  I'll leave the rest up to find/replace.

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+
+	// ImportPrivKeyCmd help.
+	"importprivkey--synopsis": "Imports a WIF-encoded private key to the 'imported' account.",
+	"importprivkey-privkey":   "The WIF-encoded private key",
+	"importprivkey-label":     "Unused (all imported addresses belong to the imported account)",
+	"importprivkey-rescan":    "Rescan the blockchain (since the genesis block) for outputs controlled by the imported key",
+
+	// KeypoolRefillCmd help.
+	"keypoolrefill--synopsis": "DEPRECATED -- This request does nothing since no keypool is maintained.",
+	"keypoolrefill-newsize":   "Unused",
```
davecgh: Unused vs. Unset, etc

jrick: Parameters vs results.

internal/rpchelp/helpdescs_en_US.go
jrick: aight I'll make them match.

internal/rpchelp/helpdescs_en_US.go
jrick: the problem is that the field is still in the json object, not missing, so the help language should indicate that the value is bogus.

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"listreceivedbyaccountresult-confirmations": "Number of block confirmations of the most recent transaction relevant to the account",
+
+	// ListReceivedByAddressCmd help.
+	"listreceivedbyaddress--synopsis":        "Returns a JSON array of objects listing wallet payment addresses and their total received amounts.",
+	"listreceivedbyaddress-minconf":          "Minimum number of block confirmations required before a transaction is considered",
+	"listreceivedbyaddress-includeempty":     "Unused",
+	"listreceivedbyaddress-includewatchonly": "Unused",
+
+	// ListReceivedByAddressResult help.
+	"listreceivedbyaddressresult-account":           "Unset",
```
davecgh: Is this true?  I'm pretty sure the result has the appropriate account.

jrick: I've intentionally avoided setting that field to anything meaningful since Bitcoin Core has deprecated it.

davecgh:
Before this PR:

> $ btcctl listreceivedbyaccount
> [
>   {
>     "account": "default",
>     "address": "S....",

After:

> $ btcctl listreceivedbyaccount
> [
>   {
>     "account": "",
>     "address": "S....",

internal/rpchelp/helpdescs_en_US.go
davecgh: It makes sense for our accounts though.

jrick:
That whole request is actually deprecated, not just the account field in the response.

davecgh:
> $ btcctl getblockhash 1000
> IjUyMDkwZmE5M2UzODMwODNkN2MwZmIyNzQ4MTZkNzc2ZGJiNTM3MmUyNWZiYjc2Njk2OTYyMWY0NGYzZDg5YjIi

davecgh:
> $ btcctl debuglevel show
> IlN1cHBvcnRlZCBzdWJzeXN0ZW1zIFtBRFhSIEFNR1IgQkNEQiBCTUdSIEJUQ0QgQ0hBTiBESVNDIE1JTlIgUEVFUiBSUENTIFNDUlAgU1JWUiBUWE1QXSI=

jrick:
Hmm those are wrong (encoding probably).

davecgh:
Looks like all of the passthough commands are not properly passing through the raw messages.

> $ btcctl decoderawtransaction "010....."
> eyJ.....fQ==

davecgh:
The `validateaddress` RPC is not detectnig public keys as being owned by the wallet.

> $ btcctl validateaddress Scd3...
> {
>  "isvalid": true,
>  "address": "Scd3...",
>  "ismine": true,
>  "pubkey": "02aadc...",
>  "iscompressed": true,
>  "account": "default"
> }

> $ btcctl validateaddress 02aadc...
{
  "isvalid": true,
  "address": "Scd3..."
}

jrick:
Seems to be the fault of the address manager (if that is a fault, I'll check what core does here).

jrick:
Correct (matching) behavior here is to return with `"isvalid": true` if it's a P2PKH or P2SH address.  So I'll change that.

jrick:
On second though, it wouldn't hurt to return isvalid, itmine, etc. for pubkeys as well.  But that can be fixed outside of this pr.

davecgh:
Confirmed it was the previous behavior too, so not related directly to this PR.  I'd like to see it changed separately as noted however.  Also of note is that, with a watching-only address, it is possible the wallet doesn't actually know the pubkey even if it knows the address, so the logic there has to be careful to not assume that existence of the address implies "ismine" for pubkeys.

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"dumpprivkey-address":   "The address to return a private key for",
+	"dumpprivkey--result0":  "The WIF-encoded private key",
+
+	// GetAccountCmd help.
+	"getaccount--synopsis": "Lookup the account name that some wallet address belongs to.",
+	"getaccount-address":   "The address to query the account for",
+	"getaccount--result0":  "The name of the account that 'address' belongs to",
+
+	// GetAccountAddressCmd help.
+	"getaccountaddress--synopsis": "Returns the most recent unused external payment address for an account.\n" +
```
davecgh: The address isn't technically unused.  It might have received funds to it already.

jrick: That's not correct either.  It should say "seen publically" instead of used.

davecgh:
Doesn't appear like settxfee is working correctly.

> $ btcctl getinfo
> {
>  ...
>   "paytxfee": 1e-05,
>  ...
> }

> $ btcctl settxfee 0.001
> true

> $ btcctl getinfo
> {
>  ...
>   "paytxfee": 0,
>  ...
> }

davecgh:
Help for getinfo claims:
```
 "keypoololdest": n,    (numeric) Unset
```

> $ btcctl getinfo
> {
>  ...
>   "keypoololdest": 1430430480,
>  ...
> }

I'm going to guess it's a time.Time zero val which isn't 0.  Set it explicitly to zero?

davecgh:
Help for getinfo claims:
```
 "keypoololdest": n,    (numeric) Unset
```

> $ btcctl getinfo
> {
>  ...
>   "keypoololdest": 1430430480,
>  ...
> }

I'm going to guess it's a time.Time zero val which isn't 0.  Set it explicitly to zero?

davecgh:
> $ btcctl getreceivedbyaddress Sbogus
> -5: checksum mismatch

> $ btcctl getreceivedbyaddress Sbogs
> -5: decoded address is of unknown format

Those should probably match or have the same prefix at least with the specific error after.

davecgh:
> $ btcctl getbestblockhash | btcctl listsinceblock -
>
> {
>  "transactions": [
>  ...
>      "confirmations": 5889,
>  ...
      "confirmations": 5888,
>  ...
      "confirmations": 5887,
>  ...
>  all the way down to 1

Then with the block hash for block 0:
> $ btcctl getblockhash 0 | btcctl listsinceblock -
> Same results...

jrick:
Haven't looked to see exactly what the bug is for that one but after giving it a quick glance I'm not happy with the implementation.  With the newer wtxmgr in, it can be done smarter, so I'll give those parts a rewrite later.

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"listunspent-maxconf":   "Maximum number of block confirmations required before a transaction output is excluded",
+	"listunspent-addresses": "If set, limits the returned details to unspent outputs received by any of these payment addresses",
+
+	// ListUnspentResult help.
+	"listunspentresult-txid":          "The transaction hash of the referenced output",
+	"listunspentresult-vout":          "The output index of the referenced output",
+	"listunspentresult-address":       "The payment address that received the output",
+	"listunspentresult-account":       "The account associated with the receiving payment address",
+	"listunspentresult-scriptPubKey":  "The output script encoded as a hexadecimal string",
+	"listunspentresult-redeemScript":  "Unset",
```
davecgh: Shouldn't this be set for p2sh addresses?

jrick: For compatibility, yes.  For just the btcjson conversion, no.

It's also saved in the db with the script crypto key and would require the entire wallet to be unlocked to access it.

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"dumpprivkey-address":   "The address to return a private key for",
+	"dumpprivkey--result0":  "The WIF-encoded private key",
+
+	// GetAccountCmd help.
+	"getaccount--synopsis": "DEPRECATED -- Lookup the account name that some wallet address belongs to.",
+	"getaccount-address":   "The address to query the account for",
+	"getaccount--result0":  "The name of the account that 'address' belongs to",
+
+	// GetAccountAddressCmd help.
+	"getaccountaddress--synopsis": "DEPRECATED -- Returns the most recent external payment address for an account that has not been seen publically.\n" +
```
davecgh: publically -> publicly

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	"signrawtransaction-flags":    "Sighash flags",
+
+	// SignRawTransactionResult help.
+	"signrawtransactionresult-hex":      "The resulting transaction encoded as a hexadecimal string",
+	"signrawtransactionresult-complete": "Whether all input signatures have been created",
+
+	// ValidateAddressCmd help.
+	"validateaddress--synopsis": "Verify that an address is valid.\n" +
+		"Extra details are returned if the address is controlled by this wallet.\n" +
+		"The following fields are valid only when the address is contrilled by this wallet (ismine=true): isscript, pubkey, iscompressed, account, addresses, hex, script, and sigsrequired.\n" +
```
davecgh: contrilled -> controlled

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -0,0 +1,399 @@
...
+	// ListReceivedByAddressResult help.
+	"listreceivedbyaddressresult-account":           "DEPRECATED -- Unset",
+	"listreceivedbyaddressresult-address":           "The payment address",
+	"listreceivedbyaddressresult-amount":            "Total amount received by the payment address valued in bitcoin",
+	"listreceivedbyaddressresult-confirmations":     "Number of block confirmations of the most recent transaction relevant to the address",
+	"listreceivedbyaddressresult-txids":             "Transaction hashes of all transactions involving this address",
+	"listreceivedbyaddressresult-involvesWatchonly": "Unset",
+
+	// ListSinceBlockCmd help.
+	"listsinceblock--synopsis":           "Returns a JSON array of objects listing details of wallet trasactions after some block.",
```
davecgh: trasactions -> transactions

davecgh:
Alright, so I've finished testing all of the RPCs.  I believe the only thing outstanding related to this PR is the  `listsinceblock` issue.  I'll make a separate issue for the other things I found which are not the result of this PR..

davecgh:
Everything looks good now.  Time for rebase/squash.  Also, please don't forget to add closes #227 to the commit message, so it'll auto close that issue when this gets merged.

OK

PR #253, opened 2015-04-23
Modify default account naming policy.

jrick:
Must be merged together with btcsuite/btcd#399.


jrick:
rebased

waddrmgr/manager.go
```diff
@@ -88,15 +94,18 @@ const (
...
-	// reservedAccountNames is a set of account names reserved for internal
-	// purposes
-	reservedAccountNames = map[string]struct{}{
-		"*":                     struct{}{},
-		DefaultAccountName:      struct{}{},
-		ImportedAddrAccountName: struct{}{},
-	}
-)
+// isReservedAccountName returns true if the account name is reserved.  Reserved
+// accounts may never be renamed, and other account may not be renamed to a
```
jrick: accounts

waddrmgr/manager_test.go
```diff
@@ -1264,13 +1264,6 @@ func testRenameAccount(tc *testContext) bool {
 	if !checkManagerError(tc.t, testName, err, wantErrCode) {
 		return false
 	}
-	// Test account name validation
-	testName = "*"
```
jrick: I'll add this test back using "imported" so it's reserved like before.

davecgh:
Other than the accounts typo and conversion to *btjcjson.RPCError which you did while I was reviewing, looks good.

OK after that typo fix and squash.  I'll merge btcsuite/btcd#399 as well.  It's

waddrmgr/manager_test.go
jrick: or not.  using "imported" always will fail with ErrDupliateAccount so it will never check for a reserved name.  Seems backwards but oh well.

jrick:
Travis will fail this one since btcsuite/btcd#399 isn't in yet.

PR #255, opened 2015-04-29
Sendpairs refactor

manan19:
~~This PR based on top of https://github.com/btcsuite/btcwallet/pull/252~~ Merged

The intent is to provide SendPairs as a helper routine to developers to be able to easily send funds from a specified account


manan19:
Is this still something that is going to be accepted?

jrick:
actually just finished reviewing it now :)

it's ok

PR #257, opened 2015-05-01
Fix waddrmgr version 4 upgrade path.

jrick:
If the account number to name index mapped the default account name to
an alias, the upgrade would not succeed and the upgrade would be
aborted (and rolled back).

This became a problem for upgrading old (pre-v3) wallets since the v3
upgrade did not rename the previous "" account to "default", but
instead just created an alias.

Fix tested by @dajohi, who ran into this issue with a wallet upgrade
from an older keystore version.


davecgh:
OK

PR #261, opened 2015-05-04
Check consistency of account id, name index on open

tuxcanfly:
This PR addresses an issue reported by @dajohi regarding missing imported addresses.


davecgh:
Shouldn't this be accompanied by a version bump?  It doesn't need to happen on every single open.

tuxcanfly:
There was a suggestion on IRC that we always have a consistency check on account indexes on startup. Right now this involves deserializing each account row so it might be inefficient and overkill. I'll check on IRC and move this to an version bump and upgrade in case that seems more favorable.

dajohi:
and maybe a flag..  ./btcwallet --check  to check consistency.

tuxcanfly:
Added a upgrade as well as flag `--check` to check the database and update the indexes.

PR #265, opened 2015-05-06
Updates for btcjson type changes.

jrick:
To increase compatibility with Bitcoin Core Wallet, additional fields
were added to and other fields made optional for the listtransactions
and gettransaction results structs.  For both, fee was changed to be
optional (including the zero value is allowed).

Requires btcsuite/btcd#427


davecgh:
[btcrpcclient](https://github.com/btcsuite/btcrpcclient) will need to be udpated as well.

jrick:
You sure? It builds and works fine

davecgh:
You're right.  I was thinking it was command updates for some reason.  Since it's only results, and there is no conversion happening for those results, it doesn't need an update.

jrick:
needs review

davecgh:
OK

PR #266, opened 2015-05-06
Remove target conf limit on listsinceblock txs.

### LINKED ISSUE #263: `listsinceblock` seems to miss unconfirmed transactions
It appears that btcwallet's `listsinceblock` call doesn't contain unconfirmed transactions known to btcd and btcwallet (checked via `gettransaction`).

jrick:
All transactions since the specified block (or the genesis block if
left unspecified) should be included in the result array

Along with this fix, update the help descriptions to mention that the
target confirmations parameter is not considered when including
transactions in the result object.  That is, transactions with a
height greater than the height of the lastblock in the result object
are still included.

Fixes #263.


davecgh:
OK

PR #267, opened 2015-05-07
Add script verification errors to signrawtransaction result.

dajohi:
This mimics Bitcoin Core commit 8ac2a4e1788426329b842eea7121b8eac7875c76


rpcserver.go
```diff
@@ -3060,18 +3060,39 @@ func SignRawTransaction(w *wallet.Wallet, chainSvr *chain.Client, icmd interface
...
+			tmpErr = err
+		}
+		if tmpErr != nil {
+			signErrors = append(signErrors,
+				btcjson.SignRawTransactionError{
+					TxID:      txIn.PreviousOutPoint.Hash.String(),
+					Vout:      txIn.PreviousOutPoint.Index,
+					ScriptSig: hex.EncodeToString(txIn.SignatureScript),
+					Sequence:  txIn.Sequence,
+					Error:     fmt.Sprintf("%v", tmpErr),
```
jrick: Don't use fmt:
```
Error: err.Error(),
```

rpcserver.go
```diff
@@ -2981,7 +2981,7 @@ func SignRawTransaction(w *wallet.Wallet, chainSvr *chain.Client, icmd interface
 	// `complete' denotes that we successfully signed all outputs and that
 	// all scripts will run to completion. This is returned as part of the
 	// reply.
-	complete := true
+	var signErrors []btcjson.SignRawTransactionError
```
jrick: Does core return [] or null when there are no errors?

rpcserver.go
```diff
@@ -3060,18 +3060,39 @@ func SignRawTransaction(w *wallet.Wallet, chainSvr *chain.Client, icmd interface
...
 		// Find out if it is completely satisfied or still needs more.
 		vm, err := txscript.NewEngine(input, msgTx, i,
 			txscript.StandardVerifyFlags)
-		if err != nil || vm.Execute() != nil {
-			complete = false
+
+		if err != nil {
+			tmpErr = err
+		} else if vm.Execute() != nil {
+			tmpErr = err
```
jrick: This doesn't look right.  Did you mean to set tmpErr to the result of vm.Execute?

rpcserver.go
jrick: For reference, core does not set this field in the json object.  Leaving it nil will have the same effect since the omitempty struct tag is specified.

internal/rpchelp/helpdescs_en_US.go
```diff
@@ -316,6 +316,14 @@ var helpDescsEnUS = map[string]string{
...
 	"signrawtransactionresult-hex":      "The resulting transaction encoded as a hexadecimal string",
 	"signrawtransactionresult-complete": "Whether all input signatures have been created",
+	"signrawtransactionresult-errors":   "Script verification errors (if exists)",
+
+	// SignRawTransactionError help.
+	"signrawtransactionerror-error":     "Verification or signing error related to the input",
+	"signrawtransactionerror-sequence":  "Script sequence number",
+	"signrawtransactionerror-scriptSig": "The hex-encoded signature script",
+	"signrawtransactionerror-txid":      "The hash of the referenced, previous transaction",
+	"signrawtransactionerror-vout":      "The index of the output to spent and used as input",
```
jrick: I'm not sure what this is trying to say.  Maybe:

"The output index of the referenced previous output"

?

And then we should change the one above to match:

"The transaction hash of the referenced previous output"

jrick:
ok

Issue #242, opened 2015-04-14
Bucket keys may not be derivable from solely public data


jrick:
Doing so subverts some of the protection intended by public data encryption.  For example, currently address hash160s are hashed again using sha256 and used as bucket keys.  While this prevents being able to go from bucket key (sha) -> address, if an attacker knows an interesting address and wanted to determine whether this wallet controlled it or not, this is as simple as looking up sha256(hash160) in the db.

Instead, keys should be crypted with the public crypto key, so if public encryption is used, this attack/privacy leak is not possible.

PR #268, opened 2015-05-12
waddrmgr: encrypt address keys

tuxcanfly:
Refs #242 

Still testing out the actual encryption, looks like the encrypted key is different each time.


tuxcanfly:
@jrick I think the encryption is different because of a random nonce. Instead I'm prepending the cryptokey to the address script and hashing the result, thus making sure the key cannot be derived from address alone.

tuxcanfly:
Working on the upgrade path to convert existing keys to the new format.

Issue #254, opened 2015-04-28
waddrmgr: implement active/inactive addresses


tuxcanfly:
>As discussed on IRC, an address can expire after a duration and become inactive while still remaining unused. Therefore, instead of assuming that addr not present in active bucket is used, we should have an explicit db flag on address row using a flag field which determines whether the address is used or not. Similarly, the address should be marked expired after a certain configurable duration.

PR #271, opened 2015-05-13
waddrmgr: add flags field on address row

tuxcanfly:
Refs #254 
WIP: Working on fetching the field from db, adding tests


jrick:
This looks like a pretty small and reasonable change, and shouldn't be difficult to merge either.  I'll review this soon so we can land it.

tuxcanfly:
Thanks. Yeah, this partially addresses the comments in #254. Still need to add the functionality to mark addresses as expired, but I think that could be done in a different PR.

PR #275, opened 2015-05-13
Prevent duplicate waddrmgr lock error.

### LINKED ISSUE #270: Seeing Error "WLLT: Could not lock wallet: address manager is locked"
Log shows `[ERR] WLLT: Could not lock wallet: address manager is locked` from time to time.

No obvious log entries around occurrence; log lists `WLLT: Marked address ...` before and after but with several seconds to minutes distance. Running with loglevel debug.

jrick:
Fixes #270.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -611,12 +611,14 @@ out:
 
 		// Select statement fell through by an explicit lock or the
 		// timer expiring.  Lock the manager here.
-		timeout = nil
-		err := w.Manager.Lock()
-		if err != nil {
-			log.Errorf("Could not lock wallet: %v", err)
-		} else {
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -606,6 +606,7 @@ out:
 			break out
 
 		case <-w.lockRequests:
+			timeout = nil
```
davecgh: This is going to be set to nil anyways below, but as discussed, I believe the order could be the issue since channel selection is not deterministic.

davecgh:
OK.  Needs rebase.

jrick:
Rebased as 472d6b0c1e13c0f76bbf33574aca881e093d9340

PR #277, opened 2015-05-13
Add fix for reorganizations from ppcwallet

cjepson:
Wallet had an issue in which blocks failed to be removed from wtxmgr by
disconnectBlock for long (>20 block) reorganizations because Rollback
failed to be called. This fix was written by mably for ppcwallet,
commit ffb1eeea39138be8a500dab4778abae3ade004a3.


jrick:
This fix isn't quite right.  The rollback will occur starting at the last recent block saved by the address manager, but the address manager may then be marked totally unsynced.  Instead, another rollback should be done in the else case that rolls back block 1 onwards (everything but the genesis block).

jrick:
Closing in favor of #281, which uses the fix I described above.

PR #279, opened 2015-05-14
Use 0 instead of -1 for zero-conf listtransactions results.

### LINKED ISSUE #278: zero-conf listtransactions results have -1 confirmations
should be 0.

jrick:
Fixes #278.


davecgh:
OK

PR #281, opened 2015-05-14
Rollback transactions past the last saved recent block.

jrick:
If a long reorganize occurs farther back than the last saved recent
block hash (currently max 20 are saved) a full rescan is triggered
since there is no guarantee the previous blocks weren't also removed
in the reorg.  In this case, the address manager was set unsynced, but
transaction history was not rolled back as well.  This commit corrects
this by unconfirming all transactions but those in the genesis block.


davecgh:
OK

PR #285, opened 2015-05-27
Move legacy under internal directory.

### LINKED ISSUE #232: Move the legacy folder under the internal folder
This will make all of the legacy, and hence unmaintained, code unavailable to external packages.

This will need to wait until after #217 is merged to prevent a bunch of conflicts.

jrick:
After Go 1.5, this will prevent consumers from importing these
packages since they are currently unmaintained.

Closes #232.


davecgh:
OK

PR #287, opened 2015-05-27
Take advantage of Go 1.5 optimized zeroing.

### LINKED ISSUE #286: Use range based zeroing for >= Go 1.5
Go 1.5 optimizes code such as the following:

```Go
for i := range slice {
        slice[i] = 0
}
```

to a duff's device to zero memory in the slice.  With Go tip, it benchmarks faster than the `zero.Bytes` function.

```
BenchmarkRange32	300000000	         4.14 ns/op
BenchmarkRange64	300000000	         4.26 ns/op
BenchmarkBytes32	200000000	         7.91 ns/op
BenchmarkBytes64	100000000	        13.6 ns/op
```

The range-based zero should be enabled for all versions but Go 1.4 using build tags.  When Go 1.5 is released, we will stop supporting Go 1.3, so only 1.4 needs the custom implementation.  After 1.6 is out, the custom implementation can be removed entirely.

jrick:
Closes #286.

### DIFF
--- a/internal/zero/array.go
+++ b/internal/zero/array.go
@@ -12,28 +12,7 @@
 // ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
 // OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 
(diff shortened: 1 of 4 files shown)


davecgh:
OK

PR #288, opened 2015-05-27
Add spendable field to listunspent result.

### LINKED ISSUE #262: `listunspent` does not set `spendable` field
In results of `listunspent`, Bitcoin core sets the field `spendable` to true for normal addresses and to false for watch only addresses. (They've added this recently in 0.10.0.) Btcwallet omits this currently.

jrick:
Fixes #262.

Requires btcsuite/btcd#440.


jrick:
This isn't quite right yet.  Bitcoin Core wallet will only mark multisig outputs as spendable ("mine") if all involved keys belong to the wallet.

jrick:
I started to work on the watching only bits to check if the outputs are actually spendable, but without the `WatchingOnly() bool` method on the managed addresses it is incredibly painful (have to attempt to access the privkey/script just to compare the error to one particular error case, while working around several others).  Even though the entire manager can be exported watching only, I don't think it's worth the effort since all of this code would be thrown out later anyways.

After #204 this will need to be updated with the new method.

davecgh:
Needs a squash and rebase, but OK after that.

PR #292, opened 2015-06-12
Remove data races from switching lock impls.

jrick:
sync.Locker cannot be safely used to switch a sync.Mutex to a noop
locker since other goroutines that attempt to lock the mutex will race
on the changing interface.  Instead, just statically dispatch
sync.Mutex methods.


jcvernaleo:
@jrick Code looks good and I tested and it takes care of the race I was seeing.

jcvernaleo:
So I'm ok with it going in.

PR #295, opened 2015-06-19
Reorder wtxmgr.CreditRecord fields.

jrick:
This reverses the order of index and amount fields to match
DebitRecord and also saves 8 bytes per instance on amd64.


davecgh:
OK

PR #298, opened 2015-06-30
Remove unnecessary address manager locks.

jrick:
In particular, this allows the use of AccountName in the passed
function when iterating over all account numbers.


davecgh:
OK (after rebase).

PR #301, opened 2015-07-05
Restore terminal on signal.

### LINKED ISSUE #296: Pressing CTRL+C while entering password leaves terminal in no echo state
Steps to reproduce the issue:
- Type in btcwallet --create
- Enter a random password but don't press enter. Press CTRL+C.
- Try typing in anything.
- Tried switching to terminal instead of terminal emulator. No change.

Temporary Fix:
- Reopen terminal emulator.
- Or if used terminal, logout/login.

Permanent Fix:
- Save terminal state as soon as btcwallet starts and reset state in case of interrupt and then shutdown.
```go
// Store terminal state.
state, err := terminal.GetState(int(os.Stdin.Fd()))
if err != nil {
	return err
}

// Restore terminal state in case of interrupt.
addInterruptHandler(func() {
	terminal.Restore(int(os.Stdin.Fd()), state)
})
```
However, things are not as easy as this. Because the interrupt doesn't get propagated i.e. btcwallet doesn't shutdown.

dajohi:
Fixes #296

### DIFF
--- a/btcwallet.go
+++ b/btcwallet.go
@@ -25,10 +25,12 @@ import (
 	"runtime"
 
 	"github.com/btcsuite/btcwallet/chain"
+	"github.com/btcsuite/golangcrypto/ssh/terminal"
 )
 
 var (
-	cfg *config
+	cfg           *config
+	terminalState *terminal.State
 )
 
 func main() {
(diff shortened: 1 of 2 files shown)


jrick:
I believe the correct fix for this would be for calls to ReadPassword to be preceded by another interrupt handler that restores the current terminal state.  sync.Once should be useful here so the code isn't run a second time on ^C later when ReadPassword read input successfully.

jrick:
Closing since this doesn't actually fix the issue (confirmed by the issue reporter) and I've been unsuccessful in attempts to find a workaround that does work.

PR #302, opened 2015-07-06
Implement Pool.UpdateWithdrawal

gsalgado:
This is the API that will eventually be exposed over the websocket interface so that members of a votingpool can pass on to the wallet the new signature lists (for a given withdrawal round) they received from other members

This depends on #305

### DIFF
--- a/votingpool/common_test.go
+++ b/votingpool/common_test.go
@@ -63,38 +63,39 @@ func TstRunWithManagerUnlocked(t *testing.T, mgr *waddrmgr.Manager, callback fun
 	callback()
 }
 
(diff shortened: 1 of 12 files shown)


davecgh:
FYI.  I plan on reviewing this and the other voting pool related PRs over the next few days.  We haven't forgot them, just been rather busy with btcd infrastructure changes.

davecgh:
I'm going to merge all of these into a `votingpool_review` branch since it's getting a bit unwieldy and we won't be able to properly review them until the current major infrastructure work is done.

PR #303, opened 2015-07-06
New struct (TxInSigs) to store the raw sigs of a tx input

gsalgado:
This allows us to store the number of required sigs for a given input as well
as the raw signatures themselves.


davecgh:

PR #305, opened 2015-07-08
Make getRawSigs work on changeAwareTxs instead of withdrawalTxs

gsalgado:
This is necessary so that we can re-generate the raw signatures in UpdateWithdrawal.

Depends on #303


davecgh:

PR #307, opened 2015-07-10
StartWithdrawal now updates WithdrawalStatus.nextInputAddr

gsalgado:
Depends on #302

PR #308, opened 2015-07-13
votingpool: Enforce constraints on addr indexes

gsalgado:
* ChangeAddress(): error if index==0
* DepositScript(): error if index >= hdkeychain.HardenedKeyStart

Also get rid of nextChangeAddress(), which was broken as a given withdrawal should only use change addresses from the active series.

Depends on #307


davecgh:

PR #309, opened 2015-07-15
New API to calculate the balance of a votingpool series

gsalgado:
Depends on #308


davecgh:

PR #310, opened 2015-07-16
Debug log the reason of connection failure

bclermont:
To let user run with more detailed debug log and figure why it fail

### DIFF
--- a/btcwallet.go
+++ b/btcwallet.go
@@ -119,8 +119,8 @@ func walletMain() error {
 		}
 		err = rpcc.Start()
 		if err != nil {
-			log.Warnf("Connection to Bitcoin RPC chain server " +
-				"unsuccessful -- available RPC methods will be limited")
+			log.Warnf("Connection to Bitcoin RPC chain server unsuccessful " +
+				"-- available RPC methods will be limited -- error: %v", err)
 		}
 		// Even if Start errored, we still add the server disconnected.
 		// All client methods will then error, so it's obvious to a


bclermont:
I just spent a while trying to change user, pass, address and so on and I had to change the code to figure what I was doing wrong

davecgh:
I think it's probably fine to include the error in the message regardless of the debug level.

For example:
```Go
log.Warnf("Connection to Bitcoin RPC chain server " +
				"unsuccessful -- available RPC methods will be "+
				"limited -- error: %v", err)
```

@jick any objections to that?

Alternatively, if the message is to be modified based on debug level, it would result in a cleaner message, in my opinion, to do it like so:

```Go
			msg := "Connection to Bitcoin RPC chain server unsuccessful -- "+
				"available RPC methods will be limited"
			if log.Level() <= btclog.DebugLevel {
				msg += "-- error: " + err.Error()
			}
			log.Warn(msg)
```

jrick:
@dajohi those seem fine

jick:
Hey, I'm afraid you've got the wrong e-mail address for somebody here.

Zack

davecgh:
@jick:  Sorry was a typo.  I meant to put @jrick.

bclermont:
Updated based on @davecgh comments

davecgh:
OK

jrick:
I'm not sure why this never got merged (sorry!) but the latest code keeps looping forever trying the connection (instead of stopping after 5 attempts or so like before) and logs an error for each unsuccessful attempt.

jick:
Guys, I'm still not the right Zack Johnson.  Please remove my address from
this list.

PR #311, opened 2015-07-21
votingpool: Update the status of partially fulfilled withdrawal requests

gsalgado:
These now include the ID of the highest series that needs to be thawed
before the requests can be fulfilled.

Depends on #309


davecgh:

PR #314, opened 2015-07-22
Add a new error when signing raw transactions specific to multisig

cjepson:
For the signing of raw transactions that pay to a P2SH multisignature
output, a new error has been added to inform the user that not enough
signatures havebeen produced in place of the non-informative txscript
error 'stack underflow'. Note that both behaviours deviate from bitcore,
which simply results the partially signed transaction without any
errors.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -3013,13 +3013,30 @@ func SignRawTransaction(w *wallet.Wallet, chainSvr *chain.Client, icmd interface
 			err = vm.Execute()
 		}
 		if err != nil {
+			localErr := err
+			class, addr, _, _ := txscript.ExtractPkScriptAddrs(
+				inputs[txIn.PreviousOutPoint],
+				activeNet.Params)
+
+			if err == txscript.ErrStackUnderflow &&
+				class == txscript.ScriptHashTy {
+				redeemScript, _ := getScript(addr[0])
+				redeemClass, _, _, _ := txscript.ExtractPkScriptAddrs(
+					redeemScript,
+					activeNet.Params)
+				if redeemClass == txscript.MultiSigTy {
+					localErr = errors.New("failed to produce enough " +
+						"signatures to satisfy the conditions of the " +
+						"multisignature script")
+				}
+			}
 			signErrors = append(signErrors,
 				btcjson.SignRawTransactionError{
 					TxID:      txIn.PreviousOutPoint.Hash.String(),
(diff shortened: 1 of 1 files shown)


davecgh:
I think the correct behavior is no error.  There are several occasions where the signer will not have all of the keys and a partially signed transaction must be produced in order to give it to the other parties to add their signatures until a final tx with all signatures is produced.

rpcserver.go
```diff
@@ -3013,13 +3013,30 @@ func SignRawTransaction(w *wallet.Wallet, chainSvr *chain.Client, icmd interface
...
 		if err != nil {
+			localErr := err
+			class, addr, _, _ := txscript.ExtractPkScriptAddrs(
+				inputs[txIn.PreviousOutPoint],
+				activeNet.Params)
+
+			if err == txscript.ErrStackUnderflow &&
+				class == txscript.ScriptHashTy {
+				redeemScript, _ := getScript(addr[0])
+				redeemClass, _, _, _ := txscript.ExtractPkScriptAddrs(
```
davecgh: It's much cheaper to use `txscript.GetScriptClass` since all you're after is the class of the script.

cjepson: Whoops, didn't know it existed. Lemme fix that too.

cjepson:
That should address that. I think I accidentally added a binary too, I'll remove it when I squash

rpcserver.go
```diff
@@ -3013,14 +3013,33 @@ func SignRawTransaction(w *wallet.Wallet, chainSvr *chain.Client, icmd interface
...
-					Error:     err.Error(),
-				})
+			multisigNotEnoughSigs := false
+			class, addr, _, _ := txscript.ExtractPkScriptAddrs(
+				inputs[txIn.PreviousOutPoint],
+				activeNet.Params)
+
+			if err == txscript.ErrStackUnderflow &&
+				class == txscript.ScriptHashTy {
+				redeemScript, _ := getScript(addr[0])
```
davecgh: Probably need to modify the if to detect 0 addresses too.  This will panic if no addrs could be extracted.

davecgh: Oh well nevermind, you checked that it's a `ScriptHashTy` already, so the address will exist.

jrick:
Looks like this is still needed.  Can you update the PR? (this code now lives in the SignTransaction method in the wallet package.)

cjepson:
Hey, didn't see this. Yeah, I'll see what I can do.

PR #316, opened 2015-08-24
Get rid of a few unnecessary functions and unexpose others that are only used internally

gsalgado:


jrick:
This looks ok.  if you rebase there over master I'll ff merge it, otherwise I'll do that next week just to get it in.

PR #317, opened 2015-08-24
Use a mutex to prevent concurrent calls to StartWithdrawl/UpdateWithdrawal

gsalgado:
Depends on #316


jrick:
Could these mutexes be added to the Pool struct instead of package globals?

gsalgado:
Unfortunately no, because Pool is not a singleton so concurrent RPC calls with the same roundID would end up creating separate Pool instances and thus be able to execute concurrently.

PR #320, opened 2015-09-16
Detect silent network drops.

jrick:
This change introduces additional network activity with the btcd
process to ensure that the network connection is not silently dropped.
Previously, if the connection was lost (e.g. wallet runs on a laptop
and connects to remote btcd, and the laptop is suspended/resumed) the
lost connection would not be detectable since all normal RPC activity
(excluding requests from btcwallet to btcd made by the user) is in the
direction of btcd to wallet in the form of websocket notifications.

Rather than simply pinging the server and waiting for the response, a
unique session ID has been added for each websocket client connection.
These session IDs are randomly generated by the btcd RPC server when a
new websocket connection is established, and are queried using the
session RPC.  If the session ID changes, it is assumed that the
connection silently reconnected, and the client is shutdown for this
case as well.

Requires btcsuite/btcd#500 and btcsuite/btcrpcclient#62

### DIFF
--- a/btcwallet.go
+++ b/btcwallet.go
@@ -123,7 +123,7 @@ func walletMain() error {
 			server.SetChainServer(rpcc)
 
 			// Start wallet goroutines and handle RPC client notifications
-			// if the wallet is not shutting down.
+			// if the server is not shutting down.
 			select {
 			case <-server.quit:
 				return
@@ -134,6 +134,7 @@ func walletMain() error {
 			// Block goroutine until the client is finished.
 			rpcc.WaitForShutdown()
 
+			wallet.SetChainSynced(false)
 			wallet.Stop()
 
 			// Reconnect only if the server is not shutting down.
--- a/btcwallet.go
+++ b/btcwallet.go
@@ -91,52 +91,57 @@ func walletMain() error {
 	// Shutdown the server if an interrupt signal is received.
 	addInterruptHandler(server.Stop)
 
-	// Create channel so that the goroutine which opens the chain server
-	// connection can pass the conn to the goroutine which opens the wallet.
-	// Buffer the channel so sends are not blocked, since if the wallet is
-	// not yet created, the wallet open goroutine does not read this.
-	chainSvrChan := make(chan *chain.Client, 1)
-
 	go func() {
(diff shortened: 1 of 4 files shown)


davecgh:
This appears to be resetting back to height 0 after a disconnect:

> 12:14:44 2015-09-17 [INF] WLLT: Started rescan from block 000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943 (height 0) for 1 address
> ...
> 12:20:34 2015-09-17 [INF] WLLT: Finished rescan for 1 address (synced to block 0000000003c34aa56f526562497c63e9cedc59dbac76c385efdd0cdbee6e3419, height 556333)
> ...
> ***Killed remote connection here***
> 12:27:19 2015-09-17 [ERR] CHNS: Failed to receive session result: &lt;nil&gt;
> 12:27:19 2015-09-17 [INF] CHNS: Established websocket RPC connection to btcd
> 12:27:19 2015-09-17 [INF] WLLT: Started rescan from block 000000000933ea01ad0ee984209779baaec3ced90fa3f408719526f8d77f4943 (height 0) for 1 address

davecgh:
It appears working properly now with the latest commit:

> 14:13:17 2015-09-17 [INF] WLLT: Finished rescan for 1 address (synced to block 0000000000025723097205ec8ae22750fec0e6fabb1eee027a61a19d5b1c2f63, height 556361)
> 14:14:17 2015-09-17 [ERR] CHNS: Failed to receive session result: the client has been shutdown
> 14:14:17 2015-09-17 [INF] CHNS: Established websocket RPC connection to btcd
> 14:14:17 2015-09-17 [INF] WLLT: Started rescan from block 0000000000025723097205ec8ae22750fec0e6fabb1eee027a61a19d5b1c2f63 (height 556361) for 1 address
> 14:14:17 2015-09-17 [INF] WLLT: Finished rescan for 1 address (synced to block 0000000000025723097205ec8ae22750fec0e6fabb1eee027a61a19d5b1c2f63, height 556361)

davecgh:
OK

jrick:
squashed+merged as 33d053c6a71003c1a72481ed21a2de6f92fcf6c7

PR #324, opened 2015-10-16
Fix signmessage RPC to match Core.

jrick:
AFAICT this function has never worked correctly due to the hash being
signed not matching the hash created by Core.  Core wallet writes
serialized strings to a double-sha256 hashing stream, while we were
using string concatination.  This produced different messages since
the message before hashing did not include compact integers (called
varints in btcsuite code) preceding each string with the string
length.

Tested by creating signed messages from btcwallet and verifying them
with Bitcoin-Qt, as well as creating signatures from Bitcoin-Qt and
verifying them with btcwallet.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -2732,9 +2732,12 @@ func SignMessage(w *wallet.Wallet, chainSvr *chain.Client, icmd interface{}) (in
 		return nil, err
 	}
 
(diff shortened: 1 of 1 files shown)


davecgh:
Tested and confirmed as working properly in both directions (all combination of verifying and creating signatures between btcwallet and Bitcoin Core).

OK.

PR #327, opened 2015-11-19
Add sweepaccount tool.

jrick:
This tool creates on-chain transactions, one per used address in a
source account, to sweep all output value to new addresses in a
different destination account.

### DIFF
--- a/cmd/sweepaccount/main.go
+++ b/cmd/sweepaccount/main.go
@@ -1,0 +1,348 @@
+// Copyright (c) 2015-2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+
+	"github.com/btcsuite/btcd/btcjson"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcrpcclient"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcwallet/internal/cfgutil"
+	"github.com/btcsuite/btcwallet/netparams"
+	"github.com/btcsuite/btcwallet/wallet/txauthor"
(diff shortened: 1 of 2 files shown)


jrick:
Example usage on testnet: http://sprunge.us/bKjh

jrick:
Once #330 is in I'd like to use that to add testnet/simnet flags and normalize the rpc connection string (currently this requires the port, which could be made optional).

jrick:
Added the new testnet and simnet options, and made the port in the rpc network address optional (uses the default mainnet/testnet/simnet wallet rpc server port if unspecified).

jrick:
Also, AmountFlag looks useful for other tools and/or btcwallet itself.  Thoughts on moving that into cfgutil @davecgh?

davecgh:
Personally, I'd just do it like dhill did in btcd where the exported config struct entry a float64 and then the conversion and error bits are in the `loadConfig` function.  Making it satisfy the go-flags interface more tightly integrates it to go-flags which I'd personally prefer to avoid.

That said, if you want to do it this way and put it over in `cfgutil`, that's fine too.  I don't have a strong opinion on it.

jrick:
Actually I'd like to see the logic in UnmarshalFlag to be moved to btcutil.NewAmountFromString and check for other units besides BTC, and then the interface method just calls that.

Satisifying the go-flags interface makes sense at this point since we're committed to it and want to use it everywhere for consistency.  cfgutil is also an internal package so it only affects this repo.

davecgh:
Improving `btcutil.NewAmountFromString` sounds like a fine idea to me.

davecgh:
I like the idea of the utility, but I haven't reviewed the code yet.

dajohi:
This tool worked great on both mainnet and testnet.  I just used it to move my funds from imported accounts to the new addresses on the chain.  Back in business.

jrick:
Rebased over latest master and rewritten to use the new txauthor package to create the unsigned transaction.

davecgh:
OK

jrick:
Well turns out this wasn't rebased over master after all.  Merged as 4ee9ce59fba801ff5036de35afac67b61e6247cc.

PR #329, opened 2015-11-23
Release 0.7.0

jrick:


davecgh:
The `btcrpcclient` dep is too old.  It should be at least 975536f.

davecgh:
OK

PR #330, opened 2015-11-25
Separate out default ports and utility funcs.

jrick:
This change moves the chain and network parameter definitions, along
with the default client and server ports, to a package for reuse by
other utilities (most notably, tools in the cmd dir).  Along with it,
functions commonly used for config parsing and validation are moved to
an internal package since they will also be useful for distributed
tools.

### DIFF
--- a/rpcserver.go
+++ b/rpcserver.go
@@ -322,9 +322,11 @@ func newRPCServer(listenAddrs []string, maxPost, maxWebsockets int64) (*rpcServe
 	// Setup TLS if not disabled.
 	listenFunc := net.Listen
 	if !cfg.DisableServerTLS {
-		// Check for existence of cert file and key file
-		if !fileExists(cfg.RPCKey) && !fileExists(cfg.RPCCert) {
-			// if both files do not exist, we generate them.
+		// Check for existence of cert file and key file.  Generate a
+		// new keypair if both are missing.
+		_, e1 := os.Stat(cfg.RPCKey)
+		_, e2 := os.Stat(cfg.RPCCert)
+		if os.IsNotExist(e1) && os.IsNotExist(e2) {
 			err := genCertPair(cfg.RPCCert, cfg.RPCKey)
 			if err != nil {
 				return nil, err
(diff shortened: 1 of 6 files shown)


rpcserver.go
```diff
@@ -322,9 +322,11 @@ func newRPCServer(listenAddrs []string, maxPost, maxWebsockets int64) (*rpcServe
 	// Setup TLS if not disabled.
 	listenFunc := net.Listen
 	if !cfg.DisableServerTLS {
-		// Check for existence of cert file and key file
-		if !fileExists(cfg.RPCKey) && !fileExists(cfg.RPCCert) {
-			// if both files do not exist, we generate them.
+		// Check for existence of cert file and key file.  Generate a
+		// new keypair if both are missing.
+		_, e1 := os.Stat(cfg.RPCKey)
```
davecgh: Any reason not using the `cfgutil.FileExists` here?

jrick: It gets needlessly messy because of the extra error paths for io errors not due to a missing file.

```
keyExists, err1 := cfgutil.FileExists(cfg.RPCKey)
certExists, err2 := cfgutil.FileExists(cfg.RPCCert)
if err1 == nil && err2 == nil && !keyExists && !certExists {
    // ...
}
`

File: walletsetup.go [hunk 2]
```diff
@@ -432,8 +432,13 @@ func createWallet(cfg *config) error {
...
 	keystorePath := filepath.Join(netDir, keystore.Filename)
 	var legacyKeyStore *keystore.Store
-	if fileExists(keystorePath) {
-		var err error
+	_, err := os.Stat(keystorePath)
+	if err != nil && !os.IsNotExist(err) {
+		// A stat error not due to a non-existant file should be
+		// returned to the caller.
+		return err
+	} else if err == nil {
```
jrick: @davecgh this one could be easily converted to use cfgutil.FileExists, however, since that error case should be returned here.

davecgh:
I don't see much point to this to be honest, but no objections.  Out of the functions moved, the only thing being used is `cfg.FileExists` and even then it's not used consistently (per our previous comments).  This seem like an extra dependency for almost no gain.

That said, code wise it's OK.

jrick:
It's an extra "dependency" (not really, just shared code in a package) because we want other tools (sweepaccount, my as of now unreleased tool to prompt and unlock the wallet, etc.) to share the default ports and config logic.  And besides those tools, my upcoming grpc server patch will use the address normalization from this change for initializing the btcd RPC.

davecgh:
OK

PR #331, opened 2015-11-25
Update README.

jrick:
Removed links to outdated btcsuite MSIs, replacing these with links to
Github releases.

Combined installation and updating instructions since they are
identical.

Added Windows to the list of operating systems that the "Build from
source" instructions work with.

Added PowerShell examples for copying the sample btcd and btcwallet
configs for both MSI and source installs.


davecgh:
OK

Issue #156, opened 2014-12-19
Resending of unmined transactions must be done in dependency order.


jrick:
If not, it's possible that transactions will be refused from mempool for spending unknown inputs (sendrawtransaction does not allow orphans).

davecgh:
@tuxcanfly will be taking this after finish #216.

PR #332, opened 2015-11-25
Correct transaction dependency ordering on startup mempool resync

cjepson:
Submission of unconfirmed transactions to daemon memory pool upon connection
would sometimes fail because their dependency ordering was incorrect. This
sorts them as a directed acyclic graph so that their ordering doesn't cause
'unknown inputs' errors upon submission to the daemon memory pool.

Corrects issue #156

### DIFF
--- a/wtxmgr/unconfirmed.go
+++ b/wtxmgr/unconfirmed.go
@@ -179,8 +179,8 @@ func (s *Store) unminedTxs(ns walletdb.Bucket) ([]*wire.MsgTx, error) {
 		return nil, err
 	}
 
-	// Transactions with no local depencies are excluded from this list, so
-	// we need to add them back now. First, find transactions with local
+	// Transactions with no local dependencies are excluded from this list,
+	// so we need to add them back now. First, find transactions with local
 	// dependencies. Then, sort those as DAGs. Finally, append all the
 	// transactions with no local dependencies and ship them out to the
 	// caller.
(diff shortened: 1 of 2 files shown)


jrick:
wire.ShaHash has value equality using == so just use that instead of IsEqual (and this matches the rest of the codebase).

Slices can be ranged over without first checking for nil (a nil slice is treated the same as a non-nil slice with 0 length).

We should error for cycles since that should only happen if you pass in invalidly signed transactions.

wtxmgr/unconfirmed.go
```diff
@@ -164,9 +164,51 @@ func (s *Store) unminedTxs(ns walletdb.Bucket) ([]*wire.MsgTx, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	order, _, err := topSortKahn(g, i)
+	if err != nil {
+		return nil, err
+	}
+
+	// Transactions with no local depencies are excluded from this list, so
```
jrick: typo: depencies

cjepson:
Recommended changes have been made, with the exception of adding an error for cycle detection. Cycles are already detected and an error is returned if they are found.

jrick:
I'd like to get this in.  Unfortunately it looks like the webpage containing the original implementation is covered under the GNU Free Documentation License, and I'm unsure if we can use the code there without violating the GPL (or any GNU copyleft license).

jrick:
I've reimplemented this change without copying the code from another source due to the license concerns.  Closing this PR and will open a new one.

PR #333, opened 2015-11-26
Add resources package for app-specific names and paths.

jrick:
This separates out application names ("btcwallet" and "btcd") and file
paths (the application data directory, wallet database file, etc.)
into a non-internal package.

There are two benefits from splitting these resources this way:

1) Distributed tools in the cmd dir and 3rd party tools can reuse
   these paths for configuration defaults.

2) Altcoins, in theory, only have to change the application names once
   and the changes are made visible everywhere.

The resources package also defines the active network variable (moved
from the main package's params.go) and the application data directory
itself.  If the network is changed, or the active datadir is modified,
the functions of the resource package that return paths relative to
these directories will use the new values.


resources/resources.go
```diff
@@ -0,0 +1,166 @@
+package resources
```
jrick: note to self: add license header

davecgh:
Not sure the active network parameters should be in resources, but other than that and the missing license header, seems fine.

jrick:
@davecgh would you prefer all the resource paths that require the network name to pass in a *netparams.Params instead of using a package global?

jrick:
Anyways, after we settle on an API for this, I'd like to get all the functions here tested to ensure they actually do return the expected values for default and custom datadirs/networks/etc.

jrick:
I'm redoing this by defining a Resources struct in this package with methods to return all of the various resources.  This is more testable and allows multiple resources to be used simultaneously.

jrick:
This is, perhaps not so strangely, turning into another config struct without many of the other config options. hmmmmm...

jrick:
rebased over the grpc branch.

jrick:
Rebased over master.

jcvernaleo:
Looks good to me (and works in quick tests).
I think this kind of change will be good for dcrwallet as well so I definitely want to see it in (but I hope you have to do the merge to dcr not me).
OK

jrick:
--create panics after this change so i'll resolve that first.

PR #336, opened 2015-12-03
Make chain.Client an interface to allow for swapable back ends

d4l3k:
Hi, 
I'm trying to make a simple bitcoin handling library using btcwallet and [Electrum](https://electrum.org/).
https://github.com/d4l3k/go-electrum

These are some refactors to make chain.Client into an interface and the existing Client into BTCRPCClient. This allows for other back ends besides btcrpcclient such as Electrum which doesn't require the entire blockchain to be stored locally.

This is an initial refactor with the fewest changes possible since I'm not that familiar with the btcwallet code base. If you think this isn't the best way to do this, please let me know which direction to look.

Thanks!

### DIFF
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -17,6 +17,7 @@
 package chain
 
 import (
+	"encoding/json"
 	"errors"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


tuxcanfly:
Thanks, I need to take a closer look but I was going to work on this for SPV mode, specifically making `chain.Client` an interface that can be used to implement a P2P client.

jrick:
The point of chain.Client was not implement SPV (it embeds a btcrpcclient.Client, so the caller expects it to behave as one, and this is used for several RPC handlers and RPC passthrough) but to implement some additional functionality and behavior on top of what the btcrpcclient package provides (using channels instead of callbacks and serializing each notification, for example).  I'm not sure how an electrum server fits in here, but for SPV I believe making chain.Client an interface is the wrong path.

After developing a much better GUI for btcwallet (sorry, not released publically yet) I've also come to the conclusion that the way wallet updates itself with btcd (or electrum, etc.) notifications is a little backwards.  instead of the wallet depending on the rpc client, I'd prefer the client to depend on the wallet and make changes to it.  This would also make the wallet itself much more testable, since it would be its own component and wouldn't need a connection to btcd (or spv, electrum...) to test modifications.

davecgh:
FWIW, I have thought the same thing for a while myself in that wallet really should be a separate component where the client should simply use its functionality and be responsible for obtaining the necessary chain state and transaction data from whatever means it wants whether that is via RPC, P2P, ZMQ, etc.

jrick:
I'm going to close this then.  I'm not opposed to having an electrum server backend, but this just isn't the development direction we want wallet going at the moment.  Instead, we should be adding exported mutator functions to wallet.Wallet which other syncing code calls to keep the wallet up to date.

d4l3k:
That makes sense.

What are your thoughts on the timeline for the refactoring? I'd like to be able to use Electrum as a backend sooner rather than later. As it is, I'm thinking about making a mock RPC backend so I can use btcwallet without these changes.

Please let me know if there's anything I can do to help.

jrick:
The refactoring won't be done for while since I have other priorities for wallet that I am working on currently.  Making this an interface is the easiest and fastest way to add Electrum server support so feel free to continue doing so if you want to get that working quickly, but it will need to be on your own fork (sorry!).

If you'd like to help with the refactoring, I will be creating a new PR this week with massive changes to the RPC server side of wallet, including moving the old server to its own package and adding a new RPC server without all the legacy Core-compatible methods that everyone hates so much.  Because it's such a huge change I'd like for any other code reorganization to happen on top of that.  Any additional review on that change would be very welcome.

PR #344, opened 2016-01-15
Votingpool review

jrick:
Opening PR for review so this doesn't get lost again.


votingpool/withdrawal.go
```diff
@@ -124,16 +127,33 @@ type withdrawalInfo struct {
...
-type TxSigs [][]RawSig
+type TxInSigs struct {
+	Required uint32
+	Sigs     []RawSig
+}
+
+func newTxInSigs(reqSigs uint32, length int) TxInSigs {
+	sigs := make([]RawSig, length)
+	for i := range sigs {
+		sigs[i] = emptyRawSig
```
jrick: These assignments are unnecessary.

votingpool/factory_test.go
```diff
@@ -243,27 +301,29 @@ func TstCreateCreditsOnNewSeries(t *testing.T, pool *Pool, amounts []int64) (uin
...
-	for i := range msgTx.TxOut {
+	for i, amount := range amounts {
+		addr := TstNewWithdrawalAddress(t, pool, seriesID, Branch(1), Index(i))
+		pkScript, err := txscript.PayToAddrScript(addr.addr)
+		if err != nil {
+			t.Fatal(err)
+		}
+		msgTx := createMsgTx(pkScript, []int64{amount})
+		txSha := msgTx.TxSha()
+		idx := 0
```
jrick: Why is the index always 0 here? (why use idx and not i?)

votingpool/withdrawal.go
gsalgado: Indeed; removed

votingpool/factory_test.go
gsalgado: Because createMsgTx() will create a tx with a single TxOut, and the Credit's Outpoint (created a few lines below) will refer to that. Maybe I could rename it to txOutIdx? Or add a comment explaining why it's always 0?

gsalgado:
BTW, I'm doing those changes you requested on my own copy of the votingpool_review branch (and I'm amending the commits), so you won't see them here unless you force pull. Maybe we should create a PR from my own votingpool_review branch to master?

votingpool/factory_test.go
jrick: OK, I see.  I had to look at the body of createMsgTx to figure out what it's doing here.  I think renaming that function could give a little more clarity here.  Perhaps createCoinbaseTx? Coinbases can only have one input, and the previous output matches that of a coinbase.

jrick:
Closing, please open one for your branch.

votingpool/factory_test.go
gsalgado: But the fact that the prevout matches that of a coinbase is just an implementation detail of that func, and it's not relevant to its callsites. I think what the name should convey is that it creates a msgtx with one txout (locked to the given pkscript) for every item in the amounts slice. I've changed it to look like this:

```golang
// createTxWithOutputs creates a MsgTx with one TxOut (locked to the given
// pkScript) for every amount in the given slice.
func createTxWithOutputs(amounts []int64, pkScript []byte) *wire.MsgTx {
    ...
}
```

I aso added a comment here and renamed idx to txOutIdx. Do you think that's clear enough?

jrick: Oh.. ok I misread that and am confusing inputs with outputs.  It's referencing the one output, since you only passed it one amount.  Why not pass a slice (or varargs) of wire.TxOut instead?

gsalgado: Yeah, that makes it a lot clearer. Done

PR #347, opened 2016-01-31
Generate a seed when creating a new wallet if none was provided.

Roasbeef:
Previously, if a nil seed was passed into loader.CreateNewWallet, a
random seed was never generated. This would cause an error within the
waddrmgr due to the seed being of invalid (0) length.


davecgh:
Looks correct to me.

OK

jrick:
ok

PR #348, opened 2016-01-31
Allow spending from the waddrmgr.ImportedAddrAccount via wallet.SpendPairs.

Roasbeef:
Previously, when creating a change address during the process of creating a new transaction an error case would be hit in the `waddrmgr` triggered by attempting to derive a new internal address from under a `waddrmgr.ImportedAddrAccount`. To remedy this error, we now use the default account for change when spending outputs from an imported key. This approach allows funds under the control of imported private keys to be protected under the wallet's seed as soon as they've been partially spent.

### DIFF
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -216,6 +216,13 @@ func createTx(eligible []wtxmgr.Credit,
 		feeEst = minimumFee(feeIncrement, szEst, msgtx.TxOut, inputs, bs.Height, disallowFree)
 	}
 
+	// If we're spending the outputs of an imported address, we default
+	// to generating change addresses from the default account.
+	prevAccount := account
+	if account == waddrmgr.ImportedAddrAccount {
+		account = waddrmgr.DefaultAccountNum
+	}
+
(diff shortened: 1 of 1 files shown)


davecgh:
I think this is fine as a temporary measure.  Ultimately, I think the wallet account handling needs to be improved such that each account can have multiple extended keys, imported keys, etc.

jrick:
Please add a warning log so that users aren't clueless about why funds magically moved from the one account to another.

Roasbeef:
Added a log warning. How's it look to you?

jrick:
That log needs to only happen when using imported for the source account.

wallet/createtx.go
```diff
@@ -234,6 +241,7 @@ func createTx(eligible []wtxmgr.Credit,
 			if err != nil {
 				return nil, err
 			}
+
```
jrick: Everything looks good now except this extra newline added.  Can you rebase+squash?

Roasbeef:
Done :+1:

PR #350, opened 2016-02-01
Always lock waddrmgr for every timeout or explicit request.

### LINKED ISSUE #349: Explicitly locking the wallet does not work and leaves the key in memory indefinitely.
Reported in decred/dcrwallet#13

jrick:
Use waddrmgr.IsError to avoid logging errors when trying to lock an
already locked wallet.

Fixes #349.


davecgh:
Looks good.

OK

PR #351, opened 2016-02-01
Return WWW-Authenticate header

jcvernaleo:
On invalide credentials return WWW-Authenticate the same way btcd does.

Pointed out by @davec in decred/dcrwallet#14


jrick:
ok

PR #354, opened 2016-02-01
Set account field in listtransactions result.

### LINKED ISSUE #353: Set account field in listtransactions result.
Reported in decred/dcrwallet#8

jrick:
Fixes #353.

### DIFF
--- a/rpc/legacyrpc/server.go
+++ b/rpc/legacyrpc/server.go
@@ -760,7 +760,8 @@ func (t relevantTx) notificationCmds(w *wallet.Wallet) []interface{} {
 		return nil
 	}
 
-	ltr := wallet.ListTransactions(details, syncBlock.Height, w.ChainParams())
+	ltr := wallet.ListTransactions(details, w.Manager, syncBlock.Height,
+		w.ChainParams())
 	ntfns := make([]interface{}, len(ltr))
 	for i := range ntfns {
 		ntfns[i] = btcjson.NewNewTxNtfn(ltr[i].Account, ltr[i])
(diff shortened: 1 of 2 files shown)


jrick:
```
=== RUN TestThrottle
--- FAIL: TestThrottle (0.05s)
rpcserver_test.go:54: status codes: want: map[200:1 429:1], got: map[200:2]
```

hmmmmm....

jrick:
@richiela this is a first attempt at fixing this issue, but i'll have to go back and compare it more closely against bitcoind behavior.  Right now it is always using the address for the tx output script (if any) to lookup the account, but this only makes sense when dealing with outputs.  It's possible that inputs (the "send" category) use the account that this tx is debiting from.  We don't have any way to do that other than look up the previous outputs and figure out what account they came from, and it would produce weird results if a tx was added to the wallet that debited from multiple accounts.

richiela:
This should be fine.  The account field isn't really useful in sends especially since sends can come from multiple inputs.  Having it on receive is the important part.

jrick:
I think that I'll use this method for setting the receive/immature/generate categories but leave the send category unset.  In bitcoin core, every wallet tx has an associated "from account" string member, but I'm unsure how core is determining that information, and would rather leave it unset that fill it with different data.

jrick:
was just waiting for OKs

jcvernaleo:
So i think I said it in the wrong place (on a commit or something) but OK from me.

jrick:
Rebased and ff-merged as b480a0a09ddc5b0fd4aa2cf8267be3f596560b9e

PR #357, opened 2016-02-02
Do not error for empty tx comments.

### LINKED ISSUE #356: Do not error in sendfrom/sendmany/sendtoaddress handlers for empty strings used for tx comments.
Reported in decred/dcrwallet#12.

jrick:
This prevents the server from returning an error when empty strings
are passed as parameters for transaction comments for the
sendfrom/sendmany/sendtoaddress RPCs.  Non-empty strings will still
cause errors since transaction comments are not saved.

Fixes #356.


davecgh:
OK

PR #358, opened 2016-02-05
Update README for gRPC changes.

jrick:
While here, remove mention of btcgui and add a disclaimer that total
compatibility with the JSON-RPC server cannot be achieved.


jcvernaleo:
Looks good to me.  OK

PR #362, opened 2016-02-11
Add option for one time TLS keys.

### LINKED ISSUE #359: Add option for one time TLS keys.
For graphical clients that manage the lifetime of the wallet process, or any other clients that always interact with a local wallet, the certificate file can change each time and will be dynamically picked up.  In this case, there is no need for a persistent TLS certpair to be saved to disk.  For these clients, the wallet could generate a new TLS cert and key each time it starts, writing the certificate to disk but NOT the key.  This would prevent rogue processes with access to the wallet data directory from using the TLS key to decrypt sensitive information passed over the RPC connection, such as the wallet passphrase.

I propose adding a new option `--onetimetlskey` to enable this mode.  If both TLS cert and key files already exist, the wallet will error on startup because this could mean the existing cert has been copied to other computers.  If neither file exists, or just the cert file exists, a new cert will be written (possibly overwriting the previous file).

jrick:
This option prevents the RPC server TLS key from ever being written to
disk.  This is performed by generating a new certificate pair each
startup and writing (possibly overwriting) the certificate but not the
key.

Closes #359.

### DIFF
--- a/config.go
+++ b/config.go
@@ -94,6 +94,7 @@ type config struct {
 	// aren't considered legacy.
 	RPCCert                string   `long:"rpccert" description:"File containing the certificate file"`
 	RPCKey                 string   `long:"rpckey" description:"File containing the certificate key"`
+	OneTimeTLSKey          bool     `long:"onetimetlskey" description:"Generate a new TLS certpair at startup, but only write the certificate to disk"`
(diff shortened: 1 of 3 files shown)


davecgh:
OK

jrick:
Thanks for review.  I'll add a line to the example config and then merge this in.

PR #363, opened 2016-02-11
If wallet already exists show path.

jcvernaleo:
This was initially pointed out by davec in
decred/dcrwallet#18


jrick:
I like the idea, but could you change it to print it as "The wallet database file `/path/to/wallet.db` already exists"?  That's a little more consistent with how the rest of the project formats errors such as these.

jrick:
the CIs are failing this but it's not from this change.  Appears to be HTTP 200 errors from some of the non-github deps.  I'll merge shortly anyways.

PR #365, opened 2016-02-12
Create results for each wallet.TransactionSummary output.

jrick:
The gRPC server requires this as it must include the properties of
every transaction output, not just those that are controlled by the
wallet.


davecgh:
OK

PR #367, opened 2016-02-15
Make walletpassphrase with timeout=0 never lock the wallet.

jrick:
This broke when the Unlock API changed to replace the timeout in
seconds with a time.Time channel.


jcvernaleo:
OK

PR #370, opened 2016-02-23
Follow specification for FundTransaction RPC.

jrick:
The specification states that even when the target output amount
cannot be reached, all available outputs should still be included in
the response and it is up to the caller to check that the target can
be satisified or not.  Follow this behavior by not erroring when the
target was not met.


davecgh:
OK

PR #372, opened 2016-02-23
Regenerate gRPC files from descriptor.

jrick:


davecgh:
I reviewed the generated code and decoded the new gzipped `fileDescriptor0` data to ensure it doesn't contain anything it's not supposed to.

OK

PR #373, opened 2016-02-23
Add Kind enum to NextAddress request.

### LINKED ISSUE #371: gRPC api needs way to set a change output script just before signing a transaction
The only way a change address can be created currently is by setting `include_change_script` to true in a FundTransaction request.  This was intended to prevent generating change addresses unless a transaction is actually being created that will need one.  However, for clients using fee estimation logic, many calls to FundTransaction may be required, but only a single change address is needed just before the transaction is signed.

I see two ways to fix this:

1. Add an enum to the NextAddress request specifying the account's address type.  An enum is better than introducing a brand new RPC in case the wallet account abstraction changes to represent more than BIP0044 accounts in the future.

2. Mark an output with a null script as change, and have wallet create a change script for it in the SignTransaction RPC.  This could be a little more efficient as it removes a round trip and it can prevent creating unnecessary change addresses, but I think the API is a little less elegant.

Background: Paymetheus is being modified to fetch a new output set for some target amount whenever the outputs are modified by the user.  This is done so that an estimated fee can be calculated and a change output created.  However, with the current FundTransaction RPC, a new change address is generated by the wallet each time the outputs are queried, resulting in far more internal addresses being generated than used.  It would make more sense to estimate the fee without this change address, and only when the tx is about to be signed, create the change output script.

jrick:
Fixes #371.

Depends on #372.


davecgh:
OK

PR #376, opened 2016-02-24
Add VersionService to gRPC server to query server versions.

jrick:


davecgh:
OK (if AppVeyor ever decides to finish).

PR #377, opened 2016-02-28
Refactor wallet transaction creation code.

jrick:
This began as a change to improve the fee calculation code and evolved
into a much larger refactor which improves the readability and
modularity of all of the transaction creation code.

Transaction fee calculations have been switched from full increments
of the relay fee to a proportion based on the transaction size.  This
means that for a relay fee of 1e3 satoshis/kB, a 500 byte transaction
is only required to pay a 5e2 satoshi fee and a 1500 byte transaction
only need pay a 1.5e3 fee.  The previous code would end up estimating
these fees to be 1e3 and 2e3 respectively.

Because the previous code would add more fee than needed in almost
every case, the transaction size estimations were optimistic
(best/smallest case) and signing was done in a loop where the fee was
incremented by the relay fee again each time the actual size of the
signed transaction rendered the fee too low.  This has switched to
using worst case transaction size estimates rather than best case, and
signing is only performed once.

Transaction input signature creation has switched from using
txscript.SignatureScript to txscript.SignTxOutput.  The new API is
able to redeem outputs other than just P2PKH, so the previous
restrictions about P2SH outputs being unspendable (except through the
signrawtransaction RPC) no longer hold.

Several new public packages have been added:

wallet/txauthor - transaction authoring and signing
wallet/txrules - simple consensus and mempool policy rule checks

Along with some internal packages:

wallet/internal/txsizes - transaction size estimation
internal/helpers - context free convenience functions

The txsizes package is internal as the estimations it provides are
specific for the algorithms used by the txrules and txauthor packages.

**PR Note**: The old transaction creation tests tested the internal APIs and will need to be rewritten or replaced for the new code.  This PR will fail the CI infrastructure tests until this is resolved.


davecgh:
I haven't looked at this specifics yet, but from reading the description I definitely give this a concept ack.  Splitting up the transaction authoring, fee calculation and estimations, and policy is something that has been sorely needed for quite some time.

Along the same lines, the transaction signing code in `btcd/txscript` needs a serious look as well.

jrick:
There's a bit of overlap in functionality between the new AddAllInputTransactions and wallet's old SignTransaction that I had to add to support signing transactions in the new grpc server.  I'd rather keep the old API around and not try to combine the two, at least at this point, because that API includes a lot of other features needed for the legacy signrawtransaction, such as setting the sighash flags and reporting all of the signature errors.

jrick:
Tests rewritten to test fee and change output value calculations.  This is ready for review now.

wallet/txauthor/author.go
```diff
@@ -0,0 +1,187 @@
...
+import (
+	"errors"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcwallet/wallet/txfees"
+
+	h "github.com/btcsuite/btcwallet/internal/helpers"
```
jcvernaleo: I don't particularly care but I think short names for imports aren't used anywhere else in the code base.

wallet/createtx.go
```diff
@@ -1,125 +1,88 @@
...
+	"github.com/btcsuite/btcd/btcec"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcutil"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet/txauthor"
 	"github.com/btcsuite/btcwallet/wtxmgr"
 )
 
-const (
```
jcvernaleo: I realize these constants aren't used any more, but comments like this have proved very helpful in the past (writing decred tests and the like).
Just wondering if this stuff is reproduced elsewhere or if maybe we could save it in documentation somewhere.  Not really sure.

jcvernaleo:
Looks like a pretty big improvement over the old stuff to me.

wallet/createtx.go
jrick: There are new consts in the txsizes package.  However, the new constants are maximum estimates rather than the minimum estimates that were previously here (and removed since they would be incorrect for the new algorithms).

jrick:
I just remembered, this change should also probably deprecate or remove the disablefree option from the config, since it no longer determines whether a transaction's priority is high enough that no fee is needed.

wallet/txauthor/author.go
jrick: The `walletrpc` package, which is generated by `protoc`, is imported using the `pb` qualifier, so that's one example.  There may be others but that's the only one I've recently used.

rpc/legacyrpc/methods.go
```diff
@@ -1384,14 +1385,36 @@ func LockUnspent(icmd interface{}, w *wallet.Wallet) (interface{}, error) {
 	return true, nil
 }
 
+func makeOutputs(pairs map[string]btcutil.Amount, chainParams *chaincfg.Params) ([]*wire.TxOut, error) {
```
davecgh: Could probably use a comment here.  It's not horribly hard to figure out the intention, but it might be useful to document the outputs are created in random order.

wallet/createtx.go
```diff
@@ -1,125 +1,87 @@
...
-			"%s input but only %v spendable", total, e.in)
+func (s secretSource) GetKey(addr btcutil.Address) (*btcec.PrivateKey, bool, error) {
+	ma, err := s.Address(addr)
+	if err != nil {
+		return nil, false, err
 	}
-	return fmt.Sprintf("insufficient funds: transaction requires %s input "+
-		"(%v output + %v fee) but only %v spendable", total, e.out,
-		e.fee, e.in)
+	privKey, err := ma.(waddrmgr.ManagedPubKeyAddress).PrivKey()
```
davecgh: This and `GetScript` below, should be checking the type assertion for validity imo.  It's way too easy to call these functions with any type of address and end up getting back a managed address of the wrong type which will lead to a panic.

wallet/createtx.go
```diff
@@ -380,122 +207,17 @@ func (w *Wallet) findEligibleOutputs(account uint32, minconf int32, bs *waddrmgr
...
-	}
-
-	return nil
-}
-
-func validateMsgTx(msgtx *wire.MsgTx, prevOutputs []wtxmgr.Credit) error {
-	for i := range msgtx.TxIn {
-		vm, err := txscript.NewEngine(prevOutputs[i].PkScript,
-			msgtx, i, txscript.StandardVerifyFlags, nil)
+func validateMsgTx(tx *wire.MsgTx, prevScripts [][]byte) error {
```
davecgh: Needs a comment.  Particularly about how the order of the scripts here has to match up with the order they are consumed by the tx inputs.  That bit is important because it means the caller has to ensure it actually reorders then correctly if it randomizes the transaction's inputs.

wallet/txauthor/cprng.go
```diff
@@ -0,0 +1,39 @@
...
+	_, err := rand.Read(buf)
+	if err != nil {
+		panic("Failed to seed prng: " + err.Error())
+	}
+
+	seed := int64(binary.LittleEndian.Uint64(buf))
+	cprng.r = mrand.New(mrand.NewSource(seed))
+}
+
+func (c *cprngType) Int31n(n int32) int32 {
```
davecgh: Exported function missing comment.

wallet/wallet.go
```diff
@@ -2101,11 +2110,11 @@ func (w *Wallet) SignTransaction(tx *wire.MsgTx, hashType txscript.SigHashType,
 			key, err := pka.PrivKey()
 			if err != nil {
 				return nil, false, err
+
```
davecgh: Extra space.

wallet/txrules/rules.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/btcsuite/btcutil"
+)
+
+// DefaultRelayFeePerKb is the default minimum relay fee policy for a mempool.
+const DefaultRelayFeePerKb btcutil.Amount = 1e3
+
+// IsDust determines whether a transaction output value and script length would
+// cause the output to be considered dust.  Transactions with dust outputs are
+// not standard and are rejected by mempools with default policies.
+func IsDust(amount btcutil.Amount, scriptSize int, relayFeePerKb btcutil.Amount) bool {
```
davecgh: Seems like this really ought to be using the code from `btcd` which isn't exported yet.  We need to get the policy code off into a policy package to make that feasible though.

wallet/txauthor/author.go
```diff
@@ -0,0 +1,200 @@
...
+type InputSource func(target btcutil.Amount) (total btcutil.Amount, inputs []*wire.TxIn, scripts [][]byte, err error)
+
+// InputSourceError describes the failure to provide enough input value from
+// unspent transaction outputs to meet a target amount.  A typed error is used
+// so input sources can provide their own implementations describing the reason
+// for the error, for example, due to spendable policies or locked coins rather
+// than the wallet not having enough available input value.
+type InputSourceError interface {
+	error
+	InputSourceError()
```
Roasbeef: What's the purpose of this method on the interface? It isn't used anywhere within this diff.

wallet/txauthor/author.go
```diff
@@ -0,0 +1,200 @@
...
+	PrevScripts [][]byte
+	TotalInput  btcutil.Amount
+	ChangeIndex int // negative if no change
+}
+
+// ChangeSource provides P2PKH change output scripts for transaction creation.
+type ChangeSource func() ([]byte, error)
+
+// NewUnsignedTransaction creates an unsigned transaction paying to one or more
+// non-change outputs.  An appropiate transaction fee is included based on the
```
Roasbeef: appropiate -> appropriate

wallet/txauthor/author.go
```diff
@@ -0,0 +1,200 @@
...
+			if err != nil {
+				return nil, err
+			}
+			if len(changeScript) > txsizes.P2PKHPkScriptSize {
+				return nil, errors.New("fee estimation requires change " +
+					"scripts no larger than P2PKH output scripts")
+			}
+			change := wire.NewTxOut(int64(changeAmount), changeScript)
+			l := len(outputs)
+			unsignedTransaction.TxOut = append(outputs[:l:l], change)
```
Roasbeef: Is the three-index slice really necessary here? I'm under the impression that three-index slices are only useful if one would like to re-slice a slice, with the new slice having only limited access to the capacity of the original slice. 

It seems that: 
```
unsignedTransaction.TxOut = append(outputs, change)
```
would suffice.

wallet/txrules/rules.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Dust is defined as an output value where the total cost to the network
+	// (output size + input size) is greater than 1/3 of the relay fee.
+	return int64(amount)*1000/(3*int64(totalSize)) < int64(relayFeePerKb)
+}
+
+// OutputIsDust determines whether a transaction output value is considered
+// dust.  Transactions with dust outputs are not standard and are rejected by
+// mempools with default policies.
+func OutputIsDust(output *wire.TxOut, relayFeePerKb btcutil.Amount) bool {
```
davecgh: Mempools also treat unspendable outputs as dust with the exception of the special case for `nulldata` (a.k.a standard `OP_RETURN`) outputs.  I don't see that logic handled in here.

davecgh: Mempools also treat unspendable outputs as dust with the exception of the special case for `nulldata` (a.k.a standard `OP_RETURN`) outputs.  I don't see that logic handled in here.

wallet/txrules/rules.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	if OutputIsDust(output, relayFeePerKb) {
+		return ErrOutputIsDust
+	}
+	return nil
+}
+
+// FeeForSerializeSize calculates the required fee for a transaction of some
+// arbitrary size given a mempool's relay fee policy.
+func FeeForSerializeSize(relayFeePerKb btcutil.Amount, txSerializeSize int) (fee btcutil.Amount) {
```
davecgh: Named return?

wallet/txauthor/author.go
```diff
@@ -0,0 +1,200 @@
...
+	txscript.ScriptDB
+	ChainParams() *chaincfg.Params
+}
+
+// AddAllInputScripts modifies transaction a transaction by adding inputs
+// scripts for each input.  Previous output scripts being redeemed by each input
+// are passed in prevPkScripts and the slice length must match the number of
+// inputs.  Private keys and redeem scripts are looked up using a SecretsSource
+// based on the previous output script.
+func AddAllInputScripts(tx *wire.MsgTx, prevPkScripts [][]byte, secrets SecretsSource) error {
```
Roasbeef: Any interest in including a similar function: `AddControlledInputScripts` (feel free to suggest a better name :tongue:), which would only attempt to sign an input if the wallet can discern if it is in control os said input? 

Such a function would be useful to prospective callers who wish to engage in multi-party collaborative transaction creation+signing (CoinJoin, Lightning, etc)

wallet/txauthor/author.go
```diff
@@ -0,0 +1,200 @@
...
+// necessary, fetchChange is called zero or one times to generate this script.
+// This function must return a P2PKH script or smaller, otherwise fee estimation
+// will be incorrect.
+//
+// If successful, the transaction, total input value spent, and all previous
+// output scripts are returned.  If the input source was unable to provide
+// enough input value to pay for every output any any necessary fees, an
+// InputSourceError is returned.
+//
+// BUGS: Fee estimation may be off when redeeming non-compressed P2PKH outputs.
```
davecgh: This isn't a bug.  It's simply a limitation of the function.

wallet/txauthor/author.go
davecgh: It was described in the comment.  It's so the caller can choose to return a more detailed error if they want while allowing easy detection of an `InputSourceError` versus some other type of error such as a database failure, etc.  Check out how the tests switch on the error type.

For example:

```Go
type CustomInputSourceError struct {
	SomeImportantField1 int
	SomeImportantField2 int
}

// Make sure it is identifiable by the caller as an `InputSourceError`.
func (c *CustomInputSourceError) InputSourceError() {}

// .... Later in the caller somewhere...

if _, ok := err.(txauthor.InputSourceError); ok {
	// Handle any type of input source error differently
	// regardless of whether it was a custom one or the
	// standard implementation provided by the package.
} else {
	// Some other type of unexpected error.  Handle it differently.
}
```

wallet/createtx.go
```diff
@@ -1,125 +1,87 @@
...
-// unspent output is eligible for spending. Leftover input funds not sent
-// to addr or as a fee for the miner are sent to a newly generated
-// address. InsufficientFundsError is returned if there are not enough
-// eligible unspent outputs to create the transaction.
-func (w *Wallet) txToPairs(pairs map[string]btcutil.Amount, account uint32, minconf int32) (*CreatedTx, error) {
-
+// txToOutputs creates a signed transaction which includes each output from
+// outputs.  Previous outputs to reedeem are chosen from the passed account's
+// UTXO set and minconf policy. An additional output may be added to return
+// change to the wallet.  An appropiate fee is included based on the wallet's
```
davecgh: appropiate -> appropriate

wallet/txauthor/author.go
davecgh: It is needed imo.  The three-index slice ensures a new output slice is created instead of modifying the caller slice which could lead to overwriting data depending on how the caller passed it in.

Observe the following:
```Go
package main

import "fmt"

func Foo(outputs []int) {
	l := len(outputs)
	foo := append(outputs[:l:l], 100)
	_ = foo
}

func Foo2(outputs []int) {
	foo := append(outputs, 100)
	_ = foo
}

func main() {
	allOutputs := make([]int, 6)
	for i := 0; i < len(allOutputs); i++ {
		allOutputs[i] = i
	}
	curOutputs := allOutputs[0:4]

	// Notice how the call to Foo does not modify the
	// backing array.
	fmt.Println("allOutputs before Foo:", allOutputs)
	Foo(curOutputs)
	fmt.Println("allOutputs after Foo:", allOutputs)

	// Notice how the call to Foo2 ends up modifying
	// the backing array.
	fmt.Println("allOutputs before Foo2:", allOutputs)
	Foo2(curOutputs)
	fmt.Println("allOutputs after Foo2:", allOutputs)
}
```
Output:
```
allOutputs before Foo: [0 1 2 3 4 5]
allOutputs after Foo: [0 1 2 3 4 5]
allOutputs before Foo2: [0 1 2 3 4 5]
allOutputs after Foo2: [0 1 2 3 100 5]
```

wallet/txauthor/author.go
```diff
@@ -0,0 +1,200 @@
...
+import (
+	"errors"
+
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcwallet/wallet/txrules"
+
+	h "github.com/btcsuite/btcwallet/internal/helpers"
```
davecgh: I'm not a fan of this.  `h` is too easy to shadow imo.

wallet/txauthor/author.go
```diff
@@ -0,0 +1,200 @@
...
+					"scripts no larger than P2PKH output scripts")
+			}
+			change := wire.NewTxOut(int64(changeAmount), changeScript)
+			l := len(outputs)
+			unsignedTransaction.TxOut = append(outputs[:l:l], change)
+			changeIndex = l
+		}
+
+		return &AuthoredTx{
+			unsignedTransaction,
```
davecgh: Please specify the field names.  It makes it much more readable and is more consistent with the code just about it that specifies them.

wallet/txauthor/author.go
davecgh: I think that's a fine idea.  The `txscript` package is already capable of doing sparse signing, so it shouldn't be too difficult to make work.

wallet/createtx.go
jrick: I'll fix but it seems unlikely, and would be a bug in txscript (which calls these functions through the SecretSource interface).

wallet/txauthor/cprng.go
jrick: It's an unexported type.  Intention was to be analogous to math/rand's Int32n, except this one is concurrent safe.

wallet/txrules/rules.go
jrick: I'm aware.  This code was ported from the paymetheus fee estimation branch and I didn't have any way to check that there.

wallet/txrules/rules.go
jrick: Not an issue imo but will change it

wallet/txauthor/author.go
jrick: I like the idea too but the code wouldn't be used anywhere in wallet currently, except maybe if we hooked it up to Wallet.SignTransaction (used by signrawtransaction).  I'll hold off on it for this PR, but it's a good suggestion.

wallet/txauthor/author.go
jrick: I'd contend that it is a bug.  We could make a better guess about the serialize size cost of inputs based on the kinds of output scripts that are returned by the input source.

wallet/txauthor/author.go
jrick: It also hides bugs if you miss a field.  I don't agree with the readability argument either (none of the types are the same, so it's impossible to assign them to the wrong field, and you can be assured that none of the fields have been missed).

But I'll change it because others feel strongly about it and I'm not about to argue a non issue.

wallet/txrules/rules.go
jrick: I'll rename IsDust to IsDustAmount, OutputIsDust to IsDustOutput (for consistency), and check for unspendable outputs in IsDustOutput.

wallet/txauthor/author.go
jrick: @davecgh is right.  The function makes no guarantees about it, so a defensive copy is necessary.

In the common case (using sendfrom/many/toaddress from the json-rpc server) it won't make any performance difference since the output array created by the legacyrpc package has the equal len and caps.

wallet/txauthor/author.go
jrick: everything is too easy to shadow :p

jrick: Also, any editor with "find all references" (e.g. https://github.com/redefiance/go-find-references) would show that it is a package name.  I don't see this being a real issue.

What is an issue is that this package is needed at all to reduce the verbosity of looping through the outputs just to sum various fields.

wallet/txauthor/author.go
Roasbeef: I see. Thanks for the illuminating example @davecgh! I recall the three-index slice was introduced in Go 1.2 or so, but wasn't' aware of some of the motivating use cases until this discussion.

davecgh:
It looks like everything has been addressed.  I'm really happy to see this refactored into more reusable and better tested code.  Nice job!

OK

PR #378, opened 2016-02-29
Use LICENSE file and short license headers.

jrick:


davecgh:
Should the years in all of the modified files be updated to include 2016 since they're being touched?  It's debatable since the code isn't actually changing.

I'm fine with it as it is too.

jrick:
I don't think the license header itself is covered under any copyright. IANAL

jcvernaleo:
I'm in favor of this for the sake of consistency (and shorter headers).

PR #384, opened 2016-03-09
Protect the relay fee field with a mutex.

### LINKED ISSUE #379: settxfee races
The legacy RPC server's settxfee RPC causes data races when modifying the wallet's relay fee.  Synchronization is needed.

jrick:
This prevents races when setting a new relay fee through the legacy
RPC server (settxfee).

Fixes #379.


davecgh:
OK

PR #385, opened 2016-03-09
Fix spelling in legacy test and docs using aspell

ysangkok:


davecgh:
Nice catches  Looks good to me.

OK

jrick:
Thanks for the fixes!

PR #387, opened 2016-03-10
Block database closes until open transactions finish.

jrick:


jrick:
Closing this in favor of a better fix.

The real issue is that the database is being closed from main while the wallet has not yet been stopped.  This appears to have broken when the wallet code was refactored out into a separate wallet package, as wallet processes were no longer being stopped from an interrupt handler.

The correct fix is to use the loader to close any loaded wallets from the interrupt handler.  This will also simplify the loader callbacks, as it will no longer need to pass the database as a parameter for the main function to close it.

PR #389, opened 2016-03-11
Default to mainnet.

### LINKED ISSUE #383: Default to mainnet
After several years of development I think it's time to change this.  The software has been used in production on mainnet for quite some time on various projects, and we have no reason to deter other users from using the software on mainnet either.

jrick:
Note that this is a breaking change since it removes the mainnet
config option, replacing it with a testnet option.  Old configuration
files that set mainnet=1 will cause the wallet to error during startup
since extraneous flags are treated as errors.

Closes #383.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,12 +54,6 @@ Wallet clients can use one of two RPC servers:
      notifications for changes to the wallet, this is the RPC server to use.
      The gRPC server is documented [here](./rpc/documentation/README.md).
 
-Mainnet support is currently disabled by default.  Use of btcwallet on
-mainnet requires passing the `--mainnet` flag on the command line or
-adding `mainnet=1` to the configuration file.  Mainnet will be enabled
-by default in a future release after further database changes and
-testing.
-
 ## Installation and updating
 
 ### Windows - MSIs Available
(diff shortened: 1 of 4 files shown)


jrick:
I wonder if the other deprecated options should be removed here as well, since if we're going to require users modify their existing configs, might as well remove all of the unused options.

README.md
```diff
@@ -85,9 +79,9 @@ go get -u -v github.com/btcsuite/btcwallet/...
 
 ## Getting Started
 
-The following instructions detail how to get started with btcwallet
-connecting to a localhost btcd.  Commands should be run in `cmd.exe`
-or PowerShell on Windows, or any terminal emulator on *nix.
+The following instructions detail how to get started with btcwallet connecting
+to a localhost testnet btcd.  Commands should be run in `cmd.exe` or PowerShell
```
davecgh: Shouldn't this all just be changed over to mainnet since it's the Getting Started section?

davecgh:
I agree with removing any deprecated options since the config will have to modified anyways.

README.md
jrick: Could, although I do worry about whether new users "getting started" will properly secure their wallet seed.

jrick: Actually, I'm ok with that.  If you're a new user and using it directly on the command line instead of through another client, I'd expect you to have some knowledge about wallet security.

davecgh: Makes sense.  I would actually like to see the documentation improved quite a bit to urge Windows users towards Paymetheus, but that's out of scope for this PR.

davecgh:
OK

PR #390, opened 2016-03-11
Disallow naming accounts the empty string.

### LINKED ISSUE #369: gRPC allows renaming accounts to the empty string.
This should be forbidden, as it is in the legacy JSON-RPC server.  We should prevent this at the waddrmgr level.

jrick:
This change only prevents creating new accounts with the empty name or
renaming an existing account to one.  Any accounts in the DB that are
already named the empty string are left untouched (and should be
renamed to something meaningful by the user).

Fixes #369.

### DIFF
--- a/waddrmgr/manager.go
+++ b/waddrmgr/manager.go
@@ -1711,6 +1711,10 @@ func (m *Manager) LastInternalAddress(account uint32) (ManagedAddress, error) {
 
 // ValidateAccountName validates the given account name and returns an error, if any.
 func ValidateAccountName(name string) error {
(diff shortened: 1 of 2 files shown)


dajohi:
OK

davecgh:
I agree with the change, but will this break existing wallets that already have an account with an empty name unless the database version is bumped and it is given a name?

jrick:
It won't.  Existing accounts already named the empty string can still be queried by that name.

davecgh:
Alright.  Looks good then.

OK

PR #391, opened 2016-03-11
Remove legacy JSON-RPC notifications.

jrick:
These notifications were added to support real time updates for
btcgui.  As the btcgui project is no longer being developed, there are
no more consumers of this API, and it makes sense to remove them given
their various issues (the largest being that notifiations are sent
unsubscribed to clients that may never be interrested in them).

A new notification server has already been added to the wallet package
to handle notifications in a RPC-server agnostic way.  This server is
the means by which the wallet notifies changes for gRPC clients.  If
per-client registered notifications are to be re-added for the
JSON-RPC server, they should be integrated with the new notification
server rather than using this legacy code.

### DIFF
--- a/rpc/legacyrpc/server.go
+++ b/rpc/legacyrpc/server.go
@@ -19,10 +19,8 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcjson"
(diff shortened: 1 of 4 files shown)


davecgh:
OK

PR #393, opened 2016-03-11
Log wallet locks and unlocks.

### LINKED ISSUE #392: Add logging for (un)locking of the wallet.
Currently, all locking and unlocking activity is unlogged except when there is some kind of unexpected error.

jrick:
Fixes #392.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -498,6 +498,11 @@ out:
 				continue
 			}
 			timeout = req.lockAfter
+			if timeout == nil {
+				log.Info("Wallet unlocked without timeout")
+			} else {
+				log.Info("Wallet temporarily unlocked")
+			}
 			req.err <- nil
 			continue
 
@@ -544,6 +549,8 @@ out:
 		err := w.Manager.Lock()
 		if err != nil && !waddrmgr.IsError(err, waddrmgr.ErrLocked) {
 			log.Errorf("Could not lock wallet: %v", err)
+		} else {
+			log.Info("Wallet locked")
 		}
 	}
 	w.wg.Done()
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -499,9 +499,9 @@ out:
 			}
 			timeout = req.lockAfter
 			if timeout == nil {
-				log.Info("Wallet unlocked without timeout")
+				log.Info("The wallet has been unlocked without a time limit")
 			} else {
-				log.Info("Wallet temporarily unlocked")
+				log.Info("The wallet has been temporarily unlocked")
 			}
 			req.err <- nil
 			continue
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -498,6 +498,11 @@ out:
 				continue
 			}
 			timeout = req.lockAfter
+			if timeout == nil {
+				log.Info("Wallet unlocked without timeout")
```
davecgh: These should probably match was `dcrwallet` says:

```Go
log.Info("The wallet has been unlocked without a time limit.")
```

wallet/wallet.go
```diff
@@ -498,6 +498,11 @@ out:
...
+			} else {
+				log.Info("Wallet temporarily unlocked")
```
davecgh:
```Go
log.Infof("The wallet has been unlocked. This is set to expire in %v.", timeout)
```

wallet/wallet.go
```diff
@@ -544,6 +549,8 @@ out:
 		err := w.Manager.Lock()
 		if err != nil && !waddrmgr.IsError(err, waddrmgr.ErrLocked) {
 			log.Errorf("Could not lock wallet: %v", err)
+		} else {
+			log.Info("Wallet locked")
```
davecgh:
```Go
log.Info("The wallet has been locked.")
```

wallet/wallet.go
jrick: There's no way to know how long the timeout is.  It's a channel, not a duration.

davecgh: Hmmm well `dcrwallet` does the print here: https://github.com/decred/dcrwallet/blob/master/rpc/legacyrpc/methods.go#L3489-L3496

I'd really prefer it to print how long it's being unlocked for like in `dcrwallet`.  That, in my opinion, is much more useful information than just saying temporarily.

jrick: I edited my comment above with why logging the duration would not be a good idea.

davecgh: We were both editing at the same time.  Perhaps the duration should be passed along in the `unlockRequets` so the log has the necessary information?

I understand it's advisory, but it's still useful.  There are really only two cases where it will be wrong:

1) It happens to be on hold the moment it expires, but even then, it will be unlocked as soon as the hold has been released, so it won't be that much of a difference
2) The wallet is manually locked before the time.  That is also fine, in my opinion, because the verbiage says "set to expire in %v".  Just because it is manually locked before it expired doesn't change the fact it is being unlocked with a timeout of X.

Anyways, it's not a big deal, though I would certainly prefer it to be more explicit in the logging.  I see a message like "The wallet has been unlocked temporarily" and the first thing that comes to my mind is "FOR HOW LONG????!??!?!"

jrick: There's no reasonable value to pass in as a timeout for extra logging information from calling code similar to the grpc server.  There is no set timeout, it simply locks the wallet as soon as it no longer needs it.

davecgh: Meh OK.

davecgh:
As mentioned, I don't like the lack of printing how long it's unlocked for, but no reason to hold this up.

OK

PR #396, opened 2016-03-16
Sort unmined transactions in their dependency order.

### LINKED ISSUE #156: Resending of unmined transactions must be done in dependency order.

jrick:
This changes the wtxmgr.Store.UnminedTxs method to sort transactions,
using the Kahn topological sort algorithm, before returning
transactions to the caller.  This is possible because transactions
form a sort of Directed Acyclic Graph (DAG) where transactions
reference other transactions to spend their outputs (multiple
referenced outputs from a single transaction spent by the same
transaction count as a single graph edge).

This prevents the possibility of orphan rejection errors when sending
unmined transactions to a full node at startup.  As these transactions
are sent using the sendrawtransaction RPC, which does not permit
orphans, this topological sort is required.

Fixes #156.

### DIFF
--- a/wtxmgr/kahnsort.go
+++ b/wtxmgr/kahnsort.go
@@ -1,0 +1,115 @@
+// Copyright (c) 2016 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package wtxmgr
+
+import (
+	"github.com/btcsuite/btcd/wire"
+)
+
+type graphNode struct {
+	value    *TxRecord
+	outEdges []*wire.ShaHash
+	inDegree int
+}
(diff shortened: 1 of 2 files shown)


cjepson:
utACK, way cleaner than my code and no extra licensing issues. :P

jcvernaleo:
I tested this on testnet a bit and it seems good to me.
OK

jrick:
The CIs will choke on that rebase since the walletdb/bdb tests are invalid and break after the recent bolt sync.  I'll open a new PR for those in a bit.

PR #398, opened 2016-03-17
Manage wallet db namespaces from wallet package.

### LINKED ISSUE #397: Remove database namespaces from wallet.Open
While it is reasonable to pass an already open database when creating a wallet (and therefore the wallet should not manage the database lifetime), the address and transaction manager namespaces should be managed by the wallet itself, rather than requiring the caller to open these first.

jrick:
This changes the wallet.Open function signature to remove the database
namespace parameters.  This is done so that the wallet package itself
is responsible for the location and opening of these namespaces from
the database, rather than requiring the caller to open these ahead of
time.

A new wallet.Create function has also been added.  This function
initializes a new wallet in an empty database, using the same
namespaces as wallet.Open will eventually use.  This relieves the
caller from needing to manage wallet database namespaces explicitly.

Fixes #397.

### DIFF
--- a/waddrmgr/manager.go
+++ b/waddrmgr/manager.go
@@ -2272,11 +2272,11 @@ func Open(namespace walletdb.Namespace, pubPassphrase []byte, chainParams *chain
 	return loadManager(namespace, pubPassphrase, chainParams)
 }
 
-// Create returns a new locked address manager in the given namespace.  The
-// seed must conform to the standards described in hdkeychain.NewMaster and will
-// be used to create the master root node from which all hierarchical
-// deterministic addresses are derived.  This allows all chained addresses in
-// the address manager to be recovered by using the same seed.
+// Create creates a new address manager in the given namespace.  The seed must
+// conform to the standards described in hdkeychain.NewMaster and will be used
+// to create the master root node from which all hierarchical deterministic
+// addresses are derived.  This allows all chained addresses in the address
+// manager to be recovered by using the same seed.
 //
 // All private and public keys and information are protected by secret keys
 // derived from the provided private and public passphrases.  The public
@@ -2289,26 +2289,26 @@ func Open(namespace walletdb.Namespace, pubPassphrase []byte, chainParams *chain
 //
 // A ManagerError with an error code of ErrAlreadyExists will be returned the
 // address manager already exists in the specified namespace.
-func Create(namespace walletdb.Namespace, seed, pubPassphrase, privPassphrase []byte, chainParams *chaincfg.Params, config *ScryptOptions) (*Manager, error) {
+func Create(namespace walletdb.Namespace, seed, pubPassphrase, privPassphrase []byte, chainParams *chaincfg.Params, config *ScryptOptions) error {
 	// Return an error if the manager has already been created in the given
 	// database namespace.
 	exists, err := managerExists(namespace)
 	if err != nil {
-		return nil, err
+		return err
 	}
 	if exists {
-		return nil, managerError(ErrAlreadyExists, errAlreadyExists, nil)
+		return managerError(ErrAlreadyExists, errAlreadyExists, nil)
 	}
 
 	// Ensure the private passphrase is not empty.
 	if len(privPassphrase) == 0 {
 		str := "private passphrase may not be empty"
-		return nil, managerError(ErrEmptyPassphrase, str, nil)
+		return managerError(ErrEmptyPassphrase, str, nil)
 	}
 
 	// Perform the initial bucket creation and database namespace setup.
 	if err := createManagerNS(namespace); err != nil {
-		return nil, err
+		return err
 	}
 
 	if config == nil {
@@ -2322,15 +2322,16 @@ func Create(namespace walletdb.Namespace, seed, pubPassphrase, privPassphrase []
 	root, err := hdkeychain.NewMaster(seed, chainParams)
 	if err != nil {
 		str := "failed to derive master extended key"
-		return nil, managerError(ErrKeyChain, str, err)
+		return managerError(ErrKeyChain, str, err)
 	}
 
 	// Derive the cointype key according to BIP0044.
 	coinTypeKeyPriv, err := deriveCoinTypeKey(root, chainParams.HDCoinType)
 	if err != nil {
 		str := "failed to derive cointype extended key"
-		return nil, managerError(ErrKeyChain, str, err)
+		return managerError(ErrKeyChain, str, err)
 	}
+	defer coinTypeKeyPriv.Zero()
 
 	// Derive the account key for the first account according to BIP0044.
 	acctKeyPriv, err := deriveAccountKey(coinTypeKeyPriv, 0)
@@ -2339,11 +2340,11 @@ func Create(namespace walletdb.Namespace, seed, pubPassphrase, privPassphrase []
 		// required hierarchy can't be derived due to invalid child.
 		if err == hdkeychain.ErrInvalidChild {
 			str := "the provided seed is unusable"
-			return nil, managerError(ErrKeyChain, str,
+			return managerError(ErrKeyChain, str,
 				hdkeychain.ErrUnusableSeed)
 		}
 
-		return nil, err
+		return err
 	}
 
 	// Ensure the branch keys can be derived for the provided seed according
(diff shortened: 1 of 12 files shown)


wallet/wallet.go
```diff
@@ -2058,13 +2059,58 @@ func (w *Wallet) ChainParams() *chaincfg.Params {
...
+	// we generate a random seed for the wallet with the recommended seed
+	// length.
+	if seed == nil {
+		hdSeed, err := hdkeychain.GenerateSeed(hdkeychain.RecommendedSeedLen)
+		if err != nil {
+			return err
+		}
+		seed = hdSeed
+	}
+	if len(seed) < hdkeychain.MinSeedBytes || len(seed) > hdkeychain.MaxSeedBytes {
```
davecgh: Goes over 80 while the call to `Create` just below, and practically everywhere else, doesn't.

waddrmgr/manager.go
```diff
@@ -2289,26 +2289,26 @@ func Open(namespace walletdb.Namespace, pubPassphrase []byte, chainParams *chain
 //
 // A ManagerError with an error code of ErrAlreadyExists will be returned the
 // address manager already exists in the specified namespace.
-func Create(namespace walletdb.Namespace, seed, pubPassphrase, privPassphrase []byte, chainParams *chaincfg.Params, config *ScryptOptions) (*Manager, error) {
+func Create(namespace walletdb.Namespace, seed, pubPassphrase, privPassphrase []byte, chainParams *chaincfg.Params, config *ScryptOptions) error {
```
davecgh: I think it is a bad idea to remove the functionality from `Create` which returns the address manager already locked.  It means that callers end up stuttering because they have to call `Open`, and if it fails they have to call it again after `Create`.  Ick.

wtxmgr/example_test.go
```diff
@@ -155,7 +155,12 @@ func Example_basicUsage() {
 	}
 
 	// Create (or open) the transaction store in the provided namespace.
-	s, err := wtxmgr.Create(ns)
+	err = wtxmgr.Create(ns)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+	s, err := wtxmgr.Open(ns)
```
davecgh: This is another example of why Create should just return the manager like it used to.  Putting more burden on the caller like this doesn't have any benefit.

davecgh:
I agree with the concept of this PR and it looks pretty good aside from the things I pointed out.

waddrmgr/manager.go
jrick: The reason I made this change was that calling code that created a manager (wallet.Create) would need a special error handling path to close the manager if it was successfully created but a later error was encounted (even though it would not be used yet, that's what wallet.Open is for).

wtxmgr/example_test.go
jrick: wtxmgr doesn't have a Close method, so perhaps this is right.  I disagree about waddrmgr (because Close is needed), and the change to wtxmgr was only done for consistency.

waddrmgr/manager.go
davecgh: Yeah, I get that, but I still think it's a bad idea to make all call sites worse to avoid a single case of `defer mgr.Close()`.

jrick: But if your calling code (unlike how the wallet package now works) actually does end up using the opened manager from the Create func, you can't use defer.  It should only defer when there was an error.  The only "sane" way to do that is to use named returns and check the returned err in a defer before deciding whether to close or not, and I know how much you love named returns.

jrick:
What about a `waddrmgr.Exists(ns walletdb.Namespace)` func? Then calling code responsible for both creating and opening the manager would only need to check Exists before calling Create, while always calling Open after?

waddrmgr/manager.go
davecgh: Looking through the PR, I can see that 8 call sites needed code added to deal with this when realistically only the one for later use in `wallet.Open`.

Instead of deferring then, why not just close the mgr returned from `Create` in the `wallet.Create` function and leave everything else untouched?

jrick: You want code like this?

```
func Create(...) (err error) {
    mgr, err := waddrmgr.Create(...)
    if err != nil {
        return err
    }
    defer func() {
        if err != nil {
            mgr.Close()
        }
    }()

    err := otherStuff()
    if err != nil {
        return err
    }

    return nil
}
```

jrick:
If this really is such an issue then we can just add a CreateAndOpen func or something.  But it is totally stupid to open up the manager when the caller does not want it opened, so we need a way to avoid that.

davecgh:
You mean the following right?

```Go
if !waddrmgr.Exists(ns) {
	if err := waddrmgr.Create(...); err != nil {
		return err
	}
}
mgr, err := waddrmgr.Open(...)
if err != nil {
	return err
}
```

Still seems like a lot of code for each caller, but it's better than Open -> fail, Create, Open.  Honestly, I'd prefer to leave the `Create` function like it is and a `CreateOnly` function to handle the specific one-off case of what you're trying to accomplish.  That way you don't have to change any of the existing code call sites, nor would the txmgr, which doesn't need to be closed, have to updated since it would still be consistent.

Overall, that would greatly reduce the change set and not break a public API.

jrick:
Yes, that's what I had in mind.

This is totally not a one off case.  It is *the* case that these functions are designed for.  Eventually I'd even like to make waddrmgr and wtxmgr (or whatever the equivalent packages are in the future after combining them) internal packages, and all callers operate through the wallet package.

waddrmgr/manager.go
davecgh: Nah.

```Go
func Create(...) error {
	err := waddrmgr.CreateNoOpen(...)
	if err != nil {
		return err
	}

	err = otherStuff()
	if err != nil {
		return err
	}

	return nil
}
```

davecgh:
Anyways, if you absolutely must change the `Create` function instead of just adding a new one to avoid breaking API compatibility, at least add a `CreateAndOpen` for the other call sites so they don't have to add a bunch of code.

jrick:
Just to reiterate the point:

```
PS C:\Users\jrick\go\src\github.com\btcsuite\btcwallet> grep -ri 'waddrmgr.Create' *
votingpool/example_test.go:     err = waddrmgr.Create(
votingpool/example_test.go:     err = waddrmgr.Create(
votingpool/factory_test.go:     err = waddrmgr.Create(mgrNamespace, seed, pubPassphrase, privPassphrase,
waddrmgr/common_test.go:        err = waddrmgr.Create(namespace, seed, pubPassphrase,
waddrmgr/manager_test.go:       err = waddrmgr.Create(mgrNamespace, seed, pubPassphrase,
waddrmgr/manager_test.go:       err = waddrmgr.Create(mgrNamespace, seed, pubPassphrase,
wallet/wallet.go:       err = waddrmgr.Create(addrMgrNamespace, seed, pubPass, privPass,
PS C:\Users\jrick\go\src\github.com\btcsuite\btcwallet> grep -ri 'wtxmgr.Create' *
votingpool/example_test.go:     err = wtxmgr.Create(wtxmgrNamespace)
votingpool/factory_test.go:     err = wtxmgr.Create(wtxmgrNamespace)
wallet/wallet.go:       return wtxmgr.Create(txMgrNamespace)
wallet/wallet.go:                       err = wtxmgr.Create(txMgrNS)
wtxmgr/example_test.go: err = wtxmgr.Create(ns)
```

Outside of tests, there is a single call to waddrmgr.Create in the entire project.  The calling function only needs to create the manager, not return an opened manager for the caller.  There would be no benefit to a CreateAndOpen function except to simplify tests.  However, in that case, I think the tests themselves should provide that convenience function, otherwise we wouldn't be testing the API that the wallet package is actually using.

The wtxmgr story is slightly different.  `wallet.Open` needs to check for the existence of the transaction manger before it can open it (or in the current case, it first opens, and handles ErrNoExists by creating and opening again).  This is required to handle tools like `dropwtxmgr`, which drop the entire namespace, signaling the wallet to recreate it and do a fresh rescan.  This is a case where we would benefit from a `wtxmgr.Exists` func.  Again, `CreateAndOpen` is not useful for this case because the wallet needs to perform other tasks in case the namespace does not exist.

jrick:
Instead of `waddrmgr.Exists` and/or `wtxmgr.Exists`, I think it would make more sense to add a function to walletdb checking whether a namespace is empty or not, as this is a better generalization for the problem.

jrick:
@davecgh are you ok with these changes now?

davecgh:
OK

PR #400, opened 2016-03-21
Print version string at startup.

jcvernaleo:
This makes btcwallet match btcd's behavior.

Initially pointed out in:
decred/dcrwallet#125


jrick:
Travis fails because of what appears to be rate limiting.

OK

PR #402, opened 2016-03-29
Use goclean like the rest of btcsuite.

jcvernaleo:
This cleans up some things caught by goclean's various tests in btcwallet and enables it's use with travis.

This catches btcwallet up with the rest of btcsuite in this regard and will help get dcrwallet to the same point.

I have not squashed this yet in case @jrick takes issue with any particular fixes (or things that I made the tests ignore).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,16 @@
 language: go
 go:
-  - 1.4.3
-  - 1.5.2
+  - 1.5.3
+  - 1.6
 sudo: false
-install: go get -d -t -v ./...
+before_install:
+  - gotools=golang.org/x/tools
+install:
+  - go get -d -t -v ./...
+  - go get -v $gotools/cmd/cover
+  - go get -v $gotools/cmd/vet
+  - go get -v github.com/bradfitz/goimports
+  - go get -v github.com/golang/lint/golint
+script:
+  - export PATH=$PATH:$HOME/gopath/bin
+  - ./goclean.sh
--- a/chain/chain.go
+++ b/chain/chain.go
@@ -342,8 +342,8 @@ out:
 		case dequeue <- next:
 			if n, ok := next.(BlockConnected); ok {
 				bs = &waddrmgr.BlockStamp{
-					n.Height,
-					n.Hash,
+					Height: n.Height,
+					Hash:   n.Hash,
 				}
 			}
 
--- a/goclean.sh
+++ b/goclean.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+# The script does automatic checking on a Go package and its sub-packages, including:
+# 1. gofmt         (http://golang.org/cmd/gofmt/)
+# 2. goimports     (https://github.com/bradfitz/goimports)
+# 3. golint        (https://github.com/golang/lint)
+# 4. go vet        (http://golang.org/cmd/vet)
+# 5. race detector (http://blog.golang.org/race-detector)
+# 6. test coverage (http://blog.golang.org/cover)
+
+set -ex
+
+# Automatic checks
+test -z "$(gofmt -l -w . | tee /dev/stderr)"
+test -z "$(goimports -l -w . | tee /dev/stderr)"
+test -z "$(golint ./... | grep -v 'ALL_CAPS\|OP_\|NewFieldVal\|RpcCommand\|RpcRawCommand\|RpcSend\|Dns\|api.pb.go\|StartConsensusRpc\|factory_test.go\|legacy' | tee /dev/stderr)"
+test -z "$(go tool vet . 2>&1 | grep -v 'Example\|newestSha\|rpcserver/server.go' | tee /dev/stderr)"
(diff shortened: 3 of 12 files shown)


davecgh:
I've reviewed this and agree with all changes.  I particularly like that you fixed the unkeyed fields.

OK

.travis.yml
```diff
@@ -1,6 +1,17 @@
...
 go:
-  - 1.4.3
-  - 1.5.2
+  - 1.5.3
+  - 1.6
 sudo: false
-install: go get -d -t -v ./...
+before_install:
+  - gotools=golang.org/x/tools
+  - if [ "$TRAVIS_GO_VERSION" = "go1.3.3" ]; then gotools=code.google.com/p/go.tools; fi
```
jrick: We're not testing against Go 1.3 anymore so I believe this can be removed.

jrick:
That is the part that i particularly dislike :)

I'd also avoid touching any of the legacy packages, those should be essentially read only at this point.  There's no need to keep updating them.

everything else has an OK though

.travis.yml
jcvernaleo: Good point.  Will kill that.

jcvernaleo:
@jrick would you like me to revert the change in legacy and just have goclean ignore it like it does with the generated stuff?

jrick:
Yep that sounds fine

jcvernaleo:
@jrick if that's all good with you now, let me know and I'll squash.

jrick:
Well neither @davecgh nor I really want to budge on the keyed struct literals, so I'll try to meet somewhere in the middle on that point.  Going through these changes again, the only ones I really dislike changing over to the keyed fields are the outpoints, since anyone even slightly familiar with Bitcoin or the btcsuite code in general knows that an output is exactly two fields: a transaction hash and an output index.  That's certainly not changing anytime soon, and in cases where it does for altcoins (e.g. decred) you absolutely must not forget to set the additional field (tree), or the entire data is probably bogus.  I'm fine with adding the keyed fields for every other type, but please leave the outpoints untouched.

Almost all of those outpoints are in the test code anyways, and it's good to have more static assertions that our test cases are valid.

jcvernaleo:
The OutPoints represent 3/4 of the keyed fields that I added.  I can special case that too but at that point we have a ton of special cases.
And I disagree with the unstated premise of your statement about "anyone even slightly familiar with Bitcoin or the btcsuite code".  That is not the entire audience.  Part of the utility of always using keyed fields is that it makes the code more legible for someone who is NOT currently familiar with the code.  Part of the utility of btcsuite is that is it (imo) a much clearer way for someone not familiar with bitcoin to learned the internals.

jrick:
It also hurts legibility for developers who can no longer assume that all fields are being correctly set.  I've wasted tons of time on these kind of struct declarations on other code reviews because I have no idea whether or not each field has been set to the correct value or not, or if any are missing.  Code readability should focus on developers first, not end users.

Really, my issue is not the naming of the struct fields.  It's that without removing the names, there is no way to guarantee that every field is set.  A function to create the struct doesn't really help either since function parameters can not be named either, and then you lose the flexibility of adding the & at the beginning depending on whether you need a pointer to the struct value, or the struct value itself.

jcvernaleo:
I never said anything about end users.  I was talking about developers who are learning the codebase.

davecgh:
I completely agree with @jcvernaleo here.  Most developers will be learning the codebase.  You have to keep in mind that very few developers are active and remain active on a given codebase long term.  The vast majority of devs are people that are just making a few contributions here and there and aren't familiar with it.

While it is true that both using and not using keyed literals can lead to issues, not using them has been proven to be more dangerous and problematic in the wild multiple times across multiple projects.  Furthermore, consistency is far more important than almost every other concern given inconsistencies are one of the biggest things that lead to bugs, much more so than changing a struct and not spending the additional 2 minutes it takes to find all instances that use the struct and verifying if they need to be updated.

You've already conceded that the struct is "certainly not changing anytime soon", but then you argue that the code should *special case* this one thing, which creates an inconsistency, makes it harder to read for just about everything except maybe the 2 or 3 people who originally wrote it, and makes it fail lint checks, etc, all just in case it ever changes and whoever changes it doesn't do what any good programmer would do and verify that all instances of the struct are updated properly.

My issue is this is a theoretical argument over a practical one.  Making the code consistent, lint clean, and more readable for new developers have a real and practical benefit.  Given up practical benefits in exchange for special casing things to avoid using keyed literals because of a theoretical possibility of a struct changing (one that everyone already agrees is highly unlikely to change anyways) in the future doesn't make any sense to me at all.

jrick:
I'm merging this, but I am still unconvinced it's a good idea.  I am always in favor of improving readability, but data consistency is always more important than readability.

In the future we need to write a static analyzer to check for struct literals with missing fields.  That would get us the best of both worlds.  (why the hell that's not part of the go language in the first place is anyone's guess.  probably because there's no versioned package manager in the standard toolchain)

Issue #404, opened 2016-04-04
Invalid input transaction hash crashes btcwallet


gmelika:
if I create a transaction with a non-existent input transaction id crashes btcwallet:

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x58599d]

goroutine 47 [running]:
panic(0xbd4f40, 0xc8200100e0)
        /usr/lib/go/src/runtime/panic.go:464 +0x3e6
github.com/btcsuite/btcwallet/wallet.(*Wallet).SignTransaction(0xc8200bac00, 0xc8202d19e8, 0xc800000001, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/ubuntu/dev/go/src/github.com/btcsuite/btcwallet/wallet/wallet.go:1943 +0xbdd
github.com/btcsuite/btcwallet/rpc/rpcserver.(*walletServer).SignTransaction(0xc8200440f8, 0x7fd4a09f5ef0, 0xc8312c2330, 0xc8202c6050, 0x0, 0x0, 0x0)
        /home/ubuntu/dev/go/src/github.com/btcsuite/btcwallet/rpc/rpcserver/server.go:505 +0x569
github.com/btcsuite/btcwallet/rpc/walletrpc._WalletService_SignTransaction_Handler(0xc9b9e0, 0xc8200440f8, 0x7fd4a09f5ef0, 0xc8312c2330, 0xc820466280, 0x0, 0x0, 0x0, 0x0)
        /home/ubuntu/dev/go/src/github.com/btcsuite/btcwallet/rpc/walletrpc/api.pb.go:1400 +0x135
google.golang.org/grpc.(*Server).processUnaryRPC(0xc820140200, 0x7fd4a09f5d38, 0xc8311ec1b0, 0xc820102540, 0xc8312021a0, 0x123f420, 0xc8312c2300, 0x0, 0x0)
        /home/ubuntu/dev/go/src/google.golang.org/grpc/server.go:497 +0xe13
google.golang.org/grpc.(*Server).handleStream(0xc820140200, 0x7fd4a09f5d38, 0xc8311ec1b0, 0xc820102540, 0xc8312c2300)
        /home/ubuntu/dev/go/src/google.golang.org/grpc/server.go:646 +0x109d
[...]

PR #405, opened 2016-04-04
invalid transaction hashes crashes btcwallet

gmelika:
this is for issue #404

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -1936,6 +1936,10 @@ func (w *Wallet) SignTransaction(tx *wire.MsgTx, hashType txscript.SigHashType,
 			prevIndex := txIn.PreviousOutPoint.Index
 			txDetails, err := w.TxStore.TxDetails(prevHash)
 			if err != nil {
+				return nil, fmt.Errorf("Cannot query previous transaction "+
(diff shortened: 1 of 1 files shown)


jrick:
Thanks for spotting this.  You're right that TxDetails will return a nil transaction and a nil error when the transaction is not found.

Since the error case is handled by returning "not found" to the caller, could you first check that the error is not nil, and encode the error to a string to report the error, and then check if txDetails is nil after the error check?

gmelika:
separated the error messages based on your feedback.  I basically copied the error text from another place in the code that did that same check

jrick:
Minor style nitpick: can you split the string over several lines (concatenating them with +) so this fits in under 80 columns?

FWIW this is not my preferred style and I admit to breaking it in some places for practical reasons (like the rpcserver package) but here there's no real reason not to keep it similar to the surrounding code.

gmelika:
no problem.  I fixed it, and squashed the commits

jrick:
Looks like you need to run go fmt on it.

gmelika:
> meh, this looks simple enough to commit as is. what could possibly go wrong, anyway

- Anonymous developer 😄

PR #406, opened 2016-04-05
Rollback namespace transactions when bucket is not found.

jrick:
This fixes a deadlock where failed transactions due to the namespaces'
bucket being missing would cause deadlocks due to bolt's mmap rwmutex
still being read locked (and no way to unlock it, since the underlying
bolt tx was not returned on failure).


davecgh:
OK

PR #407, opened 2016-04-06
Add amount to TransactionDetails

gmelika:
it would be nice to have access to the amount value in the `outputs` transactions.  This is a backwards compatible change and will not break any existing implementations.

Also, what is the reasoning behind not including the address for `mine` transactions?


jrick:
The index field is unnecessary since all outputs from the transaction are included, in the same order, so there's a 1:1 correspondence between each output and the transaction's actual outputs.

The assumption here is that the client is able to decode the raw transaction, but any wallet details regarding that transaction that are not encoded in it have to be encoded separately in the RPC response.  This is why the index and amount fields were removed, and why the addresses were never included, because if the client needs that information, it's available in the decoded transaction.

Is there a reason why you would need that information without decoding the transaction?

jrick:
Oh, addresses were included for "not mine" outputs.  I may consider removing that field entirely and make the client encode the output scripts to addresses if it needs them.

gmelika:
i am of 2 minds here.  On the one hand I can see your point - especially if you're removing the addresses field altogether.  What was confusing was that some data that could clearly be had through parsing the tx was still included in the payload but others, like `Amount`, were being hidden on the rpc layer.

However, I feel like there is no added benefit in hiding that info when it is readily available.  To me it makes sense to include everything in the output transactions: addresses (both `mine` and not), amount, account, internal...  Just thinking about it from an api user perspective, it is a convoluted process to get any output relevant info if they're checking the `mine` outputs.  They'd have to parse the transaction, loop through the outputs looking for `mine`, then reference the corresponding output in the raw tx to find that info.

Since the output info is all there anyway, it didn't make sense to me to hide the 2 most important pieces of info (addresses, and amount) and forcing the user to parse the raw transaction.

jrick:
The point is not to hide it, it's to avoid duplicating it unnecessarily.

In the case of addresses, those are not readily available and have to be parsed out of the output scripts server side in the RPC handlers.  It can be more efficient for the client to perform that work when they need it rather than the server paying the cost for it every time.

It would be good to think about APIs for clients without the ability to parse raw transactions, but I think those should be done with separate RPC calls instead of cluttering the existing ones.

I'll open up a new issue for removing the addresses field to keep this consistent.

gmelika:
Makes sense to think about this in the context of new APIs.

PR #409, opened 2016-04-07
Update TravisCI configs.

jcvernaleo:
The vet tool moved to the Go tree as of go1.5 so it does not
need to be installed seperately.  With the x/tools repo now gone,
remove the install step for vet since it fails.


davecgh:
OK

PR #411, opened 2016-04-13
Resurrect rpchelp_test.go

jrick:
This file was accidentally removed when the gRPC server was introduced
and the legacy JSON-RPC server was moved to the legacyrpc package.


davecgh:
Nice catch!

OK

PR #412, opened 2016-04-13
Rename datadir/-b options to appdata/-A.

jrick:
This removes an inconsistency with the datadir option from btcd, which
only set the directory containing the blockchain data and indexes.

### DIFF
--- a/btcwallet.go
+++ b/btcwallet.go
@@ -61,7 +61,7 @@ func walletMain() error {
 		}()
 	}
 
-	dbDir := networkDir(cfg.DataDir, activeNet.Params)
+	dbDir := networkDir(cfg.AppDataDir, activeNet.Params)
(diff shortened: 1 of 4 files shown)


jcvernaleo:
We already have this in dcrd so I'd love it to make it to btcd.  So OK by me.

jrick:
Right, plan is to get something like this into btcd as well.

davecgh:
I'd prefer to keep datadir as an alias and a big deprecated warning to give people a chance to switch over before breaking their configs.

jrick:
Reintroduced the old flag.  The new flag receives priority when both are modified from the defaults, and a warning is logged whenever the old flag is modified from the default.

If we synced the latest go-flags from upstream, then we could also use the new hidden tag to hide the option from the help.

davecgh:
OK

PR #413, opened 2016-04-13
Lookup actual home directory when expanding ~.

jrick:
This fixes home directory expansion on Windows and OS X.  Previously
on these platforms, a leading ~ would expand to %LOCALAPPDATA% on
Windows or ~/Library/Application Support/ on OS X.

While here, add support for ~otheruser expansion to expand to
otheruser's home directory.


jrick:
Rebased over master and added the expansion to a few more config options that were missing it.  I plan on leaving that as a separate commit since it's unrelated to the changes to the expansion function itself.

davecgh:
OK

PR #414, opened 2016-04-15
Cleanup TransactionDetails RPC messages.

### LINKED ISSUE #408: Remove addresses field from TransactionDetails.Output RPC message.
See discussion in #407.

Since this would be a breaking change, the fields should be renumbered as well so they start with 1.

jrick:
Remove the addresses field from TransactionDetails.Output.  It is
assumed that the caller is able to deserialize the transaction and
encode the output scripts to addresses, so this is unnecessary server
overhead and conflicts with the current API philosophy of not
duplicating data already included in another field.

Since there is no additional data included for outputs not controlled
by the wallet, remove the `mine` specifier from the Output message and
replace it with an output index.  Only include messages for controlled
outputs, rather than creating messages for both controlled and
uncontrolled outputs.  Rename the repeated field from `outputs` to
`credits` to be consistent with the `debits` field.

Bump major API version as this is a breaking change.

Closes #408.


davecgh:
This gets a no from me.  The purpose of an RPC API is to make it easy to interface with from disparate sources.  So if I were, for example, interfacing with this via `node.js` or `Java`, or `X`, I sure as heck don't want to have to write transaction deserialization code and address parsing code too.

jrick:
I'm fine with providing those details at a later time (probably in a new message, or the caller could provide options in the request to include these additional details).  However, this API is still very much in its early stages and I am trying to keep it as minimal as possible.  This means no duplicating of data that can be found elsewhere.  The entire API at this point is coded with the assumption that the client does have the ability to understand raw transactions, with this case being the odd one out.

wallet/notifications.go
```diff
@@ -371,23 +364,13 @@ type TransactionSummaryInput struct {
 	PreviousAmount  btcutil.Amount
 }
 
-// TransactionSummaryOutput describes a transaction output of a relevant
-// transaction.  When the transaction is authored by this wallet, all
-// transaction outputs are considered relevant.  The Mine field describes
-// whether outputs to these authored transactions pay back to the wallet
-// (e.g. change) or create an uncontrolled output.  For convenience, the
-// addresses (if any) of an uncontrolled output are included.
+// TransactionSummaryOutput describes wallet properties of a a transaction
```
davecgh: a is repeated.

davecgh:
I understand the reasoning for the change when viewing it from the view of Paymetheus.

However, the issue I see with this, and as you noted, the API in general, is it very much feels like it is effectively being designed specifically for `Paymetheus` as opposed for the purposes of being a full-featured rich client API.

I personally am much more interested in having the latter versus something that is finely tuned and really only works well for Paymetheus.

Anyways, as far as code, other than the typo I pointed out and has already been fixed, it looks fine.

jrick:
I agree that the API should be extended to support more kinds of applications, but it is difficult to do this without suffering from feature creep and making the entire API mediocre for everyone.  At the current time I want avoid these issues by keeping the API surface limited.  The API can be gradually changed later, either by adding new methods and/or services, or adding additional fields to existing message types.

davecgh:
OK

PR #416, opened 2016-04-18
Ignore config in current directory by default.

### LINKED ISSUE #415: Ignore config files in current directory unless explicitly requested
In the current code, configuration files (`btcwallet.conf`) in the current directory the process is run from are given precedence over the default configuration file.  This feature may introduce a security and privacy vulnerability if a malicious actor is able to place a configuration file in this directory which, e.g., runs the experimental gRPC server on a non-standard port and listens for all wallet notifications, or is able to sign a transaction in the time that the wallet was momentarily unlocked by the JSON-RPC server.

Even without the security issues, this behavior does not match btcd and could be considered surprising, so it should be changed regardless for consistency.

The config file in the current directory can be specified by using relative paths instead: `btcwallet -C ./btcwallet.conf`

jrick:
Fixes #415.

### DIFF
--- a/config.go
+++ b/config.go
@@ -222,21 +222,11 @@ func loadConfig() (*config, []string, error) {
 		DataDir:                defaultAppDataDir,
 	}
 
-	// A config file in the current directory takes precedence.
-	exists, err := cfgutil.FileExists(defaultConfigFilename)
-	if err != nil {
-		fmt.Fprintln(os.Stderr, err)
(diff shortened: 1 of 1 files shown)


jcvernaleo:
I can't test this since it isn't caught up with grpc but I agree with the idea.

jrick:
Rebased in case you want to test.

jcvernaleo:
OK

PR #417, opened 2016-04-18
Do not include zero value change outputs.

jrick:
Due to the way dust is calculated, if the transaction relay fee is
zero, then a zero output amount is not considered dust.  As the
transaction authoring code used this dust check to determine whether a
change output can be included or not, it could create unnecessary
change outputs which return no value back to the wallet.  Prevent this
by including an explicit check for zero values.


alexlyp:
Changes look good.  utACK

PR #418, opened 2016-04-19
Regenerate walletrpc package.

jrick:
Fixes build when using newer versions of the grpc packages.


jcvernaleo:
Just like in dcrwallet, not much choice on this one.  Really needs to go in.
OK

PR #419, opened 2016-04-19
Convert project to glide.

### LINKED ISSUE #399: Vendor gRPC+deps
Currently, `go get` pulls the master branches for these repos and installs them to GOPATH.  This has several issues, including the inability to run reproducible builds and vetting of changes before merging them to the project.

Possible solutions:

1. Vendor deps to btcsuite repos and rewrite import paths
2. Clone and commit these dependencies to the vendor directory
3. Use a script to clone deps to the vendor directory, but do not commit them
4. Use a project-based tool (e.g. `glide`) to manage vendored dependencies

All of these solutions come with various upsides and downsides.

Solution 1 allows us to maintain control of what changes we merge from upstream, but it requires rewriting import paths (especially bad since `protoc` generates source files that reference the canonical import paths).  It also does not solve the issue of reproducible builds as it still relies on GOPATH and the checked out version may not match what is actually required for some particular btcwallet version.

Solution 2 is the most reliable.  It provides reproducible builds since it does not rely on a GOPATH-global version of the code and that the code never changes unless explicitly checked into the project.  Out of all the solutions that use the vendor directory, this is the only one that does not require any extra tools besides the `go` tool (the project remains `go get`able).  The only major downsides to this solution is that it bloats the btcwallet repo unnecessarily and doesn't do much in terms of helping developers manage newer dependency versions.

Solution 3 is the most minimal, but requires additional bootstrapping before the project can be built and requires introducing operating system-specific build scripts to the repo.  Like solution 2, this solution also provides little help in managing updates to newer versions of the dependencies.

Solution 4 is similar to solution 3 but uses a tool designed specifically for this purpose.  Newer versions of the dependencies can be easily fetched and locked to specific commits by committing a lock file to the repo, rather than a build script with commit hashes.  Downsides include the reliance on an external tool to fetch and update the project and the inability to reliably use `go get` to initially download a repo (which might or might not work depending on whether the master branches are compatible).  This is currently my preferred solution to the vendoring problem.

jrick:
Fixes #399.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,15 +66,72 @@ https://github.com/btcsuite/btcwallet/releases
 
 ### Windows/Linux/BSD/POSIX - Build from source
 
-- If necessary, install Go according to the installation instructions
-  here: http://golang.org/doc/install.  It is recommended to add
-  `$GOPATH/bin` to your `PATH` at this point.
+Building or updating from source requires the following build dependencies:
 
-- Run the following commands to obtain and install btcd, btcwallet
-  and all dependencies:
+- **Go 1.5 or 1.6**
+
+  Installation instructions can be found here: http://golang.org/doc/install.
+  It is recommended to add `$GOPATH/bin` to your `PATH` at this point.
+
+- **Glide**
+
+  Glide is used to manage project dependencies and provide reproducible builds.
+  To install:
+
+  `go get -u github.com/Masterminds/glide`
+
+Unfortunately, the use of `glide` prevents a handy tool such as `go get` from
+automatically downloading, building, and installing the source in a single
+command.  Instead, the latest project and dependency sources must be first
+obtained manually with `git` and `glide`, and then `go` is used to build and
+install the project.
+
+**Getting the source**:
+
+For a first time installation, the project and dependency sources can be
+obtained manually with `git` and `glide` (create directories as needed):
+
 ```
-go get -u -v github.com/btcsuite/btcd/...
-go get -u -v github.com/btcsuite/btcwallet/...
+git clone https://github.com/btcsuite/btcwallet $GOPATH/src/github.com/btcsuite/btcwallet
+cd $GOPATH/src/github.com/btcsuite/btcwallet
+glide install
+```
+
+To update an existing source tree, pull the latest changes and install the
+matching dependencies:
+
+```
+cd $GOPATH/src/github.com/btcsuite/btcwallet
+git pull
+glide install
+```
+
+**Building/Installing**:
+
+The `go` tool is used to build or install (to `GOPATH`) the project.  Some
+example build instructions are provided below (all must run from the `btcwallet`
+project directory) in a `sh`-compatible shell on Unix, or PowerShell on Windows
+(`$()` subshells are not supported by `cmd`).
+
+To build and install `btcwallet` and all helper commands (in the `cmd`
+directory) to `$GOPATH/bin/`, as well as installing all compiled non-vendored
+packages to `$GOPATH/pkg/` (**use this if you are unsure which command to
+run**):
+
+```
+go install $(glide novendor)
+```
+
+To build a `btcwallet` executable and install it to `$GOPATH/bin/`:
+
+```
+go install
+```
+
+To build a `btcwallet` executable and place it in the current directory:
+
+```
+go build
 ```
 
 ## Getting Started
--- a/goclean.sh
+++ b/goclean.sh
@@ -1,38 +1,36 @@
 #!/bin/bash
 # The script does automatic checking on a Go package and its sub-packages, including:
-# 1. gofmt         (http://golang.org/cmd/gofmt/)
-# 2. goimports     (https://github.com/bradfitz/goimports)
-# 3. golint        (https://github.com/golang/lint)
-# 4. go vet        (http://golang.org/cmd/vet)
-# 5. race detector (http://blog.golang.org/race-detector)
-# 6. test coverage (http://blog.golang.org/cover)
+# 1. go fmt        (http://golang.org/cmd/gofmt/)
+# 2. golint        (https://github.com/golang/lint)
+# 3. go vet        (http://golang.org/cmd/vet)
+# 4. race detector (http://blog.golang.org/race-detector)
 
 set -ex
 
 # Automatic checks
-test -z "$(gofmt -l -w . | tee /dev/stderr)"
-test -z "$(goimports -l -w . | tee /dev/stderr)"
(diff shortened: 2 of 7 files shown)


jrick:
Once this is in and btcd is converted to use glide as well, I would like to switch away from our btcsuite package paths back to the upstream ones.  We can still maintain forked repos as a backup in case they disappear, but there is no reason to keep rewriting the import paths when we maintain control over the versions in the vendor directory.

This should be done after btcd also is converted to glide since we don't want to import the same package multiple times with different names, as some of these dependencies are shared in btcd and btcwallet code.

jcvernaleo:
Could you maybe point out in the readme that go commands that do not recurse (go build/install without the ./...) work as normal after the glide install so there is no need for the glide novendor in those cases.
More specifically, when I cross compile, I cannot use glide novendor.  This does not work properly:
env GOOS=openbsd GOARCH=amd64 go build $(glide novendor)
while this is fine:
env GOOS=openbsd GOARCH=amd64 go build

jcvernaleo:
Other than that doc request this looks good to me and works as expected on Linux and BItrig.

jrick:
sure, i'll update the readme to include both instructions

Since those instructions are the same for both first time installs and updates, i'll try to merge them into a single place too.

jcvernaleo:
Do I need to test this on OSX as well?

jrick:
Unlikely to be an issue but it wouldn't hurt.

jcvernaleo:
I tested and as expected, OSX was fine.  And new instructions look good.  Not sure if anyone else wanted to chime in but OK from me.

README.md
```diff
@@ -66,15 +66,71 @@ https://github.com/btcsuite/btcwallet/releases
...
+Building or updating from source requires the following build dependencies:
+
+- **Go 1.5 or 1.6**
+
+  Installation instructions can be found here: http://golang.org/doc/install.
+  It is recommended to add `$GOPATH/bin` to your `PATH` at this point.
+
+- **Glide**
+
+  Glide is used to manage project dependencies and provide reproducable builds.
```
davecgh: reproducible

README.md
```diff
@@ -66,15 +66,71 @@ https://github.com/btcsuite/btcwallet/releases
...
+
+**Getting the source**:
+
+For a first time installation, the project and dependency sources can be
+obtained manually with `git` and `glide` (create directories as needed):
+
+```
+git clone https://github.com/btcsuite/btcwallet $GOPATH/src/github.com/btcsuite/btcwallet
+cd $GOPATH/src/github.com/btcsuite/btcwallet
+glide install
```
davecgh: This will only install btcwallet and none of the subcommands.

README.md
```diff
@@ -66,15 +66,71 @@ https://github.com/btcsuite/btcwallet/releases
...
+To build a `btcwallet` executable and place it in the current directory:
+
+```
+go build
+```
+
+To build a `btcwallet` executable and install it to `$GOPATH/bin/btcwallet`:
+
+```
+go get
```
davecgh: `go install`

goclean.sh
```diff
@@ -1,38 +1,36 @@
...
-do
-if ls $dir/*.go &> /dev/null; then
-  go test -covermode=count -coverprofile=$dir/profile.tmp $dir
-  if [ -f $dir/profile.tmp ]; then
-    cat $dir/profile.tmp | tail -n +2 >> profile.cov
-    rm $dir/profile.tmp
-  fi
-fi
-done
+#for dir in $(find . -maxdepth 10 -not -path '.' -not -path './.git*' \
```
davecgh: I'd prefer this be fixed instead of commented out.  It allows the test coverage to be available in the build logs.  I believe all you should need to do is add a `-not -path './vendor' -type d`

README.md
jrick: `go get` does this too.

README.md
jrick: The instructions below provide various build commands, one of which includes the installing the helper commands.  This section is only about getting the latest sources for the project and vendored deps.

goclean.sh
jrick: It was commented out because it's not used at all (with coveralls.io, I mean).  I'd rather the CI not spend time running every test a second time just to create a number that's not used anywhere.

README.md
jrick: Well, `get` will download missing dependencies while `install` does not, and (if glide was run) there should be none missing, so there won't not be any difference.  This comes down to whether we want to fetch missing dependencies over the network or not.

davecgh: Correct.  That is why I was suggesting `install`, because the entire point of switching over to glide is to ensure reproducible builds.  I would rather have it fail to build/install if there are any dependencies missing which would mean that the lockfile isn't specifying them properly.

jrick: Ok, that makes sense.  I gravitated towards `get` because i never use `install` myself.

goclean.sh
davecgh: Alright.  I personally like having the actually coverage displayed in the build logs, but it's not a big deal if it's not there.

README.md
davecgh: This is confusing in my opinion.  The first thing I expect to see is a "quick start" that is the most common case which is getting the source (including all deps), building, and installing `btcwallet` _and_ all of its utilities.  As it standing, the user will have to do this and then know which section to choose.  That is not very user friendly.

jrick: I bolded that case just for them :)

> To build and install `btcwallet` and all helper commands (in the `cmd` directory) to `$GOPATH/bin/`, as well as installing all compiled non-vendored packages to `$GOPATH/pkg/` (**use this if you are unsure which command to run**):

jrick: Would you want those examples reordered so that this is the first one?

davecgh: Yes, that is exactly what I was going to suggest along with a pseudo heading note.  Perhaps:

------

To build and install `btcwallet` and all helper commands (in the `cmd`
directory) to `$GOPATH/bin/`, as well as installing all compiled non-vendored
packages to `$GOPATH/pkg/` (**use this if you are unsure which command to
run**):

```
go install $(glide novendor)
```

**NOTE**: The following are alternatives and do not need to be
executed unless you specifically want the different behavior
described.

* To build a `btcwallet` executable and place it in the current directory:

```
go build
```
**OR**

* To build a `btcwallet` executable and install it to `$GOPATH/bin/btcwallet`:

```
go install
```

-----

Another issue I see is that the `$()` syntax doesn't work under Windows `cmd`.  This should call that out, in my opinion.

davecgh:
It appears that everything has been addressed.

OK

PR #422, opened 2016-04-20
Make default config relative to appdata directory.

### LINKED ISSUE #421: Modifying the appdata flag should move the default config
This should only occur if the config itself is also not specified with a command line flag.

Unfortunately these checks only work by comparing the value after parsing against the old value, so even if the old value was explicitly passed as a flag, we would have no idea.  I consider this a bug in the flags package.

jrick:
This modifies the default configuration file to be relative to the
application data directory (configuratble with --appdata).  If there
is no configuration file in this directory, then no additional
configuration options are applied (it does not fallback to
~/.btcwallet/btcwallet.conf).

Fixes #421.


davecgh:
OK

PR #423, opened 2016-04-22
Avoid calling getrawtransaction from signrawtransaction.

### LINKED ISSUE #410: signrawtransaction should not call getrawtransaction
All relevant transactions should already be saved by the wallet so it is unnecessary to look these up from the btcd instance.

This needs to be fixed soon so that wallets running on btcd instances without the transaction index can still be used to sign raw transactions over the legacy RPC API.

jrick:
Replace all calls to getrawtransaction with gettxout.

As btcd no longer enables the transaction index by default,
getrawtransaction can no longer be depended on.  gettxout is safe to
use since it queries the utxo set.  This also means that it will
continue to work even when pruning is enabled.

This should be further improved in the future to not look up previous
output scripts over btcd rpc when they are already saved by the
wallet.

Fixes #410.

### DIFF
--- a/rpc/legacyrpc/methods.go
+++ b/rpc/legacyrpc/methods.go
@@ -1622,13 +1622,6 @@ func SignMessage(icmd interface{}, w *wallet.Wallet) (interface{}, error) {
 	return base64.StdEncoding.EncodeToString(sigbytes), nil
 }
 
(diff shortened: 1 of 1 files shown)


davecgh:
OK

PR #424, opened 2016-04-25
[WIP] Add support for new Segregated Witness standard output types

Roasbeef:
NOTE: This PR depends on both btcsuite/btcd#656 and btcsuite/btcutil#75

This PR introduces two new address types to the `waddrmgr`, and the wallet 
sub-package itself. The first address type is the native `p2wkh` (
pay-to-witness-key-hash) output type introduced as part of BIP0141 and 
the segwit soft-fork. The second address type is a `p2wkh` output nested 
_within_ a regular `p2sh` output. This second address allows older wallets
which are not yet aware of the new segwit output types to transparently 
pay to a wallet which does support them. Additionally, using this 
nested p2wkh output the wallet gains both the space+transaction
fee savings, as well as the nuisance malleability fixes.

Both address types have been implemented as special cases of the
`ManagedPubKeyAddress` since they share several traits, only
differentiating in the signing mechanism needed, and the concrete
implementation of `btcutil.Address` returned by the address.

Two new `addressType` constants have been added to `waddrmgr`’s db in
order to properly serialize and deserialize the new address types.

When spending a segwit output, the wallet also needs the input value of
the previous output script. Therefore when selecting outputs the input
value is now returned. Additionally when validating newly signed
outputs the input value as also passed into `txscript.Engine`. 

Additionally, the wallet is now able to properly spend nested and 
normal `p2wkh` outputs under its control. For regular `p2wkh` outputs, 
spending simply involves presenting the original pub key, and signature 
as the witness data. For nested `p2wkh` outputs, in addition to the above, 
the version zero witness `p2wkh` witness program is placed in the sigScript 
in order to allow clients who are aware of BIP 16 to validate the witness 
spend.

NOTE: This PR is still a WIP progress, the following work items need to
be reconciled before the tag is removed: 
- [x] Update the transaction fee and size estimates to account for `p2wkh` and nested 
  `p2wkh` outputs. 
- [ ] Both (?) RPC servers need to update to compile under the API changes. 
- [ ] Investigate adding a lookahead to the wallet which accounts for all three address types?
- [ ] Address lingering `TODO(roasbeef):` comments.

### DIFF
--- a/waddrmgr/address.go
+++ b/waddrmgr/address.go
@@ -10,10 +10,46 @@ import (
 	"sync"
 
 	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcutil"
 	"github.com/btcsuite/btcutil/hdkeychain"
 	"github.com/btcsuite/btcwallet/internal/zero"
 )
+
+// AddressType represents the various address types waddrmgr is currently able
+// to generate, and maintain.
+type AddressType uint8
+
+const (
+	// PubKeyHash is a regular p2pkh address.
+	PubKeyHash AddressType = 1 << iota
+
+	// NestedWitnessPubKey represents a p2wkh output nested within a p2sh
+	// output. Using this address type, the wallet can receive funds from
+	// other wallet's which don't yet recognize the new segwit standard
+	// output types. Receiving funds to this address maintains the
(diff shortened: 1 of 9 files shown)


Roasbeef:
Rebased to latest master. Working on finishing up those lingering TODO's.

Roasbeef:

PR #430, opened 2016-04-27
Set gRPC cipher suites sooner in node.js example.

### LINKED ISSUE #429: Example node.js client may set cipher suites environment var too late
On Windows, it appears that the GRPC_SSL_CIPHER_SUITES variable must be set before loading the grpc library with require.

jrick:
While here, also rename OpenSSL to BoringSSL since gRPC Core has
switched to that.

Fixes #429.


davecgh:
OK

PR #431, opened 2016-04-28
Remove code intended for Go 1.4 and older.

jrick:
This change only removes optimizations for older Go releases that we
no longer support in the internal/zero package.  This package should
still continue to build on older releases.


davecgh:
OK

PR #432, opened 2016-05-04
Fix credit slice indexing for transaction notifications.

jrick:
Previously, this would always check a transaction output index against
the 0th credit's index.


davecgh:
Nice catch.

OK

jrick:
Going to also bump the patch number for the grpc api.

PR #433, opened 2016-05-04
Use -b for deprecated --datadir short option.

jrick:
When 29f31725ee57fcf4a1118b420f8c115c2742da0a deprecated the datadir
option, replacing it with --appdata/-A, the short --datadir option was
unintentinally changed from -b to -D.  It was previously changed to -b
in fda2e14b99bba7e8ff4cf0f7434e1f5e8b130205 for consistency with
btcd's --datadir short option.

This change reverts the unintentional switch back to -D.


jcvernaleo:
OK

PR #434, opened 2016-05-05
Handle flags explicitly set to defaults.

jrick:
This prevents treating a flag that was explicitly set to the default
as unchanged, since the explicit set is recorded in the new
*cfgutil.ExplicitString flag type.

### DIFF
--- a/btcwallet.go
+++ b/btcwallet.go
@@ -61,7 +61,7 @@ func walletMain() error {
 		}()
 	}
 
-	dbDir := networkDir(cfg.AppDataDir, activeNet.Params)
+	dbDir := networkDir(cfg.AppDataDir.Value, activeNet.Params)
 	loader := wallet.NewLoader(activeNet.Params, dbDir)
 
 	// Create and start HTTP server to serve wallet client connections.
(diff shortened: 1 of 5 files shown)


jcvernaleo:
This looks good and I'd like us to get it in dcr* stuff too.  My only concern is putting things in the package internal/cfgutil/explicitflags.go
since we probably could use this in some of the other tools and I would hate to have to duplicate it in more than one place.  That being said, I'm not sure if there is a reasonable place for it.  btcutil maybe?

jrick:
That internal package can be used by any of the main packages in this repo.  It's only btcd/dcrd, and the extra stake-related Decred tools in other repos, that would be unable to import it.

I think btcutil is overused already as is.  Not sure where else this package could go, other than a new repo (with glide, having many repos isn't really an issue anymore).

jcvernaleo:
Fair enough.  I was mostly thinking out loud about if we might need it elsewhere (btcd/btcctl mainly).  I agree that btcutil would be an odd place for it (although I don't share you feeling about that having too much in it).  So
OK
as is.

PR #436, opened 2016-05-05
Unexport private methods from legacyrpc package.

### LINKED ISSUE #435: legacyrpc RPC handler functions should be unexported
These have no business being called by other packages (their implementations are internal to the RPC server implementation) and their signatures are free to change at any time (as has happened before).

Unexporting these would also help consistency with the gRPC server stub implementations, which are also unexported (actually, they are exported methods on an unexported type in order to implement the stub interface, but similar idea).

jrick:
This unexports every RPC handler function, as well as the
Server.PostClientRPC method which was only called by this package.

Fixes #435.

### DIFF
--- a/rpc/legacyrpc/methods.go
+++ b/rpc/legacyrpc/methods.go
@@ -78,84 +78,84 @@ var rpcHandlers = map[string]struct {
 	noHelp bool
 }{
 	// Reference implementation wallet methods (implemented)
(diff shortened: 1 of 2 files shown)


davecgh:
Awesome.  This has bothered me for quite some time now.  I'm extremely happy to see this PR.

PR #437, opened 2016-05-05
btcwallet: handle signal SIGTERM

dajohi:
When an OS reboots or shuts down, it sends all processes SIGTERM before
sending SIGKILL.  This allows btcwallet to do a proper shutdown which
most importantly closes the databases.

### DIFF
--- a/signal.go
+++ b/signal.go
@@ -22,6 +22,10 @@ var interruptHandlersDone = make(chan struct{})
 
 var simulateInterruptChannel = make(chan struct{}, 1)
 
+// signals defines the signals that are handled to do a clean shutdown.
+// Conditional compilation is used to also include SIGTERM on Unix.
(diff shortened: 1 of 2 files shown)


signal.go
```diff
@@ -22,6 +22,10 @@ var interruptHandlersDone = make(chan struct{})
 
 var simulateInterruptChannel = make(chan struct{}, 1)
 
+// signals defines the default signals to catch in order to do a proper
```
jrick: I know it says the same thing in btcd now but could we remove "proper" (that word can mean anything, really) and just say that all of these signals are handled with clean shutdown, and that this also includes SIGINT on Unix with conditional compilation?

dajohi:
updated to address comment.

jrick:
ok

PR #438, opened 2016-05-05
Include vendored packages when recursively installing.

jrick:
This improves incremental compilation times since these packages are
not recompiled unless changed.  The README and CI scripts have been
updated for these new recommended instructions.

The $(glide novendor) trick is now only used in order to run tests in
the btcwallet project itself, not tests of its vendored dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ before_install:
 install:
   - go get -v github.com/Masterminds/glide
   - glide install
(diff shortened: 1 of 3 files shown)


jrick:
Closing.  `go get $(glide novendor)` already installs vendored packages into `$GOPATH/bin/` and this change only causes other build issues.

PR #439, opened 2016-05-05
Fix description for recommended install instructions.

jrick:
The description had previously said that non-vendored packages were
not installed to $GOPATH/bin, but this is incorrect.  Remove this part
of the description so it includes all packages.


davecgh:
OK

PR #440, opened 2016-05-06
Enable vendor experiment for Go 1.5 on Travis-CI.

jrick:
Previously, the GO15VENDOREXPERIMENT environment variable was unset,
leaving travis to only use the vendored packages during builds for Go
1.6 which enabled the experiment by default.  Export the variable so
it also correctly tests under Go 1.5.


jcvernaleo:
Looks good.
OK

PR #442, opened 2016-05-06
Don't use go install $(glide novendor) in examples.

jrick:
This will build all (non-vendored) packages recursively, even if they
are not used in any of the binaries (e.g. votingpool).  Additionally,
since a subshell is no longer used, the example commands can be run in
a Windows command prompt.  I don't encourage the use cmd.exe over
alternatives such as PowerShell, but there's no reason this command
could not be run there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,5 @@ install:
   - go get -v github.com/davecgh/go-spew/spew
 script:
   - export PATH=$PATH:$HOME/gopath/bin
+  - export GO15VENDOREXPERIMENT=1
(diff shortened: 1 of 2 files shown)


jcvernaleo:
From a testing perspective I like building everything, even things that aren't used (that's what eventually convinced me to kill votingpool in dcrwallet when I got sick of fixing it) but I agree that it isn't a useful thing for most people so I think this makes sense.
OK

jrick:
Tests (as run by travis and appveyor) still build and test everything with $(glide novendor).

PR #443, opened 2016-05-06
Switch from forked to upstream flags package.

jrick:
In https://github.com/btcsuite/btcwallet/pull/419#issuecomment-212482492
I mentioned that btcwallet should wait until btcd also switches back
to upstream package paths so that duplicate packages are not included
in and bloat the btcwallet binary.  Even while btcd has not yet
switched to the upstream flags package, this change can still be made
now since it is only used by btcd's main package, which is not
included in btcwallet.

### DIFF
--- a/cmd/dropwtxmgr/main.go
+++ b/cmd/dropwtxmgr/main.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2015 The btcsuite developers
+// Copyright (c) 2015-2016 The btcsuite developers
 // Use of this source code is governed by an ISC
 // license that can be found in the LICENSE file.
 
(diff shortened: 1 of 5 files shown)


davecgh:
I don't think this should be updated to the latest upstream version without having properly reviewed and vetted all of the changes in between the latest version in the btcsuite repos and the latest upstream version.

This is just using the latest upstream commit instead of commit `167953...` which is where the btcsuite fork is properly vetted to.

jrick:
I've gone through the changes and didn't see anything suspect.  I'm also interested in the updated code for the 'hidden' tag so we can hide deprecated options such as `--datadir`.  But since updating the package and switching the package path are two different things, we can separate them out and I'll open a new PR with the update.

davecgh:
Glad to hear.  That wasn't very obvious from the PR or the description so I was concerned about introducing new unvetted code.  I like the idea of splitting them up as that will make it very explicit.  I see you've updated the PR to use the same upstream version as the forked version.

OK

PR #448, opened 2016-06-16
Fix logic for CreateTemp needing either appdata or datadir

alexlyp:


jrick:
OK

Travis failure is unrelated, appears that it is not using the vendor directory on the 1.5 build.

PR #450, opened 2016-08-07
Update for recent chainhash-related API changes.

davecgh:
**This PR requires btcsuite/btcd#729 and btcsuite/btcutil#78.**

This updates all code to make use of the new `chainhash` package since the old `wire.ShaHash` type and related functions have been removed in favor of the abstracted package.

Also, while here, rename all variables that included `sha` in their name to include `hash` instead.

Finally, update `glide.lock` to use the required version of `btcd`, `btcutil`, and `btcrpcclient`.


jcvernaleo:
OK

marcopeereboom:
ok

jrick:
ok

PR #451, opened 2016-08-10
Update for recent chaincfg API changes.

davecgh:
**This pull request requires btcsuite/btcd#732** 

Since the coinbase maturity is now allowed to be defined per chain and the old `blockchain.CoinbaseMaturity` constant has been removed, this updates the code accordingly.

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,10 +1,10 @@
 hash: 0cff4a723566a3e33910785c104bddde49d1e0ba2d51f3833599a423140f5163
-updated: 2016-08-08T14:07:48.6427885-05:00
+updated: 2016-08-12T09:43:45.22287-05:00
 imports:
 - name: github.com/boltdb/bolt
   version: 831b652a7f8dbefaf94da0eb66abd46c0c4bcf23
 - name: github.com/btcsuite/btcd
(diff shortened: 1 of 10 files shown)


glide.lock
```diff
@@ -4,7 +4,7 @@ imports:
 - name: github.com/boltdb/bolt
   version: 831b652a7f8dbefaf94da0eb66abd46c0c4bcf23
 - name: github.com/btcsuite/btcd
-  version: bd4e64d1d43bad445dd8e6577907c0c265cd83c2
+  version: a7b35d9f9e24f8cebf3dc30cf083668700690095
```
jrick: Looks like the lockfile was manually edited, since the "updated" time at the top wasn't bumped. tsk tsk

davecgh: Yeah, `glide up` tries to update all kinds of other stuff (adding entries that don't belong), so I just did it manually.

jrick: What I've done is let it update everything but then `git add -p` the deps I actually want updated.

davecgh: Ok, updated via `glide up` and ignored the other stuff.

jrick:
ok

PR #454, opened 2016-08-13
Add options --nopass and --seed.

DanielKrawisz:
Both of these options are used with --create, and they are intended
to make it easier to manage wallet creation from the command line.

--nopass means that the user is not prompted for a password to
encrypt the wallet (a default password is used instead).

--seed enables the user to provide the seed to the wallet from the
command line.

### DIFF
--- a/btcwallet.go
+++ b/btcwallet.go
@@ -86,7 +86,7 @@ func walletMain() error {
 	if !cfg.NoInitialLoad {
 		// Load the wallet database.  It must have been created already
 		// or this will return an appropriate error.
-		_, err = loader.OpenExistingWallet([]byte(cfg.WalletPass), true)
+		_, err = loader.OpenExistingWallet([]byte(cfg.WalletPass), !cfg.NoPass)
 		if err != nil {
 			log.Error(err)
 			return err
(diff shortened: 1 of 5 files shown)


jrick:
I don't want to accept this one.  Secrets (such as the seed or passphrase) should never be exposed as process arguments because they can be read by other processes.  Take this powershell script for example, that waits until a user starts a btcwallet process using this patch and then prints all of the process arguments:

```PowerShell
$timespan = new-object System.Timespan(0, 0, 1)
$query = new-object System.Management.WqlEventQuery(
	"__InstanceCreationEvent",
	$timespan,
	"TargetInstance ISA 'Win32_Process' AND TargetInstance.Name='btcwallet.exe'")
$watcher = new-object System.Management.ManagementEventWatcher($query)

$event = $watcher.WaitForNextEvent()
$event.TargetInstance.CommandLine
```

```
PS C:\Users\jrick> & .\walletwatcher.ps1
"C:\Users\jrick\Go\bin\btcwallet.exe" --testnet --create --seed 6a213e9bc477421dd19ba904f03db16c8798f2f68e566073c6865cac155f602f --nopass
```

The same basic problem exists on Unix, where process arguments are visible using tools like top and ps.

justusranvier:
Are you willing to accept any solution for creating a wallet without keyboard interaction, or will that always be an excluded use case for btcwallet?

jrick:
Other forms of input (stdin, or any other pipe) are fine.

PR #456, opened 2016-09-09
Update project dependencies.

jrick:
This updates both btcsuite and external dependencies to their latest
versions.  In particular, gRPC was updated to version 1.0.1 and bolt
to 1.3.0.

The walletrpc package needed to be regenerated for the gRPC update.


bclermont:
👍

PR #457, opened 2016-10-18
header-based-nonces

jadeblaquiere:
Move to header-based nonces for ciphrtxt network - remove references to bitcoin networks (as the block structure is now different)


jadeblaquiere:
don't merge this... intended for jadeblaquiere/ctcwallet

PR #458, opened 2016-10-27
Update for wire.NewMsgTx API change.

davecgh:
**Requires PRs btcsuite/btcutil#82 and btcsuite/btcd#817**

Also, bump glide.lock to ensure the appropriate deps are pulled in.

**NOTE: I modified `glide.yaml` and `glide.lock` temporarily until the upstream deps are in so this PR builds in the mean time.  Those will need to be reverted before this is merged**

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,10 +1,10 @@
 hash: 5efa5cef5495a0b5256f45ecb2456ea21101934ed56e780a46b1c16e30177ebc
-updated: 2016-10-27T11:04:37.4161268-04:00
+updated: 2016-10-27T11:54:19.492811-05:00
 imports:
 - name: github.com/boltdb/bolt
   version: 583e8937c61f1af6513608ccc75c97b6abdf4ff9
 - name: github.com/btcsuite/btcd
-  version: e3eeb4a34a04c8ec40298fee9f9d8fe2ce0c13ca
+  version: f6ad7eb2c963151c71de2bd37d04d9af644a891d
   subpackages:
   - blockchain
   - btcec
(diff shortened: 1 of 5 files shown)


glide.lock
```diff
@@ -44,36 +44,38 @@ imports:
...
+  - lex/httplex
   - trace
 - name: golang.org/x/sys
-  version: f64b50fbea64174967a8882830d621a18ee1548e
+  version: c200b10b5d5e122be351b67af224adc6128af5bf
   subpackages:
   - unix
 - name: google.golang.org/grpc
-  version: dd828651e45229541896bc41cd9cf2f89ac7002a
+  version: 396f8ba2a6b9039070f4ff97561f5e408828fccd
```
jrick: Looks like this pulled master instead of the released version.  Can you update the glide.yaml to set the version to `^1.0.0` and update again?

jrick: No, this is referring to the grpc repo.  It should be using ^1.0.0

glide.lock
```diff
@@ -1,10 +1,12 @@
...
-updated: 2016-10-27T11:04:37.4161268-04:00
+hash: 000de32dd5d8955199377a7216e94a8e75779e0422453d04331c6b8c1c566fe8
+updated: 2016-10-27T11:54:19.492811-05:00
 imports:
 - name: github.com/boltdb/bolt
   version: 583e8937c61f1af6513608ccc75c97b6abdf4ff9
 - name: github.com/btcsuite/btcd
-  version: e3eeb4a34a04c8ec40298fee9f9d8fe2ce0c13ca
+  version: ec78ac40328fa5bd43bddb6cd08d2b5eb7896749
+  repo: https://github.com/davecgh/btcd
```
jrick: ok after deps are in

PR #462, opened 2016-12-07
Remove -v from go test on travis.

jrick:
This just clutters the logs and makes it more difficult to determine
what the actual failing test was. For tests that do fail, all logs
from the test are already outputted.


jrick:
I opened https://github.com/btcsuite/btcwallet/issues/463 to track the race from travis.  This PR is unrelated to that issue.

PR #464, opened 2016-12-13
Minor fixes to Manager.Close

vds:


vds:
Minor fixes, checking if the the manager was already closed and removing unused error return value.

waddrmgr/manager.go
```diff
@@ -361,7 +361,11 @@ func (m *Manager) zeroSensitivePublicData() {
 // Close cleanly shuts down the manager.  It makes a best try effort to remove
 // and zero all private key and sensitive public key material associated with
 // the address manager from memory.
-func (m *Manager) Close() error {
+func (m *Manager) Close() {
+	if m.closed == true {
```
jrick: this must happen with the mutex held.

jrick: also you can just check `if m.closed {` without the `== true`.

PR #467, opened 2017-01-10
Fix reported spendable balances from coinbase outputs.

jrick:
Previously, this would not increment the spendable balance for matured
coinbase outputs and would only increment the immature balance if the
output was still immature.


davecgh:
Information from gRPC prior to this patch:

> Immature:  4950 BTC
> Spendable:  0 BTC
> Total:  5000 BTC

After:

> Immature:  4950 BTC
> Spendable:  50 BTC
> Total:  5000 BTC

PR #468, opened 2017-01-11
Drop fastsha256 in favor of crypto/sha256

dajohi:


glide.yaml
```diff
@@ -25,16 +27,19 @@ import:
...
   subpackages:
   - proto
+- package: github.com/jessevdk/go-flags
+  version: 1679536dcc895411a9f5848d9a0250be7856448c
 - package: golang.org/x/net
   subpackages:
   - context
 - package: google.golang.org/grpc
-  version: ^1.0.0
+  version: 1.0.3
```
jrick: This is fine temporarily.  I'll update grpc myself and regenerate the walletrpc package since I have all the tooling already installed.

PR #469, opened 2017-01-17
Improve wallet atomicity.

jrick:
This changes the database access APIs and each of the "manager"
packages (waddrmgr/wtxmgr) so that transactions are opened (only)
by the wallet package and the namespace buckets that each manager
expects to operate on are passed in as parameters.

This helps improve the atomicity situation as it means that many
calls to these APIs can be grouped together into a single
database transaction.

This change does not attempt to completely fix the "half-processed"
block problem.  Mined transactions are still added to the wallet
database under their own database transaction as this is how they are
notified by the consensus JSON-RPC server (as loose transactions,
without the rest of the block that contains them). It will make
updating to a fixed notification model significantly easier, as the
same "manager" APIs can still be used, but grouped into a single
atomic transaction.


jrick:
I'll get back to updating several of the btcwallet tests that fail to compile now with the new db apis (because these tests were either removed or ignored in dcrwallet...).  For now I just wanted to get btcwallet itself compiling and to finish committing the cherry-pick.

toddfries:
I updated to this and suddenly my 'listaccounts' is showing the 'getbalance' result for all accounts ..

jrick:
I see the cause of that, fix coming up shortly.

walletdb/bdb/db.go
```diff
@@ -48,34 +48,88 @@ func convertErr(err error) error {
...
+
+func (tx *transaction) ReadWriteBucket(key []byte) walletdb.ReadWriteBucket {
+	boltBucket := tx.boltTx.Bucket(key)
+	if boltBucket == nil {
+		return nil
+	}
+	return (*bucket)(boltBucket)
+}
+
+func (tx *transaction) CreateTopLevelBucket(key []byte) (walletdb.ReadWriteBucket, error) {
```
Roasbeef: Going off of the `godoc` comment in `interface.go`, the line below should actually be using `CreateBucketIfNotExists`.

wallet/createtx.go
```diff
@@ -92,61 +95,57 @@ func (s secretSource) GetScript(addr btcutil.Address) ([]byte, error) {
...
 		}
+
+		inputSource := makeInputSource(eligible)
+		changeSource := func() ([]byte, error) {
+			// Derive the change output script.  As a hack to allow spending from
+			// the imported account, change addresses are created from account 0.
+			var changeAddr btcutil.Address
+			var err error
+			if account == waddrmgr.ImportedAddrAccount {
+				changeAddr, err = w.NewChangeAddress(0)
```
Roasbeef: This line (and the one below) will cause a deadlock within the wallet as it attempts to create a write transaction within a read transaction.

wallet/wallet.go
```diff
@@ -1771,84 +2063,120 @@ func confirms(txHeight, curHeight int32) int32 {
...
+	LastConfirmation int32
+}
 
-	var (
-		amount     btcutil.Amount
-		lastConf   int32 // Confs of the last matching transaction.
-		stopHeight int32
-	)
+// TotalReceivedForAccounts iterates through a wallet's transaction history,
+// returning the total amount of decred received for all accounts.
```
Roasbeef: decred -> bitcoin

Roasbeef:

Issue #463, opened 2016-12-07
Race on waddrmgr.newSecretKey during tests


jrick:
```
==================
WARNING: DATA RACE
Write at 0x0000009259b8 by goroutine 9:
  github.com/btcsuite/btcwallet/waddrmgr.TstRunWithReplacedNewSecretKey()
      /home/travis/gopath/src/github.com/btcsuite/btcwallet/waddrmgr/internal_test.go:38 +0x79
  github.com/btcsuite/btcwallet/waddrmgr_test.testChangePassphrase()
      /home/travis/gopath/src/github.com/btcsuite/btcwallet/waddrmgr/manager_test.go:994 +0xcd
  github.com/btcsuite/btcwallet/waddrmgr_test.testManagerAPI()
      /home/travis/gopath/src/github.com/btcsuite/btcwallet/waddrmgr/manager_test.go:1350 +0x90
  github.com/btcsuite/btcwallet/waddrmgr_test.TestManager()
      /home/travis/gopath/src/github.com/btcsuite/btcwallet/waddrmgr/manager_test.go:1743 +0x8a6
  testing.tRunner()
      /home/travis/.gimme/versions/go1.7.3.linux.amd64/src/testing/testing.go:610 +0xc9

Previous read at 0x0000009259b8 by goroutine 10:
  [failed to restore the stack]

Goroutine 9 (running) created at:
  testing.(*T).Run()
      /home/travis/.gimme/versions/go1.7.3.linux.amd64/src/testing/testing.go:646 +0x52f
  testing.RunTests.func1()
      /home/travis/.gimme/versions/go1.7.3.linux.amd64/src/testing/testing.go:793 +0xb9
  testing.tRunner()
      /home/travis/.gimme/versions/go1.7.3.linux.amd64/src/testing/testing.go:610 +0xc9
  testing.RunTests()
      /home/travis/.gimme/versions/go1.7.3.linux.amd64/src/testing/testing.go:799 +0x4ba
  testing.(*M).Run()
      /home/travis/.gimme/versions/go1.7.3.linux.amd64/src/testing/testing.go:743 +0x12f
  main.main()
      github.com/btcsuite/btcwallet/waddrmgr/_test/_testmain.go:66 +0x1b8

Goroutine 10 (running) created at:
  testing.(*T).Run()
      /home/travis/.gimme/versions/go1.7.3.linux.amd64/src/testing/testing.go:646 +0x52f
  testing.RunTests.func1()
[...]

PR #475, opened 2017-06-20
Update project dependencies.

jrick:
One change of note is that this pulls in the new btclog-based logger
and removes the seelog dependency.


jrick:
Travis is hitting #463.  This isn't an issue with this PR.

PR #482, opened 2017-09-17
Added ability to use gRPC in insecure mode for testing purposes.

michaeldowling:
Added ability to use gRPC in insecure mode for testing purposes.

I did this by expanding out the if/else statement where there is a test for --noservertls.  Rather than not doing gRPC, I instead created a grpc service without a TLS config.

It still conforms to your requirements of a localhost address.

I also added a .gitignore :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,25 @@
-btcwallet
-vendor
+# Binaries for programs and plugins
+*.exe
+*.dll
+*.so
+*.dylib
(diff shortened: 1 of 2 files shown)


jrick:
Even in testing environments we do not encourage the use of gRPC without TLS.  You're welcome to continue using this patch but I don't want this change upstreamed and maintained.

PR #502, opened 2018-04-19
An argument was missing in the NewTxIn

rafaelescrich:
Test was failing because of a missing argument in the function.


Roasbeef:

PR #503, opened 2018-05-15
multi: catch up upstream to roasbeef's fork

Roasbeef:
In this PR, we aim to catch up the upstream btcwallet with the last year or so of work that has taken place in roasbeef's fork of btcwallet. The fork was created nearly 2 years ago as we needed segwit support for `lnd`, but it wasn't yet merged in. A highlight of the changes between this branch and the current master include:

  * Full proper segwit support
  * BIP 49 and BIP 84 support
  * Fully abstracted chain backend with support for: `btcd`, `neutrino`, and `bitcoind`
  * Support for seed import rescans with a birthday
  * Proper atomicity between the `wtxmgr` and `waddrmgr`


stevenroose:
:heart_eyes:

Roasbeef:
🙏

Dreamsorcerer:
If I'm reading the PR correctly, this adds support for Neutrino, and other backends, right? (I'm looking to use this for an SPV wallet in the near future).

If so, could the README get updated? It currently says:

> btcwallet is not an SPV client and requires connecting to a local or remote btcd instance for asynchronous blockchain queries and notifications over websockets.

mecampbellsoup:
Why can't LN be implemented without SegWit support again?

stevenroose:
@mecampbellsoup In theory it could, but the current design relies heavily on transactions to be inmalleable. This means that after a tx is signed, the TXID (the hash of the tx) can no longer be changed. Before SegWit it would be possible to make changes to a tx that don't make it invalid, but change the hash.

Since in Lightning, you sign transactions that spend previous transactions that are not mined yet, you need to be able to rely on the TXID of that tx. You will ff have a `txid001` and then create a new transactions spending `txid001` that you ask the other party of your channel to sign. If he could change this tx to get id `txid001b`, the second transaction would become invalid, since `txid001` will never be mined.

mecampbellsoup:
@stevenroose thanks for that; I knew it had to do with TX malleability but didn't know exactly how or why that was a problem.

So in this example someone might broadcast `txid001b` of a previously valid (but no-longer-current) LN tx in order to ... try to invalidate subsequent payments `txid002`, `txid003`?

stevenroose:
Correct!

mecampbellsoup:
@stevenroose have you seen any good write-ups about this issue and why it would be a problem for LN? A concrete example would be great to wrap my head around this 😄

stevenroose:
I think the LN paper writes about this, as well as the duplex paper.

PR #508, opened 2018-06-06
wtxmgr: remove previously replaced transactions from RBF transaction

wpaulino:
In this commit, we fix a lingering bug within the wallet where we would
correctly mark a RBF transaction within the wallet, but we would still
mark the previous transactions replaced as valid. We fix this by
removing any transactions which double spend outputs which the RBF
transaction spends.

See https://github.com/lightningnetwork/lnd/issues/763.


wpaulino:
cc @halseth

wpaulino:
Closing in favor of #509.

PR #509, opened 2018-06-09
wtxmgr: fix double spend bug

wpaulino:
In this PR, we fix a double spend bug where we would still keep track of replaced transactions once the  _replacing_ transaction confirms. This would lead to issues where the balance would not be displayed correctly within the wallet.

See lightningnetwork/lnd#763.


wtxmgr/tx_test.go
```diff
@@ -1364,3 +1364,205 @@ func TestRemoveUnminedTx(t *testing.T) {
...
+	// in the wallet's store.
+	unminedTxs, err := store.UnminedTxs(ns)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(unminedTxs) != 2 {
+		t.Fatalf("expected 2 unmined txs, got %v", len(unminedTxs))
+	}
+
+	// Then, we'll confirm the second spend with a height deep enough that
```
halseth: comment should say either first or second based on the boolean.

wtxmgr/tx_test.go
```diff
@@ -1364,3 +1364,205 @@ func TestRemoveUnminedTx(t *testing.T) {
 			len(unminedTxns))
 	}
 }
+
+// testInsertDoubleSpendTx is a helper test which double spends an output. The
```
halseth: Excellent tests! 👍

wtxmgr/tx.go
```diff
@@ -173,68 +173,12 @@ func (s *Store) moveMinedTx(ns walletdb.ReadWriteBucket, rec *TxRecord, recKey,
...
-	err = putRawBlockRecord(ns, blockKey, blockVal)
-	if err != nil {
-		return err
-	}
-
-	err = putRawTxRecord(ns, recKey, recVal)
-	if err != nil {
-		return err
-	}
+	// Fetch the mined balance in case we need to update it.
```
halseth: Is this change really just a refactor? Seems like a lot of code was removed, that I imagine did something 😛 

Maybe add a more descriptive commit message to make it easier to follow what this change is doing? :)

wtxmgr/tx.go
```diff
@@ -335,20 +274,29 @@ func (s *Store) RemoveUnminedTx(ns walletdb.ReadWriteBucket, rec *TxRecord) erro
 // exist in the unmined buckets, but unmined double spends (including mutations)
 // are removed.
 func (s *Store) insertMinedTx(ns walletdb.ReadWriteBucket, rec *TxRecord, block *BlockMeta) error {
+	// If a transaction record for this hash and block already exists, we
+	// can exit early.
+	k, v := existsTxRecord(ns, &rec.Hash, &block.Block)
```
halseth: `k` not used?

wtxmgr/tx_test.go
halseth: Maybe also add a test for the case where we have multiple unconfimed transactions stored, then a new mined tx comes in that invalidates all of them. 

For instance:
1. Spends outpoint A
2. Spends outpoint B

Now tx 3 comes in a  block and spends both A and B.

wtxmgr/unconfirmed.go
```diff
@@ -53,25 +53,31 @@ func (s *Store) removeDoubleSpends(ns walletdb.ReadWriteBucket, rec *TxRecord) e
...
 		prevOut := &input.PreviousOutPoint
 		prevOutKey := canonicalOutPoint(&prevOut.Hash, prevOut.Index)
-		doubleSpendHash := existsRawUnminedInput(ns, prevOutKey)
-		if doubleSpendHash != nil {
+
+		doubleSpendHashes := fetchUnminedInputSpendTxHashes(ns, prevOutKey)
+		for _, doubleSpendHash := range doubleSpendHashes {
+			doubleSpendVal := existsRawUnmined(ns, doubleSpendHash[:])
+			if doubleSpendVal == nil {
+				continue
```
halseth: Is this an error case?

wtxmgr/tx_test.go
wpaulino: Fixed.

wtxmgr/tx_test.go
wpaulino: Ended up modifying `TestInsertConfirmedDoubleSpendTx` to achieve this.

wtxmgr/tx.go
wpaulino: It removes a lot of duplicate code between both methods that can be condensed into one.

wtxmgr/tx.go
wpaulino: Yep, removed.

wtxmgr/unconfirmed.go
wpaulino: Nope, this can happen in the case that the spending transaction actually spends multiple outputs from the same transaction. If this is the case, they'll get added multiple times, so it's possible to not find the unconfirmed transaction in the store if it's already been deleted. I added a comment to explain this. I tried to work around it but things got ugly quickly.

wtxmgr/tx_test.go
```diff
@@ -1364,3 +1364,217 @@ func TestRemoveUnminedTx(t *testing.T) {
...
+	defer teardown()
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	dbtx, err := db.BeginReadWriteTx()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer dbtx.Commit()
```
halseth: I think we should really commit the db transaction between insertions/extractions, as this will be what happens in real use.

wtxmgr/tx_test.go
```diff
@@ -1364,3 +1364,217 @@ func TestRemoveUnminedTx(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := store.InsertTx(ns, cbRec, &b100); err != nil {
+		t.Fatal(err)
+	}
+
+	// Then, we'll create three spends spends from the same coinbase output.
+	// The first two will remain unconfirmed, while the last should confirm
+	// and remove the remaning unconfirmed from the wallet's store.
```
halseth: nit: remaning 😛

wtxmgr/tx_test.go
```diff
@@ -1364,3 +1364,217 @@ func TestRemoveUnminedTx(t *testing.T) {
...
+	}
+
+	// Then, we'll insert the confirmed spend at a height deep enough that
+	// allows us to successfully spend the coinbase output.
+	coinbaseMaturity := int32(chaincfg.TestNet3Params.CoinbaseMaturity)
+	bMaturity := BlockMeta{
+		Block: Block{Height: b100.Height + coinbaseMaturity},
+		Time:  time.Now(),
+	}
+	confirmedSpend := spendOutput(&cbRec.Hash, 0, 3e7, 7e7)
```
halseth: you can add a third unconfirmed spend to the store that spends a second coinbase output. Then you make this `confirmedSpend` spend both coinbase outputs in the same tx.

wtxmgr/tx.go
```diff
@@ -260,50 +208,43 @@ func (s *Store) moveMinedTx(ns walletdb.ReadWriteBucket, rec *TxRecord, recKey,
...
 			return err
 		}
 		err = putUnspent(ns, &cred.outPoint, &block.Block)
 		if err != nil {
 			return err
 		}
 
-		// reposition cursor before deleting, since the above puts have
-		// invalidated the cursor.
-		it.reposition(&rec.Hash, index)
```
halseth: why is this not needed anymore?

wtxmgr/tx.go
halseth: I see, so it changes the behaviour of `moveMinedTx`, but that's okay, since it is only used by `insertMinedTx`?

wtxmgr/tx.go
```diff
@@ -670,11 +670,20 @@ func (s *Store) rollback(ns walletdb.ReadWriteBucket, height int32) error {
 
 	for _, op := range coinBaseCredits {
 		opKey := canonicalOutPoint(&op.Hash, op.Index)
-		unminedKey := existsRawUnminedInput(ns, opKey)
-		if unminedKey != nil {
-			unminedVal := existsRawUnmined(ns, unminedKey)
+		unminedSpendTxHashKeys := fetchUnminedInputSpendTxHashes(ns, opKey)
+		for _, unminedSpendTxHashKey := range unminedSpendTxHashKeys {
+			unminedVal := existsRawUnmined(ns, unminedSpendTxHashKey[:])
+			// It might be the case that this spending transaction
```
halseth: 👍

wtxmgr/tx.go
wpaulino: We're no longer deleting the credits within the iterator, so I believe there should be no need for this.

wtxmgr/tx_test.go
wpaulino: Fixed.

wtxmgr/tx_test.go
wpaulino: Fixed.

wtxmgr/tx_test.go
wpaulino: Fixed.

wtxmgr/tx.go
wpaulino: Yep!

wtxmgr/tx_test.go
```diff
@@ -1496,85 +1533,122 @@ func TestInsertConfirmedDoubleSpendTx(t *testing.T) {
...
 	// At this point, we should see both unconfirmed transactions within the
 	// store.
-	unminedTxs, err := store.UnminedTxs(ns)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if len(unminedTxs) != 2 {
-		t.Fatalf("expected 2 unmined txs, got %d", len(unminedTxs))
-	}
+	commitDBTx(t, store, db, func(ns walletdb.ReadWriteBucket) {
```
halseth: update comments to reflect the change.

wtxmgr/tx.go
```diff
@@ -169,72 +169,18 @@ func Create(ns walletdb.ReadWriteBucket) error {
 
 // moveMinedTx moves a transaction record from the unmined buckets to block
 // buckets.
-func (s *Store) moveMinedTx(ns walletdb.ReadWriteBucket, rec *TxRecord, recKey, recVal []byte, block *BlockMeta) error {
+func (s *Store) moveMinedTx(ns walletdb.ReadWriteBucket, rec *TxRecord,
```
halseth: It looks like this method is no longer moving the tx from `unmined->block` buckets, but is deleting it from the unmined bucket + updating the mined balance. I think it makes sense to split it into two new methods (`deleteUnmined` and `updateMinedBalance` or similar), then call these.

wtxmgr/tx.go
halseth: nit: newline above comment

wtxmgr/unconfirmed.go
```diff
@@ -53,25 +53,36 @@ func (s *Store) removeDoubleSpends(ns walletdb.ReadWriteBucket, rec *TxRecord) e
 	for _, input := range rec.MsgTx.TxIn {
 		prevOut := &input.PreviousOutPoint
 		prevOutKey := canonicalOutPoint(&prevOut.Hash, prevOut.Index)
-		doubleSpendHash := existsRawUnminedInput(ns, prevOutKey)
-		if doubleSpendHash != nil {
+
+		doubleSpendHashes := fetchUnminedInputSpendTxHashes(ns, prevOutKey)
+		for _, doubleSpendHash := range doubleSpendHashes {
+			doubleSpendVal := existsRawUnmined(ns, doubleSpendHash[:])
+			// It might be the case that this spending transaction
```
halseth: nit: newline

wtxmgr/unconfirmed.go
```diff
@@ -83,28 +94,34 @@ func (s *Store) removeConflict(ns walletdb.ReadWriteBucket, rec *TxRecord) error
...
-	for i := uint32(0); i < numOuts; i++ {
-		k := canonicalOutPoint(&rec.Hash, i)
-		spenderHash := existsRawUnminedInput(ns, k)
-		if spenderHash != nil {
+	for i := range rec.MsgTx.TxOut {
+		k := canonicalOutPoint(&rec.Hash, uint32(i))
+		spenderHashes := fetchUnminedInputSpendTxHashes(ns, k)
+		for _, spenderHash := range spenderHashes {
+			spenderVal := existsRawUnmined(ns, spenderHash[:])
+			// It might be the case that this spending transaction
```
halseth: nit: add newline

wtxmgr/db.go
```diff
@@ -1215,18 +1215,39 @@ func (it *unminedCreditIterator) reposition(txHash *chainhash.Hash, index uint32
 //   [0:32]   Transaction hash (32 bytes)
 
 func putRawUnminedInput(ns walletdb.ReadWriteBucket, k, v []byte) error {
-	err := ns.NestedReadWriteBucket(bucketUnminedInputs).Put(k, v)
+	spendTxHashes := ns.NestedReadBucket(bucketUnminedInputs).Get(k)
```
Roasbeef: The `godoc` comment should be updated to reflect that we now store a slice of txid's rather than a single unique txid.

wtxmgr/db.go
```diff
@@ -1215,18 +1215,39 @@ func (it *unminedCreditIterator) reposition(txHash *chainhash.Hash, index uint32
...
+	spendTxHashes = append(spendTxHashes, v...)
```
Roasbeef: Also I _think_ we might actually need to copy these bytes to do an operation like this safely within `bolt`.

wtxmgr/db.go
```diff
@@ -1215,18 +1215,39 @@ func (it *unminedCreditIterator) reposition(txHash *chainhash.Hash, index uint32
...
+	rawSpendTxHashes := ns.NestedReadBucket(bucketUnminedInputs).Get(k)
+	if rawSpendTxHashes == nil {
+		return nil
+	}
+
+	// Each transaction hash is 32 bytes.
+	spendTxHashes := make([]chainhash.Hash, 0, len(rawSpendTxHashes)/32)
+	for len(rawSpendTxHashes) > 0 {
+		var spendTxHash chainhash.Hash
+		copy(spendTxHash[:], rawSpendTxHashes[:32])
```
Roasbeef: 👍 on copy

wtxmgr/tx.go
```diff
@@ -375,7 +375,7 @@ func (s *Store) insertMinedTx(ns walletdb.ReadWriteBucket, rec *TxRecord,
 	// If this transaction previously existed within the store as
 	// unconfirmed, we'll need to move it to the confirmed bucket.
 	if v := existsRawUnmined(ns, rec.Hash[:]); v != nil {
-		return s.moveMinedTx(ns, rec, block)
+		s.moveMinedTx(ns, rec, block)
```
Roasbeef: The error here is now being ignored... 

The comment above could also be updated to reflect that we don't terminate here to ensure that any unmined conflicting transactions are properly removed.

wtxmgr/tx.go
```diff
@@ -670,11 +670,20 @@ func (s *Store) rollback(ns walletdb.ReadWriteBucket, height int32) error {
...
 		opKey := canonicalOutPoint(&op.Hash, op.Index)
-		unminedKey := existsRawUnminedInput(ns, opKey)
-		if unminedKey != nil {
-			unminedVal := existsRawUnmined(ns, unminedKey)
+		unminedSpendTxHashKeys := fetchUnminedInputSpendTxHashes(ns, opKey)
+		for _, unminedSpendTxHashKey := range unminedSpendTxHashKeys {
+			unminedVal := existsRawUnmined(ns, unminedSpendTxHashKey[:])
+			// It might be the case that this spending transaction
+			// spent multiple outpoints from the previous
+			// transaction output. If so, then we won't find this
```
Roasbeef: Multiple points from the previous transaction output? That's not possible in this context afaict. Guessing you mean spends multiple outputs from the previous transaction, so this transaction will have already been removed after the first spent output.

wtxmgr/unconfirmed.go
Roasbeef: Similar comment here about updating the comment to reflect reality.

wtxmgr/unconfirmed.go
Roasbeef: Similar comment here about updating the comment to reflect reality.

wtxmgr/tx_test.go
wpaulino: Are you referring to the individual database transactions? A comment was included at the top of the test indicating that a new database transaction would be used for each interaction.

wtxmgr/tx.go
wpaulino: Good point! Fixed.

wtxmgr/tx.go
wpaulino: Fixed.

wtxmgr/unconfirmed.go
wpaulino: Fixed.

wtxmgr/unconfirmed.go
wpaulino: Fixed.

wtxmgr/tx_test.go
```diff
@@ -1364,3 +1364,217 @@ func TestRemoveUnminedTx(t *testing.T) {
...
+	// spends for this coinbase output, as we've already seen the confirmed
+	// one. Therefore, we shouldn't see any other unconfirmed transactions
+	// within it.
+	unminedTxs, err = store.UnminedTxs(ns)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if len(unminedTxs) != 0 {
+		t.Fatalf("expected 0 unmined txs, got %v", len(unminedTxs))
+	}
```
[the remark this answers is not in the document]
wpaulino: Fixed.

wtxmgr/db.go
wpaulino: Fixed.

wtxmgr/db.go
wpaulino: I don't mind taking another approach, but the tests exercise this behavior and it seems to work. We could do something like below, but the current approach is much simpler:
```
if len(spendTxHashes)/32 == 0 {
    spendTxHashes = v
} else {
    // We allocate a new slice big enough to store the previous
    // entries and the new one.
    newSlice := make([]byte, len(spendTxHashes)+32)
    copy(newSlice, spendTxHashes)
    copy(newSlice[len(newSlice)-32:], v)
    spendTxHashes = newSlice
}
```

wtxmgr/tx.go
wpaulino: Fixed.

wtxmgr/tx.go
wpaulino: Fixed, not sure where I got this wording from lol...

PR #510, opened 2018-06-13
wallet: add dryrun arg to tx create, rolling back db if set

halseth:
Useful for creating transactions for fee estimation, e.g.

### DIFF
--- a/waddrmgr/scoped_manager.go
+++ b/waddrmgr/scoped_manager.go
@@ -818,33 +818,46 @@ func (s *ScopedKeyManager) nextAddresses(ns walletdb.ReadWriteBucket,
 		}
 	}
 
-	// Finally update the next address tracking and add the addresses to
-	// the cache after the newly generated addresses have been successfully
-	// added to the db.
 	managedAddresses := make([]ManagedAddress, 0, len(addressInfo))
 	for _, info := range addressInfo {
 		ma := info.managedAddr
-		s.addrs[addrKey(ma.Address().ScriptAddress())] = ma
-
-		// Add the new managed address to the list of addresses that
-		// need their private keys derived when the address manager is
-		// next unlocked.
-		if s.rootManager.IsLocked() && !s.rootManager.WatchOnly() {
-			s.deriveOnUnlock = append(s.deriveOnUnlock, info)
-		}
-
 		managedAddresses = append(managedAddresses, ma)
 	}
 
-	// Set the last address and next address for tracking.
-	ma := addressInfo[len(addressInfo)-1].managedAddr
-	if internal {
-		acctInfo.nextInternalIndex = nextIndex
-		acctInfo.lastInternalAddr = ma
-	} else {
-		acctInfo.nextExternalIndex = nextIndex
-		acctInfo.lastExternalAddr = ma
+	// Finally, create a closure that will update the next address tracking
+	// and add the addresses to the cache after the newly generated
+	// addresses have been successfully committed to the db.
+	onCommit := func() {
+		// Since this closure will be called when the DB transaction
+		// gets committed, we won't longer be holding the manager's
+		// mutex at that point. We must therefore re-acquire it before
+		// continuing.
+		s.mtx.Lock()
+		defer s.mtx.Unlock()
+
+		for _, info := range addressInfo {
+			ma := info.managedAddr
+			s.addrs[addrKey(ma.Address().ScriptAddress())] = ma
+
+			// Add the new managed address to the list of addresses
+			// that need their private keys derived when the
+			// address manager is next unlocked.
+			if s.rootManager.IsLocked() && !s.rootManager.WatchOnly() {
+				s.deriveOnUnlock = append(s.deriveOnUnlock, info)
+			}
+		}
+
+		// Set the last address and next address for tracking.
+		ma := addressInfo[len(addressInfo)-1].managedAddr
+		if internal {
+			acctInfo.nextInternalIndex = nextIndex
+			acctInfo.lastInternalAddr = ma
+		} else {
+			acctInfo.nextExternalIndex = nextIndex
+			acctInfo.lastExternalAddr = ma
+		}
 	}
+	ns.Tx().OnCommit(onCommit)
 
 	return managedAddresses, nil
 }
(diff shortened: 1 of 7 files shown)


wallet/createtx.go
```diff
@@ -101,15 +101,25 @@ func (s secretSource) GetScript(addr btcutil.Address) ([]byte, error) {
...
 
 	chainClient, err := w.requireChainClient()
 	if err != nil {
 		return nil, err
 	}
 
 	err = walletdb.Update(w.db, func(dbtx walletdb.ReadWriteTx) error {
+		// If dry run is requested, we must make sure the database is
+		// not altered.
+		if dryRun {
```
Roasbeef: Ideally we add a test to ensure that this works properly. At glance it _seems_ that it should, but would like an additional level of certainty.

halseth: Very good idea, turned out writing a test surfaced all kinds of problems... To be continued

halseth:
Significant rewrite pushed. The added test surfaced a problem with how the `acctInfo` is loaded from the database and cached. We would not check if any error occurred during the db transaction before updating the cached account information. This would cause the cache to get out of sync with the database in case of error or rollback.

The first few commits now address this issue by passing an `onCommit` closure back to the caller, which does all the cache updates. The caller can then execute this closure once the db transaction is commited.

An alternative I explored was to add a `OnCommit(func())` method to the `ReadWriteBucket` that could achieve the same thing, but this turned out to be a much more intrusive change..

The last couple of commits add the `dryRun` option to `CreateSimpleTx`, as well as a test that makes sure the dry run doesn't inflate the number of change addresses.

Roasbeef:
Agree that the non-interface callback is less invasive, also the better route IMO.

waddrmgr/scoped_manager.go
```diff
@@ -777,55 +780,60 @@ func (s *ScopedKeyManager) nextAddresses(ns walletdb.ReadWriteBucket,
...
-	if internal {
-		acctInfo.nextInternalIndex = nextIndex
-		acctInfo.lastInternalAddr = ma
-	} else {
-		acctInfo.nextExternalIndex = nextIndex
-		acctInfo.lastExternalAddr = ma
+	// Finally return a closure that will update the next address tracking
+	// and add the addresses to the cache after the newly generated
+	// addresses have been successfully added to the db.
+	onCommit := func() {
```
Roasbeef: We'll need to be careful with this, as we need to ensure that all access to the caches are properly thread-safe. The callback MUST be executed right at the point of the caller.

Roasbeef:
Also travis is failing with a race atm.

waddrmgr/manager_test.go
```diff
@@ -1998,6 +2003,7 @@ func TestScopedKeyManagerManagement(t *testing.T) {
 	// Next, ensure that if we create an internal and external addrs for
 	// each of the default scope types, then they're derived according to
 	// their schema.
+	var onCommits []func()
```
halseth: note to self: these are never called.

halseth:
re: race on Travis. It looks like all recent PRs fails with this error.

waddrmgr/scoped_manager.go
halseth: That's true... Seems for this to be thread safe, mutexes must be added around the db access + `onCommit` calls. 

Wondering if there's a reason to keep the current style of passing in the db transaction/bucket from the outer level, or if we instead should try changing to opening the db+updating the caches only at the point where the data is put?

halseth:
Pushed change using `boltdb`'s underlying `OnCommit` call to update the cache. Should ensure thread safety!

PTAL

Roasbeef:
Needs a rebase due to the `bbolt` switch over.

halseth:
To handle a detected race, I added a fixup commit to re-acquire the `ScopedManager` lock in the `onCommit` closure.

waddrmgr/scoped_manager.go
```diff
@@ -818,33 +818,46 @@ func (s *ScopedKeyManager) nextAddresses(ns walletdb.ReadWriteBucket,
...
-		acctInfo.nextInternalIndex = nextIndex
-		acctInfo.lastInternalAddr = ma
-	} else {
-		acctInfo.nextExternalIndex = nextIndex
-		acctInfo.lastExternalAddr = ma
+	// Finally, create a closure that will update the next address tracking
+	// and add the addresses to the cache after the newly generated
+	// addresses have been successfully committed to the db.
+	onCommit := func() {
+		// Since this closure will be called when the DB transaction
```
Roasbeef: I'm not so sure about this. [The `OnCommit` callback is called from _within_ the `Commit` method](https://github.com/boltdb/bolt/blob/master/tx.go#L230), as a result, we'll still have the mutex held [since it's grabbed here](https://github.com/btcsuite/btcwallet/blob/7dc4cdcc9af3c50ce9918137878b9d69a3e43c43/waddrmgr/scoped_manager.go#L1062).

halseth: This is tricky, but I believe the order of operations will be
```
db tx opened
    mutex locked
    mutex unlocked
    onCommit called
           mutex locked
           mutex unlocked
    onCommit done
db tx committed
```

Note that the mutex is unlocked before `NextExternalAddresses` returns, while the db tx gets commited after it has returned.

Roasbeef: Yes, it's unlocked before `NextExternalAddresses` returns, but it'll only return _after_ `nextAddresses` returns. Instead, I think it's more like: 
```
mutex locked
db tx opened
    onCommit called
           mutex locked
           mutex unlocked
    onCommit done
db tx committed
mutex unlocked
```

Notice the outer locking of the mutex _outside_ the database transaction.

halseth: But `ns walletdb.ReadWriteBucket` is passed _into_ `NextExternalAddresses`, which means that the db tx gets opened **before** the mutex is locked?

Roasbeef: Ahh, yeah you're right! Thanks for walking through this, time to land this ting.

Roasbeef: I think there's still an issue with usage of `Batch`, but we don't do so atm, so it shouldn't pop up. Worst case would be address re-use AFAICT.

halseth:
Rebased!

halseth:
Made the dry-run option return tx before adding the input scripts, to ensure it cannot be broadcasted: https://github.com/btcsuite/btcwallet/pull/510/commits/4e7dcfa14b6b704154bdd961a77d178c21e0656b

PR #511, opened 2018-06-22
chain: share the same bitcoind connection between multiple rescan clients

wpaulino:
In this PR, we introduce a nice optimization to `btcwallet`'s interaction with a `bitcoind` node. This was mostly inspired by https://github.com/lightningnetwork/lnd/issues/1174. Within `lnd`, we currently have three different subsystems responsible for watching the chain: `chainntnfs`, `lnwallet`,  and `routing/chainview`. Each of these subsystems has an active RPC and ZMQ connection to the underlying `bitcoind` node. This would incur a toll on the underlying `bitcoind` node and would cause us to miss ZMQ events, which are crucial to `lnd`. This PR attempts to remedy this by reworking how each BitcoindClient retrieves these events.

We introduce a `BitcoindConn` struct which is responsible for maintaining the RPC and ZMQ connection to a `bitcoind` node. This struct will be shared among all of the `BitcoindClients` created, and it will send each event to the currently active clients. This allows us to reduce the toll on the underlying `bitcoind` node and increases the reliability of the delivery of ZMQ events.

Some other notable changes:
* Used a concurrent unbounded queue for notification handling and delivery.
* Improved overall documentation of methods and structs.


chain/queue.go
```diff
@@ -0,0 +1,97 @@
...
+					return
+				}
+			}
+		}
+	}()
+}
+
+// Stop ends the goroutine that moves items from the in channel to the out
+// channel.
+func (cq *ConcurrentQueue) Stop() {
```
Roasbeef: If the caller isn't aware, also closing the chan in and out can lead to panics. The version we have within `lnd` atm _only_ closes the quit channel due to that.

chain/bitcoind_client.go
```diff
@@ -1157,83 +1150,3 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
 
 	return notifyTx, rec, nil
 }
-
```
Roasbeef: 👍 

Should eventually also be done for the `neutrino` and `btcd` backends.

chain/bitcoind_client.go
```diff
@@ -364,6 +365,21 @@ func (c *BitcoindClient) Start() error {
 		return errors.New("mismatched networks")
 	}
 
+	// Get initial conditions.
```
Roasbeef: Why do this outside of establishing a connection? As is now, it's possible that we have stale information by the time the client is actually connected.

chain/bitcoind_client.go
```diff
@@ -235,7 +235,7 @@ func (c *BitcoindClient) LoadTxFilter(reset bool,
 	// If we reset, signal that.
 	if reset {
 		select {
-		case c.rescanUpdate <- reset:
+		case c.rescanUpdate <- struct{}{}:
```
Roasbeef: What's the diff?

chain/bitcoind_client.go
```diff
@@ -22,6 +23,13 @@ import (
 	"github.com/lightninglabs/gozmq"
 )
 
+var (
+	// ErrBitcoindClientShuttingDown is an error returned when we attempt
+	// to receive a notification for a specific item and the bitcoind client
+	// is in the middle of shutting down.
+	ErrBitcoindClientShuttingDown = errors.New("client is shutting down")
```
Roasbeef: Are we able to recover from this error?

chain/bitcoind_client.go
```diff
@@ -36,99 +32,85 @@ type BitcoindClient struct {
...
+	// bestBlock keeps track of the tip of the current best chain.
 	bestBlockMtx sync.RWMutex
 	bestBlock    waddrmgr.BlockStamp
 
-	clientMtx      sync.RWMutex
-	rescanUpdate   chan interface{}
-	startTime      time.Time
+	// notifyBlocks signals whether the client is sending block
+	// notifications to the caller.
+	notifyBlocks uint32
```
Roasbeef: Why not a bool?

chain/bitcoind_client.go
```diff
@@ -36,99 +32,85 @@ type BitcoindClient struct {
...
+
+	// rescanUpdate is a channel will be sent items that we should match
+	// transactions against while processing a chain rescan. to determine if
```
Roasbeef: extra period

chain/bitcoind_client.go
```diff
@@ -428,17 +416,116 @@ func (c *BitcoindClient) Stop() {
...
+				c.mu.Lock()
+				for _, addr := range update {
+					c.watchAddresses[addr.String()] = struct{}{}
+				}
+				c.mu.Unlock()
+			case chainhash.Hash:
+				if err := c.rescan(update); err != nil {
+					log.Errorf("Unable to complete chain "+
+						"rescan: %v", err)
+				}
```
Roasbeef: Should have a default case here even if just for logging purposes.

chain/bitcoind_client.go
```diff
@@ -428,17 +416,116 @@ func (c *BitcoindClient) Stop() {
...
+// rescan.
+//
+// NOTE: This must be called as a goroutine.
+func (c *BitcoindClient) rescanHandler() {
+	defer c.wg.Done()
+
+	for {
+		select {
+		case update := <-c.rescanUpdate:
+			switch update := update.(type) {
```
Roasbeef: We're missing the two chain hash slice instances. Atm, these are used for tx confirmation notifications.

chain/bitcoind_client.go
```diff
@@ -1117,15 +1028,12 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
 	// If we already know this is a relevant tx from a previous ntfn, we
 	// can shortcut the filter process and let the caller know the filter
 	// matches.
-	c.clientMtx.RLock()
-	if _, ok := c.memPool[tx.TxHash()]; ok {
-		c.clientMtx.RUnlock()
+	if _, ok := c.mempool[tx.TxHash()]; ok {
```
Roasbeef: godoc comment should be updated to reflect that now the caller is responsible for holding the mutex while calling this method.

chain/bitcoind_conn.go
```diff
@@ -0,0 +1,273 @@
...
+				log.Error(err)
+			}
+
+			continue out
+		}
+
+		// We have an event! We'll now determine which kind of event was
+		// read, deserialize it, and report it to the different rescan
+		// clients.
+		switch string(msgBytes[0]) {
```
Roasbeef: Should have a default case for logging purposes.

chain/bitcoind_client.go
```diff
@@ -204,36 +222,27 @@ func (c *BitcoindClient) NotifySpent(outPoints []*wire.OutPoint) error {
...
 
-// LoadTxFilter updates the transaction watchlists for the client. Acceptable
-// arguments after `reset` are any combination of []btcutil.Address,
-// []wire.OutPoint, []*wire.OutPoint, []chainhash.Hash, and []*chainhash.Hash.
-func (c *BitcoindClient) LoadTxFilter(reset bool,
-	watchLists ...interface{}) error {
+// LoadTxFilter uses the given outpoints and addresses as filters to what we
+// should match transactions against to determine if they are relevant to the
+// client. The reset argument is used to reset the current filters.
+func (c *BitcoindClient) LoadTxFilter(reset bool, outpoints []wire.OutPoint,
```
Roasbeef: This is too restrictive. As is, things beyond outpoint and addresses can be sent.

chain/bitcoind_conn.go
```diff
@@ -167,54 +184,92 @@ out:
...
 
-		// We have an event! We'll now determine which kind of event was
-		// read, deserialize it, and report it to the different rescan
+		// We have an event! We'll now ensure it is a block event,
+		// deserialize it, and report it to the different rescan
 		// clients.
-		switch string(msgBytes[0]) {
-		case "rawtx":
-			tx := &wire.MsgTx{}
+		if string(msgBytes[0]) == "rawblock" {
```
Roasbeef: Should have an else case for logging purposes.

Roasbeef:
While this is still in flight, we should also start to create the corresponding `lnd` PR to ensure that the set of integration tests that exercise `bitcoind` still pass.

chain/queue.go
wpaulino: Yeah, good point. I added it because we currently iterate over the notification channel in `wallet.chainntfns.go`, but I decided to modify that instead with a follow-up commit.

chain/bitcoind_client.go
wpaulino: I did it this way rather than within the goroutine (as it was done before) so that we actually return the error to the caller, signifying that something could potentially be wrong. FWIW, it will be updated within `ntfnHandler` as soon as a new block notification arrives from `bitcoind`.

chain/bitcoind_client.go
wpaulino: The event loop was previously not handling receiving `bool` as you can see in https://github.com/btcsuite/btcwallet/blob/master/chain/bitcoind.go#L570.

chain/bitcoind_client.go
wpaulino: Nope. The only way this error can be returned is if the caller explicitly called `Stop` on either the `BitcoindClient` or `BitcoindConn`.

chain/bitcoind_client.go
wpaulino: It needs to be atomic.

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_client.go
wpaulino: I don't think there's a need for the mutex as the map is only used within the `ntfnHandler`.

chain/bitcoind_conn.go
wpaulino: Fixed.

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_conn.go
wpaulino: Fixed.

wpaulino:
I already have a local branch for `lnd` that incorporates these changes, was just waiting on the switch to `btcsuite`.

chain/queue.go
```diff
@@ -0,0 +1,88 @@
...
+import (
+	"container/list"
+)
+
+// ConcurrentQueue is a concurrent-safe FIFO queue with unbounded capacity.
+// Clients interact with the queue by pushing items into the in channel and
+// popping items from the out channel. There is a goroutine that manages moving
+// items from the in channel to the out channel in the correct order that must
+// be started by calling Start().
+type ConcurrentQueue struct {
```
halseth: Maybe not in the scope for this PR, but would be nice to extract this into its own package/repo, such that it could be shared between `lnd` and `btcwallet`,

chain/bitcoind_client.go
```diff
@@ -36,99 +32,86 @@ type BitcoindClient struct {
...
+	// mempool keeps track of all relevant transactions that have yet to be
+	// confirmed. This is used to shortcut the filtering process of a
+	// transaction when a new confirmed transaction notification is
+	// received.
+	mempool map[chainhash.Hash]struct{}
+
+	// expiredMempool keeps track of a set of confirmed transactions along
+	// with the height at which they were included in a block. These
+	// transactions will then be removed from the mempool after a period of
+	// 288 blocks. This is done to ensure the tranasctions are safe from a
```
halseth: nit: tranasctions

chain/bitcoind_client.go
```diff
@@ -428,17 +417,131 @@ func (c *BitcoindClient) Stop() {
...
+//
+// NOTE: This must be called as a goroutine.
+func (c *BitcoindClient) rescanHandler() {
+	defer c.wg.Done()
+
+	for {
+		select {
+		case update := <-c.rescanUpdate:
+			switch update := update.(type) {
+			case struct{}:
```
halseth: The cases should probably have a short comment explaining them.

chain/bitcoind_client.go
```diff
@@ -1097,6 +1024,9 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
 	blockDetails *btcjson.BlockDetails, notify bool) (bool,
 	*wtxmgr.TxRecord, error) {
 
+	c.mu.Lock()
```
halseth: nit: move the locking down to where it is first needed

chain/bitcoind_conn.go
```diff
@@ -0,0 +1,335 @@
...
+		default:
+		}
+
+		// Poll an event from the ZMQ socket. It's possible that the
+		// connection to the socket continuously times out, so we'll
+		// prevent logging this error to prevent spamming the logs.
+		msgBytes, err := conn.Receive()
+		if err != nil {
+			err, ok := err.(net.Error)
+			if !ok || (ok && !err.Timeout()) {
```
halseth: second `ok` can be removed

chain/bitcoind_conn.go
```diff
@@ -0,0 +1,335 @@
...
```
halseth: same

Roasbeef:
Needs a rebase!

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_conn.go
wpaulino: Fixed.

chain/bitcoind_conn.go
wpaulino: Fixed.

chain/bitcoind_client.go
```diff
@@ -36,99 +32,90 @@ type BitcoindClient struct {
...
-		return chaincfg.MainNetParams.Net, nil
-	default:
-		return 0, errors.New("unknown network")
-	}
-}
-
 // GetBestBlock returns the highest block known to bitcoind.
 func (c *BitcoindClient) GetBestBlock() (*chainhash.Hash, int32, error) {
-	bcinfo, err := c.client.GetBlockChainInfo()
+	bcinfo, err := c.chainConn.client.GetBlockChainInfo()
```
halseth: you have `b.bestBlock`, any reason it is not used?

wpaulino: Not sure, just copied the existing behavior.

chain/bitcoind_client.go
```diff
@@ -1144,5 +1212,7 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
 		c.mempool[tx.TxHash()] = struct{}{}
 	}
 
-	return notifyTx, rec, nil
+	c.onRelevantTx(rec, blockDetails)
```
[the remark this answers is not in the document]
wpaulino: Nope, the logic was simply reworked to make it a bit easier to follow.

chain/bitcoind_client.go
```diff
@@ -316,69 +328,60 @@ func (c *BitcoindClient) RescanBlocks(blockHashes []chainhash.Hash) (
...
 	}
 
-	// Update outpoints.
 	select {
 	case c.rescanUpdate <- outPoints:
 	case <-c.quit:
 		return ErrBitcoindClientShuttingDown
 	}
 
-	// Kick off the rescan with the starting block hash.
```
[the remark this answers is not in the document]
wpaulino: Fixed.

chain/bitcoind_client.go
```diff
@@ -438,17 +431,151 @@ func (c *BitcoindClient) Stop() {
 	}
 
 	close(c.quit)
-	c.client.Shutdown()
+
+	// Remove this client's reference from the bitcoind connection to
+	// prevent sending notifications to it after it's been stopped.
+	c.chainConn.rescanClientsMtx.Lock()
```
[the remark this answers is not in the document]
wpaulino: Fixed.

chain/bitcoind_client.go
```diff
@@ -392,41 +384,42 @@ func (c *BitcoindClient) Start() error {
...
-	// Connect a ZMQ socket for block notifications
-	zmqClient, err := gozmq.Subscribe(c.zmqConnect, []string{"rawblock",
-		"rawtx"}, c.zmqPollInterval)
-	if err != nil {
-		return err
-	}
+	// Once the client has started successfully, we'll include it in the set
+	// of rescan clients of the backing bitcoind connection in order to
+	// received ZMQ event notifications.
+	c.chainConn.rescanClientsMtx.Lock()
```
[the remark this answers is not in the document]
wpaulino: Good point, fixed!

chain/bitcoind_conn.go
```diff
@@ -0,0 +1,335 @@
...
+
+	switch *hash {
+	case *chaincfg.TestNet3Params.GenesisHash:
+		return chaincfg.TestNet3Params.Net, nil
+	case *chaincfg.RegressionNetParams.GenesisHash:
+		return chaincfg.RegressionNetParams.Net, nil
+	case *chaincfg.MainNetParams.GenesisHash:
+		return chaincfg.MainNetParams.Net, nil
+	default:
+		return 0, errors.New("unknown network")
```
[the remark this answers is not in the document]
wpaulino: Fixed.

chain/bitcoind_conn.go
```diff
@@ -0,0 +1,335 @@
...
+				case client.zmqTxNtfns <- tx:
+				case <-client.quit:
+				case <-c.quit:
+					c.rescanClientsMtx.Unlock()
+					return
+				}
+			}
+			c.rescanClientsMtx.Unlock()
+		} else {
+			log.Warn("Received unexpected event type:", eventType)
```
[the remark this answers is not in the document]
wpaulino: Fixed.

chain/bitcoind_conn.go
```diff
@@ -0,0 +1,335 @@
...
+				case client.zmqBlockNtfns <- block:
+				case <-client.quit:
+				case <-c.quit:
+					c.rescanClientsMtx.Unlock()
+					return
+				}
+			}
+			c.rescanClientsMtx.Unlock()
+		} else {
+			log.Warn("Received unexpected event type:", eventType)
```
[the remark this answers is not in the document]
wpaulino: Fixed.

chain/bitcoind_conn.go
```diff
@@ -0,0 +1,335 @@
...
+			}
+
+			continue
+		}
+
+		// We have an event! We'll now ensure it is a block event,
+		// deserialize it, and report it to the different rescan
+		// clients.
+		eventType := string(msgBytes[0])
+		if eventType == "rawblock" {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

chain/bitcoind_client.go
```diff
@@ -684,20 +681,24 @@ mainLoop:
...
+			c.bestBlockMtx.Lock()
+			bestBlock := c.bestBlock
+			if block.Header.PrevBlock == bestBlock.Hash {
 				// No reorg. Notify the subscriber of the block.
-				bs.Hash = block.BlockHash()
-				bs.Height++
-				bs.Timestamp = block.Header.Timestamp
-				_, err = c.filterBlock(block, bs.Height, true)
+				bestBlock.Hash = block.BlockHash()
+				bestBlock.Height++
```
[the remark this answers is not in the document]
wpaulino: Fixed. A copy is made so that we can also use it within `c.reorg`. In the case that this is the next block, we'll filter the new block, modify the copy to be the new block, and finally assign it to `c.bestBlock`.

chain/bitcoind_client.go
```diff
@@ -684,20 +681,24 @@ mainLoop:
...
 					log.Error(err)
 				}
+				c.bestBlockMtx.Unlock()
 				continue mainLoop
 			}
+			c.bestBlockMtx.Unlock()
 
 			// We have a reorg.
-			err = c.reorg(bs, block)
+			err = c.reorg(bestBlock, block)
```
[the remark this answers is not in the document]
wpaulino: That's fine as it is a copy.

chain/bitcoind_client.go
wpaulino: I believe so yeah 😅

chain/bitcoind_client.go
```diff
@@ -274,9 +292,14 @@ func (c *BitcoindClient) LoadTxFilter(reset bool,
 			sendList(list)
 
 		default:
-			log.Warnf("Couldn't add item to filter: unknown type")
+			return fmt.Errorf("unable to update filter: unknown "+
+				"type %T", list)
+		}
+		if err != nil {
```
halseth: doesn't look like this can ever be non-nil

chain/bitcoind_client.go
```diff
@@ -36,99 +32,90 @@ type BitcoindClient struct {
...
+
+	// rescanUpdate is a channel will be sent items that we should match
+	// transactions against while processing a chain rescan to determine if
+	// they are relevant to the client.
+	rescanUpdate chan interface{}
+
+	// watchedAddresses, watchedOutPoints, and watchedTxs are the set of
+	// items we should match transactions against while processing a chain
+	// rescan to determine if they are relevant to the client.
+	mu               sync.RWMutex
```
halseth: nit: if this mustex is used exclusively for these maps, a better name would be `watchMtx` or similar

chain/bitcoind_client.go
```diff
@@ -36,99 +32,90 @@ type BitcoindClient struct {
...
-		return chaincfg.MainNetParams.Net, nil
-	default:
-		return 0, errors.New("unknown network")
-	}
-}
-
 // GetBestBlock returns the highest block known to bitcoind.
 func (c *BitcoindClient) GetBestBlock() (*chainhash.Hash, int32, error) {
-	bcinfo, err := c.client.GetBlockChainInfo()
+	bcinfo, err := c.chainConn.client.GetBlockChainInfo()
```
halseth: I think it is actually better to return the `bestBlock` here, as we might risk that and what gets returned from the backend to be out of sync, and I just imagine that could lead to weird issues.

chain/bitcoind_client.go
```diff
@@ -438,17 +429,157 @@ func (c *BitcoindClient) Stop() {
...
+				bestBlock.Height++
+				bestBlock.Timestamp = block.Header.Timestamp
+
+				c.bestBlockMtx.Lock()
+				c.bestBlock = bestBlock
+				c.bestBlockMtx.Unlock()
+
+				continue
+			}
+			c.bestBlockMtx.Unlock()
```
halseth: this should be removed now

chain/bitcoind_client.go
```diff
@@ -1094,15 +1041,13 @@ func (c *BitcoindClient) filterBlock(block *wire.MsgBlock, height int32,
 	// Update the expiration map by setting the block's confirmed
 	// transactions and deleting any in the mempool that were confirmed
 	// over 288 blocks ago.
-	c.clientMtx.Lock()
```
halseth: why safe to remove mutex?

chain/bitcoind_client.go
```diff
@@ -1132,68 +1077,71 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
...
-		}
-		c.memPool[tx.TxHash()] = struct{}{}
-		c.clientMtx.Unlock()
+	if !isRelevant {
+		return false, rec, nil
+	}
+
+	// If this transaction matched any of our filters but it has yet to be
+	// confirmed, we'll defer sending its notification until it has been
+	// mined as part of FilteredBlockConnected. Otherwise, we'll send a
```
halseth: "Otherwise, we'll send a notification immediately" -> I don't see that this is being done

chain/bitcoind_client.go
```diff
@@ -1132,68 +1077,71 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
...
-		if _, ok := c.memPool[tx.TxHash()]; blockDetails == nil || !ok {
-			c.onRelevantTx(rec, blockDetails)
-		}
-		c.memPool[tx.TxHash()] = struct{}{}
-		c.clientMtx.Unlock()
+	if !isRelevant {
+		return false, rec, nil
+	}
+
+	// If this transaction matched any of our filters but it has yet to be
```
halseth: Interesting, so we never forward unconfirmed transactions to the wallet?

chain/bitcoind_client.go
```diff
@@ -1121,6 +1121,14 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
 	}
 
 	// If the transaction didn't pay to any of our watched addresses, we'll
+	// check if we're currently watching for the hash of this transaction.
+	if !isRelevant {
```
halseth: nit: the outer `if` can be removed, really.

chain/bitcoind_client.go
```diff
@@ -835,50 +893,40 @@ func (c *BitcoindClient) rescan(hash chainhash.Hash) error {
...
 	if err != nil {
 		return err
 	}
-	firstHeader := lastHeader
-
-	headers := list.New()
-	headers.PushBack(lastHeader)
-
-	// We always send a RescanFinished message when we're done.
-	defer func() {
```
halseth: why don't we want to defer this?

chain/bitcoind_client.go
```diff
@@ -901,81 +952,87 @@ func (c *BitcoindClient) rescan(hash chainhash.Hash) error {
...
-				// current hash).
+				// current hash.
 				headers.Remove(headers.Back())
 				if headers.Back() != nil {
-					lastHeader = headers.Back().
-						Value.(*btcjson.
-						GetBlockHeaderVerboseResult)
-					lastHash, err = chainhash.
-						NewHashFromStr(lastHeader.Hash)
+					previousHeader = headers.Back().Value.(*btcjson.GetBlockHeaderVerboseResult)
```
halseth: line a bit long. Possible to keep on multiple lines?

chain/bitcoind_client.go
wpaulino: Mempool checks are now shared with `watchMtx`. It's documented in the struct definition.

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_client.go
wpaulino: Fixed.

PR #514, opened 2018-07-08
wallet: fix lost balance of unmined transaction for neutrino

vapopov:
Migrated from: https://github.com/lightningnetwork/lnd/pull/1501

If we use neutrino as a bitcoin backend - after opening channel we don't count credits for unmined transactions

```
~/g/s/g/l/lnd ❯ lncli connect 03193d512b010997885b232ecd6b300917e5288de8785d6d9f619a8952728c78e8@testnet-lnd.htlc.me 
{

}
~/g/s/g/l/lnd ❯ lncli walletbalance                                                                                                                                                 
{
    "total_balance": "163020941",
    "confirmed_balance": "163020941",
    "unconfirmed_balance": "0"
}
~ ❯ lncli openchannel 03193d512b010997885b232ecd6b300917e5288de8785d6d9f619a8952728c78e8 250000
{
	"funding_txid": "fa602ff6cd70a00cf803f581de04b9bb411d0f1bbb2cb0b9ad9b574fdddcdd53"
}
~/g/s/g/l/lnd ❯ lncli walletbalance                                                                                                                                                 
{
    "total_balance": "0",
    "confirmed_balance": "0",
    "unconfirmed_balance": "0"
}
```

For the flow with `bitcoind` it should work in this way, as I understand: publish tx to backand-> bticoind adds it to mempool -> bitcoind triggers `chain.RelevantTx` notification to lnd, and then we add amount of this tx as unmined credits. With neutrino we have no such notificaiton event, thats why the balance appears only after block confirmation with this transaction.

With this fix, right after opening a channel we able to see unconfirmed balance
```
~ ❯ lncli walletbalance
{
    "total_balance": "154706191",
    "confirmed_balance": "93897730",
    "unconfirmed_balance": "60808461"
}
```

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3224,6 +3224,15 @@ func (w *Wallet) PublishTransaction(tx *wire.MsgTx) error {
 	_, err = server.SendRawTransaction(tx, false)
 	switch {
 	case err == nil:
+		switch w.chainClient.(type) {
+		// For neutrino we need to trigger adding relevant tx manually
+		// because for spv client - tx data isn't received from sync peer.
+		case *chain.NeutrinoClient:
+			return walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+				return w.addRelevantTx(tx, txRec, nil)
+			})
+		}
+
 		return nil
 
 	// The following are errors returned from btcd's mempool.


wallet/wallet.go
```diff
@@ -3224,6 +3224,18 @@ func (w *Wallet) PublishTransaction(tx *wire.MsgTx) error {
 	_, err = server.SendRawTransaction(tx, false)
 	switch {
 	case err == nil:
+		switch w.chainClient.(type) {
+		// For neutrino we need to trigger adding relevant tx manually
+		// because for spv client - tx data isn't received from sync peer.
+		case *chain.NeutrinoClient:
+			rec, err := wtxmgr.NewTxRecordFromMsgTx(tx, time.Now())
```
wpaulino: Could just use the same `txRec` found above before `SendRawTransaction`.

vapopov: you are right, didn't notice it after move from another method, chnaged

wallet/wallet.go
```diff
@@ -3224,6 +3224,13 @@ func (w *Wallet) PublishTransaction(tx *wire.MsgTx) error {
 	_, err = server.SendRawTransaction(tx, false)
 	switch {
 	case err == nil:
+		switch w.chainClient.(type) {
+		// For neutrino we need to trigger adding relevant tx manually
+		// because for spv client - tx data isn't received from sync peer.
+		case *chain.NeutrinoClient:
+			return w.AddRelevantTx(txRec, nil)
```
Roasbeef: Why's a new method needed here? Why not just call the existing one as is, wrapped in a DB transaction?

Roasbeef: Looking at this method again, it seems to be the case that above, we should actually be calling `addRelevantTx` as well. This ensure that we'll always notify the clients, and also that addresses will be marked as used, and credits added.

wpaulino: I believe `addRelevantTx` is called once the `chain.RelevantTx` notification is received, which is sent once we filter it from the backend.

vapopov: @Roasbeef I've removed method

yep `addRelevantTx` is called when notification is received, for `btcd` - `onRecvTx` for `bitcoind` - `onRelevantTx`

PR #518, opened 2018-07-14
multi: update neutrino backend to use new filter format

Roasbeef:
In this PR, we update the backends, and also chain recovery logic to be aware of the new neutrino API as a result of the modifications to BIP 158. The major change is that we must now carry along the previous output script along with the outpoint everywhere as we'll now _match_ on the prev output script, but then actually _dispatch_ notifications based on the outpoint.

### DIFF
--- a/wallet/rescan.go
+++ b/wallet/rescan.go
@@ -255,9 +255,6 @@ func (w *Wallet) Rescan(addrs []btcutil.Address, unspent []wtxmgr.Credit) error 
 		}
 
 		outpoints[output.OutPoint] = outputAddrs[0]
-		if err != nil {
-			return err
-		}
 	}
 
 	job := &RescanJob{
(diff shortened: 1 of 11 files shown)


Roasbeef:
Note that this will fail travis until https://github.com/lightninglabs/neutrino/pull/74 is merged.

wallet/rescan.go
```diff
@@ -240,9 +245,19 @@ out:
...
+	for _, output := range unspent {
+		_, outputAddrs, _, err := txscript.ExtractPkScriptAddrs(
+			output.PkScript, w.chainParams,
+		)
+		if err != nil {
+			return err
+		}
+
+		outpoints[output.OutPoint] = outputAddrs[0]
+		if err != nil {
```
wpaulino: 🤔

PR #519, opened 2018-07-17
multi: update neutrino gcs filter format

cfromknecht:
In this PR, we update the backends, and also chain recovery logic to be aware of the new neutrino API as a result of the modifications to BIP 158. The major change is that we must now carry along the previous output script along with the outpoint everywhere as we'll now match on the prev output script, but then actually dispatch notifications based on the outpoint.

Replaces #518


cfromknecht:
LGTM! 🚀

PR #524, opened 2018-08-14
waddrmgr: add new DerivationInfo method to ManagedPubKeyAddress

Roasbeef:
In this PR, we add a new method `DerivationInfo` to the
`ManagedPubKeyAddress` interface. This method is meant to provide callers
with the information necessary to independently derive each key returned
by the various methods provided to derive addresses.

We also update the tests to ensure that when we create addresses, as well as read them from disk, the `DerivationInfo` return values are set properly. 

Finally, we fix a bug in the wallet code for disconnecting blocks which would previously ignore an error.

### DIFF
--- a/waddrmgr/address.go
+++ b/waddrmgr/address.go
@@ -108,7 +108,7 @@ type ManagedPubKeyAddress interface {
 	// that backs the address via traditional methods from the HD root. For
 	// imported keys, the first value will be set to false to indicate that
 	// we don't know exactly how the key was derived.
-	DerivationInfo() (bool, KeyScope, DerivationPath)
+	DerivationInfo() (KeyScope, DerivationPath, bool)
 }
 
 // ManagedScriptAddress extends ManagedAddress and represents a pay-to-script-hash
@@ -125,7 +125,7 @@ type ManagedScriptAddress interface {
 // the private key associated with the public key.
 type managedAddress struct {
 	manager          *ScopedKeyManager
-	account          uint32
+	derivationPath   DerivationPath
 	address          btcutil.Address
 	imported         bool
 	internal         bool
@@ -181,7 +181,7 @@ func (a *managedAddress) lock() {
 //
 // This is part of the ManagedAddress interface implementation.
 func (a *managedAddress) Account() uint32 {
-	return a.account
+	return a.derivationPath.Account
 }
 
 // AddrType returns the address type of the managed address. This can be used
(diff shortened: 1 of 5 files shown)


waddrmgr/address.go
```diff
@@ -310,11 +316,33 @@ func (a *managedAddress) ExportPrivKey() (*btcutil.WIF, error) {
...
+		path  DerivationPath
+	)
+
+	// If this key is imported, then we can't return any information as we
+	// don't know precisely how the key was derived.
+	if a.imported {
+		return false, scope, path
+	}
+
+	return false, a.manager.Scope(), a.derivationPath
```
cfromknecht: shouldn't the first val be `true` here?

wallet/chainntfns.go
```diff
@@ -227,12 +227,15 @@ func (w *Wallet) disconnectBlock(dbtx walletdb.ReadWriteTx, b wtxmgr.BlockMeta)
...
 			bs.Timestamp = header.Timestamp
-
 			err = w.Manager.SetSyncedTo(addrmgrNs, &bs)
 			if err != nil {
 				return err
 			}
+
 			err = w.TxStore.Rollback(txmgrNs, b.Height)
+			if err != nil {
+				return err
```
cfromknecht: 👍

waddrmgr/manager_test.go
```diff
@@ -181,6 +194,12 @@ func testManagedPubKeyAddress(tc *testContext, prefix string,
 		}
 	}
 
+	// Imported addresses should return a nil derivation info.
+	if ok, _, _ := gotAddr.DerivationInfo(); gotAddr.Imported() && ok {
```
cfromknecht: Also test that `!ok` for non-imported?

waddrmgr/address.go
```diff
@@ -103,6 +103,12 @@ type ManagedPubKeyAddress interface {
 	// ExportPrivKey returns the private key associated with the address
 	// serialized as Wallet Import Format (WIF).
 	ExportPrivKey() (*btcutil.WIF, error)
+
+	// DerivationInfo contains the information required to derive the key
+	// that backs the address via traditional methods from the HD root. For
+	// imported keys, the first value will be set to false to indicate that
+	// we don't know exactly how the key was derived.
+	DerivationInfo() (bool, KeyScope, DerivationPath)
```
cfromknecht: not a blocker, but `ok` return value would be more consistent if it were the third item

Roasbeef: Fixed!

waddrmgr/manager_test.go
Roasbeef: Checked above.

waddrmgr/address.go
Roasbeef: Fixed! Also test updated.

PR #532, opened 2018-08-28
chain/rpc: remove unnecessary ping keep alive

wpaulino:
In this commit, we remove the keep-alive logic within the handler of the
RPCClient struct as this logic already exists within the backing
rpclient.Client instance. In this case, we'd completely stop the
connection after the ping timeout (1 min), which would render the
reconnection logic within rpcclient.Client useless.

Fixes https://github.com/lightningnetwork/lnd/issues/1775.


mably:
Would be interesting to update lnd Gopkg.lock file accordingly.  Thanks for the fix 👍

PR #533, opened 2018-08-28
Fix NewAddress deadlock

halseth:
This PR together with https://github.com/lightninglabs/neutrino/pull/92 fixes https://github.com/lightninglabs/neutrino/issues/91 by ensuring we are not in the process of shutting down an ongoing rescan while trying to add addresses to it. 

NOTE: the lock around `WaitForShutdown` was originally removed in https://github.com/btcsuite/btcwallet/commit/d7ad221ee068786bb0bab4fb507b12fb87102873. With this change it will hold the lock while waiting for shutdown, but not sure if safe. @aakselrod any thoughts?


chain/neutrino.go
```diff
@@ -332,16 +332,16 @@ func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Addre
...
 	defer s.clientMtx.Unlock()
+
 	if !s.started {
 		return fmt.Errorf("can't do a rescan when the chain client " +
 			"is not started")
 	}
+
 	if s.scanning {
 		// Restart the rescan by killing the existing rescan.
 		close(s.rescanQuit)
```
wpaulino: Why do we kill the rescan if it hasn't finished? Can't we potentially miss dispatching notifications for some events?

halseth: That's a good observation! Not sure why it was originally designed like this, but would say it makes more sense to latch on to the existing one than quit the ongoing one. 

Also possible that we will miss events by just exiting an ongoing rescan without keeping the list of watched addresses around. Will see if I can trigger this situation in tests.

halseth: Turned out to be hard to mock the backends to test this. Ended up indeed adding to an existing rescan instead of killing the ongoing one. Passing in the `Rewind` height should ensure the rescan is started from the appropriate height.

halseth:
Seeing some `lnd` test errors with the newest changes. Looking into.

chain/neutrino.go
```diff
@@ -336,15 +336,35 @@ func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Addre
 		return fmt.Errorf("can't do a rescan when the chain client " +
 			"is not started")
 	}
+
+	var inputsToWatch []neutrino.InputWithScript
+	for op, addr := range outPoints {
+		addrScript, err := txscript.PayToAddrScript(addr)
+		if err != nil {
+		}
```
Roasbeef: Missing error return here.

Roasbeef:
@halseth what's the lnd error?

PR #534, opened 2018-08-29
wtxmgr: resolve duplicate entries for outputs within bucketUnspent

wpaulino:
In this PR, we resolve a lingering bug within the wallet where it's possible that an output is added as unconfirmed credit after the fact that it has already confirmed. This would lead to duplicate entries for the same output within the wallet causing double spend transactions to be crafted.

The root cause for the unconfirmed credit being added again is most likely triggered by receiving an unconfirmed `chain.RelevantTx` notification from the backend after the transaction has already confirmed. Ideally this should not happen, but that fix is a lot more involved.

Fixes https://github.com/lightningnetwork/lnd/issues/1494.

### DIFF
--- a/wtxmgr/query_test.go
+++ b/wtxmgr/query_test.go
@@ -7,6 +7,7 @@ package wtxmgr_test
 import (
 	"bytes"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"testing"
 	"time"
@@ -64,12 +65,8 @@ func deepCopyTxDetails(d *TxDetails) *TxDetails {
 	return &cpy
 }
 
-func (q *queryState) compare(t *testing.T, s *Store, ns walletdb.ReadBucket, changeDesc string) {
-	defer func() {
-		if t.Failed() {
-			t.Fatalf("Store state queries failed after '%s'", changeDesc)
-		}
-	}()
+func (q *queryState) compare(s *Store, ns walletdb.ReadBucket,
+	changeDesc string) error {
 
 	fwdBlocks := q.blocks
 	revBlocks := make([][]TxDetails, len(q.blocks))
@@ -80,17 +77,22 @@ func (q *queryState) compare(t *testing.T, s *Store, ns walletdb.ReadBucket, cha
 	checkBlock := func(blocks [][]TxDetails) func([]TxDetails) (bool, error) {
 		return func(got []TxDetails) (bool, error) {
 			if len(fwdBlocks) == 0 {
-				return false, fmt.Errorf("entered range when no more details expected")
+				return false, errors.New("entered range " +
+					"when no more details expected")
 			}
 			exp := blocks[0]
 			if len(got) != len(exp) {
-				return false, fmt.Errorf("got len(details)=%d in transaction range, expected %d", len(got), len(exp))
+				return false, fmt.Errorf("got len(details)=%d "+
+					"in transaction range, expected %d",
+					len(got), len(exp))
 			}
 			for i := range got {
-				equalTxDetails(t, &got[i], &exp[i])
-			}
-			if t.Failed() {
-				return false, fmt.Errorf("Failed comparing range of transaction details")
+				err := equalTxDetails(&got[i], &exp[i])
+				if err != nil {
+					return false, fmt.Errorf("failed "+
+						"comparing range of "+
+						"transaction details: %v", err)
+				}
 			}
 			blocks = blocks[1:]
 			return false, nil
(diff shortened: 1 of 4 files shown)


wtxmgr/tx_test.go
```diff
@@ -1724,3 +1725,138 @@ func TestInsertConfirmedDoubleSpendTx(t *testing.T) {
...
+
+// TestInsertDuplicateCreditAfterConfirm aims to test the case where a duplicate
+// unconfirmed credit is added to the store after the intial credit has already
+// confirmed. This can lead to outputs being duplicated in the store, which can
+// lead to creating double spends when querying the wallet's UTXO set.
+func TestInsertDuplicateCreditAfterConfirm(t *testing.T) {
+	t.Parallel()
+
+	store, db, teardown, err := testStore()
+	defer teardown()
```
cfromknecht: this should come after error check

wpaulino: Fixed.

wtxmgr/tx.go
```diff
@@ -405,10 +405,15 @@ func (s *Store) AddCredit(ns walletdb.ReadWriteBucket, rec *TxRecord, block *Blo
 // duplicate (false).
 func (s *Store) addCredit(ns walletdb.ReadWriteBucket, rec *TxRecord, block *BlockMeta, index uint32, change bool) (bool, error) {
 	if block == nil {
+		// Skip from adding the credit if it's already been seen either
```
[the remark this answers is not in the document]
wpaulino: Fixed.

wtxmgr/tx_test.go
```diff
@@ -1728,3 +1728,137 @@ func TestInsertConfirmedDoubleSpendTx(t *testing.T) {
...
+	spendTx := spendOutput(&cbRec.Hash, 0, 5e7, 4e7)
+	spendTxRec, err := NewTxRecordFromMsgTx(spendTx, b101.Time)
+	if err != nil {
+		t.Fatal(err)
+	}
+	commitDBTx(t, store, db, func(ns walletdb.ReadWriteBucket) {
+		if err := store.InsertTx(ns, spendTxRec, nil); err != nil {
+			t.Fatal(err)
+		}
+		err := store.AddCredit(ns, spendTxRec, nil, 1, true)
```
halseth: question: is `InsertTx` always followed by an `AddCredit`? If yes maybe they could be combined?

wtxmgr/query_test.go
```diff
@@ -525,31 +534,23 @@ func TestStoreQueries(t *testing.T) {
...
-	newState.blocks = [][]TxDetails{
-		newState.blocks[0],
-		{newState.blocks[1][0]},
-		{newState.blocks[1][1]},
-	}
-	newState.blocks[1][0].Block = b101
-	newState.txDetails[recA.Hash][1].Block = b101
-	lastState = newState
-	tests = append(tests, queryTest{
-		desc:    "mine duplicate tx A",
```
halseth: what happened to this test?

halseth:
This is an awesome find! Should hopefully get rid of the double spending issues we have been seeing! 👍

wtxmgr/tx_test.go
```diff
@@ -64,17 +64,18 @@ func testStore() (*Store, walletdb.DB, func(), error) {
...
 		return nil, nil, func() {}, err
 	}
+
 	db, err := walletdb.Create("bdb", filepath.Join(tmpDir, "db"))
 	if err != nil {
-		teardown := func() {
-			os.RemoveAll(tmpDir)
-		}
-		return nil, nil, teardown, err
+		os.RemoveAll(tmpDir)
```
cfromknecht: 👍

wtxmgr/tx.go
```diff
@@ -405,10 +405,16 @@ func (s *Store) AddCredit(ns walletdb.ReadWriteBucket, rec *TxRecord, block *Blo
...
 func (s *Store) addCredit(ns walletdb.ReadWriteBucket, rec *TxRecord, block *BlockMeta, index uint32, change bool) (bool, error) {
 	if block == nil {
+		// If the outpoint that we should mark as credit already exists
+		// within the store, either as unconfirmed or confirmed, then we
+		// have nothing left to do and can exit.
 		k := canonicalOutPoint(&rec.Hash, index)
 		if existsRawUnminedCredit(ns, k) != nil {
 			return false, nil
 		}
+		if existsRawUnspent(ns, k) != nil {
```
cfromknecht: great catch

wtxmgr/query_test.go
```diff
@@ -512,38 +512,6 @@ func TestStoreQueries(t *testing.T) {
...
-		{
-			TxRecord: *stripSerializedTx(recA),
-			Block:    makeBlockMeta(-1),
-		},
-		newState.blocks[1][0],
-	}
-	newState.txDetails[recA.Hash][0].Block = b100
-	newState.txDetails[recA.Hash] = append(newState.txDetails[recA.Hash], newState.blocks[1][0])
-	lastState = newState
-	tests = append(tests, queryTest{
```
cfromknecht: why was this removed?

wtxmgr/query_test.go
wpaulino: Previously, the store would allow storing the same hash in both unconfirmed and confirmed buckets at the same time. We no longer allow this so the test had to be removed, otherwise it'd cause a failure.

wtxmgr/query_test.go
wpaulino: See https://github.com/btcsuite/btcwallet/pull/534#discussion_r214225146.

wtxmgr/tx_test.go
wpaulino: Do you mean just combine them for the sake of the test? Outside of tests, it looks like it'd be a pretty big, possibly irrelevant change. FWIW there's a TODO for it https://github.com/btcsuite/btcwallet/blob/master/wtxmgr/tx.go#L387.

PR #535, opened 2018-08-31
wallet/wallet: notify addrs+props after db commit

cfromknecht:
This PR moves any address notifications outside of the
db transaction that creates them. This is known to have
resulted in deadlocks, since `chainClient.NotifyReceived`
could block the db transaction from committing.

Doing so also prevents the situation where we send
notifications about the new addresses, but the db txn
fails to commit and the addresses are in fact never
created.


cfromknecht:
This PR now includes a fix to prevent a testing race from modifying `newSecretKey` concurrently

PR #541, opened 2018-09-10
Wait for neutrino filter headers

halseth:
This PR makes `btcwallet` utilize the changes in https://github.com/lightninglabs/neutrino/pull/96, where we will properly wait for filter headers to be synced before considering the chain backend to be current.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,7 +12,7 @@ before_install:
 install:
   - go install -v $(glide novendor)
   - go get -v golang.org/x/tools/cmd/cover
-  - go get -v github.com/golang/lint/golint
+  - go get -v golang.org/x/lint/golint
   - go get -v github.com/davecgh/go-spew/spew
 script:
   - export PATH=$PATH:$HOME/gopath/bin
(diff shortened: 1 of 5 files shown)


Roasbeef:
Also this PR re-introduces the issue with NewAddress hanging on initial sync, since it'll wait in that upper loop until the filter headers are fully caught up. As a result, it doesn't release the DB transaction to allow the new address to be written to disk.

halseth:
That's a good point, since Neutrino won't start getting filter headers before it has all block headers, this will cause sync to "not progress" for a while from the point of view of the wallet. This should be fixed before merging this.

I think not having the wallet (or lnd) be aware of the distinction between block headers and filter headers is beneficial, since it moves all this complexity to Neutrino, and can make them more chain agnostic.

Can we start catching up filter headers before block headers are "done"? This seems like would be a win both for the above problem and speed.

Roasbeef:
> Can we start catching up filter headers before block headers are "done"? This seems like would be a win both for the above problem and speed.

 
Yep, I _think_ this should address the problem. Once the other PR #neutrino/98 is merged in, I'l test this again to ensure the behavior is correct.

halseth:
Blocked by https://github.com/lightninglabs/neutrino/pull/98 and https://github.com/lightninglabs/neutrino/pull/96

PR #548, opened 2018-09-21
wtxmgr/tx_test: extend TestRemoveUnminedTx to check balances

wpaulino:
In this commit, we extend TestRemoveUnminedTx to also account for
checking the store's total balance (confirmed and unconfirmed). It
currently ensures that the UTXO state is correct, but as a sanity check,
we'll also ensure that balances are properly updated.

### DIFF
--- a/wtxmgr/tx_test.go
+++ b/wtxmgr/tx_test.go
@@ -1324,59 +1324,142 @@ func TestRemoveUnminedTx(t *testing.T) {
 	}
 	defer teardown()
 
-	dbtx, err := db.BeginReadWriteTx()
+	// In order to reproduce real-world scenarios, we'll use a new database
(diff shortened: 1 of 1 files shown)


wtxmgr/tx_test.go
```diff
@@ -1324,59 +1324,142 @@ func TestRemoveUnminedTx(t *testing.T) {
...
+			t.Fatal(err)
+		}
+		err := store.AddCredit(ns, spendTxRec, nil, 1, true)
+		if err != nil {
+			t.Fatal(err)
+		}
+	})
+
+	// With the unconfirmed spend inserted into the store, we'll query it
+	// for its unconfirmed tranasctions to ensure it was properly added.
```
halseth: nit: tranasctions

PR #554, opened 2018-09-28
wallet/wallet: rescan from current height after recovery

halseth:
This avoids doing a costly redundant rescan from the birthday height. It
was initially done to ensure we didn't miss payments confirmed while the
recovery was in progress, but we now instead scan from the syncedTo
height, which we already have checked for our addresses.


halseth:
cc @cfromknecht

Roasbeef:
cc @cfromknecht for review

Issue #513, opened 2018-07-08
wallet: only store last N headers on disk rather than the entire chain, optimize header fetching


Roasbeef:
During the initial sync of the wallet with the chain backend, [we first wait for the chain backend to finish syncing](https://github.com/btcsuite/btcwallet/blob/master/wallet/wallet.go#L436), then we being to start to fetch headers and [record our (hash, height) index](https://github.com/btcsuite/btcwallet/blob/master/wallet/wallet.go#L484) so we can handle re-orgs. 

In a prior version of this code, we only kept the past 25 or so (hash, height) pairs as it was assumed anything beyond this range would be disastrous. However, when testing `lnd` on testnet, we frequently saw 1000+ block re-orgs, so [we moved to storing _each_ block we knew of to be able to handle a re-org of unbounded length](https://github.com/btcsuite/btcwallet/blob/master/waddrmgr/db.go#L1833). 

Instead, we should find a nice happy place, and store say only the last 10k or so blocks. This should cover us for any outlandish re-orgs, and also allow us to reduce the db size and thus mmap size of boltdb. 

Additionally, the default backend code and the foreground wallet sync code should be reconciled to eliminate the redundant I/O that currently exists during IBD. 

Finally, we can also fetch headers in batch from the chain backend, which will significantly speed up sync for backends like `bitcoind`.

PR #555, opened 2018-09-28
waddrmgr/db: delete blockhashes 10000 deep

halseth:
Addresses the first part of https://github.com/btcsuite/btcwallet/issues/513

### DIFF
--- a/waddrmgr/db.go
+++ b/waddrmgr/db.go
@@ -19,6 +19,10 @@ import (
 const (
 	// LatestMgrVersion is the most recent manager version.
 	LatestMgrVersion = 5
+
+	// maxReorgDepth is the maximum number of block hashes we will store to
+	// handle reorgs.
+	maxReorgDepth = 10000
 )
 
 var (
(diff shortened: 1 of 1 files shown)


waddrmgr/db.go
```diff
@@ -1866,6 +1870,17 @@ func putSyncedTo(ns walletdb.ReadWriteBucket, bs *BlockStamp) error {
 	if err != nil {
 		return managerError(ErrDatabase, errStr, err)
 	}
+
+	// Finally we delete any hashes for heights below our max reorg depth.
+	forgetHeight := bs.Height - maxReorgDepth
```
Roasbeef: Just so we can be sure of the behavior here, can we cast this to an int32?

halseth: `bs.Height` already is, so it will be `int32`.

waddrmgr/db.go
```diff
@@ -1866,6 +1870,17 @@ func putSyncedTo(ns walletdb.ReadWriteBucket, bs *BlockStamp) error {
...
+	if forgetHeight > 0 {
+		height := make([]byte, 4)
+		binary.BigEndian.PutUint32(height, uint32(forgetHeight))
+		if err := bucket.Delete(height); err != nil {
```
halseth: One thing we should consider is for user having old databases, this will not delete the oldest hashes. We could consider looping up to this height to delete them.

Roasbeef: I think we can add that in a follow up, as fwiw, this is mainly for new nodes coming up which will be our application and btcd/bitcoind nodes.

waddrmgr/db.go
Roasbeef: 👍

waddrmgr/db.go
Roasbeef:
```suggestion
		if err := bucket.Delete(height); err != nil {
```

halseth:
Replaced by #618

PR #561, opened 2018-10-06
wtxmgr: add HasSpendingTx method

wpaulino:
In this PR, we add a new method to the transaction store: `HasSpendingTx`. This will allow the store to quickly determine whether it knows of a transaction that spends an output. This is done through the `bucketMinedInputs` bucket that was also added as a part of this PR. Due to existing wallets not previously having this bucket created, migration support has been added to the store, along with a migration to actually populate this new bucket.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -23,8 +23,6 @@ import (
 	"github.com/btcsuite/btcd/rpcclient"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
-	"github.com/davecgh/go-spew/spew"
-
 	"github.com/btcsuite/btcutil"
 	"github.com/btcsuite/btcutil/hdkeychain"
 	"github.com/btcsuite/btcwallet/chain"
@@ -33,6 +31,7 @@ import (
 	"github.com/btcsuite/btcwallet/wallet/txrules"
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/wtxmgr"
+	"github.com/davecgh/go-spew/spew"
 )
 
 const (
@@ -3313,6 +3312,27 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 	default:
 		return nil, err
 	}
+}
+
+// GetSpendingTx determines whether the store contains a spending transaction
+// for the given outpoint. This transaction can be either confirmed or
+// unconfirmed.
+//
+// NOTE: It's possible for there not to be a spending transaction for this
+// outpoint, therefore a nil check must be used to guarantee safety to the
+// caller.
+func (w *Wallet) GetSpendingTx(op wire.OutPoint) (*wire.MsgTx, error) {
+	var (
+		spendTx *wire.MsgTx
+		err     error
+	)
+	err = walletdb.View(w.db, func(tx walletdb.ReadTx) error {
+		ns := tx.ReadBucket(wtxmgrNamespaceKey)
+		spendTx, err = w.TxStore.GetSpendingTx(ns, op)
+		return err
+	})
+
+	return spendTx, err
 }
 
 // ChainParams returns the network parameters for the blockchain the wallet
--- a/wtxmgr/migrations.go
+++ b/wtxmgr/migrations.go
@@ -16,6 +16,10 @@ var dbVersions = []dbVersion{
 		version:   1,
 		migration: nil,
 	},
+	{
+		version:   2,
+		migration: migrationBucketMinedInputs,
+	},
 }
 
 // getLatestDBVersion retrieves the most recent recent of the store.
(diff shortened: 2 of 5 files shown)


wtxmgr/tx.go
```diff
@@ -931,3 +981,114 @@ func (s *Store) Balance(ns walletdb.ReadBucket, minConf int32, syncHeight int32)
 
 	return bal, nil
 }
+
+// HasSpendingTx determines whether the store contains a spending transaction
+// for the given outpoint. This transaction can be either confirmed or
+// unconfirmed.
+func (s *Store) HasSpendingTx(ns walletdb.ReadBucket,
+	outpoint wire.OutPoint) (bool, *wire.MsgTx, error) {
```
joostjager: Can the bool return parameter be removed and a nil tx be used to indicated that there isn't a spending tx? I think the only case where you return true and a nil tx is when there is a serialization error. In that case return nil, err is good enough?

wpaulino: Agreed- fixed!

wtxmgr/tx.go
```diff
@@ -942,3 +943,122 @@ func (s *Store) Balance(ns walletdb.ReadBucket, minConf int32, syncHeight int32)
...
+	outpoint wire.OutPoint) (*wire.MsgTx, error) {
+
+	var rawSpendTx []byte
+
+	// We'll start by checking if there is an unconfirmed spending
+	// transaction within the store for the given outpoint.
+	k := canonicalOutPoint(&outpoint.Hash, outpoint.Index)
+	unconfirmedSpend, err := s.hasUnconfirmedSpendingTx(ns, k)
+	if err != nil {
+		// TODO(wilmer): check confirmed?
```
Roasbeef: ?

wtxmgr/tx.go
```diff
@@ -942,3 +943,122 @@ func (s *Store) Balance(ns walletdb.ReadBucket, minConf int32, syncHeight int32)
...
+	// If there doesn't exist an entry for this outpoint in the bucket, then
+	// the outpoint doesn't have an unconfirmed spending transaction, so we
+	// can return to the caller.
+	if len(unminedSpendTxKeys) == 0 {
+		return nil, nil
+	}
+
+	// Otherwise, an unconfirmed spending transaction does exist, so we'll
+	// attempt to fetch it.
+	unminedSpendTxKey := unminedSpendTxKeys[0]
```
Roasbeef: When will we have a multiple transactions spending the same outpoint?

wtxmgr/tx.go
```diff
@@ -942,3 +943,122 @@ func (s *Store) Balance(ns walletdb.ReadBucket, minConf int32, syncHeight int32)
...
+	// can return to the caller.
+	if spendTxKey == nil {
+		return nil, nil
+	}
+
+	// Otherwise, a confirmed spending transaction does exist, so we'll
+	// attempt to fetch it. spendTxKey also includes the input index of the
+	// transaction spending the output, which we'll be discarding as it's
+	// not needed in this case.
+	spendTx := existsRawTxRecord(ns, spendTxKey[:68])
```
Roasbeef: The 68 magic number should be eliminated in favor of a constant somewhere.

wtxmgr/tx_test.go
```diff
@@ -1945,3 +1945,116 @@ func TestAddDuplicateCreditAfterConfirm(t *testing.T) {
...
+	commitDBTx(t, store, db, func(ns walletdb.ReadWriteBucket) {
+		spendTx, err := store.HasSpendingTx(ns, cbOutpoint)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if spendTx != nil {
+			t.Fatal("expected store to not have a spending tx")
+		}
+	})
+	// checkSpendTx is a helper closure that will assert that the store
```
Roasbeef: Nit: missing newline above here.

wtxmgr/migrations.go
```diff
@@ -0,0 +1,38 @@
...
+type dbVersion struct {
+	version   uint32
+	migration func(walletdb.ReadWriteBucket) error
+}
+
+// dbVersions represents the different versions of the store, along with the
+// migrations allowing them to proceed to said versions.
+var dbVersions = []dbVersion{
+	{
+		version:   1,
```
Roasbeef: Why not zero?

wtxmgr/migrations.go
```diff
@@ -36,3 +40,63 @@ func getMigrationsToApply(version uint32) []dbVersion {
 	}
 	return migrations
 }
+
+// migrationBucketMinedInputs is the migration responsible for creating a new
```
Roasbeef: As this is a migration touches a sensitive section of the wallet (the tx tracking), we should add a test for the migration itself. This test should ensure that it can handle: upgrade from zero, upgrade to same version, upgrade w/ a higher db version (as we hit version 5 before this). 

For inspiration, check out the way we test migrations within lnd. https://github.com/lightningnetwork/lnd/blob/master/channeldb/migrations_test.go

wtxmgr/migrations.go
```diff
@@ -36,3 +40,63 @@ func getMigrationsToApply(version uint32) []dbVersion {
...
+	// entry then maps to the output it spends, which will allow us to
+	// populate the new bucket.
+	c := ns.NestedReadBucket(bucketDebits).ReadCursor()
+	for k, v := c.First(); k != nil; k, v = c.Next() {
+		// Each value of the debits bucket includes the block at which
+		// the spent output confirmed at. We're not interested in all of
+		// the information, so we'll omit it and only grab what we need
+		// (transaction hash and output index).
+		outpointSpent := make([]byte, 32+4)
+		copy(outpointSpent[:32], v[8:40])
```
Roasbeef: Similar comment re magic numbers here.

wallet/wallet.go
```diff
@@ -3315,6 +3314,27 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
 }
 
+// HasSpendingTx determines whether the store contains a spending transaction
+// for the given outpoint. This transaction can be either confirmed or
+// unconfirmed.
+//
+// NOTE: It's possible for there not to be a spending transaction for this
+// outpoint, therefore a nil check must be used to guarantee safety to the
+// caller.
+func (w *Wallet) HasSpendingTx(op wire.OutPoint) (*wire.MsgTx, error) {
```
joostjager: Now that the boolean is gone, we should probably change the name of the func too. `GetWalletSpendingTx` or something? (maybe reference wallet too to distinguish from getting any spending tx)

halseth: `SpendingTx` should be more in line with the standard Go names I reckon.

PR #562, opened 2018-10-12
gRPC v1.12.0: Register all services before call server.Serve.

nhandl3:
Reference dcrwallet for this implementation.


nhandl3:
@Roasbeef Can you assign reviewer for this?

PR #566, opened 2018-10-25
Close wallet db on shutdown

roeierez:
This PR ensure that after the wallet is shut down the database will be closed as well.
To do so the "syncWithChain" go routine was added to the wait group to make sure it competes before we attempt to close the wallet db.
In addition I added explicit db.Close call after all go routines have exited.
This is crucial to be able to use dependent systems such LND as library (such as in mobile) and run it multiple times on the same process.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -261,6 +261,10 @@ func (w *Wallet) WaitForShutdown() {
 	}
 	w.chainClientLock.Unlock()
 	w.wg.Wait()
+
+	//We can close the db now that we know all go routines have exited and
+	//we are sure no further attempts to write to the wallet db will be done.
+	w.db.Close()
 }
 
 // SynchronizingToNetwork returns whether the wallet is currently synchronizing
@@ -315,6 +319,21 @@ func (w *Wallet) activeData(dbtx walletdb.ReadTx) ([]btcutil.Address, []wtxmgr.C
 	}
 	unspent, err := w.TxStore.UnspentOutputs(txmgrNs)
 	return addrs, unspent, err
+}
+
+func (w *Wallet) onClientConnected() {
+	w.wg.Add(1)
+	go func() {
+		// At the moment there is no recourse if the rescan fails for
+		// some reason, however, the wallet will not be marked synced
+		// and many methods will error early since the wallet is known
+		// to be out of date.
+		defer w.wg.Done()
+		err := w.syncWithChain()
+		if err != nil && !w.ShuttingDown() {
+			log.Warnf("Unable to synchronize wallet to chain: %v", err)
+		}
+	}()
 }
 
 // syncWithChain brings the wallet up to date with the current chain server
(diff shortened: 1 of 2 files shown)


howardpen9:
Hi, would you support SegWit and LND?

wallet/wallet.go
```diff
@@ -261,6 +261,10 @@ func (w *Wallet) WaitForShutdown() {
 	}
 	w.chainClientLock.Unlock()
 	w.wg.Wait()
+
+	//We can close the db now that we know all go routines have exited and
+	//we are sure no further attempts to write to the wallet db will be done.
+	w.db.Close()
```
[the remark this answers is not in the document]
roeierez: Thanks for your review @halseth and sorry for my late response, somehow lost track of this issue.
To see that I understand: 
1. I will leave the onClientConnected function since it adds a waiting for the syncToChain go routine that is needed.
2. I will remove the db.Close call.

The impact of this is that users that uses btcwallet as library will be responsible to call loader.UnloadWallet on shutdown and not only wallet.Stop (as it is now in LND).
If this is what you think needed I will need to create another fix in LND to explicitly call UnloadWallet on shutdown.
LMK your thoughts.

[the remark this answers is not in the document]
roeierez: At the moment I removed the db.Close and rebased.

wallet/wallet.go
```diff
@@ -317,6 +317,22 @@ func (w *Wallet) activeData(dbtx walletdb.ReadTx) ([]btcutil.Address, []wtxmgr.C
 	return addrs, unspent, err
 }
 
+func (w *Wallet) onClientConnected(birthdayStamp *waddrmgr.BlockStamp) {
+	w.wg.Add(1)
```
halseth: this should be called the line before the goroutine

roeierez: done!

wallet/wallet.go
```diff
@@ -317,6 +321,21 @@ func (w *Wallet) activeData(dbtx walletdb.ReadTx) ([]btcutil.Address, []wtxmgr.C
...
 }
 
+func (w *Wallet) onClientConnected() {
+	w.wg.Add(1)
+	go func() {
+		// At the moment there is no recourse if the rescan fails for
+		// some reason, however, the wallet will not be marked synced
+		// and many methods will error early since the wallet is known
+		// to be out of date.
+		defer w.wg.Done()
```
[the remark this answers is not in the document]
roeierez: done!

wallet/wallet.go
roeierez: @halseth another issue with the syncWithChain function related to the shutdown is that it runs in a loop and doesn't check for ShuttingDown() during its run. The problem even get worst when the shutdown request comes at time where the code is at this point: https://github.com/btcsuite/btcwallet/blob/master/wallet/wallet.go#L460
the go routine keeps running forever stuck inside this block preventing from wallet shutdown.
I can add such kind of check inside this block and on every iteration of the outer loop. Do you think it is a good solution or can suggest a better one?

halseth: Yeah, checking the quit signal on every iteration sounds like  a good idea :)

roeierez:
Squashed.

PR #568, opened 2018-11-05
wallet: return full tx from SendOutputs

joostjager:
Sometimes it is needed to spend a specific output from the SendOutputs tx. It could be that the tx is still unconfirmed at that point. To do this, the full tx needs to be looked up by using the returned tx hash and the tx outs need to be searched to find the output indices.

This PR adds an output index mapping return value to SendOutputs.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3079,9 +3079,9 @@ func (w *Wallet) TotalReceivedForAddr(addr btcutil.Address, minConf int32) (btcu
 }
 
 // SendOutputs creates and sends payment transactions. It returns the
-// transaction hash upon success.
+// transaction upon success.
 func (w *Wallet) SendOutputs(outputs []*wire.TxOut, account uint32,
-	minconf int32, satPerKb btcutil.Amount) (*chainhash.Hash, error) {
+	minconf int32, satPerKb btcutil.Amount) (*wire.MsgTx, error) {
 
 	// Ensure the outputs to be created adhere to the network's consensus
 	// rules.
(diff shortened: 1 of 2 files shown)


joostjager:
That is fine too. I just kept it to what I need. Do we also want to return the full tx in the wallet kit rpc?

joostjager:
Modified to return full tx.

wallet/wallet.go
```diff
@@ -3081,7 +3081,7 @@ func (w *Wallet) TotalReceivedForAddr(addr btcutil.Address, minConf int32) (btcu
 // SendOutputs creates and sends payment transactions. It returns the
 // transaction hash upon success.
 func (w *Wallet) SendOutputs(outputs []*wire.TxOut, account uint32,
-	minconf int32, satPerKb btcutil.Amount) (*chainhash.Hash, error) {
+	minconf int32, satPerKb btcutil.Amount) (*wire.MsgTx, error) {
```
halseth: godoc should be updated

wallet/wallet.go
```diff
@@ -3100,7 +3100,17 @@ func (w *Wallet) SendOutputs(outputs []*wire.TxOut, account uint32,
...
 	}
 
-	return w.publishTransaction(createdTx.Tx)
+	txHash, err := w.publishTransaction(createdTx.Tx)
+	if err != nil {
+		return nil, err
+	}
+
+	// Sanity check on the returned tx hash.
+	if *txHash != createdTx.Tx.TxHash() {
```
halseth: Is this really needed?

joostjager: Not sure about it. Don't know if there is a risk here with any of the backends?

halseth: I doubt it, but I guess it doesn't hurt to do the check.

PR #570, opened 2018-11-06
waddrmgr: convert unit tests to package-level tests

wpaulino:
In this PR, we convert our unit tests to have package-level access. We do this as an effort to reduce test code duplication when we introduce migration tests which require access to specific unexported functions/methods.

By removing these test work-arounds, we also address some lingering data race issues in the test suite.

Depends on #567.


Roasbeef:
Can be rebased now that #567 has been merged!

wpaulino:
Rebased!

waddrmgr/manager_test.go
```diff
@@ -1148,7 +1169,9 @@ func testChangePassphrase(tc *testContext) bool {
 	}
 
 	// Ensure the public passphrase was successfully changed.
-	if !tc.rootManager.TstCheckPublicPassphrase(pubPassphrase2) {
+	secretKey := snacl.SecretKey{Key: &snacl.CryptoKey{}}
```
halseth: Possible to get the secret key from the root manager here? 

Hard to follow what is going on here, at least add a comment if not possible.

waddrmgr/manager_test.go
```diff
@@ -1980,16 +2003,11 @@ func TestEncryptDecryptErrors(t *testing.T) {
...
+	// Make sure to cover the ErrCrypto error path in Encrypt and Decrypt.
+	mgr.cryptoKeyPriv = &failingCryptoKey{}
+	_, err = mgr.Encrypt(CKTPrivate, []byte{})
 	checkManagerError(t, "failed encryption", err, ErrCrypto)
-
-	// Make sure to cover the ErrCrypto error path in Decrypt.
-	TstRunWithFailingCryptoKeyPriv(mgr, func() {
-		_, err = mgr.Decrypt(CKTPrivate, []byte{})
-	})
+	_, err = mgr.Decrypt(CKTPrivate, []byte{})
```
halseth: newline above

waddrmgr/manager_test.go
wpaulino: Added a comment. IIUC, the point of this snippet is to ensure that we can re-derive the public key with the new passphrase.

waddrmgr/manager_test.go
wpaulino: Fixed.

PR #571, opened 2018-11-07
wallet+waddrmgr: add migration & sanity check to populate birthday block

wpaulino:
In this PR, we add a new migration to the `waddrmgr` to populate the birthday block for existing wallets. This will deem useful when performing rescans for whatever reason, as we'll now be able to start from this point rather than the genesis block, incurring a longer rescan.

The migration is not reliable since we do not store block timestamps, so we'll need to estimate our height by looking at the genesis timestamp and assuming a block occurs every 10 minutes. This can be unsafe, and cause us to actually miss on-chain events, so a sanity check has also been added to ensure its correctness.

This sanity check performs two main checks: whether the birthday block timestamp reflects a time before the birthday timestamp and whether the delta between these two timestamps is a reasonable amount. The birthday block is then found as the first candidate that satisfies both of these conditions.

Depends on #570.

### DIFF
--- a/waddrmgr/db.go
+++ b/waddrmgr/db.go
@@ -8,6 +8,7 @@ package waddrmgr
 import (
 	"crypto/sha256"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"time"
 
@@ -1876,6 +1877,10 @@ func fetchBlockHash(ns walletdb.ReadBucket, height int32) (*chainhash.Hash, erro
 	heightBytes := make([]byte, 4)
 	binary.BigEndian.PutUint32(heightBytes, uint32(height))
 	hashBytes := bucket.Get(heightBytes)
+	if hashBytes == nil {
+		err := errors.New("block not found")
+		return nil, managerError(ErrBlockNotFound, errStr, err)
+	}
 	if len(hashBytes) != 32 {
 		err := fmt.Errorf("couldn't get hash from database")
 		return nil, managerError(ErrDatabase, errStr, err)
--- a/waddrmgr/db.go
+++ b/waddrmgr/db.go
@@ -252,9 +252,10 @@ var (
 	watchingOnlyName    = []byte("watchonly")
 
 	// Sync related key names (sync bucket).
-	syncedToName   = []byte("syncedto")
-	startBlockName = []byte("startblock")
-	birthdayName   = []byte("birthday")
+	syncedToName      = []byte("syncedto")
+	startBlockName    = []byte("startblock")
+	birthdayName      = []byte("birthday")
+	birthdayBlockName = []byte("birthdayblock")
 )
 
 // uint32ToBytes converts a 32 bit unsigned integer into a 4-byte slice in
@@ -1929,32 +1930,80 @@ func putStartBlock(ns walletdb.ReadWriteBucket, bs *BlockStamp) error {
 
 // fetchBirthday loads the manager's bithday timestamp from the database.
 func fetchBirthday(ns walletdb.ReadBucket) (time.Time, error) {
-	bucket := ns.NestedReadBucket(syncBucketName)
-
 	var t time.Time
 
-	buf := bucket.Get(birthdayName)
-	if len(buf) != 8 {
+	bucket := ns.NestedReadBucket(syncBucketName)
+	birthdayTimestamp := bucket.Get(birthdayName)
+	if len(birthdayTimestamp) != 8 {
 		str := "malformed birthday stored in database"
 		return t, managerError(ErrDatabase, str, nil)
 	}
 
-	t = time.Unix(int64(binary.BigEndian.Uint64(buf)), 0)
+	t = time.Unix(int64(binary.BigEndian.Uint64(birthdayTimestamp)), 0)
+
 	return t, nil
 }
 
 // putBirthday stores the provided birthday timestamp to the database.
 func putBirthday(ns walletdb.ReadWriteBucket, t time.Time) error {
+	var birthdayTimestamp [8]byte
+	binary.BigEndian.PutUint64(birthdayTimestamp[:], uint64(t.Unix()))
+
 	bucket := ns.NestedReadWriteBucket(syncBucketName)
-
-	buf := make([]byte, 8)
-	binary.BigEndian.PutUint64(buf, uint64(t.Unix()))
-
-	err := bucket.Put(birthdayName, buf)
-	if err != nil {
+	if err := bucket.Put(birthdayName, birthdayTimestamp[:]); err != nil {
 		str := "failed to store birthday"
 		return managerError(ErrDatabase, str, err)
 	}
+
+	return nil
+}
+
+// fetchBirthdayBlock retrieves the birthday block from the database.
+//
+// The block is serialized as follows:
+//   [0:4]   block height
+//   [4:36]  block hash
+//   [36:44] block timestamp
+func fetchBirthdayBlock(ns walletdb.ReadBucket) (BlockStamp, error) {
+	var block BlockStamp
+
+	bucket := ns.NestedReadBucket(syncBucketName)
+	birthdayBlock := bucket.Get(birthdayBlockName)
+	if birthdayBlock == nil {
+		str := "birthday block not set"
+		return block, managerError(ErrBirthdayBlockNotSet, str, nil)
+	}
+	if len(birthdayBlock) != 44 {
+		str := "malformed birthday block stored in database"
+		return block, managerError(ErrDatabase, str, nil)
+	}
+
+	block.Height = int32(binary.BigEndian.Uint32(birthdayBlock[:4]))
+	copy(block.Hash[:], birthdayBlock[4:36])
+	t := int64(binary.BigEndian.Uint64(birthdayBlock[36:]))
+	block.Timestamp = time.Unix(t, 0)
+
+	return block, nil
+}
+
+// putBirthdayBlock stores the provided birthday block to the database.
+//
+// The block is serialized as follows:
+//   [0:4]   block height
+//   [4:36]  block hash
+//   [36:44] block timestamp
+func putBirthdayBlock(ns walletdb.ReadWriteBucket, block BlockStamp) error {
+	var birthdayBlock [44]byte
+	binary.BigEndian.PutUint32(birthdayBlock[:4], uint32(block.Height))
+	copy(birthdayBlock[4:36], block.Hash[:])
+	binary.BigEndian.PutUint64(birthdayBlock[36:], uint64(block.Timestamp.Unix()))
+
+	bucket := ns.NestedReadWriteBucket(syncBucketName)
+	if err := bucket.Put(birthdayBlockName, birthdayBlock[:]); err != nil {
+		str := "failed to store birthday block"
+		return managerError(ErrDatabase, str, err)
+	}
+
 	return nil
 }
 
--- a/waddrmgr/migrations.go
+++ b/waddrmgr/migrations.go
@@ -2,7 +2,9 @@ package waddrmgr
 
 import (
 	"fmt"
+	"time"
 
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/walletdb/migration"
 )
@@ -19,6 +21,10 @@ var versions = []migration.Version{
 	{
 		Number:    5,
 		Migration: upgradeToVersion5,
+	},
+	{
+		Number:    6,
+		Migration: populateBirthdayBlock,
 	},
 }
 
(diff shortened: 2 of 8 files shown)


wallet/chainntfns.go
```diff
@@ -329,3 +344,151 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
+	header, err := chainClient.GetBlockHeader(&candidate.Hash)
+	if err != nil {
+		return fmt.Errorf("unable to get header for block hash %v: %v",
+			candidate.Hash, err)
+	}
+
+	// We'll go back a day worth of blocks in the chain until we find a
+	// block whose timestamp is below our birthday timestamp.
+	heightDelta := int32(144)
+	for birthdayTimestamp.Before(header.Timestamp) {
```
halseth: Since header timestamps are not necessarily increasing, we must be carful here. I'm not sure if there's a faster way of finding the first block with a timestamp after our birthday without scanning all blocks, but I think we can make a decent estimate.

Since a block's timestamp must be greater than the [median of the previous 11 blocks](https://en.bitcoin.it/wiki/Block_timestamp), a reasonable strategy would be to 
* find the last block with a timestamp before our birthday
* take the maximum of the previous 11 block window
* keep moving this window back until the maximum is < birthday

This lets us find the last 11 block window with all blocks before the birthday, and the next block (together with a delta, see below) should be an okay guess I think.

wallet/chainntfns.go
```diff
@@ -329,3 +344,151 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
+	// timestamp. To do so, we'll fetch the block header and check its
+	// timestamp.
+	candidate := birthdayBlock
+	header, err := chainClient.GetBlockHeader(&candidate.Hash)
+	if err != nil {
+		return fmt.Errorf("unable to get header for block hash %v: %v",
+			candidate.Hash, err)
+	}
+
+	// We'll go back a day worth of blocks in the chain until we find a
```
halseth: I think this would be a very useful utility method! `firstBlockFromTimestamp(t time.Time) BlockStamp`.

This is something we want for the initial wallet rescan as well, as we also find the block there (currently by scanning from genesis).

wallet/chainntfns.go
```diff
@@ -329,3 +344,151 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
+			header.Timestamp)
+	}
+
+	// To ensure we have a reasonable birthday block, we'll make sure it
+	// respects our birthday timestamp and it is within a reasonable delta.
+	// The birthday has already been adjusted to two days in the past of the
+	// actual birthday, so we'll make our expected delta to be within an
+	// hour of it to prevent fetching more unnecessary blocks.
+	timestampDelta := birthdayTimestamp.Sub(header.Timestamp)
+	for time.Duration(math.Abs(float64(timestampDelta))) > time.Hour {
```
halseth: Using same reference as above, maybe makes more sense to make the delta 2 hours?

wallet/wallet.go
```diff
@@ -2652,7 +2655,11 @@ func (w *Wallet) ImportPrivateKey(scope waddrmgr.KeyScope, wif *btcutil.WIF,
 			return nil
 		}
 
-		return w.Manager.SetBirthday(addrmgrNs, newBirthday)
+		err = w.Manager.SetBirthday(addrmgrNs, newBirthday)
```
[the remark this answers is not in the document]
wpaulino: Agree that this would be nice to have, but seeing how there is currently no testing logic for the wallet and this is release blocking, we should probably address this later? FWIW, `lnd` has no uses of this method.

wallet/wallet.go
```diff
@@ -657,11 +657,25 @@ func (w *Wallet) syncWithChain() error {
...
 
+	// If this was our initial sync, then a birthday stamp should have been
+	// populated, so we'll write it to disk now.
+	if birthdayStamp != nil {
+		log.Debugf("Using birthday block: height=%d, hash=%v",
+			birthdayStamp.Height, birthdayStamp.Hash)
+
+		err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+			ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+			return w.Manager.SetBirthdayBlock(ns, *birthdayStamp)
```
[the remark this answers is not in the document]
wpaulino: It's already set at the time the wallet is created. This is the only reasonable place to set the birthday block because the wallet is not aware of the chain yet (in most scenarios) when it is created. For example, in `lnd`, we create the wallet before we even establish our connection to the backend, so at this point we wouldn't be able to retrieve our birthday block.

waddrmgr/migrations.go
```diff
@@ -254,3 +260,97 @@ func migrateRecursively(src, dst walletdb.ReadWriteBucket,
...
+	}
+
+	// Now, we'll need to determine the timestamp of the genesis block for
+	// the corresponding chain.
+	genesisHash, err := fetchBlockHash(ns, 0)
+	if err != nil {
+		return fmt.Errorf("unable to fetch genesis block hash: %v", err)
+	}
+
+	var genesisTimestamp time.Time
```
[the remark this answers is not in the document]
wpaulino: Don't think so, we'd have to make this a method of the address manager, which we don't have access to.

wallet/chainntfns.go
```diff
@@ -96,6 +99,18 @@ func (w *Wallet) handleChainNotifications() {
...
 			case chain.ClientConnected:
+				// Before attempting to sync with our backend,
+				// we'll make sure that our birthday block has
+				// been set correctly to potentially prevent
+				// missing relevant events.
+				if err := w.birthdaySanityCheck(); err != nil {
+					log.Errorf("Unable to sanity check "+
+						"wallet birthday block: %v", err)
+
+					// TODO(wilmer): return here to prevent
```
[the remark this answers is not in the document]
wpaulino: Fixed.

wallet/chainntfns.go
wpaulino: Not sure if all this additional complexity is warranted for since the birthday timestamp is set to two days in the past from when it actually is. Also, `birthdaySanityCheck` only serves useful for existing wallets that have gone through the birthday block migration as the assumption used within it (1 block/10 mins) is most likely to be inaccurate. For new wallets, the correct block will be populated upon syncing with the chain.

wallet/chainntfns.go
wpaulino: I agree this would be useful, but perhaps might be out of scope considering the birthday block is now being set within `syncWithChain`, which we're planning to refactor in the future anyway.

wallet/chainntfns.go
wpaulino: Agreed - fixed!

wpaulino:
```
--- FAIL: TestManagerHigherVersion (0.00s)
	common_test.go:289: Failed to create Manager: failed to master public key: unable to decrypt
```

Not sure what's causing this yet and seems to be flaky.

wallet/wallet.go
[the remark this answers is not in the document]
halseth: Yeah, I can agree on the test.

But the outlined scenario should be fixed, the new birthday block must be set if it is less than the existing birtdayblock.

wallet/chainntfns.go
halseth: Shouldn't be much work extracting into a method though?

wallet/chainntfns.go
```diff
@@ -329,3 +341,152 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
+	}
+
+	// To ensure we have a reasonable birthday block, we'll make sure it
+	// respects our birthday timestamp and it is within a reasonable delta.
+	// The birthday has already been adjusted to two days in the past of the
+	// actual birthday, so we'll make our expected delta to be within two
+	// hours of it to account for the network-adjusted time and prevent
+	// fetching more unnecessary blocks.
+	timestampDelta := birthdayTimestamp.Sub(header.Timestamp)
+	for timestampDelta > 2*time.Hour {
```
halseth: I don't think this loop is worth it. At most we can gain 144 blocks, which doesn't seem too much extra to scan.

waddrmgr/migrations.go
```diff
@@ -254,3 +260,91 @@ func migrateRecursively(src, dst walletdb.ReadWriteBucket,
...
+			chaincfg.SimNetParams.GenesisBlock.Header.Timestamp
+
+	default:
+		return fmt.Errorf("unknown genesis hash %v", genesisHash)
+	}
+
+	// With the timestamps retrieved, we can estimate a block height by
+	// taking the difference between them and dividing by the average block
+	// time (10 minutes).
+	birthdayHeight := int32((birthdayTimestamp.Sub(genesisTimestamp).Seconds() / 600))
```
halseth: On testnet I can imagine this being pretty far off.

waddrmgr/migrations.go
```diff
@@ -254,3 +260,91 @@ func migrateRecursively(src, dst walletdb.ReadWriteBucket,
...
+// populateBirthdayBlock is a migration that attempts to populate the birthday
+// block of the wallet. This is needed so that in the event that we need to
+// perform a rescan of the wallet, we can do so starting from this block, rather
+// than from the genesis block.
+//
+// NOTE: This migration cannot guarantee the correctness of the birthday block
+// being set as we do not store block timestamps, so a sanity check must be done
+// upon starting the wallet to ensure we do not potentially miss any relevant
+// events when rescanning.
+func populateBirthdayBlock(ns walletdb.ReadWriteBucket) error {
```
halseth: with the sanity check, is this actually needed? If we set it to the current block, then the sanity check should decrease it down to the actual block.

wallet/chainntfns.go
wpaulino: It is needed due to the height estimate of the migration potentially being wrong. On mainnet this isn't such an issue, but on testnet, like you pointed out, the estimate is much farther back than it should be, so the sanity check should be able to detect this and look for a more reasonable value.

waddrmgr/migrations.go
wpaulino: Yep, see my comment above regarding this.

waddrmgr/migrations.go
wpaulino: Yeah the migration is still needed because the sanity check can only run once the wallet has been initialized. Since we'll be making another migration to force a rescan from the birthday block, these migrations happen before the wallet's been initialized and will most likely happen back-to-back for most users, so we need to make sure the birthday block exists before the following migration.

wallet/chainntfns.go
```diff
@@ -96,6 +98,16 @@ func (w *Wallet) handleChainNotifications() {
 			var err error
 			switch n := n.(type) {
 			case chain.ClientConnected:
+				// Before attempting to sync with our backend,
+				// we'll make sure that our birthday block has
+				// been set correctly to potentially prevent
+				// missing relevant events.
+				if err := w.birthdaySanityCheck(); err != nil {
+					log.Errorf("Unable to sanity check "+
```
Roasbeef: We may actually want to panic here. As is, the loop will exit, and we'll silently stop receiving any chain notifications.

wallet/chainntfns.go
Roasbeef: Let's opt for a minimal diff in the PR, when we have a use for the method elsewhere in the code base it can be extracted.

wallet/chainntfns.go
Roasbeef: As the goal is to minimize the rescan duration (since users _really_ hated the migration time from 0.5) logic that attempts to bridge the gap is warranted IMO. I think with real world tests, we'll be able to get a feel for how off the estimates are on mainnet/testnet. At that point, we can make a decisions w.r.t the fate of this fragment.

wallet/wallet.go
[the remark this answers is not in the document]

wallet/wallet.go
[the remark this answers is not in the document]
wpaulino: > Why not set this when the birthday block is discovered? Right now it seems we may never set it if we crash after finding it but before finishing the initial sync.

The birthday block can be modified later on by a rollback, which is why I opted for doing it here instead. Should we just also persist it once we find it as well?

> It should also be loaded at the beginning of this method in case it was written in a prior attempt

Why does this need to be done though?

wallet/chainntfns.go
wpaulino: They're also slightly different FWIW. At the moment, `syncWithChain` will find the first block with a timestamp greater than our birthday timestamp, while the sanity check is more just about ensuring that we have a sensible value.

wallet/chainntfns.go
wpaulino: Correct. On mainnet, the height estimate was off by about 30K blocks in the past, which is still a pretty big improve over the genesis block IMO. On testnet, it was ~1 million.

wallet/chainntfns.go
wpaulino: Fixed.

waddrmgr/sync.go
```diff
@@ -110,3 +110,17 @@ func (m *Manager) SetBirthday(ns walletdb.ReadWriteBucket,
 	m.birthday = birthday
 	return putBirthday(ns, birthday)
 }
+
+// BirthdayBlock returns the birthday block, or earlist block a key could have
```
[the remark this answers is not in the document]
wpaulino: Fixed.

waddrmgr/sync.go
```diff
@@ -110,3 +110,17 @@ func (m *Manager) SetBirthday(ns walletdb.ReadWriteBucket,
...
+// been used, for the manager.
+func (m *Manager) BirthdayBlock(ns walletdb.ReadBucket) (BlockStamp, error) {
+	return fetchBirthdayBlock(ns)
+}
+
+// SetBirthdayBlock sets the birthday block, or earlist time a key could have
```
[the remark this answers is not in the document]
wpaulino: Fixed.

waddrmgr/db.go
```diff
@@ -1929,32 +1930,67 @@ func putStartBlock(ns walletdb.ReadWriteBucket, bs *BlockStamp) error {
...
-	buf := make([]byte, 8)
-	binary.BigEndian.PutUint64(buf, uint64(t.Unix()))
+	return nil
+}
 
-	err := bucket.Put(birthdayName, buf)
-	if err != nil {
-		str := "failed to store birthday"
+// fetchBirthdayBlock retrieves the birthday block from the database.
+func fetchBirthdayBlock(ns walletdb.ReadBucket) (BlockStamp, error) {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

waddrmgr/db.go
```diff
@@ -1876,6 +1877,10 @@ func fetchBlockHash(ns walletdb.ReadBucket, height int32) (*chainhash.Hash, erro
 	heightBytes := make([]byte, 4)
 	binary.BigEndian.PutUint32(heightBytes, uint32(height))
 	hashBytes := bucket.Get(heightBytes)
+	if hashBytes == nil {
+		err := errors.New("block not found")
```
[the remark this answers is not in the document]
wpaulino: Fixed.

wallet/wallet.go
```diff
@@ -2648,11 +2651,19 @@ func (w *Wallet) ImportPrivateKey(scope waddrmgr.KeyScope, wif *btcutil.WIF,
 		// We'll only update our birthday with the new one if it is
 		// before our current one. Otherwise, we won't rescan for
 		// potentially relevant chain events that occurred between them.
-		if newBirthday.After(w.Manager.Birthday()) {
+		birthdayBlock, err := w.Manager.BirthdayBlock(addrmgrNs)
+		if err != nil {
+			return err
+		}
+		if birthdayBlock.Height >= bs.Height {
```
halseth: Should be reversed?

wallet/chainntfns.go
halseth: Ah, I see. I was assuming the candidate started out `>` birthday before the first loop.

waddrmgr/migrations.go
halseth: I might be missing something, but since we must handle the case where the estimate being way off (testnet), then shouldn't we be able to just set the birthday to the latest block we know, and let the sanity check "fix" this estimate?

wallet/wallet.go
wpaulino: Yep, fixed.

waddrmgr/migrations.go
wpaulino: In the case of consecutive migrations that depend on this one, the sanity check won't be done until the wallet is initialized. Since we also have a rescan migration upcoming, the birthday block just needs to be a more sensible value than the genesis block. We'll be updating the `btcwallet` dependency within `lnd` that contains all of these migrations as part of one commit, so users will experience the migrations one after the other within the same database transaction before the wallet is initialized.

wallet/wallet.go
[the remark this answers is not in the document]
Roasbeef: This should be moved above to when we discover the birthday, such that, it's written in the very same transaction as the call to `SyncedTo` directly after we discover the birthday. Otherwise, we can write the `SyncedTo` as the birthday height, then crash _right_ after the call to `Commit()`. If this happens, then we'll come back up again, and note the birthday stamp to be nil, and potentially write (at the bottom) an incorrect birthday height _right_ after the _actual_ birthday height. 

This won't be caught by the sanity check as no height has been written yet.

wallet/chainntfns.go
```diff
@@ -329,3 +343,161 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
+	}
+
+	// Otherwise, we have a new, better candidate, so we'll write it to
+	// disk.
+	log.Debugf("Found a new valid wallet birthday block: height=%d, hash=%v",
+		candidate.Height, candidate.Hash)
+
+	return walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+		ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+		return w.Manager.SetBirthdayBlock(ns, candidate)
```
Roasbeef: We can also set `SyncedTo` here in order to avoid a low recan from a poor estimate. On testnet for example, I had wallets start from height 380k or so.

wpaulino: Fixed.

wallet/wallet.go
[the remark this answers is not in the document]
wpaulino: Fixed. Let me know if there's anything I missed.

waddrmgr/migrations.go
```diff
@@ -254,3 +260,93 @@ func migrateRecursively(src, dst walletdb.ReadWriteBucket,
...
+	}
+
+	// With the timestamps retrieved, we can estimate a block height by
+	// taking the difference between them and dividing by the average block
+	// time (10 minutes).
+	birthdayHeight := int32((birthdayTimestamp.Sub(genesisTimestamp).Seconds() / 600))
+
+	// Now that we have the height estimate, we can fetch the corresponding
+	// block and set it as our birthday block.
+	birthdayHash, err := fetchBlockHash(ns, birthdayHeight)
```
halseth: https://github.com/btcsuite/btcwallet/pull/555 removes old block hashes, so should consider if this change make them incompatible.

wpaulino: That hasn't been merged yet, so this migration should be fine. Ideally, the way to remove the older block hashes should be done through a migration as well that follows this one.

PR #574, opened 2018-11-09
bbolt git is now owned by etcd-io and component named bbolt

jadeblaquiere:
This simply realigns to the current HEAD of bbolt.


Roasbeef:
Fixed by #578

PR #576, opened 2018-11-19
wallet/chainntnfs: set height for new birthday block candidate

wpaulino:
In this commit, we address an issue that would cause users to be stuck
in an infinite loop by fetching the same candidate birthday block due to
its height not being updated if the sanity check was attempting to fix
an estimate in the future. We fix this by setting the new candidate
height so that new candidate blocks can be fetched and tested.

### DIFF
--- a/wallet/chainntfns.go
+++ b/wallet/chainntfns.go
@@ -118,7 +118,7 @@ func (w *Wallet) handleChainNotifications() {
 				birthdayBlock, err := birthdaySanityCheck(
 					chainClient, birthdayStore,
 				)
-				if err != nil {
+				if err != nil && !waddrmgr.IsError(err, waddrmgr.ErrBirthdayBlockNotSet) {
 					err := fmt.Errorf("unable to sanity "+
 						"check wallet birthday block: %v",
 						err)
@@ -450,20 +450,15 @@ func (s *walletBirthdayStore) SetBirthdayBlock(block waddrmgr.BlockStamp) error 
 // with the backend, but before it begins syncing. This is done as the second
 // part to the wallet's address manager migration where we populate the birthday
 // block to ensure we do not miss any relevant events throughout rescans.
+// waddrmgr.ErrBirthdayBlockNotSet is returned if the birthday block has not
+// been set yet.
 func birthdaySanityCheck(chainConn chainConn,
 	birthdayStore birthdayStore) (*waddrmgr.BlockStamp, error) {
 
 	// We'll start by fetching our wallet's birthday timestamp and block.
 	birthdayTimestamp := birthdayStore.Birthday()
 	birthdayBlock, birthdayBlockVerified, err := birthdayStore.BirthdayBlock()
-	switch {
-	// If our wallet's birthday block has not been set yet, then this is our
-	// initial sync, so we'll defer setting it until then.
-	case waddrmgr.IsError(err, waddrmgr.ErrBirthdayBlockNotSet):
-		return nil, nil
-
-	// Otherwise, we'll return the error if there was one.
-	case err != nil:
+	if err != nil {
 		return nil, err
 	}
 
--- a/wallet/chainntfns.go
+++ b/wallet/chainntfns.go
@@ -10,11 +10,20 @@ import (
 	"strings"
 	"time"
 
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcwallet/chain"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/wtxmgr"
+)
+
+const (
+	// birthdayBlockDelta is the maximum time delta allowed between our
+	// birthday timestamp and our birthday block's timestamp when searching
+	// for a better birthday block candidate (if possible).
+	birthdayBlockDelta = 2 * time.Hour
 )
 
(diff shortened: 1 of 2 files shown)


wallet/chainntfns.go
```diff
@@ -442,6 +442,7 @@ func (w *Wallet) birthdaySanityCheck() (*waddrmgr.BlockStamp, error) {
 		}
 
 		candidate.Hash = *hash
+		candidate.Height = int32(newCandidateHeight)
```
halseth: can we add a test for this case?

Roasbeef: Agreed, we've caught two issues in this method that were lingering due to a lack of tests in this method. We can add a set of _minimal_ interfaces, so interfaces that contain _only_ the set of methods we need in order to minimize the foot print here.

wpaulino: Tests added!

wallet/chainntfns.go
```diff
@@ -345,30 +356,65 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
+	// We'll start by fetching our wallet's birthday timestamp and block.
 	var (
-		birthdayTimestamp     = w.Manager.Birthday()
+		birthdayTimestamp     = syncStore.Birthday()
 		birthdayBlock         waddrmgr.BlockStamp
 		birthdayBlockVerified bool
 	)
-	err = walletdb.View(w.db, func(tx walletdb.ReadTx) error {
+
+	err := walletdb.View(db, func(tx walletdb.ReadTx) error {
```
Roasbeef: Why not throw this into an interface above as well? No need to create an entire mock wallet db in this case to test the function.

wallet/chainntfns.go
```diff
@@ -345,30 +356,65 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
+}
+
+// syncStore is an interface that abstracts the wallet's sync-related
+// information required to perform a birthday block sanity check.
+type syncStore interface {
+	// Birthday returns the birthday timestamp of the wallet.
+	Birthday() time.Time
+
+	// BirthdayBlock returns the birthday block of the wallet.
+	BirthdayBlock(walletdb.ReadBucket) (waddrmgr.BlockStamp, bool, error)
```
Roasbeef: I think we can further abstract over the `walletbdb.ReadBucket` instances. The concrete implementation can make the transaction outside this function, hiding even more details.

wallet/chainntfns.go
```diff
@@ -524,13 +570,13 @@ func (w *Wallet) birthdaySanityCheck() (*waddrmgr.BlockStamp, error) {
 	log.Debugf("Found a new valid wallet birthday block: height=%d, hash=%v",
 		candidate.Height, candidate.Hash)
 
-	err = walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+	err = walletdb.Update(db, func(tx walletdb.ReadWriteTx) error {
```
Roasbeef: See comment above re abstracting even further.

wallet/chainntfns_test.go
```diff
@@ -0,0 +1,346 @@
...
+func createTestDB(t *testing.T) (walletdb.DB, func()) {
+	t.Helper()
+
+	tempDir, err := ioutil.TempDir("", "walletdb")
+	if err != nil {
+		t.Fatalf("unable to create temp dir: %v", err)
+	}
+
+	dbPath := filepath.Join(tempDir, "wallet.db")
+	db, err := walletdb.Create("bdb", dbPath)
```
Roasbeef: Don't need to create a database at all. See comments on the prior commit.

wallet/chainntfns_test.go
```diff
@@ -0,0 +1,346 @@
...
+			Hash:   chainhash.Hash{},
+			Height: 1337,
+		},
+		birthdayBlockVerified: true,
+		syncedTo: waddrmgr.BlockStamp{
+			Height: chainTip,
+		},
+	}
+
+	if _, err := birthdaySanityCheck(db, chainConn, syncStore); err != nil {
```
Roasbeef: Should check that the block is returned since we expect to pass it up to the caller.

Roasbeef:
Verified that this diff causes the tests to hang:
```
diff --git a/wallet/chainntfns.go b/wallet/chainntfns.go
index a9f9cd3..a923cae 100644
--- a/wallet/chainntfns.go
+++ b/wallet/chainntfns.go
@@ -488,7 +488,6 @@ func birthdaySanityCheck(db walletdb.DB, chainConn chainConn,
 		}
 
 		candidate.Hash = *hash
-		candidate.Height = int32(newCandidateHeight)
 		candidate.Timestamp = header.Timestamp
 
 		log.Debugf("Checking next birthday block candidate: "+
```

Roasbeef:
Changes look good to me, will wait to give @halseth a chance for another path.

wallet/chainntfns_test.go
```diff
@@ -0,0 +1,300 @@
...
+func TestBirthdaySanityCheckEmptyBirthdayBlock(t *testing.T) {
+	t.Parallel()
+
+	chainConn := &mockChainConn{}
+
+	// Our birthday store will reflect that we don't have a birthday block
+	// set, so we should not attempt a sanity check.
+	birthdayStore := &mockBirthdayStore{}
+
+	birthdayBlock, err := birthdaySanityCheck(chainConn, birthdayStore)
```
halseth: Should we rather make this method return `ErrBirthdayBlockNotSet`?

wpaulino: Fixed.

PR #577, opened 2018-11-26
wallet/wallet: redefine initial sync to birthday block not being set

wpaulino:
In this commit, we modify our initial sync assumption to depend on
whether the wallet's birthday block has been set. We can use this
assumption to prevent costly rescans from the birthday block upon
restarts if there are no UTXOs that exist within the wallet.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -855,6 +855,130 @@ func (w *Wallet) syncToBirthday() (*waddrmgr.BlockStamp, error) {
 	return birthdayStamp, nil
 }
 
+// recovery attempts to recover any unspent outputs that pay to any of our
+// addresses starting from the specified height.
+//
+// NOTE: The starting height must be at least the height of the wallet's
+// birthday or later.
+func (w *Wallet) recovery(startHeight int32) error {
+	log.Infof("RECOVERY MODE ENABLED -- rescanning for used addresses "+
+		"with recovery_window=%d", w.recoveryWindow)
+
+	// We'll initialize the recovery manager with a default batch size of
+	// 2000.
+	recoveryMgr := NewRecoveryManager(
+		w.recoveryWindow, recoveryBatchSize, w.chainParams,
+	)
+
+	// In the event that this recovery is being resumed, we will need to
+	// repopulate all found addresses from the database. For basic recovery,
+	// we will only do so for the default scopes.
+	scopedMgrs, err := w.defaultScopeManagers()
+	if err != nil {
+		return err
+	}
+	tx, err := w.db.BeginReadWriteTx()
+	if err != nil {
+		return err
+	}
+	txMgrNS := tx.ReadBucket(wtxmgrNamespaceKey)
+	credits, err := w.TxStore.UnspentOutputs(txMgrNS)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+	addrMgrNS := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+	err = recoveryMgr.Resurrect(addrMgrNS, scopedMgrs, credits)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	// We'll also retrieve our chain backend client in order to filter the
+	// blocks as we go.
+	chainClient, err := w.requireChainClient()
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	// We'll begin scanning the chain from the specified starting height.
+	// Since we assume that the lowest height we start with will at least be
+	// that of our birthday, we can just add every block we process from
+	// this point forward to the recovery batch.
+	err = w.scanChain(startHeight, func(height int32,
+		hash *chainhash.Hash, header *wire.BlockHeader) error {
+
+		recoveryMgr.AddToBlockBatch(hash, height, header.Timestamp)
+
+		// We'll checkpoint our current batch every 2K blocks, so we'll
+		// need to start a new database transaction. If our current
+		// batch is empty, then this will act as a NOP.
+		if height%recoveryBatchSize == 0 {
+			blockBatch := recoveryMgr.BlockBatch()
+			err := w.recoverDefaultScopes(
+				chainClient, tx, addrMgrNS, blockBatch,
+				recoveryMgr.State(),
+			)
+			if err != nil {
+				return err
+			}
+
+			// Clear the batch of all processed blocks.
+			recoveryMgr.ResetBlockBatch()
+
+			if err := tx.Commit(); err != nil {
+				return err
+			}
+
+			log.Infof("Recovered addresses from blocks %d-%d",
+				blockBatch[0].Height,
+				blockBatch[len(blockBatch)-1].Height)
+
+			tx, err = w.db.BeginReadWriteTx()
+			if err != nil {
+				return err
+			}
+			addrMgrNS = tx.ReadWriteBucket(waddrmgrNamespaceKey)
+		}
+
+		// Since the recovery in a way acts as a rescan, we'll update
+		// the wallet's tip to point to the current block so that we
+		// don't unnecessarily rescan the same block again later on.
+		return w.Manager.SetSyncedTo(addrMgrNS, &waddrmgr.BlockStamp{
+			Hash:      *hash,
+			Height:    height,
+			Timestamp: header.Timestamp,
+		})
+	})
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	// Now that we've reached the chain tip, we can process our final batch
+	// with the remaining blocks if it did not reach its maximum size.
+	blockBatch := recoveryMgr.BlockBatch()
+	err = w.recoverDefaultScopes(
+		chainClient, tx, addrMgrNS, blockBatch, recoveryMgr.State(),
+	)
+	if err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	// With the recovery complete, we can persist our new state and exit.
+	if err := tx.Commit(); err != nil {
+		tx.Rollback()
+		return err
+	}
+
+	log.Infof("Recovered addresses from blocks %d-%d", blockBatch[0].Height,
+		blockBatch[len(blockBatch)-1].Height)
+
+	return nil
+}
+
 // defaultScopeManagers fetches the ScopedKeyManagers from the wallet using the
 // default set of key scopes.
 func (w *Wallet) defaultScopeManagers() (
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -327,274 +327,27 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
 		return err
 	}
 
-	// Request notifications for transactions sending to all wallet
-	// addresses.
-	var (
-		addrs   []btcutil.Address
-		unspent []wtxmgr.Credit
-	)
-	err = walletdb.View(w.db, func(dbtx walletdb.ReadTx) error {
+	// To start, if we've yet to find our birthday stamp, we'll do so now.
+	if birthdayStamp == nil {
 		var err error
-		addrs, unspent, err = w.activeData(dbtx)
-		return err
-	})
-	if err != nil {
-		return err
+		birthdayStamp, err = w.syncToBirthday()
+		if err != nil {
+			return err
+		}
 	}
 
-	startHeight := w.Manager.SyncedTo().Height
-
-	// We'll mark this as our first sync if we don't have any unspent
-	// outputs as known by the wallet. This'll allow us to skip a full
-	// rescan at this height, and instead wait for the backend to catch up.
-	isInitialSync := len(unspent) == 0
-
-	isRecovery := w.recoveryWindow > 0
-	birthday := w.Manager.Birthday()
-
-	// TODO(jrick): How should this handle a synced height earlier than
-	// the chain server best block?
-
-	// When no addresses have been generated for the wallet, the rescan can
-	// be skipped.
-	//
-	// TODO: This is only correct because activeData above returns all
-	// addresses ever created, including those that don't need to be watched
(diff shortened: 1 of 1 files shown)


roeierez:
@wpaulino thanks! and following our discussion here https://github.com/lightningnetwork/lnd/issues/2215, I tested and it solves the problem.
One bug, that is not introduces by this commit, is that if the first run during "initialSync" has passed the birthday and was stopped before starting the rescan, then on the next start blocks in the range of the birthday to the best block known will be missed.  
I think a good way to solve it is to better define the initialSync to be syncToBirthday and then trigger a rescan always from the known best block.
Here is a diff that is meant to show what I mean and hope can help in shaping the solution:
https://github.com/roeierez/btcwallet/commit/bfa2e60728fc7e7bef1af112ec7dfa10bbbea1ef

wpaulino:
@roeierez agree that this is a good approach to go forward with! I think it merits its own PR though as this one aims to solve a different problem. Let me know if you'd like to propose the change yourself, otherwise I can take it over.

One thing that jumps out to me in your diff is that `syncWithChain` should be completely refactored to only sync until the birthday and not go past that since there's logic to catch up to the tip of the chain while the rescan is ongoing as well.
https://github.com/btcsuite/btcwallet/blob/e59e51f8e13c9e552686ef785edae838ca08ebd2/wallet/chainntfns.go#L166

roeierez:
> Let me know if you'd like to propose the change yourself, otherwise I can take it over.

I will be happy to introduce a new PR for this and we can take it from there.


> `syncWithChain` should be completely refactored to only sync until the birthday and not go past that

Agree! this is also the reason why I suggested rename to "syncToBirthday".

@wpaulino  If you are fine with this, I am going to start working on a proposed solution.

Roasbeef:
> then on the next start, blocks in the range between the birthday to the best block known will be missed.

By missed, do you mean missed by the rescan, or that the "initial catch up loop" will miss those blocks? If it's the latter, then this would only be an issue in the case of deep re-orgs.

roeierez:
> By missed, do you mean missed by the rescan, or that the "initial catch up loop" will miss those blocks?

I mean missing  by the rescan, in that case it will only rescan from tip as I see here: https://github.com/btcsuite/btcwallet/blob/master/wallet/wallet.go#L696

Roasbeef:
"Tip" in that case is the [wallet's synced to height](https://github.com/btcsuite/btcwallet/blob/e59e51f8e13c9e552686ef785edae838ca08ebd2/wallet/rescan.go#L272), so if we restart _right_ after the wallet's birthday has been passed, then we'll being from that point onwards.

roeierez:
I understand that and I agree that If the restart happened right after the wallet's birthday then we indeed don't have a problem. I was referring to the case where there is a gap between the birthday stamp and the point where the "initial sync" has reached. 
if the birthday is block N then on regular start (without stopping) the rescan will indeed start from block N (as I see here: https://github.com/btcsuite/btcwallet/blob/master/wallet/wallet.go#L693).
But if the first start the "initial sync" has reached "N + 100", resulting in updating the `Manager.SyncedTo()` to N + 100,   and then restarted the rescan will start from "N + 100" skipping 100 blocks (as I see here: https://github.com/btcsuite/btcwallet/blob/master/wallet/wallet.go#L696)
Do you think I am missing something?

Roasbeef:
Ahh ok, that's for that example, I get where you're describing here now.

We might as well fix the bug since this commit makes it a bit more apparent. If you throw up a commit somewhere fixing it (possibly with that refactor mentioned above, as refactoring in this section of the code is looong over due), we can cherry pick it into this PR as we can test both of them as a hole. Thanks for pointing this out!

roeierez:
@wpaulino as @Roasbeef has asked I pushed my changes here: https://github.com/btcsuite/btcwallet/compare/master...roeierez:sync
I made several commits so you can see the changes clearly enough as the last commit extract out only the initial sync logic to its own function.
The initial sync now stops at the birthday stamp and one thing I wasn't sure about is whether it holds fine also for the recovery mode. I think it is because the code triggers a rescan later from the updated tip but I figured it is good to point to mention here.
I still didn't tested it properly as first I would like to know if this is what you had in mind.
Any kind feedback is very welcome and appreciated.

wpaulino:
@roeierez thanks for getting this started! We plan to get this in ASAP for our next release, so I'll be covering the rest from here. I made sure to provide you with credit on the relevant commits. The only major thing missing from your branch was modifying the recovery logic to maintain its current behavior (add blocks to the recovery manager from the birthday to the tip of the chain).

@Roasbeef this should be ready for review now.

wallet/wallet.go
```diff
@@ -322,379 +322,414 @@ func (w *Wallet) activeData(dbtx walletdb.ReadTx) ([]btcutil.Address, []wtxmgr.C
...
+		return bestHeight >= latestCheckptHeight
+	}
 
-		log.Infof("Catching up block hashes to height %d, this will "+
-			"take a while...", logHeight)
+	for height := w.Manager.SyncedTo().Height; birthdayStamp == nil; height++ {
+		// If we've reached our best height and we're not current, we'll
+		// wait for blocks at tip to ensure we find a valid birthday
+		// stamp.
+		for height == bestHeight && !isCurrent(bestHeight) {
```
roeierez: @wpaulino wanted to point out that I am not sure the "!isCurrent(bestHeight)" is needed in the condition. This is because at this point the birthday stamp is null and we are certain that we are not "current".

Roasbeef: It's not meant to check if we're current, but if the chain backend is current or not. This is how it works in the current diff at least, since it may be the case that the chain backend is still syncing while `lnd` is connected to it.

wallet/wallet.go
```diff
@@ -697,6 +697,147 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
...
+
+			tx, err = w.db.BeginReadWriteTx()
+			if err != nil {
+				return nil, err
+			}
+			ns = tx.ReadWriteBucket(waddrmgrNamespaceKey)
+		}
+	}
+
+	// If a birthday stamp has yet to be found, we'll return an error
```
Roasbeef: Shouldn't this never happen given the birthday sanity check? In general this commit would be easier to review it if were in two parts: a move, then modification. I think we're past that point in this PR though, thanks for getting it started!

wallet/wallet.go
```diff
@@ -838,6 +838,148 @@ func (w *Wallet) syncToBirthday() (*waddrmgr.BlockStamp, error) {
 	return birthdayStamp, nil
 }
 
+// recovery attempts to recover any unspent outputs that pay to any of our
+// addresses. The birthdayStamp is used to indicate the starting point of our
+// recovery, as it's not possible for us to create any addresses before our
+// wallet's birthday.
+func (w *Wallet) recovery(birthdayStamp *waddrmgr.BlockStamp) error {
```
Roasbeef: At a glance a large degree of code duplication exists in this method and the one introduced in the prior commit.

wallet/wallet.go
```diff
@@ -838,6 +838,148 @@ func (w *Wallet) syncToBirthday() (*waddrmgr.BlockStamp, error) {
...
+	// chain backend is currently synced.
+	isCurrent := func(bestHeight int32) bool {
+		switch c := chainClient.(type) {
+		case *chain.NeutrinoClient:
+			return c.CS.IsCurrent()
+		}
+		return bestHeight >= latestCheckptHeight
+	}
+
+	for height := birthdayStamp.Height; height <= bestHeight; height++ {
```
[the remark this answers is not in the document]
Roasbeef: 👍

wallet/wallet.go
```diff
@@ -838,6 +838,148 @@ func (w *Wallet) syncToBirthday() (*waddrmgr.BlockStamp, error) {
...
+
+			tx, err = w.db.BeginReadWriteTx()
+			if err != nil {
+				return err
+			}
+			ns = tx.ReadWriteBucket(waddrmgrNamespaceKey)
+		}
+	}
+
+	// Now that we've reached the chain tip, we can process our final batch
```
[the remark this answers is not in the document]
Roasbeef: Isn't the assumption that we first do the birthday sync, _then_ recover if needed, and finally sync to the regular chain? The downside of this split up, is that we'll now scan the chain entirely twice, vs doing it once. Additionally, each time we restart the recovery logic will need to start from scratch, rather than pick up where it left off as it does now.

wallet/wallet.go
```diff
@@ -379,48 +380,45 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
...
+		}
+
+		// Otherwise, we'll mark this as our new synced height.
+		err := w.Manager.SetSyncedTo(addrmgrNs, &rollbackStamp)
+		if err != nil {
+			return err
+		}
+
+		// If the rollback happened to go beyond our birthday stamp,
+		// we'll need to find a new one by syncing with the chain again
```
Roasbeef: Why do we need to rescan to find another birthday in this case? I can see how the prior logic of using that new roll back height as the birthday isn't very precise, but it avoids this duplicated scanning.

wallet/wallet.go
roeierez: > It's not meant to check if we're current, but if the chain backend is current or not

This is why I was suggesting to remove this condition. as long as "height == bestHeight" we can assume the chain backend is not current since at this point of the code the birthday stamp is still null. Also think it is better to move this conditional waiting block to the end of the for loop. First process the block then wait.

wallet/wallet.go
wpaulino: > Shouldn't this never happen given the birthday sanity check?

In an initial sync case, the sanity check won't run.

> In general this commit would be easier to review it if were in two parts: a move, then modification.

I thought of doing that first, but it wouldn't just resulted in intermediate changes that were going to be removed anyway, so I went with this approach for the sake of simplicity.

wallet/wallet.go
wpaulino: Fixed, thanks @halseth for the suggestion!

wallet/wallet.go
wpaulino: I was thinking of the case where a deep reorg puts us significantly behind our birthday, but as these are not common, perhaps it's not worth worrying about. Ended up removing it.

wallet/wallet.go
```diff
@@ -838,6 +838,148 @@ func (w *Wallet) syncToBirthday() (*waddrmgr.BlockStamp, error) {
...
+			}
+
+			tx, err = w.db.BeginReadWriteTx()
+			if err != nil {
+				return err
+			}
+			ns = tx.ReadWriteBucket(waddrmgrNamespaceKey)
+		}
+	}
+
```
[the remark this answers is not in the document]
wpaulino: Don't think this is really needed as we'll perform a rescan at the end of `syncWithChain` which will end up catching up to any blocks we missed.

wallet/wallet.go
```diff
@@ -379,48 +380,45 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
...
 			if err != nil {
 				return err
 			}
 		}
-		return nil
+
+		// Finally, we'll rollback our transaction store to reflect the
+		// stale state. Rollback unconfirms transactions at and beyond
+		// the passed height, so add one to the new synced-to height to
+		// prevent unconfirming transactions from the synced-to block.
```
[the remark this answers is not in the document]

wallet/wallet.go
```diff
@@ -697,6 +697,147 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
...
+		"while...", logHeight)
+
+	// isCurrent is a helper function that we'll use to determine if the
+	// chain backend is currently synced.
+	isCurrent := func(bestHeight int32) bool {
+		switch c := chainClient.(type) {
+		case *chain.NeutrinoClient:
+			return c.CS.IsCurrent()
+		}
+		return bestHeight >= latestCheckptHeight
```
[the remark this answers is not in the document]
wpaulino: It definitely doesn't, but this wasn't introduced by my changes.

wallet/wallet.go
```diff
@@ -697,6 +697,147 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
...
+		}
+
+		// Checkpoint our state every 10K blocks.
+		if height%10000 == 0 {
+			if err := tx.Commit(); err != nil {
+				tx.Rollback()
+				return nil, err
+			}
+
+			tx, err = w.db.BeginReadWriteTx()
```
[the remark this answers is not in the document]
halseth: Usually scope issues like this could be solved by not capturing the `tx` variable when it is first created:
```go
tx, err := w.db.BeginReadWriteTx()
if err != nil {
	return nil, err
}
defer func() {
        tx.Rollback()
}()
```
but I think maybe that's even more confusing in this case, and we should keep it as is.

Roasbeef:
> Also possible to add a test for this code?

The tests for this area more or less reside within `lnd` and indirectly neutrino. The `lnwallet` integration tests touch much of this area, though it lives outsides of `btcwallet` (also lnd's integration tests). Generally it's something we need to improve and would be larger undertaking as it's also a major blocker in more aggressively refactoring `btcwallet` in general.

wallet/wallet.go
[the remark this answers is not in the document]
wpaulino: We still can't defer the `Rollback` since we're within the function closure, which is called on and terminates after every block.

Roasbeef:
I think this is ready to, land, final Q seems to be whether to defer the roll back or not? @wpaulino @halseth ?

wpaulino:
@Roasbeef don't think there's a way of having the `defer` there - see https://github.com/btcsuite/btcwallet/pull/577#discussion_r247771661.

Roasbeef:
Alrighty, landing then! Thanks to @roeierez for getting this fix started!

PR #582, opened 2018-12-04
 chain/bitcoind_client: extend filterTx with script spend support

wpaulino:
In this PR, we extend the BitcoindClient to properly detect the spend of an arbitrary script on-chain. This is possible by looking at a transaction's input and re-deriving the PkScript, from its signature script/witness, of the output it's attempting to spend. This is now possible due to a new `btcd` version, which includes `ComputePkScript`. Upon detecting the spend, a chain.RelevantTx will be dispatched.

Depends on https://github.com/btcsuite/btcd/pull/1359.

### DIFF
--- a/chain/bitcoind_client.go
+++ b/chain/bitcoind_client.go
@@ -1205,16 +1205,47 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
 	// to determine if this transaction is somehow relevant to the caller.
 	var isRelevant bool
 
-	// We'll start by cycling through its outputs to determine if it pays to
-	// any of the currently watched addresses. If an output matches, we'll
-	// add it to our watch list.
-	for i, out := range tx.TxOut {
-		_, addrs, _, err := txscript.ExtractPkScriptAddrs(
-			out.PkScript, c.chainParams,
+	// We'll start by checking all inputs and determining whether it spends
+	// an existing outpoint or a pkScript encoded as an address in our watch
+	// list.
+	for _, txIn := range tx.TxIn {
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -19,3 +18,5 @@ require (
 	google.golang.org/genproto v0.0.0-20181127195345-31ac5d88444a // indirect
 	google.golang.org/grpc v1.16.0
 )
+
+replace github.com/btcsuite/btcd => github.com/wpaulino/btcd v0.0.0-20181206032332-6efa63bc1c66
```
Roasbeef: Comment reminder to remove once the other PR goes in.

chain/bitcoind_client.go
```diff
@@ -1205,16 +1205,47 @@ func (c *BitcoindClient) filterTx(tx *wire.MsgTx,
...
+			// Non-standard outputs can be safely skipped.
+			continue
+		}
+		if _, ok := c.watchedAddresses[addr.String()]; ok {
+			isRelevant = true
+			break
+		}
+	}
+
+	// We'll also cycle through its outputs to determine if it pays to
```
Roasbeef: Similarly, can't we skip this loop if `isRelevant` is set above?

wpaulino: Nope, as `NotifyReceived` for outpoints will also track them for spends, so we still want that logic to happen for backwards compatibility.

cfromknecht:
Do we have any tests that can exercise this new functionality?

wpaulino:
Yeah within `chainntnfs` in `lnd`. Agree that `btcwallet` in general is lacking in tests.

Issue #585, opened 2019-01-16
Wallet doesn't sync if initial rescan starts one block behind chain backend.


roeierez:
In the cases where the wallet starts and its initial rescan only need to catch up one block, RescanFinished notification is not triggered causing it not to switch into "synced" state.
Only restart or another mined block fixes this.

I bumped into this bug using LND and here is my investigation:

The syncToChain flag in LND will only be evaluated to true if the wallet “ChainSynced” function returns true (https://github.com/btcsuite/btcwallet/blob/master/wallet/wallet.go#L281)
This flag is set after first rescan has finished here: https://github.com/btcsuite/btcwallet/blob/master/wallet/chainntfns.go#L176
The problem I see is that sometimes the first rescan just doesn’t end and only when another block is mined it triggers the “RescanFinished” notification, causes the wallet to be marked as “synced”.
The racy condition is when the chain backend (I am using neutrino) is exactly one block ahead of the wallet “SyncedTo” value. In that case the wallet starts a rescan, the rescan notifies of the next block and filteredBlock here: https://github.com/lightninglabs/neutrino/blob/master/rescan.go#L722
The RescanFinished notification is triggered inside the “onFilteredBlockConnected” callback here: https://github.com/btcsuite/btcwallet/blob/master/chain/neutrino.go#L523 but only if the “s.lastProgressSent=true”. This flag is set on the onBlockConnected callback (https://github.com/btcsuite/btcwallet/blob/master/chain/neutrino.go#L593) which happens to be called AFTER the “onFilteredBlockConnected” callback, causing it not to trigger the RescanFinished notification until after the next block arrives.

Roasbeef:
We've decided to fix this issue in neutrino itself by swapping the order of the `onBlockConnected` and `onFilteredBlockConnected` calls.

PR #586, opened 2019-01-17
Fix Initial rescan when it only needs to scan one block

roeierez:
This PR is a proposed fix for https://github.com/btcsuite/btcwallet/issues/585
The fix extracts out the logic that sends "RescanFinished" notification to a function. This function is then called from both onFilteredBlockConnected and onBlockConnected.
This will remove the dependency on the notifications order triggered from the rescan and ensure the notification is sent also in the case of the issue.

Note:
I wasn't sure if the intention of the flag "lastProgressSent" is only to make sure we have passed the birthday block or whether it is indeed needed on every restart as a pre-condition for the "RescanFinished" notification.

### DIFF
--- a/chain/neutrino.go
+++ b/chain/neutrino.go
@@ -28,11 +28,12 @@ type NeutrinoClient struct {
 	// We currently support one rescan/notifiction goroutine per client
 	rescan *neutrino.Rescan
 
-	enqueueNotification chan interface{}
-	dequeueNotification chan interface{}
-	startTime           time.Time
-	lastProgressSent    bool
-	currentBlock        chan *waddrmgr.BlockStamp
+	enqueueNotification     chan interface{}
+	dequeueNotification     chan interface{}
+	startTime               time.Time
+	lastProgressSent        bool
+	lastFilteredBlockHeader *wire.BlockHeader
+	currentBlock            chan *waddrmgr.BlockStamp
 
 	quit       chan struct{}
 	rescanQuit chan struct{}
@@ -339,6 +340,7 @@ func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Addre
 	s.scanning = true
 	s.finished = false
 	s.lastProgressSent = false
+	s.lastFilteredBlockHeader = nil
 	s.isRescan = true
 
 	bestBlock, err := s.CS.BestBlock()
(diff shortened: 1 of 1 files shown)


halseth:
Awesome catch! Do you think you are able to write a unit test to illustrate the situation?

roeierez:
> Do you think you are able to write a unit test to illustrate the situation?

@halseth this is a good idea and I'd love to do that. 
In order to test the NeutrinoClient I would need to mock the underline chain service and since it is not abstracted (it is an instance of neutrino.ChainService) I don't see a way for achieving it without first implementing the abstraction.
Do you have any idea or a better approach I can follow here?

wpaulino:
> In order to test the NeutrinoClient I would need to mock the underline chain service and since it is not abstracted (it is an instance of neutrino.ChainService) I don't see a way for achieving it without first implementing the abstraction.

Agreed. I'd say let's not worry about this for now.

As for the actual fix, perhaps reordering `OnBlockConnected` and `OnFilteredBlockConnected` (as mentioned in #585) is a better solution? Is there a reason you decided to fix things this way?

roeierez:
> Is there a reason you decided to fix things this way?

I didn't see a reason to add a strict requirement on the notifications order sent from neutrino especially when the order is not expressed in their name or nature.
Going this path will also require:
1. Making sure that in chainntfns.go receiving these notifications in different order doesn't cause any problems.
2. Changing the other chainView implementations to send notifications at the same order for consistency.

BTW, since Neutrino always sending both notifications together I think maybe consolidating the two notifications into one may be a clean solution but it is a breaking change to the neutrino interface.

chain/neutrino.go
```diff
@@ -606,6 +578,45 @@ func (s *NeutrinoClient) onBlockConnected(hash *chainhash.Hash, height int32,
...
+	if current {
+		s.finished = true
+	}
+	s.clientMtx.Unlock()
+	if current {
+		select {
+		case s.enqueueNotification <- &RescanFinished{
+			Hash:   &bs.Hash,
+			Height: bs.Height,
+			Time:   s.lastFilteredBlockHeader.Timestamp,
```
wpaulino: Reading this value without the lock will cause a race condition.

wpaulino:
Looking back at this now:

> I didn't see a reason to add a strict requirement on the notifications order sent from neutrino especially when the order is not expressed by their name or nature.

There is already an implicit order in the way they'll be called because of this line: https://github.com/btcsuite/btcwallet/blob/8b90263a619068db81994688fbdf434be39a81ba/chain/neutrino.go#L522

Roasbeef:
We've decided to fix this issue in neutrino itself by swapping the order of the `onBlockConnected` and `onFilteredBlockConnected` calls.

Roasbeef:
Ultimately went with a slightly different version: https://github.com/btcsuite/btcwallet/pull/613

Issue #588, opened 2019-01-30
`glide install` [ERROR]	Failed to find glide.yaml file in directory tree: Cannot resolve parent of / with Go1.11


BobleChinois:
Hi I experienced [some trouble](https://github.com/lightningnetwork/lnd/issues/2393) with LND recently, and I was adviced to scan my wallet file using `dropwtxmgr` in btcwallet.

I'm following install instructions here, download btcwallet git directory, but when I run `glide install` I encounter this error : 
```
[ERROR]	Failed to find glide.yaml file in directory tree: Cannot resolve parent of /
```
I'm not very familiar with Go, but I highly suspect it is related to the fact that I use the latest Go1.11, while it is advised in the README file to use 1.5 or 1.6.

Is there something I can do about it except downgrading for the time to run it and upgrading again?

`uname -a`: Linux Celeri 4.15.0-38-generic #41-Ubuntu SMP Wed Oct 10 10:59:38 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

Thanks!

wpaulino:
The project uses go modules for dependencies now. You can build the binary with `GO111MODULE=on go install ./dropwtxmgr`.

BobleChinois:
Ok it worked, thanks. I will update the README file, ok?

BernhardFuchs:
As of writing this the correct command is ```GO111MODULE=on go install -v . ./cmd/...``` run from ```/btcwallet```. Found it in this [PR](https://github.com/wallclockbuilder/btcwallet/pull/1/commits/ee6e5093abe07bb4e64c81a461d84f1cb15fa14c).
Would be nice to finally merge this.

PR #596, opened 2019-02-20
Use go modules instead of glide

wallclockbuilder:
Glide was removed by https://github.com/btcsuite/btcwallet/commit/b523fd5274a0283638807cbf107ae77b2cfe5f89 in favor of go modules. 
Update readme to reflect current reality.
 Also fixes https://github.com/btcsuite/btcwallet/issues/588

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,28 +77,19 @@ Building or updating from source requires the following build dependencies:
     experiment by setting the `GO15VENDOREXPERIMENT` environment variable to
     `1`.  This step is not required for Go 1.6.
 
-- **Glide**
-
-  Glide is used to manage project dependencies and provide reproducible builds.
-  To install:
-
-  `go get -u github.com/Masterminds/glide`
-
-Unfortunately, the use of `glide` prevents a handy tool such as `go get` from
-automatically downloading, building, and installing the source in a single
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -77,28 +77,19 @@ Building or updating from source requires the following build dependencies:
...
 install the project.
 
 **Getting the source**:
 
 For a first time installation, the project and dependency sources can be
-obtained manually with `git` and `glide` (create directories as needed):
+obtained manually with `git` and `go modules` (create directories as needed):
 
 ```
 git clone https://github.com/btcsuite/btcwallet $GOPATH/src/github.com/btcsuite/btcwallet
```
dolanor: We can actually make it easier now with modules.
as:
```
git clone https://github.com/btcsuite/btcwallet
cd btcwallet
go get -v ./cmd/...
```
and if you cloned it in the $GOPATH, go would run in go classic, and if you cloned it outside $GOPATH, go would be `module aware` and thus download the the dependencies via go modules.

And if you're used to Go and still use GOPATH, you would just do a:
```
GO111MODULE=on go get -v github.com/btcsuite/btcwallet/cmd/...
```
What do you think?

PR #597, opened 2019-02-20
 wallet/wallet: remove invalid transactions when broadcast fails

wpaulino:
In this commit, we rework how `publishTransaction` works in order to correctly handle removing invalid transactions from the wallet's unconfirmed transaction store. This is crucial as otherwise, invalid transactions can remain within the wallet and be used for further transactions, causing a chain of inaccurate transactions.

`publishTransaction` will now only return an error if the transaction fails to be broadcast and it has not been previously seen in the mempool/chain. This is intended in order to provide an easier API to callers. Any other errors when broadcasting the transaction will cause it to be removed from the wallet's unconfirmed transaction store to ensure it maintains an accurate view of the chain.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3214,7 +3214,7 @@ func (w *Wallet) SendOutputs(outputs []*wire.TxOut, account uint32,
 		return nil, err
 	}
 
-	txHash, err := w.publishTransaction(createdTx.Tx)
+	txHash, err := w.reliablyPublishTransaction(createdTx.Tx)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -2823,96 +2823,29 @@ func (w *Wallet) LockedOutpoints() []btcjson.TransactionInput {
...
+		log.Errorf("Unable to retrieve unconfirmed transactions to "+
+			"resend: %v", err)
 		return
 	}
 
 	for _, tx := range txs {
-		resp, err := chainClient.SendRawTransaction(tx, false)
+		txHash, err := w.publishTransaction(tx)
 		if err != nil {
-			// If the transaction has already been accepted into the
```
Roasbeef: Yay code deletion!

PR #602, opened 2019-03-01
wallet/wallet: accept 0 bestHeight in developer environments

### LINKED ISSUE #594: Wallet hangs when getting new address on empty simnet blockchain
All following actions assume there are btcd.conf, btcwallet.conf and btcctl.conf files with simnet=1 and the same user and password.

Steps to reproduce:
1. Make sure local btcd db and wallet.db are removed.
2. Start btcd; reports the chain state with height 0.
3. Create a wallet:  btcwallet --create
4. Start btcwallet; reports "Caught up to height 0".
5. Try to get an address needed for mining: btcctl --wallet getnewaddress
Hangs here.

A little investigation reveals btcwallet locks a mutex in a DB transaction, then enters an infinite loop in function Wallet.scanChain:

// If we've reached our best height and we're not current, we'll
// wait for blocks at tip to ensure we go through all existent
// blocks.
for height == bestHeight && !isCurrent(bestHeight) {
	time.Sleep(100 * time.Millisecond)
	_, bestHeight, err = chainClient.GetBestBlock()
	if err != nil {
		return err
	}
}

The loop is infinite because height and bestHeight are always 0 and isCurrent always returns false on a brand-new, empty simnet blockchain. The command btcctl --wallet getnewaddress then starts new DB transaction and blocks on that mutex.

It seems would be wise to provide timeout exit from that loop.

wpaulino:
In this commit, we fix a regression in the wallet when attempting to
sync new developer test chains such as regtest and simnet. The wallet
would block syncing until a block was mined, but in order to a mine a
block, an address must be generated by the wallet first. This address
generation would block as the syncing logic was already holding the
database's mutex. As a result of this, we'll now accept the genesis
block as the wallet's birthday block, which should allow the wallet to
proceed as intended.

Fixes #594.
Replaces #599.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -450,6 +450,18 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
 	return w.rescanWithTarget(addrs, unspent, nil)
 }
 
+// isDevEnv determines whether the wallet is currently under a local developer
+// environment, e.g. simnet or regtest.
+func (w *Wallet) isDevEnv() bool {
+	switch uint32(w.ChainParams().Net) {
+	case uint32(chaincfg.RegressionNetParams.Net):
+	case uint32(chaincfg.SimNetParams.Net):
+	default:
+		return false
+	}
+	return true
+}
+
 // scanChain is a helper method that scans the chain from the starting height
 // until the tip of the chain. The onBlock callback can be used to perform
 // certain operations for every block that we process as we scan the chain.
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -450,6 +451,20 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
 	return w.rescanWithTarget(addrs, unspent, nil)
 }
 
+// isDevEnv determines whether the wallet is currently under a local developer
+// environment, e.g. simnet or regtest.
+func (w *Wallet) isDevEnv() bool {
+	switch uint32(w.ChainParams().Net) {
+	case uint32(chaincfg.RegressionNetParams.Net):
+	case uint32(chaincfg.SimNetParams.Net):
+	case uint32(ltcChainCfg.RegressionNetParams.Net):
```
Roasbeef: Nothing in btcwallet is aware of LTC, we shouldn't introduce awareness here.

wpaulino:
Pushed up a slightly new version to use the last block for chains that have blocks but are not current, rather than just always using the genesis block.

PR #603, opened 2019-03-03
Ensure best block is up to date during initial sync

roeierez:
This PR fixes a case during the initial sync of the wallet where the code stops too early to scan the chain and as a consequence fails to reach the birthday block, causing the initial start to stop
with error.

The scan is done in a loop until reaches the "bestHeight", where at that point the code only exits if the chain backend also reports it is synchronized at tip. 
Problem is that since the "bestHeight" is initialized before the loop it might contain an old value causing the main loop to exit too early.

The solution is to refresh the bestHeight on every iteration and by doing so making sure it is always called after isCurrent is called and might return true.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -515,6 +515,17 @@ func (w *Wallet) scanChain(startHeight int32,
 				return err
 			}
 		}
+
+		// if we have reached our initial bestHeight then we might be in a case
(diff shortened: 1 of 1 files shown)


wpaulino:
Thanks for reporting this @roeierez! I've opened #604 to go with a slightly different approach, but I listed you as a co-author regardless.

Roasbeef:
Replaced by #604

PR #610, opened 2019-04-08
[neutrino] Stop `ChainService` on shutdown 

halseth:
This ensures that we stop the underlying `ChainService` on of the `chainClient` before we close the database it is accessing.

Previously we would close the DB here: https://github.com/btcsuite/btcwallet/blob/8b90263a619068db81994688fbdf434be39a81ba/btcwallet.go#L165, while the `ChainService` was not actually stopped. Now we ensure that the call to `NeutrinoClient.Stop()` also stops the underlying `ChainService`.


chain/neutrino.go
```diff
@@ -88,12 +93,19 @@ func (s *NeutrinoClient) Start() error {
...
-	defer s.clientMtx.Unlock()
 	if !s.started {
+		s.clientMtx.Unlock()
 		return
 	}
-	close(s.quit)
 	s.started = false
+	s.clientMtx.Unlock()
+
+	if err := s.CS.Stop(); err != nil {
```
wpaulino: Would it be better to stop it after all the goroutines have exited? This way we don't attempt to use it while it's shutting down.

halseth: Hm, initially I wanted to avoid this in order to not change current behavior by waiting for the `wg` here. But maybe it is better to just do it, I agree.

halseth: Done!

chain/neutrino.go
```diff
@@ -92,12 +93,17 @@ func (s *NeutrinoClient) Start() error {
 // Stop replicates the RPC client's Stop method.
 func (s *NeutrinoClient) Stop() {
 	s.clientMtx.Lock()
-	defer s.clientMtx.Unlock()
 	if !s.started {
+		s.clientMtx.Unlock()
```
Roasbeef: The defer can likely be left as is, don't see anything performance critical here.

chain/neutrino.go
Roasbeef: So within `WaitForShutdown` then?

chain/neutrino.go
```diff
@@ -101,6 +101,8 @@ func (s *NeutrinoClient) Stop() {
 	s.started = false
 	s.clientMtx.Unlock()
 
+	s.wg.Wait()
```
Roasbeef: This now conflicts with the role of the `WaitForShutdown `  method.

halseth:
The point of this PR was to stop the `ChainService` when stopping the `NeutrinoClient` to prevent a crash when closing the database. But looking at the interface (that includes the `WaitForShutdown` method), it looks like the client is supposed to support being stopped and started again. This is not supported by the underlying `ChainService` which would explain why we are not stopping it in the first place.

Because of this realization, I reworked the commits to start and stop the `ChainService` **outside** the client (at the same level as the database is opened/closed), to ensure we can reuse it across restarts of the client.

chain/neutrino.go
halseth: Done

chain/neutrino.go
halseth: Stop no moved out of `NeutrinoClient`, leaving the `WaitForShutdown` behavior intact.

chain/neutrino.go
halseth: Reverted.

btcwallet.go
```diff
@@ -143,8 +143,39 @@ func walletMain() error {
...
 // methods.
 func rpcClientConnectLoop(legacyRPCServer *legacyrpc.Server, loader *wallet.Loader) {
-	var certs []byte
-	if !cfg.UseSPV {
+	var (
+		certs        []byte
+		chainService *neutrino.ChainService
+	)
+
+	if cfg.UseSPV {
```
Roasbeef: So this new change doesn't affect the `btcwallet` library at all, which is what we use ln `lnd` and elsewhere. If we go this route, seems that we'll need an additional set of modifications in `lnd` to use this new startup logic.

PR #613, opened 2019-04-19
chain: dispatch RescanFinished notification after BlockConnected

wpaulino:
In this PR, we address a bug within the wallet when running with Neutrino where it wouldn't be able to mark it as synced with the chain due to not receiving a RescanFinished notification if it was a block behind the chain. This happened because of the order in which the notifications are delivered within the underlying Neutrino ChainService, FilteredBlockConnected gets dispatched before BlockConnected. This doesn't always work however because there is an implicit dependency between them in which BlockConnected must occur first.

To address it, rather than switching the order in which the notifications are dispatched, we just check if we can dispatch the final RescanFinished notification after dispatching BlockConnected.

### DIFF
--- a/chain/neutrino.go
+++ b/chain/neutrino.go
@@ -512,40 +512,7 @@ func (s *NeutrinoClient) onFilteredBlockConnected(height int32,
 	s.clientMtx.Unlock()
 
 	// Handle RescanFinished notification if required.
-	bs, err := s.CS.BestBlock()
-	if err != nil {
-		log.Errorf("Can't get chain service's best block: %s", err)
-		return
-	}
-
-	if bs.Hash == header.BlockHash() {
-		// Only send the RescanFinished notification once.
-		s.clientMtx.Lock()
-		if s.finished {
-			s.clientMtx.Unlock()
-			return
-		}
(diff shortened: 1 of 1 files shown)


chain/neutrino.go
```diff
@@ -612,6 +587,57 @@ func (s *NeutrinoClient) onBlockConnected(hash *chainhash.Hash, height int32,
...
+// again.
+func (s *NeutrinoClient) dispatchRescanFinished() {
+	bs, err := s.CS.BestBlock()
+	if err != nil {
+		log.Errorf("Can't get chain service's best block: %s", err)
+		return
+	}
+
+	s.clientMtx.Lock()
+	if bs.Hash != s.lastFilteredBlockHeader.BlockHash() {
```
roeierez: Better check LastFilteredBlockHeader for nil? Since no particular order is assumed, if onBlockConnected dispatched first the value would be nil here.

wpaulino: Ah yes, good point @roeierez. Should be fine for now since `onBlockConnected` won't be called first in any case, but agree that the check is nice to have anyway. See https://github.com/btcsuite/btcwallet/pull/614.

PR #617, opened 2019-05-10
chain: misc BitcoindClient fixes

wpaulino:
Depends on https://github.com/btcsuite/btcd/pull/1428.


chain/bitcoind_client.go
```diff
@@ -906,8 +914,6 @@ func (c *BitcoindClient) FilterBlocks(
 // the client in the watch list. This is called only within a queue processing
 // loop.
 func (c *BitcoindClient) rescan(start chainhash.Hash) error {
-	log.Infof("Starting rescan from block %s", start)
```
halseth: is this log not useful?

chain/bitcoind_client.go
```diff
@@ -1089,6 +1082,8 @@ func (c *BitcoindClient) rescan(start chainhash.Hash) error {
 		}
 	}
 
+	c.onRescanFinished(bestHash, bestHeight, time.Unix(bestHeader.Time, 0))
```
halseth: 👍

wallet/wallet.go
```diff
@@ -3414,6 +3414,14 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 	}
 
 	txid, err := chainClient.SendRawTransaction(tx, false)
+
+	// Determine if this was an RPC error thrown due to the transaction
+	// already confirming.
+	var rpcTxConfirmed bool
+	if rpcErr, ok := err.(*btcjson.RPCError); ok {
```
halseth: Possible to add a unit test for this behavior?

chain/bitcoind_client.go
wpaulino: It's already logged by the wallet itself, so opted to remove this to avoid the duplicate info.

wallet/wallet.go
wpaulino: It's somewhat already covered [here](https://github.com/lightningnetwork/lnd/blob/master/lnwallet/interface_test.go#L1435).

wallet/wallet.go
```diff
@@ -3438,19 +3446,19 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
+	// already confirmed to a btcd node over the P2P network.
+	case strings.Contains(err.Error(), "transaction already exists"):
 		fallthrough
 
-	// This error is returned when sending a transaction that has already
-	// confirmed to a bitcoind node over RPC.
-	case strings.Contains(err.Error(), "transaction already in block chain"):
+	// This error is returned when broadcasting a transaction that has
+	// already confirmed to a bitcoind node over the P2P network.
+	case strings.Contains(err.Error(), "txn-already-known"):
```
Roasbeef: We'll need to maintain the old _and_ new error messages in order to support the prior version as we do currently.

wpaulino: This is already the case for both backends with this commit. `rpcTxConfirmed` will be caught by old (< 0.18.0) and new (>= 0.18.0) `bitcoind` versions and new `btcd` versions. `transaction already exists` will be caught by old `btcd` nodes (those before https://github.com/btcsuite/btcd/pull/1428).

chain/bitcoind_client.go
```diff
@@ -945,13 +945,6 @@ func (c *BitcoindClient) rescan(start chainhash.Hash) error {
 	}
 	headers.PushBack(previousHeader)
 
-	// Queue a RescanFinished notification to the caller with the last block
-	// processed throughout the rescan once done.
-	defer c.onRescanFinished(
-		previousHash, previousHeader.Height,
-		time.Unix(previousHeader.Time, 0),
```
Roasbeef: Nice catch! `previousHash` or `previousHeader` were also never even updated, so it's not as though they would've eventually pointed to the most up to date versions.

wallet/wallet.go
halseth: I was thinking a `btcwallet` unit test with a mocked `chainClient`. We could make the mock mimic the behavior of various backends/versions, and ensure we handle the diff errors correctly.

wpaulino: Since we're doing string matching I'm not sure a unit test with a mock would work here. We'd just be including the same strings in there, which would seem to me like we're testing `string.Contains`.

wallet/wallet.go
```diff
@@ -3438,19 +3446,19 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
+	// This error is returned when sending a transaction that has already
+	// confirmed to a btcd/bitcoind node over RPC.
+	case rpcTxConfirmed:
 		fallthrough
 
 	// This error is returned when broadcasting a transaction that has
-	// already confirmed to a bitcoind node.
-	case strings.Contains(err.Error(), "txn-already-known"):
+	// already confirmed to a btcd node over the P2P network.
+	case strings.Contains(err.Error(), "transaction already exists"):
```
cfromknecht: would prefer if both of these cases were made case insensitive for robustness

wpaulino: Done.

Issue #616, opened 2019-04-23
wallet: instant start up with full node backend


Roasbeef:
Today when spinning up with a new full node, for each block in the chain, [we'll write a new synced to checkpoint to disk](https://github.com/btcsuite/btcwallet/blob/master/wallet/wallet.go#L393). This makes sync _very_ slow as we need to do this for each block, and then also increases storage as we maintain a mapping from `blockHeight -> blockHash` on disk. 

Instead, what we should do is simply _wait_ for the chain backend to sync to the chain. Then obtain and write the like 10k (a generous safety margin) blocks to disk. This relates to #555, as it starts to only maintain these 10k blocks on disk, but will will initially perform a _write_ for each block. 

With the above completed, we should have more or less instant node start up for btcd and bitcoind. Additionally, if the node is still syncing while we're running, then we'll no longer process new block mined notifications, instead just waiting until the node has been fully synced.

PR #618, opened 2019-05-10
wallet+waddrmgr: sync and store up to MaxReorgDepth blocks

wpaulino:
In this PR, we modify the wallet's block hash index to only store up to MaxReorgDepth blocks. This allows us to reduce consumed storage, as we'd be mostly storing duplicate data. We choose to store up to MaxReorgDepth to ensure we can recover from a potential long reorg.

We also add a migration that will be used by existing wallets to ensure they can adhere to the new requirement of storing up to MaxReorgDepth entries within the block hash index.

As a result of only storing up to MaxReorgDepth, we can also speed up the initial sync process for the wallet as we can just fetch the latest MaxReorgDepth blocks of the chain.

Depends on #617 to address some `bitcoind` issues.
Partly addresses https://github.com/btcsuite/btcwallet/issues/513, except we do not batch the requests to the chain backend.
Fixes https://github.com/btcsuite/btcwallet/issues/616.

NOTE: Since this is changing the initial sync logic, we should make sure to test manually and also through the set of tests within `lnd` to ensure there are no regressions.


Roasbeef:
cc @halseth

halseth:
for reference: https://github.com/btcsuite/btcwallet/pull/555

wallet/chainntfns.go
```diff
@@ -47,42 +52,45 @@ func (w *Wallet) handleChainNotifications() {
...
+
+			blockStamps = append(blockStamps, &waddrmgr.BlockStamp{
+				Hash:      *hash,
+				Height:    height,
+				Timestamp: header.Timestamp,
+			})
+			return nil
 
-			for i := startBlock.Height + 1; i <= height; i++ {
-				hash, err := client.GetBlockHash(int64(i))
```
halseth: whoa, we were doing RPC calls withing a DB tx? 😮

wallet/chainntfns.go
```diff
@@ -47,42 +52,45 @@ func (w *Wallet) handleChainNotifications() {
...
 
-				bs := waddrmgr.BlockStamp{
-					Height:    i,
-					Hash:      *hash,
-					Timestamp: header.Timestamp,
-				}
-				err = w.Manager.SetSyncedTo(ns, &bs)
+		err = walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+			ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+			for _, blockStamp := range blockStamps {
```
halseth: I know that for certain backends, the actual `GetBlockX` calls are very slow. And since we must finish them all before committing to the DB, the progress is lost if the wallet is restarted mid-way.

Should we consider writing to the DB each time N blockstamps have been fetched? (N=1 or maybe bigger?)

wpaulino: This is already called in batches of 10K from `BitcoindClient`, a ticker is used within `RPCClient` to send them periodically, and `NeutrinoClient` only sends a final `OnRescanFinished` notification. We could batch them here if it is an issue though.

waddrmgr/migrations.go
```diff
@@ -372,3 +376,37 @@ func resetSyncedBlockToBirthday(ns walletdb.ReadWriteBucket) error {
 
 	return PutSyncedTo(ns, &birthdayBlock)
 }
+
+// storeMaxReorgDepth is a migration responsible for allowing the wallet to only
+// maintain MaxReorgDepth block hashes stored in order to recover from long
+// reorgs.
+func storeMaxReorgDepth(ns walletdb.ReadWriteBucket) error {
```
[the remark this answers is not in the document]
wpaulino: Indeed we could, although it is a small one. Will wait for someone else to chime in before proceeding.

wpaulino:
Pushed a significant rewrite that addresses some comments @cfromknecht and I discussed offline where it was possible for us to miss scanning relevant blocks if the birthday was before the reorg safe height. I also included some nice refactors along the way.

PTAL @cfromknecht @halseth!

wallet/chainntfns.go
wpaulino: Yeah we do this in other places too, like the recovery logic.

wallet/rescan.go
```diff
@@ -134,9 +141,12 @@ out:
 						"currently running")
 					continue
 				}
-				w.rescanProgress <- &RescanProgressMsg{
+				select {
+				case w.rescanProgress <- &RescanProgressMsg{
 					Addresses:    curBatch.addrs,
 					Notification: n,
+				}:
+				case <-quit:
```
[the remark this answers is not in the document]
wpaulino: Done.

wallet/wallet.go
```diff
@@ -493,6 +506,85 @@ func (w *Wallet) waitUntilBackendSynced(chainClient chain.Interface) error {
...
+		birthdayBlock *waddrmgr.BlockStamp
+		left, right   = startHeight, bestHeight
+	)
+
+	// Binary search for a block that meets the birthday timestamp by a
+	// margin of +/-2 hours.
+	for {
+		// Retrieve the timestamp for the block halfway through our
+		// range.
+		mid := left + (right-left)/2
```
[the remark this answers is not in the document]
wpaulino: Heh I've always written it this way to prevent the potential overflow.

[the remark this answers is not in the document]
cfromknecht: fair enough, i suppose this will save btcwallet devs a 40000 years from now from hitting that :P

wallet/wallet.go
```diff
@@ -493,6 +506,85 @@ func (w *Wallet) waitUntilBackendSynced(chainClient chain.Interface) error {
...
+		// The block's timestamp is more than 2 hours after the
+		// birthday, so look for a lower block.
+		if header.Timestamp.Sub(birthday) > birthdayBlockDelta {
+			right = mid
+			continue
+		}
+
+		// The birthday is more than 2 hours before the block's
+		// timestamp, so look for a higher block.
+		if header.Timestamp.Sub(birthday) < -birthdayBlockDelta {
```
halseth: Do we risk this this looping if `left=right=mid`?

wallet/wallet.go
```diff
@@ -493,6 +506,85 @@ func (w *Wallet) waitUntilBackendSynced(chainClient chain.Interface) error {
...
+
+	log.Debugf("Locating suitable block for birthday %v between blocks "+
+		"%v-%v", birthday, startHeight, bestHeight)
+
+	var (
+		birthdayBlock *waddrmgr.BlockStamp
+		left, right   = startHeight, bestHeight
+	)
+
+	// Binary search for a block that meets the birthday timestamp by a
```
halseth: I think we need to be careful here, since header timestamps are not always increasing. Could do something like always jump 11 blocks back to be sure.

wallet/wallet.go
```diff
@@ -683,108 +676,137 @@ func (w *Wallet) recovery(startHeight int32) error {
...
-		tx.Rollback()
-		return err
-	}
-	addrMgrNS := tx.ReadWriteBucket(waddrmgrNamespaceKey)
-	err = recoveryMgr.Resurrect(addrMgrNS, scopedMgrs, credits)
+	err = walletdb.View(w.db, func(tx walletdb.ReadTx) error {
+		txMgrNS := tx.ReadBucket(wtxmgrNamespaceKey)
+		credits, err := w.TxStore.UnspentOutputs(txMgrNS)
+		if err != nil {
+			tx.Rollback()
```
halseth: Rollback in view tx is not needed.

wallet/wallet.go
cfromknecht: The wallet already sets its birthday to two days before the birthday in the seed, so we already have enough breathing room I think?

wallet/wallet.go
wpaulino: This is already handled in the `mid == startHeight || mid == bestHeight` check above, no?

wallet/wallet.go
wpaulino: > The wallet already sets its birthday to two days before the birthday in the seed, so we already have enough breathing room I think?

Yes I think this is fine as well.

wallet/wallet.go
wpaulino: Nice catch, fixed.

wallet/wallet.go
halseth: e.g. `left=10, right=11,mid=10` looks like it still can loop. Could be fixd by just returning if `mid==left` I think.

wallet/wallet.go
halseth: Yeah, then we should be fine.

wallet/wallet.go
```diff
@@ -493,6 +506,85 @@ func (w *Wallet) waitUntilBackendSynced(chainClient chain.Interface) error {
 	}
 }
 
+// locateBirthdayBlock returns a block that meets the given birthday timestamp
+// by a margin of +/-2 hours. This is safe to do as the timestamp is already 2
+// days in the past of the actual timestamp.
+func locateBirthdayBlock(chainClient chain.Interface,
```
halseth: method should have some tests

wallet/wallet.go
```diff
@@ -683,108 +676,136 @@ func (w *Wallet) recovery(startHeight int32) error {
...
-		// We'll checkpoint our current batch every 2K blocks, so we'll
-		// need to start a new database transaction. If our current
-		// batch is empty, then this will act as a NOP.
-		if height%recoveryBatchSize == 0 {
-			blockBatch := recoveryMgr.BlockBatch()
-			err := w.recoverDefaultScopes(
-				chainClient, tx, addrMgrNS, blockBatch,
-				recoveryMgr.State(),
-			)
+	for height < bestHeight {
```
halseth: Any specific reason not to keep the existing single loop with `if height%recoveryBatchSize == 0` check?

wpaulino: This logic had to be changed in order to handle the recovery syncing blocks that are not part of our birthday but required as part of the wallet maintaining the last 10K block hashes. This check still exists within the inner loop as `for len(recoveryMgr.BlockBatch()) < recoveryBatchSize`.

wallet/wallet.go
```diff
@@ -467,6 +475,24 @@ func (w *Wallet) isDevEnv() bool {
 	return true
 }
 
+// waitUntilBackendSynced blocks until the chain backend considers itself
+// "current".
+func (w *Wallet) waitUntilBackendSynced(chainClient chain.Interface) error {
+	// We'll poll every second to determine if our chain considers itself
+	// "current".
+	t := time.NewTicker(time.Second)
```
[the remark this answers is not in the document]
wpaulino: Does it matter since it goes out of scope? Fixed anyway.

wallet/wallet.go
```diff
@@ -467,6 +475,24 @@ func (w *Wallet) isDevEnv() bool {
 	return true
 }
 
+// waitUntilBackendSynced blocks until the chain backend considers itself
+// "current".
+func (w *Wallet) waitUntilBackendSynced(chainClient chain.Interface) error {
+	// We'll poll every second to determine if our chain considers itself
```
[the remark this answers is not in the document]
wpaulino: Already logged by `lnd`, so I'd rather avoid a duplicate message.

wallet/wallet.go
wpaulino: There are some tests within `wallet/chainntnfs_test.go` that already exercise this behavior: `TestBirthdaySanityCheckHigherEstimate` and `TestBirthdaySanityCheckLowerEstimate`. Do you think additional tests would still be nice to have?

wallet/wallet.go
wpaulino: Ah I see, good catch. Fixed.

wallet/wallet.go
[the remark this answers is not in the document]
halseth: Okay, sounds good :)

wallet/wallet.go
halseth: Looks like they call into `birthdaySanityCheck`. Since they probably cover most of the logic of this method anyway, I agree that we don't have to add extensive tests for this method. But would be nice to have a simple test targeting this method directly, just as a starting point for adding more tests down the line.

wallet/wallet.go
```diff
@@ -685,108 +678,136 @@ func (w *Wallet) recovery(startHeight int32) error {
...
+		}
+
+		if len(batch) > 0 {
+			log.Infof("Recovered addresses from blocks %d-%d",
+				batch[0].Height, batch[len(batch)-1].Height)
+		}
+
+		// Clear the batch of all processed blocks to reuse the same
+		// memory for future batches.
+		blocks = blocks[:0]
```
halseth: are `blocks` actually used anywhere?

wallet/wallet.go
halseth: So how is this different from 
```go
for height < bestHeight {
     if height >= birthday.height {
          addToBatch()
     }

     if len(blockBatch) >= batchSize || height == bestBlock {
            recoverScopes()
            resetBlockBatch()
     }
```

Not that it matters that much, but I find a single loop easier to reason about then two, and I'm trying to understand why two is needed.

wallet/wallet.go
```diff
@@ -685,108 +678,136 @@ func (w *Wallet) recovery(startHeight int32) error {
...
-	// Now that we've reached the chain tip, we can process our final batch
-	// with the remaining blocks if it did not reach its maximum size.
-	blockBatch := recoveryMgr.BlockBatch()
-	err = w.recoverDefaultScopes(
-		chainClient, tx, addrMgrNS, blockBatch, recoveryMgr.State(),
-	)
+	return nil
+}
+
+// syncHeightRange determines the best height range to sync with the chain to
```
halseth: naming suggestion: `getSyncRange`? Since current name indicates something is being synced when this method is called :P

chain/bitcoind_client.go
```diff
@@ -265,7 +265,40 @@ func (c *BitcoindClient) NotifyTx(txids []chainhash.Hash) error {
...
+		return fmt.Errorf("unable to retrieve best block: %v", err)
+	}
+	bestHeader, err := c.GetBlockHeaderVerbose(bestHash)
+	if err != nil {
+		atomic.StoreUint32(&c.notifyBlocks, 0)
+		return fmt.Errorf("unable to retrieve header for best block: "+
+			"%v", err)
+	}
+
+	c.bestBlock.Hash = *bestHash
```
halseth: no mutex needed here?

chain/bitcoind_client.go
```diff
@@ -429,22 +462,14 @@ func (c *BitcoindClient) Start() error {
 			"%v", err)
 	}
 
-	c.bestBlockMtx.Lock()
 	c.bestBlock = waddrmgr.BlockStamp{
```
halseth: sane mutex question

wallet/wallet.go
wpaulino: Done. Much cleaner now :)

wallet/wallet.go
wpaulino: They're used in the database transaction above.

wallet/wallet.go
wpaulino: Done.

chain/bitcoind_client.go
wpaulino: Fixed.

chain/bitcoind_client.go
wpaulino: I don't think it's necessary here since this is the `Start` method, but added anyway.

wallet/wallet.go
```diff
@@ -327,7 +327,20 @@ func (w *Wallet) activeData(dbtx walletdb.ReadTx) ([]btcutil.Address, []wtxmgr.C
...
+	chainClient, err := w.requireChainClient()
+	if err != nil {
+		return err
+	}
+
+	// We'll wait until the backend is synced to ensure we get the latest
+	// MaxReorgDepth blocks to store. We don't do this for development
+	// environments as we can't guarantee a lively chain.
+	if !w.isDevEnv() {
+		if err := w.waitUntilBackendSynced(chainClient); err != nil {
```
halseth: Hm, I think I would have a log here. It just makes sense to log before doing a potentially long blocking call.

wallet/wallet.go
halseth: reminder to add a simple unit test for this method.

wallet/wallet.go
```diff
@@ -685,108 +678,127 @@ func (w *Wallet) recovery(startHeight int32) error {
...
+	for height := startHeight; height <= bestHeight; height++ {
+		hash, err := chainClient.GetBlockHash(int64(height))
+		if err != nil {
+			return err
+		}
+		header, err := chainClient.GetBlockHeader(hash)
+		if err != nil {
+			return err
+		}
+		blocks = append(blocks, &waddrmgr.BlockStamp{
```
halseth: should blocks before our birthday be added to the slice? If yes, add comment what this achieves.

wpaulino: There's already a comment at the top of the loop explaining the rationale?

> Since the recovery process itself acts as rescan, we'll also update our wallet's synced state along the way to reflect the blocks we process and prevent rescanning them later on.

wallet/wallet.go
wpaulino: Made it debug to prevent the duplicate log.

wallet/wallet.go
wpaulino: Done.

wallet/rescan.go
```diff
@@ -103,18 +107,23 @@ func (b *rescanBatch) done(err error) {
...
 		case job := <-w.rescanAddJob:
 			if curBatch == nil {
 				// Set current batch as this job and send
 				// request.
 				curBatch = job.batch()
-				w.rescanBatch <- curBatch
+				select {
+				case w.rescanBatch <- curBatch:
+				case <-quit:
+					return
```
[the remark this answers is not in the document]
wpaulino: Indeed we do, fixed!

chain/bitcoind_client.go
wpaulino: It's meant to be there since `BlockStamp` can be called at any point, which acquires the lock.

PR #624, opened 2019-06-05
chain: prevent deadlock while notifying RescanFinished for NeutrinoClient

wpaulino:


chain/neutrino.go
```diff
@@ -369,6 +374,7 @@ func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Addre
 		case <-s.rescanQuit:
 			return nil
 		}
+		s.clientMtx.Lock()
```
halseth: I think we must make sure the mutex is not double unlocked on quit here, since the defer will also unlock it.

chain/neutrino.go
```diff
@@ -358,6 +358,11 @@ func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Addre
 	// notification indicating the rescan has "finished".
 	if header.BlockHash() == *startHash {
 		s.finished = true
+
+		// Release the lock while dispatching the notification since
+		// it's possible for the notificationHandler to be waiting to
+		// acquire it before receiving the notification.
+		s.clientMtx.Unlock()
```
halseth: Nice, I think this is an ok solution for now, this should fix the issue I ran into. Longer term we should move to a more granular set of mutexes (or get rid of the mutexes all together), but short term this is fine :+1:

chain/neutrino.go
wpaulino: Fixed.

chain/neutrino.go
```diff
@@ -322,8 +322,8 @@ func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Addre
 	outPoints map[wire.OutPoint]btcutil.Address) error {
 
 	s.clientMtx.Lock()
-	defer s.clientMtx.Unlock()
```
cfromknecht: it seems safer to just use a defer here? there are several fields that are protected by a mutex in this method that would now be accessed without one. Is it required that the lock be released during `WaitForShutdown`?

wpaulino: Needs to be release when dispatching the RescanFinished notification.

cfromknecht: okay, then maybe we only do that?

wpaulino: Not sure I follow. I don't know if it needs to be released for `WaitForShutdown`, so I'd rather keep the same behavior there. I wasn't a fan of releasing and re-acquiring the lock again for dispatching the notification while keeping the `defer` at the top, so I went for the cleaner (IMO) approach.

wpaulino: Never mind, fixed.

cfromknecht: there's also one other case i think
https://github.com/btcsuite/btcwallet/pull/624/files#diff-3d0490616ade8665ccb58730190699bdR334
`s.rescan` is modified under the lock, but there it accessed w/o it

wpaulino: Fixed.

chain/neutrino.go
```diff
@@ -358,17 +359,30 @@ func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Addre
...
 		case <-s.quit:
+			// We'll need to re-acquire the lock in order for the
+			// deferred release call to not panic.
+			s.clientMtx.Lock()
 			return nil
-		case <-s.rescanQuit:
+		case <-rescanQuit:
+			// We'll need to re-acquire the lock in order for the
+			// deferred release call to not panic.
+			s.clientMtx.Lock()
```
halseth: Could we instead get rid of the defer and unlock manually on each return case?

chain/neutrino.go
```diff
@@ -358,17 +359,30 @@ func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Addre
...
 			return nil
 		}
+		s.clientMtx.Lock()
```
halseth: I noticed that we are not returning the error on line 391

chain/neutrino.go
wpaulino: @cfromknecht seems to prefer the defer approach, though that might've been because there were some fields that were being used without the lock in the prior diff.

chain/neutrino.go
wpaulino: Heh good catch!

chain/neutrino.go
cfromknecht: Yeah I’m fine with not having a defer, originally thought it’d make a smaller diff but seems not :P

wpaulino: Done.

PR #625, opened 2019-06-10
Add support for pay to witness script address

yaslama:


yaslama:
@Roasbeef @wpaulino @halseth @cfromknecht What do you think about this PR? It's not intrusive at all, quite trivial, and only add standard witness address type.

waddrmgr/address.go
```diff
@@ -646,3 +646,154 @@ func newScriptAddress(m *ScopedKeyManager, account uint32, scriptHash,
...
+
+	// Account manager must be unlocked to decrypt the script.
+	if a.manager.rootManager.IsLocked() {
+		return nil, managerError(ErrLocked, errLocked, nil)
+	}
+
+	// Decrypt the script as needed.  Also, make sure it's a copy since the
+	// script stored in memory can be cleared at any time.  Otherwise,
+	// the returned script could be invalidated from under the caller.
+	return a.unlock(a.manager.rootManager.cryptoKeyScript)
```
guggero, line 779: I know the non-segwit `scriptAddress` above uses this crypto key. But I'm thinking that we might want to support watch-only witness script addresses. So I'm not sure if we should use the `cryptokeyPub` here instead? Then we don't need the check for `a.manager.rootManager.WatchOnly()` or `a.manager.rootManager.IsLocked()`.

waddrmgr/address.go
```diff
@@ -646,3 +646,154 @@ func newScriptAddress(m *ScopedKeyManager, account uint32, scriptHash,
...
+	zero.Bytes(a.scriptCT)
+	a.scriptCT = nil
+	a.scriptMutex.Unlock()
+}
+
+// Account returns the account the address is associated with.  This will always
+// be the ImportedAddrAccount constant for script addresses.
+//
+// This is part of the ManagedAddress interface implementation.
+func (a *witnessScriptAddress) Account() uint32 {
```
guggero, line 702: nit: would be nice if we could extract some of that duplicate code into a `baseScriptAddress` type that can be embedded into both types?

waddrmgr/scoped_manager.go
```diff
@@ -550,7 +551,9 @@ func (s *ScopedKeyManager) scriptAddressRowToManaged(row *dbScriptAddressRow) (M
 		str := "failed to decrypt imported script hash"
 		return nil, managerError(ErrCrypto, str, err)
 	}
-
+	if len(scriptHash) == 32 {
+		return newWitnessScriptAddress(s, row.account, scriptHash, row.encryptedScript)
```
guggero, line 555: nit: line length.

waddrmgr/scoped_manager.go
```diff
@@ -1636,6 +1639,117 @@ func (s *ScopedKeyManager) ImportScript(ns walletdb.ReadWriteBucket,
 	return scriptAddr, nil
 }
 
+// ImportWitnessScript imports a user-provided script into the address manager.  The
```
guggero, line 1642: nit: double spaces before comments (we tend to eliminate those for _new_ code).

waddrmgr/scoped_manager.go
```diff
@@ -1636,6 +1639,117 @@ func (s *ScopedKeyManager) ImportScript(ns walletdb.ReadWriteBucket,
...
+// imported script will act as a pay-to-witness-script-hash address.
+//
+// All imported script addresses will be part of the account defined by the
+// ImportedAddrAccount constant.
+//
+// When the address manager is watching-only, the script itself will not be
```
guggero, line 1648: See comment above about this sentence. Or maybe we can add a switch to decide whether this is considered to be _that_ private? So the user of the library can choose what crypto key to use? cc @Roasbeef

waddrmgr/scoped_manager.go
```diff
@@ -1636,6 +1639,117 @@ func (s *ScopedKeyManager) ImportScript(ns walletdb.ReadWriteBucket,
...
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	// The manager must be unlocked to encrypt the imported script.
+	if s.rootManager.IsLocked() && !s.rootManager.WatchOnly() {
+		return nil, managerError(ErrLocked, errLocked, nil)
+	}
+
+	// Prevent duplicates.
+	sh := sha256.Sum256(script)
```
guggero, line 1666: Combined with the comment above: we could extract this duplicated code into its own function that just takes a `baseScriptAddress` type (and the new cryptoKeyType parameter mentioned in the other comment).

guggero (commented):
Thanks for the PR and sorry for the delay in the review!

Looks pretty good, main comments are around code duplication and what key to use to encrypt the script. Some test cases to get basic coverage would be nice as well.

guggero:
Needs a rebase as well.

PR #627, opened 2019-06-20
wallet: start recovery from the wallet's best height

wpaulino:
Previously, the wallet would attempt to store the same block it checkpointed during its initial sync when performing a recovery. This would cause the previous block existence validation check to be in place, which would ultimately fail because the previous block was not stored intentionally.

To address this, we always start/resume our recovery from the wallet's best height. This also ensures that we do not rescan the same block again when resuming a recovery after a shutdown.

Fixes https://github.com/lightningnetwork/lnd/issues/3223.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -643,23 +643,24 @@ func (w *Wallet) recovery(chainClient chain.Interface,
 		return err
 	}
 
-	// We'll then need to determine the range of our recovery. This properly
-	// handles the case where we resume a previous recovery attempt after a
-	// restart.
-	startHeight, bestHeight, err := w.getSyncRange(chainClient, birthdayBlock)
+	// Fetch the best height from the backend to determine when we should
+	// stop.
+	_, bestHeight, err := chainClient.GetBestBlock()
 	if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -643,23 +643,24 @@ func (w *Wallet) recovery(chainClient chain.Interface,
...
+	// ensure we properly handle restarts. Since the recovery process itself
+	// acts as rescan, we'll also update our wallet's synced state along the
+	// way to reflect the blocks we process and prevent rescanning them
+	// later on.
 	//
 	// NOTE: We purposefully don't update our best height since we assume
 	// that a wallet rescan will be performed from the wallet's tip, which
 	// will be of bestHeight after completing the recovery process.
 	var blocks []*waddrmgr.BlockStamp
+	startHeight := w.Manager.SyncedTo().Height + 1
```
halseth: so the assumption is that everything up to this height has already been recovered?

wpaulino: Yeah. In the case of the first recovery attempt ever happening, we can potentially start from a block after our birthday block, but this should be fine as it's already an estimate 2 days prior to the actual birthday.

PR #632, opened 2019-07-09
chain: explicitly close connections in BitcoindConn's Stop method

wpaulino:
Since it's now possible for gozmq.Conn to block when calling Receive, BitcoindConn hangs upon being stopped because its goroutines are waiting for a message to be delivered. To address this, we modify it to close its ZMQ connections driving the goroutines once it's been stopped. This allows the goroutines to unblock by detecting the EOF error and exiting.

Depends on https://github.com/lightninglabs/gozmq/pull/3.


go.mod
```diff
@@ -23,3 +24,5 @@ require (
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/yaml.v2 v2.2.2 // indirect
 )
+
+replace github.com/lightninglabs/gozmq => github.com/wpaulino/gozmq v0.0.0-20190709225702-f5caaa46981f
```
Roasbeef: Can be removed now in favor of updating the dep directly.

wpaulino: Done.

PR #636, opened 2019-08-10
chain: prevent filtering blocks when BitcoindClient has an empty filter

wpaulino:


chain/bitcoind_client.go
```diff
@@ -1133,9 +1133,13 @@ func (c *BitcoindClient) rescan(start chainhash.Hash) error {
 func (c *BitcoindClient) filterBlock(block *wire.MsgBlock, height int32,
 	notify bool) []*wtxmgr.TxRecord {
 
-	// If this block happened before the client's birthday, then we'll skip
-	// it entirely.
-	if block.Header.Timestamp.Before(c.birthday) {
+	// If this block happened before the client's birthday or we have
```
Roasbeef: We can optimize this further by not fetching the block at all, and just returning `nil` if our filter is empty.

PR #647, opened 2019-09-19
Cleanup publication error matching, add ErrDoubleSpend/ErrReplacment

halseth:
This is meant to provide a unified interface for publication of transactions, regardless of the active backend.

The error types `ErrDoubleSpend` and `ErrReplacement` also gives some more information to the caller about the type of error, making it easier to take the correct action.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3294,6 +3294,25 @@ func (w *Wallet) SignTransaction(tx *wire.MsgTx, hashType txscript.SigHashType,
 	return signErrors, err
 }
 
+// ErrDoubleSpend is an error returned from PublishTransaction in case the
+// published transaction failed to propagate since it was double spending a
+// confirmed transaction or a transaction in the mempool.
+type ErrDoubleSpend struct {
+	backendError error
+}
+
+// Error returns the string representation of ErrDoubleSpend.
+//
+// NOTE: Satisfies the error interface.
+func (e *ErrDoubleSpend) Error() string {
+	return fmt.Sprintf("double spend: %v", e.backendError)
+}
+
+// Unwrap returns the underlying error returned from the backend.
+func (e *ErrDoubleSpend) Unwrap() error {
+	return e.backendError
+}
+
 // PublishTransaction sends the transaction to the consensus RPC server so it
 // can be propagated to other nodes and eventually mined.
 //
@@ -3384,7 +3403,11 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 		rpcTxConfirmed = rpcErr.Code == btcjson.ErrRPCTxAlreadyInChain
 	}
 
-	txid := tx.TxHash()
+	var (
+		txid      = tx.TxHash()
+		returnErr error
+	)
+
 	switch {
 	case err == nil:
 		return &txid, nil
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -3369,6 +3369,12 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 		return nil, err
 	}
 
+	// match is a helper method to easily string match on the error
```
Roasbeef: In the commit message: matchin -> matching

wallet/wallet.go
```diff
@@ -3443,29 +3461,70 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
+	// non-replacable transaction already in the mempool.
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L622
+	case match(err, "txn-mempool-conflict"):
+		fallthrough
+
+	// Returned by bitcoind on the RPC when broadcasting a transaction that
+	// is spending either output that is missing or already spent.
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/node/transaction.cpp#L49
+	case match(err, "missing inputs"):
+		returnErr = &ErrDoubleSpend{
```
Roasbeef: Isn't this a reversal of our prior direction? Wherein we removed the err double spend as it was deemed to not give the caller any pertinent information? Say like on re-broadcast. In the past, we had `err != ErrDoubleSpend` littered all over `lnd`.

wallet/wallet.go
```diff
@@ -3294,6 +3294,20 @@ func (w *Wallet) SignTransaction(tx *wire.MsgTx, hashType txscript.SigHashType,
...
+// published transaction failed to propagate since it was double spending a
+// confirmed transaction or a transaction in the mempool.
+type ErrDoubleSpend struct {
+	backendError error
+}
+
+// Error returns the string representation of ErrDoubleSpend.
+//
+// NOTE: Satisfies the error interface.
+func (e *ErrDoubleSpend) Error() string {
```
Roasbeef: Add an `Unwrap` method for our go 1.13 fam?

wallet/wallet.go
```diff
@@ -3499,6 +3513,26 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 			backendError: err,
 		}
 
+	// Returned by bitcoind when a replacement transaction did not have
+	// enough fee.
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L830
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L894
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L904
+	case match(err, "insufficient fee"):
```
Roasbeef: What about when it is rejected for other reasons? As an example if it creates a chain of transactions that's deemed "too long".

wallet/wallet.go
halseth: Those will still be handled by the above cases (`"transaction already exists"`, `"txn-already-known"` etc) and return a nil-error.

Note that this  change doesn't change anything about what we consider a publication error. Previously we would just return these errors directly, and do the string matching in lnd. Now we'll still return the error, it is just wrapped within `ErrDoubleSpend`.

wallet/wallet.go
halseth: Done: https://github.com/btcsuite/btcwallet/pull/647/commits/784134b4688c709780dd8b8e3cb28d3ba85b47dc

wallet/wallet.go
halseth: I intentionally didn't add it since I assume we won't be running into those situations, but agree that we might as well add it when at it. Done here: https://github.com/btcsuite/btcwallet/pull/647/commits/d131e20be1b1a16767893be343e3366e5544bbde

halseth:
Review address. PTAL @Roasbeef @wpaulino

wallet/wallet.go
```diff
@@ -3443,29 +3461,70 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
+	// We received an error not matching any of the above cases.
+	default:
+		returnErr = fmt.Errorf("unmatched backend error: %v", err)
 	}
+
+	// If the transaction was rejected for whatever other reason, then
+	// we'll remove it from the transaction store, as otherwise, we'll
+	// attempt to continually re-broadcast it, and the UTXO state of the
+	// wallet won't be accurate.
+	dbErr := walletdb.Update(w.db, func(dbTx walletdb.ReadWriteTx) error {
```
wpaulino: This was already the case before these changes, but we should move the transaction removal to `reliablyPublishTransaction` since that's where we add it in the first place.

halseth: I agree, but it's not totally straightforward since we also want to remove it in the non-error `"txn-already-known"` case above.

halseth:
Rebased!

wallet/wallet.go
wpaulino: The removal could be extracted into a method and it could then be used in cases where we rebroadcast and receive `txn-already-known`. Not a blocker though.

PR #653, opened 2019-10-23
wtxmgr: improve error message for failed tx record retrieval

wpaulino:
This change is motivated by https://github.com/lightningnetwork/lnd/issues/3624. We now include the transaction we failed to locate to help track down how the wallet entered its invalid state.

### DIFF
--- a/wtxmgr/tx.go
+++ b/wtxmgr/tx.go
@@ -7,6 +7,7 @@ package wtxmgr
 
 import (
 	"bytes"
+	"fmt"
 	"time"
 
 	"github.com/btcsuite/btcd/blockchain"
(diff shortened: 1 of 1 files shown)


wtxmgr/tx.go
```diff
@@ -768,7 +770,8 @@ func (s *Store) UnspentOutputs(ns walletdb.ReadBucket) ([]Credit, error) {
 		var rec TxRecord
 		err = readRawTxRecord(&op.Hash, recVal, &rec)
 		if err != nil {
-			return err
+			return fmt.Errorf("unable to retrieve transaction %v: "+
```
halseth: distinguish the two by having this read `unable to read raw tx record: %v`?

wpaulino: Good idea, done.

PR #655, opened 2019-10-28
Fix broadcast existing transaction

roeierez:
This PR ensures the wallet won't enter an inconsistent state by exiting early in the `addReleventTx` flow in case the transaction is new and it already exists in the wallet.
Without this fix, In the case the existing transaction is already confirmed on-chain the flow updates the `bucketUnminedCredits` but without adding en entry also to the `bucketUnmined` resulting in
inconsistent state.
One example for error due to this state is at this point https://github.com/btcsuite/btcwallet/blob/master/wtxmgr/tx.go#L767
where the next line passes nil to the `readRawTxRecord`. When this flow is executed on startup it causes a panic in lnd.
I verified this fixes the issue in lnd: https://github.com/lightningnetwork/lnd/issues/3640
where the sweeper tries to rebroadcast the same transaction that was already confirmed.

### DIFF
--- a/wtxmgr/tx.go
+++ b/wtxmgr/tx.go
@@ -404,7 +404,9 @@ func (s *Store) addCredit(ns walletdb.ReadWriteBucket, rec *TxRecord, block *Blo
 		if existsRawUnminedCredit(ns, k) != nil {
 			return false, nil
 		}
-		if existsRawUnspent(ns, k) != nil {
+		if _, tv := latestTxRecord(ns, &rec.Hash); tv != nil {
+			log.Tracef("Ignoring credit for existing confirmed transaction %v",
+				rec.Hash.String())
 			return false, nil
 		}
 		v := valueUnminedCredit(btcutil.Amount(rec.MsgTx.TxOut[index].Value), change)
--- a/wtxmgr/tx_test.go
+++ b/wtxmgr/tx_test.go
@@ -162,7 +162,7 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: 0,
 			unc: btcutil.Amount(TstRecvTx.MsgTx().TxOut[0].Value),
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
+				{
 					Hash:  *TstRecvTx.Hash(),
 					Index: 0,
 				}: {},
@@ -189,7 +189,7 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: 0,
 			unc: btcutil.Amount(TstRecvTx.MsgTx().TxOut[0].Value),
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
+				{
 					Hash:  *TstRecvTx.Hash(),
 					Index: 0,
 				}: {},
@@ -216,7 +216,7 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: btcutil.Amount(TstRecvTx.MsgTx().TxOut[0].Value),
 			unc: 0,
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
+				{
 					Hash:  *TstRecvTx.Hash(),
 					Index: 0,
 				}: {},
@@ -241,7 +241,7 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: btcutil.Amount(TstRecvTx.MsgTx().TxOut[0].Value),
 			unc: 0,
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
+				{
 					Hash:  *TstRecvTx.Hash(),
 					Index: 0,
 				}: {},
@@ -257,7 +257,7 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: 0,
 			unc: btcutil.Amount(TstRecvTx.MsgTx().TxOut[0].Value),
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
+				{
 					Hash:  *TstRecvTx.Hash(),
 					Index: 0,
 				}: {},
@@ -284,7 +284,7 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: btcutil.Amount(TstDoubleSpendTx.MsgTx().TxOut[0].Value),
 			unc: 0,
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
+				{
 					Hash:  *TstDoubleSpendTx.Hash(),
 					Index: 0,
 				}: {},
@@ -343,7 +343,7 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: 0,
 			unc: btcutil.Amount(TstSpendingTx.MsgTx().TxOut[0].Value),
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
+				{
 					Hash:  *TstSpendingTx.Hash(),
 					Index: 0,
 				}: {},
@@ -369,11 +369,11 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: 0,
 			unc: btcutil.Amount(TstSpendingTx.MsgTx().TxOut[0].Value + TstSpendingTx.MsgTx().TxOut[1].Value),
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
+				{
 					Hash:  *TstSpendingTx.Hash(),
 					Index: 0,
 				}: {},
-				wire.OutPoint{
+				{
 					Hash:  *TstSpendingTx.Hash(),
 					Index: 1,
 				}: {},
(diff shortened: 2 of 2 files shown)


wpaulino:
Thanks for taking the time to look into this @roeierez! I read the background you included in https://github.com/lightningnetwork/lnd/issues/3640, but I'm not sure if these changes will actually address it. The reason why is because a similar check was previously added to mitigate this (https://github.com/btcsuite/btcwallet/pull/534/commits/fbe82c3531344a0422bc5454f82ed44c83e93124), but perhaps there's something else. Are you able to add a test case to verify that this does indeed solve the issue?

roeierez:
> The reason why is because a similar check was previously added to mitigate this ([fbe82c3](https://github.com/btcsuite/btcwallet/commit/fbe82c3531344a0422bc5454f82ed44c83e93124))

@wpaulino  It seems that this commit was overridden by this commit https://github.com/btcsuite/btcwallet/pull/633/commits/44d818d81334f55059c909cb77f733f99cf031e6 which introduces the problem. Now the transaction is not added to the mempool but added to the `bucketUnminedCredits` bucket.
To me it looks like the main reasons for bugs at this area is that `AddCredit` is not part of `InsertTx` and changes in the former are not consistent with the later but it will require a bigger change which is better to be done after this.
For sure I need to add a test and I see now that based on the current code/history I need to put the check inside `addCredit` and not in the outer flow (although it looks more natural to put it once on start).
Working on this now so let me know if you have a different take on this.

wpaulino:
> It seems that this commit was overridden by this commit 44d818d which introduces the problem. Now the transaction is not added to the mempool but added to the bucketUnminedCredits bucket.

How does https://github.com/btcsuite/btcwallet/commit/44d818d81334f55059c909cb77f733f99cf031e6 introduce the problem? The check in `addCredit` still stands to prevent re-adding it as unconfirmed credit if it's already confirmed.

> To me it looks like the main reasons for bugs at this area is that AddCredit is not part of InsertTx and changes in the former are not consistent with the later but it will require a bigger change which is better to be done after this.

Definitely, this is something that has been brought up in code review before but never got done.

roeierez:
> How does [44d818d](https://github.com/btcsuite/btcwallet/commit/44d818d81334f55059c909cb77f733f99cf031e6) introduce the problem? The check in `addCredit` still stands to prevent re-adding it as unconfirmed credit if it's already confirmed.

I see. 
To my understanding In the case `block = nil` `addCredit` executes two checks
https://github.com/btcsuite/btcwallet/blob/master/wtxmgr/tx.go#L404 -  for existence of unmined credit. 
https://github.com/btcsuite/btcwallet/blob/master/wtxmgr/tx.go#L407 - for existence of confirmed unspent.

One question: In the case the transaction involved is both confirmed and spent, wouldn't both checks will return nil and `putRawUnminedCredit`  will be executed here? https://github.com/btcsuite/btcwallet/blob/master/wtxmgr/tx.go#L411

wpaulino:
> One question: In the case the transaction involved is both confirmed and spent, wouldn't both checks will return nil and putRawUnminedCredit will be executed here?

Yeah, good point! That seems to be the culprit.

roeierez:
> Yeah, good point! That seems to be the culprit.

Great! to verify I looked at my wallet and lnd state and indeed the sweep transaction was spent which reveals now the missing piece regarding lightningnetwork/lnd#3640

Anyway, I am going to see what is the best way to add a check inside `addCredit` and also add a test case.

wpaulino:
I also found another bug that helped mask the issue. The unconfirmed inputs are not removed from the store in [`deleteUnminedTx`](https://github.com/btcsuite/btcwallet/blob/7abdd4f8ad7dcf22c5dd90d123b0c137f93d3879/wtxmgr/tx.go#L281), causing the unconfirmed credits to be skipped due to the [unconfirmed spend check](https://github.com/btcsuite/btcwallet/blob/7abdd4f8ad7dcf22c5dd90d123b0c137f93d3879/wtxmgr/tx.go#L754). I'll address this as a separate PR.

roeierez:
@wpaulino I moved the check inside `addCredit` and added test case.

wtxmgr/tx.go
```diff
@@ -404,7 +404,9 @@ func (s *Store) addCredit(ns walletdb.ReadWriteBucket, rec *TxRecord, block *Blo
 		if existsRawUnminedCredit(ns, k) != nil {
 			return false, nil
 		}
-		if existsRawUnspent(ns, k) != nil {
+		if _, tv := latestTxRecord(ns, &rec.Hash); tv != nil {
+			log.Infof("addCredit for new transaction %v is ignored as it "+
```
wpaulino: Nit: no need for this log. Or perhaps demote to `trace` and reword to `Ignoring credit for existing confirmed transaction %v`.

wtxmgr/tx_test.go
```diff
@@ -1522,6 +1522,83 @@ func TestInsertMempoolTxAlreadyConfirmed(t *testing.T) {
...
+	}
+	amt := int64(1e7)
+	spend := spendOutput(&cbRec.Hash, 0, amt)
+	spendRec, err := NewTxRecordFromMsgTx(spend, time.Now())
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	commitDBTx(t, store, db, func(ns walletdb.ReadWriteBucket) {
+
```
wpaulino: Nit: extra newline.

wtxmgr/tx_test.go
```diff
@@ -1522,6 +1522,83 @@ func TestInsertMempoolTxAlreadyConfirmed(t *testing.T) {
...
+	// now we check that there no unminedCredit exists for the original tx.
+	commitDBTx(t, store, db, func(ns walletdb.ReadWriteBucket) {
+		k := canonicalOutPoint(&cbRec.Hash, 0)
+		if existsRawUnminedCredit(ns, k) != nil {
+			t.Fatalf("expected output to not exist " +
+				"in unmined credit bucket")
+		}
+	})
+
```
wpaulino: Nit: extra newline.

wtxmgr/tx_test.go
```diff
@@ -1522,6 +1522,83 @@ func TestInsertMempoolTxAlreadyConfirmed(t *testing.T) {
...
+			t.Fatal(err)
+		}
+
+		// We now adding the original transaction as mempool to simulate
+		// a real case where trying to broadcast a tx after it has been
+		// confirmed and spent
+		if err := store.InsertTx(ns, cbRec, nil); err != nil {
+			t.Fatal(err)
+		}
+
```
wpaulino: Nit: extra newline.

wtxmgr/tx_test.go
```diff
@@ -1522,6 +1522,83 @@ func TestInsertMempoolTxAlreadyConfirmed(t *testing.T) {
...
+func TestInsertMempoolTxAfterSpentOutput(t *testing.T) {
+	t.Parallel()
+
+	store, db, teardown, err := testStore()
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer teardown()
+
+	// First we add a confirmed transaction to the wallet
```
wpaulino: Nit: end comments with period. Couple other cases below.

wtxmgr/tx_test.go
```diff
@@ -162,7 +162,7 @@ func TestInsertsCreditsDebitsRollbacks(t *testing.T) {
 			bal: 0,
 			unc: btcutil.Amount(TstRecvTx.MsgTx().TxOut[0].Value),
 			unspents: map[wire.OutPoint]struct{}{
-				wire.OutPoint{
```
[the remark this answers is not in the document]
roeierez: Yes, I ran gofmt.

wtxmgr/tx.go
roeierez: Fixed.

wtxmgr/tx_test.go
roeierez: Fixed.

wtxmgr/tx_test.go
roeierez: Fixed.

wtxmgr/tx_test.go
roeierez: Fixed.

roeierez:
Squashed.

wtxmgr/tx_test.go
```diff
@@ -1522,6 +1522,80 @@ func TestInsertMempoolTxAlreadyConfirmed(t *testing.T) {
 	checkStore()
 }
 
+func TestInsertMempoolTxAfterSpentOutput(t *testing.T) {
```
wpaulino: One last thing: would be nice to provide a small description for the test as a comment like the other tests within this file.

roeierez: Done.

wpaulino:
@roeierez here's the issue I was referring to by [my previous comment](https://github.com/btcsuite/btcwallet/pull/655#issuecomment-547221417): https://github.com/btcsuite/btcwallet/pull/658.

wtxmgr/tx_test.go
```diff
@@ -1522,6 +1522,82 @@ func TestInsertMempoolTxAlreadyConfirmed(t *testing.T) {
...
+		err := store.InsertTx(ns, spendRec, &b101)
+		if err != nil {
+			t.Fatal(err)
+		}
+		err = store.AddCredit(ns, spendRec, &b101, 0, false)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		// We now adding the original transaction as mempool to simulate
```
halseth: do this in a separate db transaction?

PR #660, opened 2019-11-08
build+chain: reuse buffers when reading ZMQ messages from bitcoind

wpaulino:
`bitcoind` notifies transactions once they're accepted into the mempool and once they're confirmed in a block. Previously, reading a message from ZMQ would allocate a buffer with the size of the message. This can cause nodes to perform a large number of allocations within a small amount periodically (3000 300B allocations every 10 mins on average), which can cause a lot of GC pressure on lower resourced nodes. To remedy this, we introduce two static buffers, one for blocks and another for transactions, that will be reused for every message read. Each is constrained by its maximum expected size.

### DIFF
--- a/chain/bitcoind_conn.go
+++ b/chain/bitcoind_conn.go
@@ -16,6 +16,16 @@ import (
 	"github.com/lightninglabs/gozmq"
 )
 
+const (
+	// rawBlockZMQCommand is the command used to receive raw block
+	// notifications from bitcoind through ZMQ.
+	rawBlockZMQCommand = "rawblock"
+
+	// rawTxZMQCommand is the command used to receive raw transaction
+	// notifications from bitcoind through ZMQ.
+	rawTxZMQCommand = "rawtx"
+)
+
 // BitcoindConn represents a persistent client connection to a bitcoind node
 // that listens for events read from a ZMQ connection.
 type BitcoindConn struct {
@@ -79,7 +89,7 @@ func NewBitcoindConn(chainParams *chaincfg.Params,
 	// concern to ensure one type of event isn't dropped from the connection
 	// queue due to another type of event filling it up.
 	zmqBlockConn, err := gozmq.Subscribe(
-		zmqBlockHost, []string{"rawblock"}, zmqPollInterval,
+		zmqBlockHost, []string{rawBlockZMQCommand}, zmqPollInterval,
 	)
 	if err != nil {
 		return nil, fmt.Errorf("unable to subscribe for zmq block "+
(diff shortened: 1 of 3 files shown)


chain/bitcoind_conn.go
```diff
@@ -16,6 +16,28 @@ import (
...
+	// notifications from bitcoind through ZMQ.
+	rawTxZMQCommand = "rawtx"
+
+	// maxRawBlockSize is the maximum size for a raw block received from
+	// bitcoind through ZMQ.
+	maxRawBlockSize = 4e6
+
+	// maxRawTxSize is the maximum size for a raw standard transaction
+	// received from bitcoind through ZMQ.
+	maxRawTxSize = 4e5
```
cfromknecht: is there any harm in just using another 4MB? a malicious miner could sever bitcoind communication simultaneously if they violate policy, i would vote for safety here.

chain/bitcoind_conn.go
```diff
@@ -254,9 +297,15 @@ func (c *BitcoindConn) txEventHandler() {
 		default:
 		}
 
+		// Set up the buffers we expect our messages to consume. ZMQ
+		// messages from bitcoind include three parts: the command, the
+		// data, and the sequence number.
+		var command [len(rawTxZMQCommand)]byte
+		var seqNum [seqNumLen]byte
```
cfromknecht: lift these up to where `data` is initialized? can all use a single var statement, same for blocks

chain/bitcoind_conn.go
wpaulino: > is there any harm in just using another 4MB?

Not at all.

> a malicious miner could sever bitcoind communication simultaneously if they violate policy, i would vote for safety here.

Seems like we'd just be unable to deserialize the transaction, which would only be crucial if the transaction was somehow relevant to us (unlikely?). Definitely an easy way to mitigate such harm by just increasing this to 4MB though, fixed.

chain/bitcoind_conn.go
wpaulino: Done.

chain/bitcoind_conn.go
```diff
@@ -164,6 +186,21 @@ func (c *BitcoindConn) blockEventHandler() {
...
+	//
+	// We'll allocate a fixed data slice that we'll reuse when reading
+	// blocks from bitcoind through ZMQ. There's no need to recycle this
+	// slice (zero out) after using it, as further reads will overwrite the
+	// slice and we'll only be deserializing the bytes needed.
+	var (
+		command [len(rawBlockZMQCommand)]byte
+		seqNum  [seqNumLen]byte
+		data    = make([]byte, maxRawBlockSize)
+		bufs    = [][]byte{command[:], data, seqNum[:]}
```
cfromknecht: If I’m not mistaken, `bufs` needs to be reinstantiated with every for loop, otherwise the slices can be shortened over time

wpaulino: Yep, fixed.

Roasbeef:
Can be updated now that the dependent `zmq` PR has been merged in.

wpaulino:
@Roasbeef rebased.

PR #661, opened 2019-11-09
Fix #652: Create walletdb with no-freelist-sync option

### LINKED ISSUE #652: no-freelist-sync option not passed to RPC when creating a temp wallet via btcd
I believe recent merge https://github.com/btcsuite/btcwallet/pull/648 broke btcd when trying to create a temp wallet. It looks like a RPC call to btcwallet fails with this error message: 

```
Unable to create wallet: invalid arguments to bdb.Create -- expected database path and no-freelist-sync option
```

I believe btcwallet should be less strict [here](https://github.com/btcsuite/btcwallet/pull/648/files#diff-7a56bf5fd9751a56fbe6ea8277a16650R19) and issue a warning only or alternatively code path triggering this should be made aware of this new option.

For reference see:
https://github.com/btcsuite/btcwallet/pull/648#issuecomment-539237987
https://travis-ci.org/darwin/simverse/builds/594826741#L4574

Moved from: https://github.com/btcsuite/btcd/issues/1473

angrypie:
Fix #652

### DIFF
--- a/walletsetup.go
+++ b/walletsetup.go
@@ -219,7 +219,7 @@ func createSimulationWallet(cfg *config) error {
 	fmt.Println("Creating the wallet...")
 
 	// Create the wallet database backed by bolt db.
-	db, err := walletdb.Create("bdb", dbPath)
+	db, err := walletdb.Create("bdb", dbPath, true)
 	if err != nil {
 		return err
 	}


walletsetup.go
```diff
@@ -219,7 +219,7 @@ func createSimulationWallet(cfg *config) error {
 	fmt.Println("Creating the wallet...")
 
 	// Create the wallet database backed by bolt db.
-	db, err := walletdb.Create("bdb", dbPath)
+	db, err := walletdb.Create("bdb", dbPath, true)
```
Roasbeef: Rather than _always_ set it, this should be read as a new config argument.

angrypie: Do you mean creating new such argument in the application config?

PR #667, opened 2019-12-14
Create watch-only address managers and accounts

devrandom:
This PR allows the creation of managers and accounts that are watch-only.  The state of the database after creation would be identical to the state after calling `Manager.ConvertToWatchingOnly`, assuming accounts with the right xpubs were created in the former case.

Of course, creating a watch-only database is safer, since you don't have to worry about private data ending up in a swap file or the system having been compromised before creation.

### DIFF
--- a/waddrmgr/db.go
+++ b/waddrmgr/db.go
@@ -850,6 +850,7 @@ func forEachAccount(ns walletdb.ReadBucket, scope *KeyScope,
 }
 
 // fetchLastAccount retrieves the last account from the database.
+// If no accounts, returns twos-complement representation of -1, so that the next account is zero
 func fetchLastAccount(ns walletdb.ReadBucket, scope *KeyScope) (uint32, error) {
 	scopedBucket, err := fetchReadScopeBucket(ns, scope)
 	if err != nil {
@@ -859,6 +860,9 @@ func fetchLastAccount(ns walletdb.ReadBucket, scope *KeyScope) (uint32, error) {
 	metaBucket := scopedBucket.NestedReadBucket(metaBucketName)
 
 	val := metaBucket.Get(lastAccountName)
+	if val == nil {
+		return (1 << 32) - 1, nil
+	}
 	if len(val) != 4 {
 		str := fmt.Sprintf("malformed metadata '%s' stored in database",
 			lastAccountName)
--- a/waddrmgr/error.go
+++ b/waddrmgr/error.go
@@ -33,10 +33,6 @@ var (
 	// errWatchingOnly is the common error description used for the
 	// ErrWatchingOnly error code.
 	errWatchingOnly = "address manager is watching-only"
-
-	// errMustBeWatchingOnly is the common error description used for the
-	// ErrMustBeWatchingOnly error code.
-	errMustBeWatchingOnly = "address manager must be watching-only"
 )
 
 // ErrorCode identifies a kind of error.
@@ -94,11 +90,6 @@ const (
 	// account manager to have access to private data, was requested on
 	// a watching-only account manager.
 	ErrWatchingOnly
-
-	// ErrMustBeWatchingOnly indicates that an operation, which requires the
-	// account manager to be watching only, was requested on
-	// a non-watching-only account manager.
-	ErrMustBeWatchingOnly
 
 	// ErrInvalidAccount indicates that the requested account is not valid.
 	ErrInvalidAccount
--- a/waddrmgr/error.go
+++ b/waddrmgr/error.go
@@ -33,6 +33,10 @@ var (
 	// errWatchingOnly is the common error description used for the
 	// ErrWatchingOnly error code.
 	errWatchingOnly = "address manager is watching-only"
+
+	// errMustBeWatchingOnly is the common error description used for the
+	// ErrMustBeWatchingOnly error code.
+	errMustBeWatchingOnly = "address manager must be watching-only"
 )
 
 // ErrorCode identifies a kind of error.
@@ -90,6 +94,11 @@ const (
 	// account manager to have access to private data, was requested on
 	// a watching-only account manager.
 	ErrWatchingOnly
+
+	// ErrMustBeWatchingOnly indicates that an operation, which requires the
+	// account manager to be watching only, was requested on
+	// a non-watching-only account manager.
+	ErrMustBeWatchingOnly
 
 	// ErrInvalidAccount indicates that the requested account is not valid.
 	ErrInvalidAccount
--- a/waddrmgr/manager.go
+++ b/waddrmgr/manager.go
@@ -446,46 +446,51 @@ func (m *Manager) NewScopedKeyManager(ns walletdb.ReadWriteBucket,
 
 	var rootPriv *hdkeychain.ExtendedKey
 	if !m.watchingOnly {
-		// If the manager is locked, then we can't create a new scoped manager.
+		// If the manager is locked, then we can't create a new scoped
+		// manager.
 		if m.locked {
 			return nil, managerError(ErrLocked, errLocked, nil)
 		}
 
-		// Now that we know the manager is unlocked, we'll need to fetch the
-		// root master HD private key. This is required as we'll be attempting
-		// the following derivation: m/purpose'/cointype'
+		// Now that we know the manager is unlocked, we'll need to
+		// fetch the root master HD private key. This is required as
+		// we'll be attempting the following derivation:
+		// m/purpose'/cointype'
 		//
-		// Note that the path to the coin type is requires hardened derivation,
-		// therefore this can only be done if the wallet's root key hasn't been
-		// neutered.
+		// Note that the path to the coin type is requires hardened
+		// derivation, therefore this can only be done if the wallet's
+		// root key hasn't been neutered.
 		masterRootPrivEnc, _, err := fetchMasterHDKeys(ns)
 		if err != nil {
 			return nil, err
 		}
 
-		// If the master root private key isn't found within the database, but
-		// we need to bail here as we can't create the cointype key without the
-		// master root private key.
+		// If the master root private key isn't found within the
+		// database, but we need to bail here as we can't create the
+		// cointype key without the master root private key.
 		if masterRootPrivEnc == nil {
 			return nil, managerError(ErrWatchingOnly, "", nil)
 		}
 
-		// Before we can derive any new scoped managers using this key, we'll
-		// need to fully decrypt it.
-		serializedMasterRootPriv, err := m.cryptoKeyPriv.Decrypt(masterRootPrivEnc)
+		// Before we can derive any new scoped managers using this
+		// key, we'll need to fully decrypt it.
+		serializedMasterRootPriv, err :=
+			m.cryptoKeyPriv.Decrypt(masterRootPrivEnc)
 		if err != nil {
-			str := fmt.Sprintf("failed to decrypt master root serialized private key")
+			str := fmt.Sprintf("failed to decrypt master root " +
+				"serialized private key")
 			return nil, managerError(ErrLocked, str, err)
 		}
 
-		// Now that we know the root priv is within the database, we'll decode
-		// it into a usable object.
+		// Now that we know the root priv is within the database,
+		// we'll decode it into a usable object.
 		rootPriv, err = hdkeychain.NewKeyFromString(
 			string(serializedMasterRootPriv),
 		)
 		zero.Bytes(serializedMasterRootPriv)
 		if err != nil {
-			str := fmt.Sprintf("failed to create master extended private key")
+			str := fmt.Sprintf("failed to create master extended " +
+				"private key")
 			return nil, managerError(ErrKeyChain, str, err)
 		}
 	}
@@ -515,9 +520,9 @@ func (m *Manager) NewScopedKeyManager(ns walletdb.ReadWriteBucket,
 	}
 
 	if !m.watchingOnly {
-		// With the database state created, we'll now derive the cointype key
-		// using the master HD private key, then encrypt it along with the
-		// first account using our crypto keys.
+		// With the database state created, we'll now derive the
+		// cointype key using the master HD private key, then encrypt
+		// it along with the first account using our crypto keys.
 		err = createManagerKeyScope(
 			ns, scope, rootPriv, m.cryptoKeyPub, m.cryptoKeyPriv,
 		)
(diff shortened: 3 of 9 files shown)


waddrmgr/manager.go
```diff
@@ -532,6 +532,61 @@ func (m *Manager) NewScopedKeyManager(ns walletdb.ReadWriteBucket, scope KeyScop
 	return m.scopedManagers[scope], nil
 }
 
+// NewWatchingOnlyScopedKeyManager creates a new watching-only scoped key manager
+// from the root manager.
+// A scoped key manager is a sub-manager that only has the coin type key of a
+// particular coin type and BIP0043 purpose. This is useful as it enables
+// callers to create an arbitrary BIP0043 like schema with a stand alone
+// manager.
+func (m *Manager) NewWatchingOnlyScopedKeyManager(ns walletdb.ReadWriteBucket, scope KeyScope,
```
devrandom: This could be merged with `NewScopedKeyManager`, with a conditional on `watchingOnly`.  WDYT?

Roasbeef: SGTM, there's a ton of duplication as is.

waddrmgr/db.go
```diff
@@ -859,6 +860,9 @@ func fetchLastAccount(ns walletdb.ReadBucket, scope *KeyScope) (uint32, error) {
 	metaBucket := scopedBucket.NestedReadBucket(metaBucketName)
 
 	val := metaBucket.Get(lastAccountName)
+	if val == nil {
```
Roasbeef: Why wasn't this needed before? Or is this for the case where we create the new wallet as pure watch-only?

waddrmgr/manager.go
```diff
@@ -532,6 +532,61 @@ func (m *Manager) NewScopedKeyManager(ns walletdb.ReadWriteBucket, scope KeyScop
...
+	addrSchema ScopeAddrSchema) (*ScopedKeyManager, error) {
+
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	// Now we'll fetch the scope bucket
```
Roasbeef: Columns should be wrapped to 80 chars, but I think this'll fall out in any case due to the above comment.

waddrmgr/manager.go
```diff
@@ -1820,3 +1875,103 @@ func Create(ns walletdb.ReadWriteBucket, seed, pubPassphrase, privPassphrase []b
...
+// All public keys and information are protected by secret keys
+// derived from the provided public passphrases.  The public
+// passphrase is required on subsequent opens of the address manager.
+//
+// If a config structure is passed to the function, that configuration will
+// override the defaults.
+//
+// A ManagerError with an error code of ErrAlreadyExists will be returned the
+// address manager already exists in the specified namespace.
+func CreateWatchOnly(ns walletdb.ReadWriteBucket,
```
Roasbeef: Similar to the new scope key manager method, there's a lot of duplication here with the `Create` method, as a result I think the two of them can be heavily consolidated.

waddrmgr/manager_test.go
```diff
@@ -1067,6 +1068,10 @@ func testMarkUsed(tc *testContext) bool {
 		},
 	}
 
+	if !doScript {
+		tests = tests[0:1]
```
Roasbeef: Alternatively, we can add a new `bool` to the table tests above, and skip them below based on that bool.

waddrmgr/manager_test.go
```diff
@@ -1906,6 +1931,146 @@ func TestManager(t *testing.T) {
 	}
 }
 
+// TestManagerNewWatchOnly is similar to TestManager, but uses the
+// CreateWatchOnly to test watch-only wallets created from export
```
Roasbeef: export -> exported

waddrmgr/scoped_manager.go
```diff
@@ -1326,6 +1342,74 @@ func (s *ScopedKeyManager) newAccount(ns walletdb.ReadWriteBucket,
 	return putLastAccount(ns, &s.scope, account)
 }
 
+// NewAccountWatchingOnly is similar to NewAccount, but for watch-only wallets.
+func (s *ScopedKeyManager) NewAccountWatchingOnly(ns walletdb.ReadWriteBucket, name string,
+	pubKey *hdkeychain.ExtendedKey) (uint32, error) {
+	if !s.rootManager.WatchOnly() {
```
Roasbeef: This should be removed, as otherwise it means that users are unable to import _additional_ keys into an existing full wallet which IMO is a compelling use case.

waddrmgr/manager.go
guggero: Yes, that would definitely be preferable. You could split the content into multiple smaller methods and call them depending on the flag. This would avoid large chunks of code inside if conditions.

waddrmgr/manager.go
guggero: Agreed. This will also make it possible to use table driven tests and reduce code duplication there. See comment further below.

waddrmgr/manager_test.go
```diff
@@ -9,6 +9,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"github.com/btcsuite/btcutil/hdkeychain"
```
guggero: nit: import needs to be grouped with the other external imports

waddrmgr/manager_test.go
guggero: Can this be consolidated with `TestManager` by converting to table driven test? Similar to `testImportScript` where there are two test cases with different settings but same execution code.

waddrmgr/manager_test.go
```diff
@@ -2454,10 +2619,77 @@ func TestNewRawAccount(t *testing.T) {
...
+		return scopedMgr.NewRawAccountWatchingOnly(ns, accountNum, accountKey)
+	})
+	if err != nil {
+		t.Fatalf("unable to create new account: %v", err)
+	}
+
+	testNewRawAccount(t, mgr, db, accountNum, scopedMgr)
+}
+
+func testNewRawAccount(t *testing.T, mgr *Manager, db walletdb.DB, accountNum uint32, scopedMgr *ScopedKeyManager) {
```
guggero: nit: line too long (same for other changes in the PR).

waddrmgr/db.go
ksedgwic: I think this only occurs when the wallet is created as pure watch-only.

waddrmgr/manager.go
ksedgwic: Fixed.

waddrmgr/manager.go
ksedgwic: Addressed.

waddrmgr/manager.go
ksedgwic: Fixed.

waddrmgr/manager_test.go
ksedgwic: Fixed.

waddrmgr/manager_test.go
ksedgwic: Fixed.

waddrmgr/scoped_manager.go
ksedgwic: Done.

waddrmgr/manager_test.go
ksedgwic: Fixed.

waddrmgr/manager_test.go
ksedgwic: Fixed.

waddrmgr/manager.go
devrandom: Let us know if looks good.

waddrmgr/db.go
guggero: Ah, because no default accounts are created, right.

waddrmgr/scoped_manager.go
```diff
@@ -1326,6 +1342,70 @@ func (s *ScopedKeyManager) newAccount(ns walletdb.ReadWriteBucket,
 	return putLastAccount(ns, &s.scope, account)
 }
 
+// NewAccountWatchingOnly is similar to NewAccount, but for watch-only wallets.
+func (s *ScopedKeyManager) NewAccountWatchingOnly(ns walletdb.ReadWriteBucket, name string,
```
guggero: nit: line too long

devrandom:
> Maybe it's a bit late to suggest this, but while figuring out how exactly one would use this new functionality I couldn't help but wonder if it wouldn't be possible to create a new manager with the master xpub directly instead of later adding accounts.

Account derivation is hardened, so it's not possible to derive account xpubs from the parent(s).

guggero:
Right, I didn't remember that. Then I think the current design of the PR makes a lot of sense :+1:

waddrmgr/error.go
```diff
@@ -33,6 +33,10 @@ var (
 	// errWatchingOnly is the common error description used for the
 	// ErrWatchingOnly error code.
 	errWatchingOnly = "address manager is watching-only"
+
+	// errMustBeWatchingOnly is the common error description used for the
+	// ErrMustBeWatchingOnly error code.
+	errMustBeWatchingOnly = "address manager must be watching-only"
```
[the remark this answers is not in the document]
ksedgwic: Unneeded; removed.

waddrmgr/error.go
```diff
@@ -91,6 +95,11 @@ const (
 	// a watching-only account manager.
 	ErrWatchingOnly
 
+	// ErrMustBeWatchingOnly indicates that an operation, which requires the
+	// account manager to be watching only, was requested on
+	// a non-watching-only account manager.
+	ErrMustBeWatchingOnly
```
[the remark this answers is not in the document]
ksedgwic: Unneeded; removed.

waddrmgr/manager.go
```diff
@@ -430,53 +430,56 @@ func (m *Manager) Close() {
...
-	// the following derivation: m/purpose'/cointype'
-	//
-	// Note that the path to the coin type is requires hardened derivation,
-	// therefore this can only be done if the wallet's root key hasn't been
-	// neutered.
-	masterRootPrivEnc, _, err := fetchMasterHDKeys(ns)
-	if err != nil {
-		return nil, err
-	}
+		// Now that we know the manager is unlocked, we'll need to fetch the
```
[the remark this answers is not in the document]
ksedgwic: Fixed.

waddrmgr/manager.go
```diff
@@ -1743,52 +1725,101 @@ func Create(ns walletdb.ReadWriteBucket, seed, pubPassphrase, privPassphrase []b
...
-	if err != nil {
-		str := "failed to derive master extended key"
-		return managerError(ErrKeyChain, str, err)
-	}
-	rootPubKey, err := rootKey.Neuter()
-	if err != nil {
-		str := "failed to neuter master extended key"
-		return managerError(ErrKeyChain, str, err)
-	}
+		// Generate the private passphrase salt.  This is used when hashing
```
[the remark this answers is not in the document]
ksedgwic: Fixed.

waddrmgr/manager.go
```diff
@@ -1631,11 +1636,18 @@ func createManagerKeyScope(ns walletdb.ReadWriteBucket,
...
+// Create creates a new address manager in the given namespace.
+//
+// The seed must conform to the standards described in
+// hdkeychain.NewMaster and will be used to create the master root
+// node from which all hierarchical deterministic addresses are
+// derived.  This allows all chained addresses in the address manager
+// to be recovered by using the same seed.
+//
+// If the provided seed value is nil the address manager will be
+// created in watchingOnly mode in which case no default account or
```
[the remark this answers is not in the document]
ksedgwic: Fixed.

waddrmgr/manager.go
```diff
@@ -1631,11 +1636,18 @@ func createManagerKeyScope(ns walletdb.ReadWriteBucket,
...
+// scoped manager are created - it is up to the caller to create a new
+// one with NewAccount and NewScopedKeyManager
```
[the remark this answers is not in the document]
ksedgwic: Fixed.

waddrmgr/scoped_manager.go
ksedgwic: Fixed.

ksedgwic:
I think everything is addressed, pls let me know if I missed something ...

guggero:
Thanks, looks better now!

Just one last question: I started implementing something on top of this and noticed that currently there is no way to create a new _wallet_ wit a `nil` seed, because by default a random seed is created in the `wallet.Create` function. Maybe I still don't understand how you intend to use this functionality. Is the idea to create a new wallet normally, create all the key scopes and accounts, then turn it into a watch-only wallet?

ksedgwic:
No, the idea is definitely to create the wallet without a seed and add accounts using XPUB from the remote signer.  Investigating ...

ksedgwic:
Oliver, I think we missed that.  We'll add that to this PR straightaway.  Thanks for catching!

ksedgwic:
Did a force push to rebase on master, fixes coming ...

ksedgwic:
Ok, I've added `Loader.CreateNewWatchingOnlyWallet` and `Wallet.CreateWatchingOnly`, thoughts?

wallet/wallet.go
```diff
@@ -3620,23 +3620,45 @@ func (w *Wallet) Database() walletdb.DB {
...
+// either.
+func CreateWatchingOnly(db walletdb.DB, pubPass []byte,
+	params *chaincfg.Params, birthday time.Time) error {
+
+	return createInternal(
+		db, pubPass, nil, nil, params, birthday, true,
+	)
+}
+
+func createInternal(db walletdb.DB, pubPass, privPass, seed []byte,
```
guggero, line 3643: nit: This can be called just lowercase `create` as not exporting it basically means the same as "internal".

wallet/loader.go
```diff
@@ -100,6 +100,25 @@ func (l *Loader) RunAfterLoad(fn func(*Wallet)) {
...
+// since the wallet is watching-only.
+func (l *Loader) CreateNewWatchingOnlyWallet(pubPassphrase []byte,
+	bday time.Time) (*Wallet, error) {
+
+	return l.createNewWalletInternal(
+		pubPassphrase, nil, nil, bday, true,
+	)
+}
+
+func (l *Loader) createNewWalletInternal(pubPassphrase, privPassphrase,
```
guggero, line 119: nit: This can be called just lowercase `createNewWallet` as not exporting it basically means the same as "internal".

ksedgwic:
Force pushed rebase onto master, about to fix nits ...

wallet/wallet.go
ksedgwic: Done.

wallet/loader.go
ksedgwic: Done.

ksedgwic:
ping, everything look ok with this?

guggero:
cc @Roasbeef.

waddrmgr/manager_test.go
```diff
@@ -2454,10 +2581,145 @@ func TestNewRawAccount(t *testing.T) {
...
+	}
+
+	testNewRawAccount(t, mgr, db, accountNum, scopedMgr)
+}
+
+// TestNewRawAccountHybrid is similar to TestNewRawAccountWatchingOnly
+// except that the manager is created normally with a seed. This test
+// shows that watch-only accounts can be added to managers with
+// non-watch-only accounts.
+func TestNewRawAccountHybrid(t *testing.T) {
```
Roasbeef, line 2656: 👍

Roasbeef:
Y'all can follow the progress for the remaining `lnd` changes to support a watch-only node and a segregated signer here: https://github.com/lightningnetwork/lnd/issues/3943

PR #671, opened 2020-01-04
walletdb: add sequence methods from bbolt to the main bucket interface

Roasbeef:
In this PR, we add the trio of sequence based methods that `bbolt` ships with to the main bucket interface. We do this in order to easily allow `walletdb` as is to be slotted into place where `bbolt` is currently used, without sacrificing any functionality.

Along the way we fix some issues with sub-modules that resulted in packges not being tested against the latest set of sub-modules. We also resolve an issue with `goclean.sh` that prevented any packages which were themselves sub-modules from being run.


halseth:
Replaced by https://github.com/btcsuite/btcwallet/pull/672

PR #672, opened 2020-01-05
walletdb: add sequence methods from bbolt to the main bucket interface

cfromknecht:
In this PR, we add the trio of sequence based methods that `bbolt` ships with to the main bucket interface. We do this in order to easily allow walletdb as is to be slotted into place where bbolt is currently used, without sacrificing any functionality.

Along the way we fix some issues with sub-modules that resulted in packges not being tested against the latest set of sub-modules. We also resolve an issue with `goclean.sh` that prevented any packages which were themselves sub-modules from being run.

### DIFF
--- a/walletdb/bdb/db.go
+++ b/walletdb/bdb/db.go
@@ -342,6 +342,19 @@ func (db *db) Close() error {
 	return convertErr((*bbolt.DB)(db).Close())
 }
 
+// Batch is similar to the package-level Update method, but it will attempt to
+// optismitcally combine the invocation of several transaction functions into a
+// single db write transaction.
+//
+// This function is part of the walletdb.Db interface implementation.
+func (db *db) Batch(f func(tx walletdb.ReadWriteTx) error) error {
+	return (*bbolt.DB)(db).Batch(func(btx *bbolt.Tx) error {
+		interfaceTx := transaction{btx}
+
+		return f(&interfaceTx)
+	})
+}
+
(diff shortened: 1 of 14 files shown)


walletdb/bdb/db.go
```diff
@@ -68,7 +68,7 @@ func (tx *transaction) ReadWriteBucket(key []byte) walletdb.ReadWriteBucket {
 }
 
 func (tx *transaction) CreateTopLevelBucket(key []byte) (walletdb.ReadWriteBucket, error) {
-	boltBucket, err := tx.boltTx.CreateBucket(key)
+	boltBucket, err := tx.boltTx.CreateBucketIfNotExists(key)
```
cfromknecht: is this change relevant to the pr?

wtxmgr/tx_test.go
```diff
@@ -51,7 +51,7 @@ func testDB() (walletdb.DB, func(), error) {
 	if err != nil {
 		return nil, func() {}, err
 	}
-	db, err := walletdb.Create("bdb", filepath.Join(tmpDir, "db"))
+	db, err := walletdb.Create("bdb", filepath.Join(tmpDir, "db"), true)
```
cfromknecht: should these be part of the prior commit?

walletdb/bdb/db.go
Roasbeef: Discussed offline that without this change, there isn't a way to conditionally obtain/create a top-level bucket.

PR #674, opened 2020-01-17
waddrmgr+wallet: use fast scrypt options for regtest and simnet

guggero:
In unit tests we can pass in the fast scrypt options directly.
This is not true for integration tests of external projects that
use btcwallet. But we can speed up these tests by using the fast
scrypt options if the regtest or simnet chain params are used as
they should never be set in production wallet creation.

### DIFF
--- a/waddrmgr/common_test.go
+++ b/waddrmgr/common_test.go
@@ -33,11 +33,7 @@ var (
 
 	// fastScrypt are parameters used throughout the tests to speed up the
 	// scrypt operations.
-	fastScrypt = &ScryptOptions{
-		N: 16,
-		R: 8,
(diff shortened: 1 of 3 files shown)


guggero:
Never mind, sorry... While debugging an error, I found out that we also just can do this in our integration tests:
```
        scryptN, scryptR, scryptP := 16, 8, 1
	waddrmgr.SetSecretKeyGen(func(passphrase *[]byte,
		config *waddrmgr.ScryptOptions) (*snacl.SecretKey, error) {
		return snacl.NewSecretKey(passphrase, scryptN, scryptR, scryptP)
	})
```

PR #677, opened 2020-02-18
chain: wait for bitcoind rpc to warm up

sangaman:
This modifies the `Start` method to handle the `RPC_IN_WARMUP` error code of `-28` by retrying the rpc call to determine the current network every second until it either succeeds or fails with a different code. The current behavior fails and terminates the connection upon receiving this error code. This change allows for connecting to a recently started bitcoind node and starting the client while bitcoind is still warming up.

Related issues: https://github.com/lightningnetwork/lnd/issues/1533 & https://github.com/ExchangeUnion/xud-docker/issues/195

I first tried to approach this issue with changes to the lnd codebase, but the relevant code is what's being modified here. Attempting to call the `Start()` method from lnd a second time, for instance, doesn't work because of the up front check that determines whether `Start()` has already been called and immediately returns if so.

I've tested this locally by pointing btcwallet in go.mod to my local repository and unlocking lnd moments after starting bitcoind. Instead of shutting down, lnd will keep trying bitcoind until it gets a non-error response and then continue with its initialization process. This would save some grief when bringing a node online and starting bitcoind & lnd at roughly the same time.


chain/bitcoind_conn.go
```diff
@@ -139,10 +140,24 @@ func (c *BitcoindConn) Start() error {
...
-		return err
+	for {
+		net, err = c.getCurrentNet()
+		if err == nil {
+			// No error means bitcoind is responsive and we can proceed.
+			break
+		} else if err.(*btcjson.RPCError).Code == -28 {
+			// Error code -28 indicates the bitcoind rpc is warming up.
+			// We will wait and recheck periodically until rpc is ready.
+			time.Sleep(time.Second * 1)
```
halseth: this looks okay, the only thing I'm worried about is bitcoind hanging, always returning this code. In that case it looks like we'll deadlock here.

wpaulino: The alternative would be to crash, which is what happens now. This approach looks good to me, maybe just add a log like `Waiting for bitcoind RPC ready signal` that's only logged once if we ever run into this error code.

sangaman: There could perhaps be some upper limit of retries before returning an error. A couple of minutes perhaps? But yeah the alternative is crashing, and if ever bitcoind is stuck in this rpc warming up status, something is seriously wrong and requires intervention. I can't say I've ever seen bitcoind get stuck in this point, even if there's something wrong like db corruption it'll eventually crash in my experience.

wpaulino: @sangaman in your experience, what has been the longest bitcoind has been in this warmup state? Waiting something like 5 mins sounds good to me.

sangaman: In my experience, even on a low-powered ARM machine this warm-up stage takes about a minute or two by default, rarely much more than than that. However, it's also possible to change the config to increase the amount of checks that are done before starting the bitcoind server, theoretically this could take hours if you had it do very deep/thorough checks.

wpaulino: In that case, I'm fine with blocking indefinitely as long as a log statement exists.

halseth: Agreed :)

Roasbeef:
What's needed to unblock this?

halseth:
Add a log, then I think it's good.

ping @sangaman

sangaman:
Just saw this ping, I'll add a log statement and review the feedback. Sorry this fell off my radar.

sangaman:
I updated this PR to use the newly available `btcjson.ErrRPCInWarmup` variable (which equals `-28`) instead of writing in `-28` and I added logic to log `Waiting for bitcoind RPC to finish warming up...` the first time we see that code and `Bitcoind finished warming up and RPC is ready` after the code goes away. Ping @halseth @wpaulino

halseth (approved):
LGTM 🌡️

chain/bitcoind_conn.go
```diff
@@ -139,10 +140,32 @@ func (c *BitcoindConn) Start() error {
...
-	net, err := c.getCurrentNet()
-	if err != nil {
-		return err
+	btcInWarmup := false
+	for {
+		net, err = c.getCurrentNet()
+		if err == nil {
+			// No error means bitcoind is responsive and we can proceed.
+			break
+		} else if err.(*btcjson.RPCError).Code == btcjson.ErrRPCInWarmup {
```
wpaulino, line 155: We should do a soft type assertion here to prevent a panic if `err` is not of the type `*btcjson.RPCError`:
```
} else if rpcErr, ok := err.(*btcjson.RPCError); ok && rpcErr.Code == btcjson.ErrRPCInWarmup {
```

sangaman: Thanks, makes sense. Done.

wpaulino (approved):

wpaulino:
cc @Roasbeef

sangaman:
Should I continue with this PR? I do think it's a useful feature. It looks like I'll need to resolve some conflicts with pruned node logic.

wpaulino:
@sangaman yes! It's already gotten approvals, @Roasbeef is just the only one with merge access. I can take another look once rebased.

sangaman:
@wpaulino  @Roasbeef  I rebased the changes, the code didn't really need to change I just had to move it to keep up with the updated structure.

chain/bitcoind_conn.go
```diff
@@ -147,9 +147,29 @@ func NewBitcoindConn(cfg *BitcoindConfig) (*BitcoindConn, error) {
 	}
 
 	// Verify that the node is running on the expected network.
-	net, err := getCurrentNet(client)
-	if err != nil {
-		return nil, err
+	var (
+		net wire.BitcoinNet
+	)
```
wpaulino, line 152: Nit: this can be collapsed into a single line `var net wire.BitcoinNet`.

chain/bitcoind_conn.go
```diff
@@ -147,9 +147,29 @@ func NewBitcoindConn(cfg *BitcoindConfig) (*BitcoindConn, error) {
...
+	btcInWarmup := false
+	for {
+		net, err = getCurrentNet(client)
+		if err == nil {
+			// No error means bitcoind is responsive and we can proceed.
+			break
+		} else if rpcErr, ok := err.(*btcjson.RPCError); ok && rpcErr.Code == btcjson.ErrRPCInWarmup {
```
wpaulino, line 159: Nit: line exceeds 80 character limit.

chain/bitcoind_conn.go
sangaman: Thanks, a leftover from when this also declared `err` which didn't exactly work any more since it's being used above.

chain/bitcoind_conn.go
sangaman: If I split this into two lines, for instance by moving everything after the semicolon to its own line, then `make lint` doesn't pass and tells me to run `gofmt`, but if I do that it moves this `else if` statement back to one line. So I'm not sure what to do, it feels like I'm fighting against the style rules defined by the project.

wpaulino: So the `else if` can just become an `if` since the clause above breaks out. We can then split the line at `&&`.

sangaman: Thank you for that bit of hand holding, that works out.

chain/bitcoind_conn.go
schildbach: Yes, isn't bitcoind in that warmup state for the entire initial blockchain sync (which can take days, or more)?

I think it should wait indefinitely. However the sleep could maybe be increased slowly to a max of 10 minutes or so.

sangaman: No I don't think so, the warm up state is just while it's verifying the db and blocks on disk. Once it starts accepting and validating new blocks, it's past the warm up state and will respond to rpc requests normally.

PR #681, opened 2020-03-13
wallet: wait for chain sync on Neutrino recovery

guggero:
Normally the wallet doesn't wait for the chain backend to be synced
on regtest/simnet because there we cannot be certain if we are at
the chain tip, as there are no other nodes to compare to. For
Neutrino, this is a bit different because we rely on the cfheader
server to tell us what it thinks the chain tip is. For a wallet
recovery on Neutrino we therefore need to make sure we are at least
synced up to what the server thinks is the tip.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -336,10 +336,21 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
 		return err
 	}
 
+	// Neutrino relies on the information given to it by the cfheader server
+	// so it knows exactly whether it's synced up to the server's state or
+	// not, even on dev chains. To recover a Neutrino wallet, we need to
+	// make sure it's synced before we start scanning for addresses,
+	// otherwise we might miss some if we only scan up to its current sync
+	// point.
+	neutrinoRecovery := chainClient.BackEnd() == "neutrino" &&
+		w.recoveryWindow > 0
+
 	// We'll wait until the backend is synced to ensure we get the latest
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -653,6 +664,9 @@ func (w *Wallet) recovery(chainClient chain.Interface,
 	if err != nil {
 		return err
 	}
+	log.Debugf("Chain backend has best block %d which is current: %v",
+		bestHeight, chainClient.IsCurrent())
+	chainClient.NotifyBlocks()
```
halseth: why is this change needed?

wallet/wallet.go
```diff
@@ -336,10 +336,21 @@ func (w *Wallet) syncWithChain(birthdayStamp *waddrmgr.BlockStamp) error {
...
+		w.recoveryWindow > 0
+
 	// We'll wait until the backend is synced to ensure we get the latest
 	// MaxReorgDepth blocks to store. We don't do this for development
-	// environments as we can't guarantee a lively chain.
-	if !w.isDevEnv() {
+	// environments as we can't guarantee a lively chain, except for
+	// Neutrino, where the cfheader server tells us what it believes the
+	// chain tip is.
+	if !w.isDevEnv() || neutrinoRecovery {
```
halseth: Should consider removing this special casing at one point, but for now this is an okay fix I think.

wallet/wallet.go
guggero: Oops, that was some debugging/trial-and-error code I committed by accident. Removed.

PR #682, opened 2020-03-17
wallet: rescan and recover addresses in all relevant key scopes

wpaulino:
Previously, the wallet would determine the key scope to use for change addresses by locating the one compatible with P2WPKH addresses, but this wasn't always safe like in the case when multiple key scopes that supported these addresses existed within the address manager, leading the change address to be created outside of the intended key scope.

Wallets affected by this now need to ensure they scan the chain when performing rescans or recovery attempts for addresses within _all_ existing key scopes, rather than just the default ones, to reflect their proper balance. For newer wallets not affected by this, they should only scan for addresses within their default key scopes.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -2946,7 +2946,7 @@ func (w *Wallet) NewChangeAddress(account uint32,
 	err = walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
 		addrmgrNs := tx.ReadWriteBucket(waddrmgrNamespaceKey)
 		var err error
-		addr, err = w.newChangeAddress(addrmgrNs, account)
+		addr, err = w.newChangeAddress(addrmgrNs, account, scope)
 		return err
 	})
 	if err != nil {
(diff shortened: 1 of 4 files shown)


wallet/wallet.go
```diff
@@ -303,27 +303,71 @@ func (w *Wallet) SetChainSynced(synced bool) {
...
+// created within a non-default key scope if it shared an address schema with a
+// default one. In order to ensure older wallets reflect their proper balance,
+// we gate whether they should scan for all existing key scopes by their
+// birthday height.
+func (w *Wallet) shouldOnlyScanDefaultKeyScopes(birthday *waddrmgr.BlockStamp) bool {
+	const (
+		// Wallets with a birthday after this height are not susceptible
+		// to the bug described, so they can safely only scan the
+		// default key scopes.
+		mainNetHeight = 623000
```
Roasbeef: How were these heights obtained? The heights when the prior active addr change landed in master?

Roasbeef: Instead, I think we should just have this be based off of a database migration version which detects if a change address was ever created under a non-default scope. If so, then we need to check for everything constantly, otherwise we can only start to watch the default scopes from there.

wpaulino:
@Roasbeef I've added a migration with some unit tests. I'll be testing this over the next few days on some nodes, but let me know if this is what you had in mind.

PR #688, opened 2020-03-31
chain: batch filter fetches for neutrino chain client

wpaulino:
This greatly speeds up wallet recovery with a neutrino backend, as we'll avoid unnecessary round trips for each block filter fetch.

### DIFF
--- a/chain/neutrino.go
+++ b/chain/neutrino.go
@@ -199,6 +199,11 @@ func (s *NeutrinoClient) FilterBlocks(
 	// the filter returns a positive match, the full block is then requested
 	// and scanned for addresses using the block filterer.
 	for i, blk := range req.Blocks {
+		// TODO(wilmer): Investigate why polling it still necessary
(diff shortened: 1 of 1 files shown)


chain/neutrino.go
```diff
@@ -199,6 +199,11 @@ func (s *NeutrinoClient) FilterBlocks(
 	// the filter returns a positive match, the full block is then requested
 	// and scanned for addresses using the block filterer.
 	for i, blk := range req.Blocks {
+		// TODO(wilmer): Investigate why polling it still necessary
```
cfromknecht: isn't this because the filter headers can lag behind the block headers?

wpaulino: I'm not sure since the header chains should be caught up to tip before we attempt recovery:
https://github.com/btcsuite/btcwallet/blob/60299130c647f27394e818079e2a0705e6aa8e60/wallet/wallet.go#L351

PR #689, opened 2020-04-03
build: revert update to walletdb v1.3.0

wpaulino:
The v1.3.0 version of the walletdb module contained a bug fix within boltdb to allow the use of go1.14, but it uncovered a GC issue. Rather than reverting the commit itself that upgraded the dependency, we opt to revert the change ourselves since we'll need additional changes to ensure we're pinned to the intended walletdb version (v1.2.0).


go.mod
```diff
@@ -22,7 +22,10 @@ require (
 	google.golang.org/grpc v1.18.0
 )
 
-replace github.com/btcsuite/btcwallet/walletdb => ./walletdb
+// Prevent using the replace directive as it'll point to the latest walletdb
```
Roasbeef, line 25: Why's this needed?

Roasbeef: Nvm.

PR #690, opened 2020-04-03
walletdb: revert bolt v1.3.4

wpaulino:


walletdb/bdb/db.go
```diff
@@ -9,7 +9,7 @@ import (
 	"os"
 
 	"github.com/btcsuite/btcwallet/walletdb"
-	"go.etcd.io/bbolt"
+	"github.com/coreos/bbolt"
```
Roasbeef, line 12: I think this needs to be the other path?

wpaulino: I reverted the commit this time and this was part of it. Seems to build just fine.

PR #696, opened 2020-04-17
wtxmgr: add labels to transactions

### LINKED ISSUE #693: Add User Specified Labels to Published Transactions
#### Feature Request
The ability to add a user-specified label to the transactions that btcwallet publishes is a useful feature for users with the requirement to audit/ reconcile their on-chain activity. This feature request arises from the need to store labels for on chain transactions to allow users to more easily audit their nodes in [lnd](https://github.com/lightningnetwork/lnd/issues/2597).

I would like to work on this myself, but this change would require a database migration, so I’d like to open up discussion before submitting a PR. Specifically whether this is a change that would be welcome in btcwallet, and whether the approach outlined below is suitable. 

#### Proposed Approach
The current schema for writing transactions to disk (copied from wtxmgr/db.go) ends with a variable length field, which means a new field cannot simply be added. 
```
Key:
  [0:32]  Transaction hash (32 bytes)
  [32:36] Block height (4 bytes)
  [36:68] Block hash (32 bytes)
Value:
  [0:8]   Received time (8 bytes)
  [8:]    Serialized transaction (varies)
```

Migration:
* Iterate through each txRecord, reading from disk
* Get length of serialized transaction, and write to disk in LV format:
  ```
  [0:8]           Received time (8 bytes)
  [8:16]          Len(Serialized transaction)
  [16: len(tx)]  Serialized transaction
  ```

This migration will allow for the addition of new fields. In the case of a user provided label, which is optional, a similar LV format would be used:
```
[0:8]           Received time (8 bytes)
[8:16]         Len(Serialized transaction)
[+len(tx)]    Serialized transaction
[+8]:          Len(user label)
[+len(label)]   User Label
```

carlaKC:
This change adds optional, user provided labels to transactions. Labels are written to their own bucket, keyed by transaction hash. This approach was taken to avoid the need for migration in the `TxRecord`. 

For the legacy rpc, the existing `Comment` parameter is used. Some fields also have a `CommentTo` field; I'm unsure what this field is for, and it is not present for all send endpoints so I have left it as unimplemented. 

Ready for a first pass of review, however I suspect that my I've generated the protos with the wrong protoc version. `serverchanges.md` does not mention a specific version?

Fixes #693

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3118,7 +3118,7 @@ func (w *Wallet) SendOutputs(outputs []*wire.TxOut, account uint32,
 		return nil, err
 	}
 
-	txHash, err := w.reliablyPublishTransaction(createdTx.Tx)
+	txHash, err := w.reliablyPublishTransaction(createdTx.Tx, "")
 	if err != nil {
 		return nil, err
 	}
@@ -3311,8 +3311,8 @@ func (e *ErrReplacement) Unwrap() error {
 //
 // This function is unstable and will be removed once syncing code is moved out
 // of the wallet.
-func (w *Wallet) PublishTransaction(tx *wire.MsgTx) error {
-	_, err := w.reliablyPublishTransaction(tx)
+func (w *Wallet) PublishTransaction(tx *wire.MsgTx, label string) error {
+	_, err := w.reliablyPublishTransaction(tx, label)
 	return err
 }
 
@@ -3321,7 +3321,9 @@ func (w *Wallet) PublishTransaction(tx *wire.MsgTx) error {
 // relevant database state, and finally possible removing the transaction from
 // the database (along with cleaning up all inputs used, and outputs created) if
 // the transaction is rejected by the backend.
-func (w *Wallet) reliablyPublishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
+func (w *Wallet) reliablyPublishTransaction(tx *wire.MsgTx,
+	label string) (*chainhash.Hash, error) {
+
 	chainClient, err := w.requireChainClient()
 	if err != nil {
 		return nil, err
@@ -3336,7 +3338,19 @@ func (w *Wallet) reliablyPublishTransaction(tx *wire.MsgTx) (*chainhash.Hash, er
 		return nil, err
 	}
 	err = walletdb.Update(w.db, func(dbTx walletdb.ReadWriteTx) error {
-		return w.addRelevantTx(dbTx, txRec, nil)
+		if err := w.addRelevantTx(dbTx, txRec, nil); err != nil {
+			return err
+		}
+
+		// If the tx label is empty, we can return early.
+		if len(label) == 0 {
+			return nil
+		}
+
+		// If there is a label we should write, get the namespace key
+		// and record it in the tx store.
+		txmgrNs := dbTx.ReadWriteBucket(wtxmgrNamespaceKey)
+		return w.TxStore.PutTxLabel(txmgrNs, tx.TxHash(), label)
 	})
 	if err != nil {
 		return nil, err
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -65,6 +65,16 @@ var (
 	// down.
 	ErrWalletShuttingDown = errors.New("wallet shutting down")
 
+	// ErrUnknownTransaction is returned when an attempt is made to label
+	// a transaction that is not known to the wallet.
+	ErrUnknownTransaction = errors.New("cannot label transaction not " +
+		"known to wallet")
+
+	// ErrTxLabelExists is returned when a transaction already has a label
+	// and an attempt has been made to label it without setting overwrite
+	// to true.
+	ErrTxLabelExists = errors.New("transaction already labelled")
+
 	// Namespace bucket keys.
 	waddrmgrNamespaceKey = []byte("waddrmgr")
 	wtxmgrNamespaceKey   = []byte("wtxmgr")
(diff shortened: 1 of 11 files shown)


Roasbeef:
cc @wpaulino

wtxmgr/query.go
```diff
@@ -296,52 +296,13 @@ func (s *Store) rangeBlockTransactions(ns walletdb.ReadBucket, begin, end int32,
...
 			}
-			detail := TxDetails{
-				Block: BlockMeta{
-					Block: block.Block,
-					Time:  block.Time,
-				},
-			}
-			err := readRawTxRecord(&txHash, v, &detail.TxRecord)
+
+			detail, err := s.minedTxDetails(ns, &txHash, k, v)
```
wpaulino, line 342: 👍

wallet/wallet.go
```diff
@@ -3336,7 +3338,19 @@ func (w *Wallet) reliablyPublishTransaction(tx *wire.MsgTx) (*chainhash.Hash, er
...
+
+		// If the tx label is empty, we can return early.
+		if len(label) == 0 {
+			return nil
+		}
+
+		// If there is a label we should write, get the namespace key
+		// and record it in the tx store.
+		txmgrNs := dbTx.ReadWriteBucket(wtxmgrNamespaceKey)
+		return w.TxStore.PutTxLabel(txmgrNs, tx.TxHash(), label)
```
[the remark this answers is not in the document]
carlaKC: Originally put this in `reliablyPublishTransaction` because it's only called for user initiated sends (where it's possible to provide a label). `addRelevantTx` is also called when we get chain notifications, so we would need to check whether there is an existing label, and only overwrite it if the label provided is not empty. Otherwise can end up with user publishing tx with label, tx confirms on chain, label overwritten when `addRelevantTx` is called called by `handleChainNotifications` with label="". 

nbd to move it down and add a check, but labels feel more high level to me, so it made sense to only write on specific use case of a user send.

carlaKC:
>We generally don't make changes to the RPC interface as we don't interact with it, so I'd be fine with leaving those out for now. 

Removed 👍 

> Is there a PR on the lnd side yet to see if these changes will be sufficient for what's needed?

POC PR up, the api change + adding a parameter to `SendOutputs` (which would be required for legacy rpc anyway).

> This may be a valid concern to address if users are relying on these labels.

What do you think of just copying the labels up in the `dropwtxmgr` tool? This could be done optionally, with a `--drop-labels` flag that can be set if you do want to clear them out. Also thinking about an independent `LabelTransaction` endpoint that could be used to label a tx (or relabel with a `--overwrite` option), which could be used to selectively relabel after dropping.

wtxmgr/tx.go
```diff
@@ -935,3 +959,76 @@ func (s *Store) Balance(ns walletdb.ReadBucket, minConf int32, syncHeight int32)
...
+	return putTxLabel(ns, txid, label)
+}
+
+// putTxLabel writes a transaction label to disk, keyed by transaction hash:
+// [0:32] Transaction hash (32 bytes)
+//
+// The label itself is written to disk in length value format:
+// [0:2] Label length
+// [2: +len] Label
+func putTxLabel(ns walletdb.ReadWriteBucket, txid chainhash.Hash,
```
wpaulino, line 977: Nit: doesn't seem like we need an exported and unexported version of this method? Just having the exported would be fine.

rpc/api.proto
```diff
@@ -237,6 +237,10 @@ message SignTransactionResponse {
 
 message PublishTransactionRequest {
 	bytes signed_transaction = 1;
+
+	// An optional string label to store with the transaction, limited to 500
+	// characters.
+	string label = 2;
```
wpaulino, line 246: The RPC changes are still present in the diff.

carlaKC:
> We could go ahead and expose the method required for the endpoint now to avoid another update dependency update in lnd.

Method added. One open question here is whether we want to add a check that this is actually a tx relevant to the wallet? There isn't much harm in allowing people to label transactions that aren't theirs (perhaps even a desirable feature for some), but as is they won't be able to access them unless we add a dedicated lookup function which just gets the label.

wallet/wallet.go
```diff
@@ -1591,6 +1596,45 @@ func (w *Wallet) PubKeyForAddress(a btcutil.Address) (*btcec.PublicKey, error) {
...
+	err := walletdb.View(w.db, func(tx walletdb.ReadTx) error {
+		txmgrNs := tx.ReadBucket(wtxmgrNamespaceKey)
+
+		_, err := wtxmgr.FetchTxLabel(txmgrNs, hash)
+		return err
+	})
+
+	switch err {
+	// If no labels have been written yet, we can silence the error.
+	// Likewise if there is no error, we do not need to do any overwrite
```
wpaulino, line 1615: Did you mean "no label" here instead of "no error"?

wallet/wallet.go
```diff
@@ -1591,6 +1596,45 @@ func (w *Wallet) PubKeyForAddress(a btcutil.Address) (*btcec.PublicKey, error) {
...
+// call will fail if the label is too long, or if the transaction already has
+// a label and the overwrite boolean is not set.
+func (w *Wallet) LabelTransaction(hash chainhash.Hash, label string,
+	overwrite bool) error {
+
+	// Check whether we already have a label for this transaction.
+	err := walletdb.View(w.db, func(tx walletdb.ReadTx) error {
+		txmgrNs := tx.ReadBucket(wtxmgrNamespaceKey)
+
+		_, err := wtxmgr.FetchTxLabel(txmgrNs, hash)
```
wpaulino, line 1609: We should check the transaction exists before this.

wallet/wallet.go
```diff
@@ -65,6 +65,16 @@ var (
 	// down.
 	ErrWalletShuttingDown = errors.New("wallet shutting down")
 
+	// errUnknownTransaction is returned when an attempt is made to label
+	// a transaction that is not known to the wallet.
+	errUnknownTransaction = errors.New("cannot label transaction not " +
```
Roasbeef, line 70: Similar comment here re returning non-exported errors, which means that users need to use string matching rather than comparing against the variable directly.

wtxmgr/tx.go
```diff
@@ -935,3 +959,70 @@ func (s *Store) Balance(ns walletdb.ReadBucket, minConf int32, syncHeight int32)
...
+		return err
+	}
+
+	// We expect the label length to be limited on creation, so we can
+	// store the label's length as a uint16.
+	labelLen := uint16(len(label))
+
+	var buf bytes.Buffer
+
+	var b [2]byte
```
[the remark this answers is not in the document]
carlaKC: Mostly future proofing against needing a migration to add information to the bucket.

carlaKC:
> inconsistency w.r.t exporting or not exporting the newly added errors.

Just an oversight, updated to export all new errors.

PR #697, opened 2020-05-04
walletdb/bdb: attempt to set initial memory map size when opening 

Roasbeef:
In this commit, we start to set the initial memory map size when opening
the database. This helps with bulk insertions or migrations of the
database, as if we set this to a larger value, then bbolt needs to remap
less often. Remapping can be memory intensive as it needs to copy over
the entire existing database.

For 32-bit systems, we take care to clamp this value to ensure we don't
exceed the largest addressable memory on such systems.

In this commit, also we modify the Create and Open methods for bbolt to
accept the bbolt config struct directly. With this change, we don't need
to update this file each time we want to expose a new config option for
the caller to set.


walletdb/bdb/db.go
```diff
@@ -375,6 +387,30 @@ func openDB(dbPath string, create bool, options *bbolt.Options) (walletdb.DB, er
...
+
+		// We'll aim to set the initial memory map to 2x the size of
+		// the database as it exists on disk, meaning the DB size can
+		// double without us having to remap everything.
+		mmapSize := dbFileInfo.Size() * 2
+
+		// If we're on a 32-bit system, then we'll need to limit this
+		// value to ensure we don't run into errors down the line.
+		if !is64Bit {
+			mmapSize = max32BitMapSize
```
Crypt-iQ, line 410: If we always set this for 32-bit systems, this will fail bc the process may be using other virtual memory. Also the mmap is a contiguous block of vmem, so even if this value is lower, this problem can still occur. Part of the reason why I think 32-bit support should be dropped

walletdb/bdb/db.go
```diff
@@ -375,6 +387,30 @@ func openDB(dbPath string, create bool, options *bbolt.Options) (walletdb.DB, er
 	// Specify bbolt freelist options to reduce heap pressure in case the
 	// freelist grows to be very large.
 	options.FreelistType = bbolt.FreelistMapType
+
+	if !create {
+		// The other value that we want to set is the initial memory
+		// map size.  When bolt expands the mmap, it actually copies
+		// over the entire database. By setting this to a large value
```
Crypt-iQ, line 394: *value larger than zero*

PR #700, opened 2020-05-15
wallet: introduce persistent output leases

wpaulino:
This PR allows for the ability to lease an output to a particular ID for a limited amount of time, ensuring that no other processes can use said output for their coin selection needs. An output can either be unlocked manually, or lazily whenever required.


wtxmgr/tx.go
```diff
@@ -1026,3 +1111,70 @@ func FetchTxLabel(ns walletdb.ReadBucket, txid chainhash.Hash) (string, error) {
...
+// invocations of this call.
+//
+// Outputs can be unlocked before their expiration through `UnlockOutput`.
+// Otherwise, they are unlocked lazily through calls which iterate through all
+// known outputs, e.g., `Balance`, `UnspentOutputs`.
+//
+// If the output is not known, ErrUnknownOutput is returned. If the output has
+// already been locked to a different ID, then ErrOutputAlreadyLocked is
+// returned.
+func (s *Store) LockOutput(ns walletdb.ReadWriteBucket, id uint64,
```
Roasbeef, line 1140: Similar Q here re ID.

wtxmgr/tx_test.go
```diff
@@ -2355,3 +2355,338 @@ func TestTxLabel(t *testing.T) {
...
+				})
+
+				// Now lock the confirmed utxo, we should no
+				// longer see any utxos available.
+				lock(t, s, ns, 1, confirmedOutPoint, nil)
+				assertUtxos(t, s, ns, nil)
+
+				// Wait for the output locks to expire for the
+				// utxos to become available once again.
+				<-time.After(lockDuration)
```
Roasbeef, line 2630: If time is mocked out, these could just be insta sends, which'll speed up the tests.

wallet/wallet.go
```diff
@@ -1481,9 +1481,9 @@ type Balances struct {
 // outputs must be iterated.
 func (w *Wallet) CalculateAccountBalances(account uint32, confirms int32) (Balances, error) {
 	var bals Balances
-	err := walletdb.View(w.db, func(tx walletdb.ReadTx) error {
+	err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
```
Roasbeef, line 1484: Kinda weird now that something like just reading balances now can mutate the database and requires a write transaction.

wallet/wallet.go
```diff
@@ -2852,6 +2852,42 @@ func (w *Wallet) LockedOutpoints() []btcjson.TransactionInput {
...
+// Otherwise, they are unlocked lazily through calls which iterate through all
+// known outputs, e.g., `CalculateBalance`, `ListUnspent`.
+//
+// If the output is not known, ErrUnknownOutput is returned. If the output has
+// already been locked to a different ID, then ErrOutputAlreadyLocked is
+// returned.
+//
+// NOTE: This differs from LockOutpoint in that outputs are locked for a limited
+// amount of time and their locks are persisted to disk.
+func (w *Wallet) LeaseOutput(id uint64, op wire.OutPoint) (time.Time, error) {
```
Roasbeef, line 2870: These should also call `LockedOutpoint` so things are in-sync. `LockedOutputs` should also refer to this.

Roasbeef, line 2870: `ListUnspent` would then be able to continue to read the in-memory map and apply output locked-ness as normal.

wtxmgr/db.go
```diff
@@ -1287,6 +1288,89 @@ func deleteRawUnminedInput(ns walletdb.ReadWriteBucket, outPointKey []byte,
...
+		_ = lockedCredits.Delete(k)
+		return 0, time.Time{}, false
+	}
+
+	return lockID, expiration, true
+}
+
+// lockOutput creates a lock for `duration` over an output assigned to the `id`,
+// preventing it from becoming eligible for coin selection.
+func lockOutput(ns walletdb.ReadWriteBucket, id uint64,
```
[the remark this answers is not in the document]
wpaulino: No mapping, just went with uint64 as that seemed enough.

wtxmgr/db.go
```diff
@@ -1287,6 +1288,89 @@ func deleteRawUnminedInput(ns walletdb.ReadWriteBucket, outPointKey []byte,
...
+
+	// Create the corresponding bucket if necessary.
+	lockedCredits, err := ns.CreateBucketIfNotExists(bucketLockedOutputs)
+	if err != nil {
+		str := "failed to create locked credits bucket"
+		return time.Time{}, storeError(ErrDatabase, str, err)
+	}
+
+	// Determine the absolute time at which the output lock expires.
+	expiry := time.Now().Add(duration)
```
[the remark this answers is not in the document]
wpaulino: It'd have to be turned into a module first. The logic here was simple enough IMO to test without such an interface though.

wallet/wallet.go
wpaulino: I'm not sure that's safe? The lease lock will expire at some point, but that's not the case with `LockedOutpoint`, unless we detect when it has expired and remove it ourselves.

wtxmgr/db.go
```diff
@@ -1287,6 +1288,89 @@ func deleteRawUnminedInput(ns walletdb.ReadWriteBucket, outPointKey []byte,
 	return nil
 }
 
+// isLockedOutput determines whether an output is locked. If it is, its assigned
+// ID is returned, along with its absolute expiration time. If the output lock
+// exists, but its expiration has been met, then it's removed.
+func isLockedOutput(ns walletdb.ReadWriteBucket, op wire.OutPoint) (uint64,
```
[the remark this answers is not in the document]
wpaulino: Found a way of getting around this in the latest diff. `isLockedOutput` no longer deletes anything, but still evaluates the expiration to determine if it's been reached or not. Deletion now only occurs once on startup.

wallet/wallet.go
wpaulino: No longer an issue.

wallet/wallet.go
Roasbeef: Well, I guess it depends on the goals here. I had envisioned we would have a sort of unified abstraction for al of this, but it's also feasible that we just have "locking" and "leasing" be distinct classes of actions w.r.t the wallet.

wtxmgr/db.go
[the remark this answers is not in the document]
Roasbeef: Gotcha, it is "a lot" of space, but IMO not large enough that applications won't collide easily. Depending on what the interface is like on the `lnd` side, they may just try with an ID of `0`, that fails, then guess at some other ID. With a "hash" like object, we have a much larger space and can also be more sure that all applications can easily pick a unique ID.

wallet/wallet.go
Roasbeef: The main gap here is that as is (still need to glance at the latest diff), things like evaluating the balance will examine _leased_ outputs, but not _locked_ outputs.

wtxmgr/db.go
```diff
@@ -1287,6 +1288,124 @@ func deleteRawUnminedInput(ns walletdb.ReadWriteBucket, outPointKey []byte,
...
+// deserializeLockedOutput deserializes the value of a locked output.
+func deserializeLockedOutput(v []byte) (uint64, time.Time) {
+	lockID := byteOrder.Uint64(v[:8])
+	expiry := time.Unix(int64(byteOrder.Uint64(v[8:])), 0)
+	return lockID, expiry
+}
+
+// isLockedOutput determines whether an output is locked. If it is, its assigned
+// ID is returned, along with its absolute expiration time. If the output lock
+// exists, but its expiration has been met, then it's removed.
```
Roasbeef, line 1308: Comment here not super accurate in that it's the caller that will eventually remove them.

wtxmgr/db.go
```diff
@@ -1287,6 +1288,124 @@ func deleteRawUnminedInput(ns walletdb.ReadWriteBucket, outPointKey []byte,
...
+	// Retrieve the output lock, if any, and extract the relevant fields.
+	k := canonicalOutPoint(&op.Hash, op.Index)
+	v := lockedOutputs.Get(k)
+	if v == nil {
+		return 0, time.Time{}, false
+	}
+	lockID, expiration := deserializeLockedOutput(v)
+
+	// If the output lock has already expired, delete it now.
+	if !time.Now().Before(expiration) {
```
Roasbeef, line 1328: Use `After` here for clarity?

wpaulino: `After` doesn't handle `time.Now == expiration`.

wtxmgr/db.go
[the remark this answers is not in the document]
wpaulino: It was turned into a module, so this is done now.

wtxmgr/db.go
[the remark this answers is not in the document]
wpaulino: Made it 32 bytes.

PR #702, opened 2020-05-28
dropwtxmgr: add drop labels option

carlaKC:
Add an option to drop your tx labels in `dropwtxmgr`. By default, we keep labels to prevent accidental deletion of important tx information that cannot be obtained from chain. This change pulls some of the label read/write logic out into separate logic in `wtxmgr` so we do not have to duplicate it. 

Tested with some labelled txns in lnd - drop wtxmgr, labels remain; drop with `--droplables` and labels are gone.

### DIFF
--- a/cmd/dropwtxmgr/main.go
+++ b/cmd/dropwtxmgr/main.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"path/filepath"
 
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcutil"
 	"github.com/btcsuite/btcwallet/waddrmgr"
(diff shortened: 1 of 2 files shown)


carlaKC:
PTAL @wpaulino 🙏

cmd/dropwtxmgr/main.go
```diff
@@ -109,6 +114,16 @@ func mainInt() int {
 	fmt.Println("Dropping btcwallet transaction history")
```
wpaulino, line 114: Could you add `with/without labels` to this print?

wpaulino:
cc @Roasbeef

Roasbeef:
Bobby Droptables ;)

PR #711, opened 2020-06-19
wallet: add new PSBT funding and finalizing methods

guggero:
Depends on https://github.com/btcsuite/btcutil/pull/178

Adds the methods `FundPsbt` which can perform coin selection and/or fee estimation to fund outputs in a PSBT and `FinalizePsbt` which signs known inputs of a PSBT packet.

We also add `FetchInputInfo` and `ComputeInputScript` which previously resided in `lnd` to make them available to the PSBT signing code and other users of `btcwallet`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ require (
 	github.com/btcsuite/btcd v0.20.1-beta.0.20200513120220-b470eee47728
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
+	github.com/btcsuite/btcutil/psbt v1.0.3-0.20200826194809-5f93e33af2b0
 	github.com/btcsuite/btcwallet/wallet/txauthor v1.0.0
 	github.com/btcsuite/btcwallet/wallet/txrules v1.0.0
 	github.com/btcsuite/btcwallet/walletdb v1.3.3
--- a/wallet/createtx_test.go
+++ b/wallet/createtx_test.go
@@ -18,6 +18,14 @@ import (
 	"github.com/btcsuite/btcwallet/wtxmgr"
 )
 
+var (
+	testBlockHash, _ = chainhash.NewHashFromStr(
+		"00000000000000017188b968a371bab95aa43522665353b646e41865abae" +
+			"02a4",
+	)
+	testBlockHeight int32 = 276425
+)
+
 // TestTxToOutput checks that no new address is added to he database if we
 // request a dry run of the txToOutputs call. It also makes sure a subsequent
 // non-dry run call produces a similar transaction to the dry-run.
@@ -45,41 +53,7 @@ func TestTxToOutputsDryRun(t *testing.T) {
 			txOut,
 		},
 	}
-
-	var b bytes.Buffer
-	if err := incomingTx.Serialize(&b); err != nil {
-		t.Fatalf("unable to serialize tx: %v", err)
-	}
-	txBytes := b.Bytes()
-
-	rec, err := wtxmgr.NewTxRecord(txBytes, time.Now())
-	if err != nil {
-		t.Fatalf("unable to create tx record: %v", err)
-	}
-
-	// The block meta will be inserted to tell the wallet this is a
-	// confirmed transaction.
-	blockHash, _ := chainhash.NewHashFromStr(
-		"00000000000000017188b968a371bab95aa43522665353b646e41865abae02a4")
-	block := &wtxmgr.BlockMeta{
-		Block: wtxmgr.Block{Hash: *blockHash, Height: 276425},
-		Time:  time.Unix(1387737310, 0),
-	}
-
-	if err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
-		ns := tx.ReadWriteBucket(wtxmgrNamespaceKey)
-		err = w.TxStore.InsertTx(ns, rec, block)
-		if err != nil {
-			return err
-		}
-		err = w.TxStore.AddCredit(ns, rec, block, 0, false)
-		if err != nil {
-			return err
-		}
-		return nil
-	}); err != nil {
-		t.Fatalf("failed inserting tx: %v", err)
-	}
+	addUtxo(t, w, incomingTx)
 
 	// Now tell the wallet to create a transaction paying to the specified
 	// outputs.
@@ -173,5 +147,45 @@ func TestTxToOutputsDryRun(t *testing.T) {
 	if !bytes.Equal(change2.PkScript, change3.PkScript) {
 		t.Fatalf("dry-run using different change address " +
 			"than wet run")
+	}
+}
+
+// addUtxo add the given transaction to the wallet's database marked as a
+// confirmed UTXO .
+func addUtxo(t *testing.T, w *Wallet, incomingTx *wire.MsgTx) {
+	var b bytes.Buffer
+	if err := incomingTx.Serialize(&b); err != nil {
+		t.Fatalf("unable to serialize tx: %v", err)
+	}
+	txBytes := b.Bytes()
+
+	rec, err := wtxmgr.NewTxRecord(txBytes, time.Now())
+	if err != nil {
+		t.Fatalf("unable to create tx record: %v", err)
+	}
+
+	// The block meta will be inserted to tell the wallet this is a
+	// confirmed transaction.
+	block := &wtxmgr.BlockMeta{
+		Block: wtxmgr.Block{
+			Hash:   *testBlockHash,
+			Height: testBlockHeight,
+		},
+		Time: time.Unix(1387737310, 0),
+	}
+
+	if err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+		ns := tx.ReadWriteBucket(wtxmgrNamespaceKey)
+		err = w.TxStore.InsertTx(ns, rec, block)
+		if err != nil {
+			return err
+		}
+		err = w.TxStore.AddCredit(ns, rec, block, 0, false)
+		if err != nil {
+			return err
+		}
+		return nil
+	}); err != nil {
+		t.Fatalf("failed inserting tx: %v", err)
 	}
 }
--- a/wallet/psbt.go
+++ b/wallet/psbt.go
@@ -1,0 +1,305 @@
+// Copyright (c) 2020 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package wallet
+
+import (
+	"bytes"
+	"fmt"
+
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcutil/psbt"
+	"github.com/btcsuite/btcwallet/wallet/txauthor"
+	"github.com/btcsuite/btcwallet/wallet/txrules"
+	"github.com/btcsuite/btcwallet/wtxmgr"
+)
+
+// FundPsbt creates a fully populated PSBT packet that contains enough inputs to
+// fund the outputs specified in the passed in packet with the specified fee
+// rate. If there is change left, a change output from the wallet is added.
+//
+// NOTE: If the packet doesn't contain any inputs, coin selection is performed
+// automatically. If the packet does contain any inputs, it is assumed that full
+// coin selection happened externally and no additional inputs are added. If the
+// specified inputs aren't enough to fund the outputs with the given fee rate,
+// an error is returned.
+//
+// NOTE: A caller of the method should hold the global coin selection lock of
+// the wallet. However, no UTXO specific lock lease is acquired for any of the
+// selected/validated inputs by this method. It is in the caller's
+// responsibility to lock the inputs before handing the partial transaction out.
+func (w *Wallet) FundPsbt(packet *psbt.Packet, account uint32,
+	feeSatPerKB btcutil.Amount) error {
+
+	// Make sure the packet is well formed. We only require there to be at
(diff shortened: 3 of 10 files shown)


go.mod
```diff
@@ -4,6 +4,7 @@ require (
 	github.com/btcsuite/btcd v0.20.1-beta.0.20200513120220-b470eee47728
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
 	github.com/btcsuite/btcutil v0.0.0-20190425235716-9e5f4b9a998d
+	github.com/btcsuite/btcutil/psbt v1.0.2
```
Roasbeef, line 7: I think we'll need to tag a new version on the `btcutil` side correct?

wallet/createtx_test.go
```diff
@@ -26,7 +25,7 @@ func TestTxToOutputsDryRun(t *testing.T) {
 	defer cleanup()
 
 	// Create an address we can use to send some coins to.
-	addr, err := w.CurrentAddress(0, waddrmgr.KeyScopeBIP0044)
+	addr, err := w.CurrentAddress(0, waddrmgr.KeyScopeBIP0084)
```
Roasbeef, line 28: Why the change in key-scope? To force a segwit addr?

wallet/signer.go
```diff
@@ -0,0 +1,102 @@
...
+
+import (
+	"github.com/btcsuite/btcd/btcec"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+)
+
+// Tweaker is a function type that can be used to pass in a callback for
```
Roasbeef, line 15: Prefix with `PrivKey`? ;) 

So `PrivKeyTweaker`, otherwise there're some undesirable connotations w.r.t the colloquial usage of this term for native english speakers lol.

wallet/signer_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+	if len(witness) != 2 {
+		t.Fatalf("unexpected witness stack length, got %d, wanted %d",
+			len(witness), 2)
+	}
+
+	// Finally verify that the created witness is valid.
+	outgoingTx.TxIn[0].Witness = witness
+	outgoingTx.TxIn[0].SignatureScript = script
+	err = validateMsgTx(
```
Roasbeef, line 97: 👍

wallet/psbt.go
```diff
@@ -0,0 +1,298 @@
...
+
+// FundPsbt creates a fully populated PSBT packet that contains enough inputs to
+// fund the outputs specified in the passed in packet with the specified fee
+// rate. If there is change left, a change output from the wallet is added.
+//
+// NOTE: If the packet doesn't contain any inputs, coin selection is performed
+// automatically. If the packet does contain any inputs, it is assumed that full
+// coin selection happened externally and no additional inputs are added. If the
+// specified inputs aren't enough to fund the outputs with the given fee rate,
+// an error is returned. No lock lease is acquired for any of the
```
Roasbeef, line 29: So the caller should be holding some sort of coin selection mutex before it calls this function? I guess things _are_ serialized through `CreateSimpleTx` partially, however I don't think that does any internal coin selection locking itself.

wallet/psbt.go
```diff
@@ -0,0 +1,298 @@
...
+	"github.com/btcsuite/btcwallet/wallet/txrules"
+	"github.com/btcsuite/btcwallet/wtxmgr"
+)
+
+// FundPsbt creates a fully populated PSBT packet that contains enough inputs to
+// fund the outputs specified in the passed in packet with the specified fee
+// rate. If there is change left, a change output from the wallet is added.
+//
+// NOTE: If the packet doesn't contain any inputs, coin selection is performed
+// automatically. If the packet does contain any inputs, it is assumed that full
```
Roasbeef, line 26: Hmm, I guess this behavior (no partial coin selection) isn't _too_ restrictive. Also seems hard to know hoc much we shoul dtry to pay for w/ partial coin selection, and we also won't know what the full fee would be ahead of time.

wallet/psbt.go
```diff
@@ -0,0 +1,298 @@
...
+	}
+
+	// Let's find out the amount to fund first.
+	amt := int64(0)
+	for _, output := range txOut {
+		amt += output.Value
+	}
+
+	// addInputInfo is a helper function that fetches the UTXO information
+	// of an input and attaches it to the PSBT packet.
```
Roasbeef, line 65: So we assume all witness utxos? We _do_ currently support nested p2wkh which _technically_ is a non-witness utxo? 

Actually, I guess it isn't since there's no signature in the P2SH portion, and the witness signature will properly sign the input value.

wallet/psbt.go
```diff
@@ -0,0 +1,298 @@
...
+			return fmt.Errorf("error creating funding TX: %v", err)
+		}
+
+		// Copy over the inputs now then collect all UTXO information
+		// that we can and attach them to the PSBT as well. We don't
+		// include the witness as the resulting PSBT isn't expected not
+		// should be signed yet.
+		txIn = tx.Tx.TxIn
+		packet.UnsignedTx.TxIn = tx.Tx.TxIn
+		err = addInputInfo()
```
Roasbeef, line 109: How about just passing the txin as an argument to this func? This way we don't need to mess around with swapping out the variables the closure closes over.

wallet/psbt.go
```diff
@@ -0,0 +1,298 @@
...
+		if err != nil {
+			return err
+		}
+
+	// If there are inputs, we need to check if they're sufficient and add
+	// a change output if necessary.
+	default:
```
Roasbeef, line 116: What's the intend scenario where this method would be used? Just as a sanity check? If the transaction already has inputs, then perhaps it's safe a safe assumption that it also has computed fees and has a change address?

wallet/psbt.go
```diff
@@ -0,0 +1,298 @@
...
+		}
+
+		// Serialize the witness format from the stack representation to
+		// the wire representation.
+		var witnessBytes bytes.Buffer
+		err = psbt.WriteTxWitness(&witnessBytes, witness)
+		if err != nil {
+			return fmt.Errorf("error serializing witness: %v", err)
+		}
+		packet.Inputs[idx].FinalScriptWitness = witnessBytes.Bytes()
```
wpaulino, line 286: Somewhat unrelated, but it's unfortunate some of these fields PSBT fields are undocumented...

wallet/createtx_test.go
guggero: No, I don't think I actually wanted to commit this change. Fixed, thanks!

wallet/signer.go
guggero: Oh, I wasn't aware of that particular colloquial usage :sweat_smile: Had to google it and now I'm kind of disappointed that "Breaking Bad" didn't teach me that term :joy:

wallet/psbt.go
guggero: True, had to look up most of these fields too. The library kind of assumes you are familiar with the BIP. We should fix that in the future.

wallet/psbt.go
```diff
@@ -0,0 +1,298 @@
...
+	// ones to sign. If there is any input without witness data that we
+	// cannot sign because it's not our UTXO, this will be a hard failure.
+	tx := packet.UnsignedTx
+	sigHashes := txscript.NewTxSigHashes(tx)
+	for idx, txIn := range tx.TxIn {
+		in := packet.Inputs[idx]
+
+		// We can only sign witness inputs. Everything else must already
+		// have been signed or the finalize step later will fail.
+		if in.WitnessUtxo == nil {
```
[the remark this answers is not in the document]
guggero: Yes, from the perspective of a signer, a NP2WKH looks exactly like a native SegWit output except that `FinalScriptSig` is set to a non-empty slice.

wallet/psbt.go
guggero: Nested P2WKH are witness outputs. The only difference on the signer side is that they have the `FinalScriptSig` set to a non-empty slice.

We will have to change this in a future bugfix to fix the PSBT security flaw: https://github.com/spesmilo/electrum/issues/6257#issuecomment-647736461 (also see https://github.com/lightningnetwork/lnd/issues/4400).

I'd rather not do it in this PR because a change in the `btcutil/psbt` library is also necessary to remove some of the sanity checks.

wallet/psbt.go
guggero: Not necessarily. My goal was to replicate `bitcoind`'s `walletcreatefundedpsbt` where you can either leave the UTXO list empty or select one or more UTXOs. So to give the user a basic level of "coin control" without them also being required to calculate the fees and select a change address. This is especially handy now that we have labels on the transactions, selecting which UTXO to spend should be a feature of all on-chain calls of the wallet IMO.

wallet/psbt.go
guggero: I added another note to the godoc comment to mention that the global coin selection lock must be held.

wallet/psbt.go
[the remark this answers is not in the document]
wpaulino: Not sure if I'm missing something, but I don't see that being done here.

wallet/psbt_test.go
```diff
@@ -0,0 +1,302 @@
...
+func TestFinalizePsbt(t *testing.T) {
+	w, cleanup := testWallet(t)
+	defer cleanup()
+
+	// Create an address we can use to send some coins to.
+	addr, err := w.CurrentAddress(0, waddrmgr.KeyScopeBIP0084)
+	if err != nil {
+		t.Fatalf("unable to get current address: %v", addr)
+	}
+	p2shAddr, err := txscript.PayToAddrScript(addr)
```
wpaulino, line 245: Should be P2WKH?  Could you add a nested P2WKH test case as well? This also applies to `FundPsbt`.

wallet/psbt.go
```diff
@@ -0,0 +1,300 @@
...
+		inputSource := makeInputSource(credits)
+
+		// We also need a change source which needs to be able to insert
+		// a new change addresse into the database.
+		dbtx, err := w.db.BeginReadWriteTx()
+		if err != nil {
+			return err
+		}
+		addrmgrNs := dbtx.ReadWriteBucket(waddrmgrNamespaceKey)
+		changeSource := func() ([]byte, error) {
```
wpaulino, line 145: I believe this logic also exists elsewhere, can we refactor it into a method and use it here?

wallet/signer.go
Roasbeef: 😂😂

wallet/psbt.go
Roasbeef: Looking at `psbt/finalizer.go` a bit more, it looks like it expects you to only set the `RedeemScript`, then it'll construct the `sigScript` itself manually: https://github.com/btcsuite/btcutil/blob/master/psbt/finalizer.go#L417, then emplaced here: https://github.com/btcsuite/btcutil/blob/master/psbt/finalizer.go#L453

wallet/psbt.go
```diff
@@ -0,0 +1,300 @@
...
+
+		// Serialize the witness format from the stack representation to
+		// the wire representation.
+		var witnessBytes bytes.Buffer
+		err = psbt.WriteTxWitness(&witnessBytes, witness)
+		if err != nil {
+			return fmt.Errorf("error serializing witness: %v", err)
+		}
+		packet.Inputs[idx].FinalScriptWitness = witnessBytes.Bytes()
+		packet.Inputs[idx].WitnessScript = sigScript
```
Roasbeef, line 289: This should be the witness script instead. It looks like things get by atm since this isn't a multi-sig output, and it only checks these fields for nil-ness when it goes to finalize it.

wallet/psbt_test.go
guggero: Ah, you're right. I assigned the script to `WitnessScript` instead of `FinalScriptSig`. The added test case caught that now. Thanks!

wallet/psbt.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		// Serialize the witness format from the stack representation to
+		// the wire representation.
+		var witnessBytes bytes.Buffer
+		err = psbt.WriteTxWitness(&witnessBytes, witness)
+		if err != nil {
+			return fmt.Errorf("error serializing witness: %v", err)
+		}
+		packet.Inputs[idx].FinalScriptWitness = witnessBytes.Bytes()
```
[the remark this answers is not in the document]
guggero: Yes indeed, thanks!

wallet/psbt.go
guggero: Added a test case that uses a nested P2WKH input and that caught the problem. Should work correctly now.

wallet/psbt.go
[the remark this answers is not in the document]
guggero: Ah, you're right. I assigned the script to `WitnessScript` instead of `FinalScriptSig`. The added test case caught that now. Thanks!

sako0938:
Can you show me how to us this PR with my current LND, so I can test PSBT funding with latest electrum releases?

guggero:
See https://github.com/lightningnetwork/lnd/pull/4389, that adds PSBT signing to `lnd`. Though you won't be able to use PSBTs created with Electrum, because they already added the bugfix for CVE-2020-14199 (see https://github.com/spesmilo/electrum/issues/6257#issuecomment-647736461). I'm working on a fix for that in `btcutil/psbt` as well.

guggero:
I updated the PR to include the compatibility fix for patched HW wallets: https://github.com/btcsuite/btcutil/pull/178

I also made sure we always include the non-witness UTXO in our PSBTs as well to make sure we produce packets that can be signed by HW wallets with the newest firmware.

guggero:
Dependency updated.

sako0938:
Can’t wait to test this out in LND!

PR #716, opened 2020-08-13
wallet/txauthor: fix bug in dust calculation

Roasbeef:
In this commit, we fix a 3 year old bug in dust calculation. Before this
commit, the target fee of the transaction to be crafted would be used to
determine dust. If the target fee is very high, then this would cause
over all higher fees, as we'd base that dust computation off of that fee
rate, rather than the min relay fee.

To fix this, we always use the min relay fee at all times when computing
dust.


Roasbeef:
cc @wpaulino

PR #718, opened 2020-08-25
wallet: only request spend notification for our outputs

wpaulino:
We would previously request spend notifications for all transaction outputs, leading to irrelevant transactions being found in the wallet's transaction store.

Motivated by https://github.com/lightningnetwork/lnd/issues/3746.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3447,7 +3447,35 @@ func (w *Wallet) reliablyPublishTransaction(tx *wire.MsgTx,
 	if err != nil {
 		return nil, err
 	}
+
+	// Along the way, we'll extract our relevant destination addresses from
+	// the transaction.
+	var ourAddrs []btcutil.Address
 	err = walletdb.Update(w.db, func(dbTx walletdb.ReadWriteTx) error {
+		txmgrNs := dbTx.ReadWriteBucket(wtxmgrNamespaceKey)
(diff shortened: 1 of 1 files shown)


wpaulino:
cc @Roasbeef

Roasbeef (approved):
LGTM 🔋

Roasbeef:
cc @halseth

wallet/wallet.go
```diff
@@ -3469,27 +3496,8 @@ func (w *Wallet) reliablyPublishTransaction(tx *wire.MsgTx,
...
-		if err != nil {
-			// Non-standard outputs can safely be skipped because
-			// they're not supported by the wallet.
-			continue
-		}
-
-		if err := chainClient.NotifyReceived(addrs); err != nil {
-			return nil, err
-		}
+	if err := chainClient.NotifyReceived(ourAddrs); err != nil {
```
halseth, line 3499: Will this still be okay in the case where we send a 1 output tx to a destination that is not ours? We don't rely on confirmation being anywhere for that case?

halseth (commented):
Only a question, otherwise LGTM

wallet/wallet.go
wpaulino: Good question, it should be notified as any addresses we're watching that get spent also request a notification on the input spending it for when it confirms.

PR #720, opened 2020-09-26
wallet: allow wallet to be created from extended master root key directly

guggero:
To make it possible to create or restore a wallet from the extended master root key only, we refactor the loader, wallet and address manager to accept the root key directly instead of the seed.

### DIFF
--- a/wallet/loader.go
+++ b/wallet/loader.go
@@ -13,6 +13,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcutil/hdkeychain"
 	"github.com/btcsuite/btcwallet/internal/prompt"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/walletdb"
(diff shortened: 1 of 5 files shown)


guggero:
Rebased. cc @carlaKC, @wpaulino

wallet/wallet.go
```diff
@@ -3730,25 +3731,27 @@ func CreateWatchingOnly(db walletdb.DB, pubPass []byte,
...
-			hdSeed, err := hdkeychain.GenerateSeed(
-				hdkeychain.RecommendedSeedLen)
-			if err != nil {
-				return err
-			}
-			seed = hdSeed
+	// If no root key was provided, we create one now from a random seed.
+	// But only if this is not a watching-only wallet where the accounts are
+	// created individually from their xpubs.
+	if !isWatchingOnly && rootKey == nil {
```
wpaulino, line 3742: We should check the `rootKey` is a private key if one is given.

wpaulino (approved):

wallet/wallet.go
guggero: Yes, good idea. Done.

wallet/loader.go
```diff
@@ -145,8 +146,42 @@ func (l *Loader) OnWalletCreated(fn func(walletdb.ReadWriteTx) error) {
 func (l *Loader) CreateNewWallet(pubPassphrase, privPassphrase, seed []byte,
 	bday time.Time) (*Wallet, error) {
 
+	var (
+		rootKey *hdkeychain.ExtendedKey
+		err     error
+	)
+
+	// If a seed was specified, we check its length now. If no seed is
```
Roasbeef, line 154: 👍

Roasbeef (approved):
LGTM 🧩

PR #721, opened 2020-10-01
wallet: use constant input source for change calculation

guggero:
To fix a bug where specifying multiple UTXOs that are by themselves
large enough to satisfy the output amount would lead to the rest of them
being added to fees, we need to provide the transaction author with a
constant list of UTXOs. If we didn't, the author would only consider one
input and calculate the change based on that alone. But since we'd add
all inputs to the PSBT, the rest of the amounts would go to fees.

### DIFF
--- a/wallet/psbt.go
+++ b/wallet/psbt.go
@@ -19,7 +19,9 @@ import (
 
 // FundPsbt creates a fully populated PSBT packet that contains enough inputs to
 // fund the outputs specified in the passed in packet with the specified fee
-// rate. If there is change left, a change output from the wallet is added.
+// rate. If there is change left, a change output from the wallet is added and
+// the index of the change output is returned. Otherwise no additional output
+// is created and the index -1 is returned.
 //
 // NOTE: If the packet doesn't contain any inputs, coin selection is performed
 // automatically. If the packet does contain any inputs, it is assumed that full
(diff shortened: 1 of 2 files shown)


wallet/psbt.go
```diff
@@ -186,10 +190,22 @@ func (w *Wallet) FundPsbt(packet *psbt.Packet, account uint32,
...
+		return 0, fmt.Errorf("could not sort PSBT: %v", err)
 	}
 
-	return nil
+	// The change output index might have changed after the sorting. We need
+	// to find our index again.
+	changeIndex := int32(-1)
+	if changeTxOut != nil {
+		for idx, txOut := range packet.UnsignedTx.TxOut {
+			if psbt.TxOutsEqual(changeTxOut, txOut) {
```
Roasbeef, line 208: Ah ok, we do a deep comparison here.

wallet/psbt_test.go
```diff
@@ -269,10 +272,12 @@ func TestFundPsbt(t *testing.T) {
 				case bytes.Equal(script, testScriptP2WSH):
 					p2wshIndex = idx
 
-				default:
-					changeIndex = idx
 				}
 			}
+			totalIn := int64(0)
+			for _, txIn := range packet.Inputs {
+				totalIn += txIn.WitnessUtxo.Value
```
Roasbeef, line 285: Does it a tally but no assertion using these new values?

wallet/psbt_test.go
```diff
@@ -296,15 +302,17 @@ func TestFundPsbt(t *testing.T) {
 					txOuts[p2wshIndex].PkScript)
 			}
 
-			// Finally, check the change output size and that it
-			// belongs to the wallet.
-			expectedFee := int64(368)
-			expectedChange := 1000000 - 150000 - expectedFee
-			if txOuts[changeIndex].Value != expectedChange {
+			// Finally, check the change output size and fee.
+			fee := totalIn - totalOut
```
Roasbeef, line 306: Ah ok, here it is!

Roasbeef (approved):
LGTM 🦋

wpaulino (approved):

PR #722, opened 2020-10-05
psbt: don't add scriptSig to txIn

guggero:
Because of an incorrect test, it wasn't discovered that the scriptSig
field was being set on the unsigned TX inputs for a nested SegWit input.
This PR fixes the bug and also refactors the test so it would have
caught this specific bug.

### DIFF
--- a/wallet/psbt_test.go
+++ b/wallet/psbt_test.go
@@ -43,7 +43,7 @@ func TestFundPsbt(t *testing.T) {
 	}
 
 	// Also create a nested P2WKH address we can use to send some coins to.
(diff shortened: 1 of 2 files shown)


wallet/psbt_test.go
```diff
@@ -318,6 +281,46 @@ func TestFundPsbt(t *testing.T) {
 	}
 }
 
+func assertTxInputs(t *testing.T, packet *psbt.Packet,
```
Roasbeef, line 284: 👍

Roasbeef (approved):
LGTM 🦆

PR #723, opened 2020-10-24
Export dropwtxmgr functionality as utility function

guggero:
We create a more generic copy of the `dropwtxmgr` command's functionality
and export it as the `wallet.DropTransactionHistory` utility function.
It removes all transaction history from the given wallet to force a
full chain rescan. Optionally the user-defined transaction labels can be
preserved.


wpaulino (approved):

Roasbeef (approved):
LGTM 📍

PR #731, opened 2021-02-15
wtxmgr: ignore ErrBucketNotFound error upon locked outputs bucket deletion

wpaulino:
This error would be seen when an old wallet that has yet to update is performing the latest wtxmgr migration. It's possible for the locked outputs bucket to not exist if outputs haven't been locked before, so we should its deletion correctly.


guggero (approved):
Nice catch, LGTM :tada:

Roasbeef (approved):
LGTM 🌺

PR #732, opened 2021-02-17
wallet: support derived public key import

wpaulino:
In this PR, we build upon https://github.com/btcsuite/btcwallet/pull/667 to introduce support for importing master/derived public keys into the wallet. This has been a much requested feature by `lnd` users that will allow them to minimize their hot wallet risks. Master public keys, commonly referred to as `xpub`s, must have a derivation path of the form `m/purpose'/coin_type'/account'`, while derived public keys must have a derivation path of the form `m/purpose'/coin_type'/account'/branch/index`. As of now, care must be taken to ensure any watch-only accounts/derived public keys imported have not been used, as the functionality to rescan for their previous activity is not supported yet.


Roasbeef:
cc @guggero

waddrmgr/scoped_manager.go
```diff
@@ -308,69 +308,92 @@ func (s *ScopedKeyManager) loadAccountInfo(ns walletdb.ReadBucket,
...
+	branch, index := ExternalBranch, acctInfo.nextExternalIndex
 	if index > 0 {
 		index--
 	}
 	lastExtAddrPath := DerivationPath{
 		InternalAccount: account,
-		Account:         acctKeyPub.ChildIndex(),
 		Branch:          branch,
 		Index:           index,
+		Account:         acctInfo.acctKeyPub.ChildIndex(),
```
guggero, line 396: nit: preserve field order as this is moved back up again in a later commit?

waddrmgr/scoped_manager.go
```diff
@@ -308,69 +308,92 @@ func (s *ScopedKeyManager) loadAccountInfo(ns walletdb.ReadBucket,
...
-		acctName:          row.name,
-		acctKeyEncrypted:  row.privKeyEncrypted,
-		acctKeyPub:        acctKeyPub,
-		nextExternalIndex: row.nextExternalIndex,
-		nextInternalIndex: row.nextInternalIndex,
-	}
+	var acctInfo *accountInfo
+	watchOnly := s.rootManager.WatchOnly()
+	private := !s.rootManager.isLocked()
+	switch row := rowInterface.(type) {
```
guggero, line 326: So if some wallet actually started using the watch only accounts this still shouldn't break because they'd still be stored as the old, default account type? And new watch-only accounts would be stored under the correct, new type?

wallet/import.go
```diff
@@ -1,13 +1,193 @@
...
+		_, err = scopedMgr.NewAccountWatchingOnly(
+			ns, name, masterPubKey, masterKeyFingerprint,
+		)
+		return err
+	})
+}
+
+// ImportPubKey imports a single derived public key into the address manager. An
+// optional block can be provided to determine where the rescan should begin.
+func (w *Wallet) ImportPubKey(pubKey *hdkeychain.ExtendedKey) error {
```
guggero, line 158: I'm missing the context on how we're planning on using this, so this is just a side note: If we pass in a default BIP44 extended key (`tpub`), this will cause the address to be of type `PubKeyHash`. So either we only use this with BIP84 extended keys (`vpub`) or need an address type override parameter here as well.

guggero (commented):
Did a first pass, looks pretty good!
Since I wasn't very familiar with the code base it took me a while to grok all the changes. It is easy to underestimate the complexity here. So great work on this, the diff reads well once one understands the context.

wallet/import.go
```diff
@@ -1,13 +1,191 @@
...
 
+// pubKeyHDVersion defines the different public key HD versions supported that
+// can be imported into the wallet.
+type pubKeyHDVersion uint32
+
+const (
+	mainnetBIP0044 pubKeyHDVersion = 0x0488b21e // xpub
+	testnetBIP0044 pubKeyHDVersion = 0x043587cf // tpub
+
+	mainnetBIP0049Plus pubKeyHDVersion = 0x049d7cb2 // ypub
```
Roasbeef, line 22: Note that bip49 plus was somethign i actually made up. If someone imports an actual "ypub" that used nested p2sh for its change addresses, then we'd miss the funds on rescan.

wallet/import.go
```diff
@@ -1,13 +1,191 @@
...
+
+// ImportAccount imports an account backed by a master extended public key. The
+// master key fingerprint denotes the fingerprint of the root key corresponding
+// to the master public key (also known as the key with derivation path m/).
+// This may be required by some hardware wallets for proper identification and
+// signing. The address type can usually be inferred from the key's version, but
+// in the case of legacy versions (xpub, tpub), an address type must be
+// specified as we intend to not support importing BIP-44 keys into the wallet
+// using the legacy pay-to-pubkey (P2PK) scheme.
+func (w *Wallet) ImportAccount(name string, masterPubKey *hdkeychain.ExtendedKey,
```
Roasbeef, line 124: Is this _actually_ the master xpub? Given derivation for BIP 44 is hardened all the way down to the actual account itself...

wallet/import.go
```diff
@@ -1,13 +1,191 @@
...
+		return err
+	})
+}
+
+// ImportPubKey imports a single derived public key into the address manager.
+// The address type can usually be inferred from the key's version, but in the
+// case of legacy versions (xpub, tpub), an address type must be specified as we
+// intend to not support importing BIP-44 keys into the wallet using the legacy
+// pay-to-pubkey (P2PK) scheme.
+func (w *Wallet) ImportPubKey(pubKey *hdkeychain.ExtendedKey,
```
Roasbeef, line 157: If this is a plain public key, do we actually need the extra information in the `ExtendedKey` struct?

Roasbeef (changes requested):

waddrmgr/scoped_manager.go
```diff
@@ -915,7 +915,7 @@ func (s *ScopedKeyManager) extendAddresses(ns walletdb.ReadWriteBucket,
 	}
 
 	// Derive the appropriate branch key and ensure it is zeroed when done.
-	branchKey, err := acctKey.Child(branchNum)
+	branchKey, err := acctKey.Derive(branchNum)
```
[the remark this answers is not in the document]
Roasbeef: An alternative to this would be for us to do a "migration" in order to detect any instances (by scanning the key range), and then picking if we should use the new derivation or not from there. However, this doesn't protect us from the scenario where a user imports from scratch, which puts us back at a new `aezeed` version.

waddrmgr/scoped_manager.go
```diff
@@ -308,69 +308,92 @@ func (s *ScopedKeyManager) loadAccountInfo(ns walletdb.ReadBucket,
...
-			str := fmt.Sprintf("failed to create extended private "+
-				"key for account %d", account)
-			return nil, managerError(ErrKeyChain, str, err)
+			str := fmt.Sprintf("failed to decrypt public key for "+
+				"account %d", account)
+			return nil, managerError(ErrCrypto, str, err)
 		}
-		acctInfo.acctKeyPriv = acctKeyPriv
+
+		watchOnly = watchOnly || true
```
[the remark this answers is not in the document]
wpaulino: Yeah, I've simplified them now.

wallet/import.go
```diff
@@ -1,13 +1,193 @@
...
+func (w *Wallet) validateExtendedPubKey(pubKey *hdkeychain.ExtendedKey,
+	isMasterKey bool) error {
+
+	// Private keys are not allowed.
+	if pubKey.IsPrivate() {
+		return errors.New("private keys cannot be imported")
+	}
+
+	// Public key must have a version corresponding to the current chain.
+	if !pubKey.IsForNet(w.chainParams) {
```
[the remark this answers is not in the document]
wpaulino: Fixed.

waddrmgr/scoped_manager.go
wpaulino: > So if some wallet actually started using the watch only accounts this still shouldn't break because they'd still be stored as the old, default account type? And new watch-only accounts would be stored under the correct, new type?

Correct. I haven't actually confirmed this, but as @Roasbeef mentioned, this would mean someone interacted with the wallet without going through lnd.

wallet/import.go
wpaulino: I had this fixed but wasn't pushed up at the time of your review. Should be good now, we do the same as in `ImportAccount`.

waddrmgr/scoped_manager.go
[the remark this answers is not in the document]
wpaulino: Ah, thanks for the explanation. Should be good now.

waddrmgr/address.go
```diff
@@ -177,11 +177,12 @@ func (a *managedAddress) lock() {
...
 
-// Account returns the account number the address is associated with.
+// InternalAccount returns the internal account number the address is associated
+// with.
 //
 // This is part of the ManagedAddress interface implementation.
-func (a *managedAddress) Account() uint32 {
-	return a.derivationPath.Account
+func (a *managedAddress) InternalAccount() uint32 {
+	return a.derivationPath.InternalAccount
```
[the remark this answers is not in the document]
wpaulino: Correct. The distinction was necessary for imported accounts, as they could have a 0 account number, but internally they'd be mapped to a 1 if a default account already existed before the import. This would propagate all the way down to the derivation path that gets populated in the PSBTs, resulting in a signature for the wrong key.

waddrmgr/scoped_manager.go
```diff
@@ -1686,7 +1686,8 @@ func (s *ScopedKeyManager) importPublicKey(ns walletdb.ReadWriteBucket,
 	// The start block needs to be updated when the newly imported address
 	// is before the current one.
 	s.rootManager.mtx.Lock()
-	updateStartBlock := bs.Height < s.rootManager.syncState.startBlock.Height
+	updateStartBlock := bs != nil &&
```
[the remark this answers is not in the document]
wpaulino: No, this was just required by `ImportPublicKey` as a block stamp is not exposed there yet.

wallet/import.go
wpaulino: TIL. This seems like a problem then, how can we differentiate the two? We could leave it up to the user to explicitly tell us if it's originating from another wallet, in which case we'd use and scan for P2SH change addresses, but is there anything better we could do?

wallet/import.go
wpaulino: No it's not. I kind of conflated the meaning of master xpub here with the account xpub across all my changes. I'll update those to be account xpub instead.

wallet/import.go
wpaulino: Not really, I went with it here to keep things consistent. Should we change it?

waddrmgr/scoped_manager.go
guggero: Yes, lnd users wouldn't be affected. I meant other wallets that use this code base (if there are any?), wouldn't we change the behavior for them, possibly?

wallet/import.go
Roasbeef: Yeah back then, it didn't make sense to me to still use a nested p2wkh address there, given it was a _change_ address. We'd either not support "ypub" (still makes me lol tbh), or also add in "proper" BIP 49, which is just a new key scope definition:
```go
KeyScopeBIP0049 = KeyScope{
    Purpose: 49, 
    Coin: 0, 
}

ScopeAddrMap = map[KeyScope]ScopeAddrSchema{
    KeyScopeBip0049: {
        ExternalAddrType: NestedWitnessPubKey, 
        InternalAddrType: NestedWitnessPubKey,
    }, 
}
```

wallet/import.go
Roasbeef: Hmm, no strong prefrecne, but IMO the API woul dbe more self documenting if it was just a regular public key, given we don't use any of the other information that's carried along with an extended pubkey.

waddrmgr/manager.go
```diff
@@ -1416,12 +1422,12 @@ func deriveAccountKey(coinTypeKey *hdkeychain.ExtendedKey,
...
 	// Derive the external branch as the first child of the account key.
-	if _, err := acctKey.Child(ExternalBranch); err != nil {
+	if _, err := acctKey.DeriveNonStandard(ExternalBranch); err != nil {
 		return err
 	}
 
-	// Derive the external branch as the second child of the account key.
-	_, err := acctKey.Child(InternalBranch)
+	// Derive the internal branch as the second child of the account key.
+	_, err := acctKey.DeriveNonStandard(InternalBranch)
```
Roasbeef, line 1467: 👍

waddrmgr/scoped_manager.go
```diff
@@ -1577,30 +1606,42 @@ func (s *ScopedKeyManager) ImportPrivateKey(ns walletdb.ReadWriteBucket,
...
+		s.addrSchema.ExternalAddrType,
+	)
+	if err != nil {
+		return nil, err
 	}
+	managedAddr.imported = imported
+
+	// Add the new managed address to the cache of recent addresses and
+	// return it.
+	s.addrs[addrKey(managedAddr.Address().ScriptAddress())] = managedAddr
```
Roasbeef, line 1746: 👍

waddrmgr/db.go
```diff
@@ -809,6 +826,78 @@ func serializeDefaultAccountRow(encryptedPubKey, encryptedPrivKey []byte,
...
+	// 4 bytes encrypted pubkey len + encrypted pubkey + 4 bytes master key
+	// fingerprint + 4 bytes next external index + 4 bytes next internal
+	// index + 4 bytes name len + name
+	pubLen := uint32(len(encryptedPubKey))
+	nameLen := uint32(len(name))
+	rawData := make([]byte, 20+pubLen+nameLen)
+	binary.LittleEndian.PutUint32(rawData[0:4], pubLen)
+	copy(rawData[4:4+pubLen], encryptedPubKey)
+	offset := 4 + pubLen
+	binary.LittleEndian.PutUint32(rawData[offset:offset+4], masterKeyFingerprint)
```
Roasbeef, line 889: Occurs to me that we can make all this a bit more readable by instead using `binary.Read/Write`, but we can save that for a future refactoring.

wallet/import.go
```diff
@@ -1,13 +1,232 @@
...
+	err = walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+		ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+		addr, err = scopedKeyManager.ImportPublicKey(ns, pk, nil)
+		return err
+	})
+	if err != nil {
+		return err
+	}
+
+	err = w.chainClient.NotifyReceived([]btcutil.Address{addr.Address()})
```
Roasbeef, line 221: 👍

wallet/import.go
```diff
@@ -1,13 +1,232 @@
...
+		account, err = scopedMgr.NewAccountWatchingOnly(
+			ns, name, accountPubKey, masterKeyFingerprint,
+		)
+		return err
+	})
+	if err != nil {
+		return waddrmgr.KeyScope{}, 0, err
+	}
+
+	return keyScope, account, nil
```
Roasbeef, line 178: Perhaps specify that what's being returned here is the _internal_ (auto incremented) account? Alternatively, we could retuyrn the full account info struct here as well.

Roasbeef (changes requested):
Really close now! Added a suggestions w.r.t  how to handle bip49plus vs the regular bip 49. I think that's the only proper blocking comment I have in this diff.

waddrmgr/db.go
wpaulino: Done.

wallet/import.go
wpaulino: Returned the full account info struct instead.

wallet/import.go
wpaulino: Converted to use standard public keys.

wallet/import.go
Roasbeef: Resolved this offline as we'll just modify the existing scope since wallet instances today don't ever use that scope.

Roasbeef:
I think this is only missing the change to "undo" the creation of the BIP 49+ key scope. If we don't want to break some other callers (like lnd perhaps), then we can just change the mapping in `ScopeAddrMap`.

wpaulino:
@Roasbeef @guggero PTAL! A few things have changed in the latest diff:

* New address schema override per account to support importing standard BIP-0049 keys
* BIP-0044 and BIP-0049 keys imported with the `NestedWitnessPubkey` address type now use the standard BIP-0049 address schema instead of our own (BIP-0049Plus)
* BIP-0049 keys imported with the `WitnessPubkey` address type now use the BIP-0049Plus address schema
* Account public keys now use the proper HD version for the default wallet accounts -- this isn't done for imported accounts to maintain the key consistent with what the caller provided.

Roasbeef:
@wpaulino oddly, the repo doesn't actually have any build caches...

Roasbeef:
I ran it locally and eventually got this after running it twice: 
```
/Users/roasbeef/gocode/src/github.com/btcsuite/btcwallet/waddrmgr/manager.go:1786:26: should drop = nil from declaration of var privParams; it is the zero value
/Users/roasbeef/gocode/src/github.com/btcsuite/btcwallet/waddrmgr/manager.go:1788:32: should drop = nil from declaration of var cryptoKeyPrivEnc; it is the zero value
/Users/roasbeef/gocode/src/github.com/btcsuite/btcwallet/waddrmgr/manager.go:1789:34: should drop = nil from declaration of var cryptoKeyScriptEnc; it is the zero value
+ test -z '/Users/roasbeef/gocode/src/github.com/btcsuite/btcwallet/waddrmgr/manager.go:1786:26: should drop = nil from declaration of var privParams; it is the zero value
/Users/roasbeef/gocode/src/github.com/btcsuite/btcwallet/waddrmgr/manager.go:1788:32: should drop = nil from declaration of var cryptoKeyPrivEnc; it is the zero value
/Users/roasbeef/gocode/src/github.com/btcsuite/btcwallet/waddrmgr/manager.go:1789:34: should drop = nil from declaration of var cryptoKeyScriptEnc; it is the zero value'
```

I don't think this has been popping up for any other PRs though?

Roasbeef:
The first time it ran it actually modified a set of checked in protos, then after running it again it finally sent out a proper error.

waddrmgr/db.go
```diff
@@ -102,6 +109,18 @@ type dbDefaultAccountRow struct {
...
+// dbWatchOnlyAccountRow houses additional information stored about a watch-only
+// account in the databse.
+type dbWatchOnlyAccountRow struct {
+	dbAccountRow
+	pubKeyEncrypted      []byte
+	masterKeyFingerprint uint32
+	nextExternalIndex    uint32
+	nextInternalIndex    uint32
+	name                 string
+	addrSchema           *ScopeAddrSchema
```
Roasbeef, line 121: 👍

Roasbeef (approved):
LGTM 🏊‍♀️

guggero (changes requested):
I think we need to use a default version when serializing instead of returning an error. This currently causes the lnd itests to fail.

waddrmgr/scoped_manager.go
```diff
@@ -2085,3 +2143,49 @@ func (s *ScopedKeyManager) ForEachInternalActiveAddress(ns walletdb.ReadBucket,
...
+	case wire.MainNet:
+		switch s.scope {
+		case KeyScopeBIP0044:
+			version = HDVersionMainNetBIP0044
+		case KeyScopeBIP0049Plus:
+			version = HDVersionMainNetBIP0049
+		case KeyScopeBIP0084:
+			version = HDVersionMainNetBIP0084
+		default:
+			return nil, fmt.Errorf("unsupported scope %v", s.scope)
```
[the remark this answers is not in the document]
wpaulino: I had already fixed this, just didn't update the dependency in lnd again.

guggero (approved):
Travis is failing, otherwise LGTM :100:

wpaulino:
The linter doesn't produce any failures locally, and it doesn't seem to return what's wrong on Travis. Waiting for @Roasbeef to report back.

Roasbeef:
I can confirm that `./goclean.sh` passes for me locally, not sure what's up with Travis....

Merging so we can see if the issue perists once it's running against master.

PR #733, opened 2021-02-18
waddrmgr+wallet: support transaction creation and signing for watch-only accounts

wpaulino:
This PR aims to address the missing gaps for watch-only accounts to actually be used in the context of transaction creation and signing. It also includes some minor changes and surfaces some additional account information that will serve useful for higher level applications and their use of watch-only accounts.

Depends on https://github.com/btcsuite/btcwallet/pull/732.

### DIFF
--- a/rpc/legacyrpc/methods.go
+++ b/rpc/legacyrpc/methods.go
@@ -1359,13 +1359,14 @@ func makeOutputs(pairs map[string]btcutil.Amount, chainParams *chaincfg.Params) 
 // It returns the transaction hash in string format upon success
 // All errors are returned in btcjson.RPCError format
 func sendPairs(w *wallet.Wallet, amounts map[string]btcutil.Amount,
-	account uint32, minconf int32, feeSatPerKb btcutil.Amount) (string, error) {
+	keyScope waddrmgr.KeyScope, account uint32, minconf int32,
+	feeSatPerKb btcutil.Amount) (string, error) {
 
 	outputs, err := makeOutputs(amounts, w.ChainParams())
 	if err != nil {
 		return "", err
 	}
-	tx, err := w.SendOutputs(outputs, account, minconf, feeSatPerKb, "")
+	tx, err := w.SendOutputs(outputs, &keyScope, account, minconf, feeSatPerKb, "")
 	if err != nil {
 		if err == txrules.ErrAmountNegative {
 			return "", ErrNeedPositiveAmount
@@ -1433,7 +1434,7 @@ func sendFrom(icmd interface{}, w *wallet.Wallet, chainClient *chain.RPCClient) 
 		cmd.ToAddress: amt,
 	}
 
-	return sendPairs(w, pairs, account, minConf,
+	return sendPairs(w, pairs, waddrmgr.KeyScopeBIP0044, account, minConf,
 		txrules.DefaultRelayFeePerKb)
 }
 
@@ -1475,7 +1476,7 @@ func sendMany(icmd interface{}, w *wallet.Wallet) (interface{}, error) {
 		pairs[k] = amt
 	}
 
-	return sendPairs(w, pairs, account, minConf, txrules.DefaultRelayFeePerKb)
+	return sendPairs(w, pairs, waddrmgr.KeyScopeBIP0044, account, minConf, txrules.DefaultRelayFeePerKb)
 }
 
 // sendToAddress handles a sendtoaddress RPC request by creating a new
@@ -1511,7 +1512,7 @@ func sendToAddress(icmd interface{}, w *wallet.Wallet) (interface{}, error) {
 	}
 
 	// sendtoaddress always spends from the default account, this matches bitcoind
-	return sendPairs(w, pairs, waddrmgr.DefaultAccountNum, 1,
+	return sendPairs(w, pairs, waddrmgr.KeyScopeBIP0044, waddrmgr.DefaultAccountNum, 1,
 		txrules.DefaultRelayFeePerKb)
 }
 
--- a/rpc/legacyrpc/methods.go
+++ b/rpc/legacyrpc/methods.go
@@ -1293,20 +1293,14 @@ func listAllTransactions(icmd interface{}, w *wallet.Wallet) (interface{}, error
 func listUnspent(icmd interface{}, w *wallet.Wallet) (interface{}, error) {
 	cmd := icmd.(*btcjson.ListUnspentCmd)
 
-	var addresses map[string]struct{}
-	if cmd.Addresses != nil {
-		addresses = make(map[string]struct{})
-		// confirm that all of them are good:
-		for _, as := range *cmd.Addresses {
-			a, err := decodeAddress(as, w.ChainParams())
-			if err != nil {
-				return nil, err
-			}
-			addresses[a.EncodeAddress()] = struct{}{}
+	if cmd.Addresses != nil && len(*cmd.Addresses) > 0 {
+		return nil, &btcjson.RPCError{
+			Code:    btcjson.ErrRPCInvalidParameter,
+			Message: "Filtering by addresses has been deprecated",
 		}
 	}
 
-	return w.ListUnspent(int32(*cmd.MinConf), int32(*cmd.MaxConf), addresses)
+	return w.ListUnspent(int32(*cmd.MinConf), int32(*cmd.MaxConf), "")
 }
 
 // lockUnspent handles the lockunspent command.
--- a/waddrmgr/scoped_manager.go
+++ b/waddrmgr/scoped_manager.go
@@ -2186,6 +2186,22 @@ func (s *ScopedKeyManager) ForEachInternalActiveAddress(ns walletdb.ReadBucket,
 	return nil
 }
 
+// IsWatchOnlyAccount determines if the given account belonging to this scoped
+// manager is set up as watch-only.
+func (s *ScopedKeyManager) IsWatchOnlyAccount(ns walletdb.ReadBucket,
+	account uint32) (bool, error) {
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	acctInfo, err := s.loadAccountInfo(ns, account)
+	if err != nil {
+		return false, err
+	}
+
+	return acctInfo.acctKeyPriv == nil, nil
+}
+
 // cloneKeyWithVersion clones an extended key to use the version corresponding
 // to the manager's key scope. This should only be used for non-watch-only
 // accounts as they are stored within the database using the legacy BIP-0044
--- a/waddrmgr/scoped_manager.go
+++ b/waddrmgr/scoped_manager.go
@@ -72,6 +72,12 @@ type DerivationPath struct {
 	// Index is the final child in the derivation path. This denotes the
 	// key index within as a child of the account and branch.
 	Index uint32
+
+	// MasterKeyFingerprint represents the fingerprint of the root key (also
+	// known as the key with derivation path m/) corresponding to the
+	// account public key. This may be required by some hardware wallets for
+	// proper identification and signing.
+	MasterKeyFingerprint uint32
 }
 
 // KeyScope represents a restricted key scope from the primary root key within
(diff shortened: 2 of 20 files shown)


wpaulino:
cc @Roasbeef @guggero

wallet/psbt.go
```diff
@@ -227,6 +232,8 @@ func (w *Wallet) FundPsbt(packet *psbt.Packet, account uint32,
 //
 // NOTE: This method does NOT publish the transaction after it's been finalized
 // successfully.
+//
+// TODO: require account and check if watch only to avoid signing.
```
Roasbeef: Require account?

rpc/legacyrpc/methods.go
```diff
@@ -1372,13 +1372,14 @@ func makeOutputs(pairs map[string]btcutil.Amount, chainParams *chaincfg.Params)
 // It returns the transaction hash in string format upon success
 // All errors are returned in btcjson.RPCError format
 func sendPairs(w *wallet.Wallet, amounts map[string]btcutil.Amount,
-	account uint32, minconf int32, feeSatPerKb btcutil.Amount) (string, error) {
+	keyScope waddrmgr.KeyScope, account uint32, minconf int32,
```
Roasbeef, line 1369: If we already have the account here (which is the internal auto incremented (?) account), then isn't that sufficient? Given that we can look up that account, and then use the keyscopes it defines.

Roasbeef: Or is there a case where you'd want to force an account to use a different keyscope than what we store with it?

wallet/createtx.go
```diff
@@ -239,8 +250,14 @@ func (w *Wallet) findEligibleOutputs(dbtx walletdb.ReadTx, account uint32, minco
...
-		_, addrAcct, err := w.Manager.AddrAccount(addrmgrNs, addrs[0])
-		if err != nil || addrAcct != account {
+		scopedMgr, addrAcct, err := w.Manager.AddrAccount(addrmgrNs, addrs[0])
+		if err != nil {
+			continue
+		}
+		if keyScope != nil && scopedMgr.Scope() != *keyScope {
+			continue
+		}
+		if addrAcct != account {
```
Roasbeef, line 282: Given that the internal account (maybe we should find another term for this to make it less confusing, accountID?) is unique, isn't this check enough, and we don't need the keyscope check above as well?

wallet/psbt.go
```diff
@@ -24,17 +25,21 @@ import (
...
+// externally and no additional inputs are added. If the specified inputs aren't
+// enough to fund the outputs with the given fee rate, an error is returned.
 //
 // NOTE: A caller of the method should hold the global coin selection lock of
 // the wallet. However, no UTXO specific lock lease is acquired for any of the
 // selected/validated inputs by this method. It is in the caller's
 // responsibility to lock the inputs before handing the partial transaction out.
-func (w *Wallet) FundPsbt(packet *psbt.Packet, account uint32,
-	feeSatPerKB btcutil.Amount) (int32, error) {
+func (w *Wallet) FundPsbt(packet *psbt.Packet, keyScope *waddrmgr.KeyScope,
```
Roasbeef, line 42: Same question here, but seems the comment indicates that it's possible for several scoped key managers to have the same default account of 0, so this additional filtration step is needed?

Roasbeef (changes requested):
One step closer! The diff reads well, but I'm having a hard time wrapping my head around exactly what an `account uint32` means in various areas of the codebase now. Perhaps we should introduce a _new_ term like `accountID` that refers to the auto incrementing account number within a given scoped key manager?

guggero (approved):
LGTM, barring Laolu's comments :tada:

wallet/utxos.go
```diff
@@ -143,9 +151,19 @@ func (w *Wallet) FetchInputInfo(prevOut *wire.OutPoint) (*wire.MsgTx,
...
-	}, confs, nil
+			Value:    txDetail.TxRecord.MsgTx.TxOut[prevOut.Index].Value,
+			PkScript: pkScript,
+		}, &psbt.Bip32Derivation{
+			PubKey:               pubKeyAddr.PubKey().SerializeCompressed(),
+			MasterKeyFingerprint: 0, // TODO
+			Bip32Path: []uint32{
+				keyScope.Purpose + hdkeychain.HardenedKeyStart,
+				keyScope.Coin + hdkeychain.HardenedKeyStart,
+				derivationPath.Account,
```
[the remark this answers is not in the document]
wpaulino: It's already accounted for in `DerivationInfo`.

wallet/psbt.go
wpaulino: This was a WIP TODO, it was addressed in a later commit.

rpc/legacyrpc/methods.go
wpaulino: Account numbers are not guaranteed to be consistent throughout the different key scopes except for the default account. The key scope requirement here along with the internal account number allow us to uniquely identify an account.

wallet/createtx.go
wpaulino: They're not unique, each scoped key manager tracks accounts individually, which is why the key scope is required as well.

wallet/psbt.go
wpaulino: Rephrased the comment. Ideally, the key scope wouldn't be optional (note the pointer type) in order to uniquely identify an account and only spend inputs from it. For the case of funding a transaction from the PoV of lnd however, there are cases where we'd want to fund it with a single call of `FundPsbt` that can spend inputs from the two default key scopes it supports (BIP-0049 and BIP-0084).

waddrmgr/scoped_manager.go
```diff
@@ -342,7 +340,9 @@ func (s *ScopedKeyManager) loadAccountInfo(ns walletdb.ReadBucket,
 		nextInternalIndex: row.nextInternalIndex,
 	}
 
-	if !s.rootManager.isLocked() {
+	watchOnly := s.rootManager.WatchOnly() || len(acctInfo.acctKeyEncrypted) == 0
```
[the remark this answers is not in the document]
wpaulino: Thanks for the pointer, I fixed a case I introduced.

Roasbeef:
Needs a rebase!

wpaulino:
@Roasbeef rebased, though Travis keeps failing. Good thing we have https://github.com/btcsuite/btcwallet/pull/738 now.

guggero:
I've tested this with the `lnd` PR and a Coldcard. Turns out my PSBT code I added a few months back was incomplete and didn't fully support signing NP2WKH inputs because the redeem script was missing.
Can we add the following patch to this PR to fix it?

```
diff --git a/wallet/psbt.go b/wallet/psbt.go
index fc9085b4..a0617d82 100644
--- a/wallet/psbt.go
+++ b/wallet/psbt.go
@@ -104,9 +104,21 @@ func (w *Wallet) FundPsbt(packet *psbt.Packet, keyScope *waddrmgr.KeyScope,
 			}
 
 			// We don't want to include the witness or any script
-			// just yet.
+			// on the unsigned TX just yet.
 			packet.UnsignedTx.TxIn[idx].Witness = wire.TxWitness{}
 			packet.UnsignedTx.TxIn[idx].SignatureScript = nil
+
+			// For nested P2WKH we need to add the redeem script to
+			// the input, otherwise an offline wallet won't be able
+			// to sign for it. For normal P2WKH this will be nil.
+			addr, witnessProgram, _, err := w.scriptForOutput(utxo)
+			if err != nil {
+				return fmt.Errorf("error fetching UTXO "+
+					"script: %v", err)
+			}
+			if addr.AddrType() == waddrmgr.NestedWitnessPubKey {
+				packet.Inputs[idx].RedeemScript = witnessProgram
+			}
 		}
 
 		return nil
diff --git a/wallet/signer.go b/wallet/signer.go
index 390022bf..881c8510 100644
--- a/wallet/signer.go
+++ b/wallet/signer.go
@@ -5,6 +5,7 @@
 package wallet
 
 import (
+	"fmt"
 	"github.com/btcsuite/btcd/btcec"
@@ -12,30 +13,24 @@ import (
 )
 
-// PrivKeyTweaker is a function type that can be used to pass in a callback for
-// tweaking a private key before it's used to sign an input.
-type PrivKeyTweaker func(*btcec.PrivateKey) (*btcec.PrivateKey, error)
-
-// ComputeInputScript generates a complete InputScript for the passed
-// transaction with the signature as defined within the passed SignDescriptor.
-// This method is capable of generating the proper input script for both
-// regular p2wkh output and p2wkh outputs nested within a regular p2sh output.
-func (w *Wallet) ComputeInputScript(tx *wire.MsgTx, output *wire.TxOut,
-	inputIndex int, sigHashes *txscript.TxSigHashes,
-	hashType txscript.SigHashType, tweaker PrivKeyTweaker) (wire.TxWitness,
-	[]byte, error) {
+// scriptForOutput returns the address, witness program and redeem script for a
+// given UTXO. An error is returned if the UTXO does not belong to our wallet or
+// it is not a managed pubKey address.
+func (w *Wallet) scriptForOutput(
+	output *wire.TxOut) (waddrmgr.ManagedPubKeyAddress, []byte, []byte,
+	error) {
 
 	// First make sure we can sign for the input by making sure the script
 	// in the UTXO belongs to our wallet and we have the private key for it.
 	walletAddr, err := w.fetchOutputAddr(output.PkScript)
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, nil, err
 	}
 
-	pka := walletAddr.(waddrmgr.ManagedPubKeyAddress)
-	privKey, err := pka.PrivKey()
-	if err != nil {
-		return nil, nil, err
+	pubKeyAddr, ok := walletAddr.(waddrmgr.ManagedPubKeyAddress)
+	if !ok {
+		return nil, nil, nil, fmt.Errorf("address %s is not a "+
+			"p2wkh or np2wkh address", walletAddr.Address())
 	}
 
 	var (
@@ -46,8 +41,8 @@ func (w *Wallet) ComputeInputScript(tx *wire.MsgTx, output *wire.TxOut,
 	switch {
 	// If we're spending p2wkh output nested within a p2sh output, then
 	// we'll need to attach a sigScript in addition to witness data.
-	case pka.AddrType() == waddrmgr.NestedWitnessPubKey:
-		pubKey := privKey.PubKey()
+	case walletAddr.AddrType() == waddrmgr.NestedWitnessPubKey:
+		pubKey := pubKeyAddr.PubKey()
 		pubKeyHash := btcutil.Hash160(pubKey.SerializeCompressed())
 
 		// Next, we'll generate a valid sigScript that will allow us to
@@ -58,18 +53,18 @@ func (w *Wallet) ComputeInputScript(tx *wire.MsgTx, output *wire.TxOut,
 			pubKeyHash, w.chainParams,
 		)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, nil, err
 		}
 		witnessProgram, err = txscript.PayToAddrScript(p2wkhAddr)
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, nil, err
 		}
 
 		bldr := txscript.NewScriptBuilder()
 		bldr.AddData(witnessProgram)
 		sigScript, err = bldr.Script()
 		if err != nil {
-			return nil, nil, err
+			return nil, nil, nil, err
 		}
 
 	// Otherwise, this is a regular p2wkh output, so we include the
@@ -81,6 +76,32 @@ func (w *Wallet) ComputeInputScript(tx *wire.MsgTx, output *wire.TxOut,
 		witnessProgram = output.PkScript
 	}
 
+	return pubKeyAddr, witnessProgram, sigScript, nil
+}
+
+// PrivKeyTweaker is a function type that can be used to pass in a callback for
+// tweaking a private key before it's used to sign an input.
+type PrivKeyTweaker func(*btcec.PrivateKey) (*btcec.PrivateKey, error)
+
+// ComputeInputScript generates a complete InputScript for the passed
+// transaction with the signature as defined within the passed SignDescriptor.
+// This method is capable of generating the proper input script for both
+// regular p2wkh output and p2wkh outputs nested within a regular p2sh output.
+func (w *Wallet) ComputeInputScript(tx *wire.MsgTx, output *wire.TxOut,
+	inputIndex int, sigHashes *txscript.TxSigHashes,
+	hashType txscript.SigHashType, tweaker PrivKeyTweaker) (wire.TxWitness,
+	[]byte, error) {
+
+	walletAddr, witnessProgram, sigScript, err := w.scriptForOutput(output)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	privKey, err := walletAddr.PrivKey()
+	if err != nil {
+		return nil, nil, err
+	}
+
 	// If we need to maybe tweak our private key, do it now.
 	if tweaker != nil {
 		privKey, err = tweaker(privKey)
```

Roasbeef:
Nice find @guggero!

Roasbeef (approved):
LGTM 🐠

PR #734, opened 2021-02-19
wallet: allow using external wallet db

bhandras:
This PR extends the `btcwallet` with new `Create` and `Open` functions that allow users passing an external database. This is useful when using other `walletdb.DB` implementations than `bbolt`.

### DIFF
--- a/wallet/loader.go
+++ b/wallet/loader.go
@@ -58,6 +58,7 @@ type Loader struct {
 	wallet         *Wallet
 	localDB        bool
 	walletExists   func() (bool, error)
+	walletCreated  func(db walletdb.ReadWriteTx) error
 	db             walletdb.DB
 	mu             sync.Mutex
 }
@@ -129,6 +130,15 @@ func (l *Loader) RunAfterLoad(fn func(*Wallet)) {
 	}
 }
 
+// OnWalletCreated adds a function that will be executed the wallet structure
+// is initialized in the wallet database. This is useful if users want to add
+// extra fields in the same transaction (eg. to flag wallet existence).
+func (l *Loader) OnWalletCreated(fn func(walletdb.ReadWriteTx) error) {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+	l.walletCreated = fn
+}
+
 // CreateNewWallet creates a new wallet using the provided public and private
 // passphrases.  The seed is optional.  If non-nil, addresses are derived from
 // this seed.  If nil, a secure random seed is generated.
@@ -187,16 +197,17 @@ func (l *Loader) createNewWallet(pubPassphrase, privPassphrase,
 
 	// Initialize the newly created database for the wallet before opening.
 	if isWatchingOnly {
-		err := CreateWatchingOnly(
+		err := CreateWatchingOnlyWithCallback(
 			l.db, pubPassphrase, l.chainParams, bday,
+			l.walletCreated,
 		)
 		if err != nil {
 			return nil, err
 		}
 	} else {
-		err := Create(
+		err := CreateWithCallback(
 			l.db, pubPassphrase, privPassphrase, seed,
-			l.chainParams, bday,
+			l.chainParams, bday, l.walletCreated,
 		)
 		if err != nil {
 			return nil, err
--- a/wallet/loader.go
+++ b/wallet/loader.go
@@ -6,6 +6,7 @@ package wallet
 
 import (
 	"errors"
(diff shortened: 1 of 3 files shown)


wallet/loader.go
```diff
@@ -108,8 +108,37 @@ func (l *Loader) RunAfterLoad(fn func(*Wallet)) {
 func (l *Loader) CreateNewWallet(pubPassphrase, privPassphrase, seed []byte,
 	bday time.Time) (*Wallet, error) {
 
+	defer l.mu.Unlock()
+	l.mu.Lock()
```
Roasbeef, line 112: Hmm usually in practice we _flip_ this order: so we always lock first, as otherwise an early rewind can cause a panic due to unlock of a mutex that isn't locked.

wallet/loader.go
```diff
@@ -108,8 +108,37 @@ func (l *Loader) RunAfterLoad(fn func(*Wallet)) {
...
+		return nil, err
+	}
+
+	return l.createNewWallet(
+		db, pubPassphrase, privPassphrase, seed, bday, false,
+	)
+}
+
+// CreateNewWalletInDB creates a new wallet similarly to CreateNewWallet but
+// in an externally provided db.
```
Roasbeef, line 129: in -> with?

wallet/loader.go
Roasbeef: Applies to all other instances here as well, likely harmless in practice, but jumped out at me at first glance as the ordering just stuck out a bit.

wallet/loader.go
```diff
@@ -134,6 +177,12 @@ func (l *Loader) createNewWallet(pubPassphrase, privPassphrase,
 		return nil, ErrLoaded
 	}
 
+	return l.createNewWallet(
+		db, pubPassphrase, nil, nil, bday, true,
+	)
+}
+
+func (l *Loader) createLocalDB() (walletdb.DB, error) {
```
Roasbeef, line 185: Missing a `godoc` comment.

Roasbeef: Applies to the newly created `createNewWallet` method as well.

wallet/loader.go
```diff
@@ -108,8 +108,37 @@ func (l *Loader) RunAfterLoad(fn func(*Wallet)) {
...
 	bday time.Time) (*Wallet, error) {
 
+	defer l.mu.Unlock()
+	l.mu.Lock()
+
+	if l.wallet != nil {
+		return nil, ErrLoaded
+	}
+
+	db, err := l.createLocalDB()
```
Roasbeef, line 118: Alternatively, what if we jsut added the database as new argument to `NewLoader` where we take all the other non-wallet specific data? This would reduce the size of the diff, as we'd just need to check if that field is nil or not. If we don't want to explicitly break the API here, than we can use a functional argument, which also opens up the way for us extending the methods further in the future.

Roasbeef (changes requested):
Main comment is re doing this at the `NewLoader` level, which would save us from needing to add variants of existing methods for w/ DB and w/o db.

guggero (approved):
Nice, LGTM :tada:

bhandras:
> Nice, LGTM tada

Thanks for the LGTM @guggero!
I'm working on an upgrade to this PR, to be able to hook into the txn that creates the wallet with an external template fn to allow users to "decorate" the wallet with additional data (useful when using an external wallet). Will re-request review soon.

guggero (approved):
re-ACK!

Travis is failing because of a pre-existing flake. Can you remove the `t.Parallel()` on L1816 in `waddrmgr/manager_test.go`? The manager is messing with a global variable which causes the tests to fail randomly if they're run in parallel.
Removing the parallelism even seems to speed up the test on my machine.

bhandras:
> waddrmgr/manager_test.go

Thanks for the re-review! Flake fix added as a separate commit, thanks for the investigation, I didn't even notice it as I only checked the tests locally.

wallet/loader.go
```diff
@@ -134,48 +161,56 @@ func (l *Loader) createNewWallet(pubPassphrase, privPassphrase,
 		return nil, ErrLoaded
 	}
 
-	dbPath := filepath.Join(l.dbDirPath, WalletDBName)
-	exists, err := fileExists(dbPath)
+	exists, err := l.WalletExists()
```
Roasbeef, line 174: 👍

Roasbeef (approved):
LGTM 🥇

PR #736, opened 2021-03-12
wallet: list leases and parameterize duration

joostjager:
Allows settings arbitrary lock durations. Also adds a `ListLeases` method to get an overview of all current leases. This call is particularly useful to not lose track of leases when longer durations are used.

### DIFF
--- a/wtxmgr/tx.go
+++ b/wtxmgr/tx.go
@@ -125,6 +125,14 @@ type TxRecord struct {
 	SerializedTx []byte // Optional: may be nil
 }
 
+// LockedOutput is a type that contains an outpoint of an UTXO and its lock
+// lease information.
+type LockedOutput struct {
+	Outpoint   wire.OutPoint
+	LockID     LockID
+	Expiration time.Time
+}
+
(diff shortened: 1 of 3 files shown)


guggero (approved):
Very nice addition, LGTM :100:

guggero:
cc @wpaulino

wtxmgr/tx.go
```diff
@@ -1162,7 +1159,7 @@ func isKnownOutput(ns walletdb.ReadWriteBucket, op wire.OutPoint) bool {
 // already been locked to a different ID, then ErrOutputAlreadyLocked is
 // returned.
 func (s *Store) LockOutput(ns walletdb.ReadWriteBucket, id LockID,
-	op wire.OutPoint) (time.Time, error) {
+	op wire.OutPoint, duration time.Duration) (time.Time, error) {
```
Roasbeef, line 1162: This'll require a version bump for the `wtxmgr` module, but we can handle that once this PR itself lands.

Roasbeef (approved):
LGTM 🦕

PR #737, opened 2021-03-23
chain: request pruned blocks from backend peers

wpaulino:
To minimally support wallets connected to pruned nodes, we add a new subsystem that can be integrated with chain clients to request blocks that the server has already pruned. This is done by connecting to the server's full node peers and querying them directly. Ideally, this is a capability supported by the server, though this is not yet possible with bitcoind.


wpaulino:
cc @Roasbeef

chain/network_block.go
```diff
@@ -0,0 +1,536 @@
...
+					continue
+				case <-s.quit:
+					return
+				}
+			}
+
+			log.Debugf("Found %v candidate bitcoind peers", len(peers))
+		}
+
+		// With our set of candidate peers refreshed, we'll start with
```
Roasbeef, line 275: So I take it attempting to re-use the existing query interface added to neutrino didn't end up being workable? It already supports being able to multi-plex several requests across a fixed number of peers. cc @halseth

Roasbeef: For reference: https://github.com/lightninglabs/neutrino/tree/master/query

chain/network_block.go
```diff
@@ -0,0 +1,536 @@
...
+
+	var eligible []btcjson.GetPeerInfoResult
+	for _, peer := range peers {
+		rawServices, err := hex.DecodeString(peer.Services)
+		if err != nil {
+			return nil, err
+		}
+
+		services := wire.ServiceFlag(binary.BigEndian.Uint64(rawServices))
+		if services&wire.SFNodeNetwork != wire.SFNodeNetwork {
```
Roasbeef, line 397: We also want to make sure the peers can send us witness data.

chain/network_block.go
Roasbeef: It even has a peer rank abstraction that can be used to base prioritization on latency, then punish/rewards those tht actually respond quickly.

Roasbeef (changes requested):
Completed an initial pass. Stopped short when I realized how much code we can dave here by re-using the existing query manager we wrote for neutrino. What's stopping us from re-using that package? it would save a lot of code, and we'd be able to focus our testing to a single package and ideally improve that single site which'll benefit both btcwallet and neutrino independently.

halseth:
If the it's to be used also outside of Neutrino it _could be_ it needs to be abstracted in a way, but the `query` package was created with re-usability in mind, so I think it would be cool to do it :)

chain/network_block.go
```diff
@@ -0,0 +1,536 @@
...
+// handle an incoming block request.
+type blockReq struct {
+	blocks    []*chainhash.Hash
+	blockChan chan *wire.MsgBlock
+}
+
+// queryPeer represents a Bitcoin network peer that we'll query for blocks.
+type queryPeer struct {
+	*peer.Peer
+	quit chan struct{}
```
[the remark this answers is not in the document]
wpaulino: For callers waiting on the peer, in this case the caller is the WorkManager.

wpaulino:
@Roasbeef @halseth this is ready for a proper look now -- I've re-written it to use Neutrino's query API. I've tested this on both mainnet and testnet without issues. I have a WIP branch for lnd to do this, but it depends on a global daemon-level block cache (being worked on by an external contributor) to work properly, otherwise bitcoind peers stall during initial graph sync as we're sending them multiple requests for the same block.

Roasbeef:
>  I have a WIP branch for lnd to do this, but it depends on a global daemon-level block cache (being worked on by an external contributor) to work properly, otherwise bitcoind peers stall during initial graph sync as we're sending them multiple requests for the same block.

Ahh interesting find! I wonder if the same thing happens on the neutrino side since I've observed at times that requests to fetch blocks aren't properly de-duplicated. Isn't it the case though that w/ the current query API stuff, we'll eventually rotate that out to another peer due to a timeout? Or is it that we only use a single peer right now so there's no actual peer to rotate out to?

chain/pruned_block_dispatcher.go
```diff
@@ -0,0 +1,528 @@
...
+
+	_ = d.workManager.Stop()
+}
+
+// pollPeers continuously polls the backend node for new peers to establish
+// connections to.
+func (d *PrunedBlockDispatcher) pollPeers() {
+	defer d.wg.Done()
+
+	if err := d.connectToPeers(); err != nil {
```
Roasbeef, line 203: Do we need to do this if there aren't actually any blocks that need fetching yet? Otherwise we'll create some TCP connections (and keep them alive?) that we'll never actually need to use (say it's a full full node).

chain/pruned_block_dispatcher_test.go
```diff
@@ -0,0 +1,563 @@
...
+type addr struct {
+	net, address string
+}
+
+func (m addr) Network() string { return m.net }
+func (m addr) String() string  { return m.address }
+
+// pipe turns two mock connections into a full-duplex connection similar to
+// net.Pipe to allow pipe's with (fake) addresses.
+func pipe(c1, c2 *conn) (*conn, *conn) {
```
Roasbeef, line 63: Very cool usage of `io.Pipe` here!

chain/pruned_block_dispatcher_test.go
```diff
@@ -0,0 +1,563 @@
...
+		hashes[i] = &blockHash
+		blocks[blockHash] = block
+
+		prevHash = &blockHash
+	}
+
+	return hashes, blocks
+}
+
+// networkBlockTestHarness is a harness used to facilitate the testing of the
```
Roasbeef, line 96: Excellent work on this test harness! It really makes the resulting tests much easier to fllow, and also extend in the future as well.

chain/bitcoind_conn.go
```diff
@@ -142,16 +186,36 @@ func NewBitcoindConn(cfg *BitcoindConfig) (*BitcoindConn, error) {
...
-		client:        client,
-		zmqBlockConn:  zmqBlockConn,
-		zmqTxConn:     zmqTxConn,
-		rescanClients: make(map[uint64]*BitcoindClient),
-		quit:          make(chan struct{}),
+	// Only initialize the PrunedBlockDispatcher when the connected bitcoind
+	// node is pruned.
+	var prunedBlockDispatcher *PrunedBlockDispatcher
+	if chainInfo.Pruned {
+		prunedBlockDispatcher, err = NewPrunedBlockDispatcher(
```
Roasbeef, line 193: Ah ok, safe to ignore my other comment re having extra unused TCP connections sitting around.

Roasbeef (commented):
Super happy we were able to re-use the query manager from neutrino in the end. I've completed an initial pass and plan to spin up a new testnet bitcoind node to give things a spin first hand.

wpaulino:
>> I have a WIP branch for lnd to do this, but it depends on a global daemon-level block cache (being worked on by an external contributor) to work properly, otherwise bitcoind peers stall during initial graph sync as we're sending them multiple requests for the same block.

>Ahh interesting find! I wonder if the same thing happens on the neutrino side since I've observed at times that requests to fetch blocks aren't properly de-duplicated. Isn't it the case though that w/ the current query API stuff, we'll eventually rotate that out to another peer due to a timeout? Or is it that we only use a single peer right now so there's no actual peer to rotate out to?

I looked into this further and there was actually a bug in my changes when handling concurrent requests for the same block. I've addressed it in the latest diff and the block cache is no longer necessary to have this functionality work properly. You should be able to test with https://github.com/lightningnetwork/lnd/pull/5154.

chain/pruned_block_dispatcher.go
```diff
@@ -0,0 +1,554 @@
...
+	// upon connection establishment.
+	defaultPeerReadyTimeout = 15 * time.Second
+
+	// requiredServices are the requires services we require any candidate
+	// peers to signal such that we can retrieve pruned blocks from them.
+	requiredServices = wire.SFNodeNetwork | wire.SFNodeWitness
+
+	// prunedNodeService is the service bit signaled by pruned nodes on the
+	// network.
+	prunedNodeService wire.ServiceFlag = 2 << 10
```
Roasbeef, line 39: Ah yeah, we don't have this defined in `btcd` yet given we don't yet implement a pruned operating mode for it.

chain/pruned_block_dispatcher.go
```diff
@@ -0,0 +1,554 @@
...
+
+	// Make sure the channel is buffered enough to handle all blocks.
+	blockChan := make(chan *wire.MsgBlock, len(blocks))
+
+	d.blockMtx.Lock()
+	defer d.blockMtx.Unlock()
+
+	// Each GetData message can only include up to wire.MaxInvPerMsg invs,
+	// and each block consumes a single inv.
+	var reqs []*query.Request
```
Roasbeef, line 445: Style nit: can reside in a single `var` statement.

chain/pruned_block_dispatcher.go
```diff
@@ -0,0 +1,554 @@
...
+	var getData *wire.MsgGetData
+	for i, block := range blocks {
+		if getData == nil {
+			getData = wire.NewMsgGetData()
+		}
+
+		if _, ok := d.blocksQueried[*block]; !ok {
```
Roasbeef, line 452: What about adding a `debug` log here and below so we can track the number of concurrent requests for a single block in practice when this is active? So one here when we need a new query, and another when we just need to append the block channel as there's already an active query for that hash.

Roasbeef:
cc @carlaKC

chain/pruned_block_dispatcher_test.go
halseth: Agreed, great work 👍

chain/pruned_block_dispatcher_test.go
```diff
@@ -0,0 +1,593 @@
...
+	localConns  map[string]net.Conn // Connections to peers.
+	remoteConns map[string]net.Conn // Connections from peers.
+
+	dialedPeer    chan struct{}
+	queriedPeer   chan struct{}
+	blocksQueried map[chainhash.Hash]int
+
+	shouldReply uint32 // 0 == true, 1 == false
+
+	*PrunedBlockDispatcher
```
halseth, line 115: I think it would be clearer if this had its own field perhaps?

chain/pruned_block_dispatcher_test.go
```diff
@@ -0,0 +1,593 @@
...
+// query requests the given blocks from the PrunedBlockDispatcher.
+func (h *networkBlockTestHarness) query(blocks []*chainhash.Hash,
+	opts ...query.QueryOption) (<-chan *wire.MsgBlock, <-chan error) {
+
+	h.t.Helper()
+
+	blockChan, errChan := h.Query(blocks, opts...)
+	select {
+	case err := <-errChan:
+		require.NoError(h.t, err)
```
halseth, line 298: this select seems uneccessary as we don't expect it to happen?

chain/pruned_block_dispatcher_test.go
```diff
@@ -0,0 +1,593 @@
...
+
+	h.t.Helper()
+
+	select {
+	case block := <-blockChan:
+		blockHash := block.BlockHash()
+		_, ok := h.blocksQueried[blockHash]
+		require.True(h.t, ok)
+
+		expBlock, ok := h.blocks[blockHash]
```
halseth, line 389: check ok

chain/pruned_block_dispatcher_test.go
```diff
@@ -0,0 +1,593 @@
...
+	t.Parallel()
+
+	const numBlocks = wire.MaxInvPerMsg + 1
+
+	h := newNetworkBlockTestHarness(t, numBlocks, 1, 1)
+	h.start()
+	defer h.stop()
+
+	// Request all blocks. We extend the timeout of the query as this will
+	// take longer than the default 30s timeout.
```
halseth, line 488: why does it take this long in the test environment?

chain/bitcoind_conn.go
```diff
@@ -38,21 +38,49 @@ const (
 	seqNumLen = 4
 )
 
+// BitcoindConfig contains all of the parameters required to establish a
```
halseth, line 47: 👍

chain/bitcoind_conn.go
```diff
@@ -472,6 +539,45 @@ func (c *BitcoindConn) RemoveClient(id uint64) {
...
+	rpcErr, ok := err.(*btcjson.RPCError)
+	return ok && rpcErr.Code == btcjson.ErrRPCMisc &&
+		rpcErr.Message == errBlockPrunedStr
+}
+
+// GetBlock returns a raw block from the server given its hash. If the server
+// has already pruned the block, it will be retrieved from one of its peers.
+func (c *BitcoindConn) GetBlock(hash *chainhash.Hash) (*wire.MsgBlock, error) {
+	block, err := c.client.GetBlock(hash)
+	if !isBlockPrunedErr(err) || c.prunedBlockDispatcher == nil {
```
halseth, line 554: I think this would be easier to read if we just immediatly return the block if `err == nil`, then go on to handle the error cases.

halseth (approved):
Wow, this turned out to be pretty clean, very easy to follow changes and written in a very testable way! 👍 

A few comments and suggestions, when those are fixed this LGTM ✅

carlaKC (commented):
Just a few questions from me, missing some context. Great tests for this one! 

Tested out with the lnd #5154 and the issue described in #2997 (lnd goes down, channel closed, blocks pruned) and closed channel is detected 🎉

chain/pruned_block_dispatcher.go
```diff
@@ -0,0 +1,554 @@
...
+	}()
+}
+
+// SubscribeRecvMsg adds a OnRead subscription to the peer. All bitcoin messages
+// received from this peer will be sent on the returned channel. A closure is
+// also returned, that should be called to cancel the subscription.
+//
+// NOTE: This method exists to satisfy the query.Peer interface.
+func (p *queryPeer) SubscribeRecvMsg() (<-chan wire.Message, func()) {
+	return p.msgsRecvd, func() {}
```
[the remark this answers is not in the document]
wpaulino: The lifetime of the subscription is bound to the lifetime of the `queryPeer` struct and its corresponding `query.Worker` spawned by the `query.WorkManager`, so the closure returned only gets called after the peer disconnects.

The closure exists as part of the API because of how neutrino handles message subscriptions, as it needs to clean up some state internally. This isn't needed here, which is why an empty closure is returned.

chain/pruned_block_dispatcher.go
```diff
@@ -0,0 +1,554 @@
...
+		// next if any.
+		if (len(getData.InvList) > 0 && i == len(blocks)-1) ||
+			len(getData.InvList) == wire.MaxInvPerMsg {
+
+			reqs = append(reqs, &query.Request{
+				Req:        getData,
+				HandleResp: d.handleResp,
+			})
+			getData = nil
+			continue
```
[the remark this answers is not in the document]
wpaulino: Nope, removed.

chain/pruned_block_dispatcher.go
```diff
@@ -0,0 +1,554 @@
...
+		return query.Progress{
+			Progressed: false,
+			Finished:   false,
+		}
+	}
+	delete(d.blocksQueried, blockHash)
+
+	// Launch a goroutine to notify all clients of the block as we don't
+	// want to potentially block our workManager.
+	go func() {
```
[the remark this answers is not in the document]
wpaulino: Done.

chain/bitcoind_conn.go
wpaulino: Done.

chain/pruned_block_dispatcher_test.go
wpaulino: It's run in parallel along with other tests and 50k channel sends occur (which is slower when running with race enabled). There's also only one query peer to simplify the test scenario. The test when ran by itself without race takes 50s on a MacBook Pro.

chain/pruned_block_dispatcher_test.go
wpaulino: Within `PrunedBlockDispatcher.Query`, we have a call to `newRequest` which can return an error immediately. The select exists to handle that case.

chain/pruned_block_dispatcher_test.go
wpaulino: Done.

chain/pruned_block_dispatcher.go
```diff
@@ -0,0 +1,554 @@
...
+	"github.com/btcsuite/btcd/peer"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/neutrino/query"
+	"github.com/lightningnetwork/lnd/ticker"
+)
+
+const (
+	// defaultRefreshPeersInterval represents the default polling interval
+	// at which we attempt to refresh the set of known peers.
+	defaultRefreshPeersInterval = 5 * time.Second
```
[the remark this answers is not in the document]
wpaulino: Bumped to 30s. Does that work or did you have something else in mind?

chain/pruned_block_dispatcher.go
wpaulino: Done, let me know if that's what you had in mind.

Roasbeef:
Needs a rebase to clear the `go.mod`!

harding:
I quickly skimmed this diff and didn't see validation of the block's contents.  E.g., I assume you have the header hash of a block on the best chain from your trusted bitcoind and you use that hash to request that block from an untrusted peer.  I'm guessing your code from neutrino knows to validate the block's implicit merkle tree against the explicit merkle root.

What might not be obvious is that there's a number of ways to create a valid block whose merkle tree contains ambiguities that could be exploited to send you invalid block contents (e.g. I'm aware of [CVE-2012-2459](https://bitcointalk.org/index.php?topic=102395.0), [internal nodes as a tx](http://soc1024.ece.illinois.edu/BTCRelayAudit.pdf), and [CVE-2017-12842](https://bitslog.com/2018/06/09/leaf-node-weakness-in-bitcoin-merkle-tree-design/); there may be more).  This isn't an issue for LND when it receives blocks from a trusted bitcoind because modern full nodes are resistant to all of those attacks; it also isn't an issue (IMO) for Neutrino when used as a lightweight client because there are cheaper ways to attack SPV security; but it could represent a reduction in security for full-node-backed LND instances using this PR.

Happily, it has an easy fix: even when a block has been pruned, bitcoind keeps a record of its transaction count.  If you ensure that both bitcoind and your LND/neutrino agree about the number of transactions, you prevent any attacks that reinterpret a tx as a merkle node or vice-versa.  E.g.:

    $ bitcoin-cli getblock $( bitcoin-cli getblockhash 123456 )
    error code: -1
    error message:
    Block not available (pruned data)

    $ bitcoin-cli getblockheader $( bitcoin-cli getblockhash 123456 ) true | jq .nTx
    13

Related:

- https://github.com/bitcoin/bitcoin/pull/13451
- https://github.com/bitcoin/bitcoin/pull/13452

Roasbeef:
@harding that's a _really_ good point harding! Thankfully we can re-use `btcd`'s block sanity checks here easily as they don't require a UTXO set to validate against: https://pkg.go.dev/github.com/btcsuite/btcd/blockchain#CheckBlockSanity. Neutrino uses the same function whenever it needs to fetch blocks itself.

harding:
@Roasbeef checkBlockSanity looks pretty good to me.  It appears to be explicitly designed to be impervious to CVE-2012-2459.  I think it also makes internal node re-interpretation computationally infeasible, although I'd have to diagram that out to be sure.

wpaulino:
Thanks for pointing this out @harding! I've updated the PR to validate them and added a test case.

Roasbeef (approved):
LGTM 🌵

PR #740, opened 2021-03-30
wallet: add random coin selection

joostjager:
This PR adds a new coin selection strategy based on random picks.

### DIFF
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -7,6 +7,7 @@ package wallet
 
 import (
 	"fmt"
+	"math/rand"
 	"sort"
 
 	"github.com/btcsuite/btcd/btcec"
@@ -29,10 +30,6 @@ func (s byAmount) Less(i, j int) bool { return s[i].Amount < s[j].Amount }
 func (s byAmount) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
 
 func makeInputSource(eligible []wtxmgr.Credit) txauthor.InputSource {
-	// Pick largest outputs first.  This is only done for compatibility with
-	// previous tx creation code, not because it's a good idea.
-	sort.Sort(sort.Reverse(byAmount(eligible)))
-
 	// Current inputs and their total value.  These are closed over by the
 	// returned input source and reused across multiple calls.
 	currentTotal := btcutil.Amount(0)
@@ -110,7 +107,8 @@ func (s secretSource) GetScript(addr btcutil.Address) ([]byte, error) {
 // the database. A tx created with this set to true will intentionally have no
 // input scripts added and SHOULD NOT be broadcasted.
 func (w *Wallet) txToOutputs(outputs []*wire.TxOut, keyScope *waddrmgr.KeyScope,
-	account uint32, minconf int32, feeSatPerKb btcutil.Amount, dryRun bool) (
+	account uint32, minconf int32, feeSatPerKb btcutil.Amount,
+	coinSelectionStrategy CoinSelectionStrategy, dryRun bool) (
 	tx *txauthor.AuthoredTx, err error) {
 
 	chainClient, err := w.requireChainClient()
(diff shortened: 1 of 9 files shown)


wpaulino (approved):

wpaulino:
> wallet/wallet.go:138:2: U1000: field `coinSelectionStrategy` is unused (unused)
	coinSelectionStrategy CoinSelectionStrategy

Linter isn't happy.

> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

Ran the tests locally and they all passed. Looks like we'll need to increase our travis build timeout.

joostjager:
Removed unused field `coinSelectionStrategy`

wpaulino:
cc @Roasbeef

Roasbeef:
Weird error, restarting the build: 
```
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```

joostjager:
@Roasbeef on master, one of the package test times is close to 10 mins:

`ok  	github.com/btcsuite/btcwallet/wallet	575.716s`

Roasbeef:
Ah interesting, ran it locally and it took a bit of time as well. Perhaps the cache was cleared causing Travis to need to re-run it from scratch.

wallet/createtx.go
```diff
@@ -286,11 +287,30 @@ func (w *Wallet) findEligibleOutputs(dbtx walletdb.ReadTx,
 		if addrAcct != account {
 			continue
 		}
+
+		// Skip inputs that do not raise the total transaction output
```
Roasbeef, line 302: I don't think we should have this type of logic at this layer, instead it should be up to the caller to decide if these inputs should be skipped or not. As implemented, this may lead to a change in behavior resulting in outputs that would've been spent as normal, now requiring manual code to fetch and sign for in order to sweep.

Roasbeef (changes requested):

wallet/createtx.go
joostjager: Moved filtering for positively yielding inputs to the caller `Wallet.txToOutputs`. But not completely sure if this is what you meant though. Do you mean a change in behavior for future other users of `findEligibleOutputs`?

Roasbeef: I mean that there're classes of transactions the _existing_ code would produce, that this no longer will by default (since it enforces only positive yield inputs being added). As a result, people wouldn't be able to use the normal "sweep all" behavior to clear _all_ UTXOs independent of the net cost/yield (they just want to pay down their UTXO debt and aren't sensitive to the cost. 

By "this layer" I meant that imo this should be up to the caller if this should be done or not. Every once in a while when fees dip low, I have people reach out to make to figure out how to clear out all their fragmented UTXOs. With this change (to only every include inputs with positive yield), they'd have to write a custom script using the sub-servers to sweep things.

joostjager: Ah, got it. Updated branch to keep the old behavior unchanged.

It is unclear to me though why users would want to spend money to clear out fragmented utxos.

danielalexiuc: The main reason at the moment is that the current 'largest' coin selection strategy produces an ever-increasing UTXO count of dust like transactions... when you end up with a massive number of UTXOs it's hard to get a feel for how much of it is unspendable, how much to consolidate when blockchain fees are low, and how bad of a problem this really is. So it's good to be able to reset the UTXO count back to 1, even if some UTXOs are wasted in the process.

I guess this use-case will go away once we can use the 'random' strategy though... or hopefully an even better strategy in the future? Ideally I'd lilke to target an overall UTXO count range, and the strategy would do the most intelligent thing possible to keep it near that.

joostjager:
Added variable copy to please linter. There was no actual problem with the code.

Now build times out again at 10m.

Roasbeef:
@joostjager can you rebase it on top of master? We've extended the race condition timeout there which seemed to work for the last PR that landed.

joostjager:
@Roasbeef rebased, but that wasn't enough. Travis times out after not receiving output for a seemingly hard-coded duration of 10 minutes. The wallet test produces no output for longer than that on CI:

`ok  	github.com/btcsuite/btcwallet/wallet	615.442s`

Fixed it by adding `travis_wait`. An alternative fix could be to run the tests in verbose mode.

guggero:
https://github.com/btcsuite/btcwallet/pull/749 has increased the timeout to 20 minutes, which seems to be okay for master. Are you sure it doesn't work without `travis_wait`?

joostjager:
@guggero I am sure it doesn't work. This is the failed build without `travis_wait` and a go test timeout of 20m: https://travis-ci.org/github/btcsuite/btcwallet/builds/771638340?utm_source=github_status&utm_medium=notification

In #749, the total test duration may have exceeded 10m, but the wallet test stayed below:
`ok  	github.com/btcsuite/btcwallet/wallet	581.318s`

Roasbeef (approved):
LGTM 🦧

PR #742, opened 2021-04-12
chain: fix alignment of atomic integers in bitcoind conn struct

wpaulino:
As reported by https://github.com/lightningnetwork/lnd/issues/5196, the
new atomic integer isn't properly aligned which can caus panics on
32-bit operating systems.. Tof fix this issue, we move the 64-bit
integer to lay after the two 32-bit integers at the top of the struct.


Roasbeef (approved):
LGTM 🍀

PR #743, opened 2021-04-15
waddrmgr+wallet: allow account import for simnet wallets

wpaulino:
Simnet was previously left out as it didn't have defined HD versions for some of our key scopes. To allow testing importing accounts into simnet wallets, we fall back to use the mainnet HD versions.

This commit also addresses an issue with simnet wallets that would arise whenever ScopedKeyManager.AccountProperties was invoked:
`failed to retrieve account public key: unsupported net SimNet`

### DIFF
--- a/wallet/import.go
+++ b/wallet/import.go
@@ -37,7 +37,9 @@ func keyScopeFromPubKey(pubKey *hdkeychain.ExtendedKey,
 	// force the standard BIP-0049 derivation scheme (nested witness pubkeys
 	// everywhere), while a witness address type will force the standard
 	// BIP-0084 derivation scheme.
-	case waddrmgr.HDVersionMainNetBIP0044, waddrmgr.HDVersionTestNetBIP0044:
+	case waddrmgr.HDVersionMainNetBIP0044, waddrmgr.HDVersionTestNetBIP0044,
+		waddrmgr.HDVersionSimNetBIP0044:
+
 		if addrType == nil {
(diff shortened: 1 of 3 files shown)


wpaulino:
cc @Roasbeef

wallet/import.go
```diff
@@ -111,6 +113,15 @@ func (w *Wallet) isPubKeyForNet(pubKey *hdkeychain.ExtendedKey) bool {
 			version == waddrmgr.HDVersionTestNetBIP0049 ||
 			version == waddrmgr.HDVersionTestNetBIP0084
 
+	// For simnet, we'll also allow the mainnet versions since simnet
+	// doesn't have defined versions for some of our key scopes, and the
+	// mainnet versions are usually used as the default regardless of the
+	// network/key scope.
+	case wire.SimNet:
```
Roasbeef, line 120: Why not use the testnet here? IIUC, regtest wasn't affected by this as it fell through to the case above (OG testnet net magic).

Roasbeef (changes requested):

Roasbeef (approved):
LGTM 🦖

Roasbeef:
```
--- FAIL: TestPrunedBlockDispatcherMultipleQueryPeers (5.00s)
    pruned_block_dispatcher_test.go:497: expected nil err to signal completion
```

wpaulino:
Might be a one-off flake? Haven't seen it in #744 or locally.

PR #744, opened 2021-04-16
chain: implement GetNodeAddresses fallback for PrunedBlockDispatcher

wpaulino:
It's possible for bitcoind instances to only have connections to pruned nodes after its initial block download, which are incompatible with the PrunedBlockDispatcher. This would result in GetBlock requests for pruned blocks to never resolve. Since bitcoind also exposes a GetNodeAddresses RPC, which returns random reachable addresses from its address manager, we can leverage it to obtain a new candidate set of peers that we otherwise wouldn't obtain through GetPeers.

### DIFF
--- a/chain/pruned_block_dispatcher.go
+++ b/chain/pruned_block_dispatcher.go
@@ -99,6 +99,11 @@ type PrunedBlockDispatcherConfig struct {
 	// GetPeers retrieves the active set of peers known to the backend node.
 	GetPeers func() ([]btcjson.GetPeerInfoResult, error)
 
+	// GetNodeAddresses returns random reachable addresses known to the
+	// backend node. An optional number of addresses to return can be
+	// provided, otherwise 8 are returned by default.
+	GetNodeAddresses func(*int32) ([]btcjson.GetNodeAddressesResult, error)
+
 	// PeerReadyTimeout is the amount of time we'll wait for a query peer to
 	// be ready to receive incoming block requests. Peers cannot respond to
 	// requests until the version exchange is completed upon connection
@@ -256,7 +261,7 @@ func (d *PrunedBlockDispatcher) pollPeers() {
 			// If we do, attempt to establish connections until
 			// we've reached our target number.
 			if err := d.connectToPeers(); err != nil {
-				log.Warnf("Unable to establish peer "+
+				log.Warnf("Failed to establish peer "+
 					"connections: %v", err)
 				continue
 			}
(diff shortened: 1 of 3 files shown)


wpaulino:
cc @Roasbeef @halseth

chain/pruned_block_dispatcher.go
```diff
@@ -277,90 +281,149 @@ func (d *PrunedBlockDispatcher) connectToPeers() error {
...
+	// `getnodeaddresses` RPC to receive random reachable addresses. We'll
+	// also filter out any that do not meet our requirements. The nil
+	// argument will return a default number of addresses, which is
+	// currently 8. We don't care how many addresses are returned as long as
+	// 1 is returned, since this will be polled regularly if needed.
+	nodeAddrs, err := d.cfg.GetNodeAddresses(nil)
+	if err != nil {
+		return err
+	}
+	addrs = filterNodeAddrs(nodeAddrs)
```
Roasbeef, line 312: Seems we're assuming here that all these connections will succeed rather than looping like we do above to continue to get new (unique) addresses until we reach out connection goal.

chain/pruned_block_dispatcher.go
```diff
@@ -277,90 +281,149 @@ func (d *PrunedBlockDispatcher) connectToPeers() error {
...
 		if err != nil {
-			return fmt.Errorf("unable to configure query peer %v: "+
-				"%v", peer.Addr, err)
+			log.Debugf("Failed connecting to peer %v: %v", addr, err)
 		}
-		if err := d.connectToPeer(queryPeer); err != nil {
-			log.Debugf("Failed connecting to peer %v: %v",
-				peer.Addr, err)
-			continue
+		if !needMore {
```
halseth, line 297: should perhaps not use this bool if we error above

chain/pruned_block_dispatcher.go
```diff
@@ -99,6 +99,10 @@ type PrunedBlockDispatcherConfig struct {
 	// GetPeers retrieves the active set of peers known to the backend node.
 	GetPeers func() ([]btcjson.GetPeerInfoResult, error)
 
+	// GetNodeAddresses returns random reachable addresses known to the
+	// backend node.
+	GetNodeAddresses func(*int32) ([]btcjson.GetNodeAddressesResult, error)
```
halseth, line 104: should explain (or remove) the optional argument

chain/pruned_block_dispatcher.go
halseth: I guess the assumption is that this will be run at regular intervals

halseth (approved):
utACK 👍

chain/pruned_block_dispatcher.go
wpaulino: Yeah, `connectToPeers` is currently polled every 30s, so it would be handled there. The loop above for peer addresses works the same way.

Roasbeef (approved):
LGTM 🎨

PR #745, opened 2021-04-27
walletdb: add ForEachBucket to the ReadTx with bbolt implementation

bhandras:
This commit extends the ReadTx (and ReadWriteTx) interface with
ForEachBucket which can be used to iterate through all top level
buckets. This is a missing piece from the walletdb abstraction which
will allow us to iterate all keys in a walletdb opening the possibility
to build generic tools to browse and edit walletdb files regardless of
the underlying driver.


Roasbeef:
cc @guggero

guggero (approved):
Nice, LGTM :1st_place_medal:

Roasbeef (approved):
LGTM 🛵

PR #749, opened 2021-05-14
internal/prompt: allow compilation in js/wasm environment

guggero:
The wallet loader has a dependency to the internal/prompt package for
prompting the user for certain inputs (e.g. wallet password or new
seed).
This makes it impossible for projects that use the wallet as a
dependency and always provide those inputs as parameters to compile for
JavaScript/WebAssembly targets because the prompt code uses some
terminal functionality that is not available in JS syscalls.
By providing a JS specific implementation that just returns an error we
can compile the dependent projects.
Adding acutal support for prompting the user in the browser is currently
not planned as that can easily be circumvented by providing all inputs
as parameters.

cc @Roasbeef @wpaulino

### DIFF
--- a/internal/prompt/prompt_js.go
+++ b/internal/prompt/prompt_js.go
@@ -1,0 +1,32 @@
+// Copyright (c) 2015-2021 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package prompt
+
+import (
+	"bufio"
+	"fmt"
+
+	"github.com/btcsuite/btcwallet/internal/legacy/keystore"
+)
+
+func ProvideSeed() ([]byte, error) {
+	return nil, fmt.Errorf("prompt not supported in WebAssembly")
+}
+
+func ProvidePrivPassphrase() ([]byte, error) {
+	return nil, fmt.Errorf("prompt not supported in WebAssembly")
+}
+
+func PrivatePass(_ *bufio.Reader, _ *keystore.Store) ([]byte, error) {
+	return nil, fmt.Errorf("prompt not supported in WebAssembly")
+}
+
+func PublicPass(_ *bufio.Reader, _, _, _ []byte) ([]byte, error) {
+	return nil, fmt.Errorf("prompt not supported in WebAssembly")
+}
+
+func Seed(_ *bufio.Reader) ([]byte, error) {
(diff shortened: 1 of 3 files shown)


internal/prompt/prompt_js.go
```diff
@@ -0,0 +1,32 @@
...
+
+import (
+	"bufio"
+	"fmt"
+
+	"github.com/btcsuite/btcwallet/internal/legacy/keystore"
+)
+
+func ProvideSeed() ([]byte, error) {
+	return nil, fmt.Errorf("prompt not supported in WebAssembly")
```
wpaulino, line 15: Will these ever be invoked? Any reason to not panic instead?

wpaulino (commented):
> No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

Looks like we need to increase our Travis build timeout?

internal/prompt/prompt_js.go
Roasbeef: I'm guessing we don't need these directly as we don't plan to run the raw `btcwallet` binary (where these are used) directly in the WASM context.

guggero: Yeah, we should never hit this code path. I personally don't really like calling `panic()` in general but I'm fine with either option as it shouldn't matter for our use case anyway.

guggero:
I increased the test timeout to 20 minutes, let's see if that's enough for the race tests.

internal/prompt/prompt_js.go
wpaulino: A panic would give us a clear indication that this was called when not expected, in contrast to an error which could go undetected, but I agree that it doesn't make much of a difference in this case.

wpaulino (approved):

internal/prompt/prompt_js.go
guggero: Good point. Though I found that panics in WASM aren't that easy to handle or find since the process just aborts and you don't even get a nice stack trace. So an error might make a tiny bit more sense in this specific context?

Roasbeef (approved):
LGTM 🍒

PR #751, opened 2021-05-28
Prevent duplicate notifications for (un)confirmed transactions

bjarnemagnussen:
### Background

Relevant confirmed and unconfirmed transactions may be notified multiple times. The reason seems to be that there are multiple paths that lead to entrance into the `addRelevantTx` function. This will then always trigger a notification regardless if the transaction was already recorded or not.

This PR addresses this issue by adding a new method `InsertTxCheckIfExists` to the `TxStore`, which allows `addRelevantTx` to skip all the logic only required for new transactions, including notification of transactions that were already recorded.

This should fix lightningnetwork/lnd#2267

Arguably adding better testing of this can be added, which I have only done in this PR in a simplistic manner but would improve on.

### DIFF
--- a/wallet/chainntfns.go
+++ b/wallet/chainntfns.go
@@ -280,9 +280,17 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord, 
 	// relevant.  This assumption will not hold true when SPV support is
 	// added, but until then, simply insert the transaction because there
 	// should either be one or more relevant inputs or outputs.
-	err := w.TxStore.InsertTx(txmgrNs, rec, block)
+	exists, err := w.TxStore.InsertTxCheckIfExists(txmgrNs, rec, block)
 	if err != nil {
 		return err
+	}
+
+	// If the transaction has already been recorded, we can return early.
+	// Note: Returning here is safe as we're within the context of an atomic
+	// database transaction, so we don't need to worry about the MarkUsed
+	// calls below.
+	if exists {
+		return nil
 	}
 
 	// Check every output to determine whether it is controlled by a wallet
--- a/wtxmgr/tx.go
+++ b/wtxmgr/tx.go
@@ -55,6 +55,10 @@ var (
 	// ErrOutputUnlockNotAllowed is an error returned when an output unlock
 	// is attempted with a different ID than the one which locked it.
 	ErrOutputUnlockNotAllowed = errors.New("output unlock not alowed")
+
+	// ErrDuplicateTx is returned when attempting to record a mined or
+	// unmined transaction that is already recorded.
+	ErrDuplicateTx = errors.New("transaction already exists")
 )
 
 // Block contains the minimum amount of data to uniquely identify any block on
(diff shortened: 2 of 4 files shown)


wpaulino (approved):

wallet/chainntfns.go
```diff
@@ -280,11 +280,16 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
 	// should either be one or more relevant inputs or outputs.
-	err := w.TxStore.InsertTx(txmgrNs, rec, block)
+	exists, err := w.TxStore.InsertTxCheckIfExists(txmgrNs, rec, block)
 	if err != nil {
 		return err
 	}
 
+	// If the transaction has already been recorded, we can return early.
+	if exists {
+		return nil
```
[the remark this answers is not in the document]
bjarnemagnussen: Thanks for your review! I have added your clarification to the comment

wpaulino:
cc @Roasbeef

wallet/chainntfns.go
```diff
@@ -280,11 +280,19 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
 	// relevant.  This assumption will not hold true when SPV support is
 	// added, but until then, simply insert the transaction because there
 	// should either be one or more relevant inputs or outputs.
-	err := w.TxStore.InsertTx(txmgrNs, rec, block)
+	exists, err := w.TxStore.InsertTxCheckIfExists(txmgrNs, rec, block)
 	if err != nil {
```
Roasbeef, line 284: Typo in commit message InserTxCheckIfExists -> InsertTxCheckIfExists

Also commit body should be wrapped to 80 chars

wtxmgr/tx.go
```diff
@@ -355,11 +359,30 @@ func (s *Store) deleteUnminedTx(ns walletdb.ReadWriteBucket, rec *TxRecord) erro
...
 // history.  If block is nil, the transaction is considered unspent, and the
 // transaction's index must be unset.
-func (s *Store) InsertTx(ns walletdb.ReadWriteBucket, rec *TxRecord, block *BlockMeta) error {
+func (s *Store) InsertTx(ns walletdb.ReadWriteBucket, rec *TxRecord,
+	block *BlockMeta) error {
+	_, err := s.InsertTxCheckIfExists(ns, rec, block)
+	return err
+}
+
+// InsertTxCheckIfExists records a transaction as belonging to a wallet's
```
Roasbeef, line 368: Not the most scrutable name, can't think of any other good alternatives, so it should do...

Roasbeef (approved):
Solid fix!

Re testing, I think the final thing we want here is an integration test at the `lnd` level to confirm that the motivating issue is resolved: users no longer receive double transaction notifications on the RPC level.

bjarnemagnussen:
I have changed the commit title and message to address the issue.

> Re testing, I think the final thing we want here is an integration test at the `lnd` level to confirm that the motivating issue is resolved: users no longer receive double transaction notifications on the RPC level.

How would the procedure be for this? Merge this PR and make a PR in the LND repo that updates the dependency and adds the integration test; or first create a PR in the LND repo that uses this fork as dependency (with integration tests) and if it works merge this PR and then update the dependency inside the PR of the LND repo?

wallet/chainntfns.go
bjarnemagnussen: Changed!

Roasbeef:
> How would the procedure be for this? Merge this PR and make a PR in the LND repo that updates the dependency and adds the integration test

Yeah this exactly!

bjarnemagnussen:
I was looking into adding tests for this, but I cannot re-create this issue in the testing framework. Always only one notification is received on the `lntest.HarnessNode`. 

Of course manually building LND and using the gRPC `SubscribeTransactions` does as expected deliver duplicate notifications if not updated to the new `btcwallet` dependency, which resolves this.

Roasbeef:
> Of course manually building LND and using the gRPC SubscribeTransactions does as expected deliver duplicate notifications if not updated to the new btcwallet dependency, which resolves this.

You mean that if you make a test with this patch, then duplicate notifications aren't received? Isn't that to be expected? So it would be something like:
  * Add the test (it fails) 
  * Update go.mod (it now passes)

? 

Or you mean that the test framework itself is suppressing double notification somewhere in the higher level logic it offers?

bjarnemagnussen:
I wanted to add a test case that would trigger if duplicate notifications ever happen again. So I tried to add the test cases before updating the dependency in `go.mod` and realized that something higher up in the logic in the testing framework is preventing the duplicate notifications from happening. 

First I lazily tried to just add a select-case statement to check for any additional notifications here:
https://github.com/lightningnetwork/lnd/blob/7176a212f76f3539d0c0b89bdc4457e647fdf7c2/lnwallet/test/test_interface.go#L1514-L1529

I then also created an integration test where I created the nodes from the `lntest.NetworkHarness.NewNode` method and then set up everything to listen to notifications and creating them via `lntest.NetworkHarness.SendCoins`, etc. However this also didn't trigger duplicate notifications. 

The original issue is as described that "[the] transaction gets notified independent from the block notification once it’s confirmed and it’s also notified when processing the block notification containing the transaction." And I think in the test framework maybe only one of these notification hooks are actually active? I don't see how to change that though.

Roasbeef:
Interesting, are you able to reproduce at the _unit test_ level with `SubscribeTransactions`? So this would be directly in btcwallet, skipping the `lnd` wrapper RPC logic.

bjarnemagnussen:
> Interesting, are you able to reproduce at the _unit test_ level with `SubscribeTransactions`? So this would be directly in btcwallet, skipping the `lnd` wrapper RPC logic.

Inside LND I am not sure where the test files exist for btcwallet? It seems that the tests only exist for the lnwallet interface (for which the issue of not firing duplicate notifications persists)?

PR #757, opened 2021-07-19
wallet+walletdb: remove manual DB transactions, use custom `View/Update` implementation when provided

guggero:
This PR makes it possible to use a remote DB backend such as `etcd` as the wallet DB backend.
Because a remote DB backend might implement its own logic for committing or rolling back a DB transaction, we shouldn't create manual transactions by using `BeginReadTx` or `BeginReadWriteTx`.
Instead we should always use the `walletdb.View()` or `walletdb.Update()` calls.

We also add a new `ExtendedDB` interface to `walletdb` that can specify its own `View()` and `Update()` methods. If such an extended DB is given to the `walletdb` package, its methods are used instead of the default implementation in `walletdb`.

cc @bhandras, @wpaulino


wallet/createtx.go
```diff
@@ -109,148 +109,159 @@ func (s secretSource) GetScript(addr btcutil.Address) ([]byte, error) {
...
-			output := output
-
-			if !inputYieldsPositively(&output, feeSatPerKb) {
-				continue
-			}
-
-			positivelyYielding = append(positivelyYielding, output)
+	var tx *txauthor.AuthoredTx
+	err = walletdb.Update(w.db, func(dbtx walletdb.ReadWriteTx) error {
+		addrmgrNs, changeSource, err := w.addrMgrWithChangeSource(
```
joostjager, line 127: This doesn't look thread-safe, but from our offline discussion I understand that etcd no longer guarantees that this code isn't executed concurrently like it was the case for bbolt.

I think this applies to other changes too.

bhandras: Yes, one key thing about using the `View` and `Update` closures is that we only guarantee that transactions are serializable, but there's no guarantee of a single `Update` at a time. Also the closure may retry so any logic inside has to be aware of this.

joostjager: Echoing the offline discussion here: my suggestion would be to for now use the familiar global write lock with etcd that bbolt uses too to avoid the risk of breaking anything. The etcd backend plugin itself is probably more than enough risk already, let alone if the app code would need to be changed too. Data safety above performance.

guggero: Discussed this a bit more offline. The current plan is to add a lock specifically for the `etcd` DB instance that uses the `walletdb` to avoid concurrency issues. If this slows us down too much in `lnd` then we might need to re-architecture the wallet somewhat to support the reset callbacks and account for proper thread safety.

joostjager: Are you sure that the other database instances are completely safe to use without the lock? As a user, my preference would be to wrap them all in a lock for the first release. Lightning data is critical.

guggero: > Are you sure that the other database instances are completely safe to use without the lock?

No, not yet. This will need to be made sure and is being discussed in other PRs.
But assuming that there is a global lock for the wallet DB, can we proceed with the PR as is?

joostjager: Yes looks good then.

Roasbeef: > This doesn't look thread-safe and from our offline discussion I understand that etcd no longer guarantees that this code isn't executed concurrently like it was the case for bbolt.

Check out `txCreator` this only ever runs serially.

wallet/createtx.go
```diff
@@ -109,148 +109,159 @@ func (s secretSource) GetScript(addr btcutil.Address) ([]byte, error) {
...
-		for _, output := range eligible {
-			output := output
-
-			if !inputYieldsPositively(&output, feeSatPerKb) {
-				continue
-			}
-
-			positivelyYielding = append(positivelyYielding, output)
+	var tx *txauthor.AuthoredTx
+	err = walletdb.Update(w.db, func(dbtx walletdb.ReadWriteTx) error {
```
bhandras, line 126: I think there's a (preexisting) race here with the `BlockStamp()` since we may be past the block by the time the transaction commits. It's not related to the `Update` refactor and not sure if it's porblematic.

walletdb/interface.go
```diff
@@ -207,23 +208,60 @@ type DB interface {
...
+}
+
 // View opens a database read transaction and executes the function f with the
 // transaction passed as a parameter.  After f exits, the transaction is rolled
 // back.  If f errors, its error is returned, not a rollback error (if any
 // occur).
 func View(db DB, f func(tx ReadTx) error) error {
+	// Allow external/remote DB driver implementations to overwrite the
+	// default view behavior.
+	if edb, ok := db.(ExtendedDB); ok {
```
bhandras, line 261: To be fully safe I think we need some deeper changes, because there are some cases in the wallet where in the `View` and `Update` closures mutate collections inside the closure. We won't retry them since everything is serial and we also use the walletdb rwlocked in LND but the `reset` closure could be used to reset those collections where needed.

walletdb/interface.go
```diff
@@ -207,23 +208,60 @@ type DB interface {
...
+	// attempt to optimistically combine the invocation of several
 	// transaction functions into a single db write transaction.
 	Batch(func(tx ReadWriteTx) error) error
 }
 
+// ExtendedDB is a special version of the main DB interface that provides their
+// own implementation of the View and Update methods. The walletdb package will
+// use those methods instead of the package global View/Update functions if the
+// given DB is compatible with this extended interface.
+type ExtendedDB interface {
```
bhandras, line 227: I'm thinking if it is feasible to instead of introducing the `ExtendedDB` interface, we could just change the `DB` interface to include `View` and `Update` with reset and we could keep the old manual txn based `walletdb.View` and `walletdb.Update` intact to simplify upgrades for other implementors? Then internally in the wallet we could just use the new extended `DB`. wdyt?

bhandras (approved):
Changes look good with the assumption that the wallet really is locked for serial access. Some suggestions I put in comments that we may consider doing to make it fully safe for database retries (which won't ever happen until serial access is guaranteed).

wallet/createtx.go
guggero: I think that's okay. If a new block comes in while we're in this DB transaction, we just won't "see" any UTXOs in that block. I think with the wallet's own (internal) unique coin lock maybe we can't process new blocks while a wallet operation is ongoing anyway.

walletdb/interface.go
guggero: Yeah I guess we'll have to look at every DB transaction individually to make sure they're either serialized by the wallet's own internal lock or are `reset()` capable if not. I implemented your suggestion about adding the `View()` and `Update()` methods to the DB interface directly. That will make it possible to start migrating/upgrading individual code blocks in the future.

walletdb/interface.go
guggero: Good idea, simplifies quite a lot! I didn't go ahead and rename/change anything in the larger code base. But we can now do so package by package if we wish to.

walletdb/interface.go
```diff
@@ -205,16 +206,39 @@ type DB interface {
 
 	// Close cleanly shuts down the database and syncs all data.
 	Close() error
+
+	// PrintStats returns all collected stats pretty printed into a string.
+	PrintStats() string
```
bhandras, line 211: Nice!

guggero:
Bump, @wpaulino, @Roasbeef

walletdb/error.go
```diff
@@ -39,6 +39,10 @@ var (
 
 	// ErrInvalid is returned if the specified database is not valid.
 	ErrInvalid = errors.New("invalid database")
+
+	// ErrDryRunRollBack is returned if a database transaction should be
+	// rolled back because its changes were a dry-run only.
+	ErrDryRunRollBack = errors.New("dry run only; should roll back")
```
Roasbeef, line 45: 👍

walletdb/interface.go
```diff
@@ -205,16 +206,39 @@ type DB interface {
...
+
+	// View opens a database read transaction and executes the function f
+	// with the transaction passed as a parameter. After f exits, the
+	// transaction is rolled back. If f errors, its error is returned, not a
+	// rollback error (if any occur). The passed reset function is called
+	// before the start of the transaction and can be used to reset
+	// intermediate state. As callers may expect retries of the f closure
+	// (depending on the database backend used), the reset function will be
+	// called before each retry respectively.
+	View(f func(tx ReadTx) error, reset func()) error
```
Roasbeef, line 221: Hmm, so which one (the interface method or the package method) is a caller intended to use?

Roasbeef (approved):
Changes look good to me, just one comment about the addition of the interface methods, and how callers are meant to decide which of them to use. 

In the future, I guess we could make a v2/v3 version that pushes all callers towards the interface variant.

walletdb/interface.go
guggero: Good question. Added a comment to both.

walletdb/interface.go
```diff
@@ -205,16 +206,45 @@ type DB interface {
...
+	// if f did not error, the transaction is committed. Otherwise, if f did
+	// error, the transaction is rolled back. If the rollback fails, the
+	// original error returned by f is still returned. If the commit fails,
+	// the commit error is returned. As callers may expect retries of the f
+	// closure (depending on the database backend used), the reset function
+	// will be called before each retry respectively.
+	//
+	// NOTE: For new code, this method should be used directly instead of
+	// the package level Update() function.
+	Update(f func(tx ReadWriteTx) error, reset func()) error
```
bhandras, line 237: I think we should also add `Batch(f func(rx ReadWriteTx) error, reset func()) error` to this interface.

bhandras: And then add the `reset()` to the `bdb` implementation.

Roasbeef: Do we need that if we already have `Batch`? Given that due to the nature of `Batch`, the caller needs to be inherently able to handle retried DB transactions.

PR #758, opened 2021-08-03
mod+wallet: use btcd/mempool IsDust for calculating mempool dust

Crypt-iQ:
We can now get rid of our incorrect dust calculation which did not give exact values for segwit outputs as it was based on spending a P2PKH output instead.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/btcsuite/btcwallet
 
 require (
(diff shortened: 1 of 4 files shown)


Crypt-iQ:
@Roasbeef

Roasbeef (approved):
LGTM 🌲

PR #759, opened 2021-08-13
waddrmgr: add new DeriveFromKeyPathCache method for faster key retrieval

Roasbeef:
In this commit, we add a new method `DeriveFromKeyPathCache` that gives
callers a way to more quickly obtain a private key they know they'll be
using frequently. This method lets a caller avoid the write database
transaction as well as the EC operations to derive the key itself (BIP
32).


Roasbeef:
cc @guggero @arshbot

waddrmgr/manager_test.go
```diff
@@ -2761,3 +2762,104 @@ func testNewRawAccount(t *testing.T, _ *Manager, db walletdb.DB,
...
+			return err
+		}
+		err = Create(
+			ns, seed, pubPassphrase, privPassphrase,
+			&chaincfg.MainNetParams, fastScrypt, time.Time{},
+		)
+		if err != nil {
+			return err
+		}
+		mgr, err = Open(ns, pubPassphrase, &chaincfg.MainNetParams)
```
guggero, line 2790: Return on err here too?

waddrmgr/manager_test.go
```diff
@@ -2761,3 +2762,104 @@ func testNewRawAccount(t *testing.T, _ *Manager, db walletdb.DB,
...
+		}
+
+		derivedKey, err = managedAddr.(ManagedPubKeyAddress).PrivKey()
+		if err != nil {
+			return err
+		}
+
+		return nil
+	})
+	if err != nil {
```
guggero, line 2841: It looks like we have `testify/require` in the dependencies already, could use it here as well.

waddrmgr/scoped_manager.go
```diff
@@ -579,6 +593,75 @@ func (s *ScopedKeyManager) AccountProperties(ns walletdb.ReadBucket,
...
+// back to the normal variant.
+func (s *ScopedKeyManager) DeriveFromKeyPathCache(kp DerivationPath,
+) (*btcec.PrivateKey, error) {
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	// First, try to look up the key itself in the proper cache, if the key
+	// is here, then we don't need to do anything further.
+	privKeyVal, err := s.privKeyCache.Get(kp)
```
guggero, line 624: So even though we only use `InternalAccount`, `Branch` and `Index` of the derivation path, we still use the full derivation path as the key here. Likely not a problem if we always use this the same way. So not sure if adding a new struct just for the things we actually need as the key is overkill...

waddrmgr/scoped_manager.go
```diff
@@ -579,6 +593,75 @@ func (s *ScopedKeyManager) AccountProperties(ns walletdb.ReadBucket,
...
+	if err == nil {
+		return privKeyVal.(*cachedKey).key, nil
+	}
+
+	// If the key isn't already in the cache, then we'll try to look up the
+	// account info in the cache, if this fails, then we exit here as we
+	// can't move forward without creating a DB transaction, and the point
+	// of this method is to avoid that.
+	acctInfo, ok := s.acctInfo[kp.InternalAccount]
```
guggero, line 633: Ah, I wasn't aware that we already have this account cache that also has the extended account key. I assume getting rid of two levels of derivation is performance gain enough to warrant the additional private key cache? Or is getting rid of the DB transaction already enough and we could just do the derivation each time instead? Just thinking aloud here...

guggero (approved):
Great performance improvement, this should have quite the impact!

LGTM, just a few non-blocking comments :tada:

waddrmgr/scoped_manager.go
arshbot: I don't really see a reason to separate the two caches except for maybe code simplicity, since on retrieval you'd have to go by account first then attempt to get by index. Either way, I think the derivation on the first try (when memory and db are out of sync) is necessary

waddrmgr/scoped_manager.go
```diff
@@ -579,6 +593,75 @@ func (s *ScopedKeyManager) AccountProperties(ns walletdb.ReadBucket,
...
+// Size returns the size of this element. Rather than have the cache limit
+// based on bytes, we simply report that each element is of size 1, meaning we
+// can set our cached based on the amount of keys we want to store, rather than
+// the total size of all the keys.
+func (c *cachedKey) Size() (uint64, error) {
+	return 1, nil
+}
+
+// DeriveFromKeyPathCache is identical to DeriveFromKeyPath, however it'll fail
+// if the account refracted in the DerivationPath isn't already in the
```
arshbot, line 612: How does one ensure an account is specifically cached?

arshbot (commented):
nits and general questions

waddrmgr/scoped_manager.go
```diff
@@ -579,6 +593,75 @@ func (s *ScopedKeyManager) AccountProperties(ns walletdb.ReadBucket,
...
+// in-memory cache. Callers looking for faster private key retrieval can opt to
+// call this method, which may fail if things aren't in the cache, then fall
+// back to the normal variant.
+func (s *ScopedKeyManager) DeriveFromKeyPathCache(kp DerivationPath,
```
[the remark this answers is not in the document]
Roasbeef: > nit: meaing behind kp? misread it as pk for private key a few times

Key path, copied over from the original method below this one.

waddrmgr/scoped_manager.go
Roasbeef: > Since btcsuite is supposed to be attached to a single btcd node, would there ever be a situation where we'd need to store pks of different networks (testnet, regtest, mainnet)?

No, since a single instance is only every tracking a single chain. 

> . So not sure if adding a new struct just for the things we actually need as the key is overkill...

You mean a new struct that only has the internal account? We use things pretty consistently on the caller side of things, only setting the `InternalAccount` as that's what's actually used to derive the keys.

waddrmgr/scoped_manager.go
Roasbeef: > Ah, I wasn't aware that we already have this account cache that also has the extended account key. I assume getting rid of two levels of derivation is performance gain enough to warrant the additional private key cache? 

The idea here is just skip all the extra derivation all together since it's a pretty easy win. If we cached the account info only, we'd still need to re-derive actual final key which means the extra EC+hashing operations. 

> I don't really see a reason to separate the two caches except for maybe code simplicity, since on retrieval you'd have to go by account first then attempt to get by index

One cache is just maintaining the account information, and the top-level account key for a given key family. This cache saves us from the expensive `scrypt` operations to re-derive the encryption key to decrypt the master private key, but doesn't let us skip the extra BIP-32 operations needed to derive the actual key we need. 

Take a look at `loadAccountInfo` to see al the operations we skip by checking this cache, and `deriveKey` for the operations we skip by caching the key itself. Why have the overhead of a series of asymmetric crypto operations each time we need to: sign a new state, create an invoice, decode error responses?

waddrmgr/scoped_manager.go
Roasbeef: See the `lnd` PR, the caller can either look up the account, or cal `DeriveFromKeyPath` (which'll cache the account). Added an extra line in the comment making this more clear.

Roasbeef:
PTAL @guggero @arshbot

waddrmgr/scoped_manager.go
guggero: > You mean a new struct that only has the internal account?

Yes, only the internal account and no `MasterKeyFingerprint`. But as I said, as long as we call the cached method in a consistent manner concerning the fields being set, this shouldn't be a problem.

waddrmgr/scoped_manager.go
```diff
@@ -51,6 +52,14 @@ const (
 	HDVersionSimNetBIP0044 HDVersion = 0x0420bd3a // spub
 )
 
+const (
+	// privKeyCacheSize is the default size of the LRU cache that we'll use
+	// to cache private keys to avoid DB and EC operations within the
+	// wallet. With the default sisize, we'll allocate up to 320 KB to
+	// caching private keys (ignoring pointer overhead, etc).
+	defaultPrivKeyCacheSize = 10_0000
```
guggero, line 60: nit: underscore in wrong place :sweat_smile:

guggero (approved):
LGTM :tada:

PR #761, opened 2021-08-26
CI: switch from Travis to GitHub actions

guggero:
It seems like Travis isn't working correctly anymore so we attempt to switch the CI checks over to GitHub actions.

An example run can be seen here: https://github.com/guggero/btcwallet/actions/runs/1169645359

cc @Roasbeef


Roasbeef (approved):
LGTM 🐊

PR #764, opened 2021-09-06
wallet: watch-only account import improvements

guggero:
This PR fixes two issues that arose when trying to run a purely watch-only wallet and import accounts from xpubs only:

1. Coin selection requires the wallet to be unlocked. Since there is no unlock in watch-only mode, we need to circumvent that check.
2. When importing default accounts, the scope is well-known in advance and we don't need to take the detour via the address type. So we add a new `ImportAccountWithScope` method that both defines the key scope and address schema to use instead of guessing that information from the address type.

cc @Roasbeef

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -1169,17 +1169,27 @@ out:
 	for {
 		select {
 		case txr := <-w.createTxRequests:
-			heldUnlock, err := w.holdUnlock()
-			if err != nil {
-				txr.resp <- createTxResponse{nil, err}
-				continue
+			// If the wallet can be locked because it contains
+			// private key material, we need to prevent it from
+			// doing so while we are assembling the transaction.
+			release := func() {}
+			if !w.Manager.WatchOnly() {
+				heldUnlock, err := w.holdUnlock()
+				if err != nil {
+					txr.resp <- createTxResponse{nil, err}
+					continue
+				}
+
+				release = heldUnlock.release
 			}
+
 			tx, err := w.txToOutputs(
 				txr.outputs, txr.keyScope, txr.account,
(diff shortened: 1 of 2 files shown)


wallet/wallet.go
```diff
@@ -1169,6 +1169,20 @@ out:
...
+			// selection in that mode.
+			if w.Manager.WatchOnly() {
+				tx, err := w.txToOutputs(
+					txr.outputs, txr.keyScope, txr.account,
+					txr.minconf, txr.feeSatPerKB,
+					txr.coinSelectionStrategy, txr.dryRun,
+				)
+				txr.resp <- createTxResponse{tx, err}
+				continue
+			}
```
Roasbeef, line 1184: Seems can consolidate the code a bit here w/ something like: 
```go
var release func()
if !w.Manager.WatchOnly() {
    heldUnlock, err := w.holdUnlock()
    .... 
    release := heldUnlock.release()
}

tx, err := w.txToOutputs(
	txr.outputs, txr.keyScope, txr.account,
	txr.minconf, txr.feeSatPerKB,
	txr.coinSelectionStrategy, txr.dryRun,
)

if release != nil {
    release()
}

txr.resp <- createTxResponse{tx, err}
```

Thoughts?

Roasbeef (changes requested):
Nice fix and re-work, one small suggestion here to consolidate the code a bit more in the coin selection loop.

wallet/wallet.go
guggero: Great idea, done!

guggero:
Reworked the first commit as suggested.

Roasbeef (approved):
LGTM 📡

PR #766, opened 2021-09-23
wallet: add minimum confirmation parameter to FundPsbt

bjarnemagnussen:
Prior to this PR the `FundPsbt` method was using coin selection with a fixed requirement of at least one confirmation. Setting the new `minConfs` parameter is used for the coin selection when no inputs are provided to define this minimum confirmation requirement.


bjarnemagnussen:
Related to lightningnetwork/lnd#5779

guggero (approved):
Thanks for the fix, LGTM :tada:

Roasbeef:
We should probably start to use a sub-module for the wallet package, given all the changes to the published API lately.

Roasbeef (approved):
LGTM 🐞

PR #775, opened 2021-10-26
Fixed a bug where unconfirmed txn notifications were missing a label

ennmichael:
Fixed an issue where unconfirmed txn notifications were sometimes missing the appropriate label. The issue was that `addRelevantTx` broadcasts the notification, and then after that `PutTxLabel` updates the label, which leads to a race condition, the result being that notifications for unconfirmed transactions will sometimes not have a label, but if you call `GetTransactions` the label will in fact be there. Calling `PutTxLabel` before `addRelevantTx` fixes the bug.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3468,19 +3468,16 @@ func (w *Wallet) reliablyPublishTransaction(tx *wire.MsgTx,
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


Roasbeef:
cc @carlaKC

carlaKC (approved):
LGTM, great catch!

PR #776, opened 2021-10-27
Add witness script address support

guggero:
Replaces/continues https://github.com/btcsuite/btcwallet/pull/625.

Adds a new witness script address type with witness v0 and v1 support.
We also add a flag `isSecretScript` that can be set to false to make it possible to use a script address in a watch-only wallet.

cc @Roasbeef, @positiveblue

### DIFF
--- a/waddrmgr/address.go
+++ b/waddrmgr/address.go
@@ -508,7 +508,7 @@ type scriptAddress struct {
 	account         uint32
 	address         *btcutil.AddressScriptHash
 	scriptEncrypted []byte
-	scriptCT        []byte
+	scriptClearText []byte
 	scriptMutex     sync.Mutex
 }
 
@@ -524,7 +524,7 @@ func (a *scriptAddress) unlock(key EncryptorDecryptor) ([]byte, error) {
 	a.scriptMutex.Lock()
 	defer a.scriptMutex.Unlock()
 
-	if len(a.scriptCT) == 0 {
+	if len(a.scriptClearText) == 0 {
 		script, err := key.Decrypt(a.scriptEncrypted)
 		if err != nil {
 			str := fmt.Sprintf("failed to decrypt script for %s",
@@ -532,20 +532,20 @@ func (a *scriptAddress) unlock(key EncryptorDecryptor) ([]byte, error) {
 			return nil, managerError(ErrCrypto, str, err)
 		}
 
-		a.scriptCT = script
+		a.scriptClearText = script
 	}
 
-	scriptCopy := make([]byte, len(a.scriptCT))
-	copy(scriptCopy, a.scriptCT)
+	scriptCopy := make([]byte, len(a.scriptClearText))
+	copy(scriptCopy, a.scriptClearText)
 	return scriptCopy, nil
 }
 
-// lock zeroes the associated clear text private key.
+// lock zeroes the associated clear text script.
 func (a *scriptAddress) lock() {
 	// Zero and nil the clear text script associated with this address.
 	a.scriptMutex.Lock()
-	zero.Bytes(a.scriptCT)
-	a.scriptCT = nil
+	zero.Bytes(a.scriptClearText)
+	a.scriptClearText = nil
 	a.scriptMutex.Unlock()
 }
 
--- a/waddrmgr/address.go
+++ b/waddrmgr/address.go
@@ -47,6 +47,10 @@ const (
 	// WitnessPubKey represents a p2wkh (pay-to-witness-key-hash) address
 	// type.
 	WitnessPubKey
+
+	// WitnessScript represents a p2wsh (pay-to-witness-script-hash) address
+	// type.
+	WitnessScript
 )
 
 // ManagedAddress is an interface that provides acces to information regarding
@@ -54,7 +58,8 @@ const (
 // type may provide further fields to provide information specific to that type
 // of address.
 type ManagedAddress interface {
-	// Account returns the internal account the address is associated with.
+	// InternalAccount returns the internal account the address is
+	// associated with.
 	InternalAccount() uint32
 
 	// Address returns a btcutil.Address for the backing address.
@@ -322,7 +327,7 @@ func (a *managedAddress) ExportPrivKey() (*btcutil.WIF, error) {
 	return btcutil.NewWIF(pk, a.manager.rootManager.chainParams, a.compressed)
 }
 
-// Derivationinfo contains the information required to derive the key that
+// DerivationInfo contains the information required to derive the key that
 // backs the address via traditional methods from the HD root. For imported
 // keys, the first value will be set to false to indicate that we don't know
 // exactly how the key was derived.
@@ -502,8 +507,9 @@ func newManagedAddressFromExtKey(s *ScopedKeyManager,
 	return managedAddr, nil
 }
 
-// scriptAddress represents a pay-to-script-hash address.
-type scriptAddress struct {
+// baseScriptAddress represents the common fields of a pay-to-script-hash and
+// a pay-to-witness-script-hash address.
+type baseScriptAddress struct {
 	manager         *ScopedKeyManager
 	account         uint32
 	address         *btcutil.AddressScriptHash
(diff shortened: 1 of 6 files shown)


waddrmgr/address.go
```diff
@@ -633,9 +652,12 @@ func (a *scriptAddress) Script() ([]byte, error) {
 	return a.unlock(a.manager.rootManager.cryptoKeyScript)
 }
 
+// Enforce scriptAddress satisfies the ManagedScriptAddress interface.
+var _ ManagedScriptAddress = (*scriptAddress)(nil)
+
 // newScriptAddress initializes and returns a new pay-to-script-hash address.
 func newScriptAddress(m *ScopedKeyManager, account uint32, scriptHash,
-	scriptEncrypted []byte) (*scriptAddress, error) {
+	scriptEncrypted []byte) (ManagedScriptAddress, error) {
```
positiveblue, line 660: nit: Usually it's better to follow the rule of `accept interfaces, return structs` so the caller knows exactly what s/he is getting without the need of try/error casting.

waddrmgr/address.go
```diff
@@ -645,9 +667,131 @@ func newScriptAddress(m *ScopedKeyManager, account uint32, scriptHash,
...
+}
+
+// Enforce witnessScriptAddress satisfies the ManagedScriptAddress interface.
+var _ ManagedScriptAddress = (*witnessScriptAddress)(nil)
+
+// newWitnessScriptAddress initializes and returns a new
+// pay-to-witness-script-hash address.
+func newWitnessScriptAddress(m *ScopedKeyManager, account uint32, scriptHash,
+	scriptEncrypted []byte, witnessVersion byte,
+	isSecretScript bool) (ManagedScriptAddress, error) {
```
positiveblue, line 765: nit: same comment as before

positiveblue (approved):
LGTM but I have some unknowns so another couple of eyes are more than welcomed @Roasbeef

waddrmgr/scoped_manager.go
```diff
@@ -789,17 +790,35 @@ func (s *ScopedKeyManager) importedAddressRowToManaged(row *dbImportedAddressRow
...
 		str := "failed to decrypt imported script hash"
 		return nil, managerError(ErrCrypto, str, err)
 	}
-
 	return newScriptAddress(s, row.account, scriptHash, row.encryptedScript)
 }
 
+// scriptAddressRowToManaged returns a new managed address based on script
+// address data loaded from the database.
+func (s *ScopedKeyManager) witnessScriptAddressRowToManaged(
```
Roasbeef, line 807: `godoc` comment doesn't match

waddrmgr/scoped_manager.go
```diff
@@ -789,17 +790,35 @@ func (s *ScopedKeyManager) importedAddressRowToManaged(row *dbImportedAddressRow
...
+	row *dbWitnessScriptAddressRow) (ManagedAddress, error) {
+
+	// Use the crypto public key to decrypt the imported script hash.
+	scriptHash, err := s.rootManager.cryptoKeyPub.Decrypt(row.encryptedHash)
```
Roasbeef, line 811: Shouldn't this only preferentially attempt to decrypt it?

Roasbeef: Ah ok scratch this, I se where it happens below

waddrmgr/scoped_manager.go
```diff
@@ -2055,13 +2117,19 @@ func (s *ScopedKeyManager) ImportScript(ns walletdb.ReadWriteBucket,
...
+	}
+
 	// Encrypt the script for storage in database using the crypto script
 	// key when not a watching-only address manager.
 	var encryptedScript []byte
-	if !s.rootManager.WatchOnly() {
-		encryptedScript, err = s.rootManager.cryptoKeyScript.Encrypt(
-			script,
-		)
+	if !s.rootManager.WatchOnly() || !isSecretScript {
```
Roasbeef, line 2131: So then we don't encrypt at all if `isSecretScript` is `true`? In `scriptAddressRowToManaged`, we currently _always_ attempt to decrypt.

waddrmgr/scoped_manager.go
```diff
@@ -2107,15 +2185,25 @@ func (s *ScopedKeyManager) ImportScript(ns walletdb.ReadWriteBucket,
...
+			s, ImportedAddrAccount, scriptHash, encryptedScript,
+		)
+	}
 	if err != nil {
 		return nil, err
 	}
-	if !s.rootManager.WatchOnly() {
-		scriptAddr.scriptCT = make([]byte, len(script))
-		copy(scriptAddr.scriptCT, script)
+	if !s.rootManager.WatchOnly() || !isSecretScript {
```
Roasbeef, line 2204: Seems backwards, see comment above

Roasbeef (changes requested):
Completed an initial pass, main thin gI need to double back on is to ensure I understand the new control flow w.r.t not always using the main public crypt key for encryption

waddrmgr/db.go
```diff
@@ -155,6 +156,22 @@ type dbScriptAddressRow struct {
...
+
+	// witnessVersion is the version of the witness script.
+	witnessVersion byte
+
+	// isSecretScript denotes whether the script is considered to be "secret"
+	// and encrypted with the script encryption key or "public" and
+	// therefore only encrypted with the public encryption key.
+	isSecretScript  bool
+	encryptedHash   []byte
+	encryptedScript []byte
```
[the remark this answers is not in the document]
guggero: Good idea! I added a comment that the script isn't parsed or interpreted here and that it's up to the actual implementation on what its content is. That should be enough for us to use a TLV for a script path P2TR addr.

waddrmgr/address.go
```diff
@@ -645,9 +667,131 @@ func newScriptAddress(m *ScopedKeyManager, account uint32, scriptHash,
...
+
+	a.manager.mtx.Lock()
+	defer a.manager.mtx.Unlock()
+
+	// Account manager must be unlocked to decrypt the script.
+	if a.isSecretScript && a.manager.rootManager.IsLocked() {
+		return nil, managerError(ErrLocked, errLocked, nil)
+	}
+
+	cryptoKey := a.manager.rootManager.cryptoKeyScript
```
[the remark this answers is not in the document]
guggero: This key isn't new. It existed before for encrypting P2SH scripts. But since it is not available in watch-only wallets, we add a switch to use the "public" encryption key instead of this "secret" script encryption key.

waddrmgr/address.go
```diff
@@ -119,6 +124,10 @@ type ManagedScriptAddress interface {
 
 	// Script returns the script associated with the address.
 	Script() ([]byte, error)
+
+	// base returns the common part of a pay-to-script-hash and a
+	// pay-to-witness-script-hash address.
+	base() *baseScriptAddress
```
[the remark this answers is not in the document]
guggero: This was mainly to avoid a type switch in `importScriptAddress`, but I was able to solve it in a different way now.

waddrmgr/scoped_manager.go
guggero: Yeah, there was something backward. I fixed it. We now either require the manager to be unlocked (=cannot be watch-only) or the script to be public. And then we always encrypt, the question is just with what key.

waddrmgr/scoped_manager.go
```diff
@@ -2029,16 +2053,63 @@ func (s *ScopedKeyManager) toImportedPublicManagedAddress(
...
+	// a wallet is watch-only then the script must be encrypted with the
+	// public encryption key.
+	if isSecretScript && s.rootManager.WatchOnly() {
+		return nil, managerError(ErrWatchingOnly, errWatchingOnly, nil)
+	}
+
+	// Witness script addresses use a SHA256.
+	var scriptHash []byte
+	switch addrType {
+	case WitnessScript:
```
Roasbeef, line 2105: We'd need to modify this if we _wanted_ to support nested P2SH on this level, but I don't think we have a direct need for that, and can simplify the implementation by leaving that out.

waddrmgr/scoped_manager.go
```diff
@@ -2055,18 +2126,21 @@ func (s *ScopedKeyManager) ImportScript(ns walletdb.ReadWriteBucket,
...
-		if err != nil {
-			str := fmt.Sprintf("failed to encrypt script for %x",
-				scriptHash)
-			return nil, managerError(ErrCrypto, str, err)
-		}
+	// If a key isn't considered to be "secret", we encrypt it with the
+	// public key, so we can create script addresses that also work in
+	// watch-only mode.
+	cryptoKey := s.rootManager.cryptoKeyScript
+	if !isSecretScript {
```
Roasbeef, line 2133: 👍

Roasbeef (approved):
LGTM 🌂

PR #779, opened 2021-11-19
wallet: add GetTransaction returning data for any transaction

bjarnemagnussen:
This PR adds a method to the wallet that allows to get transaction data on any transaction given its ID. If the transaction exists in the transaction store, it will be returned with the data from the database, otherwise data from the backend is returned as a `TransactionSummary`.

This method can be useful in situations where one is interested in a specific transaction only, instead of e.g. all transactions from the `GetTransactions` method. It also adds new functionality as it can return data from the backend of a transaction not in the internal wallet.


wallet/wallet.go
```diff
@@ -2373,6 +2373,175 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+
+// GetTransaction returns data of any transaction given its id. A mined
+// transaction is returned in a Block structure which records properties about
+// the block. A bool is also returned to denote if the transaction previously
+// existed in the database or not.
+func (w *Wallet) GetTransaction(txHash *chainhash.Hash) (*GetTransactionResult,
+	bool, error) {
+
+	var res GetTransactionResult
+	chainClient := w.chainClient
```
guggero, line 2392: Need to hold the `w.chainClientLock` mutex here.

wallet/wallet.go
```diff
@@ -2373,6 +2373,175 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+		}
+
+		// The byte is reversed due to different endianness in response
+		// and call.
+		reversed := make([]byte, len(blockHashBytes))
+		for i, n := range blockHashBytes {
+			j := len(blockHashBytes) - i - 1
+			reversed[j] = n
+		}
+		blockHash, err = chainhash.NewHash(reversed)
```
guggero, line 2446: Can use `chainhash.NewHashFromStr` directly instead of reversing it yourself.

wallet/wallet.go
```diff
@@ -2373,6 +2373,175 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+			if err != nil {
+				return nil, false, err
+			}
+			height = header.Height
+		case *chain.BitcoindClient:
+			height, err = client.GetBlockHeight(blockHash)
+			if err != nil {
+				return nil, false, err
+			}
+		case *chain.NeutrinoClient:
```
guggero, line 2464: No need to include the Neutrino client here, we bailed out earlier for that chain client type.

wallet/wallet.go
```diff
@@ -2373,6 +2373,175 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+// structure for a mined transaction.
+type GetTransactionResult struct {
+	MinedTransaction   *Block
+	UnminedTransaction *TransactionSummary
+}
+
+// GetTransaction returns data of any transaction given its id. A mined
+// transaction is returned in a Block structure which records properties about
+// the block. A bool is also returned to denote if the transaction previously
+// existed in the database or not.
```
guggero, line 2387: Mention that the transaction index needs to be turned on for this to work?

lifesgoodallthetime (approved):

bjarnemagnussen:
@guggero Thanks a lot for taking your time to give me the feedback! I have added a commit with changes.

The idea with this function is to a) provide a more performant lookup of a single relevant transaction given its id, but also b) as a convenience function if the transaction index is set on the backend to allow polling for an arbitrary transaction.

Regarding the transaction index I have made a little code change to continue fetching a transaction from the database, even if the transaction index is not set on the backend. Only if the transaction cannot be found in the database _and_ the backend does not have the transaction index will the function call fail.

wallet/wallet.go
bjarnemagnussen: Thanks! I cannot believe I overlooked this method...

wallet/wallet.go
```diff
@@ -2373,6 +2373,179 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+	if err != nil {
+		return nil, false, err
+	}
+
+	var res GetTransactionResult
+
+	// In case the transaction is unconfirmed it could due to race condition
+	// confirm during the time we receive it from the backend and check it
+	// in the database.
+	// We therefore store the best block height and will use it for quering the
```
guggero, line 2403: nit: s/quering/querying (and also line length)

wallet/wallet.go
```diff
@@ -2373,6 +2373,179 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+// that does not exist in the internal wallet.
+func (w *Wallet) GetTransaction(txHash *chainhash.Hash) (*GetTransactionResult,
+	bool, error) {
+
+	chainClient, err := w.requireChainClient()
+	if err != nil {
+		return nil, false, err
+	}
+
+	var res GetTransactionResult
```
guggero, line 2398: nit: move to just above where this is actually used? So before the `switch`?

wallet/wallet.go
```diff
@@ -2373,6 +2373,175 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+
+		rangeFn := func(details []wtxmgr.TxDetails) (bool, error) {
+			// TODO: probably should make RangeTransactions not
+			// reuse the details backing array memory.
+			dets := make([]wtxmgr.TxDetails, len(details))
+			copy(dets, details)
+			details = dets
+
+			for i := range details {
+				dbTx := makeTxSummary(dbtx, w, &details[i])
```
guggero, line 2486: Hmm, could this just be:

```
for i, txDetail := range details {
        txDetail := txDetail
        dbTx := makeTxSummary(dbtx, w, &txDetail)
```

to avoid copying the whole slice?

guggero (approved):
Can you please squash the commits?
Other than a few minor nits, this looks good to me :tada:

wallet/wallet.go
bjarnemagnussen: I was a bit confused about this TODO comment. Note though, that I used the code from the `GetTransactions` method above here. But I have made your code suggestion in the squashed commit.

bjarnemagnussen:
Thanks @guggero! I have incorporated your suggestions in the squashed commit.

wallet/wallet.go
```diff
@@ -2373,6 +2373,174 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+	UnminedTransaction *TransactionSummary
+}
+
+// GetTransaction returns data of any transaction given its id. A mined
+// transaction is returned in a Block structure which records properties about
+// the block. A bool is also returned to denote if the transaction previously
+// existed in the database or not.
+// The backend must have a transaction index to poll a confirmed transaction
+// that does not exist in the internal wallet.
+func (w *Wallet) GetTransaction(txHash *chainhash.Hash) (*GetTransactionResult,
```
Roasbeef, line 2390: Unclear w.r.t the purpose of this method: it seems to rely on the RPC calls, and will return a reply even if the wallet isn't found in the wallet. 

Where would this be used that https://pkg.go.dev/github.com/btcsuite/btcwallet/wtxmgr#Store.TxDetails wouldn't be?

wallet/wallet.go
```diff
@@ -2373,6 +2373,174 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+						bestHeight = details[i].Block.Height
+						blockHash = &details[i].Block.Hash
+					}
+					return true, nil
+				}
+			}
+			return false, nil
+		}
+
+		return w.TxStore.RangeTransactions(
```
Roasbeef, line 2495: Why range over transactions vs looking up the specific-one we want? See the method I linked above

Roasbeef (changes requested):

wallet/wallet.go
bjarnemagnussen: This could be useful for applications that use `btcwallet` as dependency. E.g. in LND a remotely force closed transaction is not added to the wallet/store and therefore won't show up via `GetTransactions`. It could therefore be convenient to have a function that gets the transaction from the wallet/store if it exists and falls back to the backend otherwise.

wallet/wallet.go
bjarnemagnussen: I have changed the code to make use of `Store.TxDetails` by first checking if the transaction exists in the store, and only if not make use of the backend RPC.

guggero:
This can now rebased after https://github.com/btcsuite/btcwallet/pull/890 is in. The only remaining diff would be that if we don't have the transaction in the wallet that we would go and ask the chain backend for it.

wallet/wallet.go
```diff
@@ -2522,6 +2522,10 @@ type GetTransactionResult struct {
 func (w *Wallet) GetTransaction(txHash chainhash.Hash) (*GetTransactionResult,
 	error) {
 
+	var bestHeight int32
+	var blockHash *chainhash.Hash
+
+	var summary *TransactionSummary
```
guggero, line 2528: nit: can unify all these into a single `var ( ... )` block.

wallet/wallet.go
```diff
@@ -2554,6 +2560,86 @@ func (w *Wallet) GetTransaction(txHash chainhash.Hash) (*GetTransactionResult,
...
+
+		txRaw, err := hex.DecodeString(txResult.Hex)
+		if err != nil {
+			return nil, err
+		}
+
+		summary = &TransactionSummary{
+			Hash:        &txHash,
+			Transaction: txRaw,
+			MyInputs:    make([]TransactionSummaryInput, 0),
```
guggero, line 2595: Can just leave those nil instead of creating an empty slice IMO.

wallet/wallet.go
```diff
@@ -2554,6 +2560,86 @@ func (w *Wallet) GetTransaction(txHash chainhash.Hash) (*GetTransactionResult,
...
+
+		// Get the transaction data from the backend endpoint.
+		var txResult *btcjson.TxRawResult
+		switch client := chainClient.(type) {
+		case *chain.RPCClient:
+			txResult, err = client.GetRawTransactionVerbose(&txHash)
+			if err != nil {
+				return nil, err
+			}
+		case *chain.BitcoindClient:
```
guggero, line 2580: What about Neutrino? Those clients would currently panic on the `txResult.Hex` access below.

wallet/wallet.go
```diff
@@ -2554,6 +2560,86 @@ func (w *Wallet) GetTransaction(txHash chainhash.Hash) (*GetTransactionResult,
...
+
+		// Set the TransactionSummary in the GetTransactionResult.
+		res.Summary = *summary
+		res.Height = bestHeight
+		res.BlockHash = summary.Hash
+		res.Timestamp = summary.Timestamp
+
+		// Calculate the number of confirmations.
+		if res.Height != -1 {
+			res.Confirmations = bestHeight - res.Height + 1
```
guggero, line 2638: I don't think this is correct. `bestHeight` and `res.Height` are both the exact same value at this point.

guggero (commented):
Thanks for the update. Main comment is about making sure this doesn't panic for Neutrino clients. Not sure we can give them a useful response though, as it would require re-scanning the chain for the transaction. So not sure what to do about that. What would this extended call be used for in your use case in the first place?

wallet/wallet.go
mrfelton: Updated in updated commit

wallet/wallet.go
mrfelton: updated in latest commit

mrfelton:
> Thanks for the update. Main comment is about making sure this doesn't panic for Neutrino clients. Not sure we can give them a useful response though, as it would require re-scanning the chain for the transaction. So not sure what to do about that. What would this extended call be used for in your use case in the first place?

The use case is as @bjarnemagnussen mentioned in https://github.com/btcsuite/btcwallet/pull/779#discussion_r760966298.

wallet/wallet.go
guggero: See line 2639.

mrfelton: lol. I've updated this to fetch the chain tip and then compare the heights to get the number of confirmation.

Roasbeef (changes requested):
>  but more specifically, to enhance LND's GetTransaction RPC to be able to fetch transactions that are not known to LND

Why not just use `getrawtransaction` then? `GetTransaction` (an API method of the wallet) is intended only to fetch transactions that are relevant to the wallet. 

This PR takes that API method and adds in additional network calls (and for neutrino new p2p messages being sent), which significantly expands the scope of the method.

mrfelton:
That's certainly another valid approach, and is actually more akin to what we are doing now (we are no longer activly using this enhancement).

> GetTransaction (an API method of the wallet) is intended only to fetch transactions that are relevant to the wallet.

IIRC, the transactions we were interested in were at one point in some way relevant to the wallet (e.g. a remote force close transaction, or transactions related to RBF). Part of the rationale for making this accessible through LND rather than calling through to Bitcoind directly was for simplicity. LND generally provides everything needed to operate on top of it, and there are multiple instances where LND proxies through to Bitcoind to make something from the underlaying available. We did not want to open up direct connections to our Bitcoind instances and complicate things with needing to manage the multiple connections.

Issue #780, opened 2021-11-23
btcwallet 0.13 wtxmgr go.mod require needs update


chappjc:
https://github.com/btcsuite/btcwallet/blob/559f6eff0e52f7a617382eec0d0a41c835c0709f/go.mod#L12

The tagged btcwallet v0.13.0 currently specifies wtxmgr v1.3.0 in the go.mod, however a newer revision of wtxmgr that has the `InsertTxCheckIfExists` method is actually required.  A consumer must add a replace to their project's go.mod if they wish to use btcwallet v0.13.0.  e.g.

`github.com/btcsuite/btcwallet/wtxmgr v1.3.1-0.20211113022102-559f6eff0e52`

Consumers can work around this as above, but perhaps a tag and updated `require` is something for the next patch release.  I find that when tagging releases, that removing `replace` statements is helpful to smoke out such things, however minor they are.

Alternatively, btcwallet might consider removing `wtxmgr` as a standalone module, in favor of absorbing it into the main module.  I'm not familiar with use cases for wtxmgr that might be sensible without the root module in use as well.

chappjc:
Specifically, I guess I'm requesting:
- a `wtxmgr/v1.4.0` tag
- updated go.mod so the main btcwallet module uses that
- a `v0.13.1` tag for the main btcwallet module, preferably at a revision with `replace` statements removed to confirm the `requires` are sufficient

I'm eager to update to v0.13, but it is somewhat awkward to need a `replace` or `require` to hoist the wtxmgr module to a newer _untagged_ revision.

Thanks!  The [0.13 change list](https://github.com/btcsuite/btcwallet/compare/v0.12.0...v0.13.0) looks really nice.  In particular I'm curious about if and how https://github.com/btcsuite/btcwallet/commit/34da5cfae82320e569f696323015bbdba0d202c0 affects peering for light clients.

PR #782, opened 2021-12-03
update wtxmgr require for a replace-free release

chappjc:
Resolves https://github.com/btcsuite/btcwallet/issues/780

This is essentially what I would propose to resolve issue https://github.com/btcsuite/btcwallet/issues/780.
Except, there would ideally be a `wtxmgr/v1.4.0` tag  instead of a require to for an untagged revision.

The btcd required update is not strictly necessary, and should actually probably be skipped.
On the other hand, the neutrino update to v0.13.0 is more important as there is a critical bug fixed in that release.


chappjc:
Closing since other module tasks related to the new btcd/btcutil submodule that will obsolete these changes.

I would just respectfully request that either (a) the next tagged release could be cut without the replaces so that Go consumers (not just the btcwallet app when built from the root of the repository) can `require` the btcwallet module without additional `replace` or `require` directives to get compatible wtxmgr etc. submodules, or (b) ensure that any replaces aren't needed to build.

Plus it's nice for install instructions if you could do `go install` to get the btcwallet app without manually checking out the workspace and building from the main module folder (repo root).

```
$  go install -v github.com/btcsuite/btcwallet@v0.13.0
go install: github.com/btcsuite/btcwallet@v0.13.0 (in github.com/btcsuite/btcwallet@v0.13.0):
	The go.mod file for the module providing named packages contains one or
	more replace directives. It must not contain directives that would cause
	it to be interpreted differently than if it were the main module.
```

Also would be good to bump the versions in version.go.

PR #786, opened 2022-01-23
chain: add rpc polling events option

ellemouton:
height based polling


ellemouton:
sorry, meant to open against own repo as an example

Issue #787, opened 2022-01-25
Migrate to btcd/btcutils submodule


3nprob:
As per https://github.com/btcsuite/btcd/pull/1785, `btcutil` needs to migrate as well in order to afford upgrading the `btcd` dependency. Attempting to use current `btcd@master`: 

```
$ GO111MODULE=on go install -v . ./cmd/...
github.com/btcsuite/btcwallet/internal/legacy/keystore
github.com/btcsuite/btcwallet/wallet/txrules
github.com/lightninglabs/neutrino
# github.com/btcsuite/btcwallet/wallet/txrules
wallet/txrules/rules.go:30:23: cannot use relayFeePerKb (type "github.com/btcsuite/btcutil".Amount) as type "github.com/btcsuite/btcd/btcutil".Amount in argument to mempool.IsDust
# github.com/btcsuite/btcwallet/internal/legacy/keystore
internal/legacy/keystore/keystore.go:2795:3: cannot use addresses (type []"github.com/btcsuite/btcd/btcutil".Address) as type []"github.com/btcsuite/btcutil".Address in field value
internal/legacy/keystore/keystore.go:2877:15: cannot use addresses (type []"github.com/btcsuite/btcd/btcutil".Address) as type []"github.com/btcsuite/btcutil".Address in assignment
# github.com/lightninglabs/neutrino
../../../../pkg/mod/github.com/lightninglabs/neutrino@v0.12.1/blockmanager.go:2718:35: cannot use stubBlock (type *"github.com/btcsuite/btcutil".Block) as type *"github.com/btcsuite/btcd/btcutil".Block in argument to blockchain.CheckProofOfWork
../../../../pkg/mod/github.com/lightninglabs/neutrino@v0.12.1/query.go:988:42: cannot use block (type *"github.com/btcsuite/btcutil".Block) as type *"github.com/btcsuite/btcd/btcutil".Block in argument to blockchain.CheckBlockSanity
../../../../pkg/mod/github.com/lightninglabs/neutrino@v0.12.1/rescan.go:931:53: cannot use relevantTxs (type []*"github.com/btcsuite/btcutil".Tx) as type []*"github.com/btcsuite/btcd/btcutil".Tx in argument to ro.ntfn.OnFilteredBlockConnected
[...]

PR #788, opened 2022-01-25
Update btcd

3nprob:
Moving btcutil required as per #787.

Keeping in draft hoping that https://github.com/lightninglabs/neutrino/pull/242 will be merged and forking it will not be necessary.


go.mod
```diff
@@ -1,27 +1,29 @@
...
+	github.com/btcsuite/btcd/btcutil v1.0.1-0.20220111222628-c5e6d3cf5f56
+	github.com/btcsuite/btcd/btcutil/psbt v1.0.1-0.20220111222628-c5e6d3cf5f56
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
-	github.com/btcsuite/btcutil v1.0.3-0.20210929233259-9cdf59f60c51
-	github.com/btcsuite/btcutil/psbt v1.0.3-0.20210929233259-9cdf59f60c51
 	github.com/btcsuite/btcwallet/wallet/txauthor v1.1.0
 	github.com/btcsuite/btcwallet/wallet/txrules v1.1.0
 	github.com/btcsuite/btcwallet/wallet/txsizes v1.1.0
 	github.com/btcsuite/btcwallet/walletdb v1.3.5
 	github.com/btcsuite/btcwallet/wtxmgr v1.3.0
```
chappjc, line 12: For a Go consumer of the btcwallet module to use btcwallet as an import where the replaces here aren't applied, this version of wtxmgr is too old.  I had to update it to a pseudo version like `v1.3.1-0.20211201210108-79de92f527dc` in https://github.com/btcsuite/btcwallet/pull/782, which your PR would hopefully replace.  Ideally the maintainers would tag this a wtxmgr 1.4.0 (or whatever), but otherwise a pseudo-version is needed either here in the go.mod of the consumer.

But, the btcwallet **submodules that also require btcutil** and btcd are:
- wallet/txauthor
- wtxmgr
- wallet/txrules

I think I would update these submodules requires first, then come back to this main btcwallet go.mod to have it point to new pseudo-versions of these updated submodules.

wallet/txauthor/author_test.go
```diff
@@ -8,7 +8,7 @@ import (
 	"testing"
 
 	"github.com/btcsuite/btcd/wire"
-	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcd/btcutil"
```
chappjc, line 11: `wallet/txauthor` has it's own go.mod to update this as well.  Will be needed after the neutrino update in https://github.com/lightninglabs/neutrino/pull/242 because neutrino does not require the main btcwallet module, just `wallet/txauthor` and `wxtmgr` (`walletdb`, too but it does not require btcutil)

wallet/txrules/rules.go
```diff
@@ -12,7 +12,7 @@ import (
 	"github.com/btcsuite/btcd/mempool"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
-	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcd/btcutil"
```
chappjc, line 15: `wallet/txrules` has it's own go.mod to update this as well.

wtxmgr/tx.go
```diff
@@ -16,7 +16,7 @@ import (
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
-	"github.com/btcsuite/btcutil"
+	"github.com/btcsuite/btcd/btcutil"
```
chappjc, line 19: `wtxmgr` has it's own go.mod to update this as well.

go.mod
3nprob: Feedback appreciated; I'm currently a bit stuck ironing out all these in a way that makes everything align and builds when imported in lnd. Hopefully I'll make some progress.

wtxmgr/tx.go
3nprob: Should be addressed now

go.mod
3nprob: I *think* I've addressed all your points now. All checks seem to work out as well. Unfortunately I have not been able to figure out a way to get everything to line up without also simultaneously using an updated version of neutrino (and same thing there, making this a bit cyclical).

Would be really good if we can make some further progress here so any assistance to get this mergable is very much appreciated.

3nprob:
Can we get this CI'd?

go.mod
chappjc: Just to clarify that I'm not a maintainer here, just another fellow consumer with an interest in seeing the updates.

In general I think the way forward needs to involve a little back and forth between the repos.  With the wallet submodules (wallet/txauthor, wtxmgr, wallet/txrules) updated for the new btcutil, that would let neutrino update it's requires on those wallet submodules (in https://github.com/lightninglabs/neutrino/pull/242, but without replaces), then come back to btcwallet and update it's neutrino require.

These gymnastics are happening because btcd is still v0.x, not yet following semantic versioning.  **The btcutil move is a breaking change to consumers with btcutil types in their exported APIs**, and normally in Go land that would require a change to the semantic import versioning (SIV) in such consumers, e.g. btcd/v2 so that btcd consumers are not borken.

Unfortunately, the same breaking change I've described in btcd is about to happen to these three wallet submodules that are v1.x.  So, it all depends on how the maintainers want to handle this (allow breaking changes thus violating SIV or bump these wallet submodules to /v2 and update any requires on them).

3nprob: > Just to clarify that I'm not a maintainer here, just another fellow consumer with an interest in seeing the updates.

That was clear - the request for assistance was more a shout in the void than directly addressing you personally, though of course good to see you're still interested in this as well :)

I think I finally got what you were already trying to say above. Taking a new shot for the first step of that lock-step change. Really hoping maintainers will pop in here soon as quite a lot of things are blocked by this right now.

Roasbeef:
@3nprob I can take this over and handle pushing out all the updated sub-module tags if you'd like?

Roasbeef:
> I think I would update these submodules go.mod requires first, then come back to this main btcwallet go.mod to have it point to new pseudo-versions of these updated submodules.

Yes this is what we usually do: update the sub-modules, push a new tag, _then_ update the top-level `go.mod`.

Roasbeef:
> In general I think the way forward needs to involve a little back and forth between the repos. With the wallet submodules (wallet/txauthor, wtxmgr, wallet/txrules) updated for the new btcutil, that would let neutrino update it's requires on those wallet submodules (in https://github.com/lightninglabs/neutrino/pull/242, but without replaces), then come back to btcwallet and update it's neutrino require.

Yep this is usually how we do things. Admittedly, as we maintain btcwallet+neutrino+lnd, etc -- we at times cut a few corners when it comes to modules/tagging. Though happy to clean things up, so things are less annoying for other dependent users like y'all.

Roasbeef:
> I can take this over and handle pushing out all the updated sub-module tags if you'd like?

Started to port everything over and confirmed that we _are_ indeed in module dep hell rn....

This should remedy things though: https://github.com/btcsuite/btcd/pull/1805

After that, we need a smol `btcutil` tag, then we _should_ be able to update everything (hit this snag when I tried locally).

Roasbeef:
This PR updates _just_ the sub-modules: https://github.com/btcsuite/btcwallet/pull/789

Roasbeef:
This has been done in existing PRs, closing, thanks for starting this!

PR #789, opened 2022-02-07
multi: update sub-modules to use new btcsuite/btcd/btcutil repo

Roasbeef:
In this commit, we update our sub-modules (but _not_ the main `btcwallet` module) to use the new `btcutil` import path and APIs. We also stop using top level repalce directives in the main `go.mod` file, as this enables us to update the sub-modules _in tree_, but still use the older tagged version. Removing these directives forced us to correct a module issue, where we were pinned to an older version of the `wtxmgr` package, but were using APIs of the newer version. 

Once this lands, we can update the `neutrino` module, tag that, then update the top-level module.


Roasbeef:
The `TestPrunedBlockDispatcherMultipleQueryPeers` test seems to be kinda flaky. It failed for me, then travis, but then locally on master it passes more consistently (but is still somewhat flaky).

Roasbeef:
cc @3nprob @chappjc

go.mod
```diff
@@ -20,21 +20,15 @@ require (
...
 replace github.com/btcsuite/btcwallet/walletdb => ./walletdb
 
-replace github.com/btcsuite/btcwallet/wtxmgr => ./wtxmgr
-
-replace github.com/btcsuite/btcwallet/wallet/txauthor => ./wallet/txauthor
-
-replace github.com/btcsuite/btcwallet/wallet/txrules => ./wallet/txrules
-
-replace github.com/btcsuite/btcwallet/wallet/txsizes => ./wallet/txsizes
+replace github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.22.0-beta.0.20210803133449-f5a1fb9965e4
```
Roasbeef, line 31: I need to add this as otherwise one of the modules would bump into those bogus tags we recently removed. I should be able to remove this one I update the top-level modules in the next PR.

chappjc: I guess that next PR would also be the btcec/v2 stuff?  Trying to update our app's deps and it looks like the next thing it complains about is

```
    github.com/btcsuite/btcwallet/waddrmgr imports
    github.com/btcsuite/btcd/btcec: module github.com/btcsuite/btcd@latest found (v0.18.1), but does not contain package github.com/btcsuite/btcd/btcec
```

I thought I'd be able to have my app require btcd `master`, but that would seem to be impossible since on master it had a btcec/v2 but no btcec(/v1)

3nprob:
Thanks for picking up the effort @Roasbeef, good to start seeing some light in the dph tunnel.

This already looks like steps in the right direction. What's your stance on https://github.com/btcsuite/btcd/pull/1802 ? I think that will help dependents all over for a smoother experience.

PR #792, opened 2022-02-14
Taproot: Add taproot address and signing capabilities

guggero:
~Depends on https://github.com/lightningnetwork/lnd/pull/6285 (for the TLV module).~

Adds taproot key spend and script spend addresses.

Merge order for this saga of PRs is:
 - https://github.com/btcsuite/btcd/pull/1787
 - https://github.com/lightningnetwork/lnd/pull/6285
 - This PR
 - https://github.com/lightningnetwork/lnd/pull/6263

### DIFF
--- a/wallet/import.go
+++ b/wallet/import.go
@@ -404,6 +404,59 @@ func (w *Wallet) ImportPublicKey(pubKey *btcec.PublicKey,
 	return nil
 }
 
+// ImportTaprootScript imports a user-provided taproot script into the address
+// manager. The imported script will act as a pay-to-taproot address.
+func (w *Wallet) ImportTaprootScript(scope waddrmgr.KeyScope,
+	tapscript *waddrmgr.Tapscript, bs *waddrmgr.BlockStamp,
+	witnessVersion byte, isSecretScript bool) (waddrmgr.ManagedAddress,
+	error) {
+
+	manager, err := w.Manager.FetchScopedKeyManager(scope)
+	if err != nil {
+		return nil, err
+	}
+
+	// The starting block for the key is the genesis block unless otherwise
+	// specified.
+	if bs == nil {
+		bs = &waddrmgr.BlockStamp{
+			Hash:      *w.chainParams.GenesisHash,
+			Height:    0,
+			Timestamp: w.chainParams.GenesisBlock.Header.Timestamp,
+		}
+	} else if bs.Timestamp.IsZero() {
+		// Only update the new birthday time from default value if we
+		// actually have timestamp info in the header.
+		header, err := w.chainClient.GetBlockHeader(&bs.Hash)
+		if err == nil {
+			bs.Timestamp = header.Timestamp
+		}
+	}
+
+	// TODO: Perform rescan if requested.
+	var addr waddrmgr.ManagedAddress
+	err = walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+		ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+		addr, err = manager.ImportTaprootScript(
+			ns, tapscript, bs, witnessVersion, isSecretScript,
+		)
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	log.Infof("Imported address %v", addr.Address())
+
+	err = w.chainClient.NotifyReceived([]btcutil.Address{addr.Address()})
+	if err != nil {
+		return nil, fmt.Errorf("unable to subscribe for address "+
+			"notifications: %v", err)
+	}
+
+	return addr, nil
+}
+
 // ImportPrivateKey imports a private key to the wallet and writes the new
 // wallet to disk.
 //


Roasbeef:
Dependent PR has been merged.

wallet/txauthor/go.mod
```diff
@@ -4,7 +4,18 @@ go 1.12
...
+
+replace github.com/btcsuite/btcd => github.com/guggero/btcd v0.20.1-beta.0.20220214155024-51fd177d32c0
+
+replace github.com/btcsuite/btcd/btcec/v2 => github.com/guggero/btcd/btcec/v2 v2.0.0-20220214155024-51fd177d32c0
+
+// The old version of ginko that's used in btcd imports an ancient version of
+// gopkg.in/fsnotify.v1 that isn't go mod compatible. We fix that import error
+// by replacing ginko (which is only a test library anyway) with a more recent
+// version.
+replace github.com/onsi/ginkgo => github.com/onsi/ginkgo v1.14.2
```
Roasbeef, line 21: Is this a permanent thing?

go.mod
```diff
@@ -30,6 +30,12 @@ require (
 
 replace github.com/btcsuite/btcwallet/walletdb => ./walletdb
 
+replace github.com/btcsuite/btcd => github.com/guggero/btcd v0.20.1-beta.0.20220214155024-51fd177d32c0
+
+replace github.com/btcsuite/btcd/btcec/v2 => github.com/guggero/btcd/btcec/v2 v2.0.0-20220214155024-51fd177d32c0
+
+replace github.com/btcsuite/btcwallet/wallet/txauthor => ./wallet/txauthor
```
Roasbeef, line 37: FWIW this was removed in a prior PR, but I can see how it's helpful to fully integrate things in one swoop. Otherwise, we can make a new PR that adds the new `txscript` stuff, then does a tag, then to have it integrated here ultimately.

wallet/createtx.go
```diff
@@ -416,11 +416,22 @@ func (w *Wallet) addrMgrWithChangeSource(dbtx walletdb.ReadWriteTx,
 // validateMsgTx verifies transaction input scripts for tx.  All previous output
 // scripts from outputs redeemed by the transaction, in the same order they are
 // spent, must be passed in the prevScripts slice.
-func validateMsgTx(tx *wire.MsgTx, prevScripts [][]byte, inputValues []btcutil.Amount) error {
-	hashCache := txscript.NewTxSigHashes(tx)
+func validateMsgTx(tx *wire.MsgTx, prevScripts [][]byte,
+	inputValues []btcutil.Amount) error {
```
Roasbeef, line 420: Nice that we already have this stuff available here, we somewhat worried how much extra plumbing we'd need to add everywhere to obtain all the additional prev out values.

wallet/psbt.go
```diff
@@ -390,6 +390,39 @@ func (w *Wallet) FinalizePsbt(keyScope *waddrmgr.KeyScope, account uint32,
 	return nil
 }
 
+// PsbtPrevOutputFetcher returns a txscript.PrevOutFetcher built from the UTXO
+// information in a PSBT packet.
+func PsbtPrevOutputFetcher(packet *psbt.Packet) *txscript.MultiPrevOutFetcher {
```
Roasbeef, line 395: 👍 

I think we can keep the PSBT scope minimal here so we'll push off the interpretation of all the new fields to another PR.

waddrmgr/address.go
```diff
@@ -269,6 +273,9 @@ func (a *managedAddress) PubKey() *btcec.PublicKey {
 // pubKeyBytes returns the serialized public key bytes for the managed address
 // based on whether or not the managed address is marked as compressed.
 func (a *managedAddress) pubKeyBytes() []byte {
+	if a.addrType == TaprootPubKey {
```
Roasbeef, line 276: 👍

waddrmgr/manager_test.go
```diff
@@ -2957,3 +2957,127 @@ func TestDeriveFromKeyPathCache(t *testing.T) {
...
+		Account:         hdkeychain.HardenedKeyStart,
+		Branch:          1,
+		Index:           0,
+	}
+
+	assertAddressDerivation(
+		t, db, func(ns walletdb.ReadWriteBucket) (ManagedAddress, error) {
+			return scopedMgr.DeriveFromKeyPath(ns, externalPath)
+		},
+		"bc1p07pv8kuk9jyqr982cveq276w0rffpcva3plpresq4jzxd9nxdczszqtyre",
```
Roasbeef, line 3024: These are from BIP 86?

wallet/txauthor/author.go
```diff
@@ -309,6 +323,54 @@ func spendWitnessKeyHash(txIn *wire.TxIn, pkScript []byte,
 	return nil
 }
 
+// spendTaprootKeyHash generates, and sets a valid witness for spending the
```
Roasbeef, line 326: _technically_ there's no key hash here, as the main pubkey is used in the output script, and the witness is _just_ the signature (and not also the public key)

wallet/txauthor/author.go
```diff
@@ -309,6 +323,54 @@ func spendWitnessKeyHash(txIn *wire.TxIn, pkScript []byte,
...
+// passed pkScript with the specified input amount. The input amount *must*
+// correspond to the output value of the previous pkScript, or else verification
+// will fail since the new sighash digest algorithm defined in BIP0341 includes
+// the input value in the sighash.
+func spendTaprootKeyHash(txIn *wire.TxIn, pkScript []byte,
+	inputValue int64, chainParams *chaincfg.Params, secrets SecretsSource,
+	tx *wire.MsgTx, hashCache *txscript.TxSigHashes, idx int) error {
+
+	// First obtain the key pair associated with this p2wkh address.
+	_, addrs, _, err := txscript.ExtractPkScriptAddrs(pkScript, chainParams)
```
Roasbeef, line 336: Need to double check that this function was updated to extra the public key. However in this case, we can just use the public key as it's encoded in the pkScript.

wallet/txauthor/author.go
```diff
@@ -309,6 +323,54 @@ func spendWitnessKeyHash(txIn *wire.TxIn, pkScript []byte,
...
+// spendTaprootKeyHash generates, and sets a valid witness for spending the
+// passed pkScript with the specified input amount. The input amount *must*
+// correspond to the output value of the previous pkScript, or else verification
+// will fail since the new sighash digest algorithm defined in BIP0341 includes
+// the input value in the sighash.
+func spendTaprootKeyHash(txIn *wire.TxIn, pkScript []byte,
+	inputValue int64, chainParams *chaincfg.Params, secrets SecretsSource,
+	tx *wire.MsgTx, hashCache *txscript.TxSigHashes, idx int) error {
+
+	// First obtain the key pair associated with this p2wkh address.
```
Roasbeef, line 335: `p2wkh` comment is dated

wallet/txauthor/author.go
```diff
@@ -309,6 +323,54 @@ func spendWitnessKeyHash(txIn *wire.TxIn, pkScript []byte,
...
+	// With the concrete address type, we can now generate the
+	// corresponding witness program to be used to generate a valid witness
+	// which will allow us to spend this output.
+	witnessProgram, err := txscript.PayToAddrScript(p2trAddr)
+	if err != nil {
+		return err
+	}
+	witnessScript, err := txscript.TaprootWitnessSignature(
+		tx, hashCache, idx, inputValue, witnessProgram,
+		txscript.SigHashDefault, privKey,
```
Roasbeef, line 363: Was thinking about this when implementing, but is there any reason for us to _also_ generate 64 byte signatures when the sighash is all?

go.mod
```diff
@@ -30,6 +30,12 @@ require (
 
 replace github.com/btcsuite/btcwallet/walletdb => ./walletdb
 
+replace github.com/btcsuite/btcd => github.com/guggero/btcd v0.20.1-beta.0.20220214155024-51fd177d32c0
```
3nprob, line 33: What's behind the fork of `btcd` and `btcec/v2`?  Do you have link to diff vs upstream?

wallet/txauthor/go.mod
```diff
@@ -4,7 +4,18 @@ go 1.12
...
 	github.com/btcsuite/btcd v0.22.0-beta.0.20220204213055-eaf0459ff879
+	github.com/btcsuite/btcd/btcec/v2 v2.1.0
 	github.com/btcsuite/btcd/btcutil v1.1.0
 	github.com/btcsuite/btcwallet/wallet/txrules v1.2.0
 	github.com/btcsuite/btcwallet/wallet/txsizes v1.1.0
 )
+
+replace github.com/btcsuite/btcd => github.com/guggero/btcd v0.20.1-beta.0.20220214155024-51fd177d32c0
+
+replace github.com/btcsuite/btcd/btcec/v2 => github.com/guggero/btcd/btcec/v2 v2.0.0-20220214155024-51fd177d32c0
```
3nprob, line 15: Same question as above

go.mod
guggero: Just a few fixes I found while end-to-end testing https://github.com/btcsuite/btcd/pull/1787. Might not be necessary after the latest update on that PR.

wallet/txauthor/go.mod
guggero: Yeah, this is because of the ancient version of `goleveldb` that is used. Would be fixed by https://github.com/btcsuite/btcd/pull/1780 I think.

waddrmgr/manager_test.go
guggero: Yes. Added a comment at the top.

go.mod
guggero: See https://github.com/guggero/btcd/tree/schnorr-sig, this is the current diff (which will hopefully be integrated into btcd#1787)

wallet/txauthor/author.go
guggero: Hmm, I'd say yes. Since this is an explicit call to `TaprootWitnessSignature`. Or what do you mean exactly?

waddrmgr/address.go
```diff
@@ -785,3 +806,34 @@ func newWitnessScriptAddress(m *ScopedKeyManager, account uint32, scriptHash,
 		isSecretScript: isSecretScript,
 	}, nil
 }
+
+// taprootScriptAddress represents a pay-to-taproot address that commits to a
+// script.
+type taprootScriptAddress struct {
```
3nprob, line 812: Can this be removed?

go.mod
3nprob: How about opening a new PR for that (targeting https://github.com/Roasbeef/btcd/tree/taproot-impl until it is merged)? the branch/PR is already quite large, but it would be great to be able to review that separately already IMO :)

waddrmgr/address.go
guggero: This PR isn't ready for review, I'm actively working on stuff.

wallet/txauthor/author.go
Roasbeef: Basically to optimize the user's sig size for them behind the scenes. I updated the `RawTxInTaprootSignature` to _always_ append the sig type if not default in the latest version of the PR.

go.mod
Roasbeef: I think his fork should now be more or less identical to my main PR.

guggero: All my changes were either incorporated into https://github.com/btcsuite/btcd/pull/1787 or moved to another repo. So there is no need for another branch/PR anymore.

wallet/txauthor/author.go
```diff
@@ -309,6 +323,54 @@ func spendWitnessKeyHash(txIn *wire.TxIn, pkScript []byte,
...
+
+	tapKey := txscript.ComputeTaprootKeyNoScript(pubKey)
+	p2trAddr, err := btcutil.NewAddressTaproot(
+		schnorr.SerializePubKey(tapKey), chainParams,
+	)
+	if err != nil {
+		return err
+	}
+
+	// With the concrete address type, we can now generate the
```
Roasbeef, line 354: FWIW, we could save a few lines here just by using the passed `pkScript` here below when calling `TaprootWitnessSignature`.

waddrmgr/db.go
```diff
@@ -1622,6 +1623,11 @@ func fetchAddressByHash(ns walletdb.ReadBucket, scope *KeyScope,
 		return deserializeScriptAddress(row)
 	case adtWitnessScript:
 		return deserializeWitnessScriptAddress(row)
+	case adtTaprootScript:
```
Roasbeef, line 1626: Maybe I'm missing something, but iirc we already did the base keyspend address integration here?

waddrmgr/scoped_manager.go
```diff
@@ -2084,15 +2086,17 @@ func (s *ScopedKeyManager) ImportWitnessScript(ns walletdb.ReadWriteBucket,
...
+		ns, digest[:], script, bs, WitnessScript, witnessVersion,
+		isSecretScript,
 	)
 }
 
 // importScriptAddress imports a new pay-to-script or pay-to-witness-script
 // address.
 func (s *ScopedKeyManager) importScriptAddress(ns walletdb.ReadWriteBucket,
-	script []byte, bs *BlockStamp, addrType AddressType,
+	scriptIdent []byte, script []byte, bs *BlockStamp, addrType AddressType,
```
Roasbeef, line 2130: Alternatively, this could be a closure type that takes the script and returns an identifier. Then we can have pre-made versions for p2sh, and p2wsh, etc.

waddrmgr/address.go
```diff
@@ -534,6 +534,14 @@ func newManagedAddressFromExtKey(s *ScopedKeyManager,
 	return managedAddr, nil
 }
 
+// clearTextSetter is a non-exported interface to identify script types that
+// allow their clear text script to be set.
+type clearTextSetter interface {
```
Roasbeef, line 554: Clear text here is just a top-level identifier that isn't the hashed script?

waddrmgr/address.go
```diff
@@ -55,6 +55,10 @@ const (
 
 	// TaprootPubKey represents a p2tr (pay-to-taproot) address type.
 	TaprootPubKey
+
+	// TaprootScript represents a p2tr (pay-to-taproot) address type that
+	// commits to a script and not just a single key.
+	TaprootScript
```
Roasbeef, line 61: Update godoc above to mention that it assumes usage of BIP 86? (for `TaprootPubkey`).

Roasbeef: I wonder if we _really_ need to distinguish the two diff pubkey types here though...

Still getting through the PR, but this seems to assume that any imported taproot addrs that have items in the script tree will never spend the keyspend path?

waddrmgr/tapscript.go
```diff
@@ -0,0 +1,80 @@
...
+
+const (
+	// TapscriptTypeFullTree is the type of tapscript that knows its full
+	// tree with all individual leaves present.
+	TapscriptTypeFullTree TapscriptType = 0
+
+	// TapscriptTypePartialReveal is the type of tapscript that only knows
+	// a single revealed leaf and the merkle/inclusion proof for the rest of
+	// the tree.
+	TapscriptTypePartialReveal TapscriptType = 1
```
Roasbeef, line 22: Very cool 😎

waddrmgr/tapscript.go
```diff
@@ -0,0 +1,80 @@
...
+		return txscript.ComputeTaprootOutputKey(
+			t.ControlBlock.InternalKey, rootHash[:],
+		), nil
+
+	case TapscriptTypePartialReveal:
+		if len(t.RevealedScript) == 0 {
+			return nil, fmt.Errorf("revealed script missing")
+		}
+
+		rootHash := t.ControlBlock.RootHash(t.RevealedScript)
```
Roasbeef, line 72: `RootHash` usage here is now clear! My PR should have this method added now.

waddrmgr/tlv.go
```diff
@@ -0,0 +1,271 @@
...
+	typeTapscriptType           tlv.Type = 1
+	typeTapscriptControlBlock   tlv.Type = 2
+	typeTapscriptLeaves         tlv.Type = 3
+	typeTapscriptRevealedScript tlv.Type = 4
+
+	typeTapLeafVersion tlv.Type = 1
+	typeTapLeafScript  tlv.Type = 2
+)
+
+// tlvEncodeTaprootScript encodes the given internal key and full set of taproot
```
Roasbeef, line 22: Happy to see everything nicely fit into place! 

TLV FTW 😎

Roasbeef (changes requested):
Excellent PR! 

Completed an initial review pass through the latest set of commits. I like how we don't bind the caller/user into a particular way and allow them to either use a single script path or just import the entire tree for tracking/spending purposes. This'll really be useful once we start to implement more advanced multi-sig and other script operations using the wallet. 

Ultimately, we may want to start (from an `lnd` perspective) _also_ importing our scripts into the wallet, but possibly with a non-default account type. In any case, that's beyond the scope of this PR. 

Only lingering design question IMO is if we really need to distinguish between pure-keyspend and a mixed variant for addresses. I guess as is, we assume usage of BIP 86, so anything that doesn't use that should use this new import API instead.

Roasbeef:
@carlaKC

waddrmgr/db.go
guggero: Not 100% sure what you mean. We do add the `TaprootPubKey` address type in this PR. But because that is just a different representation of a private/public key pair, there doesn't need to be any specific database code for it. Same as there is no specific p2wkh serialization/de-serialization code here, only p2wsh.

waddrmgr/scoped_manager.go
guggero: Not sure if this was what you had in mind, but I added something to that effect with `Identity`. Can't really be more generic since the input isn't the same in all cases (raw script bytes vs. `*TapScript` struct or pub key).

waddrmgr/address.go
guggero: It's just an internal optimization (to avoid decrypting the script multiple times) that was previously done through a direct struct type assert and is now done through a private interface. Updated the name to make it a bit more clear what exactly is being set.

waddrmgr/address.go
guggero: Updated the godoc.

I think we should distinguish between the two, yes. The first (`TaprootPubKey`) is just what the user will use on the wallet level for receiving/sending. The second (`TaprootScript`) is what we'll use internally for things we sign over the RPC. In theory we don't really need to store those scripts in the wallet DB (we've been using witness scripts for a while now and haven't stored any of them in the wallet DB, we just derive them when needed).

So basically:
 - `TaprootPubKey`: the wallet knows the private key of the internal key and we use this through `ComputeInputScript` mainly
 - `TaprootScript`: the wallet doesn't know the private key of the internal key (we use `DeriveKey`/`DeriveNextKey`) and we use this through `SignOutputRaw` mainly. We only use this to store a script and detect incoming funds.

But maybe I'm missing a use case that we're planning to support?

waddrmgr/db.go
Roasbeef: I was confused, I think you've clarified things lower re the two diff taproot addr types.

waddrmgr/address.go
Roasbeef: > But maybe I'm missing a use case that we're planning to support?

Nope, I think you've cleared things up nicely, thanks!

waddrmgr/scoped_manager.go
```diff
@@ -129,6 +130,31 @@ func (k KeyScope) String() string {
 	return fmt.Sprintf("m/%v'/%v'", k.Purpose, k.Coin)
 }
 
+// Identity is a closure that returns the identifier of an address.
+type Identity func() []byte
```
Roasbeef, line 134: Nice, this abstraction is much clearer for mer as a reader/reviewer.

Roasbeef (approved):
Amazing PR! 

The diff is looking really good from my PoV now. I think the only two things we'll want to do as follow up are:
  1. Ensure that on the lnd side, we have proper recovery working, which may need some additional modifications to the recovery/rescan logic. I think since it works at mostly a higher level and lets the waddrmgr derive the addresses it _should_ be ok. It can also make optimizations like not scan for taproot addrs if an address was created before a certain height, but that may generate some false positives. 
  2. (this one is a bonus imo) implement the PSBT parsing and assembly functionality.

guggero:
Rebased after the dependent PR has been merged.
Still need to verify the recovery of taproot addresses on the `lnd` side. But can always merge and create a bugfix later (will need a PR to remove the local replaces after pushing a tag anyway).

Roasbeef:
> [](https://github.com/guggero)Still need to verify the recovery of taproot addresses on the lnd side. But can always merge and create a bugfix later (will need a PR to remove the local replaces after pushing a tag anyway).

Definitely, I think we can easily parametrize our _existing_ integration tests for that, as they already test p2wkh/np2wkh, etc.

guggero:
> Definitely, I think we can easily parametrize our existing integration tests for that, as they already test p2wkh/np2wkh, etc.

I added that integration test to https://github.com/lightningnetwork/lnd/pull/6263, no issues found.

I also added a commit that exposes the `ImportTaprootScript` method on the _wallet_ level (and not only on the scoped manager level). That method is now also used in a unit test in `lnd`.

I think with those two things tested end-to-end, this PR is ready to be merged.

positiveblue (approved):
Great job on this one @guggero, really excited to see this ready to be merged :tada:

waddrmgr/address.go
```diff
@@ -796,38 +811,110 @@ var _ ManagedScriptAddress = (*witnessScriptAddress)(nil)
...
+// AddrType returns the address type of the managed address. This can be used
+// to quickly discern the address type without further processing
+//
+// This is part of the ManagedAddress interface implementation.
+func (a *taprootScriptAddress) AddrType() AddressType {
+	return TaprootScript
+}
+
+// Address returns the btcutil.Address which represents the managed address.
+// This will be a pay-to-witness-script-hash address.
```
carlaKC, line 894: Comment copied?

carlaKC (approved):
Great PR! Did a high level santiy pass, looks great. 

I think that the distinction between `TaprootAddress` and `TaprootScript` discussed [here](https://github.com/btcsuite/btcwallet/pull/792/commits/85b35fa5c049376bf2a6bdd9efd01dde279b5a01#r821015734) will be useful, the two different use cases make sense to me.

waddrmgr/scoped_manager.go
```diff
@@ -2085,14 +2105,15 @@ func (s *ScopedKeyManager) ImportWitnessScript(ns walletdb.ReadWriteBucket,
...
+		ns, WitnessScriptHashIdentity(script), script, bs,
+		WitnessScript, witnessVersion, isSecretScript,
 	)
 }
 
 // importScriptAddress imports a new pay-to-script or pay-to-witness-script
 // address.
 func (s *ScopedKeyManager) importScriptAddress(ns walletdb.ReadWriteBucket,
-	script []byte, bs *BlockStamp, addrType AddressType,
+	identity Identity, script []byte, bs *BlockStamp, addrType AddressType,
```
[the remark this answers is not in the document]
guggero: See comment here: https://github.com/btcsuite/btcwallet/pull/792#discussion_r812496834
Gives us the ability to use nicely named, pre-defined identity functions.

Issue #793, opened 2022-02-28
Add TxOut output scripts and values to ListLeasedOutputs


hitanshu-mehta:
Currently, [ListLeasedOutputs](https://github.com/btcsuite/btcwallet/blob/50ea8d1125561279f9c9af1e46bea16ca414b4a5/wallet/wallet.go#L2714)  only return lock id, expiration, and outpoint of locked Txn output. 
A proposal is to add output scripts and values in the response. The use case for these data has been raised at [lnd](https://github.com/lightningnetwork/lnd/issues/5202). It will increase the visibility of locked funds and can be helpful with auditing.

guggero:
Thanks for opening the issue. I agree that additional information of the UTXO would be great to have in the list of leased/locked outputs. IMO you can just start implementing and then create a PR. Ping me for a review.
There's quite some overlap of maintainers between `lnd` and `btcwallet` so I don't think you need to wait for "official approval" before starting the implementation if a feature is seen as useful in `lnd`.

PR #794, opened 2022-03-03
Add `Value` and `PkScript` in response of `ListLeasedOutputs` method.

hitanshu-mehta:
In this commit, we have created new struct to store the result of `ListLeasedOutputs`
method. We call `TxDetails` to get value and pkscript of output. Other output
fields are fetched as before.

Fix: #793 

Signed-off-by: Hitanshu Mehta <hitanshu99amehta@gmail.com>

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -2379,7 +2379,7 @@ type AccountResult struct {
 	TotalBalance btcutil.Amount
 }
 
-// AccountsResult is the resutl of the wallet's Accounts method.  See that
+// AccountsResult is the result of the wallet's Accounts method.  See that
 // method for more details.
 type AccountsResult struct {
 	Accounts           []AccountResult
@@ -2709,17 +2709,45 @@ func (w *Wallet) ListUnspent(minconf, maxconf int32,
 	return results, err
 }
 
+// ListLeasedOutputResult is a single result for the Wallet.ListLeasedOutputs method.
+// See that method for more details.
+type ListLeasedOutputResult struct {
+	*wtxmgr.LockedOutput
+	Value    int64
+	PkScript []byte
+}
+
 // ListLeasedOutputs returns a list of objects representing the currently locked
 // utxos.
-func (w *Wallet) ListLeasedOutputs() ([]*wtxmgr.LockedOutput, error) {
-	var outputs []*wtxmgr.LockedOutput
+func (w *Wallet) ListLeasedOutputs() ([]*ListLeasedOutputResult, error) {
+	var results []*ListLeasedOutputResult
 	err := walletdb.View(w.db, func(tx walletdb.ReadTx) error {
 		ns := tx.ReadBucket(wtxmgrNamespaceKey)
-		var err error
-		outputs, err = w.TxStore.ListLockedOutputs(ns)
-		return err
+		outputs, err := w.TxStore.ListLockedOutputs(ns)
+		if err != nil {
+			return err
+		}
+
+		for _, output := range outputs {
+			details, err := w.TxStore.TxDetails(ns, &output.Outpoint.Hash)
+			if err != nil {
+				return err
+			}
+
+			txOut := details.MsgTx.TxOut[output.Outpoint.Index]
+
+			result := &ListLeasedOutputResult{
+				LockedOutput: output,
+				Value:        txOut.Value,
+				PkScript:     txOut.PkScript,
+			}
+
+			results = append(results, result)
+		}
+
+		return nil
 	})
-	return outputs, err
+	return results, err
 }
 
 // DumpPrivKeys returns the WIF-encoded private keys for all addresses with


hitanshu-mehta:
@guggero PR is ready for review :)

wallet/wallet.go
```diff
@@ -2709,17 +2709,37 @@ func (w *Wallet) ListUnspent(minconf, maxconf int32,
 	return results, err
 }
 
+type ListLeasedOutputResult struct {
```
guggero, line 2712: Should we call this `LockedOutputDetail` instead? Also, some Godoc comment would be good.

wallet/wallet.go
```diff
@@ -2709,17 +2709,37 @@ func (w *Wallet) ListUnspent(minconf, maxconf int32,
...
+	wtxmgr.LockedOutput
```
guggero, line 2713: I think we can embed a pointer here instead.

wallet/wallet.go
```diff
@@ -2709,17 +2709,37 @@ func (w *Wallet) ListUnspent(minconf, maxconf int32,
...
 // utxos.
-func (w *Wallet) ListLeasedOutputs() ([]*wtxmgr.LockedOutput, error) {
-	var outputs []*wtxmgr.LockedOutput
+func (w *Wallet) ListLeasedOutputs() ([]*ListLeasedOutputResult, error) {
+	var results []*ListLeasedOutputResult
 	err := walletdb.View(w.db, func(tx walletdb.ReadTx) error {
 		ns := tx.ReadBucket(wtxmgrNamespaceKey)
-		var err error
-		outputs, err = w.TxStore.ListLockedOutputs(ns)
+		outputs, err := w.TxStore.ListLockedOutputs(ns)
```
guggero, line 2724: Need to check for the error here, since it will be overwritten in the loop below.

wallet/wallet.go
```diff
@@ -2709,17 +2709,37 @@ func (w *Wallet) ListUnspent(minconf, maxconf int32,
...
+
+		for _, output := range outputs {
+			details, err := w.TxStore.TxDetails(ns, &output.Outpoint.Hash)
+			if err != nil {
+				return err
+			}
+			result := &ListLeasedOutputResult{
+				LockedOutput: *output,
+				Value:        details.MsgTx.TxOut[output.Outpoint.Index].Value,
```
guggero, line 2733: nit: line too long. Extract `details.MsgTx.TxOut[output.Outpoint.Index]` as a variable to shorten?

wallet/wallet.go
guggero: Then on line 2740 it should just be `return nil`

guggero (commented):
Thanks for the PR!
Good idea to just create a new struct with the original info embedded in it. The even "cleaner" way would probably be to add those two fields to the `wtxmgr.LockedOutput` struct directly. But that requires an additional PR since `wtxmgr` is a submodule and would need its version bumped. So I'm fine with creating a new struct.

wallet/wallet.go
hitanshu-mehta: I named it `ListLeasedOutputResult` because it is more consistent with return types of other functions. For eg. [AccountBalanceResult](https://github.com/btcsuite/btcwallet/blob/50ea8d1125561279f9c9af1e46bea16ca414b4a5/wallet/wallet.go#L2377) , [ListUnspentResult](https://github.com/btcsuite/btcwallet/blob/50ea8d1125561279f9c9af1e46bea16ca414b4a5/wallet/wallet.go#L2583), [AccountBalanceResult](https://github.com/btcsuite/btcwallet/blob/50ea8d1125561279f9c9af1e46bea16ca414b4a5/wallet/wallet.go#L2461). `LockedOutputDetail` also seems good  to me though.

wallet/wallet.go
```diff
@@ -2722,22 +2724,27 @@ func (w *Wallet) ListLeasedOutputs() ([]*ListLeasedOutputResult, error) {
...
+		if err != nil {
+			return err
+		}
 
 		for _, output := range outputs {
 			details, err := w.TxStore.TxDetails(ns, &output.Outpoint.Hash)
 			if err != nil {
 				return err
 			}
+			value := details.MsgTx.TxOut[output.Outpoint.Index].Value
```
guggero, line 2736: Why not just `txOut := details.MsgTx.TxOut[output.Outpoint.Index]` and then below `Value: txOut.Value`...?
Also, empty lines are nice to get some "breathing room". They should be used to group statements together logically. So I'd suggest adding one after the `if err != nil { ... }` block and before `results = append(results, result)`.

hitanshu-mehta: I agree with both points. Sorry for making silly mistakes 🤦🏻‍♂️

guggero (approved):
Changes look good to me, thanks for the feature.

The commits should be squashed into a single one, but other than that the PR is good to go I think.

wallet/wallet.go
guggero: No worries ;-)

hitanshu-mehta:
Thank you for the quick review!! Commits are squashed.

Roasbeef (approved):
LGTM 🎊

PR #795, opened 2022-03-08
chain: RPC polling for block and tx events

ellemouton:
This PR reworks and replaces #784. It allows you to poll Bitcoind's RPC interface for new blocks and transactions instead of using the ZMQ interface. 

Thanks to @orbitalturtle for the original PR! 

Closes #784

### DIFF
--- a/chain/bitcoind_conn.go
+++ b/chain/bitcoind_conn.go
@@ -1,20 +1,16 @@
 package chain
 
 import (
-	"bytes"
 	"fmt"
-	"io"
 	"net"
 	"sync"
 	"sync/atomic"
-	"time"
 
 	"github.com/btcsuite/btcd/btcjson"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/rpcclient"
 	"github.com/btcsuite/btcd/wire"
-	"github.com/lightninglabs/gozmq"
 	"github.com/lightningnetwork/lnd/ticker"
 )
 
@@ -61,17 +57,9 @@ type BitcoindConfig struct {
 	// server.
 	Pass string
 
-	// ZMQBlockHost is the IP address and port of the bitcoind's rawblock
-	// listener.
-	ZMQBlockHost string
-
-	// ZMQTxHost is the IP address and port of the bitcoind's rawtx
-	// listener.
-	ZMQTxHost string
-
-	// ZMQReadDeadline represents the read deadline we'll apply when reading
-	// ZMQ messages from either subscription.
-	ZMQReadDeadline time.Duration
+	// ZMQConfig holds the configuration settings required for setting up
+	// zmq connections to bitcoind.
+	ZMQConfig *ZMQConfig
 
 	// Dialer is a closure we'll use to dial Bitcoin peers. If the chain
 	// backend is running over Tor, this must support dialing peers over Tor
@@ -106,16 +94,12 @@ type BitcoindConn struct {
 	// NOTE: This is nil when the bitcoind node is not pruned.
 	prunedBlockDispatcher *PrunedBlockDispatcher
 
-	// zmqBlockConn is the ZMQ connection we'll use to read raw block
-	// events.
-	zmqBlockConn *gozmq.Conn
-
-	// zmqTxConn is the ZMQ connection we'll use to read raw transaction
-	// events.
-	zmqTxConn *gozmq.Conn
+	// events handles the block and transaction events that are received or
+	// retrieved from bitcoind.
+	events BitcoindEvents
 
 	// rescanClients is the set of active bitcoind rescan clients to which
-	// ZMQ event notfications will be sent to.
+	// ZMQ event notifications will be sent to.
 	rescanClientsMtx sync.Mutex
 	rescanClients    map[uint64]*BitcoindClient
 
@@ -164,28 +148,6 @@ func NewBitcoindConn(cfg *BitcoindConfig) (*BitcoindConn, error) {
 			"pruned: %v", err)
 	}
 
-	// Establish two different ZMQ connections to bitcoind to retrieve block
-	// and transaction event notifications. We'll use two as a separation of
-	// concern to ensure one type of event isn't dropped from the connection
-	// queue due to another type of event filling it up.
-	zmqBlockConn, err := gozmq.Subscribe(
-		cfg.ZMQBlockHost, []string{rawBlockZMQCommand},
-		cfg.ZMQReadDeadline,
-	)
-	if err != nil {
-		return nil, fmt.Errorf("unable to subscribe for zmq block "+
-			"events: %v", err)
-	}
-
-	zmqTxConn, err := gozmq.Subscribe(
-		cfg.ZMQTxHost, []string{rawTxZMQCommand}, cfg.ZMQReadDeadline,
-	)
-	if err != nil {
-		zmqBlockConn.Close()
-		return nil, fmt.Errorf("unable to subscribe for zmq tx "+
-			"events: %v", err)
-	}
-
 	// Only initialize the PrunedBlockDispatcher when the connected bitcoind
 	// node is pruned.
 	var prunedBlockDispatcher *PrunedBlockDispatcher
@@ -207,15 +169,20 @@ func NewBitcoindConn(cfg *BitcoindConfig) (*BitcoindConn, error) {
 		}
 	}
 
-	return &BitcoindConn{
+	bc := &BitcoindConn{
 		cfg:                   *cfg,
 		client:                client,
 		prunedBlockDispatcher: prunedBlockDispatcher,
-		zmqBlockConn:          zmqBlockConn,
-		zmqTxConn:             zmqTxConn,
 		rescanClients:         make(map[uint64]*BitcoindClient),
 		quit:                  make(chan struct{}),
-	}, nil
+	}
+
+	bc.events, err = NewBitcoindEventSubscriber(cfg)
+	if err != nil {
+		return nil, err
+	}
+
+	return bc, nil
 }
 
 // Start attempts to establish a RPC and ZMQ connection to a bitcoind node. If
@@ -238,10 +205,10 @@ func (c *BitcoindConn) Start() error {
 	}
 
 	c.wg.Add(2)
-	go c.blockEventHandler()
-	go c.txEventHandler()
+	go c.sendBlockToClients()
+	go c.sendTxToClients()
 
-	return nil
+	return c.events.Start()
 }
 
 // Stop terminates the RPC and ZMQ connection to a bitcoind node and removes any
@@ -257,8 +224,10 @@ func (c *BitcoindConn) Stop() {
 
 	close(c.quit)
 	c.client.Shutdown()
-	c.zmqBlockConn.Close()
-	c.zmqTxConn.Close()
+
+	if err := c.events.Stop(); err != nil {
+		log.Errorf("error shutting down bitcoind events: %w", err)
+	}
 
 	if c.prunedBlockDispatcher != nil {
 		c.prunedBlockDispatcher.Stop()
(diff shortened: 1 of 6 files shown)


ellemouton:
cc @guggero (I cant request review in this repo)

chain/bitcoind_events_test.go
```diff
@@ -0,0 +1,434 @@
...
+func testNotifications(t *testing.T, miner *rpctest.Harness,
+	client *BitcoindClient) {
+
+	script, _, err := randPubKeyHashScript()
+	require.NoError(t, err)
+
+	tx, err := miner.CreateTransaction(
+		[]*wire.TxOut{{Value: 1000, PkScript: script}}, 5, false,
+	)
+	if err != nil {
```
guggero, line 123: nit: use require here and a few other places in this file.

chain/bitcoind_events_test.go
```diff
@@ -0,0 +1,434 @@
...
+	}
+
+	// Since the miners have been disconnected, we expect not to get any
+	// notifications from our client since our client is connected to
+	// miner1.
+	select {
+	case ntfn := <-ntfns:
+		t.Fatalf("received a notification of type %T but expected,"+
+			" none", ntfn)
+	case <-time.After(time.Millisecond * 10):
```
guggero, line 246: This seems quite short. Should we wait for a bit longer, just to be sure? Maybe half a second? Shouldn't add too much to the overall test duration, I think.

chain/bitcoind_conn.go
```diff
@@ -268,207 +234,34 @@ func (c *BitcoindConn) Stop() {
...
-			// event type. To prevent from logging it, we'll make
-			// sure it conforms to the ASCII standard.
-			if eventType == "" || !isASCII(eventType) {
-				continue
-			}
-
-			log.Warnf("Received unexpected event type from %v "+
-				"subscription: %v", rawBlockZMQCommand,
-				eventType)
+			return
```
guggero, line 252: Use `defer` for the unlock instead?

chain/bitcoind_events.go
```diff
@@ -319,3 +357,306 @@ func (b *bitcoindZMQEvents) Stop() error {
...
+		// it.
+		if _, ok := m.txs[tx.TxHash()]; ok {
+			delete(m.txs, tx.TxHash())
+		}
+	}
+}
+
+// contains returns true if the given transaction hash is already in our
+// mempool.
+func (m *mempool) contains(hash *chainhash.Hash) bool {
```
guggero, line 635: I think we should avoid passing a pointer here since that comes from a range/loop variable. Should probably de-reference it outside of this method (same for `add()`).

guggero (commented):
Great PR, turned out very clean! The refactoring alone is nice and the added polling feature will greatly simplify the setup for many users.
Curious to see how this performs under mainnet loads!

ellemouton:
Thanks for the quick review on this @guggero 💪 
I have updated it and have also opened the LND pr that makes use of this branch

chain/bitcoind_events.go
```diff
@@ -319,3 +354,306 @@ func (b *bitcoindZMQEvents) Stop() error {
...
+// clean removes any of the given transactions from the mempool if they are
+// found there.
+func (m *mempool) clean(txs []*wire.MsgTx) {
+	m.Lock()
+	defer m.Unlock()
+
+	for _, tx := range txs {
+		// If the transaction is in our mempool map, we need to delete
+		// it.
+		if _, ok := m.txs[tx.TxHash()]; ok {
```
guggero, line 624: nit: no need for the check, `delete()` will just ignore the key if it doesn't exist in the map.

guggero (approved):
Great PR, LGTM :100:

chain/bitcoind_events.go
ellemouton: fixed!

chain/bitcoind_conn.go
```diff
@@ -106,16 +94,11 @@ type BitcoindConn struct {
...
 	prunedBlockDispatcher *PrunedBlockDispatcher
 
-	// zmqBlockConn is the ZMQ connection we'll use to read raw block
-	// events.
-	zmqBlockConn *gozmq.Conn
-
-	// zmqTxConn is the ZMQ connection we'll use to read raw transaction
-	// events.
-	zmqTxConn *gozmq.Conn
+	// events
```
Roasbeef: Incomplete comment?

chain/bitcoind_conn.go
```diff
@@ -268,205 +234,32 @@ func (c *BitcoindConn) Stop() {
...
+func (c *BitcoindConn) sendBlockToClients(block *wire.MsgBlock) {
+	c.rescanClientsMtx.Lock()
+	defer c.rescanClientsMtx.Unlock()
 
-	for {
-		// Before attempting to read from the ZMQ socket, we'll make
-		// sure to check if we've been requested to shut down.
+	for _, client := range c.rescanClients {
 		select {
+		case client.blockNtfns <- block:
```
Roasbeef, line 250: Are these channels always buffered?

chain/bitcoind_conn.go
```diff
@@ -268,205 +234,32 @@ func (c *BitcoindConn) Stop() {
...
+func (c *BitcoindConn) sendTxToClients(tx *wire.MsgTx) {
+	c.rescanClientsMtx.Lock()
+	defer c.rescanClientsMtx.Unlock()
 
-	for {
-		// Before attempting to read from the ZMQ socket, we'll make
-		// sure to check if we've been requested to shut down.
+	for _, client := range c.rescanClients {
 		select {
+		case client.txNtfns <- tx:
```
Roasbeef, line 265: Same question here re channel buffering.

chain/bitcoind_events.go
```diff
@@ -21,10 +37,29 @@ type BitcoindEvents interface {
 
 // NewBitcoindEventSubscriber initialises a new BitcoinEvents object impl
 // depending on the config passed.
-func NewBitcoindEventSubscriber(cfg *BitcoindConfig,
+func NewBitcoindEventSubscriber(cfg *BitcoindConfig, client *rpcclient.Client,
 	notifyBlock func(block *wire.MsgBlock),
 	notifyTx func(tx *wire.MsgTx)) (BitcoindEvents, error) {
 
+	if cfg.PollingConfig != nil && cfg.PollingConfig.Enable {
```
Roasbeef, line 44: Add a sanity check to ensure that _both_ the polling and zmq configs aren't populated?

chain/bitcoind_events.go
```diff
@@ -319,3 +354,304 @@ func (b *bitcoindZMQEvents) Stop() error {
...
+func (b *bitcoindRPCPollingEvents) evictOldTransactions() {
+	defer b.wg.Done()
+
+	ticker := time.NewTicker(b.cfg.MempoolEvictionAge)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ticker.C:
+			b.mempool.removeOldTxs(b.cfg.MempoolEvictionAge)
```
Roasbeef, line 593: Why's this needed if we can count on new blocks to actually include those transactions? Also why do we need to add them to a mempool in either case?

Roasbeef:
Build+linter are failing.

Roasbeef (changes requested):
Nice work carrying out this much requested feature! Completed an initial pass, mainly need to wrap my ahead around the new "mempool" usage in the polling backend. There're a few places where we can also tighten up our input validation logic as well.

chain/bitcoind_events_test.go
```diff
@@ -0,0 +1,375 @@
...
+
+	_, err = client.SendRawTransaction(tx, true)
+	require.NoError(t, err)
+
+	ntfns := client.Notifications()
+
+	miner.Client.Generate(1)
+
+	// We cant be sure what order the notifications will arrive in, but
+	// we know that we expect three of them. We expect one of each of the
```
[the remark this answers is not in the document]
ellemouton: ah yes, you are correct. Updating 👍

chain/bitcoind_conn.go
ellemouton: no they are not. But they were not before either and this commit is meant to just be a refactor (see `zmqTxNtfns` and `zmqBlockNtfns`).
Should I change them to be buffered?

chain/bitcoind_events.go
ellemouton: if we want to api to be the same as the zmq one then the transaction notifications need to include mempool transactions right? if we one sent tx notifications for txs we  get from a block then it is no longer the same as the zmq api. And since we dont know if all txs in the mempool will get confirmed, we need a way to ensure that our mempool doesnt keep expanding. 

Agreed that this is not amazing. Open to suggestions for how to improve it!

ellemouton:
thanks for the review @Roasbeef 🚀 
Updated and responded to some comments.

ellemouton:
the fails have this message: `'go get' is no longer supported outside a module.`....

chappjc:
> the fails have this message: `'go get' is no longer supported outside a module.`....

Have a look at https://github.com/btcsuite/btcwallet/pull/800
I hit the issue too when CI started pulling latest go

ellemouton:
ah! awesome, thanks @chappjc !

ellemouton:
hmmm: 

`$ GO111MODULE=on go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@v1.18.0`
```
go: github.com/golangci/golangci-lint/cmd/golangci-lint@v1.18.0 (in github.com/golangci/golangci-lint@v1.18.0):
        The go.mod file for the module providing named packages contains one or
        more replace directives. It must not contain directives that would cause
        it to be interpreted differently than if it were the main module.

```

guggero:
Hmm, that seems to be new with go `1.18`? Seems like golangci-lint `v1.24.0` is the oldest version that doesn't have a `replace` directive. Might need to bump the linter to that?

Roasbeef:
https://github.com/golang/go/issues/44840

guggero:
I'll create a PR to fix the linter.

Roasbeef:
Can be rebased now that https://github.com/btcsuite/btcwallet/pull/800 is in.

chain/bitcoind_conn.go
Roasbeef: > no they are not. But they were not before either and this commit is meant to just be a refactor (see zmqTxNtfns and zmqBlockNtfns).

Interesting, I don't think it matter as much in the polling implementation, but for the zmq implementation, buffered channels here might help to mitigate some of the issues w/ the zmq high water mark setting in bitcond.

chain/bitcoind_events.go
Roasbeef: Hmm, yeah correct that we need the mempool notification to be intact for the `btcwallet` side of things. 

Lemmie take another look at this particular implementation to see if there're areas we can improve.

chain/bitcoind_conn.go
Roasbeef: I think the area that would matter more in this case is the transaction notifications: at times the mempool arrival rate can be somewhat high.

chain/bitcoind_events.go
```diff
@@ -0,0 +1,321 @@
...
+	"time"
+
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/gozmq"
+)
+
+// BitcoindEvents is the interface that must be satisfied by any type that
+// serves bitcoind block and transactions events.
+type BitcoindEvents interface {
+	Start() error
```
Roasbeef, line 34: Hmm, fwiw this interface as defined, doesn't really say much (anything w/ a start + stop method). What if we modified the interface to instead define two methods (in addition to these) which each return a channel to listen on for events? 

So something like this: 
```go
type BitcoindEvents interface {
    TxNotifications() <-chan *wire.MsgTx
    BlockNotifications() <-chan *wire.MsgBlock
}
```

I imagine this interface would also make it easier to write generic unit tests as well.

chain/bitcoind_events.go
```diff
@@ -0,0 +1,321 @@
...
+		case rawBlockZMQCommand:
+			block := &wire.MsgBlock{}
+			r := bytes.NewReader(bufs[1])
+			if err := block.Deserialize(r); err != nil {
+				log.Errorf("Unable to deserialize block: %v",
+					err)
+				continue
+			}
+
+			b.notifyBlock(block)
```
Roasbeef, line 237: So with my suggestion above, this here would instead select on the channel write or quit.

chain/bitcoind_events.go
```diff
@@ -319,3 +359,300 @@ func (b *bitcoindZMQEvents) Stop() error {
...
+				}
+
+				// notify the client of the new block.
+				b.notifyBlock(newBlock)
+
+				// From our local mempool map, let's remove each
+				// of the transactions that are confirmed in
+				// this new block, since they are no longer in
+				// the mempool.
+				b.mempool.clean(newBlock.Transactions)
```
Roasbeef, line 514: If we do this, then why do we need that time based eviction? 

I'm not sure so that time based eviction is either correct either. Consider that here we remove the items from the mempool, but the _next_ time around (when we poll the mempool below), we'd potentially re-notify the same transaction. Consider a scenario in the future where a transaction sits around in the mempool for hours, depending on our mempool purge interval, we'd repeatedly re-notify that transaction...

Roasbeef (changes requested):
The latest diff is looking really good! 

Two main comments now:
 1. Massaging the methods of the new interfaces to be more indicative w.r.t what the interface actually _does_ 
 2. Properly avoiding duplicate notifications for transactions in the mempool

Roasbeef:
Re #2, It think one subtly here is that the bitcoind/btcd streaming implementations are based on a _mempool level hook_, which gets called each time a transaction is added. Therefore they don't need to worry about these duplicate notifications in theory (both still run into at-least-once delivery semantics). In other words, we may need to implement stricter filtering on this level, while also navigating memory blow up concerns (node online for 30 days retains that 30 day set of items in the mepool). 

Re constraining memory in this instance, most bitcoind instances allocate 300 MB to the mempool. Assuming 400 bytes for a transaction, this gives us max 750k transactions, which at 32 bytes per transaction (track the txid) gives us an upper bound of 24 MB...

chain/bitcoind_events.go
ellemouton: lemme know what you think of the latest method. it pretty much ensures that we have an exact reflection of the latest mempool

ellemouton (commented):
Thanks for the helpful review @Roasbeef ! 

Let me know what you think of the latest updates.

chain/bitcoind_events.go
```diff
@@ -339,3 +377,329 @@ func (b *bitcoindZMQEvents) txEventHandler() {
...
+				case <-b.quit:
+					return
+				}
+			}
+
+			// Now, we clear our internal mempool of any unmarked
+			// transactions. These are all the transactions that
+			// we still have in the mempool but that were not
+			// returned in the latest GetRawMempool query.
+			b.mempool.deleteUnmarked()
```
Roasbeef, line 612: Ah ok, so we're essentially computing the set difference as we scan through mempool, then removing those items at the very end.

Roasbeef (approved):
LGTM ❄️

PR #797, opened 2022-03-21
txsizes: count change output in estimating tx size

unnoisy-gypsy:
In the estimation of transaction size, there is an uncommon scenario in which the required space for serializing the count of outputs is calculated wrong since the change output is not counted in the count of outputs.

The source of the problem is just an unused variable which this PR fixes it.


Roasbeef (changes requested):
Great fix! 

Only final ask here is a test case that reproduces the size divergence and shows that this PR fixes things. It's clear from the diff that if we added a change output that it wouldn't be accounted for, but ideally we can have that assertion locked in at the test level.

PR #798, opened 2022-03-22
mod: bump wallet/txauthor, remove local replace

guggero:
Now that we have a new tagged version of wallet/txauthor, we no longer
need the local replace directive.


Roasbeef (approved):
LGTM 🌂

jcvernaleo (approved):
Nice.

OK

PR #799, opened 2022-04-01
waddrmgr: fix used address detection for p2tr addrs

guggero:
This patch makes sure that p2tr addresses are detected as being used
correctly.


positiveblue (approved):
Good catch

PR #783, opened 2021-12-03
legacyrpc: add address type arg to getNewAddress

chappjc:
Resolves https://github.com/btcsuite/btcwallet/issues/646 and https://github.com/btcsuite/btcwallet/issues/762
<s>Requires https://github.com/btcsuite/btcd/pull/1776 (see the `replace` to my btcd repository that would be removed if that is merged). The btcctl from that branch would also be needed for testing this.</s>

This handles the `AddressType` argument in the `getNewAddress` and `getRawChangeAddress` RPC handlers. It recognizes "legacy", "p2sh-segwit", and "bech32" to match Bitcoin Core's RPC options.  These correspond to the bip44, bip49 "plus", and bip84 paths.  These are the `waddrmgr.DefaultKeyScopes`.

The `validateaddress` RPC is already able to recognize these addresses, and it may be used to verify the addresses returned by both `getnewaddress` and `getrawchangeaddress`.

This adds `btcjson.ErrRPCWalletInvalidAddressType` with a code (-5) and message ("unknown address type") to match Bitcoin Core's v22.

PR #800, opened 2022-04-09
CI: fix Makefile not working with go 1.18

chappjc:
go get is no longer used to install or build things. Use go install
instead.  Also specify a version so the module's go.mod is not
touched when installing the CI tools.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -11,37 +11,29 @@ linters-settings:
     simplify: true
 
 linters:
-  enable-all: true
-  disable:
-    # Global variables are used in many places throughout the code base.
-    - gochecknoglobals
-
-    # Some lines are over 80 characters on purpose and we don't want to make them
-    # even longer by marking them as 'nolint'.
-    - lll
-
-    # We don't care (enough) about misaligned structs to lint that.
-    - maligned
-
-    # We have long functions, especially in tests. Moving or renaming those would
-    # trigger funlen problems that we may not want to solve at that time.
-    - funlen
-
-    # Disable for now as we haven't yet tuned the sensitivity to our codebase
-    # yet.  Enabling by default for example, would also force new contributors to
-    # potentially extensively refactor code, when they want to smaller change to
-    # land.
-    - gocyclo
-
-    # Instances of table driven tests that don't pre-allocate shouldn't trigger
-    # the linter.
+  disable-all: true
+  enable:
+    - asciicheck
+    - deadcode
+    - dupl
+    - errcheck
+    - goimports
+    - gosec
+    - gosimple
+    - govet
+    - ineffassign
+    - nolintlint
     - prealloc
-
-    # Init functions are used by loggers throughout the codebase.
-    - gochecknoinits
+    - staticcheck
+    - structcheck
+    - typecheck
+    - unconvert
+    - unused
+    - varcheck
     
-    # Explicit types are okay.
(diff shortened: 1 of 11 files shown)


guggero (approved):
LGTM :tada:

Roasbeef:
Approved for a CI run, should be automatic after this gets in (github new contributor clause).

Roasbeef:
I wonder if we'll run into this here? https://github.com/golang/go/issues/44840

guggero:
It looks like you need to use at least `golangci-lint v1.40.0` to get this fixed. But that will require a few of the new rules that were introduced in that version to be addressed as well.

chappjc:
I'll update the linter and fix whatever.  Just a few and I'll be on it...

chappjc:
Oh nevermind... https://github.com/btcsuite/btcwallet/pull/795#issuecomment-1102997537

guggero:
> Oh nevermind... [#795 (comment)](https://github.com/btcsuite/btcwallet/pull/795#issuecomment-1102997537)

Actually, I just found your PR instead that aims to fix it. So I didn't start on a PR myself.

chappjc:
Oh, haha, let me try latest golangci-lint locally and see what needs to update.

chappjc:
Yikes it's a LOT, at least if you use the default enabled linters.  
To pass as-is needs `--disable errcheck,deadcode,unused,ineffassign,staticcheck,varcheck,structcheck`
I have a feeling most of those except perhaps `errcheck`, which is super annoying, are worth satisfying, but it looks to be a lot of work.

chappjc:
But the few lint issues thrown by the current version of golangci-lint are fixed in https://github.com/btcsuite/btcwallet/pull/783/commits/1d1c00321dca8c396f1558665f2d3cc5d2118db6 (part of https://github.com/btcsuite/btcwallet/pull/783).

guggero:
If a lot of errors come from the same (new) linter, I think an option might also be to disable the linter for now and then fix the issues in a later commit.

chappjc:
K, lemme see how many would need to be disabled for btcwallet using golangci-ling 1.45.2, which I think is needed for go 1.18 practically

guggero:
I tried and it looked like `v1.40.0` is the oldest version that actually compiles with golang 1.18.

chappjc:
OK wow, I'd suggest whitelisting linters (--disable-all and selective --enable) instead of the blacklisting (--enable-all and selective --disable).  There are more than 20 linters (so far) that are going to be extremely difficult to satisfy.

Alternative, use the 10 default linters and add any extras you want `[deadcode errcheck gosimple govet ineffassign staticcheck structcheck typecheck unused varcheck]`

<details>
  <summary>just some that need to be disabled</summary>
```
    - errlint

    - paralleltest
    - exhaustive
    - nestif
    - gocognit
    - cyclop
    - tparallel
    - testpackage
    - gci
    - errname
    - godox
    - wrapcheck
    - varnamelen
    - gomnd
    - ifshort
    - ireturn
    - nilnil
    - nlreturn
    - stylecheck
    - goerr113
    - exhaustivestruct
 ```
</details>

Roasbeef:
Approved another CI run...

Roasbeef:
CI looking pretty good...

Roasbeef (approved):
LGTM 🏉

PR #801, opened 2022-05-04
chain: fix double-spend removal not being triggered

qustavo:
Unspent txns that are double-spent to an address not controlled by us (hence, not being watched) are not detected by the wallet and remain listed by `ListUnspent` even when the block has been confirmed and does not include the initial (double-spent) tx.

The fix has been implemented on the bitcoin client but with some expansion in `chain.Interface` it could be extracted out, probably by using `NotifySpent`.

Related to https://github.com/lightningnetwork/lnd/issues/5244

### DIFF
--- a/chain/bitcoind_client.go
+++ b/chain/bitcoind_client.go
@@ -59,10 +59,11 @@ type BitcoindClient struct {
 	// watchedAddresses, watchedOutPoints, and watchedTxs are the set of
 	// items we should match transactions against while processing a chain
 	// rescan to determine if they are relevant to the client.
-	watchMtx         sync.RWMutex
-	watchedAddresses map[string]struct{}
-	watchedOutPoints map[wire.OutPoint]struct{}
-	watchedTxs       map[chainhash.Hash]struct{}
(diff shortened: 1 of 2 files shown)


saubyk:
cc: @positiveblue

positiveblue:
Hi @qustavo thank you for this PR. I took a look at it, the idea seems good but I think it would be useful to add some tests for this. We should make sure this works during reorgs too.

qustavo:
Hi @positiveblue I agree, but I'm not sure how to create a re-org test environment, could you give me any direction about where to start?

Roasbeef:
@qustavo here're some tests in lnd that exercise re-orgs for the wallet: https://github.com/lightningnetwork/lnd/blob/master/lnwallet/test/test_interface.go#L2027

ziggie1984:
I will take this over if you don't mind @qustavo ?

guggero:
Closing due to inactivity, feel free to take over, @ziggie1984.

qustavo:
go for it @ziggie1984

PR #802, opened 2022-05-11
wallet: add Taproot fields to FundPsbt

guggero:
This bumps the `btcd/btcutil/psbt` library to `v1.1.4` that supports new Taproot fields in PSBTs.
We then add those fields to Taproot inputs, which, in the wallet's case, will only ever be BIP0086 inputs.

cc @Roasbeef, @positiveblue.


positiveblue:
LGTM but I do not have rights in this repo to approve it :+1:

Roasbeef (approved):
LGTM 🪡

PR #803, opened 2022-05-31
wallet: always set regular BIP 32 info for taproot PSBT inputs

Roasbeef:
Fixes https://github.com/lightningnetwork/lnd/issues/6579. 

`lnd` always checks that the main derivation exists, and uses that to derive the signing keys. An alternative would be to fix the issue on the `lnd` side, but perhaps this is a better solution assuming that these fields aren't mutually exclusive from the PoV of PSBT?


guggero (approved):
My bad, forgot to add the legacy path.

tACK, tested that `bitcoind` and `lnd` still are able to sign everything correctly.

PR #804, opened 2022-06-02
build: update to btcd v0.23.0 and neutrino v0.14.2

Roasbeef:

PR #805, opened 2022-06-06
build+chain: update to latest btcd, neutrino and use TCP conn instead of io.Pipe

Crypt-iQ:


Roasbeef:
Replaces https://github.com/btcsuite/btcwallet/pull/804

Roasbeef:
I'm gonna tag a new release on the btcd side that includes two bug fixes, then we can update this, merged it in, then tag the final (?) release candidate for lnd 0.15.

guggero (commented):
Seems unrelated to your changes, but the tests currently fail with:

```
     rpchelp_test.go:59: Cannot generate 'en_US' help for method 'getnewaddress': missing description for 'getnewaddress-addresstype'
    rpchelp_test.go:59: Cannot generate 'en_US' help for method 'getrawchangeaddress': missing description for 'getrawchangeaddress-addresstype'
    rpchelp_test.go:46: Regenerate help texts with 'go generate' after fixing
```

Or is that because things changed in the new `btcd` version?

Crypt-iQ:
I think it's because of the btcd bump

chappjc:
This PR was intended to merge with the corresponding btcjson changes that added the address type arg: https://github.com/btcsuite/btcwallet/pull/783/files
Part of that is regenerating the rpcserverhelp.go file

chappjc:
So I had to delete block_headers.bin, neutrino.db, and reg_filter_headers.bin, then recover, but SPV seems to be working again.
Is that expected when updating neutrino from 0.13.2 to 0.14.2?  Not terribly inconvenient, just unexpected.

Crypt-iQ:
> So I had to delete block_headers.bin, neutrino.db, and reg_filter_headers.bin, then recover, but SPV seems to be working again. Is that expected when updating neutrino from 0.13.2 to 0.14.2? Not terribly inconvenient, just unexpected.

That really shouldn't be the case, it should be a smooth upgrade

Crypt-iQ:
I think this branch needs to wait for btcd 0.23.1 to be tagged

cmd/sweepaccount/main.go
```diff
@@ -196,7 +196,9 @@ func makeDestinationScriptSource(rpcClient *rpcclient.Client, accountName string
 	// GetNewAddress always returns a P2PKH address since it assumes
 	// BIP-0044.
 	newChangeScript := func() ([]byte, error) {
-		destinationAddress, err := rpcClient.GetNewAddress(accountName)
+		destinationAddress, err := rpcClient.GetNewAddress(
+			accountName, "legacy",
+		)
```
chappjc, line 201: You'll be able to revert this when you move to btcd/v0.23.1.

I also see you have the `sweepaccount` binary accidentally checked in.

Roasbeef:
Can be rebased now that the new `btcd` version has been tagged!

Roasbeef:
@chappjc above was that using connect peer or addpeer?

Roasbeef (approved):
LGTM 🐞

chappjc:
> @chappjc above was that using connect peer or addpeer?

Yep, a `connect` peer.
 
> We added a neutrino sub-server to lnd (gRPC access to neutrino) in 0.15. IMO we should move that over to the neutrino repo, then other projects can issue commands to help diagnose or unstick things when stuff like that happens.

Hmm, I'm not sure I see what happened here.  I restarted with the same connect peer, just after deleting the db/dat files, and it got past it by pulling everything all over.

PR #811, opened 2022-09-19
Update txsizes for correct taproot size estimation

benthecarman:


wallet/txauthor/go.mod
```diff
@@ -6,5 +6,5 @@ require (
 	github.com/btcsuite/btcd v0.22.0-beta.0.20220316175102-8d5c75c28923
 	github.com/btcsuite/btcd/btcutil v1.1.1
 	github.com/btcsuite/btcwallet/wallet/txrules v1.2.0
-	github.com/btcsuite/btcwallet/wallet/txsizes v1.1.0
+	github.com/btcsuite/btcwallet/wallet/txsizes v1.2.1
```
guggero, line 9: I think this will also result in a change in `go.sum`, you might need to run `go mod tidy` in the `wallet/txauthor` directory.

guggero (approved):
LGTM :tada:

wallet/txauthor/go.mod
benthecarman: done

Roasbeef (approved):
LGTM 🦋

Commit message nit: can you add `wallet/txauthor` as a prefix to your message and make the first letter of the title lower case? thx

benthecarman:
> Commit message nit: can you add wallet/txauthor as a prefix to your message and make the first letter of the title lower case? thx

done

Roasbeef:
Ok the final tag has been pushed (`wallet/txauthor/v1.3.0`). We can now update on the lnd side. 

I'll also tag a new `btcwallet` version as well then we can update to that since that should smooth out the update.

Roasbeef:
I've done this here: https://github.com/btcsuite/btcwallet/pull/812

PR #812, opened 2022-09-22
build: update to latest versions of txauthor and txsizes

Roasbeef:
This fixes an error in taproot based fee estimation.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,9 +7,9 @@ require (
 	github.com/btcsuite/btcd/btcutil/psbt v1.1.4
 	github.com/btcsuite/btcd/chaincfg/chainhash v1.0.1
 	github.com/btcsuite/btclog v0.0.0-20170628155309-84c8d2346e9f
-	github.com/btcsuite/btcwallet/wallet/txauthor v1.2.3
+	github.com/btcsuite/btcwallet/wallet/txauthor v1.3.2
 	github.com/btcsuite/btcwallet/wallet/txrules v1.2.0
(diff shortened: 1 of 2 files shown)


wallet/txauthor/author.go
```diff
@@ -95,7 +95,7 @@ func NewUnsignedTransaction(outputs []*wire.TxOut, feeRatePerKb btcutil.Amount,
 
 	targetAmount := SumOutputValues(outputs)
 	estimatedSize := txsizes.EstimateVirtualSize(
-		0, 1, 0, outputs, changeSource.ScriptSize,
+		0, 0, 1, 0, outputs, changeSource.ScriptSize,
```
guggero, line 98: hmm, unfortunately this change won't take effect until we push yet another version of the `txauthor` package.

Roasbeef: Ah true, since we updated `txauthor` in the past PR that added this to `txsizes`...

Roasbeef:
Ok I made #813 that fixed the issue in the package. 

I thought we fixed this in the past, but we need to start building/running tests for each of the modules individually. Otherwise, if we update a sub-module, then we just run the tests from the PoV main module, which depends on an earlier version of the module, so this change isn't detected.

Roasbeef:
Ok, so the new path is:
  * Merge #814 -- tag a new version 
  * Fix the `txauthor` package, update to above version 
  * Merge that -- tag new version
  * Update this PR after the steps above are done

guggero (approved):

PR #813, opened 2022-09-23
wallet/txauthor: fix call to EstimateVirtualSize

Roasbeef:
The CI didn't pick this up before since the main wallet was still using an older version. We'll tag a new txauthor version after this, which'll let us actually update the wallet and eventually lnd.


Roasbeef:
Once this is in, we'll tag `wallet/txauthor/v1.3.1` then we can proceed with #812.

PR #816, opened 2022-09-26
txsize: add WU for stack size

guggero:
The input witness estimation for a P2TR was missing the single weight unit for the stack size encoded as a compact int.
This was balanced out by the worst case estimation of the Schnorr signature which can contain the sighash flag, so this wasn't discovered earlier.

@Roasbeef this requires a new tag to be pushed: `wallet/txsizes/v1.2.3`.
cc @benthecarman.


Roasbeef (approved):
LGTM 🛤

PR #817, opened 2022-09-26
wallet: add script size for P2TR change output

guggero:
This commit adds a script size for a P2TR change output. Without it, the change size for a P2TR output will always be counted as 0 which leads to invalid transactions.

Depends on https://github.com/btcsuite/btcwallet/pull/816.

cc @Roasbeef, @benthecarman.


Roasbeef:
Replaced by https://github.com/btcsuite/btcwallet/pull/818

PR #820, opened 2022-10-15
Bug/rescan datarace

### LINKED ISSUE #819: NeutrinoClient data race
## Problem
The `NeturinoClient` `Rescan` method locks, sets `rescan` to `nil` then unlocks itself before resetting the value.
As the client state is `scanning = true` yet `rescan = nil` when the lock is released there is a race between the executing `Rescan` method and any `NotifyReceived` calls that are waiting to obtain the lock.  If a `NotifyReceived` call picks up the lock in this state, it will panic when it tries to run `rescan.Update` as `scanning` is `true`.

https://github.com/btcsuite/btcwallet/blob/master/chain/neutrino.go#L351

__Originally analyzed and found by @guggero in https://github.com/lightningnetwork/lnd/issues/6985__

```
--- FAIL: TestLightningWallet (7.74s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x7d47b7]

goroutine 6 [running]:
testing.tRunner.func1.2({0x9c34e0, 0xfdbfe0})
	/opt/hostedtoolcache/go/1.19.0/x64/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.19.0/x64/src/testing/testing.go:1399 +0x39f
panic({0x9c34e0, 0xfdbfe0})
	/opt/hostedtoolcache/go/1.19.0/x64/src/runtime/panic.go:884 +0x212
github.com/lightninglabs/neutrino.(*Rescan).Update(0x0, {0xc00064a248, 0x1, 0xc00129a4e0?})
	/home/runner/work/go/pkg/mod/github.com/lightninglabs/neutrino@v0.14.2/rescan.go:1391 +0x57
github.com/btcsuite/btcwallet/chain.(*NeutrinoClient).NotifyReceived(0xc0000fa120, {0xc0012bc060?, 0x1, 0x1})
	/home/runner/work/go/pkg/mod/github.com/btcsuite/btcwallet@v0.16.1/chain/neutrino.go:457 +0x159
github.com/btcsuite/btcwallet/wallet.(*Wallet).NewAddress(0xc0001ce480, 0x0, {0x0?, 0x9f16e0?})
	/home/runner/work/go/pkg/mod/github.com/btcsuite/btcwallet@v0.16.1/wallet/wallet.go:3020 +0x17a
github.com/lightningnetwork/lnd/lnwallet/btcwallet.(*BtcWallet).NewAddress(0xc0005a8180, 0x50?, 0x0, {0xa88a60?, 0x7?})
	/home/runner/work/lnd/lnd/lnwallet/btcwallet/btcwallet.go:556 +0x10c
github.com/lightningnetwork/lnd/lnwallet/test.loadTestCredits(0xc000078720, 0xc000340380, 0x14, 0x0?)
	/home/runner/work/lnd/lnd/lnwallet/test/test_interface.go:275 +0x287
github.com/lightningnetwork/lnd/lnwallet/test.createTestWallet({0xc00031ff80?, 0xa87ad1?}, 0x5?, 0xfe8080, {0xc43560, 0xc0001ca180}, {0xc48820, 0xc0005a8180}, {0xc43b00, 0xc0011c41c8}, ...)
	/home/runner/work/lnd/lnd/lnwallet/test/test_interface.go:378 +0x4a9
github.com/lightningnetwork/lnd/lnwallet/test.runTests(_, _, {_, _}, _, {{0xc00002eba0, 0xe}, {0xa873bc, 0x2}, {0xa879c6, ...}, ...}, ...)
	/home/runner/work/lnd/lnd/lnwallet/test/test_interface.go:3398 +0x2633
github.com/lightningnetwork/lnd/lnwallet/test.TestLightningWallet(0xc000007ba0, {0xa89361, 0x8})
	/home/runner/work/lnd/lnd/lnwallet/test/test_interface.go:3064 +0x8ef
github.com/lightningnetwork/lnd/lnwallet/test/neutrino_test.TestLightningWallet(0x0?)
	/home/runner/work/lnd/lnd/lnwallet/test/neutrino/neutrino_test.go:12 +0x25
testing.tRunner(0xc000007ba0, 0xb9f9b8)
	/opt/hostedtoolcache/go/1.19.0/x64/src/testing/testing.go:1446 +0x10b
created by testing.(*T).Run
	/opt/hostedtoolcache/go/1.19.0/x64/src/testing/testing.go:1493 +0x35f
FAIL	github.com/lightningnetwork/lnd/lnwallet/test/neutrino	7.770s
FAIL
```

### Related Work

See PR #695
See PR #533

## Solution 1
Make a change against `btcwallet` to address the state management issues:
1. make the entire `Rescan` method a critical path (i.e., only unlock via a single defer call)
2. make the entire `NotifyReceived` method a critical path
3. remove the locking in the `notificationHandler` to prevent a deadlock caused by making `Rescan` completely critical.  This locking isn't necessary as it attempts to read the state of the `rescanErr` channel.  But if `rescanErr` is nil, then there is no `Rescan` object created yet so no errors to read.  Also, reading from a `nil` channel will simply block but the read is wrapped in a select statement with other exits.

### Risks
- there are no unit tests on the neutrino client in btcwallet and even a targeted change like this is hard to know if it introduced any regressions

MStreet3:
### Background
There is datarace in the neutrino client implementation that
is causing test flakes in `lnd`.  Specifically, there was a shared
struct property `rescan` whose updates resulted in a data race with
the possibility of a panic when `rescan` was `nil`.  Further investigation
revealed race conditions on the `rescanQuit` and `rescanErr`
props of the client as well.

### Changes
This PR modifies the design of the neutrino client to better reflect
the requirement that only a single `Rescan` object instance be
running at a time.  To that end, this PR removes the `rescan` property
on the client and replaces it with two buffered channels (`cap` = 1), one each for
a rescanner and the rescanner's signal channel.  In this way there is no
need to maintain the state `scanning`.  Now any method can
infer that the client is scanning if it is able to read a scanner from the channel.

This PR resolves a data race on the `rescanErr` channel via a
new error chan consumer method that takes all new error channels
from rescanner restarts and feeds any read errors onto `rescanErr`.

### Testing
This PR adds two interfaces `rescan.Interface` and `NeutrinoChainService`
to abstract out the dependency on structs from the `neutrino` package.

This PR adds a file of mock implementations for these interfaces and uses them
in a new unit test file.  The unit tests added are not intended to capture all
the functionality of the client, but to demonstrate that the data races are resolved
and these tests should be run with the `-race` flag enabled.

### Reviewing
1. Checkout 93ca6856b389ea492e787250c2fb3beaa3ece411 and run tests.
`TestNeutrinoClientNotifyReceivedRescan` will fail with a segmentation fault 
in the neutrino client due to`rescan` being set to `nil`.

2. Checkout 6e1d604ce18c21c9e9f08a500301248a5b466739 and verify that
the segmentation fault is fixed and `TestNeutrinoClientNotifyReceivedRescan`
passes, but that `-race` flag testing fails.

3. Checkout 56af455326cbe9dfacf81dfbb19ef3c3b8043501 run tests and
verify that `TestNeutrinoClientNotifyReceivedRescan` passes with `-race` flag enabled.

All changes after 56af455326cbe9dfacf81dfbb19ef3c3b8043501 are cosmetic and
cleanup related.

fixes #819

### DIFF
--- a/chain/mocks_test.go
+++ b/chain/mocks_test.go
@@ -30,6 +30,8 @@ var (
 	_ NeutrinoChainService = (*mockChainService)(nil)
 )
 
+// newMockNeutrinoClient constructs a neutrino client with a mock chain
+// service implementation and mock rescan Interface implementation.
 func newMockNeutrinoClient(t *testing.T,
 	opts ...func(*mockRescanner)) *NeutrinoClient {
 	t.Helper()
@@ -54,10 +56,10 @@ func newMockNeutrinoClient(t *testing.T,
 	}
 }
 
+// mockRescanner is a mock implementation of a rescan Interface for use in
+// tests.
 type mockRescanner struct {
 	updateArgs *list.List
-	errs       []error
-	rescanQuit <-chan struct{}
 }
 
 func (m *mockRescanner) Start() <-chan error {
@@ -74,6 +76,8 @@ func (m *mockRescanner) Update(opts ...neutrino.UpdateOption) error {
 	return nil
 }
 
+// mockChainService is a mock implementation of a chain service for use in
+// tests.
 type mockChainService struct{}
 
 func (m *mockChainService) Start() error {
--- a/chain/mocks_test.go
+++ b/chain/mocks_test.go
@@ -49,10 +49,10 @@ func newMockNeutrinoClient(t *testing.T,
 	}
 
 	return &NeutrinoClient{
-		CS:           &mockChainService{},
-		newRescanner: newRescanner,
-		rescanCh:     make(chan rescan.Interface, 1),
-		rescanQuitCh: make(chan chan struct{}, 1),
+		CS:            &mockChainService{},
+		newRescanFunc: newRescanner,
+		rescanCh:      make(chan rescan.Interface, 1),
+		rescanQuitCh:  make(chan chan struct{}, 1),
 	}
 }
 
--- a/chain/mocks_test.go
+++ b/chain/mocks_test.go
@@ -34,8 +34,21 @@ func newMockNeutrinoClient(t *testing.T,
 	opts ...func(*mockRescanner)) *NeutrinoClient {
 	t.Helper()
 
+	// newRescanner returns a mockRescanner with options set by the
+	// newMockNeutrinoClient function
+	newRescanner := func(ro ...neutrino.RescanOption) rescan.Interface {
+		mrs := &mockRescanner{
+			updateArgs: list.New(),
+		}
+		for _, o := range opts {
+			o(mrs)
+		}
+		return mrs
+	}
+
 	return &NeutrinoClient{
-		CS: &mockChainService{},
+		CS:           &mockChainService{},
+		newRescanner: newRescanner,
 	}
 }
 
--- a/chain/mocks_test.go
+++ b/chain/mocks_test.go
@@ -1,0 +1,152 @@
+package chain
+
+import (
+	"container/list"
+	"errors"
+	"testing"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/gcs"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/chain/internal/rescan"
+	"github.com/lightninglabs/neutrino"
+	"github.com/lightninglabs/neutrino/banman"
+	"github.com/lightninglabs/neutrino/headerfs"
+)
+
+// Define variables for mocks
+var (
+	ErrNotImplemented = errors.New("not implemented")
+	testBestBlock     = &headerfs.BlockStamp{
+		Height: 42,
+	}
+)
+
+// Define static type checks
+var (
+	_ rescan.Interface     = (*mockRescanner)(nil)
+	_ NeutrinoChainService = (*mockChainService)(nil)
+)
+
+func newMockNeutrinoClient(t *testing.T,
+	opts ...func(*mockRescanner)) *NeutrinoClient {
+	t.Helper()
+
+	return &NeutrinoClient{
+		CS: &mockChainService{},
+	}
+}
+
+type mockRescanner struct {
+	updateArgs *list.List
+	errs       []error
+	rescanQuit <-chan struct{}
+}
+
+func (m *mockRescanner) Start() <-chan error {
+	errs := make(chan error)
+	return errs
+}
+
+func (m *mockRescanner) WaitForShutdown() {
+	// no-op
+}
+
+func (m *mockRescanner) Update(opts ...neutrino.UpdateOption) error {
+	m.updateArgs.PushBack(opts)
+	return nil
+}
+
+type mockChainService struct{}
+
+func (m *mockChainService) Start() error {
+	return nil
+}
+
+func (m *mockChainService) GetBlock(chainhash.Hash,
+	...neutrino.QueryOption) (*btcutil.Block, error) {
+	return nil, ErrNotImplemented
+}
+
+func (m *mockChainService) GetBlockHeight(*chainhash.Hash) (int32, error) {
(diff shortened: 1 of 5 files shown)


chain/mocks_test.go
```diff
@@ -0,0 +1,180 @@
...
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/neutrino"
+	"github.com/lightninglabs/neutrino/banman"
+	"github.com/lightninglabs/neutrino/headerfs"
+)
+
+var (
+	ErrNotImplemented                      = errors.New("not implemented")
+	_                 NeutrinoRescanner    = (*mockRescanner)(nil)
```
guggero, line 20: nit: separate the variable definitions and the type assertions into their own groups so the formatting looks a bit less "stretched"?

chain/mocks_test.go
```diff
@@ -0,0 +1,180 @@
...
+	_                 NeutrinoChainService = (*mockChainService)(nil)
+	testBestBlock                          = &headerfs.BlockStamp{
+		Height: 42,
+	}
+)
+
+func newMockNeutrinoClient(t *testing.T,
+	opts ...func(*mockRescanner)) *NeutrinoClient {
```
guggero, line 28: nit: add newline after multi-line function definition for better readability.

chain/mocks_test.go
```diff
@@ -0,0 +1,180 @@
...
+	t.Helper()
+	var (
+		chainParams  = &chaincfg.Params{}
```
guggero, line 31: A package global variable called `chainParams` is already defined, can just use that one.

chain/mocks_test.go
```diff
@@ -0,0 +1,180 @@
...
+	return nil
+}
+
+type mockChainService struct{}
+
+func (m *mockChainService) Start() error {
+	return nil
+}
+
+func (m *mockChainService) GetBlock(
```
guggero, line 89: nit: formatting.

chain/neutrino.go
```diff
@@ -755,3 +872,36 @@ out:
 	close(s.dequeueNotification)
 	s.wg.Done()
 }
+
+// getNewRescanner injects the Rescanner constructor when called and defaults
+// to using neutrino.NewRescan when unspecified.
+func (s *NeutrinoClient) getNewRescanner() NewRescanner {
+	if s.newRescanner == nil {
+		s.newRescanner = func(ropts ...neutrino.RescanOption) NeutrinoRescanner {
```
guggero, line 880: nit: line length (tab width should be set to 8 spaces).

chain/neutrino.go
```diff
@@ -335,55 +386,65 @@ func (s *NeutrinoClient) pollCFilter(hash *chainhash.Hash) (*gcs.Filter, error)
...
-	// If the wallet is already fully caught up, or the rescan has started
-	// with state that indicates a "fresh" wallet, we'll send a
-	// notification indicating the rescan has "finished".
-	if header.BlockHash() == *startHash {
-		s.clientMtx.Lock()
-		s.finished = true
-		rescanQuit := s.rescanQuit
-		s.clientMtx.Unlock()
+	select {
+	case rescanner := <-s.rescannerCh:
```
guggero, line 425: All these changes to the rescan logic are very hard to reason about for a reviewer. Is there any way to split this big commit up into small pieces that logically build upon each other to help understand all the individual changes?

guggero (changes requested):
Thanks a lot for your work on this! The rescan logic is definitely something that needs to be cleaned up.
But I have to admit, this is very hard to review. All the changes are in a single commit so it's almost impossible to reason about everything at the same time. Even with the new tests added that makes it very hard to make sure no dead lock or new data races are introduced with the changes. And that the formatting doesn't match the rest of the code base doesn't really help the review either.

Can you please split things up a bit more for easier review? Thanks!

MStreet3:
Converting to draft state to address comments from @guggero

MStreet3:
See #822

PR #822, opened 2022-10-26
chain: fix NeutrinoClient segfault on NotifyReceived call

### LINKED ISSUE #819: NeutrinoClient data race

MStreet3:
Fixes #819
Replaces #820

### Background

There is datarace in the neutrino client implementation that is causing test
flakes in `lnd`. The shared `rescan` object and the granular
locking strategy allow for the possibility of a segmentation fault when
`NotifyReceived` attempts to call `Update` on a `rescan` that is `nil`.  This
is because the `clientMtx` lock is released after `rescan` is set to `nil`.

### Changes

The solution here is to be minimally invasive.  Add a new `rescanMtx` mutex
and require that the `NotifyReceived` method and the `Rescan` method
hold the `rescanMtx` lock for the duration of their execution.  Locking and
unlocking of the client is unaffected and changes to the `rescan` object
become atomic.

### Testing

This PR adds two interfaces `rescanner` and `nutrinoChainService`
to abstract out the dependency on structs from the `neutrino` package.

This PR adds a file of mock implementations for these interfaces and uses them
in a new unit test file.

### DIFF
--- a/chain/chainservice.go
+++ b/chain/chainservice.go
@@ -1,0 +1,40 @@
+package chain
+
+import (
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/gcs"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/neutrino"
+	"github.com/lightninglabs/neutrino/banman"
+	"github.com/lightninglabs/neutrino/headerfs"
+)
+
+// NeutrinoChainService is an interface that encapsulates all the public
+// methods of a *neutrino.ChainService
+type NeutrinoChainService interface {
+	Start() error
+	GetBlock(chainhash.Hash, ...neutrino.QueryOption) (*btcutil.Block, error)
+	GetBlockHeight(*chainhash.Hash) (int32, error)
+	BestBlock() (*headerfs.BlockStamp, error)
+	GetBlockHash(int64) (*chainhash.Hash, error)
+	GetBlockHeader(*chainhash.Hash) (*wire.BlockHeader, error)
+	IsCurrent() bool
+	SendTransaction(*wire.MsgTx) error
+	GetCFilter(chainhash.Hash, wire.FilterType,
+		...neutrino.QueryOption) (*gcs.Filter, error)
+	GetUtxo(...neutrino.RescanOption) (*neutrino.SpendReport, error)
+	BanPeer(string, banman.Reason) error
+	IsBanned(addr string) bool
+	AddPeer(*neutrino.ServerPeer)
+	AddBytesSent(uint64)
+	AddBytesReceived(uint64)
+	NetTotals() (uint64, uint64)
+	UpdatePeerHeights(*chainhash.Hash, int32, *neutrino.ServerPeer)
+	ChainParams() chaincfg.Params
+	Stop() error
+	PeerByAddr(string) *neutrino.ServerPeer
+}
+
+var _ NeutrinoChainService = (*neutrino.ChainService)(nil)
--- a/chain/mocks_test.go
+++ b/chain/mocks_test.go
@@ -1,0 +1,157 @@
+package chain
+
+import (
+	"container/list"
+	"errors"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/gcs"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/lightninglabs/neutrino"
+	"github.com/lightninglabs/neutrino/banman"
+	"github.com/lightninglabs/neutrino/headerfs"
+)
+
+var (
+	errNotImplemented = errors.New("not implemented")
+	testBestBlock     = &headerfs.BlockStamp{
+		Height: 42,
+	}
+)
+
+var (
+	_ rescanner            = (*mockRescanner)(nil)
+	_ NeutrinoChainService = (*mockChainService)(nil)
+)
+
+// newMockNeutrinoClient constructs a neutrino client with a mock chain
+// service implementation and mock rescanner interface implementation.
+func newMockNeutrinoClient() *NeutrinoClient {
+	// newRescanFunc returns a mockRescanner
+	newRescanFunc := func(ro ...neutrino.RescanOption) rescanner {
+		return &mockRescanner{
+			updateArgs: list.New(),
+		}
+	}
+
+	return &NeutrinoClient{
+		CS:        &mockChainService{},
+		newRescan: newRescanFunc,
+	}
+}
+
+// mockRescanner is a mock implementation of a rescanner interface for use in
+// tests.  Only the Update method is implemented.
+type mockRescanner struct {
+	updateArgs *list.List
+}
+
+func (m *mockRescanner) Update(opts ...neutrino.UpdateOption) error {
+	m.updateArgs.PushBack(opts)
+	return nil
+}
+
+func (m *mockRescanner) Start() <-chan error {
+	return nil
+}
+
+func (m *mockRescanner) WaitForShutdown() {
+	// no-op
+}
+
+// mockChainService is a mock implementation of a chain service for use in
+// tests.  Only the Start, GetBlockHeader and BestBlock methods are implemented.
+type mockChainService struct {
+}
+
+func (m *mockChainService) Start() error {
+	return nil
+}
+
+func (m *mockChainService) BestBlock() (*headerfs.BlockStamp, error) {
+	return testBestBlock, nil
+}
+
+func (m *mockChainService) GetBlockHeader(
+	*chainhash.Hash) (*wire.BlockHeader, error) {
+
+	return &wire.BlockHeader{}, nil
+}
+
+func (m *mockChainService) GetBlock(chainhash.Hash,
+	...neutrino.QueryOption) (*btcutil.Block, error) {
+
+	return nil, errNotImplemented
+}
+
+func (m *mockChainService) GetBlockHeight(*chainhash.Hash) (int32, error) {
+	return 0, errNotImplemented
+}
+
+func (m *mockChainService) GetBlockHash(int64) (*chainhash.Hash, error) {
+	return nil, errNotImplemented
+}
+
+func (m *mockChainService) IsCurrent() bool {
+	return false
+}
+
+func (m *mockChainService) SendTransaction(*wire.MsgTx) error {
+	return errNotImplemented
+}
+
+func (m *mockChainService) GetCFilter(chainhash.Hash,
+	wire.FilterType, ...neutrino.QueryOption) (*gcs.Filter, error) {
+
+	return nil, errNotImplemented
+}
+
+func (m *mockChainService) GetUtxo(
+	_ ...neutrino.RescanOption) (*neutrino.SpendReport, error) {
+
+	return nil, errNotImplemented
+}
+
+func (m *mockChainService) BanPeer(string, banman.Reason) error {
+	return errNotImplemented
+}
+
+func (m *mockChainService) IsBanned(addr string) bool {
+	panic(errNotImplemented)
+}
+
+func (m *mockChainService) AddPeer(*neutrino.ServerPeer) {
+	panic(errNotImplemented)
+}
+
+func (m *mockChainService) AddBytesSent(uint64) {
+	panic(errNotImplemented)
+}
+
+func (m *mockChainService) AddBytesReceived(uint64) {
+	panic(errNotImplemented)
+}
+
+func (m *mockChainService) NetTotals() (uint64, uint64) {
+	panic(errNotImplemented)
+}
+
+func (m *mockChainService) UpdatePeerHeights(*chainhash.Hash,
+	int32, *neutrino.ServerPeer,
+) {
+	panic(errNotImplemented)
+}
+
+func (m *mockChainService) ChainParams() chaincfg.Params {
+	panic(errNotImplemented)
+}
+
+func (m *mockChainService) Stop() error {
+	panic(errNotImplemented)
+}
+
+func (m *mockChainService) PeerByAddr(string) *neutrino.ServerPeer {
+	panic(errNotImplemented)
+}
--- a/chain/neutrino.go
+++ b/chain/neutrino.go
@@ -20,9 +20,9 @@ import (
 	"github.com/lightninglabs/neutrino/headerfs"
 )
 
-// NeutrinoClient is an implementation of the btcwalet chain.Interface interface.
+// NeutrinoClient is an implementation of the btcwallet chain.Interface interface.
 type NeutrinoClient struct {
-	CS *neutrino.ChainService
+	CS NeutrinoChainService
 
 	chainParams *chaincfg.Params
 
--- a/chain/neutrino.go
+++ b/chain/neutrino.go
@@ -27,8 +27,14 @@ type NeutrinoClient struct {
 	chainParams *chaincfg.Params
 
 	// We currently support only one rescan/notification goroutine per client.
+	// Therefore there can only be one instance of the rescan object and
+	// the rescanMtx synchronizes its access.  Calls to the NotifyReceived
+	// and Rescan methods of the client must hold the rescan mutex lock for
+	// the length of their execution to ensure that all operations that
+	// affect the rescan object are atomic.
 	rescan    rescanner
 	newRescan newRescanFunc
+	rescanMtx sync.Mutex
 
 	enqueueNotification     chan interface{}
 	dequeueNotification     chan interface{}
@@ -46,6 +52,14 @@ type NeutrinoClient struct {
 	finished   bool
 	isRescan   bool
 
+	// The clientMtx synchronizes access to the state variables of the client.
+	//
+	// TODO(mstreet3): Currently the clientMtx synchronizes access to the
+	// rescanQuit and rescanErr channels, which cancel the current rescan
+	// goroutine when closed and is updated each time a new rescan goroutine
+	// is created, respectively.  All state related to the rescan goroutine
+	// should ideally be synchronized by the same lock or via some other
+	// shared mechanism.
 	clientMtx sync.Mutex
 }
 
@@ -351,6 +365,10 @@ func (s *NeutrinoClient) pollCFilter(hash *chainhash.Hash) (*gcs.Filter, error) 
 func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Address,
 	outPoints map[wire.OutPoint]btcutil.Address) error {
 
+	// Obtain and hold the rescan mutex lock for the duration of the call.
+	s.rescanMtx.Lock()
+	defer s.rescanMtx.Unlock()
+
 	s.clientMtx.Lock()
 	if !s.started {
 		s.clientMtx.Unlock()
@@ -458,6 +476,10 @@ func (s *NeutrinoClient) NotifyBlocks() error {
 
 // NotifyReceived replicates the RPC client's NotifyReceived command.
 func (s *NeutrinoClient) NotifyReceived(addrs []btcutil.Address) error {
+	// Obtain and hold the rescan mutex lock for the duration of the call.
+	s.rescanMtx.Lock()
+	defer s.rescanMtx.Unlock()
+
 	s.clientMtx.Lock()
 
 	// If we have a rescan running, we just need to add the appropriate
--- a/chain/neutrino.go
+++ b/chain/neutrino.go
@@ -100,24 +100,36 @@ func (s *NeutrinoClient) Start() error {
 	s.clientMtx.Lock()
 	defer s.clientMtx.Unlock()
 	if !s.started {
+		// Reset the client state.
 		s.enqueueNotification = make(chan interface{})
 		s.dequeueNotification = make(chan interface{})
 		s.currentBlock = make(chan *waddrmgr.BlockStamp)
 		s.quit = make(chan struct{})
 		s.started = true
+
+		// Go place a ClientConnected notification onto the queue.
 		s.wg.Add(1)
 		go func() {
+			defer s.wg.Done()
+
 			select {
 			case s.enqueueNotification <- ClientConnected{}:
 			case <-s.quit:
 			}
 		}()
+
+		// Go launch the notification handler.
+		s.wg.Add(1)
 		go s.notificationHandler()
 	}
 	return nil
 }
 
 // Stop replicates the RPC client's Stop method.
+//
+// TODO(mstreet3): The Stop method does not cancel the long-running rescan
+// goroutine.  This is a memory leak.  Stop should shutdown the rescan goroutine
+// and reset the scanning state of the NeutrinoClient to false.
 func (s *NeutrinoClient) Stop() {
 	s.clientMtx.Lock()
 	defer s.clientMtx.Unlock()
--- a/chain/neutrino.go
+++ b/chain/neutrino.go
@@ -26,8 +26,9 @@ type NeutrinoClient struct {
 
 	chainParams *chaincfg.Params
 
-	// We currently support one rescan/notifiction goroutine per client
-	rescan *neutrino.Rescan
+	// We currently support only one rescan/notification goroutine per client.
+	rescan    rescanner
+	newRescan newRescanFunc
 
 	enqueueNotification     chan interface{}
 	dequeueNotification     chan interface{}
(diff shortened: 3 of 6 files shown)


chain/internal/rescan/rescan.go
```diff
@@ -0,0 +1,20 @@
+package rescan
+
+import "github.com/lightninglabs/neutrino"
+
+type Updater interface {
```
guggero, line 7: Nice to extract those out into an interface. Though I don't think we need an extra sub package. If we want to make them internal, we just don't export them (so `rescanUpdater`, `rescanStarter`, `rescan` as names).

Also, all interfaces should get Godoc comments to explain what they represent.

chain/neutrino.go
```diff
@@ -14,6 +14,7 @@ import (
 	"github.com/btcsuite/btcd/rpcclient"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/chain/internal/rescan"
```
guggero, line 17: Importing this package as `rescan` will cause some variables that are called the same to become shadowing. One more reason to not add those interfaces to a new sub package.

chain/neutrino.go
```diff
@@ -48,6 +48,8 @@ type NeutrinoChainService interface {
 	PeerByAddr(string) *neutrino.ServerPeer
 }
 
+var _ NeutrinoChainService = (*neutrino.ChainService)(nil)
```
guggero: This can be done in the commit that introduces the interface.

chain/mocks_test.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/chain/internal/rescan"
+	"github.com/lightninglabs/neutrino"
+	"github.com/lightninglabs/neutrino/banman"
+	"github.com/lightninglabs/neutrino/headerfs"
+)
+
+// Define variables for mocks
```
guggero, line 19: nit: incomplete comments in several places of the file (missing punctuation). Also, this comment just repeats the code below and doesn't really add any useful information.

chain/mocks_test.go
```diff
@@ -0,0 +1,152 @@
...
+)
+
+// Define static type checks
+var (
+	_ rescan.Interface     = (*mockRescanner)(nil)
+	_ NeutrinoChainService = (*mockChainService)(nil)
+)
+
+func newMockNeutrinoClient(t *testing.T,
+	opts ...func(*mockRescanner)) *NeutrinoClient {
```
guggero, line 36: nit: add a newline after a multi-line function definition. We try to adhere to [`lnd`'s code formatting rules](https://github.com/lightningnetwork/lnd/blob/master/docs/code_formatting_rules.md) here as well, at least for new code.

chain/neutrino.go
```diff
@@ -785,3 +782,17 @@ out:
 	close(s.dequeueNotification)
 	s.wg.Done()
 }
+
+// getNewRescanner injects the Rescanner constructor when called and defaults
+// to using neutrino.NewRescan when unspecified.
+func (s *NeutrinoClient) getNewRescanner() rescan.New {
```
guggero: I think this function should be passed in when creating the client. See my main comment with the proposed diff.

chain/neutrino_test.go
```diff
@@ -76,3 +77,71 @@ func TestNeutrinoClientNotifyReceived(t *testing.T) {
...
+	err := nc.Start()
+	require.NoError(t, err)
+
+	// Launch the wanted number of goroutines, wait for them to finish and
+	// signal all done.
+	wg.Add(wantRoutines)
+	go func() {
+		defer close(done)
+		defer wg.Wait()
+		for i := 0; i < wantRoutines; i++ {
```
guggero, line 130: Very nice to have more test coverage for the concurrent code!

chain/neutrino.go
```diff
@@ -77,6 +77,7 @@ type NeutrinoClient struct {
 	isRescan   bool
 
 	clientMtx sync.Mutex
+	rescanMtx sync.Mutex
```
guggero, line 81: Can you add a Godoc comment here describing what variables exactly this mutex is guarding?

chain/neutrino.go
```diff
@@ -105,18 +105,25 @@ func (s *NeutrinoClient) Start() error {
...
 		s.enqueueNotification = make(chan interface{})
 		s.dequeueNotification = make(chan interface{})
 		s.currentBlock = make(chan *waddrmgr.BlockStamp)
 		s.quit = make(chan struct{})
 		s.started = true
+
+		// Go place a ClientConnected notification onto the queue.
 		s.wg.Add(1)
 		go func() {
+			defer s.wg.Done()
```
guggero, line 118: Nice catch!

Can you add an empty line after the `defer` call to visually separate it better from the `select` statement?

chain/internal/rescan/rescan.go
```diff
@@ -18,7 +18,11 @@ type Interface interface {
 	Starter
 	Updater
 
+	// WaitForShutdown blocks until the underlying rescan object is shutdown.
```
guggero, line 21: This commit should be squashed with the commit that introduced the code. A PR should be structured in a way that explains to the reviewer how the pieces fit together, not what versions of changes each file went through during development. So things like renames or adding comments to code that is introduced in the same PR should always be squashed with the commit that introduces the code. Otherwise a reviewer needs to keep all different names in memory, which makes it harder.

chain/neutrino_test.go
```diff
@@ -14,7 +14,8 @@ import (
 // can sequentially Start and Stop without errors or races.
 func TestNeutrinoClientSequentialStartStop(t *testing.T) {
 	var (
-		ctx, cancel   = context.WithTimeout(context.Background(), 1*time.Second)
+		ctx, cancel = context.WithTimeout(context.Background(),
```
guggero, line 18: nit: formatting of multi-line function call.

chain/neutrino.go
```diff
@@ -797,16 +797,22 @@ out:
...
-	if s.newRescanner == nil {
-		s.newRescanner = func(ropts ...neutrino.RescanOption) rescan.Interface {
-			cs := &neutrino.RescanChainSource{
-				ChainService: s.CS.(*neutrino.ChainService),
-			}
-			return neutrino.NewRescan(cs, ropts...)
+func (s *NeutrinoClient) getNewRescanFunc() rescan.NewFunc {
+	if s.newRescanFunc == nil {
+		// Must cast CS as *neutrino.ChainService
+		cs := s.CS.(*neutrino.ChainService)
```
guggero: This cast can be avoided when we instead pass in the whole function that creates a rescan (see my diff).

chain/neutrino.go
```diff
@@ -107,14 +107,12 @@ func (s *NeutrinoClient) Start() error {
 	s.clientMtx.Lock()
 	defer s.clientMtx.Unlock()
 	if !s.started {
-		// Reset the client state.
```
guggero: Why are we removing those comments again?

guggero (commented):
Thank you for re-structuring the PR! I left a few suggestions for how to improve it even more for easier review.
But I was able to understand all the changes, nice work! Sorry it took so long to get to the first pass review.

I think we can further reduce the diff by passing in the `newRescanFunc` in the `NewNeutrinoClient`. That removes the need for any type casting or having two versions of the `cs` member around. See my diff with the proposed changes (might need some additional cleanup, this is just to get my idea across):

[pr-822.diff.txt](https://github.com/btcsuite/btcwallet/files/10031471/pr-822.diff.txt)

chain/neutrino.go
MStreet3: I added these earlier when fixing the race condition in start/stop and removed them in this commit just to make the diff against master smaller, wasn't sure if they were useful enough to be kept.  I can add them back. 👍🏾

chain/internal/rescan/rescan.go
MStreet3: noted, thanks

chain/neutrino.go
```diff
@@ -356,6 +362,10 @@ func (s *NeutrinoClient) pollCFilter(hash *chainhash.Hash) (*gcs.Filter, error)
 func (s *NeutrinoClient) Rescan(startHash *chainhash.Hash, addrs []btcutil.Address,
 	outPoints map[wire.OutPoint]btcutil.Address) error {
 
+	// Obtain and hold the rescan mutex lock for the duration of the call.
+	s.rescanMtx.Lock()
```
guggero, line 374: Very nice and straightforward solution!

guggero (commented):
Very nice, only nits left :tada: Thanks for all your work on this!

guggero:
@ellemouton could you take a look at this as well please?

chain/neutrino.go
MStreet3: Thanks 🙏🏾

chappjc (commented):
The `rescanMtx` resolution seems fine, but this PR introduces some notable and breaking API changes and abstraction, which create some challenges and confusion for consumers.

The abstraction seems unnecessary, except for test stubbing, but even then I don't think the constructors and exported types need to change. Namely, the test code can and does set the fields directly with its mock stubs (without using the constructors), since e.g. `newMockNeutrinoClient` manually creates the `NeutrinoClient`, and the newly added `NewRescanFunc` isn't used.  Could the exported constructor, which consumers like btcwallet use, deal with setting up the rescanner fields from the provided chain service?

This is only from a cursory scan, so I apologize in advance if my assessment is off base, but I don't think this improves the API in its present state, and I suspect the goals of testing with a mock chain service can still be achieved without the API changes.

chain/neutrino.go
```diff
@@ -45,20 +52,45 @@ type NeutrinoClient struct {
...
+	// shared mechanism.
 	clientMtx sync.Mutex
 }
 
 // NewNeutrinoClient creates a new NeutrinoClient struct with a backing
 // ChainService.
 func NewNeutrinoClient(chainParams *chaincfg.Params,
-	chainService *neutrino.ChainService) *NeutrinoClient {
+	chainService neutrinoChainService,
+	newRescan newRescanFunc) *NeutrinoClient {
```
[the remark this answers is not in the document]
MStreet3: I see your point about changing the NewNeutrinoClient constructor. It's feasible to let this constructor define the rescan constructor internally. 

I don't see an issue with exporting the neutrinoChainService interface. I left it private assuming all consumers would just use `*neutrino.ChainService` (I.e., no other concrete implementations exist or plan to exist).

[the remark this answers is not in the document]
chappjc: > I left it private assuming all consumers would just use `*neutrino.ChainService` (I.e., no other concrete implementations exist or plan to exist).

That's what I gathered.  In that case, I don't see a reason to use the interface in the constructor's input args.  Just changing the `CS` field type would suffice since the tests aren't using the constructor anyway. Even that is a breaking API change, but not as disruptive I think since that field probably shouldn't have been exported anyway.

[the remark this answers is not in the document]
MStreet3: Right right, let me give it a try. Would be good to reduce the surface area of the change.

chappjc:
@MStreet3 Thanks for restoring the `NewNeutrinoClient` function signature in https://github.com/btcsuite/btcwallet/compare/a175d3fd326a898cf5cd4c16a01cd32860fbddd6..a6507c0ba3d9b62d563df9fb51c6a6c956f53edc.  That gives me great relief.  Ideally there would be no breaking changes, but I don't expect that `CS` field type change to have a great impact.

My only other comments are mainly stylistic, and that's not my place.

One strange (existing) thing I noticed in neutrino.go is that `s.scanning` is never (re)set to false, I think.  That seems to permanently affect the flow of `NotifyBlocks` and `NotifyReceived` if a rescan hand been done and completed at any point.  Does that seem intentional?  The multitude of undocumented `bool`s (`started`, `scanning`, `finished`, `isRescan`) are confusing.

MStreet3:
> One strange (existing) thing I noticed in neutrino.go is that `s.scanning` is never (re)set to false, I think.  That seems to permanently affect the flow of `NotifyBlocks` and `NotifyReceived` if a rescan hand been done and completed at any point.  Does that seem intentional?  The multitude of undocumented `bool`s (`started`, `scanning`, `finished`, `isRescan`) are confusing.

I think that's intentional. `scanning` just means that the rescan goroutine is listening. Calls to update actually send some new options that trigger a "rescan". Once the rescan goroutine is running, it's not torn down except in calls to the Rescan method, but that always creates a new routine. I don't have enough context to advocate for or against the setup. This is why the fix for the seg fault here is intended be small. 

The Stop method of the neutrino client should stop the rescan goroutine but I don't think it's caused any issues yet to warrant a PR.

chain/rescan.go
```diff
@@ -0,0 +1,42 @@
+package chain
+
+import "github.com/lightninglabs/neutrino"
+
+var _ rescanner = (*neutrino.Rescan)(nil)
+
+// updater is the interface that wraps the Update method of a rescan object.
+
+// Update targets a long-running rescan/notification client with updateable
```
guggero, line 9: The Godoc comment for the `Update` method should be right above the method definition. So it's duplicated now, can just move this to replace what's on L14. Same for the `Start()` method of the `starter`.

chain/mocks_test.go
```diff
@@ -0,0 +1,171 @@
...
+	// newRescanFunc returns a mockRescanner
+	newRescanFunc := func(ro ...neutrino.RescanOption) rescanner {
+		return &mockRescanner{
+			updateArgs: list.New(),
+		}
+	}
+
+	return &NeutrinoClient{
+		CS:        &mockChainService{},
+		newRescan: newRescanFunc,
```
guggero, line 41: Ah, this works too, nice. Not breaking the interface is definitely a plus (though it wouldn't have mattered to our use case too much as we'd need to update `lnd` anyway).

chain/neutrino_test.go
```diff
@@ -0,0 +1,190 @@
...
+	// channel.
+	executeCalls := func(n int) <-chan struct{} {
+		var (
+			done = make(chan struct{})
+			wg   sync.WaitGroup
+		)
+
+		go func() {
+			defer close(done)
+			defer wg.Wait()
```
guggero, line 157: I think the `wg.Wait()` should come before the `close(done)`, otherwise the `select` below will return even though the wait might have still blocked. Maybe the wait should be right after the loop instead?

chain/neutrino.go
```diff
@@ -46,6 +52,13 @@ type NeutrinoClient struct {
 	finished   bool
 	isRescan   bool
 
+	// The clientMtx synchronizes access to the state variables of the client.
+	// TODO(mstreet3): Currently the clientMtx synchronizes access to the
```
guggero, line 56: nit: add an empty comment line before a TODO so it stands out a bit more.

guggero (commented):
Nice update, this is almost ready IMO :tada:

chain/neutrino_test.go
MStreet3: defer statements are executed from the bottom up, we wait and then close done.

chain/mocks_test.go
```diff
@@ -0,0 +1,171 @@
...
+
+// mockRescanner is a mock implementation of a rescanner interface for use in
+// tests.  Only the Update method is implemented.
+type mockRescanner struct {
+	updateArgs *list.List
+}
+
+func (m *mockRescanner) Start() <-chan error {
+	errs := make(chan error)
+	return errs
```
chappjc, line 53: `return make(chan error)`?

chain/neutrino_test.go
```diff
@@ -0,0 +1,209 @@
...
+	"github.com/btcsuite/btcd/wire"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestNeutrinoClientSequentialStartStop ensures that the client
+// can sequentially Start and Stop without errors or races.
+func TestNeutrinoClientSequentialStartStop(t *testing.T) {
+	var (
+		timeout      = time.After(1 * time.Second)
```
chappjc, line 19: CI can be very laggy, almost as if the VMs randomly pause, but if 1sec works, ok.  I'd be surprised if that's not often too short.  Would it hurt to go with 5 or even 10 seconds?  It only slows a test that would fail, not a passing one.

chain/neutrino_test.go
```diff
@@ -0,0 +1,209 @@
...
+
+	// Define closures to wrap desired neutrino client method calls.
+
+	// callRescan calls the Rescan() method and asserts it completes
+	// with no errors. Rescan() is called with the hash of an empty header
+	// on each call.
+	var (
+		emptyHeader = &wire.BlockHeader{}
+		startHash   = emptyHeader.BlockHash()
+	)
```
chappjc, line 123: You'd be fine with just `startHash := new(wire.BlockHeader).BlockHash()` and no `emptyHeader` declared, but NBD.

chappjc (commented):
LGTM, just a few mostly style related comments.
Have yet to actually test in our app, but I see no obvious issues with the interface unchanged.

chain/chainservice.go
```diff
@@ -0,0 +1,40 @@
...
+	GetBlockHeight(*chainhash.Hash) (int32, error)
+	BestBlock() (*headerfs.BlockStamp, error)
+	GetBlockHash(int64) (*chainhash.Hash, error)
+	GetBlockHeader(*chainhash.Hash) (*wire.BlockHeader, error)
+	IsCurrent() bool
+	SendTransaction(*wire.MsgTx) error
+	GetCFilter(chainhash.Hash, wire.FilterType,
+		...neutrino.QueryOption) (*gcs.Filter, error)
+	GetUtxo(...neutrino.RescanOption) (*neutrino.SpendReport, error)
+	BanPeer(string, banman.Reason) error
```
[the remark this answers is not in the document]
MStreet3: yes, so the first time I implemented this it was much slimmer, but I saw that this created a failure in LND.  This is because LND has some code that is directly accessing the `CS` field of a `NeutrinoClient`.  The consumer in LND expected the `CS` field to be the single `*neutrino.ChainService` concrete implementation, but in reality it was the slim interface.  So to be able to allow LND to update the version of `btcwallet` with this change I had to expose EVERY public method of `*neutrino.ChainService` event though a `NeutrinoClient` _only_ uses a few of the methods.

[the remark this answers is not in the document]
chappjc: Ohhhh, ok.  That's interesting.

I'd just note that lnd can assert that CS satisfies an even broader interface itself (the assertion applies to the underlying type, which can have more methods).  It seems like the lnd-btcwallet relationship is quite tight.  This partly explains why `CS` is even exported.

chain/rescan.go
```diff
@@ -0,0 +1,41 @@
...
+	starter
+	updater
+
+	// WaitForShutdown blocks until the underlying rescan object is shutdown.
+	// Close the quit channel before calling WaitForShutdown.
+	WaitForShutdown()
+}
+
+// updater is the interface that wraps the Update method of a rescan object.
+
```
guggero, line 20: nit: extra newline here and below.

chain/neutrino_test.go
```diff
@@ -0,0 +1,219 @@
...
+	timeout := time.After(maxDur)
+	select {
+	case <-timeout:
+		t.Fatal("timed out")
+	case <-executeCalls(wantMsgs):
+		// Ensure that exactly wantRoutines number of calls were made
+		// by counting the results on the message channel.
+		close(msgCh)
+		for str := range msgCh {
+			assert.NotNil(t, str)
```
guggero, line 212: A string cannot be nil, so the `Contains` below is enough IMO.

guggero (approved):
Very nice, LGTM :tada:

guggero (approved):
Thanks a lot for all the quick updates on this. LGTM :tada: 
Before we merge this, can you please update (or create?) the PR on the `lnd` side to make sure all neutrino integration tests still pass?

MStreet3:
@guggero the corresponding LND PR with updated btcwallet is here: https://github.com/lightningnetwork/lnd/pull/7049

ellemouton (approved):
Very nice! Love the new abstractions that make things way more testable. Great job!

guggero:
@MStreet3 can you do a rebase please?
Also, @chappjc could I ask you for another round of review please, since you did a first round a while ago.

chappjc (approved):
Still builds and runs fine in my use case.

With limited familiarity with btcwallet's code, the atomicity changes look fine too.  I don't see any lock ordering pitfalls that could lead to deadlock, but it's a risk of nested locking.

I think see some ancient data races on the `rescanQuit` field in the notification handlers like `onBlockConnected`, but that's unrelated to the issue being solved in this PR.

PR #823, opened 2022-10-26
Bump btcd version to v0.23.2

vitsalis:
Version v0.23.2 introduces a [bugfix](https://github.com/btcsuite/btcd/pull/1896) for script witness item sizes. Without this fix, btcwallet fails to handle the testnet block at height 2350076.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,7 +1,7 @@
 module github.com/btcsuite/btcwallet
 
 require (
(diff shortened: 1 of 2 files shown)


guggero (approved):

Roasbeef:
Can you run `go mod tidy`? Then this way the `go.sum` will be updated.

vitsalis:
Sure, done.

PR #825, opened 2022-11-01
waddrmgr:  add new ValidatableManagedAddress for fault detection during addr/key derivation 

Roasbeef:
In this commit, we add a new `ValidatableManagedAddress` which is intended to catch potential faults during the addr/key derivation process. Examples of such faults include: hardware issues, faulty memory, and cosmic ray bit flips. 

Once we generate a new address, we'll call the `Validate` method that performs a series of sanity check including: the pubkey in m emory matches the one derived from the private key, the key derives the same addr in memory, we can generate a valid signature using the private key validated against the public key in memory. 

We'll also _re derive_ the original child key (from the branch) and perform the same checks above once again.

Finally, at the very end, we'l re-read the address from disk and ensure that it matches the address we're holding in memory. 

This changed was inspired by this bitcoind PR: https://github.com/bitcoin/bitcoin/pull/5224


waddrmgr/manager_test.go
```diff
@@ -3137,3 +3137,148 @@ func assertAddressDerivation(t *testing.T, db walletdb.DB,
...
+		},
+		{
+			// In this final case, we'll have the signature generator
+			// return a signature that's actually invalid for the public
+			// key.
+			name: "wrong priv key",
+			mutator: func(addr *managedAddress, priv *btcec.PrivateKey) {
+				// To trigger an invalid private key, we'll
+				// modify the internal decrypted private key.
+				addr.privKeyCT[0] ^= 0x01
```
guggero, line 3222: Ah, I see why you used the `privKeyCT` directly before. But I think this test should still work the same when using `PrivKey()` in `Validate()`.

guggero (commented):
Very nice improvement :100:

waddrmgr/address.go
```diff
@@ -402,6 +420,92 @@ func (a *managedAddress) DerivationInfo() (KeyScope, DerivationPath, bool) {
...
+
+	// For the newer taproot addr type, we'll generate a schnorr signature
+	// to verify against.
+	case TaprootPubKey:
+		sig, err = schnorr.Sign(addrPrivKey, msg[:])
+		if err != nil {
+			return fmt.Errorf("unable to generate validate "+
+				"schnorr sig: %v", err)
+		}
+	}
```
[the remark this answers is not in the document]
Roasbeef: Hmm, sure, this way at least it'll return an error and prevent new addresses from being created if we don't know how to validate them. There's actually a lingering panic here (or maybe just verify against zero value sig) if it gets thru w/o hitting either of the existing cases.

waddrmgr/address.go
```diff
@@ -402,6 +420,92 @@ func (a *managedAddress) DerivationInfo() (KeyScope, DerivationPath, bool) {
...
+
+	// With the two above checks, we'll now execute our final sanity check:
+	// make sure we can generate a signature that verifies under the target
+	// public key.
+	//
+	// TODO(roasbeef): potentially run _all_ checks then see which one
+	// fails?
+	var sig signature
+
+	addrPrivKey, _ := btcec.PrivKeyFromBytes(a.privKeyCT)
```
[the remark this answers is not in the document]
Roasbeef: > I think we should use a.PrivKey() here instead, in case the private key hasn't been decrypted yet?

So in this case (at least for lnd), the wallet is always unlocked and the key decrypted. I had it using `PrivKey`, but there's a test where it makes a "raw watch only" addr w/o unlocking the wallet. In this case, a call to `PrivKey()` fail since the wallet isn't unlocked. It'll actually deadlock since it needs a mutex (had made a private func w/o a lock to test that out). 

Also this is only called from `newManagedAddress`, which _always_ have the private key available. See where `privKeyCT is set for the first time.

waddrmgr/manager_test.go
Roasbeef: See above re the issue w/ `PrivKey` and one of the tests. 

I also had things flipped: I'd use the passed in private key to generate a signature against the public key of the addr. However, that made it hard to write a tests here as modifying the passed in private key would actually cause the earlier tests to fail. 

One way to get around this would be to abstract out the calls to `Sign` behind some interface, then have that pass a bogus signature.

Roasbeef:
Thx for the review @guggero! Revised things a bit, PTAL.

Roasbeef:
cc @sputn1ck

waddrmgr/address.go
```diff
@@ -402,6 +420,92 @@ func (a *managedAddress) DerivationInfo() (KeyScope, DerivationPath, bool) {
...
+	// to verify against.
+	case TaprootPubKey:
+		sig, err = schnorr.Sign(addrPrivKey, msg[:])
+		if err != nil {
+			return fmt.Errorf("unable to generate validate "+
+				"schnorr sig: %v", err)
+		}
+
+	default:
+		fmt.Errorf("unable to validate addr, unknown type: %v",
```
guggero, line 499: Missing the `return` keyword here!

waddrmgr/address.go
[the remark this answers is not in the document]
guggero: Ah, I see. Makes sense given the context of the lock and how we use it. I wonder if we should export the `Validate()` method at all then? Since it makes these assumptions in the first place.

waddrmgr/address.go
```diff
@@ -402,6 +420,92 @@ func (a *managedAddress) DerivationInfo() (KeyScope, DerivationPath, bool) {
...
+func (a *managedAddress) Validate(msg [32]byte, priv *btcec.PrivateKey) error {
+	// First, we'll obtain the mapping public key from the target private
+	// key. This key should match up with the public key we store
+	// internally.
+	//
+	// This can potentially catch an error in the original mapping of the
+	// private key to a public key.
+	basePubKey := priv.PubKey()
+	if !a.pubKey.IsEqual(basePubKey) {
+		return fmt.Errorf("%w: expected %v, got %v", ErrPubKeyMismatch,
```
sputn1ck, line 451: nit: use `%x`

waddrmgr/address.go
```diff
@@ -402,6 +420,92 @@ func (a *managedAddress) DerivationInfo() (KeyScope, DerivationPath, bool) {
...
+	// an address of the corresponding type, that matches up w/ what we've
+	// stored internally.
+	//
+	// This can potentially catch a hardware/software error when mapping
+	// the public key to a Bitcoin address.
+	addr, err := newManagedAddressWithoutPrivKey(
+		a.manager, a.derivationPath, a.pubKey, a.compressed, a.addrType,
+	)
+	if err != nil {
+		return fmt.Errorf("unable to re-create addr: %v", err)
```
sputn1ck, line 466: nit: `%w`

waddrmgr/address.go
```diff
@@ -402,6 +420,92 @@ func (a *managedAddress) DerivationInfo() (KeyScope, DerivationPath, bool) {
...
+	// verify against.
+	case NestedWitnessPubKey, PubKeyHash, WitnessPubKey:
+		sig = ecdsa.Sign(addrPrivKey, msg[:])
+
+	// For the newer taproot addr type, we'll generate a schnorr signature
+	// to verify against.
+	case TaprootPubKey:
+		sig, err = schnorr.Sign(addrPrivKey, msg[:])
+		if err != nil {
+			return fmt.Errorf("unable to generate validate "+
```
sputn1ck, line 494: nit: `%w`

waddrmgr/address.go
```diff
@@ -629,6 +629,52 @@ func newManagedAddress(s *ScopedKeyManager, derivationPath DerivationPath,
...
+	if _, err := rand.Read(msg[:]); err != nil {
+		return nil, fmt.Errorf("unable to read random "+
+			"challenge for addr validation: %w", err)
+	}
+
+	// We'll first validate things against the private key we got
+	// from the original derivation.
+	err = managedAddr.Validate(msg, privKey)
+	if err != nil {
+		return nil, fmt.Errorf("addr validation for addr=%v "+
```
sputn1ck, line 646: nit: `Address` implements `Sting()`, is `%s` better here?

waddrmgr/address.go
```diff
@@ -629,6 +629,52 @@ func newManagedAddress(s *ScopedKeyManager, derivationPath DerivationPath,
...
+		return nil, fmt.Errorf("unable to re-derive "+
+			"key: %w", err)
+	}
+	freshPrivKey, err := rederivedKey.ECPrivKey()
+	if err != nil {
+		return nil, fmt.Errorf("unable to gen priv key: %w", err)
+	}
+	err = managedAddr.Validate(msg, freshPrivKey)
+	if err != nil {
+		return nil, fmt.Errorf("addr validation for addr=%v "+
```
sputn1ck, line 673: same q as above

waddrmgr/scoped_manager.go
```diff
@@ -1156,6 +1156,13 @@ func (s *ScopedKeyManager) nextAddresses(ns walletdb.ReadWriteBucket,
 		}
 
 		if ma.Address().String() != diskAddr.Address().String() {
+			// The address didn't match up, so we'll manually
+			// delete it from the cache.
+			delete(
```
sputn1ck, line 1161: nit: seems like the wrong commit

sputn1ck (commented):
LGTM, found some nits.

waddrmgr/address.go
Roasbeef: I think the end result would be the same? Since `%v` will try to call `String` if it exists.

PR #828, opened 2022-11-08
mod: bump to latest version of btcd

guggero:
This commit bumps to the latest version of btcd that fixes a key mutation issue when signing for Taproot outputs.


guggero:
Needs another update, will be replaced.

PR #832, opened 2022-11-14
wallet: log txid:index for leases where tx is unknown

Roasbeef:
Debugging a scenario where a lease exists on disk, but the tx associated with that output no longer does. This can potentially happen if a leased output is somehow spent before the lease, or if the lease itself was for an unconfirmed output which was RBF'd (thus the txid changed).


guggero (approved):
LGTM :tada:

bhandras (approved):

PR #833, opened 2022-11-14
wallet: default to using P2TR for change outputs

Roasbeef:


chappjc:
With `CreateSimpleTx` or `SendOutputs`, how would we indicate to use all utxos to fund the txns, but to use a p2wpkh change output?

wallet/createtx.go
```diff
@@ -354,7 +354,7 @@ func (w *Wallet) addrMgrWithChangeSource(dbtx walletdb.ReadWriteTx,
 
 	// Determine the address type for change addresses of the given account.
 	if changeKeyScope == nil {
-		changeKeyScope = &waddrmgr.KeyScopeBIP0084
+		changeKeyScope = &waddrmgr.KeyScopeBIP0086
```
chappjc, line 357: Docs on `addrMgrWithChangeSource` and `txToOutputs` both still indicate P2WPKH will be used for change if keyscope not specified.  Docs aside, I'm uncertain how to handle this in our app where we want all utxos selected for funding, but we expect the change to be P2WPKH.

May optional/variadic arg for `CreateSimpleTx` and `SendOutputs`?  Or maybe differently named exported methods with change type args?

Roasbeef:
> With CreateSimpleTx or SendOutputs, how would we indicate to use all utxos to fund the txns, but to use a p2wpkh change output?

Good question, w/ the way things are set up rn, that doesn't appear to be possible. Given that if you specify the `p2wkh` scope, then it'll only select inputs that match that.

A bigger change here would be to modify `CreateSimpeTx` further to allow more granular options (so key scope for inputs, and one for the change addr).

Roasbeef:
Tests fail rn due to the inflexibility described above (`TestFundPsbt`): the tests assume a `p2wkh` addr, but `p2tr` is used now by default.

wallet/createtx.go
Roasbeef: > May optional/variadic arg for CreateSimpleTx and SendOutputs?

Leaning towards this option, so then a potential change would be:
  * New optional arg for what the change addr scope should be 
  * If it isn't set, then it'll use the scope selected earlier (either default or specified scope) 
  * if set, then that's used for change addrs too 

Then for our desired use case (on the lnd side), we'd leave the existing arg, and start setting this new change arg.

guggero:
Yeah, sorry about those hard coded fees that make the test fail. Here's a fix for the unit test that uses dynamic fees:
[833-diff.txt](https://github.com/btcsuite/btcwallet/files/10014603/833-diff.txt)

wallet/createtx.go
Roasbeef: Ok did the the optional arg thing @chappjc, can you check out the latest diff? Want to make sure it still works for y'all's use case. I've also added a test to exercise the feature.

Roasbeef:
Ok this is ready for another pass. Also added a new test to confirm that taproot outputs are used by default, and also that the new custom change stuff works too.

wallet/psbt_test.go
```diff
@@ -7,7 +7,9 @@ package wallet
 import (
 	"bytes"
 	"encoding/hex"
-	"strings"
+	"github.com/btcsuite/btcwallet/wallet/txrules"
+	"github.com/btcsuite/btcwallet/wallet/txsizes"
```
guggero, line 13: Oops, my bad, these need to be grouped with the other imports.

wallet/wallet.go
```diff
@@ -1224,6 +1225,26 @@ out:
...
+
+// TxCreateOption...
+type TxCreateOption func(*txCreateOptions)
+
+// defaultTxCreateOptions...
+func defaultTxCreateOptions() *txCreateOptions {
+	return &txCreateOptions{}
+}
+
+// WithCustomChangeScope...
```
guggero, line 1241: nit: incomplete GoDoc.

wallet/wallet.go
```diff
@@ -1235,15 +1256,25 @@ out:
...
-	outputs []*wire.TxOut, minconf int32, satPerKb btcutil.Amount,
-	coinSelectionStrategy CoinSelectionStrategy, dryRun bool) (
-	*txauthor.AuthoredTx, error) {
+// the database. A tx created with this set to true SHOULD NOT be broadcast.
+func (w *Wallet) CreateSimpleTx(coinSelectKeyScope *waddrmgr.KeyScope,
+	account uint32, outputs []*wire.TxOut, minconf int32,
+	satPerKb btcutil.Amount, coinSelectionStrategy CoinSelectionStrategy,
+	dryRun bool, optFuncs ...TxCreateOption) (*txauthor.AuthoredTx, error) {
+
+	opts := defaultTxCreateOptions()
```
guggero, line 1270: Nice :+1:

guggero (approved):
LGTM :tada:

wallet/wallet.go
```diff
@@ -1235,15 +1256,25 @@ out:
...
-// the database. A tx created with this set to true SHOULD NOT be broadcasted.
-func (w *Wallet) CreateSimpleTx(keyScope *waddrmgr.KeyScope, account uint32,
-	outputs []*wire.TxOut, minconf int32, satPerKb btcutil.Amount,
-	coinSelectionStrategy CoinSelectionStrategy, dryRun bool) (
-	*txauthor.AuthoredTx, error) {
+// the database. A tx created with this set to true SHOULD NOT be broadcast.
+func (w *Wallet) CreateSimpleTx(coinSelectKeyScope *waddrmgr.KeyScope,
+	account uint32, outputs []*wire.TxOut, minconf int32,
+	satPerKb btcutil.Amount, coinSelectionStrategy CoinSelectionStrategy,
+	dryRun bool, optFuncs ...TxCreateOption) (*txauthor.AuthoredTx, error) {
```
chappjc, line 1268: Makes sense.  Users of `SendOutputs`, like myself, would benefit from the new are too though.
It looks simple enough to adapt and instead do `CreateSimpleTx` directly, followed by `PublishTransaction`... just noting.
Trying this PR out with a `replace` now to see.  Thanks for adapting!

chappjc: Seems I can write `SendOutputs` ourselves like the following, which is close, but not quite the same.

```go

func (w *spvWallet) sendOutputs(outputs []*wire.TxOut, keyScope *waddrmgr.KeyScope, account uint32, minconf int32,
	satPerKb btcutil.Amount, coinSelectionStrategy wallet.CoinSelectionStrategy, label string) (*wire.MsgTx, error) {
	// Ensure the outputs to be created adhere to the network's consensus rules.
	for _, output := range outputs {
		err := txrules.CheckOutput(output, txrules.DefaultRelayFeePerKb)
		if err != nil {
			return nil, err
		}
	}

	// Create the transaction and broadcast it to the network. The
	// transaction will be added to the database in order to ensure that we
	// continue to re-broadcast the transaction upon restarts until it has
	// been confirmed.
	changeTypeOpt := wallet.WithCustomChangeScope(&waddrmgr.KeyScopeBIP0084)
	createdTx, err := w.wallet.CreateSimpleTx(keyScope, account, outputs, minconf,
		satPerKb, coinSelectionStrategy, false, changeTypeOpt)
	if err != nil {
		return nil, err
	}

	err = w.wallet.PublishTransaction(createdTx.Tx, label)
	if err != nil {
		return nil, err
	}

	return createdTx.Tx, nil
}
```

chappjc: It's also worth noting that although the new signature for `CreateSimpleTx`, is close and the final arg is optional, it is still a breaking change because `interface` definitions will no longer match.

That's not terribly hard to adapt to, but it is something to consider about why a method with a different name could be better.
If a project uses btcwallet directly, plus it uses another project that also uses btcwallet and this method, you cannot update btcwallet until that _other project_ you required updates (all depending on if/how the projects define interfaces that `btcwallet.Wallet` implements).

e.g.

```go
type BTCWallet interface {
	PublishTransaction(tx *wire.MsgTx, label string) error
	ListUnspent(minconf, maxconf int32, acctName string) ([]*btcjson.ListUnspentResult, error)
	... etc
	Unlock(passphrase []byte, lock <-chan time.Time) error
	SendOutputs(outputs []*wire.TxOut, keyScope *waddrmgr.KeyScope, account uint32, minconf int32,
		satPerKb btcutil.Amount, coinSelectionStrategy wallet.CoinSelectionStrategy, label string) (*wire.MsgTx, error)
	// **** ERROR: "does not implement" - have / want" at compile:
	CreateSimpleTx(coinSelectKeyScope *waddrmgr.KeyScope, account uint32,
		outputs []*wire.TxOut, minconf int32, satPerKb btcutil.Amount,
		coinSelectionStrategy wallet.CoinSelectionStrategy, dryRun bool,
		optFuncs ...wallet.TxCreateOption) (*txauthor.AuthoredTx, error)
}
```

Roasbeef: > It looks simple enough to adapt and instead do CreateSimpleTx directly, followed by PublishTransaction... just noting.

The functional difference being that `SendOutputs` will publish for you? Or is it the extra tx rules checks?

Roasbeef: Ah ok you'd need to apply that label manually when broadcasting, but yeah that looks functionally the same.

chappjc: Yeah, the label, and I suppose the other things are that we don't get the txhash check when we use `PublishTransaction` because it eats the txhash output.  There's also a watch only wallet check, but that doesn't really affect use I think.
NBD, just not an exact clone of `SendOutputs`

PR #834, opened 2022-11-17
wallet: update sub modules to latest btcd version

Roasbeef:


wallet/txauthor/go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/btcsuite/btcwallet/wallet/txauthor
 go 1.12
 
 require (
-	github.com/btcsuite/btcd v0.23.1
+	github.com/btcsuite/btcd v0.23.3
```
guggero, line 6: Shouldn't we use `v0.23.4` to also include the tap tweak fix?

guggero: We do actually use the `txscript` package in `txauthor`, `txrules` and `txsizes`, that's why I think we should go to the latest version in these submodules.

Roasbeef: Done!

guggero (approved):

PR #835, opened 2022-11-18
wallet: fix argument order in call to txToOutputs

Roasbeef:
The initial version had the arguments swapped. This doesn't show up if both are nil, but can cause some instances to fail that otherwise wouldn't.


chappjc (approved):
ACK. All test either use both nil or same key scope.  Args are definitely coin selection followed by change.

guggero (approved):

PR #836, opened 2022-11-19
wallet: ensure change scope is uniform if coin select scope is set

Roasbeef:
In this commit, we fix a regression that was introduced as a result of the new change scope param. Of cases like spending from an imported account with PSBT, the change scope should be the same as the coin selection scope to ensure that a proper change address is available.

Example failure scenario: imported account uses the p2wkh scope, no other imported accounts, but then we fail when we go to make a change addr as the imported account hasn't had a p2tr account created.


guggero (approved):

positiveblue (approved):
LGTM  :+1:

PR #837, opened 2022-11-28
wallet: add output info to change output in FundPsbt

guggero:
In case we create a change output when funding a PSBT we want to include its derivation path and public key information in the PSBT in order to make the output information more useful and verbose.

This is used in order to create exclusion proofs for our change addresses in Taro when funding any anchor TX through PSBT.

cc @bhandras.

### DIFF
--- a/wallet/psbt.go
+++ b/wallet/psbt.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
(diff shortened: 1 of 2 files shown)


wallet/psbt.go
```diff
@@ -294,6 +309,52 @@ func addInputInfoSegWitV1(in *psbt.PInput, utxo *wire.TxOut,
 	}}
 }
 
+// addOutputInfo adds the BIP32 derivation info for an output from our internal
+// wallet.
+func addOutputInfo(out *psbt.POutput, txOut *wire.TxOut,
```
bhandras, line 314: nit: could be slightly more readable as ```getOutputInfo(*wire.TxOut, waddrmgr.ManagedPubKeyAddres) (*psbt.POutput, error)```

wallet/psbt_test.go
```diff
@@ -293,6 +293,19 @@ func TestFundPsbt(t *testing.T) {
 				t, tc.expectedChangeBeforeFee,
 				txOuts[changeIndex].Value+int64(expectedFee),
 			)
+
+			changeTxOut := txOuts[changeIndex]
+			changeOutput := packet.Outputs[changeIndex]
+			require.NotEmpty(t, changeOutput.Bip32Derivation)
```
bhandras, line 299: nit: could also add some additional checks apart from `NotEmpty`.

bhandras (approved):
LGTM :tada:

wallet/psbt.go
```diff
@@ -294,6 +309,52 @@ func addInputInfoSegWitV1(in *psbt.PInput, utxo *wire.TxOut,
...
+		},
+	}
+	out.Bip32Derivation = []*psbt.Bip32Derivation{
+		derivation,
+	}
+
+	// Include the Taproot derivation path as well if this is a P2TR output.
+	if txscript.IsPayToTaproot(txOut.PkScript) {
+		out.TaprootBip32Derivation = []*psbt.TaprootBip32Derivation{{
+			XOnlyPubKey:          derivation.PubKey[1:],
```
Roasbeef, line 348: Style nit: can capture this as `schnorrKey` above since it's used below also for `TaprootInternalKey`.

Roasbeef (commented):
Easy diff, reads well, just a small comment in addition to what Andras left.

wallet/psbt.go
guggero: Yeah, makes sense.

Roasbeef (approved):
LGTM ⏰

PR #838, opened 2022-11-30
Fix addresses from non-default scope being detected

guggero:
Fixes https://github.com/lightningnetwork/lnd/issues/6434.

This PR fixes a bug where we'd detect funds sent to a non-default
scope address when adding a relevant TX for another output (e.g. the
change from sending to the non-default scope address). This would lead
to not recognizing the spend of that output later on and the output
would remain as a "ghost UTXO".

### DIFF
--- a/chain/bitcoind_client.go
+++ b/chain/bitcoind_client.go
@@ -599,10 +599,11 @@ func (c *BitcoindClient) ntfnHandler() {
 	for {
 		select {
 		case tx := <-c.txNtfns:
-			_, _, err := c.filterTx(tx, nil, true)
+			txDetails := btcutil.NewTx(tx)
+			_, _, err := c.filterTx(txDetails, nil, true)
 			if err != nil {
 				log.Errorf("Unable to filter transaction %v: %v",
-					tx.TxHash(), err)
+					txDetails.Hash(), err)
 			}
 
 		case newBlock := <-c.blockNtfns:
@@ -1185,16 +1186,19 @@ func (c *BitcoindClient) filterBlock(block *wire.MsgBlock, height int32,
 		// Update the index in the block details with the index of this
 		// transaction.
 		blockDetails.Index = i
-		isRelevant, rec, err := c.filterTx(tx, blockDetails, notify)
+		txDetails := btcutil.NewTx(tx)
+		isRelevant, rec, err := c.filterTx(
+			txDetails, blockDetails, notify,
+		)
 		if err != nil {
 			log.Warnf("Unable to filter transaction %v: %v",
-				tx.TxHash(), err)
+				*txDetails.Hash(), err)
 			continue
 		}
 
 		if isRelevant {
 			relevantTxs = append(relevantTxs, rec)
-			confirmedTxs[tx.TxHash()] = struct{}{}
+			confirmedTxs[*txDetails.Hash()] = struct{}{}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


wallet/chainntfns.go
```diff
@@ -305,6 +307,21 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
...
+				// to non-default scopes in other places either,
+				// so detecting them here would mean we'd also
+				// not properly detect them as spent later.
+				scopedManager, _, err := w.Manager.AddrAccount(
+					addrmgrNs, addr,
+				)
+				if err != nil {
+					return err
+				}
+				if !waddrmgr.IsDefaultScope(scopedManager.Scope()) {
```
Roasbeef, line 321: So the scenario here is us picking up something like a plain force close that sends to a `p2kh` addr in lnd's bricksquad scope?

chain/bitcoind_client.go
```diff
@@ -599,10 +599,11 @@ func (c *BitcoindClient) ntfnHandler() {
 	for {
 		select {
 		case tx := <-c.txNtfns:
-			_, _, err := c.filterTx(tx, nil, true)
+			txHash := tx.TxHash()
```
Roasbeef, line 602: Alternatively, we can use `btcutil.Tx` as a wrapper here which will actually cache the hash by default at all times. Either way gives us the desired end result tho.

wallet/chainntfns.go
guggero: In this explicit scenario it were P2WKH addresses created from the public keys returned from `walletrpc.DeriveAddress` (derives from the bricksquad scope) which were used in the `bos create-balanced-channel` scenario. And they only showed up as ghost UTXOs if the user also ran a TX reset with the `--reset-wallet-transactions` flag.

guggero: Force closes of static_remote_key channels wouldn't be affected because even though it's an address derived from the same bricksquad scope, the TX spending from that address again would go into the wallet and be seen as a "relevant TX" while in this scenario it wouldn't as the spending TX was the P2WSH funding TX that the wallet didn't see as its "own".

chain/bitcoind_client.go
guggero: Ah, wasn't aware of that. Sounds good, updated.

guggero:
cc  @positiveblue.

wallet/chainntfns.go
```diff
@@ -305,6 +307,21 @@ func (w *Wallet) addRelevantTx(dbtx walletdb.ReadWriteTx, rec *wtxmgr.TxRecord,
 		for _, addr := range addrs {
 			ma, err := w.Manager.Address(addrmgrNs, addr)
 			if err == nil {
```
positiveblue, line 309: nit: it might make sense to simplify how the error checks are done, and avoid one identation

```
ma, err := w.Manager.Address(addrmgrNs, addr)

// Missing addresses are skipped.  
if waddrmgr.IsError(err, waddrmgr.ErrAddressNotFound) {
        continue
}
// Other errors should be propagated.
if err != nil {
        return err
}

// Prevent addresses from non-default scopes to
// be detected here. We don't watch funds sent
// to non-default scopes in other places either,
// so detecting them here would mean we'd also
// not properly detect them as spent later.
scopedManager, _, err := w.Manager.AddrAccount(
...
```

chain/bitcoind_client.go
```diff
@@ -599,10 +599,11 @@ func (c *BitcoindClient) ntfnHandler() {
 	for {
 		select {
 		case tx := <-c.txNtfns:
-			_, _, err := c.filterTx(tx, nil, true)
+			txDetails := btcutil.NewTx(tx)
```
positiveblue, line 602: :+1:

positiveblue (approved):
Good catch :100:

I guess `IsDefaultScope` can be used now in other functions like `ForEachRelevantActiveAddress`.

wallet/chainntfns.go
guggero: Good idea, done.

PR #840, opened 2022-12-09
chainntfns: Don't panic.

### LINKED ISSUE #827: chainntfns: Panics during initial sync blow up consumers.
Specifically, if wallet is locked during recovery btcwallet panics. Trying to use the wallet blows up whatever app is using it if a user simply locks the wallet at the wrong time. I also don't think there is a way to check if btcwallet is currently doing recovery.

I wonder if it's possible to remove the panic logic in `*Wallet.handleChainNotifications` altogether and handle errors more gracefully?

The point of panic that is causing some pain https://github.com/btcsuite/btcwallet/blob/27f244e8454076f5afcc5be6266c3762f902800f/wallet/chainntfns.go#L115-L119

The problematic recovery that we can not currently escape unless shutdown. Locking the wallet here panics. https://github.com/btcsuite/btcwallet/blob/27f244e8454076f5afcc5be6266c3762f902800f/wallet/wallet.go#L665-L792

JoeGruffins:
Unexpected panics are a problem for consumers. Remove them. If the birthday does not look correct, log the error and continue with a nil birthday. If there is a problem with initial sync, log and try again later.

closes #827

### DIFF
--- a/config.go
+++ b/config.go
@@ -37,31 +37,33 @@ const (
 )
 
 var (
-	btcdDefaultCAFile  = filepath.Join(btcutil.AppDataDir("btcd", false), "rpc.cert")
-	defaultAppDataDir  = btcutil.AppDataDir("btcwallet", false)
-	defaultConfigFile  = filepath.Join(defaultAppDataDir, defaultConfigFilename)
-	defaultRPCKeyFile  = filepath.Join(defaultAppDataDir, "rpc.key")
-	defaultRPCCertFile = filepath.Join(defaultAppDataDir, "rpc.cert")
-	defaultLogDir      = filepath.Join(defaultAppDataDir, defaultLogDirname)
+	btcdDefaultCAFile       = filepath.Join(btcutil.AppDataDir("btcd", false), "rpc.cert")
+	defaultAppDataDir       = btcutil.AppDataDir("btcwallet", false)
+	defaultConfigFile       = filepath.Join(defaultAppDataDir, defaultConfigFilename)
+	defaultRPCKeyFile       = filepath.Join(defaultAppDataDir, "rpc.key")
+	defaultRPCCertFile      = filepath.Join(defaultAppDataDir, "rpc.cert")
+	defaultLogDir           = filepath.Join(defaultAppDataDir, defaultLogDirname)
+	defaultSyncLoopInterval = time.Minute
 )
 
 type config struct {
 	// General application behavior
-	ConfigFile      *cfgutil.ExplicitString `short:"C" long:"configfile" description:"Path to configuration file"`
-	ShowVersion     bool                    `short:"V" long:"version" description:"Display version information and exit"`
-	Create          bool                    `long:"create" description:"Create the wallet if it does not exist"`
-	CreateTemp      bool                    `long:"createtemp" description:"Create a temporary simulation wallet (pass=password) in the data directory indicated; must call with --datadir"`
-	AppDataDir      *cfgutil.ExplicitString `short:"A" long:"appdata" description:"Application data directory for wallet config, databases and logs"`
-	TestNet3        bool                    `long:"testnet" description:"Use the test Bitcoin network (version 3) (default mainnet)"`
-	SimNet          bool                    `long:"simnet" description:"Use the simulation test network (default mainnet)"`
-	SigNet          bool                    `long:"signet" description:"Use the signet test network (default mainnet)"`
-	SigNetChallenge string                  `long:"signetchallenge" description:"Connect to a custom signet network defined by this challenge instead of using the global default signet test network -- Can be specified multiple times"`
-	SigNetSeedNode  []string                `long:"signetseednode" description:"Specify a seed node for the signet network instead of using the global default signet network seed nodes"`
-	NoInitialLoad   bool                    `long:"noinitialload" description:"Defer wallet creation/opening on startup and enable loading wallets over RPC"`
-	DebugLevel      string                  `short:"d" long:"debuglevel" description:"Logging level {trace, debug, info, warn, error, critical}"`
-	LogDir          string                  `long:"logdir" description:"Directory to log output."`
-	Profile         string                  `long:"profile" description:"Enable HTTP profiling on given port -- NOTE port must be between 1024 and 65536"`
-	DBTimeout       time.Duration           `long:"dbtimeout" description:"The timeout value to use when opening the wallet database."`
+	ConfigFile       *cfgutil.ExplicitString `short:"C" long:"configfile" description:"Path to configuration file"`
+	ShowVersion      bool                    `short:"V" long:"version" description:"Display version information and exit"`
+	Create           bool                    `long:"create" description:"Create the wallet if it does not exist"`
+	CreateTemp       bool                    `long:"createtemp" description:"Create a temporary simulation wallet (pass=password) in the data directory indicated; must call with --datadir"`
+	AppDataDir       *cfgutil.ExplicitString `short:"A" long:"appdata" description:"Application data directory for wallet config, databases and logs"`
+	TestNet3         bool                    `long:"testnet" description:"Use the test Bitcoin network (version 3) (default mainnet)"`
+	SimNet           bool                    `long:"simnet" description:"Use the simulation test network (default mainnet)"`
+	SigNet           bool                    `long:"signet" description:"Use the signet test network (default mainnet)"`
+	SigNetChallenge  string                  `long:"signetchallenge" description:"Connect to a custom signet network defined by this challenge instead of using the global default signet test network -- Can be specified multiple times"`
+	SigNetSeedNode   []string                `long:"signetseednode" description:"Specify a seed node for the signet network instead of using the global default signet network seed nodes"`
+	NoInitialLoad    bool                    `long:"noinitialload" description:"Defer wallet creation/opening on startup and enable loading wallets over RPC"`
+	DebugLevel       string                  `short:"d" long:"debuglevel" description:"Logging level {trace, debug, info, warn, error, critical}"`
+	LogDir           string                  `long:"logdir" description:"Directory to log output."`
+	Profile          string                  `long:"profile" description:"Enable HTTP profiling on given port -- NOTE port must be between 1024 and 65536"`
+	DBTimeout        time.Duration           `long:"dbtimeout" description:"The timeout value to use when opening the wallet database."`
+	SyncLoopInterval time.Duration           `long:"syncloopinterval" description:"Specify the amount of time to wait between errors during initial sync. Valid time units are {s, m, h}. Default is one minute."`
 
 	// Wallet options
 	WalletPass string `long:"walletpass" default-mask:"-" description:"The public wallet password -- Only required if the wallet was created with one"`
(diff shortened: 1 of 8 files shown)


wallet/chainntfns.go
```diff
@@ -107,15 +106,32 @@ func (w *Wallet) handleChainNotifications() {
 					chainClient, birthdayStore,
 				)
 				if err != nil && !waddrmgr.IsError(err, waddrmgr.ErrBirthdayBlockNotSet) {
-					panic(fmt.Errorf("unable to sanity "+
-						"check wallet birthday block: %v",
-						err))
```
JoeGruffins, line 112: It looks like the wallet is ok with a nil birthday below, so I do not think it is necessary to panic here.

wallet/chainntfns.go
```diff
@@ -107,15 +106,32 @@ func (w *Wallet) handleChainNotifications() {
...
+							log.Errorf("unable to synchronize "+
+								"wallet to chain, trying again "+
+								"in one minute: %v", err)
+							break
+						}
+						break out
+					case <-w.quitChan():
+						return
+					}
 				}
```
JoeGruffins, line 135: Very dumb loop, but I think better than the alternative of blowing up. In the case of the wallet being locked, it loops until unlocked. 

However, testing on testnet, I sometimes see an error like `022-12-09 21:22:38.865 [ERR] BTCW: unable to synchronize wallet to chain, trying again in one minute: unable to persist initial sync data: failed to store sync information 00000000000000d47ad9520e3aff5563601586cb4e19ede996827839e03a8326: failed to fetch block hash for height 1998551: block not found` after forcing a loop with a wallet lock.

Unsure atm if that is a result of these changes.

chappjc (commented):
These explicit `panic`s certainly need to go.

The reason it's so problematic is because they are _unrecoverable_ for the caller of `SynchronizeRPC`.  That means it kills the app that is using that `wallet.Wallet` instance.

To illustrate why only `SynchronizeRPC` can catch these, here's an alternative approach that least gracefully shutdowns the wallet and chain service on `panic`s bubbling up from `handleChainNotifications`:

```diff
diff --git a/wallet/wallet.go b/wallet/wallet.go
index 32bd4fa..dc576c7 100644
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -210,7 +205,15 @@ func (w *Wallet) SynchronizeRPC(chainClient chain.Interface) {
        // make changes from the RPC client) and not have to stop and
        // restart them each time the client disconnects and reconnets.
        w.wg.Add(4)
-       go w.handleChainNotifications()
+       go func() {
+               defer func() {
+                       if pv := recover(); pv != nil {
+                               log.Criticalf("handleChainNotifications died:\n\n%v\n\n", pv)
+                               w.Stop() // also stops chainClient
+                       }
+               }()
+               w.handleChainNotifications()
+       }()
        go w.rescanBatchHandler()
        go w.rescanProgressHandler()
        go w.rescanRPCHandler()
```

The caller of `SynchronizeRPC` cannot do the above because it's no longer on the same call stack as `handleChainNotifications`.

But really, each of the `panic`s in `handleChainNotifications` could be replaced with a `log.Critical(...)` -> `w.Stop()`.

**So there seem to be a few possible resolutions.**  The two above ideas cause a graceful shutdown instead of an app crash.  The one in this PR goes a bit further and attempts to retry.  Some of the errors are indeed transient, but some are not.  I'm not sure what's best yet, but this PR does look reasonable to me.

chappjc:
Accidentally commented on the linked issue regarding my testing.  Comment below.

I hit the `handleChainNotifications` panic twice today, induced by "couldn't retrieve block" both time, so I gave this diff a test.  Naturally it wouldn't reproduce without meddling, so I tried by locking the wallet during rescan.  It worked (mostly)!

```
<manually lock it here, with running rescan>
2023-05-08 17:35:11.535 [INF] BTCW: Recovered addresses from blocks 686014-688013
2023-05-08 17:35:11.535 [ERR] BTCW: unable to synchronize wallet to chain, trying again in one minute: unable to perform wallet recovery: recovery: forced shutdown
2023-05-08 17:35:11.535 [INF] BTCW: The wallet has been locked
... <manually unlocked it here>
2023-05-08 17:35:22.643 [INF] BTCW: The wallet has been unlocked without a time limit
... <retry timer fires>
2023-05-08 17:36:12.536 [INF] BTCW: RECOVERY MODE ENABLED -- rescanning for used addresses with recovery_window=250
2023-05-08 17:36:12.725 [INF] BTCW: Seed birthday surpassed, starting recovery of wallet from height=688014 hash=0000000000000000000ae672f5308b44eb64d263cfad57b1850031c4f3fec3d1 with recovery-window=250
2023-05-08 17:36:12.744 [INF] BTCW: Scanning 2000 blocks for recoverable addresses
```

I was worried it would crash straight away since some methods are strictly call-once, but `w.syncWithChain` is alright it seems.

The issue I had was that my balance was off.  I think it must have failed to register for some addresses after automatically restarting? I still think this is way way better than a panic, esp. since a consumer cannot `recover` it.

chappjc:
If the little retry loop seems problematic for any reason, I think the panic should be removed so it at least gracefully dies, using maybe `w.Stop()`, so at least the whole app (potentially something bigger than the btcwallet cli) doesn't crash.

How are you leaning regarding a fix for this explicit panics, @guggero ?  A partial list of ways the panic has been observed is in https://github.com/btcsuite/btcwallet/issues/827#issuecomment-1312239484
It seems like most of the errors that throw this panic would be transient, so my feeling is the retry loop is appropriate assuming `w.syncWithChain` can be called again without reconstructing the `Wallet` from scratch.

guggero:
I agree that the panics should be removed. But changing the behavior with the retry loop might not be the desired behavior for all consumers. Maybe we should make that configurable?
Also, the loop should be extracted into a dedicated method to avoid the unreadable many levels of indentation.

JoeGruffins:
Just rebased.

config.go
```diff
@@ -249,10 +250,10 @@ func parseAndSetDebugLevels(debugLevel string) error {
...
 //
 // The configuration proceeds as follows:
-//      1) Start with a default config with sane settings
-//      2) Pre-parse the command line to check for an alternative config file
-//      3) Load configuration file overwriting defaults with any specified options
-//      4) Parse CLI options and overwrite/add any specified options
+//  1. Start with a default config with sane settings
+//  2. Pre-parse the command line to check for an alternative config file
+//  3. Load configuration file overwriting defaults with any specified options
+//  4. Parse CLI options and overwrite/add any specified options
```
JoeGruffins, line 256: Sorry, my editor is formatting this on its own. Will revert...

JoeGruffins:
Added an option to the config so now it will panic as usual by default. Removed not panicking for the birthday since that is not strictly the problem at the moment. I attempted to move the new logic into its own function so that it reads easier. Let me know if I need to do more.

https://github.com/btcsuite/btcwallet/compare/5635cfb395563ba396c92011cde02f39858671b8..e1c5660f9c68153d0f426ca9b09c127fbde80fdb

config.go
JoeGruffins: reverted https://github.com/btcsuite/btcwallet/compare/e1c5660f9c68153d0f426ca9b09c127fbde80fdb..75db9f23dc42c84faa67b7ed9b3367da8f71f94d

JoeGruffins:
Use the config value here too https://github.com/btcsuite/btcwallet/compare/75db9f23dc42c84faa67b7ed9b3367da8f71f94d..c65786678b9b183e8251b8c768031e8c5cccbbef

guggero:
I didn't mean making the panicking configurable but whether or not we retry and potentially how long we wait for each retry (a minute might be way too long during an integration test).

JoeGruffins:
> I didn't mean making the panicking configurable but whether or not we retry and potentially how long we wait for each retry (a minute might be way too long during an integration test).

If not retry then what? Just error log and return?

I think the reason I made this a loop is because there's not currently a way, besides panicking, to tell the node something is wrong and we need to shut down now. That's why it panics now I guess. `handleChainNotifications()` does not return an error.

Also are the config options in the right place? config -> loader -> wallet open is correct? So, two options, whether to return/loop and how long to wait between loops? Should I leave the birthday panic for now?

Thanks for reviewing this pr btw!

JoeGruffins:
re @chappjc 

> But really, each of the panics in handleChainNotifications could be replaced with a log.Critical(...) -> w.Stop()

I think I looked at this but this does not let the consumer know the wallet has stopped? I can look again as it has been a while, but we need to know that the wallet has decided to stop and we need to restart it or something.

guggero:
I'm not really familiar with the whole notification logic of `btcwallet`. But I took a closer look and my recommendation would be:
 - Replace the first panic (from `birthdaySanityCheck`) with a simple error log statement.
 - Keep the retry loop, but extract it into its own function. Make the wait time configurable (replacing the `NoSyncPanic` field). Only log any errors occurring while re-trying and don't panic.

JoeGruffins:
No longer panicking at all, like the beginning. Allow the interval to be settable https://github.com/btcsuite/btcwallet/compare/c65786678b9b183e8251b8c768031e8c5cccbbef..569923cfa0c3176b5b21e997851f52c6dc83f186

JoeGruffins:
> Keep the retry loop, but extract it into its own function.

Like it is currently, or are you referring to something else?

wallet/loader.go
```diff
@@ -86,7 +88,7 @@ func NewLoader(chainParams *chaincfg.Params, dbDirPath string,
 // to store the wallet. Given that the external DB may be shared an additional
 // function is also passed which will override Loader.WalletExists().
 func NewLoaderWithDB(chainParams *chaincfg.Params, recoveryWindow uint32,
-	db walletdb.DB, walletExists func() (bool, error)) (*Loader, error) {
+	db walletdb.DB, syncLoopInterval time.Duration, walletExists func() (bool, error)) (*Loader, error) {
```
guggero, line 91: Would be nice to adhere to the 80 character line length limit (using 8 spaces for tabs) more or less observed throughout the codebase.

guggero (commented):
Looks good, thanks! Just a nit about line length throughout most of the diff.

JoeGruffins:
Bringing code down to 80 columns https://github.com/btcsuite/btcwallet/compare/569923cfa0c3176b5b21e997851f52c6dc83f186..617bd67931792f2bc523322f7e489edb15e5141f

chappjc:
The code makes sense, but are you guys sure you want breaking changes to the API?  Perhaps the fallout isn't bad: https://pkg.go.dev/github.com/btcsuite/btcwallet/wallet?tab=importedby  Mostly lnd and btcwallet forks.

guggero:
Yeah, i thought about that too. Functional options would be great, as that wouldn't break the API.

JoeGruffins:
Should I change this further? Do you mean like new methods `OpenWithoutPanic` and `OpenExistingWalletWithoutPanic`?

guggero:
> Should I change this further? Do you mean like new methods `OpenWithoutPanic` and `OpenExistingWalletWithoutPanic`?

If you have time, that would be great. I mean something like `WithLoopSyncInterval()`. To make the change as non-breaking (code compilation and behavior) as possible, maybe the default should be no retry and the functional option would enable the retry loop by specifying the retry interval.

I can also take this over if you're otherwise busy.

JoeGruffins:
> I can also take this over if you're otherwise busy.

If you have a plan that would be great! Feel free to use these changes or not! We just want the panics gone.

guggero:
> If you have a plan that would be great! Feel free to use these changes or not! We just want the panics gone.

Okay, thanks for your initial fix!
Replacement PR is here: https://github.com/btcsuite/btcwallet/pull/870, kept you as the original author for the first commit.

easymoney89:
You guys all doing a great job learning

PR #842, opened 2023-01-18
wallet: add ability to use custom change scope for FundPsbt

Torakushi:
In this commit, we add the possibility to use a custom change scope for FundPsbt. This allows the user to specify a scope for the change address instead of using the default one (BIP0086 at the moment).


wallet/psbt.go
```diff
@@ -43,7 +44,18 @@ import (
 // responsibility to lock the inputs before handing the partial transaction out.
 func (w *Wallet) FundPsbt(packet *psbt.Packet, keyScope *waddrmgr.KeyScope,
 	minConfs int32, account uint32, feeSatPerKB btcutil.Amount,
-	coinSelectionStrategy CoinSelectionStrategy) (int32, error) {
+	coinSelectionStrategy CoinSelectionStrategy, optFuncs ...TxCreateOption) (int32, error) {
+
+	opts := defaultTxCreateOptions()
```
Torakushi, line 49: This logic is the same in CreateSimpleTx. Maybe we can put this logic in a function but IMO it may decrease the readability

wallet/psbt.go
```diff
@@ -131,6 +143,7 @@ func (w *Wallet) FundPsbt(packet *psbt.Packet, keyScope *waddrmgr.KeyScope,
 		tx, err = w.CreateSimpleTx(
 			keyScope, account, packet.UnsignedTx.TxOut, minConfs,
 			feeSatPerKB, coinSelectionStrategy, false,
+			WithCustomChangeScope(opts.changeKeyScope),
```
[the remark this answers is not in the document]
joostjager: Then the `opts` definition above can be moved into the input-specified branch below.

wallet/psbt_test.go
```diff
@@ -354,6 +417,23 @@ func assertTxInputs(t *testing.T, packet *psbt.Packet,
...
+	// By default (changeScope == nil), the script should be a pay-to-taproot one.
+	switch changeScope {
+	case nil, &waddrmgr.KeyScopeBIP0086:
+		require.True(t, txscript.IsPayToTaproot(pkScript))
+	case &waddrmgr.KeyScopeBIP0049Plus, &waddrmgr.KeyScopeBIP0084:
+		require.True(t, txscript.IsPayToWitnessPubKeyHash(pkScript))
+	case &waddrmgr.KeyScopeBIP0044:
+		require.True(t, txscript.IsPayToPubKeyHash(pkScript))
+	default:
+		t.Fatalf("change scope not handled: %s", changeScope.String())
```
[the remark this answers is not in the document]
Torakushi: Took the format from here (to be consistent):
https://github.com/btcsuite/btcwallet/pull/842/files#diff-27d3fe2fb40c848223c31d69d979958b5b6c0a3ebe74870c7ed6d852fdf2bdc8R411

But maybe it is better indeed

wallet/psbt.go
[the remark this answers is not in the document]
Torakushi: Good idea indeed, we avoid building twice the opts

wallet/psbt_test.go
```diff
@@ -245,7 +386,7 @@ func TestFundPsbt(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			changeIndex, err := w.FundPsbt(
 				tc.packet, nil, 1, 0, tc.feeRateSatPerKB,
-				CoinSelectionLargest,
+				CoinSelectionLargest, WithCustomChangeScope(tc.changeKeyScope),
```
joostjager, line 389: Perhaps don't pass a modifier at all if `tc.changeKeyScope` is nil, because in real use it's probably not set either? It doesn't matter too much, because the modifier is a no op with nil.

wallet/psbt_test.go
```diff
@@ -217,6 +218,146 @@ func TestFundPsbt(t *testing.T) {
...
+			UnsignedTx: &wire.MsgTx{
+				TxIn: []*wire.TxIn{{
+					PreviousOutPoint: utxo1,
+				}},
+			},
+			Inputs: []psbt.PInput{{}},
+		},
+		feeRateSatPerKB:         20000,
+		validatePackage:         true,
+		changeKeyScope:          &waddrmgr.KeyScopeBIP0084,
```
joostjager, line 248: Is it worth testing all the different scopes because it seems that the relevant code paths are no different?

wallet/psbt.go
```diff
@@ -176,11 +178,23 @@ func (w *Wallet) FundPsbt(packet *psbt.Packet, keyScope *waddrmgr.KeyScope,
...
 		inputSource := constantInputSource(credits)
 
+		// Build the TxCreateOption to retrieve the change scope
+		opts := defaultTxCreateOptions()
+		for _, optFunc := range optFuncs {
+			optFunc(opts)
+		}
+
+		// If the change scope isn't set, then it should be the same as the
+		// coin selection scope in order to match existing behavior.
```
joostjager, line 188: "existing behavior", does that still refer to something meaningful after this is merged?

wallet/psbt.go
```diff
@@ -23,8 +23,9 @@ import (
 // FundPsbt creates a fully populated PSBT packet that contains enough inputs to
 // fund the outputs specified in the passed in packet with the specified fee
 // rate. If there is change left, a change output from the wallet is added and
-// the index of the change output is returned. Otherwise no additional output
-// is created and the index -1 is returned.
+// the index of the change output is returned. If no custom change scope is
+// specified, the BIP0086 will be used.
+// Otherwise no additional output is created and the index -1 is returned.
```
joostjager, line 28: New line above, or continue on the line above?

wallet/psbt_test.go
joostjager: Otherwise a parameterized test may be nice, but possibly overkill.

wallet/psbt.go
Torakushi: Actually yes.
Before this PR we passed `keyScope` as the `changeScope`.
Maybe "previous behaviour" is better?

wallet/psbt_test.go
joostjager: Up to you

wallet/psbt.go
joostjager: What I mean is that a reader of this comment can't know what the previous behavior was. It's more appropriate in the commit message perhaps.

wallet/psbt_test.go
joostjager: No strong opinion on generic assert. I do think that it is nice that you also assert the default change scope.

wallet/psbt_test.go
```diff
@@ -354,6 +497,24 @@ func assertTxInputs(t *testing.T, packet *psbt.Packet,
...
 
+func assertChangeOutputScope(t *testing.T, pkScript []byte,
+	changeScope *waddrmgr.KeyScope) {
+
+	// We check if the pkScript has the right type.
+	// By default (changeScope == nil), the script should be a pay-to-taproot one.
+	switch changeScope {
+	case nil, &waddrmgr.KeyScopeBIP0086:
+		require.True(t, txscript.IsPayToTaproot(pkScript))
+	case &waddrmgr.KeyScopeBIP0049Plus, &waddrmgr.KeyScopeBIP0084:
```
joostjager, line 508: Newlines above cases?

wallet/psbt_test.go
Torakushi: IMO, I would let like it is as it is a no-op (and we avoid a `if` right before)

+ `WithCustomChangeScope(nil)` and `nil` use the same path in the code

joostjager (approved):

wallet/psbt_test.go
```diff
@@ -354,6 +392,27 @@ func assertTxInputs(t *testing.T, packet *psbt.Packet,
 	}
 }
 
+func assertChangeOutputScope(t *testing.T, pkScript []byte,
+	changeScope *waddrmgr.KeyScope) {
+
+	// We check if the pkScript has the right type.
```
guggero, line 398: nit: this part of the comment could be the function's godoc comment. And the next line is overflowing the 80 char limit slightly.

guggero (commented):
Very nice, thanks a lot for the changes!
Only nits from my end. Looking forward to the PR that includes this in `lnd`.

wallet/psbt_test.go
```diff
@@ -245,7 +281,7 @@ func TestFundPsbt(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			changeIndex, err := w.FundPsbt(
 				tc.packet, nil, 1, 0, tc.feeRateSatPerKB,
-				CoinSelectionLargest,
+				CoinSelectionLargest, WithCustomChangeScope(tc.changeKeyScope),
```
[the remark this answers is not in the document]
Torakushi: yes thank you! I forgot indeed

Torakushi:
> Very nice, thanks a lot for the changes!
> Only nits from my end. Looking forward to the PR that includes this in lnd.

You are most welcome! :) 

The LND PR should arrive soon!

wallet/psbt_test.go
```diff
@@ -306,6 +343,8 @@ func TestFundPsbt(t *testing.T) {
 			// to a change output.
 			require.EqualValues(t, 1, b32d.Bip32Path[3])
 
+			assertChangeOutputScope(t, changeTxOut.PkScript, tc.changeKeyScope)
```
guggero, line 346: nit: line too long.

guggero (approved):
LGTM :tada: 

Can you please take the PR out of draft mode?

@Roasbeef would be great if you could take a final glance and then merge, thanks!

Torakushi:
> Can you please take the PR out of draft mode?

Okay! We thought that it could be a good idea to first do the LND PR before merging this one (to be sure that this btcwallet PR suits our needs)

guggero:
> We thought that it could be a good idea to first do the LND PR before merging this one.

Good point, makes sense to me.

Torakushi:
I've changed the comment in psbt.go:
https://github.com/btcsuite/btcwallet/pull/842/files#diff-b490a4d01f4b5de72521472b9b97220b310cd24523e36ae5ac27ea60cb35d209R27

Apparently, this was not clear enough (better explanation!)

Torakushi:
>>We thought that it could be a good idea to first do the LND PR before merging this one.

>Good point, makes sense to me.

I took this PR out of draft mode as the LND PR seems to be on the right track!
https://github.com/lightningnetwork/lnd/pull/7348

In theory, I don't think comments in the LND PR will require changes in the btcsuite PR 😄.

PR #845, opened 2023-02-11
build: update to use new generic neutrino/cache package 

Roasbeef:
We also update the min Go version to 1.18, and start to build against Go 1.20.


guggero (approved):
LGTM :tada:

PR #851, opened 2023-03-14
wallet: improve logging around transaction broadcast failures

guggero:
Because the returned error is not always logged by the caller, we want to make sure we log the _cause_ of the failure when transaction broadcasting fails.
In addition we also log the raw transaction as hex in order to allow manual broadcasting in specific cases.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3667,6 +3667,7 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 	// node that already has it in their mempool.
 	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L590
 	case match(err, "txn-already-in-mempool"):
+		log.Infof("%v: tx already in mempool", txid)
 		return &txid, nil
 
 	// If the transaction has already confirmed, we can safely remove it
@@ -3704,6 +3705,7 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 				"from unconfirmed store: %v", tx.TxHash(), dbErr)
 		}
 
+		log.Infof("%v: tx already confirmed", txid)
 		return &txid, nil
 
 	// If the transactions is invalid since it attempts to double spend a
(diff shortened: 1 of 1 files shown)


guggero:
cc @yyforyongyu.

wallet/wallet.go
```diff
@@ -3704,6 +3710,7 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 				"from unconfirmed store: %v", tx.TxHash(), dbErr)
 		}
 
+		log.Infof("%v: tx already in confirmed", txid)
```
yyforyongyu, line 3713:
```suggestion
		log.Infof("%v: tx already confirmed", txid)
```

wallet/wallet.go
```diff
@@ -3809,6 +3819,8 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 	} else {
 		log.Infof("Removed invalid transaction: %v",
 			spew.Sdump(tx))
+		log.Infof("Removed invalid transaction: %x",
```
yyforyongyu, line 3822: Could the above `spew` be removed now?

wallet/wallet.go
```diff
@@ -3667,6 +3672,7 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 	// node that already has it in their mempool.
 	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L590
 	case match(err, "txn-already-in-mempool"):
+		log.Infof("%v: tx already in mempool", txid)
```
yyforyongyu, line 3675: Should these be warning or info?

wallet/wallet.go
```diff
@@ -3646,8 +3646,13 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 	var (
 		txid      = tx.TxHash()
 		returnErr error
+		txRaw     bytes.Buffer
 	)
 
+	// The serialized transaction is for logging only, don't fail on the
+	// error.
+	_ = tx.Serialize(&txRaw)
```
yyforyongyu, line 3654: Not sure how costly is this `Serialize` method, but it seems we can call it at L3822 to save a few CPU cycles.

wallet/wallet.go
guggero: I think that's nice and useful to have as well since it allows you to quickly see more details about the TX. And the hex is useful if you want to manually re-broadcast for whatever reason.

wallet/wallet.go
guggero: Given that the original `Removed invalid transaction` is info, I went with info as well. Warn would probably lead to too many questions from concerned users. Normally this isn't cause for concern as it could just be a sweep attempt that was later replaced with a higher fee transaction.

wallet/wallet.go
guggero: Good point, done.

yyforyongyu (approved):
LGTM🥳

Roasbeef (approved):
LGTM 👑

PR #852, opened 2023-03-20
Verify psbt input data before finalizing

ziggie1984:
This relates to lightningnetwork/lnd#7527 and btcsuite/btcd#1964.

This would be my solution where I would fix the problem in lnd, happy to here your thoughts.

Or should we maybe directly check this in the lnd project/package ?

### DIFF
--- a/wallet/psbt.go
+++ b/wallet/psbt.go
@@ -383,8 +383,9 @@ func (w *Wallet) FinalizePsbt(keyScope *waddrmgr.KeyScope, account uint32,
 	packet *psbt.Packet) error {
 
 	// Let's check that this is actually something we can and want to sign.
-	// We need at least one input and one output.
-	err := psbt.VerifyInputOutputLen(packet, true, true)
+	// We need at least one input and one output. In addition each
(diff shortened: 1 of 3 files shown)


wallet/psbt.go
```diff
@@ -462,9 +463,9 @@ func (w *Wallet) FinalizePsbt(keyScope *waddrmgr.KeyScope, account uint32,
 			if keyScope == nil {
 				// If a key scope wasn't specified, then coin
 				// selection was performed from the default
-				// wallet accounts (NP2WKH, P2WKH), so any key
-				// scope provided doesn't impact the result of
-				// this call.
+				// wallet accounts (NP2WKH, P2WKH, P2TR), so any
+				// key scope provided doesn't impact the result
+				// of this call.
```
positiveblue, line 468: :+1:

guggero:
I pushed a new tag `v1.1.8` for the `github.com/btcsuite/btcd/btcutil/psbt` package.

guggero (approved):
LGTM, thanks for the fix :tada:

PR #853, opened 2023-04-03
chain: add `GetRawMempool` and `GetRawTransaction` to `BitcoindClient`

yyforyongyu:
This PR adds more RPC methods to `BitcoindClient`.

### DIFF
--- a/chain/bitcoind_client.go
+++ b/chain/bitcoind_client.go
@@ -184,11 +184,23 @@ func (c *BitcoindClient) IsCurrent() bool {
 	return bestHeader.Timestamp.After(time.Now().Add(-isCurrentDelta))
 }
 
-// GetRawTransactionVerbose returns a transaction from the tx hash.
+// GetRawTransactionVerbose returns a TxRawResult from the tx hash.
 func (c *BitcoindClient) GetRawTransactionVerbose(
 	hash *chainhash.Hash) (*btcjson.TxRawResult, error) {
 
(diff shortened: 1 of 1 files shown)


chain/bitcoind_client.go
```diff
@@ -184,13 +184,25 @@ func (c *BitcoindClient) IsCurrent() bool {
...
 }
 
+// GetRawTransaction returns a `btcutil.Tx` from the tx hash.
+func (c *BitcoindClient) GetRawTransaction(
+	hash *chainhash.Hash) (*btcutil.Tx, error) {
+
+	return c.chainConn.client.GetRawTransaction(hash)
+}
+
+// GetRawMempool returns the raw mempool.
```
Roasbeef, line 201: Do we also need to expand the top level interface so this can be used in a uniform manner? Or perhaps we need a sub-interface here so callers can type assert to see if this information is implemented? 

Alternatively, they can type assert on the interface value itself I guess.

yyforyongyu: Tried to add it to the interface, then realized neutrino doesn't implement the methods here. I guess at some point adding more methods to the `Interface` was paused because they are not implemented in `neutrino`?

Roasbeef: Ah yeah makes sense. Yeah I think at somepoint, we'll do a lot of refactoring and really clean all this up.

Roasbeef (approved):
LGTM 🐙

PR #855, opened 2023-04-13
Notify mempool

yyforyongyu:
Depends on #854 

This PR extends the mempool usage in the bitcoind event to make sure that if a subscription happens after the relevant tx is added into the mempool, the notification will still be sent about the relevant tx.


chain/bitcoind_events.go
```diff
@@ -645,13 +646,22 @@ func (b *bitcoindRPCPollingEvents) txEventHandlerRPC() {
...
 type mempool struct {
 	sync.RWMutex
+
+	// txs stores the txids in the mempool.
 	txs map[chainhash.Hash]bool
+
+	// inputs stores the inputs of the txids in the mempool. This is
+	// created for fast lookup.
+	//
+	// TODO(yy): create similar maps for outputs.
```
Roasbeef, line 759: Outputs to be used to finding new scripts created in the mempool?

chain/bitcoind_events.go
```diff
@@ -723,5 +754,54 @@ func (m *mempool) deleteUnmarked() {
...
+		m.removeInputs(hash)
+	}
+}
+
+// removeInputs takes a txid and removes the inputs of the tx from the
+// mempool's inputs map.
+//
+// NOTE: must be used inside a lock.
+func (m *mempool) removeInputs(tx chainhash.Hash) {
+	for outpoint, txid := range m.inputs {
```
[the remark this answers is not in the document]
Roasbeef: Yep sounds good to leave for future optimization. One way would be a two layer map so txid -> outpoint. Don't think we need to worry about this in the short term though.

chain/bitcoind_events.go
```diff
@@ -268,6 +292,11 @@ func (b *bitcoindZMQEvents) blockEventHandler() {
 				return
 			}
 
+			// From our local mempool map, let's remove each of the
+			// transactions that are confirmed in this new block,
+			// since they are no longer in the mempool.
+			b.mempool.clean(block.Transactions)
```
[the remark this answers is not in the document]
Roasbeef: Yeah good point, I think we've seen this duplicate tx notifications elsewhere as well. It's also the case that we _do_ handle blocks vs txns in diff goroutines as well. I think this is why we went with the approach of marked vs unmarked earlier as well?

chain/bitcoind_client.go
```diff
@@ -245,12 +245,46 @@ func (c *BitcoindClient) NotifyReceived(addrs []btcutil.Address) error {
...
+	// Send the outpoints so the client will cache them.
 	select {
 	case c.rescanUpdate <- outPoints:
 	case <-c.quit:
 		return ErrBitcoindClientShuttingDown
 	}
 
+	// Now we do a quick check in current mempool to see if we already have
+	// txes that spends the given outpoints.
+	for _, op := range outPoints {
```
Roasbeef, line 257: I thought the idea was to push this up a bit into lnd's existing chain notification system?

chain/bitcoind_events_test.go
```diff
@@ -408,7 +408,6 @@ func waitForBlockNtfn(t *testing.T, ntfns <-chan interface{},
 			switch ntfnType := nftn.(type) {
 			case BlockConnected:
 				if !connected {
-					fmt.Println("???")
```
Roasbeef, line 305: 🤣

chain/bitcoind_events.go
```diff
@@ -662,29 +672,50 @@ func (m *mempool) clean(txs []*wire.MsgTx) {
...
 	m.RLock()
 	defer m.RUnlock()
 
 	_, ok := m.txs[hash]
 	return ok
 }
 
+// containsInput returns true if the given transaction hash is already in our
+// mempool.
+func (m *mempool) containsInput(op wire.OutPoint) (chainhash.Hash, bool) {
```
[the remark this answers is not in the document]
yyforyongyu: updated

chain/bitcoind_events.go
```diff
@@ -459,6 +459,11 @@ func (b *bitcoindRPCPollingEvents) Start() error {
 		return err
 	}
 
+	// Load the mempool so we don't miss transactions.
+	if err := b.loadMempool(); err != nil {
```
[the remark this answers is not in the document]
yyforyongyu: guess it's just an optimization thing. Suppose the first ticker ticks after 10s, and we subscribe, then it can give us results instantly.

chain/bitcoind_events.go
```diff
@@ -365,6 +394,15 @@ func (b *bitcoindZMQEvents) txEventHandler() {
...
 			}
 
+			// Add the transaction to our local mempool.
+			//
+			// NOTE: unlike how it's handled in polling events, we
+			// won't mark or unmark the tx in mempool as the tx is
+			// supposed to be sent by bitcoind in real-time, thus
+			// we don't need to fetch the raw mempool to decide
+			// whether the tx is in mempool or not.
+			b.mempool.add(tx)
```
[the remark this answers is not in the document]
yyforyongyu: updated

chain/bitcoind_events.go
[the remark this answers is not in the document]
yyforyongyu: cool added todo

chain/bitcoind_events.go
```diff
@@ -723,5 +754,54 @@ func (m *mempool) deleteUnmarked() {
...
+			m.inputs[outpoint] = tx.TxHash()
+			continue
+		}
+
+		log.Tracef("Input %s was spent in tx %s, now spent in %s",
+			outpoint, oldTxid, tx.TxHash())
+
+		// If the input has been spent in an old tx, we need to
+		// overwrite it.
+		m.inputs[outpoint] = tx.TxHash()
```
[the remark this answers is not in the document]
yyforyongyu: > This needs to remove whatever was stored in m.inputs[outpoint] from m.txs so m.txs doesn't grow.

This is not `updateInputs`'s job to clean txids as it's only responsible for updating inputs. And `m.txs` should not be growing due to it's purged either due to `Clean` or `deleteUnmarked`.

Technically we should never end here, so maybe an error would be more appropriate.

chain/bitcoind_events.go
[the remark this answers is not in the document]
yyforyongyu: Cool now updated the logic to take advantage of the marks. And this is now what happens,
- when a new block comes in, it will first delete all the unmarked transactions from the previous round.
- if it read block notification first, then it unmarks all the transactions from the block. When the same tx is received in rawtx again, it won't be processed.

chain/bitcoind_events.go
yyforyongyu: yeah, updated the docs

chain/bitcoind_client.go
yyforyongyu: I think we want to do it here as we already have the mempool data available.

chain/bitcoind_zqm_events.go
```diff
@@ -319,3 +402,189 @@ func (b *bitcoindZMQEvents) txEventHandler() {
...
+		// sure to check if we've been requested to shut down.
+		select {
+		case <-b.quit:
+			return
+		default:
+		}
+
+		// Poll an event from the ZMQ socket.
+		var (
+			bufs = [][]byte{eventHash[:], eventType[:]}
```
Crypt-iQ, line 446: Why does this omit the command?

chain/bitcoind_zqm_events.go
```diff
@@ -319,3 +402,189 @@ func (b *bitcoindZMQEvents) txEventHandler() {
...
+
+		// Block connected.
+		case `C`:
+
+		// Block disconnected.
+		case `D`:
+
+		// Transaction is removed from mempool but NOT because it's
+		// confirmed.
+		case `R`:
```
Crypt-iQ, line 486: I think both `R` and `A` need to read the mempool sequence int or else it will affect the next time we call `Receive`

Crypt-iQ (commented):
I think a better approach for now would be to poll the mempool for the zmq variant. We can totally get rid of that behavior later and just rely on `sequence`+`tx`+`block` topics to construct the mempool, but I think in the short-term it is easiest to poll.

chain/bitcoind_zqm_events.go
Crypt-iQ: This should be `bufs = [][]byte{command[:], data, seqNum[:]}`

Then we'd go ahead and parse `eventHash`, `eventType`, and `eventSeq` (not `seqNum`) from `data`

chain/bitcoind_client.go
Roasbeef: Yep makes sense.

chain/bitcoind_conn.go
```diff
@@ -23,6 +23,10 @@ const (
 	// notifications from bitcoind through ZMQ.
 	rawTxZMQCommand = "rawtx"
 
+	// sequenceZMQCommand is the command used to receive sequence
+	// notifications from bitcoind through ZMQ.
+	sequenceZMQCommand = "sequence"
```
Roasbeef, line 28: IIUC, this is only in _newer_ versions of `bitcoind`, so we'd need to do a version check if we were trying to enable it in a graceful manner.

chain/bitcoind_events_test.go
```diff
@@ -338,6 +338,8 @@ func setupBitcoind(t *testing.T, minerAddr string,
 
 	zmqBlockHost := "ipc:///" + tempBitcoindDir + "/blocks.socket"
 	zmqTxHost := "ipc:///" + tempBitcoindDir + "/tx.socket"
+	zmqSequenceHost := "ipc:///" + tempBitcoindDir + "/sequence.socket"
```
Roasbeef, line 424: Will need documentation on the `lnd` side to set. But as mentioned above, I don't think we can actually rely on this unless we start to be stricter with teh bitcoind version required. IIUC, we support bitcoind 22 onwards (post taproot)

chain/bitcoind_rpc_events.go
```diff
@@ -126,6 +131,30 @@ func (b *bitcoindRPCPollingEvents) BlockNotifications() <-chan *wire.MsgBlock {
...
+
+	// Check whether the input is in mempool.
+	txid, found := b.mempool.containsInput(op)
+	if !found {
+		return chainhash.Hash{}, false
+	}
+
+	// Check if the tx spending the input is still marked. Unmarked txes
+	// are pending removal and are technically not in the mempool.
+	marked, found := b.mempool.containsTx(txid)
```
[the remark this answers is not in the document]
yyforyongyu: yeah since it's updated inside a lock they must be consistent. I think this was copied from zmq part, removed now and reverted `containsTx` back.

chain/bitcoind_zqm_events.go
```diff
@@ -0,0 +1,321 @@
+package chain
```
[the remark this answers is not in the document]
yyforyongyu: 🤦🏻

chain/bitcoind_conn.go
yyforyongyu: Removed, will add the check if we ever decide to implement it.

Crypt-iQ (approved):

chain/mempool.go
```diff
@@ -179,3 +184,82 @@ func (m *mempool) updateInputs(tx *wire.MsgTx) {
...
+		m.add(tx.MsgTx())
+	}
+
+	return nil
+}
+
+// UpdateMempoolTxes takes a slice of transactions from the current mempool and
+// use it to update its internal mempool. It returns a slice of transactions
+// that's new to its internal mempool.
+func (m *mempool) UpdateMempoolTxes(txids []*chainhash.Hash) []*wire.MsgTx {
```
Roasbeef, line 218: 👍

Roasbeef (approved):
LGTM 💸

PR #859, opened 2023-04-26
chain: add additional logging around new mempool logic

Roasbeef:
In this commit, we start to log how long it takes to load+synchronize the mempool. We also demote some logs to the warn logging level. We've seen these calls fail, seemingly due to active replacement attempts in the mempool.

### DIFF
--- a/chain/bitcoind_rpc_events.go
+++ b/chain/bitcoind_rpc_events.go
@@ -237,6 +237,11 @@ func (b *bitcoindRPCPollingEvents) txEventHandlerRPC() {
 	for {
 		select {
 		case <-ticker.C:
+			log.Debugf("Reconciling mempool spends with node " +
+				"mempool...")
+
+			now := time.Now()
+
 			// After each ticker interval, we poll the mempool to
 			// check for transactions we haven't seen yet.
 			txs, err := b.client.GetRawMempool()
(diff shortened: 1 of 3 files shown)


chain/bitcoind_zmq_events.go
```diff
@@ -399,6 +399,11 @@ func (b *bitcoindZMQEvents) mempoolPoller() {
 	for {
 		select {
 		case <-ticker.C:
+			log.Debugf("Reconcliing mempool spends with node " +
```
Roasbeef, line 402: Could make a sort of `logExecutionTime` func closure for this and the other instances.

chain/bitcoind_rpc_events.go
```diff
@@ -237,6 +237,11 @@ func (b *bitcoindRPCPollingEvents) txEventHandlerRPC() {
 	for {
 		select {
 		case <-ticker.C:
+			log.Debugf("Reconcliing mempool spends with node " +
```
yyforyongyu, line 240:
```suggestion
			log.Debugf("Reconciling mempool spends with node " +
```

chain/bitcoind_zmq_events.go
yyforyongyu, line 402:
```suggestion
			log.Debugf("Reconciling mempool spends with node " +
```

yyforyongyu (changes requested):
Good stuff!

PR #860, opened 2023-04-26
chain: make mempool loading async, use fine grained locking

Roasbeef:
For nodes with very large mempools, or node running on meager hardware,
loading the mempoool might take some time. To mitigate this, in this
commit, we move to make loading the mempool async. We also start to use
more fine grained locking to avoid holding the internal mutex the entire
time while we're processing the mempool diff.

If we make the [sync map in lnd a module](https://github.com/lightningnetwork/lnd/blob/fdc1ae9d57326cc82899e4c76d5069b14a764b96/lnutils/sync_map.go#L9), then we should be able to use it 
here as well. Alternatively, we can just copy things over.


chain/mempool.go
```diff
@@ -198,23 +198,26 @@ func (m *mempool) LoadMempool() error {
...
-
-	for _, txHash := range txs {
-		// Grab full mempool transaction from hash.
-		tx, err := m.client.GetRawTransaction(txHash)
-		if err != nil {
-			log.Warnf("unable to fetch transaction %s for "+
-				"mempool: %v", txHash, err)
-			continue
+	go func() {
+
```
yyforyongyu, line 202: nit: could remove the extra newline

chain/mempool.go
```diff
@@ -198,23 +198,26 @@ func (m *mempool) LoadMempool() error {
...
+			tx, err := m.client.GetRawTransaction(txHash)
+			if err != nil {
+				log.Warnf("unable to fetch transaction %s for "+
+					"mempool: %v", txHash, err)
+				continue
+			}
+
+			// Add the transaction to our local mempool.
+			m.Lock()
+			m.add(tx.MsgTx())
```
yyforyongyu, line 214: there's a `m.Add` can be used here

chain/mempool.go
```diff
@@ -257,7 +260,9 @@ func (m *mempool) UpdateMempoolTxes(txids []*chainhash.Hash) []*wire.MsgTx {
 		// bitcoind. We do this so that if that call happens to
 		// initially fail, then we will retry it on the next interval
 		// since it is still not in our local mempool.
+		m.Lock()
```
yyforyongyu, line 263: same here re using `m.Add`

chain/bitcoind_rpc_events.go
```diff
@@ -237,6 +237,11 @@ func (b *bitcoindRPCPollingEvents) txEventHandlerRPC() {
 	for {
 		select {
 		case <-ticker.C:
+			log.Debugf("Reconcliing mempool spends with node " +
```
bhandras, line 240: nit: Reconcliing

chain/mempool.go
```diff
@@ -99,6 +99,16 @@ func (m *mempool) add(tx *wire.MsgTx) {
 	m.updateInputs(tx)
 }
 
+// UnmarkAll un-marks all the transactions in the mempool. This should be done
+// just before we re-evaluate the contents of our local mempool comared to the
```
bhandras, line 108: nit: comared

bhandras (approved):
LGTM 🎉

PR #861, opened 2023-04-27
chain: parallelize mempool loading + reconciliation

Roasbeef:


chain/mempool.go
```diff
@@ -238,19 +251,30 @@ func (m *mempool) LoadMempool() error {
...
-		for _, txHash := range txs {
-			// Grab full mempool transaction from hash.
-			tx, err := m.client.GetRawTransaction(txHash)
-			if err != nil {
-				log.Warnf("unable to fetch transaction %s for "+
-					"mempool: %v", txHash, err)
-				continue
-			}
+		var eg errgroup.Group
+		eg.SetLimit(runtime.NumCPU())
```
positiveblue, line 255: This is basically limiting the number of gorotuines that can be executed at the same time in this error group right?

Do we do that because we do not want to saturate the sever with too many `m.client.GetRawTransaction`? Just checking if it makes sense to increase this number to `>NumCPU` while limiting it.

chain/mempool.go
```diff
@@ -118,7 +118,7 @@ func (m *mempool) add(tx *wire.MsgTx) {
 }
 
 // UnmarkAll un-marks all the transactions in the mempool. This should be done
-// just before we re-evaluate the contents of our local mempool comared to the
+// just before we re-evaluate the contents of our local mempool compared to the
```
positiveblue, line 121: solved

positiveblue (approved):
Thanks! :+1:

chain/mempool.go
Roasbeef: Yep, also bitcoind has a default number of rpc threads of 4, so any other requests will block beyond that value.

PR #864, opened 2023-05-09
chain: optimize removal of inputs in mempool

yyforyongyu:
This commit implements a new struct, `cachedInputs`, to provide faster lookup when removing inputs based on a given transaction. Instead of iterating all the inputs to decide whether they should be removed, we now cache `txid -> inputs` so we can update our local mempool faster by iterating much less data.

### DIFF
--- a/chain/mempool.go
+++ b/chain/mempool.go
@@ -12,6 +12,69 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+// cachedInputs caches the inputs of the transactions in the mempool. This is
+// used to provide fast lookup between txids and inputs.
+type cachedInputs struct {
+	// inputs provides a fast lookup from input -> txid.
+	inputs map[wire.OutPoint]chainhash.Hash
+
+	// txids provides a fast lookup from txid -> inputs.
+	txids map[chainhash.Hash]map[wire.OutPoint]struct{}
+}
+
+// newCachedInputs creates a new cachedInputs.
+func newCachedInputs() *cachedInputs {
+	return &cachedInputs{
+		inputs: make(map[wire.OutPoint]chainhash.Hash),
+		txids:  make(map[chainhash.Hash]map[wire.OutPoint]struct{}),
+	}
+}
+
+// hasInput returns the txid and a boolean to indicate the given input is
+// found.
+func (c *cachedInputs) hasInput(op wire.OutPoint) (chainhash.Hash, bool) {
+	txid, ok := c.inputs[op]
+	return txid, ok
+}
+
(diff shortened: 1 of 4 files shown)


chain/mempool.go
```diff
@@ -12,6 +12,71 @@ import (
...
+func (c *cachedInputs) hasInput(op wire.OutPoint) (chainhash.Hash, bool) {
+	txid, ok := c.inputs[op]
+	return txid, ok
+}
+
+// addInput adds the given input to our cached inputs. If the input already
+// exists, the `inputs` map will be overwritten and the `txids` map will be
+// updated.
+func (c *cachedInputs) addInput(op wire.OutPoint, txid chainhash.Hash) {
+	oldTxid, ok := c.inputs[op]
```
guggero, line 44: nit: `ok` not used. Or if it's used on L58, then it's not very clear whether it's overwritten in L50, so should maybe use a different name?

guggero (approved):
Very nice, should make cleanup way faster!

chain/mempool.go
```diff
@@ -12,6 +12,69 @@ import (
...
 )
 
+// cachedInputs caches the inputs of the transactions in the mempool. This is
+// used to provide fast lookup between txids and inputs.
+type cachedInputs struct {
+	// inputs provides a fast lookup from input -> txid.
+	inputs map[wire.OutPoint]chainhash.Hash
+
+	// txids provides a fast lookup from txid -> inputs.
+	txids map[chainhash.Hash]map[wire.OutPoint]struct{}
```
Roasbeef, line 22: In theory the second level here can just be an index? Since the txid is the first part of the outpoint.

chain/mempool.go
```diff
@@ -189,15 +251,8 @@ func (m *mempool) deleteUnmarked() {
...
-//
-// TODO(yy): create a txid -> [inputs] map to make this faster.
 func (m *mempool) removeInputs(tx chainhash.Hash) {
-	for outpoint, txid := range m.inputs {
-		if txid.IsEqual(&tx) {
-			// NOTE: it's safe to delete while iterating go map.
-			delete(m.inputs, outpoint)
-		}
-	}
+	m.inputs.removeInputsFromTx(tx)
```
Roasbeef, line 255: 👍

chain/mempool.go
yyforyongyu: this map stores `spendingTxid: inputs being spent`, so the outpoint will be different from the txid.

Roasbeef (approved):
LGTM ❄️

PR #870, opened 2023-05-25
wallet: don't panic, retry instead

guggero:
Replaces #840.

Fixes https://github.com/btcsuite/btcwallet/issues/827.

Main diff vs. #840:
 - Don't expose the sync retry interval to the user level config, choose a sane default instead (5 seconds sounds reasonable to me for most cases)
 - Make all changes non-breaking to existing API users (add new `OpenWithRetry()` function to wallet, add functional options to loader)

cc  @JoeGruffins, @chappjc, @Roasbeef

### DIFF
--- a/wallet/chainntfns.go
+++ b/wallet/chainntfns.go
@@ -6,7 +6,6 @@ package wallet
 
 import (
 	"bytes"
-	"fmt"
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
@@ -84,6 +83,42 @@ func (w *Wallet) handleChainNotifications() {
 		return err
 	}
 
+	waitForSync := func(birthdayBlock *waddrmgr.BlockStamp) error {
+		// We start with a retry delay of 0 to execute the first attempt
+		// immediately.
+		var retryDelay time.Duration
+		for {
+			select {
+			case <-time.After(retryDelay):
+				// Set the delay to the configured value in case
+				// we actually need to re-try.
+				retryDelay = w.syncRetryInterval
+
+				// Sync may be interrupted by actions such as
+				// locking the wallet. Try again after waiting a
+				// bit.
+				err = w.syncWithChain(birthdayBlock)
+				if err != nil {
+					if w.ShuttingDown() {
+						return ErrWalletShuttingDown
+					}
+
+					log.Errorf("Unable to synchronize "+
+						"wallet to chain, trying "+
+						"again in %s: %v",
+						w.syncRetryInterval, err)
+
+					continue
+				}
+
+				return nil
+
+			case <-w.quitChan():
+				return ErrWalletShuttingDown
+			}
+		}
+	}
+
 	for {
 		select {
 		case n, ok := <-chainClient.Notifications():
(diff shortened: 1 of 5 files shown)


wallet/chainntfns.go
```diff
@@ -106,17 +141,23 @@ func (w *Wallet) handleChainNotifications() {
 				birthdayBlock, err := birthdaySanityCheck(
 					chainClient, birthdayStore,
 				)
-				if err != nil && !waddrmgr.IsError(err, waddrmgr.ErrBirthdayBlockNotSet) {
-					panic(fmt.Errorf("unable to sanity "+
```
Roasbeef, line 110: Are we able to reliably produce this error? I think it was left unresolved for so long as it seemed to just fix itself on restart. 

I think the common case was something like moving from a full node backend to neutrino or the other way around?

guggero: Ah, also https://github.com/lightningnetwork/lnd/issues/6292.

Roasbeef: Yeah I understand what we're trying to fix. I'm curious if we have a way to reliably reproduce it in a test environment, so we can directly test this patch. At a glance should do what it says on the tin, but whenever I tried I wasn't able to hit the issue reliably.

guggero: Ah, sorry, I misread.

Yes, I was able to reproduce this quite easily: Sync `lnd` to a local regtest network, mine a couple of blocks. Then stop everything and wipe test regtest block chain but keep the `lnd` state (forcing the wallet to be synced to a different chain). On next startup `lnd` will panic with `panic: unable to synchronize wallet to chain: -8: Block height out of range`.

So I can see how this might also happen on mainnet depending on the error `bitcoind` returns.

With the patch it doesn't panic anymore but re-try (which in this example won't ever succeed, but on mainnet if it's just a temporary startup error might help):

```
2023-05-30 13:49:51.466 [INF] LTND: Waiting for chain backend to finish sync, start_height=119
2023-05-30 13:49:56.462 [ERR] LNWL: Unable to synchronize wallet to chain, trying again in 5s: -8: Block height out of range
2023-05-30 13:50:01.463 [ERR] LNWL: Unable to synchronize wallet to chain, trying again in 5s: -8: Block height out of range
2023-05-30 13:50:06.464 [ERR] LNWL: Unable to synchronize wallet to chain, trying again in 5s: -8: Block height out of range
2023-05-30 13:50:11.466 [ERR] LNWL: Unable to synchronize wallet to chain, trying again in 5s: -8: Block height out of range
```

chappjc: I don't know about the birthday block error, but there were a few ways to hit the `syncWithChain` panic listed in https://github.com/btcsuite/btcwallet/issues/827#issuecomment-1312239484.

I often hit the `couldn't retrieve block <hash> from network` panic if I try to rescan/reseed a wallet using SPV.  Briefly interrupting your machine's network to force a peer drop when a block is requested can do it, depending on timing.

Roasbeef (approved):
LGTM 👑

PR #871, opened 2023-06-03
Handle mempool minimum fee error

ziggie1984:
Relates to https://github.com/lightningnetwork/lnd/issues/7156.
and https://github.com/lightninglabs/neutrino/pull/278

This PR introduces a new error which handles the case where our backend reports either a `min relay fee not met` error or a `mempool min fee not met`.  I decided to go with a new error because handling this case as a  `ErrDoubleSpend` feels not right because it is not really a doublespend. One could decide whether we want to include the `min relay fee not met` case, but this was a nice way to test it locally, but I find it reasonable to include it too.

With this PR in place we can check for this error in higher level software like `lnd` and behave differently in some situations as menitoned in https://github.com/lightningnetwork/lnd/issues/7156.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3533,6 +3533,25 @@ func (e *ErrReplacement) Unwrap() error {
 	return e.backendError
 }
 
+// ErrMempoolFee is an error returned from PublishTransaction in case the
+// published transaction failed to propagate since it did not match the
+// current mempool fee requirement.
+type ErrMempoolFee struct {
+	backendError error
+}
+
+// Error returns the string representation of ErrMempoolFee.
+//
+// NOTE: Satisfies the error interface.
+func (e *ErrMempoolFee) Error() string {
+	return fmt.Sprintf("mempool fee not met: %v", e.backendError)
+}
+
+// Unwrap returns the underlying error returned from the backend.
+func (e *ErrMempoolFee) Unwrap() error {
+	return e.backendError
+}
+
 // PublishTransaction sends the transaction to the consensus RPC server so it
 // can be propagated to other nodes and eventually mined.
 //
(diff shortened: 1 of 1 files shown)


wallet/wallet.go
```diff
@@ -3533,6 +3533,25 @@ func (e *ErrReplacement) Unwrap() error {
...
+
+// Error returns the string representation of ErrMempoolFee.
+//
+// NOTE: Satisfies the error interface.
+func (e *ErrMempoolFee) Error() string {
+	return fmt.Sprintf("mempool fee not met: %v", e.backendError)
+}
+
+// Unwrap returns the underlying error returned from the backend.
+func (e *ErrMempoolFee) Unwrap() error {
```
Roasbeef, line 3551: 👍

wallet/wallet.go
```diff
@@ -3788,6 +3807,34 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
+			"fees, consider raising your bitcoind maxmempool "+
+			"config value", txid)
+
+	// Returned by bitcoind when a transaction does not meet the fee
+	// requirements to be accepted into mempool because the policy of the
+	// mempool has a higher min relay fee. Default value for bitcoind is
+	// 1000 sat/kvbyte but this is configurable so we log the appropriate
+	// bitcoind config setting.
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L514
+	case match(err, "min relay fee not met"):
```
Roasbeef, line 3830: Do we also want to attempt to catch the various RBF related errors? So eg: replacement transaction's fee isn't high enough.

wallet/wallet.go
```diff
@@ -3788,6 +3807,34 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 			backendError: err,
 		}
 
+	// Returned by bitcoind when a transaction does not meet the fee
+	// requirements to be accepted into mempool. This happens when the
+	// mempool reached its limits and is now purging low fee transactions.
+	// Increasing the mempool size limit (maxmempool) will solve this
+	// problem so we log the appropriate bitcoind config setting.
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L510
+	case match(err, "mempool min fee not met"):
```
Roasbeef, line 3816: btcd also has some RBF rejection logic, as well as the old priority and the static min fee setting: https://github.com/btcsuite/btcd/blob/9c16d23918b15c468c5647c388b9b7db3bc48dc7/mempool/mempool.go#L1137-L1171

ziggie1984: RBF rejection logic is already cover by the code I think? 
```
        // Returned by btcd when replacement transaction was rejected for
	// whatever reason.
	// https://github.com/btcsuite/btcd/blob/130ea5bddde33df32b06a1cdb42a6316eb73cff5/mempool/mempool.go#L841
	// https://github.com/btcsuite/btcd/blob/130ea5bddde33df32b06a1cdb42a6316eb73cff5/mempool/mempool.go#L854
	// https://github.com/btcsuite/btcd/blob/130ea5bddde33df32b06a1cdb42a6316eb73cff5/mempool/mempool.go#L875
	// https://github.com/btcsuite/btcd/blob/130ea5bddde33df32b06a1cdb42a6316eb73cff5/mempool/mempool.go#L896
	// https://github.com/btcsuite/btcd/blob/130ea5bddde33df32b06a1cdb42a6316eb73cff5/mempool/mempool.go#L913
	case match(err, "replacement transaction"):
		returnErr = &ErrReplacement{
			backendError: err,
		}
```

but added the 2 additional cases for btcd where we do not accept the transaction because of the minrelayfee according to the link you provided.

wallet/wallet.go
ziggie1984: ok I see its already remapped to `ErrDoubleSpend`: https://github.com/lightningnetwork/lnd/blob/master/lnwallet/btcwallet/btcwallet.go#L1168

wallet/wallet.go
```diff
@@ -3788,6 +3807,48 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
+	// requirements to be accepted into mempool because the policy of the
+	// mempool has a higher min relay fee. Default value for bitcoind is
+	// 1000 sat/kvbyte but this is configurable so we log the appropriate
+	// bitcoind config setting.
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L514
+	case match(err, "min relay fee not met"):
+		returnErr = &ErrMempoolFee{
+			backendError: err,
+		}
+		log.Infof("%v: tx not accepted into mempool because of low "+
```
yyforyongyu, line 3848: think this log is a bit redundant since we now have this explicit error.

yyforyongyu (approved):
LGTM👍

wallet/wallet.go
yyforyongyu: to me the error will be handled by the caller - either update the fee or just log it, so seems unnecessary.

ziggie1984: sounds good removed both cases of the logging.

bitromortac (approved):
LGTM :four_leaf_clover:

Roasbeef (approved):
LGTM 🎊

PR #872, opened 2023-06-14
chain: handle replacement and duplicate txes properly

yyforyongyu:
Previously when adding a replacement tx to the local mempool, when the txid's nested map becomes empty, we forgot to delete the key from the root map. And when adding the same tx to the `txids` map(which happens in ZMQ), we may treat the tx as a replacement tx. These are now fixed and the related unit tests are added/updated to reflect the fix.

### DIFF
--- a/chain/mempool.go
+++ b/chain/mempool.go
@@ -52,11 +52,19 @@ func (c *cachedInputs) addInput(op wire.OutPoint, txid chainhash.Hash) {
 	// Check if the input already exists.
 	oldTxid, existed := c.inputs[op]
 
-	// If existed, update the old txid to remove the input.
-	if existed {
+	// If the oldTxid exists and is different from the current txid, we
+	// need to update the oldTxid's inputs map.
+	isReplacement := false
+	if existed && oldTxid != txid {
+		isReplacement = true
+	}
+
+	// If the input is replaced, update the old txid to remove the input.
+	if isReplacement {
 		log.Tracef("Input %s was spent in tx %s, now spent in %s",
 			op, oldTxid, txid)
 
+		// Delete the input from the nested map under this old tx.
 		delete(c.txids[oldTxid], op)
 	}
 
--- a/chain/mempool.go
+++ b/chain/mempool.go
@@ -178,6 +178,12 @@ func (m *mempool) containsInput(op wire.OutPoint) (chainhash.Hash, bool) {
 //
 // NOTE: must be used inside a lock.
 func (m *mempool) add(tx *wire.MsgTx) {
+	// Skip coinbase inputs.
+	if blockchain.IsCoinBaseTx(tx) {
+		log.Debugf("Skipping coinbase tx %v", tx.TxHash())
+		return
+	}
+
 	hash := tx.TxHash()
 
 	// Add the txid to the mempool map.
@@ -268,12 +274,6 @@ func (m *mempool) removeInputs(tx chainhash.Hash) {
 //
 // NOTE: must be used inside a lock.
 func (m *mempool) updateInputs(tx *wire.MsgTx) {
-	// Skip coinbase inputs.
-	if blockchain.IsCoinBaseTx(tx) {
-		log.Debugf("Skipping coinbase tx %v", tx.TxHash())
-		return
-	}
-
 	// Iterate the tx's inputs.
 	for _, input := range tx.TxIn {
 		outpoint := input.PreviousOutPoint
--- a/chain/mempool.go
+++ b/chain/mempool.go
@@ -7,7 +7,6 @@ import (
 
 	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
-	"github.com/btcsuite/btcd/rpcclient"
 	"github.com/btcsuite/btcd/wire"
 	"golang.org/x/sync/errgroup"
 )
@@ -101,7 +100,7 @@ type mempool struct {
 	inputs *cachedInputs
 
 	// client is the rpc client that we'll use to query for the mempool.
-	client *rpcclient.Client
+	client rpcClient
 
 	// initFin is a channel that will be closed once the mempool has been
 	// initialized.
@@ -109,7 +108,7 @@ type mempool struct {
 }
 
 // newMempool creates a new mempool object.
-func newMempool(client *rpcclient.Client) *mempool {
+func newMempool(client rpcClient) *mempool {
 	return &mempool{
 		txs:     make(map[chainhash.Hash]bool),
 		inputs:  newCachedInputs(),
--- a/chain/mempool_test.go
+++ b/chain/mempool_test.go
@@ -1,6 +1,7 @@
 package chain
 
 import (
+	"math"
 	"testing"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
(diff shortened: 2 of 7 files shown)


chain/mempool.go
```diff
@@ -52,12 +52,25 @@ func (c *cachedInputs) addInput(op wire.OutPoint, txid chainhash.Hash) {
...
-	if existed {
+	if isReplacement {
 		log.Tracef("Input %s was spent in tx %s, now spent in %s",
 			op, oldTxid, txid)
 
+		// Delete the input from the nested map under this old tx.
 		delete(c.txids[oldTxid], op)
+
+		// If the old tx has no more inputs, delete it.
+		if len(c.txids[oldTxid]) == 0 {
```
Crypt-iQ, line 71: `removeInputsFromTx` will delete this, so it's not necessary

Crypt-iQ (commented):
LGTM besides one comment

chain/mempool.go
```diff
@@ -52,11 +52,19 @@ func (c *cachedInputs) addInput(op wire.OutPoint, txid chainhash.Hash) {
 	// Check if the input already exists.
 	oldTxid, existed := c.inputs[op]
 
+	// If the oldTxid exists and is different from the current txid, we
```
Roasbeef, line 54: Not 100% related to this diff: but can't we just use as integer as the key for the second level of the map? Since given the txid and the index, you already know the input. 

This could allow us to cut down on memory usage a bit, as in _practice_ transaction can't have up to 65k outputs. I think the practical limit is more like 4-5k outputs given current weight constraints.

chain/mempool.go
```diff
@@ -52,11 +52,19 @@ func (c *cachedInputs) addInput(op wire.OutPoint, txid chainhash.Hash) {
...
+	// need to update the oldTxid's inputs map.
+	isReplacement := false
+	if existed && oldTxid != txid {
+		isReplacement = true
+	}
+
 	// If existed, update the old txid to remove the input.
-	if existed {
+	if isReplacement {
```
Roasbeef, line 62: Update comment above?

chain/mempool_test.go
```diff
@@ -81,13 +128,76 @@ func TestCachedInputsAddInput(t *testing.T) {
...
+	require.Len(c.inputs, 1)
+
+	// Expect two transactions.
+	require.Len(c.txids, 2)
+
+	// The new txid should be present.
+	require.Len(c.txids[replacedTx.TxHash()], 1)
+
+	// The old txid should be an empty map.
+	require.Empty(c.txids[tx.TxHash()])
```
Roasbeef, line 159: When would this old `txid` be removed? IIUC, we'd only want to notify/play the new one, since that is what would be available from `GetRawTransactions`.

chain/mempool.go
```diff
@@ -178,6 +178,12 @@ func (m *mempool) containsInput(op wire.OutPoint) (chainhash.Hash, bool) {
 //
 // NOTE: must be used inside a lock.
 func (m *mempool) add(tx *wire.MsgTx) {
+	// Skip coinbase inputs.
```
Roasbeef, line 180: 👍

chain/mempool_test.go
```diff
@@ -81,13 +128,76 @@ func TestCachedInputsAddInput(t *testing.T) {
...
+}
+
+// TestCachedInputsAddInputTwice checks that when the same tx is added again it
+// won't change the cachedInputs's internal state.
+func TestCachedInputsAddInputTwice(t *testing.T) {
```
Roasbeef, line 164: Confirmed this test failed if the above diff is modified to not use `isReplacement`:
```
=== RUN   TestCachedInputsAddInputTwice
    mempool_test.go:202:
        	Error Trace:	/Users/roasbeef/gocode/src/github.com/btcsuite/btcwallet/chain/mempool_test.go:202
        	Error:      	"map[]" should have 1 item(s), but has 0
        	Test:       	TestCachedInputsAddInputTwice
```

I would've thought `TestCachedInputsAddInputReplacement` would fail as well though, as iiuc, the issue was that we weren't also removing the _old_ now replaced `txid`.

chain/bitcoind_zmq_events.go
```diff
@@ -358,9 +358,6 @@ func (b *bitcoindZMQEvents) txEventHandler() {
 				continue
 			}
 
-			// Add the tx to mempool.
```
Roasbeef, line 361: Typo in commit message: `LookupInputSpend` -> `LookupInputSpend`

Roasbeef: Also I don't follow how the second good case differs from the first bad case. What would cause the tx to be marked as seen in the bad case, but not the good case?

chain/mempool.go
```diff
@@ -101,15 +100,15 @@ type mempool struct {
 	inputs *cachedInputs
 
 	// client is the rpc client that we'll use to query for the mempool.
-	client *rpcclient.Client
+	client rpcClient
```
Roasbeef, line 103: 👍

chain/mocks_test.go
```diff
@@ -155,3 +156,23 @@ func (m *mockChainService) Stop() error {
 func (m *mockChainService) PeerByAddr(string) *neutrino.ServerPeer {
 	panic(errNotImplemented)
 }
+
+// mockRPCClient mocks the rpcClient interface.
+type mockRPCClient struct {
+	mock.Mock
```
Roasbeef, line 162: 👍

chain/mempool_test.go
yyforyongyu: fwiw the issue was when adding the same tx multiple times, we'd delete it every second time we added it.

chain/mempool.go
```diff
@@ -11,74 +11,168 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+// txIndex defines the transaction index type. We use uint16 here as the max
+// allowed output index cannot exceed 65535.
+type txIndex uint16
```
Crypt-iQ, line 16: I think a transaction can have more than 65535 outputs if:
- it's non-standard
- it has many `OP_RETURN` outputs with small data pushes
- it takes up the entire block

For this reason, I don't think we can have this be a `uint16` because this type of transaction would cause a panic on all LND nodes. Another interesting tidbit is that this tx would be disallowed from being a channel open by LN if it included the funding output since it exceeds the 2-byte output index.

Crypt-iQ (commented):
IMO only the first two commits should go in 0.17. The other commits would take more review time than I have and aren't as critical as the memory leak fix.

yyforyongyu:
> IMO only the first two commits should go in 0.17. The other commits would take more review time than I have and aren't as critical as the memory leak fix.

You mean the first three? The redundant `Add` needs to be removed otherwise the mempool will keep growing between polling intervals.

chain/mempool.go
yyforyongyu: Cool will move this into a different PR.

chain/bitcoind_zmq_events.go
yyforyongyu: I think this needs to be here otherwise the mempool will have confirmed txes stored?

Crypt-iQ: I think the next run of the mempool poller will get rid of it, which would make it not that much of an issue (still not ideal). Where does `filterTx` mark a tx as seen -- I can't find it?

yyforyongyu: > Where does filterTx mark a tx as seen 

[Here](https://github.com/btcsuite/btcwallet/blob/master/chain/bitcoind_client.go#L1386) when it puts it into its(yet another...) mempool. And when this tx shows up again as unconfirmed it won't be notified because [here](https://github.com/btcsuite/btcwallet/blob/master/chain/bitcoind_client.go#L1301).

chain/mempool_test.go
```diff
@@ -312,3 +313,93 @@ func TestMempool(t *testing.T) {
...
+			{PreviousOutPoint: op3},
+		},
+	}
+
+	// Now add all the transactions.
+	m.add(tx0)
+	m.add(tx1)
+	m.add(tx2)
+
+	// Check transactions are updated, mempool should now contain the three
```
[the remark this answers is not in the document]
yyforyongyu: fixed

yyforyongyu:
Updated the PR to focus on the memory leak fix. The last commit adds a unit test to check that the mempool state is updated as expected, and the only change is the newly added interface so I think it belongs here. cc @Crypt-iQ

Crypt-iQ (approved):

chain/mempool.go
```diff
@@ -101,15 +100,15 @@ type mempool struct {
 	inputs *cachedInputs
 
 	// client is the rpc client that we'll use to query for the mempool.
-	client *rpcclient.Client
```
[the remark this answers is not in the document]
yyforyongyu: Added!

chain/mempool.go
Roasbeef: Ah true about non-standard tricks (eg: you can only have a single `OP_RETURN` in a transaction with standardness). So we'll either need to bump up that integer size, or try to roll something a bit more custom (basically a series of byte arrays with custom limbs).

Roasbeef (approved):
LGTM 🐕

PR #873, opened 2023-06-19
Optimize mempool maps

yyforyongyu:
Depends on #872, this PR removes a redundant `Add(tx)` for bitcoind ZMQ events, and optimized the memory usage of mempool based on the fact that, multiple inputs may share the same tx hash that leaves room for optimization.

### DIFF
--- a/chain/bitcoind_zmq_events.go
+++ b/chain/bitcoind_zmq_events.go
@@ -358,9 +358,6 @@ func (b *bitcoindZMQEvents) txEventHandler() {
 				continue
 			}
 
-			// Add the tx to mempool.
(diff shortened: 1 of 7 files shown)


yyforyongyu:
Unfortunately the optimization doesn't work based pprof, will investigate more.

yyforyongyu:
Replaced by #877 which focuses on removing the redundant `Add`. I don't think the maps in `mempool` can be optimized more, given **most of the inputs don't share the same tx hash, and most spending txes are consuming inputs from different parent txes**. Plus the space overhead of using nested maps, my pprof says it actually takes more space.

PR #878, opened 2023-06-22
wallet: refactor broadcast error code.

ziggie1984:
This relates to https://github.com/lightninglabs/neutrino/pull/278 and https://github.com/lightningnetwork/lnd/pull/7746

I will follow @Roasbeef advice reuse the code in the Rebroadcaster Neutrino package. For this we need to refactor the code of the error checking so that we do not depend on the wallet. 
This lead to the introduction of 2 new error types to separate the logging form the error mapping.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3560,6 +3560,42 @@ func (e *ErrMempoolFee) Unwrap() error {
 	return e.backendError
 }
 
+// ErrAlreadyConfirmed is an error returned from PublishTransaction in case
+// a transaction is already confirmed in the blockchain.
+type ErrAlreadyConfirmed struct {
+	backendError error
+}
+
+// Error returns the string representation of ErrAlreadyConfirmed.
+//
+// NOTE: Satisfies the error interface.
+func (e *ErrAlreadyConfirmed) Error() string {
+	return fmt.Sprintf("tx already confirmed: %v", e.backendError)
+}
+
+// Unwrap returns the underlying error returned from the backend.
+func (e *ErrAlreadyConfirmed) Unwrap() error {
+	return e.backendError
+}
+
+// ErrInMempool is an error returned from PublishTransaction in case a
+// transaction is already in the mempool.
+type ErrInMempool struct {
+	backendError error
+}
+
+// Error returns the string representation of ErrInMempool.
+//
+// NOTE: Satisfies the error interface.
+func (e *ErrInMempool) Error() string {
+	return fmt.Sprintf("tx already in mempool: %v", e.backendError)
+}
+
+// Unwrap returns the underlying error returned from the backend.
+func (e *ErrInMempool) Unwrap() error {
+	return e.backendError
+}
+
 // PublishTransaction sends the transaction to the consensus RPC server so it
 // can be propagated to other nodes and eventually mined.
 //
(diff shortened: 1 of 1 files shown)


saubyk:
cc:  @yyforyongyu for review

wallet/wallet.go
```diff
@@ -3560,6 +3560,42 @@ func (e *ErrMempoolFee) Unwrap() error {
...
+// a transaction is already confirmed in the blockchain.
+type ErrAlreadyConfirmed struct {
+	backendError error
+}
+
+// Error returns the string representation of ErrAlreadyConfirmed.
+//
+// NOTE: Satisfies the error interface.
+func (e *ErrAlreadyConfirmed) Error() string {
+	return fmt.Sprintf("mempool fee not met: %v", e.backendError)
```
yyforyongyu, line 3573: the message is wrong.

wallet/wallet.go
```diff
@@ -4095,3 +3972,186 @@ func OpenWithRetry(db walletdb.DB, pubPass []byte, cbs *waddrmgr.OpenCallbacks,
...
+	// already confirmed to a btcd node over the P2P network.
+	// https://github.com/btcsuite/btcd/blob/130ea5bddde33df32b06a1cdb42a6316eb73cff5/mempool/mempool.go#L1036
+	case match(err, "transaction already exists"):
+		fallthrough
+
+	// This error is returned when broadcasting a transaction that has
+	// already confirmed to a bitcoind node over the P2P network.
+	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L648
+	case match(err, "txn-already-known"):
+		return &ErrAlreadyConfirmed{}
```
yyforyongyu, line 4037: assign the `backendError` here?

wallet/wallet.go
```diff
@@ -4095,3 +3972,186 @@ func OpenWithRetry(db walletdb.DB, pubPass []byte, cbs *waddrmgr.OpenCallbacks,
 
 	return w, nil
 }
+
+// MapBroadcastBackendError maps the different backend errors when broadcasting
+// a transaction to the bitcoin network to an internal error type.
+func MapBroadcastBackendError(err error) error {
```
yyforyongyu, line 3978: Is it intended to export this method?

yyforyongyu (changes requested):
Thanks for the PR! Left some comments and nits.

wallet/wallet.go
```diff
@@ -3655,70 +3691,24 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
-	// https://github.com/btcsuite/btcd/blob/130ea5bddde33df32b06a1cdb42a6316eb73cff5/mempool/mempool.go#L953
-	case match(err, "already have transaction"):
-		fallthrough
+	returnErr = MapBroadcastBackendError(err)
 
-	// This error is returned when broadcasting a transaction to a bitcoind
-	// node that already has it in their mempool.
-	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/validation.cpp#L590
-	case match(err, "txn-already-in-mempool"):
+	switch returnErr.(type) {
```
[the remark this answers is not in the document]
ziggie1984: not sure, in this case it would work because the error mapping does only return an empty error type without including the backend error for those two examples, but this could easily change so I think we would need to use errors.As instead which brings a bit of overhead because we would need to declare the variables as well because errors.As does only accept a non-nil target, so I would keep it as is ? Or should we always try to use errors.As instead of the switch err.(type) statement?

wallet/wallet.go
ziggie1984: yes because we wanna reuse the mapping code for the Rebroadcaster which is currently in the neutrino package.

wallet/wallet.go
ziggie1984: added it for all cases.

wallet/wallet.go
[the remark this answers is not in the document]
Roasbeef: > Or should we always try to use errors.As instead of the switch err.(type) statement?

I think the "modern Go" best practice is to prefer `errors.As` as it can handle wrapped errors, while type assertions can only go one level deep.

wallet/wallet.go
Roasbeef: Yep, goal here is to be able to have the rebroadcaster gain a custom hook to handle errors other than p2p errors.

wallet/wallet.go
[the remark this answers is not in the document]
Roasbeef: This would look something like this with `errors.As`:
```go
returnErr = MapBroadcastBackendError(err)

var errInMempool *ErrInMempool
var errAlreadyConfirmed *ErrAlreadyConfirmed

switch {
case errors.As(returnErr, &errInMempool):
	log.Infof("%v: tx already in mempool", txid)
	return &txid, nil

case errors.As(returnErr, &errAlreadyConfirmed):
	dbErr := walletdb.Update(w.db, func(dbTx walletdb.ReadWriteTx) error {
		txmgrNs := dbTx.ReadWriteBucket(wtxmgrNamespaceKey)
		txRec, err := wtxmgr.NewTxRecordFromMsgTx(tx, time.Now())
		if err != nil {
			return err
		}
		return w.TxStore.RemoveUnminedTx(txmgrNs, txRec)
	})
	if dbErr != nil {
		log.Warnf("Unable to remove confirmed transaction %v "+
			"from unconfirmed store: %v", tx.TxHash(), dbErr)
	}

	log.Infof("%v: tx already confirmed", txid)

	return &txid, nil
}
```

Roasbeef (approved):
Diff looks good to me, compared side by side to ensure all the errors are upheld with their existing paths. Just a minor thing re a space, and preferred usage of `errors.As`.

wallet/wallet.go
```diff
@@ -3560,6 +3560,42 @@ func (e *ErrMempoolFee) Unwrap() error {
...
+
+// Error returns the string representation of ErrAlreadyConfirmed.
+//
+// NOTE: Satisfies the error interface.
+func (e *ErrAlreadyConfirmed) Error() string {
+	return fmt.Sprintf("tx already confirmed: %v", e.backendError)
+}
+
+// Unwrap returns the underlying error returned from the backend.
+func (e *ErrAlreadyConfirmed) Unwrap() error {
```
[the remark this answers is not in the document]
ziggie1984: Currently errors like "insufficient fee" are classified as `ErrReplacement` which we map in lnd as `DoubleSpend` and remove the transaction. So I think all errors we anticipated have a Category now, the rest is treated as unknown backend error.

PR #882, opened 2023-07-13
chain: fix error match typo

yyforyongyu:
Fix a typo found in error string matching.

### DIFF
--- a/chain/mempool.go
+++ b/chain/mempool.go
@@ -15,7 +15,7 @@ import (
 
 // txNotFoundErr is an error returned from bitcoind's `getrawtransaction` RPC
 // when the requested txid cannot be found.
-const txNotFoundErr = "-5: No such mempool or blockchain Transaction"
+const txNotFoundErr = "-5: No such mempool or blockchain transaction"
 
 // cachedInputs caches the inputs of the transactions in the mempool. This is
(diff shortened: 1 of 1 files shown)


chain/mempool.go
```diff
@@ -15,7 +15,7 @@ import (
 
 // txNotFoundErr is an error returned from bitcoind's `getrawtransaction` RPC
 // when the requested txid cannot be found.
-const txNotFoundErr = "-5: No such mempool or blockchain Transaction"
+const txNotFoundErr = "-5: No such mempool or blockchain transaction"
```
Roasbeef, line 18: IIUC, we do a `ToLower()` in any case?

Roasbeef: Ah looks like not, perhaps we should add that for future proofing?

yyforyongyu: you mean add a test?

guggero: > you mean add a test?

That as well. But I think Laolu meant just doing `strings.ToLower()` on both strings before comparing them, to avoid such issues in the future.

yyforyongyu: oh gotya!

Roasbeef (approved):
LGTM 🥩

PR #885, opened 2023-08-05
wallet: allow manually setting the wallet birthday

### LINKED ISSUE #883: Allow setting the start block for address rescans
#### Description

In the event that the wallet.db file is deleted, the same wallet can be restored using the seed. Addresses that need their transactions indexed can be added to the wallet with calls to `getnewaddress`. Upon restarting, btcwallet will perform a rescan on those addresses.

The issue here is that the rescan begins from the "birthdate" of when the addresses were added to the wallet. So if they are newly added, their older transaction history is missed during the rescan.

#### To Reproduce

1. Create a btcwallet with `btcwallet --create`
2. Make `getnewaddress` call to add an address to the wallet
3. Fund the return address and verify that btcwallet recognized the deposit
4. Delete the wallet.db file
5. Repeat steps 1-3 (may have to wait ~30min before repeating)

btcwallet will think the balance is zero and that the addresses have no transactions because the rescan started at a block above the block that the funding transaction was in.

#### Tasks

Add a flag to specify the blockhash/blockheight where the rescan should start. This flag should make the address creation date irrelevant to where to start the rescan, which would allow newly created addresses with past transactions to be properly accounted for.

josephawallace:
Adds a flag that allows for the wallet's birthday to be specified when being created. This gives the user control over where rescans begin for a BTC wallet that is recovered with a seed. This is meant to resolve #883.

### DIFF
--- a/config.go
+++ b/config.go
@@ -43,6 +43,7 @@ var (
 	defaultRPCKeyFile  = filepath.Join(defaultAppDataDir, "rpc.key")
 	defaultRPCCertFile = filepath.Join(defaultAppDataDir, "rpc.cert")
 	defaultLogDir      = filepath.Join(defaultAppDataDir, defaultLogDirname)
+	defaultBirthday    = time.Now().Unix()
 )
 
 type config struct {
@@ -65,6 +66,7 @@ type config struct {
 
 	// Wallet options
 	WalletPass string `long:"walletpass" default-mask:"-" description:"The public wallet password -- Only required if the wallet was created with one"`
+	Birthday   int64  `long:"birthday" default-mask:"-" description:"Specify a unix timestamp to be the wallet birthday -- Will be used as the approximate starting time for wallet rescans"`
 
 	// RPC client options
 	RPCConnect       string                  `short:"c" long:"rpcconnect" description:"Hostname/IP and port of btcd RPC server to connect to (default localhost:8334, testnet: localhost:18334, simnet: localhost:18556)"`
@@ -278,6 +280,7 @@ func loadConfig() (*config, []string, error) {
 		BanDuration:            neutrino.BanDuration,
 		BanThreshold:           neutrino.BanThreshold,
 		DBTimeout:              wallet.DefaultDBTimeout,
+		Birthday:               defaultBirthday,
 	}
 
 	// Pre-parse the command line options to see if an alternative config
--- a/rpc/rpcserver/server.go
+++ b/rpc/rpcserver/server.go
@@ -694,13 +694,13 @@ func (s *loaderServer) CreateWallet(ctx context.Context, req *pb.CreateWalletReq
 	}
 
 	// Use the current time for the birthday when the request's is empty.
-	bday := req.Birthday
-	if bday == 0 {
-		bday = time.Now().Unix()
+	bday := time.Now()
+	if req.Birthday != 0 {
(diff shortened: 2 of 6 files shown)


rpc/rpcserver/server.go
```diff
@@ -693,8 +693,14 @@ func (s *loaderServer) CreateWallet(ctx context.Context, req *pb.CreateWalletReq
 		pubPassphrase = []byte(wallet.InsecurePubPassphrase)
 	}
 
+	// Use the current time for the birthday when the request's is empty.
+	bday := req.Birthday
```
guggero, line 697: How about:
```
bday := time.Now()
if req.Birthday != 0 {
    bday := time.Unix(req.Birthday, 0)
}
```

That then shortens the call below.

config.go
```diff
@@ -43,6 +43,7 @@ var (
 	defaultRPCKeyFile  = filepath.Join(defaultAppDataDir, "rpc.key")
 	defaultRPCCertFile = filepath.Join(defaultAppDataDir, "rpc.cert")
 	defaultLogDir      = filepath.Join(defaultAppDataDir, defaultLogDirname)
+	defaultBirthday    = time.Now().Unix()
```
guggero, line 46: Shouldn't the default be a very early timestamp instead? Otherwise a user might fear their coins are gone if they don't know to set this parameter and are trying to restore from an old seed.

sample-btcwallet.conf
```diff
@@ -15,6 +15,12 @@
 ; directory for mainnet and testnet wallets, respectively.
 ; appdata=~/.btcwallet
 
+; Unix timestamp representing when the wallet was created, and therefore
+; the earliest time the wallet could have participated in any transactions.
+; This timestamp will be used to estimate a block height that is an appropriate
+; starting place to scan for transactions related to this wallet.
+; birthday=1231006505
```
guggero, line 22: Should mention the default value/default behavior here.

walletsetup.go
```diff
@@ -188,7 +188,7 @@ func createWallet(cfg *config) error {
 	}
 
 	fmt.Println("Creating the wallet...")
-	w, err := loader.CreateNewWallet(pubPass, privPass, seed, time.Now())
+	w, err := loader.CreateNewWallet(pubPass, privPass, seed, time.Unix(cfg.Birthday, 0))
```
guggero, line 191: Again, I'm not familiar with `btcwallet` as a standalone wallet. When is this code run vs. the `CreateWallet` RPC call?

guggero (commented):
Just some thoughts, nothing blocking IMO.

config.go
josephawallace: I used the current time as the default because that was what was previously used whenever a new wallet was created:
```
fmt.Println("Creating the wallet...")
w, err := loader.CreateNewWallet(pubPass, privPass, seed, time.Now())
if err != nil {
    return err
}
```
--->
```
fmt.Println("Creating the wallet...")
w, err := loader.CreateNewWallet(pubPass, privPass, seed, time.Unix(cfg.Birthday, 0))
if err != nil {
    return err
}
```

walletsetup.go
josephawallace: When being used as a standalone tool, you can create a wallet from the command line with `btcwallet --create`, which will run that code in that snippet

PR #886, opened 2023-08-30
chain: use `GetRawTransactionLazy` in mempool poller

yyforyongyu:
Depends on https://github.com/btcsuite/btcd/pull/2027

We de-prioritize the RPC calls in mempool in case it blocks `bitcoind`'s rpc workers.


yyforyongyu:
Closed due [comments](https://github.com/btcsuite/btcd/pull/2027#issuecomment-1704577429).

PR #887, opened 2023-09-07
chain: detect and use gettxspendingprevout RPC instead of mempool

Crypt-iQ:
This change uses the gettxspendingprevout RPC if the bitcoind version is greater or equal to version 24.0.0. In this case, the mempool is no longer used for zmq clients. The original behavior must be kept for rpc polling clients since they notify on new transactions and can only learn about them from polling the bitcoind mempool and keeping track of its own btcwallet mempool.

### DIFF
--- a/chain/bitcoind_events.go
+++ b/chain/bitcoind_events.go
@@ -1,6 +1,7 @@
 package chain
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
@@ -62,5 +63,39 @@ func NewBitcoindEventSubscriber(cfg *BitcoindConfig,
 			"rpcpolling is disabled")
 	}
 
-	return newBitcoindZMQEvents(cfg.ZMQConfig, client)
+	// Check if the bitcoind node is on a version that has the
+	// gettxspendingprevout RPC. If it does, then we don't need to maintain
+	// a mempool for ZMQ clients.
+	hasRPC, err := hasSpendingPrevoutRPC(client)
+	if err != nil {
+		return nil, err
+	}
+
+	return newBitcoindZMQEvents(cfg.ZMQConfig, client, hasRPC)
+}
+
+// hasSpendingPrevoutRPC returns whether or not the bitcoind has the newer
+// gettxspendingprevout RPC.
+func hasSpendingPrevoutRPC(client *rpcclient.Client) (bool, error) {
+	// Fetch the bitcoind version.
+	resp, err := client.RawRequest("getnetworkinfo", nil)
+	if err != nil {
+		return false, err
+	}
+
+	info := struct {
+		Version int64 `json:"version"`
+	}{}
+
+	if err := json.Unmarshal(resp, &info); err != nil {
+		return false, err
+	}
+
+	// Bitcoind returns a single value representing the semantic version:
+	// 10000 * CLIENT_VERSION_MAJOR + 100 * CLIENT_VERSION_MINOR
+	// + 1 * CLIENT_VERSION_BUILD
+	//
+	// The gettxspendingprevout call was added in version 24.0.0, so we
+	// return for versions >= 240000.
+	return info.Version >= 240000, nil
 }
--- a/chain/bitcoind_zmq_events.go
+++ b/chain/bitcoind_zmq_events.go
@@ -2,6 +2,7 @@ package chain
 
 import (
 	"bytes"
+	"encoding/json"
 	"fmt"
 	"io"
 	"math/rand"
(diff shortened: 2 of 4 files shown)


chain/bitcoind_zmq_events.go
```diff
@@ -174,16 +189,85 @@ func (b *bitcoindZMQEvents) BlockNotifications() <-chan *wire.MsgBlock {
...
+// gettxspendingprevout call.
+type getTxSpendingPrevOutReq struct {
+	Txid string `json:"txid"`
+	Vout uint32 `json:"vout"`
+}
+
+// getTxSpendingPrevOutResp is the rpc response format for bitcoind's
+// gettxspendingprevout call. It returns the "spendingtxid" if one exists in
+// the mempool.
+type getTxSpendingPrevOutResp struct {
```
Roasbeef, line 202: Can also add these to `btcjson` assuming the fields are stable-ish. Also see that other issue to refactor things using tpe params so we can just do things like pass this type in and have it handle the rest of tbe boiler plate.

chain/bitcoind_conn.go
```diff
@@ -174,6 +175,14 @@ func NewBitcoindConn(cfg *BitcoindConfig) (*BitcoindConn, error) {
 		}
 	}
 
+	// Check if the bitcoind node is on a version that has the
+	// gettxspendingprevout RPC. If it does, then we don't need to maintain
+	// a mempool for ZMQ clients.
+	hasRPC, err := hasSpendingPrevoutRPC(client)
```
Roasbeef, line 181: What about the non-zmq variant? In that case, the mempool is only useful for detecting unconf transactions, and we don't need to periodically reconcile.

Roasbeef (changes requested):

chain/bitcoind_conn.go
Crypt-iQ: It already has to poll the mempool and build a local one to not give duplicate notifications, so the call to `LookupInputSpend` just checks the btcwallet mempool. For that reason, I don't think it makes sense to use the new RPC?

chain/bitcoind_zmq_events.go
```diff
@@ -174,16 +189,85 @@ func (b *bitcoindZMQEvents) BlockNotifications() <-chan *wire.MsgBlock {
...
+	// Otherwise, we aren't maintaining a mempool and can use the
+	// gettxspendingprevout RPC. Create an RPC-style prevout that will be
+	// the lone item in an array.
+	prevoutReq := &getTxSpendingPrevOutReq{
+		Txid: op.Hash.String(), Vout: op.Index,
+	}
+
+	// The RPC takes an array of prevouts so we have an array with a single
+	// item since we don't yet batch calls to LookupInputSpend.
+	var prevoutArr [1]*getTxSpendingPrevOutReq
```
guggero, line 230: Why not just:
```
prevoutArr := []*getTxSpendingPrevOutReq{prevoutReq}
```

guggero (commented):
Looks quite good, should massively help with performance and memory usage.
Just one style related comment, pending @Roasbeef's comments.

chain/bitcoind_zmq_events.go
Crypt-iQ: thanks, done

chain/bitcoind_zmq_events.go
```diff
@@ -122,22 +129,30 @@ func newBitcoindZMQEvents(cfg *ZMQConfig,
...
+
+	return zmqEvents, nil
 }
 
 // Start spins off the bitcoindZMQEvent goroutines.
 func (b *bitcoindZMQEvents) Start() error {
 	// Load the mempool so we don't miss transactions.
-	if err := b.mempool.LoadMempool(); err != nil {
-		return err
+	if b.mempool != nil {
```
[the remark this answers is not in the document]
Crypt-iQ: think I did this

chain/bitcoind_events.go
```diff
@@ -36,7 +36,7 @@ var _ rpcClient = (*rpcclient.Client)(nil)
 // NewBitcoindEventSubscriber initialises a new BitcoinEvents object impl
 // depending on the config passed.
 func NewBitcoindEventSubscriber(cfg *BitcoindConfig,
-	client *rpcclient.Client) (BitcoindEvents, error) {
+	client *rpcclient.Client, hasRPC bool) (BitcoindEvents, error) {
```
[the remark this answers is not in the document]
Crypt-iQ: added

chain/bitcoind_zmq_events.go
Crypt-iQ: which issue to refactor things? also, does this mean it would be a new `btcjson` command?

Roasbeef: So this issue: https://github.com/btcsuite/btcd/issues/1934

Non-blocking though.

Roasbeef (approved):
LGTM 🐩

PR #889, opened 2023-09-22
chain: use neutrino filters to speed up bitcoind seed recovery

Roasbeef:
In this commit, we use neutrino filters to speed up bitcoind seed
recovery. We use the recently created `maybeShouldFetchBlock` function
to check the filters to see if we need to fetch a block at all. This
saves us from fetching, decoding, then scanning the block contents if we
know nothing is present in them.

At this point, we can also further consolidate the `FilterBlocks`
methods between the two backends, as they're now identical.

Another follow up here would be to start to prefetch both filters and blocks. We know that we'll need to fetch every filter, so we can fetch them all in a single RPC request (batch call). Once we have the filters, we can sort of check them all in parallel, but need to mind the `FoundOutPoints` field, as once we have something like a change addr we know of, we want to then watch that for subsequent spends.


chain/btcd.go
```diff
@@ -197,6 +197,44 @@ func (c *RPCClient) BlockStamp() (*waddrmgr.BlockStamp, error) {
...
 }
 
+// maybeShouldFetchBlock returns true if the contents of a block *might* have
+// some items that match our watch list. This uses the neutrino filters to do a
+// quick loop up to see if the block has relevant items to avoid downloading
+// the full block.
+func maybeShouldFetchBlock(rpc *rpcclient.Client, blk wtxmgr.BlockMeta,
+	watchList [][]byte) (bool, error) {
+
+	rawFilter, err := rpc.GetCFilter(&blk.Hash, wire.GCSFilterRegular)
```
guggero, line 207: If block filters are turned off in `bitcoind`, will this return an error? Maybe we need to do a feature discovery first.

Roasbeef: True, we only want to try if the bitcoind node has it.

Roasbeef: Updated! Turns out bitcoind also has a different name for the RPC to fetch the filter, so added another layer of abstraction to handle that.

guggero (commented):
Needs a couple of fixes. Was able to get it running with a small patch and can confirm this massively speeds up scanning blocks.

My quick benchmark, scanning 2000 blocks took:
Without this PR: ~1m 44s
With this PR: ~26s

PR #890, opened 2023-09-23
wallet: add gettransaction

ErikEk:
This PR adds a method to the wallet that allows to get transaction data on any transaction given its ID. Useful for (https://github.com/lightningnetwork/lnd/pull/7654).

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -77,6 +77,9 @@ var (
 	// to true.
 	ErrTxLabelExists = errors.New("transaction already labelled")
 
+	// ErrNoTx is returned when a transaction can not be found.
+	ErrNoTx = errors.New("can not find transaction")
+
 	// ErrTxUnsigned is returned when a transaction is created in the
 	// watch-only mode where we can select coins but not sign any inputs.
 	ErrTxUnsigned = errors.New("watch-only wallet, transaction not signed")
@@ -2462,6 +2465,56 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
 		return w.TxStore.RangeTransactions(txmgrNs, start, end, rangeFn)
 	})
 	return &res, err
+}
+
+// GetTransactionResult returns a summary of the transaction along with
+// other block properties.
+type GetTransactionResult struct {
+	Summary       TransactionSummary
+	Height        int32
+	BlockHash     *chainhash.Hash
+	Confirmations int32
+	Timestamp     int64
+}
+
+// GetTransaction returns detailed data of a transaction given its id. In addition it
+// returns properties about its block.
+func (w *Wallet) GetTransaction(txHash chainhash.Hash) (*GetTransactionResult,
+	error) {
+
+	var res GetTransactionResult
+	err := walletdb.View(w.db, func(dbtx walletdb.ReadTx) error {
+		txmgrNs := dbtx.ReadBucket(wtxmgrNamespaceKey)
+
+		txDetail, err := w.TxStore.TxDetails(txmgrNs, &txHash)
+		if err != nil {
+			return err
+		}
+
+		// If the transaction was not found we return an error.
+		if txDetail == nil {
+			return fmt.Errorf("%w: txid %v", ErrNoTx, txHash)
+		}
+
+		res = GetTransactionResult{
+			Summary:       makeTxSummary(dbtx, w, txDetail),
+			Timestamp:     txDetail.Block.Time.Unix(),
+			Confirmations: txDetail.Block.Height,
+		}
+
+		// If it is a confirmed transaction we set the corresponding
+		// block height and hash.
+		if txDetail.Block.Height != -1 {
+			res.Height = txDetail.Block.Height
+			res.BlockHash = &txDetail.Block.Hash
+		}
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &res, nil
 }
 
 // AccountResult is a single account result for the AccountsResult type.
--- a/wallet/wallet_test.go
+++ b/wallet/wallet_test.go
@@ -7,14 +7,25 @@ import (
 
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/wtxmgr"
+	"github.com/stretchr/testify/require"
 
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
 )
 
 var (
 	TstSerializedTx, _ = hex.DecodeString("010000000114d9ff358894c486b4ae11c2a8cf7851b1df64c53d2e511278eff17c22fb7373000000008c493046022100995447baec31ee9f6d4ec0e05cb2a44f6b817a99d5f6de167d1c75354a946410022100c9ffc23b64d770b0e01e7ff4d25fbc2f1ca8091053078a247905c39fce3760b601410458b8e267add3c1e374cf40f1de02b59213a82e1d84c2b94096e22e2f09387009c96debe1d0bcb2356ffdcf65d2a83d4b34e72c62eccd8490dbf2110167783b2bffffffff0280969800000000001976a914479ed307831d0ac19ebc5f63de7d5f1a430ddb9d88ac38bfaa00000000001976a914dadf9e3484f28b385ddeaa6c575c0c0d18e9788a88ac00000000")
 	TstTx, _           = btcutil.NewTxFromBytes(TstSerializedTx)
 	TstTxHash          = TstTx.Hash()
+
+	TstMinedTxBlockHeight        = int32(279143)
+	TstMinedSignedTxBlockDetails = &wtxmgr.BlockMeta{
+		Block: wtxmgr.Block{
+			Hash:   *TstTxHash,
+			Height: TstMinedTxBlockHeight,
+		},
+		Time: time.Now(),
+	}
 )
 
 // TestLocateBirthdayBlock ensures we can properly map a block in the chain to a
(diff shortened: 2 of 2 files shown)


wallet/wallet.go
```diff
@@ -2464,6 +2464,73 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+			blockHash = &txDetail.Block.Hash
+		}
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var res GetTransactionResult
```
guggero, line 2515: Have this be the only variable outside of the `walletdb.View` call and move the `switch` below inside the callback? That way it's more clear what the result will be if `txDetail` is nil.

guggero (commented):
Thanks for the addition, should definitely be useful for different use cases in `lnd`.

wallet/wallet.go
```diff
@@ -2464,6 +2464,73 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+// structure for a mined transaction.
+type GetTransactionResult struct {
+	MinedTransaction   Block
+	UnminedTransaction TransactionSummary
+}
+
+// GetTransaction returns data of any transaction given its id. A mined
+// transaction is returned in a Block structure which records properties about
+// the block.
+func (w *Wallet) GetTransaction(txHash *chainhash.Hash) (*GetTransactionResult,
```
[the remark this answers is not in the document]
ErikEk: Yes, that makes sence.

wallet/wallet.go
ErikEk: Great. I'll do that and return an error if the transaction isn't found.

ErikEk: Done. Now I return an error if the transaction isn't found instead.

wallet/wallet.go
```diff
@@ -2464,6 +2464,66 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+			res.MinedTransaction = Block{
+				Hash:         blockHash,
+				Height:       bestHeight,
+				Timestamp:    summary.Timestamp,
+				Transactions: []TransactionSummary{summary},
+			}
+		}
+
+		return nil
+	})
```
guggero, line 2522: I think we should do an explicit `return nil, err` if the err is not nil here.

wallet/wallet.go
```diff
@@ -2464,6 +2464,66 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+		if txDetail == nil {
+			return fmt.Errorf("can not find the transaction")
+		}
+
+		// Otherwise, we create a summary of the transaction details.
+		summary := makeTxSummary(dbtx, w, txDetail)
+
+		var bestHeight int32
+		var blockHash *chainhash.Hash
+		// If it is a confirmed transaction we set the corresponding
```
guggero, line 2500: nit: add newline before comment.

wallet/wallet.go
```diff
@@ -2464,6 +2464,66 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+
+		// If the transaction was not found we return an error.
+		if txDetail == nil {
+			return fmt.Errorf("can not find the transaction")
+		}
+
+		// Otherwise, we create a summary of the transaction details.
+		summary := makeTxSummary(dbtx, w, txDetail)
+
+		var bestHeight int32
```
guggero, line 2498: nit: could group multiple `var` statements in a single block.
for example:
```
var (
    bestHeight int32
    blockHash *chainhash.Hash
    summary = makeTxSummary(...)
)
```

wallet/wallet.go
```diff
@@ -2464,6 +2464,66 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+// structure for a mined transaction.
+type GetTransactionResult struct {
+	MinedTransaction   Block
+	UnminedTransaction TransactionSummary
+}
+
+// GetTransaction returns data of any transaction given its id. A mined
+// transaction is returned in a Block structure which records properties about
+// the block.
+func (w *Wallet) GetTransaction(txHash chainhash.Hash) (*GetTransactionResult,
```
guggero, line 2478: Can we add a unit test for this please?

guggero (commented):
Getting pretty close!
Main ask is a unit test.

wallet/wallet.go
ErikEk: Alright. Will add a test.

yyforyongyu:
feel free to ping me when it's ready for review

ErikEk:
@yyforyongyu Just pushed the changes. Please take a look.

wallet/wallet.go
```diff
@@ -2464,6 +2464,69 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
 	return &res, err
 }
 
+// GetTransactionResult returns a transaction in the UnminedTransaction
+// field if it is unmined, or in the MinedTransaction field as a Block
+// structure for a mined transaction.
+type GetTransactionResult struct {
```
yyforyongyu, line 2470: Just thinking about the use case of this method, it's easier if we just give a flat result back, something like,
```go
type GetTransactionResult struct {
	TransactionSummary

	// Height...
	Height int32

	// BlockHash...
	BlockHash chainhash.Hash
}
```
Callers can check the `Height` field to decide whether it's mined or not, or, if we wanna be explicit, we can add one more field, something like,

```go
type GetTransactionResult struct {
	TransactionSummary

	// Height...
	Height int32

	// BlockHash...
	BlockHash chainhash.Hash

	// Confirmations...
	Confirmations uint32
```

What do you think? This way we can also get rid of the field `Transactions: []TransactionSummary{summary}` which can be confusing.

wallet/wallet_test.go
```diff
@@ -205,3 +206,43 @@ func TestLabelTransaction(t *testing.T) {
...
+	t.Parallel()
+
+	w, cleanup := testWallet(t)
+	defer cleanup()
+
+	// Create a new tx record.
+	rec, err := wtxmgr.NewTxRecord(
+		TstSerializedTx, time.Now(),
+	)
+	if err != nil {
```
yyforyongyu, line 222: We are pushing the usage of `require` atm, so we wanna do `require.NoError` here and below.

wallet/wallet_test.go
```diff
@@ -205,3 +206,43 @@ func TestLabelTransaction(t *testing.T) {
 		})
 	}
 }
+
+// TestGetTransaction writes a tx to disk and then compares
+// the hash to the response from the GetTransaction function.
+func TestGetTransaction(t *testing.T) {
```
yyforyongyu, line 212: The test could be more comprehensive, I think we should at least test three cases,
- when it's not found
- when it's not mined
- when it's confirmed

yyforyongyu (changes requested):
Thanks for the PR! Left some comments about tests and the returned format.

wallet/wallet_test.go
ErikEk: Makes sense. I will add them.

wallet/wallet.go
ErikEk: The rationale was to reuse the same importing technique as ListTransactionDetails uses in lnd (https://github.com/lightningnetwork/lnd/blob/91d910872d9a5d1490026a3a9f92a7afaa9af069/lnwallet/btcwallet/btcwallet.go#L1450).

ErikEk: Sounds good. I'll provide an updated pr when I am done with the new tests.

wallet/wallet.go
```diff
@@ -2464,6 +2467,56 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+		txmgrNs := dbtx.ReadBucket(wtxmgrNamespaceKey)
+
+		txDetail, err := w.TxStore.TxDetails(txmgrNs, &txHash)
+		if err != nil {
+			return err
+		}
+
+		// If the transaction was not found we return an error.
+		if txDetail == nil {
+			return ErrNoTx
```
yyforyongyu, line 2496: nit: `return fmt.Errorf("%w: txid %v", ErrNoTx, txHash)`

wallet/wallet_test.go
```diff
@@ -205,3 +206,89 @@ func TestLabelTransaction(t *testing.T) {
...
+
+	tests := []struct {
+		name string
+
+		// Whether the transaction should be known to the wallet.
+		txKnown bool
+
+		// Whether the transaction is mined.
+		minedTx bool
```
yyforyongyu, line 221: this field can be removed since it's not used.

wallet/wallet_test.go
```diff
@@ -205,3 +206,89 @@ func TestLabelTransaction(t *testing.T) {
...
+						)
+
+						return w.TxStore.InsertTx(ns, rec, test.blockMeta)
+
+					})
+				require.NoError(t, err, "could not insert tx")
+			}
+
+			tx, err := w.GetTransaction(*TstTxHash)
+			if err != test.expectedErr {
```
yyforyongyu, line 282: can use `require.ErrorIs`

wallet/wallet_test.go
```diff
@@ -205,3 +206,89 @@ func TestLabelTransaction(t *testing.T) {
...
+				t.Fatalf("expected: %v, got: %v", test.expectedErr, err)
+			}
+			if test.txKnown {
+				require.True(t, tx.Summary.Hash.IsEqual(TstTxHash))
+			}
+			// Test if the transaction is mined.
+			if test.blockMeta != nil {
```
yyforyongyu, line 289: This is testing that when it's mined, the `Height` field is set, but it's not testing when it's not mined, this field is not set.

wallet/wallet_test.go
```diff
@@ -205,3 +206,89 @@ func TestLabelTransaction(t *testing.T) {
...
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			w, cleanup := testWallet(t)
+			defer cleanup()
+
+			// If the transaction should be known to the store, we
+			// write txdetail to disk.
+			if test.txKnown {
```
yyforyongyu, line 261: this approach works, tho imo a better version would be creating testing a `txMined` and `txUnmined`, then define the table tests as 
```go
{
  name string
  txid chainhash.Hash
  expectedTx *GetTransactionResult
  expectedErr error
}
```
Then you construct three txids - `txMined = ...`, `txUnmined= ...`, and `txUnknown = ...`. And finally check return returned values from `GetTransaction`,
```go
require.ErrorIs(t, tc.expectedErr, err)
require.Equal(t, tc.expectedTx, tx)
```

ErikEk: @yyforyongyu I have updated the test code. I am not sure if you wanted me to add all transactions to the wallet storage  at once or one for each test cycle like I have done now?

ErikEk: I had issues storing the full GetTransactionResults array and opted for comparing the txid and block height as validators.

wallet/wallet.go
```diff
@@ -2464,6 +2467,56 @@ func (w *Wallet) GetTransactions(startBlock, endBlock *BlockIdentifier,
...
+			res.Height = txDetail.Block.Height
+			res.BlockHash = &txDetail.Block.Hash
+		}
+
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+	return &res, err
```
guggero, line 2517: nit: Here it should explicitly be `return &res, nil`.

guggero (approved):
Nice, LGTM :tada:

wallet/wallet.go
ErikEk: Fixed.

wallet/wallet_test.go
```diff
@@ -205,3 +216,90 @@ func TestLabelTransaction(t *testing.T) {
 		})
 	}
 }
+
+// TestGetTransaction tests if we can fetch a mined, an existing
+// and a non-existing transaction from the wallet like we expect.
+func TestGetTransaction(t *testing.T) {
+	t.Parallel()
+	rec, err := wtxmgr.NewTxRecord(TstSerializedTx, time.Now())
+	require.NoError(t, err)
```
yyforyongyu, line 225: nit: add a newline here

wallet/wallet_test.go
```diff
@@ -205,3 +216,90 @@ func TestLabelTransaction(t *testing.T) {
...
+			},
+			expectedErr: ErrNoTx,
+		},
+	}
+	for _, test := range tests {
+		test := test
+
+		t.Run(test.name, func(t *testing.T) {
+			w, cleanup := testWallet(t)
+			defer cleanup()
```
yyforyongyu, line 283: nit: add a newline here

yyforyongyu (approved):
LGTM🙏

ErikEk:
@yyforyongyu PR updated.

PR #891, opened 2023-10-13
chain+wallet: fix race over `RescanProgress` message

yyforyongyu:
There's a race found in `lnd` side re multiple goroutines accessing the `RescanProgress`, [details in this PR.](https://github.com/lightningnetwork/lnd/pull/8094)

It also fixes the case where `bitcoind` is still loading blocks, causing the `NewBitcoindConn` to return an error.

### DIFF
--- a/chain/bitcoind_conn.go
+++ b/chain/bitcoind_conn.go
@@ -1,10 +1,13 @@
 package chain
 
 import (
+	"errors"
 	"fmt"
 	"net"
+	"strings"
 	"sync"
 	"sync/atomic"
+	"time"
 
 	"github.com/btcsuite/btcd/btcjson"
 	"github.com/btcsuite/btcd/chaincfg"
@@ -38,7 +41,20 @@ const (
 	// errBlockPrunedStr is the error message returned by bitcoind upon
 	// calling GetBlock on a pruned block.
 	errBlockPrunedStr = "Block not available (pruned data)"
+
+	// errStillLoadingCode is the error code returned when an RPC request
+	// is made but bitcoind is still in the process of loading or verifying
+	// blocks.
+	errStillLoadingCode = "-28"
+
+	// bitcoindStartTimeout is the time we wait for bitcoind to finish
+	// loading and verifying blocks and become ready to serve RPC requests.
+	bitcoindStartTimeout = 30 * time.Second
 )
+
+// ErrBitcoindStartTimeout is returned when the bitcoind daemon fails to load
+// and verify blocks under 30s during startup.
+var ErrBitcoindStartTimeout = errors.New("bitcoind start timeout")
 
 // BitcoindConfig contains all of the parameters required to establish a
 // connection to a bitcoind's RPC.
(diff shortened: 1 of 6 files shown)


chain/bitcoind_conn.go
```diff
@@ -306,9 +313,56 @@ func (c *BitcoindConn) sendTxToClients() {
...
+	// we exit with it.
+	if !strings.Contains(err.Error(), errCode) {
+		return nil, err
+	}
+
+	// Otherwise, we'd retry calling getblockhash or time out after 30s.
+	for {
+		select {
+		// Timeout after 30s.
+		case <-time.After(30 * time.Second):
```
guggero, line 343: Should we extract this into a constant at least or even make it configurable?

guggero (approved):
utACK, just one nit :tada:

chain/bitcoind_conn.go
```diff
@@ -306,9 +313,56 @@ func (c *BitcoindConn) sendTxToClients() {
...
+// getBlockHashDuringStartup is used to call the getblockhash RPC during
+// startup. It catches the case where bitcoind is still in the process of
+// loading blocks, which returns the following error,
+// - "-28: Loading block index..."
+// - "-28: Verifying blocks..."
+// In this case we'd retry every second until we time out after 30 seconds.
+func getBlockHashDuringStartup(
+	client *rpcclient.Client) (*chainhash.Hash, error) {
+
+	errCode := "-28"
```
Crypt-iQ, line 325: nit: extract to constant above?

Crypt-iQ (approved):

chain/bitcoind_conn.go
yyforyongyu: updated to make it a constant

chain/bitcoind_conn.go
yyforyongyu: updated to make it a constant

chain/bitcoind_conn.go
```diff
@@ -311,9 +327,55 @@ func (c *BitcoindConn) sendTxToClients() {
...
+	// we exit with it.
+	if !strings.Contains(err.Error(), errStillLoadingCode) {
+		return nil, err
+	}
+
+	// Otherwise, we'd retry calling getblockhash or time out after 30s.
+	for {
+		select {
+		// Timeout after 30s.
+		case <-time.After(bitcoindStartTimeout):
```
guggero, line 356: Sorry, I missed this before. Because we create a new channel with `time.After()` within the select, this starts ticking afresh for each loop we make. So the one second one will always and forever tick first and this one will never get to tick. We need to create the channel outside of the loop if we want to cancel re-trying after 30 seconds.

guggero (changes requested):
Missed a crucial thing in my first review. Once that's fixed the PR is ready to go though, very nice fixes!

chain/bitcoind_conn.go
yyforyongyu: oh shoot my bad, now fixed!

guggero (approved):
LGTM :tada:

PR #899, opened 2024-01-05
chain: add `TestMempoolAccept` to chain interface

yyforyongyu:
`TestMempoolAccept` is now added to chain interface so it can be accessed via chain io in `lnd`.


wallet/wallet.go
```diff
@@ -4131,7 +4131,7 @@ func MapBroadcastBackendError(err error) error {
 	//
 	// https://github.com/bitcoin/bitcoin/blob/9bf5768dd628b3a7c30dd42b5ed477a92c4d3540/src/node/transaction.cpp#L49
 	// https://github.com/bitcoin/bitcoin/blob/0.20/src/validation.cpp#L642
-	case match(err, "missing inputs") ||
+	case match(err, "missing-inputs") ||
```
Roasbeef, line 4134: 😣

Roasbeef: Wondering what we can do in the future to catch stuff like this...

Ofc it's the case that string error matching is fragile to being with, wish we had some concrete error codes here.

Roasbeef: Or should we be looking at one of these sources instead?
  * https://github.com/bitcoin/bitcoin/blob/4baa162dbb3c6464e196a4a21fe63794859021b4/src/util/error.cpp#L19
  * https://github.com/bitcoin/bitcoin/blob/4baa162dbb3c6464e196a4a21fe63794859021b4/src/consensus/tx_verify.cpp#L172

Roasbeef: Also fwiw, this gives a 404 now: 
> https://github.com/bitcoin/bitcoin/blob/0.20/src/validation.cpp#L642

yyforyongyu: These two are my primary source when comparing RPC errors, as I'm not sure how `bitcoind` converts them internally, so I can only rely on the end results from its rpc responses,
- https://github.com/bitcoin/bitcoin/blob/master/test/functional/data/invalid_txs.py
- https://github.com/bitcoin/bitcoin/blob/master/test/functional/mempool_accept.py

Also updated the link, and added a new method to properly handle error matching. Meanwhile plan to work on something in `btcd/btcjson` - this way we can have a single place to track the bitcoind errors and handles the matching then turning them into defined error types.

saubyk:
cc: @ellemouton

go.mod
```diff
@@ -1,11 +1,11 @@
 module github.com/btcsuite/btcwallet
 
 require (
-	github.com/btcsuite/btcd v0.23.5-0.20230711222809-7faa9b266231
+	github.com/btcsuite/btcd v0.23.5-0.20231215221805-96c9fd8078fd
```
ellemouton, line 4: think we can update now (unless you wanna wait for a tag on that side)

ellemouton (approved):
lgtm!

go.mod
yyforyongyu: cool updated!

Roasbeef (approved):
LGTM 🍭

PR #900, opened 2024-01-12
wallet: export coin selection strategy code for re-use

guggero:
We want to be able to use the available coin selection strategies outside of just the coin selection methods offered by the wallet. So this PR refactors the code to be re-used.

### DIFF
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -21,16 +21,8 @@ import (
 	"github.com/btcsuite/btcwallet/wtxmgr"
 )
 
-// byAmount defines the methods needed to satisify sort.Interface to
-// sort credits by their output amount.
-type byAmount []wtxmgr.Credit
-
-func (s byAmount) Len() int           { return len(s) }
-func (s byAmount) Less(i, j int) bool { return s[i].Amount < s[j].Amount }
-func (s byAmount) Swap(i, j int)      { s[i], s[j] = s[j], s[i] }
-
-func makeInputSource(eligible []wtxmgr.Credit) txauthor.InputSource {
-	// Current inputs and their total value.  These are closed over by the
+func makeInputSource(eligible []Coin) txauthor.InputSource {
+	// Current inputs and their total value. These are closed over by the
 	// returned input source and reused across multiple calls.
 	currentTotal := btcutil.Amount(0)
 	currentInputs := make([]*wire.TxIn, 0, len(eligible))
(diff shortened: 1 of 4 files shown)


wallet/createtx.go
```diff
@@ -41,15 +33,23 @@ func makeInputSource(eligible []wtxmgr.Credit) txauthor.InputSource {
 		[]btcutil.Amount, [][]byte, error) {
 
 		for currentTotal < target && len(eligible) != 0 {
-			nextCredit := &eligible[0]
```
Roasbeef, line 44: Not sure why before a pointer was obtained here, just to then get the addr again to call into `NewTxIn`...

wallet/createtx.go
```diff
@@ -446,3 +429,76 @@ func validateMsgTx(tx *wire.MsgTx, prevScripts [][]byte,
...
+
+func (s sortByAmount[T]) Len() int { return len(s) }
+func (s sortByAmount[T]) Less(i, j int) bool {
+	return s[i].Output().Value < s[j].Output().Value
+}
+func (s sortByAmount[T]) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
+
+// ArrangeCoins takes a list of coins and arranges them according to the
+// specified coin selection strategy and fee rate.
+func ArrangeCoins[T SelectableCoin](coinSelectionStrategy CoinSelectionStrategy,
```
Roasbeef, line 469: Type param usage doesn't feel absolutely necessary here? At least in the context of the wallet today, seems like we could jut swap in `SelectableCoin` as the interface are here instead of `T`. I think reaching for the type param here would make more sense if we already had several other instances of something like `ArrangeCoins` that accepted a diff concrete type. 

Will take a look at the other dependent PRs here as well. Not a blocker, just mainly curious.

Roasbeef (approved):
LGTM 🩰

wallet/createtx.go
```diff
@@ -446,3 +433,76 @@ func validateMsgTx(tx *wire.MsgTx, prevScripts [][]byte,
...
+	eligible []T, feeSatPerKb btcutil.Amount) ([]T, error) {
+
+	switch coinSelectionStrategy {
+	// Pick largest outputs first.
+	case CoinSelectionLargest:
```
Roasbeef, line 474: We could also go one step further here and make `CoinSelectionStrategy` itself an interface. This would allow other custom selection strategies in theory, such as spending the oldest coins first, or selecting random coins, etc.

wallet/createtx.go
guggero: Yeah. Looking closer at this function also makes me think something's pretty weird in the way it is implemented.
If we can call the input source multiple times but keep the slices and total around, wouldn't that mean that we could never select _fewer_ inputs in a subsequent call that we've selected in a previous call?
Not sure if the underlying coin selection algorithm actually operates under the same assumption...

wallet/createtx.go
guggero: The reason for using the type parameter here is because Golang doesn't allow using a struct slice when an interface slice is expected.
For example, if `ArrangeCoins` took `eligigle []SelectableCoin`, we couldn't pass in a slice of `[]*SelectableCredit`.
Not a big deal here since we're wrapping the `wtxmgr.Credit` anyway.
But in `lnd` we could pass in `chanfunding.Coin` directly if we use the typed parameter.
But wrapping that wouldn't be a big deal either, so I can go with the interface if you prefer.

wallet/createtx.go
guggero: Makes sense, will do.

wallet/createtx.go
guggero: I resolved this by just moving the `chanfunding.Coin` declaration into the wallet. I think it makes sense to have a specific type that contains just exactly what's needed for coin selection.

wallet/wallet.go
```diff
@@ -89,17 +89,33 @@ var (
 	wtxmgrNamespaceKey   = []byte("wtxmgr")
 )
 
-type CoinSelectionStrategy int
+// Coin represents a spendable UTXO which is available for coin selection.
+type Coin struct {
```
Roasbeef, line 93: 👍

Roasbeef (approved):
LGTM 🦊

PR #901, opened 2024-01-15
mod: bump Golang min version to v1.18 in all submodules

guggero:
Because the Go module behavior was changed in newer versions of Golang, running `go mod tidy` does produce a diff with those versions.
We'll want to have the tighter dependency logic and for that reason update all submodules to require at least `Go 1.18` and run `go mod tidy` on all the modules.

cc @yyforyongyu

After merging the PR we'll want to push the following **new** tags:
 - `wallet/txauthor/v1.3.4`
 - `wallet/txrules/v1.2.1`
 - `wallet/txsizes/v1.2.4`
 - `walletdb/v1.4.1`
 - `wtxmgr/v1.5.1`

After pushing those tags, we'll need a new PR that bumps the main `go.mod` to all those versions.


yyforyongyu (approved):
LGTM🙏

Roasbeef (approved):
LGTM 🏸

PR #902, opened 2024-01-22
refactor: add `make help` to display the usage for project Makefile

Halimao:
Similar to btcsuite/btcd#2107

This pr added a new `help` target for displaying which commonly used commands that the Makefile supported. This is useful when a new developer goes to find a helpful tool for testing or building etc.

image


guggero (approved):
Thanks, LGTM :tada:

PR #903, opened 2024-01-24
chain: PrunedBlockDispatcher bugfix.

ziggie1984:
This addresses  #https://github.com/lightningnetwork/lnd/issues/8250.

(Still need to update the unittests).

So this does not address the related time issue for now which is acutally the case why the requests where failing in the first place.

I would add the new timing logic and introduce the `getblockfrompeer` command for bitcoind >24 (iirc) in a followup PR.

### DIFF
--- a/chain/bitcoind_conn.go
+++ b/chain/bitcoind_conn.go
@@ -469,10 +469,16 @@ func (c *BitcoindConn) GetBlock(hash *chainhash.Hash) (*wire.MsgBlock, error) {
 		return nil, err
 	}
 
+	// cancelChan is needed in case a block request with the same hash is
+	// already registered and fails via the returned errChan. Because we
+	// don't register a new request in this case this cancelChan is used
+	// to signal the failure of the dependant block request.
+	cancelChan := make(chan error, 1)
+
 	// Now that we know the block has been pruned for sure, request it from
 	// our backend peers.
 	blockChan, errChan := c.prunedBlockDispatcher.Query(
-		[]*chainhash.Hash{hash},
+		[]*chainhash.Hash{hash}, cancelChan,
 	)
 
 	for {
@@ -482,11 +488,31 @@ func (c *BitcoindConn) GetBlock(hash *chainhash.Hash) (*wire.MsgBlock, error) {
 
 		case err := <-errChan:
 			if err != nil {
+				// An error was returned for this block request.
+				// We have to make sure we remove the blockhash
+				// from the list of queried blocks.
+				// Moreover because in case a block is requested
+				// more than once no redundant block requests
+				// are registered but rather a reply channel is
+				// added to the pending block request. This
+				// means we need to cancel all dependent
+				// `GetBlock` calls via the cancel channel when
+				// the fetching of the block was NOT successful.
+				c.prunedBlockDispatcher.CancelRequest(
+					*hash, err,
+				)
+
 				return nil, err
 			}
 
 			// errChan fired before blockChan with a nil error, wait
 			// for the block now.
+
+		// The cancelChan is only used when there is already a pending
+		// block request for this hash and that block request fails via
+		// the error channel above.
+		case err := <-cancelChan:
+			return nil, err
 
 		case <-c.quit:
 			return nil, ErrBitcoindClientShuttingDown
--- a/chain/pruned_block_dispatcher.go
+++ b/chain/pruned_block_dispatcher.go
@@ -143,7 +143,14 @@ type PrunedBlockDispatcher struct {
 	// NOTE: The blockMtx lock must always be held when accessing this
 	// field.
 	blocksQueried map[chainhash.Hash][]chan *wire.MsgBlock
-	blockMtx      sync.Mutex
+
+	// blockQueryCancel signals the cancellation of a `GetBlock` request.
+	//
+	// NOTE: The blockMtx lock must always be held when accessing this
+	// field.
+	blockQueryCancel map[chainhash.Hash][]chan<- error
+
+	blockMtx sync.Mutex
 
 	// currentPeers represents the set of peers we're currently connected
 	// to. Each peer found here will have a worker spawned within the
@@ -198,12 +205,13 @@ func NewPrunedBlockDispatcher(cfg *PrunedBlockDispatcherConfig) (
 			NewWorker: query.NewWorker,
 			Ranking:   query.NewPeerRanking(),
 		}),
-		blocksQueried:  make(map[chainhash.Hash][]chan *wire.MsgBlock),
-		currentPeers:   make(map[string]*peer.Peer),
-		bannedPeers:    make(map[string]struct{}),
-		peersConnected: peersConnected,
-		timeSource:     blockchain.NewMedianTime(),
-		quit:           make(chan struct{}),
+		blocksQueried:    make(map[chainhash.Hash][]chan *wire.MsgBlock),
+		blockQueryCancel: make(map[chainhash.Hash][]chan<- error),
+		currentPeers:     make(map[string]*peer.Peer),
+		bannedPeers:      make(map[string]struct{}),
+		peersConnected:   peersConnected,
+		timeSource:       blockchain.NewMedianTime(),
+		quit:             make(chan struct{}),
 	}, nil
 }
 
@@ -502,9 +510,10 @@ func (d *PrunedBlockDispatcher) banPeer(peer string) {
 
 // Query submits a request to query the information of the given blocks.
 func (d *PrunedBlockDispatcher) Query(blocks []*chainhash.Hash,
+	cancelChan chan<- error,
 	opts ...query.QueryOption) (<-chan *wire.MsgBlock, <-chan error) {
 
-	reqs, blockChan, err := d.newRequest(blocks)
+	reqs, blockChan, err := d.newRequest(blocks, cancelChan)
 	if err != nil {
 		errChan := make(chan error, 1)
 		errChan <- err
@@ -515,14 +524,18 @@ func (d *PrunedBlockDispatcher) Query(blocks []*chainhash.Hash,
 	if len(reqs) > 0 {
 		errChan = d.workManager.Query(reqs, opts...)
 	}
+
 	return blockChan, errChan
 }
 
 // newRequest construct a new query request for the given blocks to submit to
 // the internal workManager. A channel is also returned through which the
 // requested blocks are sent through.
-func (d *PrunedBlockDispatcher) newRequest(blocks []*chainhash.Hash) (
-	[]*query.Request, <-chan *wire.MsgBlock, error) {
+//
+// NOTE: The cancelChan must be buffered.
+func (d *PrunedBlockDispatcher) newRequest(blocks []*chainhash.Hash,
+	cancelChan chan<- error) ([]*query.Request, <-chan *wire.MsgBlock,
+	error) {
 
 	// Make sure the channel is buffered enough to handle all blocks.
 	blockChan := make(chan *wire.MsgBlock, len(blocks))
@@ -550,6 +563,10 @@ func (d *PrunedBlockDispatcher) newRequest(blocks []*chainhash.Hash) (
 		} else {
 			log.Debugf("Received new request for pending query of "+
 				"block %v", *block)
+
+			d.blockQueryCancel[*block] = append(
+				d.blockQueryCancel[*block], cancelChan,
+			)
 		}
 
 		d.blocksQueried[*block] = append(
@@ -614,6 +631,8 @@ func (d *PrunedBlockDispatcher) handleResp(req, resp wire.Message,
 			Finished:   false,
 		}
 	}
+	copyblockChans := make([]chan *wire.MsgBlock, len(blockChans))
+	copy(copyblockChans, blockChans)
 
 	err := blockchain.CheckBlockSanity(
 		btcutil.NewBlock(block), d.cfg.ChainParams.PowLimit,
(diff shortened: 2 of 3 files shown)


chain/pruned_block_dispatcher.go
```diff
@@ -501,10 +508,10 @@ func (d *PrunedBlockDispatcher) banPeer(peer string) {
 }
 
 // Query submits a request to query the information of the given blocks.
-func (d *PrunedBlockDispatcher) Query(blocks []*chainhash.Hash,
+func (d *PrunedBlockDispatcher) Query(blocks []*chainhash.Hash, chanelChan chan<- error,
```
Roasbeef, line 511: `chanelChan` -> `cancelChan`?

chain/bitcoind_conn.go
```diff
@@ -469,10 +469,16 @@ func (c *BitcoindConn) GetBlock(hash *chainhash.Hash) (*wire.MsgBlock, error) {
 		return nil, err
 	}
 
+	// cancelChan is needed in case a block request with the same hash is
+	// already registered and fails via the returned errChan. Because we
+	// don't register a new request in this case this cancelChan is used
+	// to signal the failure of the dependant block request.
+	cancelChan := make(chan error, 1)
```
Roasbeef, line 476: 👍 

So just for my understanding, the issue before was that:
   * `N` goroutines try to fetch same block hash `H`
   * goroutine `N_0` gets there first
   * goroutines `N_1` to `N_i` block and wait for the response 
   * the query for goroutine `N_0` times out, that one exits
   * the others just sit there, and basically end up blocking forver
   * `synced_to_chain: false`

chain/pruned_block_dispatcher.go
```diff
@@ -678,3 +692,42 @@ func (d *PrunedBlockDispatcher) handleResp(req, resp wire.Message,
...
+	d.blockMtx.Unlock()
+
+	// In case there are dependant goroutines depending on this block
+	// request we make sure we cancel them.
+	if ok {
+		d.wg.Add(1)
+		go func() {
+			defer d.wg.Done()
+
+			for _, cancel := range cancelChans {
```
Roasbeef, line 724: I think this'll block if `cancelChans` is nil.

Roasbeef: Ah nvm, we check `ok` above. Alternatively, we could have the function short circuit and eliminate some of the nesting here.

chain/bitcoind_conn.go
ziggie1984: Exactly this could happen but was acutally not observed by me in the wild but it defintely could happen if two requests for the same block come in shortly after another.

What I observed was that because we did not delete:

```
delete(d.blocksQueried, blockHash)
```
when we failed the request due to a timeout, we would never register a new request although there was no request going on in the background so we would wait forever for this new `getblock` calls.

chain/bitcoind_conn.go
```diff
@@ -482,12 +488,32 @@ func (c *BitcoindConn) GetBlock(hash *chainhash.Hash) (*wire.MsgBlock, error) {
 
 		case err := <-errChan:
 			if err != nil {
+				// An error was returned for this block request.
+				// We have to make sure we remove the blockhash
+				// from the list of queried blocks.
+				// Moreover because in case a block is requested
+				// more than onced no redundant block requests
```
C-Otto, line 495:
```suggestion
				// more than once no redundant block requests
```

chain/bitcoind_conn.go
```diff
@@ -482,12 +488,32 @@ func (c *BitcoindConn) GetBlock(hash *chainhash.Hash) (*wire.MsgBlock, error) {
...
+					*hash, err,
+				)
+
 				return nil, err
 			}
 
 			// errChan fired before blockChan with a nil error, wait
 			// for the block now.
 
+		// The  cancelChan is only used when there is already a pending
```
C-Otto, line 511:
```suggestion
		// The cancelChan is only used when there is already a pending
```

chain/pruned_block_dispatcher.go
```diff
@@ -614,6 +631,8 @@ func (d *PrunedBlockDispatcher) handleResp(req, resp wire.Message,
 			Finished:   false,
 		}
 	}
+	copyblockChans := make([]chan *wire.MsgBlock, len(blockChans))
```
ziggie1984, line 634: I think we must deepcopy here, because its a slice, otherwise we might run into a race below ?

chain/pruned_block_dispatcher.go
```diff
@@ -678,3 +697,58 @@ func (d *PrunedBlockDispatcher) handleResp(req, resp wire.Message,
...
+
+	d.blockMtx.Lock()
+
+	// Before removing the block hash we get the cancelChans which were
+	// registered for block requests that had already an ongoing pending
+	// request.
+	cancelChans, ok := d.blockQueryCancel[blockHash]
+	var copycancelChans []chan<- error
+	if ok {
+		copycancelChans = make([]chan<- error, len(cancelChans))
```
ziggie1984, line 730: Same here as above, I think deep copy is necessary.

chain/pruned_block_dispatcher.go
yyforyongyu: yeah I think so since `blockChans` is just a pointer to the slice keyed using `blockHash` in map `blocksQueried`.

chain/pruned_block_dispatcher.go
```diff
@@ -550,6 +563,10 @@ func (d *PrunedBlockDispatcher) newRequest(blocks []*chainhash.Hash) (
 		} else {
 			log.Debugf("Received new request for pending query of "+
 				"block %v", *block)
+
+			d.blockQueryCancel[*block] = append(
```
yyforyongyu, line 567: I think this assumes the first `GetBlock` request always gets processed first, but I don't think it's true. Take @Roasbeef 's example further,
- goroutine `N_0` gets there first, followed by `N_1` closely
- `workManager` finishes querying for `N_1` first, returning an error
- `N_0` will now hang there because it was never registered a `cancelChan`, as it was the first seen request by the dispatcher.

I think we need to have `blockQueryCancel` for every block, either new or not.

yyforyongyu: Down to the rabbit hole - does one failed `GetBlock` request means another one would also fail for the same block? I guess it depends on the implementation details of `workManager.Query`, as we may choose to send this block inv msg to different peers. I think a better way to handle this is, cache a previous success `GetBlock` result, similar to the `blockcache` in `lnd`, future PRs tho.

yyforyongyu (commented):
Thanks for the PR! Left a comment about how we add the `cancelChan`.

For future concerns, although `GetBlock` may ask for the same block, we should treat each request independently. We can use something like a `requestID` to handle sending back the corresponding response, to make sure one request doesn't interfere with others. On top of that we can think about optimizations such as caching a previous success block response for future requests. Then ofc there's the new RPC `getblockfrompeer` we can use for pruned nodes.

C-Otto:
Note that caching is currently done in lnd: `blockcache/blockcache.go`

chain/pruned_block_dispatcher.go
ziggie1984: I think race condition you are describing will not happen because the [newRequest](https://github.com/btcsuite/btcwallet/blob/master/chain/pruned_block_dispatcher.go#L524) method guards with the lock against successive goroutines being registered. The first block request which aquires the lock will send out the request to the internal working manager so there will be no race condition.

> I think we need to have blockQueryCancel for every block, either new or not.

We could add it but its not necessary because the initial (first) block request will use the error channel if the request fails.

ziggie1984: maybe I am not seeing something tho ?

yyforyongyu: > The first block request which aquires the lock will send out the request to the internal working manager so there will be no race condition

Yeah but the lock only guarantees the registration of the block in the map. Once `d.newRequest` finishes, there's no guarantee. There's also no guarantee that ` d.workManager.Query(reqs, opts...)` will return in order, since `workManager.Query` would just put the job in a batch in an async manner.

> 

> And because we never request more than one block in what request we don't have to care about the case where we have mutliple blocks in the INV msg.

I mean even for the querying the same block multiple times we should treat each attempt differently. The mental model is different here - we think it's the same request if the block hash is the same, while in reality it's a different request, and that needs to be reflected in the design.

chain/pruned_block_dispatcher.go
```diff
@@ -395,7 +396,20 @@ func (d *PrunedBlockDispatcher) connectToPeer(addr string) (bool, error) {
...
+	// peers to fetch blocks from.
+	sort.Slice(peers, func(i, j int) bool {
+		return peers[i].PingTime < peers[j].PingTime
+	})
+
 	for _, peer := range peers {
+		// We cannot use the inbound peers here because the referenced
+		// port in the `addr` field is not the listen port for the p2p
+		// connection but a random outgoing port of the peer.
+		if peer.Inbound {
```
yyforyongyu, line 410: o damn this is here the whole time...could we maybe add a debug log here?

chain/pruned_block_dispatcher.go
yyforyongyu: ok nvm didn't realize the `reqs` are only created for the first block request, so this won't be an issue.

chain/pruned_block_dispatcher_test.go
```diff
@@ -449,16 +502,64 @@ func TestPrunedBlockDispatcherQuerySameBlock(t *testing.T) {
...
+	blockChans := make([]<-chan *wire.MsgBlock, 0, numRequests)
+	errChans := make([]<-chan error, 0, numRequests)
+	cancelChans := make([]<-chan error, 0, numRequests)
+
+	// We want to force a timeout.
+	h.disablePeerReplies()
+
 	for i := 0; i < numRequests; i++ {
-		blockChan, errChan := h.query(h.hashes)
+		// The default retry number is 2 and is defined in the neutrino
```
yyforyongyu, line 545: 👍

chain/pruned_block_dispatcher_test.go
```diff
@@ -406,6 +421,38 @@ func (h *prunedBlockDispatcherHarness) assertPeerReplied(
...
+	h.t.Helper()
+
+	select {
+	case <-blockChan:
+		h.t.Fatalf("expected no reply from peer")
+
+	case err := <-errChan:
+		require.ErrorIs(h.t, err, expectedErr)
+		for _, hash := range h.hashes {
+			h.dispatcher.CancelRequest(*hash, err)
```
yyforyongyu, line 439: Feels weird that we are re-implementing the logic in `GetBlock`, but I see there's no unit tests for `GetBlock` so nvm, we need to do more work here to patch tests in future PRs tho.

yyforyongyu (approved):
LGTM 🌹

chain/pruned_block_dispatcher_test.go
ziggie1984: Agree fully, this felt not right to me as well. Will write an issue to move the block cache into `btcwallet` and enhance the test coverage especially `GetBlock`

ziggie1984: https://github.com/btcsuite/btcwallet/issues/904

chain/pruned_block_dispatcher.go
```diff
@@ -678,3 +697,52 @@ func (d *PrunedBlockDispatcher) handleResp(req, resp wire.Message,
...
+
+	d.blockMtx.Lock()
+
+	// Before removing the block hash we get the cancelChans which were
+	// registered for block requests that had already an ongoing pending
+	// request.
+	cancelChans, ok := d.blockQueryCancel[blockHash]
+	var copycancelChans []chan<- error
+	if ok {
+		copycancelChans = make([]chan<- error, len(cancelChans))
```
Roasbeef, line 749: Why is this copy necessary? This variable isn't used at all either.

chain/pruned_block_dispatcher.go
```diff
@@ -678,3 +697,52 @@ func (d *PrunedBlockDispatcher) handleResp(req, resp wire.Message,
...
+	delete(d.blockQueryCancel, blockHash)
+
+	d.blockMtx.Unlock()
+
+	// In case there are goroutines depending on this block request we
+	// make sure we cancel them.
+	// We do this in a goroutine to not block the initial request.
+	if copycancelChans != nil {
+		d.wg.Add(1)
+		go failDependant(cancelChans)
```
Roasbeef, line 765: I think this should be `copycancelChans`?

chain/pruned_block_dispatcher.go
```diff
@@ -678,3 +697,52 @@ func (d *PrunedBlockDispatcher) handleResp(req, resp wire.Message,
...
+	// block can be registered again in the future.
+	delete(d.blocksQueried, blockHash)
+	delete(d.blockQueryCancel, blockHash)
+
+	d.blockMtx.Unlock()
+
+	// In case there are goroutines depending on this block request we
+	// make sure we cancel them.
+	// We do this in a goroutine to not block the initial request.
+	if copycancelChans != nil {
```
Roasbeef, line 763: Re above, this can just be `ok`.

chain/pruned_block_dispatcher.go
ziggie1984: Nice catch, its used below now below `failDependant`

Roasbeef (approved):
LGTM 💣

PR #905, opened 2024-01-25
wallet: downgrade `spew` log to debug and limit its size

yyforyongyu:
This commit removes using `spew` on production as the tx log can be extremely large. For instance, when running [this build](https://github.com/lightningnetwork/lnd/actions/runs/7656820228/job/20866038765) the CI crashed because of this [test](https://github.com/lightningnetwork/lnd/pull/8345/commits/11bafe84ff624e0d7a2d56ef00be89567a7d4cbd), that also caused my local terminal to crash, because we are printing a very large invalid tx.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -3819,15 +3819,23 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
 		log.Warnf("Unable to remove invalid transaction %v: %v",
 			tx.TxHash(), dbErr)
 	} else {
-		// The spew output is pretty nice to directly see how many
-		// inputs/outputs of what type there are.
-		log.Infof("Removed invalid transaction: %v", spew.Sdump(tx))
+		log.Infof("Removed invalid transaction: %v", tx.TxHash())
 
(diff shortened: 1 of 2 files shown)


wallet/wallet.go
```diff
@@ -3819,15 +3819,21 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
+		// The serialized transaction is for logging only, don't fail
+		// on the error.
 		var txRaw bytes.Buffer
 		_ = tx.Serialize(&txRaw)
-		log.Infof("Removed invalid transaction: %x", txRaw.Bytes())
+
+		// Optionally log the tx in debug when the size is manageable.
+		if txRaw.Len() < 1_000_000 {
+			log.Debugf("Removed invalid transaction: %v \n hex=%x",
+				spew.Sdump(tx), txRaw.Bytes())
```
Roasbeef, line 3832: We can also use `newLogClosure` here. This way, spew is only called if the debug level would have us print the line. As is, it'll always be called to eval the args of the function.

yyforyongyu: Done!

Roasbeef (approved):
LGTM 🧊

PR #907, opened 2024-01-27
Add option to allow for dust when checking outputs

benthecarman:
This seems standard in most wallet libraries nowadays to give users the option to allow for dust if they know what they are doing.

Particularly for my usecase I am trying to make OP_RETURNs larger than 80 bytes and btcwallet will mark it as a non-standard dust tx, so this would allow me to get around it.

### DIFF
--- a/wallet/txrules/rules.go
+++ b/wallet/txrules/rules.go
@@ -39,14 +39,14 @@ var (
 
 // CheckOutput performs simple consensus and policy tests on a transaction
 // output.
-func CheckOutput(output *wire.TxOut, relayFeePerKb btcutil.Amount) error {
+func CheckOutput(output *wire.TxOut, relayFeePerKb btcutil.Amount, allowDust bool) error {
 	if output.Value < 0 {
 		return ErrAmountNegative
 	}
 	if output.Value > btcutil.MaxSatoshi {
 		return ErrAmountExceedsMax
 	}
-	if IsDustOutput(output, relayFeePerKb) {
+	if !allowDust && IsDustOutput(output, relayFeePerKb) {
 		return ErrOutputIsDust
 	}
 	return nil


guggero:
You mentioned you need this for `OP_RETURN` outputs. But those should already be exempt by [the check here](https://github.com/btcsuite/btcwallet/blob/master/wallet/txrules/rules.go#L26).
So I don't really see why this is needed? Or how exactly are you creating the `OP_RETURN` outputs?

benthecarman:
If the op return output is over the standard 80 byte limit it is considered dust.

guggero:
Ah yes. So you'd need a non-standard implementation of a chain backend to propagate such an output in the first place...

benthecarman:
Yeah I use Libre relay

wallet/txrules/rules.go
```diff
@@ -39,14 +39,14 @@ var (
 
 // CheckOutput performs simple consensus and policy tests on a transaction
 // output.
-func CheckOutput(output *wire.TxOut, relayFeePerKb btcutil.Amount) error {
+func CheckOutput(output *wire.TxOut, relayFeePerKb btcutil.Amount, allowDust bool) error {
```
guggero, line 42: Since this is a breaking API change, should we just create a new function `CheckOutputAllowDust` that has the boolean flag? Then this function can call into the new one with `false`.

benthecarman: Damn never saw this comment, updated!

benthecarman:
this might actually be a better fix: https://github.com/btcsuite/btcd/pull/2380

wallet/txrules/rules.go
```diff
@@ -40,15 +40,27 @@ var (
...
+		return err
+	}
+
+	if IsDustOutput(output, relayFeePerKb) {
+		return ErrOutputIsDust
+	}
+	return nil
+}
+
+// CheckOutput performs simple consensus and policy tests on a transaction
```
kcalvinalvin, line 55: The comment here should be updated to match the function.

kcalvinalvin:
> this might actually be a better fix: [btcsuite/btcd#2380](https://github.com/btcsuite/btcd/pull/2380)

https://github.com/btcsuite/btcd/pull/2380#discussion_r2103209095

I think if the main goal is to fix btcwallet considering op_returns as dust, I think changing the behavior in btcwallet is the way to go because of the aforementioned comment here.

PR #909, opened 2024-02-06
wallet: return ErrNotMine for script addresses

guggero:
We recently added the ManagedTaprootScriptAddress type which is not a ManagedPubKeyAddress and therefore doesn't store the derivation info. When trying to fetch the input info for such an address we run into the case where we return nil as the error which causes issues further up in the call stack.
This commit fixes the problem by returning an actual error and not the err variable that is nil due to the previous check.

### DIFF
--- a/wallet/utxos.go
+++ b/wallet/utxos.go
@@ -135,7 +135,7 @@ func (w *Wallet) FetchInputInfo(prevOut *wire.OutPoint) (*wire.MsgTx,
 	}
 	pubKeyAddr, ok := addr.(waddrmgr.ManagedPubKeyAddress)
 	if !ok {
-		return nil, nil, nil, 0, err
+		return nil, nil, nil, 0, ErrNotMine
 	}
 	keyScope, derivationPath, _ := pubKeyAddr.DerivationInfo()
 


wallet/utxos.go
```diff
@@ -135,7 +135,7 @@ func (w *Wallet) FetchInputInfo(prevOut *wire.OutPoint) (*wire.MsgTx,
 	}
 	pubKeyAddr, ok := addr.(waddrmgr.ManagedPubKeyAddress)
 	if !ok {
-		return nil, nil, nil, 0, err
+		return nil, nil, nil, 0, ErrNotMine
```
Roasbeef, line 138: So do we need to modify this routine to support the other types? Or is this just a way to catch this case in some higher level code?

Roasbeef: Ah also weird interaction re the `err` variable shadowing 🙈

guggero: I don't think we need to handle the other types, really. This is mainly about returning the _wrong_ error (it's always `nil` which is what breaks things at the call sites).

Roasbeef (approved):
LGTM

PR #910, opened 2024-02-23
chain: add mempool lookup for outpoints

yyforyongyu:
This PR adds a new method `LookupInputMempoolSpend` to directly check whether a given input has been spent or not in the mempool.

### DIFF
--- a/chain/bitcoind_events_test.go
+++ b/chain/bitcoind_events_test.go
@@ -67,6 +67,9 @@ func TestBitcoindEvents(t *testing.T) {
 			// mempool.
 			btcClient = setupBitcoind(t, addr, test.rpcPolling)
 			testNotifySpentMempool(t, miner1, btcClient)
+
+			// Test looking up mempool for input spent.
+			testLookupInputMempoolSpend(t, miner1, btcClient)
 		})
 	}
 }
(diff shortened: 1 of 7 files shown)


chain/bitcoind_events_test.go
```diff
@@ -214,6 +217,46 @@ func testNotifySpentMempool(t *testing.T, miner *rpctest.Harness,
 	}
 }
 
+// testLookupInputMempoolSpend tests that LookupInputMempoolSpend returns the
+// correct tx hash and whether the input has been spent in the mempool.
+func testLookupInputMempoolSpend(t *testing.T, miner *rpctest.Harness,
+	client *BitcoindClient) {
+
+	require := require.New(t)
```
Roasbeef, line 225: Maybe call it like `rt` to avoid shadowing the import path?

chain/bitcoind_events_test.go
```diff
@@ -214,6 +217,46 @@ func testNotifySpentMempool(t *testing.T, miner *rpctest.Harness,
 	}
 }
 
+// testLookupInputMempoolSpend tests that LookupInputMempoolSpend returns the
+// correct tx hash and whether the input has been spent in the mempool.
+func testLookupInputMempoolSpend(t *testing.T, miner *rpctest.Harness,
```
Roasbeef, line 222: Can we refactor these tests to also exercise the `btcd` version? Or is this already covered by the `btcd` integration tests themselves?

chain/bitcoind_events_test.go
yyforyongyu: Done.

chain/bitcoind_events_test.go
yyforyongyu: Yeah that's covered there, but I definitely agree we should add a new test `btcd_test.go` here to check the RPC client behavior, maybe something to do when we decide to refactor the `chain` package.

saubyk:
cc: @guggero @Crypt-iQ for review please

chain/bitcoind_zmq_events.go
```diff
@@ -501,28 +484,7 @@ func (b *bitcoindZMQEvents) mempoolPoller() {
...
-	resp, err := client.RawRequest(
-		"gettxspendingprevout", []json.RawMessage{req},
-	)
-	if err != nil {
-		return chainhash.Hash{}, false
-	}
-
-	var prevoutResps []getTxSpendingPrevOutResp
-	err = json.Unmarshal(resp, &prevoutResps)
+	prevoutResps, err := client.GetTxSpendingPrevOut([]wire.OutPoint{op})
```
Chinwendu20, line 487: Nice.

chain/btcd.go
```diff
@@ -91,6 +93,109 @@ func NewRPCClient(chainParams *chaincfg.Params, connect, user, pass string, cert
...
+// NewRPCClientWithConfig creates a client connection to the server based on
+// the config options supplised.
+//
+// The connection is not established immediately, but must be done using the
+// Start method.  If the remote server does not operate on the same bitcoin
+// network as described by the passed chain parameters, the connection will be
+// disconnected.
+func NewRPCClientWithConfig(cfg *RPCClientConfig) (*RPCClient, error) {
+	// Make sure the config is supplied.
+	if cfg == nil {
```
guggero, line 151: nit: the way `Go` works, this check could be in the `validate()` method.

guggero (approved):
utACK, LGTM :tada:

Roasbeef:
Can be rebased with the `btcd` version merged!

chain/btcd.go
yyforyongyu: Won't that cause a nil deference? Or you mean before initializing the client we do a validation on config in `lnd`?

guggero: No, the thing is, in Go you can call a method on a struct, even if the struct itself is nil.
So you can do:
```
// validate checks the required config options are set.
func (r *RPCClientConfig) validate() error {
	// Make sure the config is supplied.
	if r == nil {
		return errors.New("missing rpc config")
	}

	// Make sure retry attempts is positive.
	if r.ReconnectAttempts < 0 {
		return errors.New("reconnectAttempts must be positive")
	}
	...
```

yyforyongyu: Ohhh cool, TILed, thanks!

Roasbeef: Yeh at time a check on the nil pointer can actually be useful.

Roasbeef (approved):
LGTM 🎢

PR #911, opened 2024-02-29
WIP: testing docker linter

ffranr:
I couldn't reproduce the CI lint results (no lint errors) locally.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,9 +3,10 @@ PKG := github.com/btcsuite/btcwallet
 LINT_PKG := github.com/golangci/golangci-lint/cmd/golangci-lint
 GOACC_PKG := github.com/ory/go-acc
 GOIMPORTS_PKG := golang.org/x/tools/cmd/goimports
+TOOLS_DIR := tools
 
+GOPATH := $(shell go env GOPATH)
 GO_BIN := ${GOPATH}/bin
-LINT_BIN := $(GO_BIN)/golangci-lint
 GOACC_BIN := $(GO_BIN)/go-acc
 
 LINT_COMMIT := v1.46.0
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -3,9 +3,10 @@ PKG := github.com/btcsuite/btcwallet
 LINT_PKG := github.com/golangci/golangci-lint/cmd/golangci-lint
 GOACC_PKG := github.com/ory/go-acc
 GOIMPORTS_PKG := golang.org/x/tools/cmd/goimports
+TOOLS_DIR := tools
 
+GOPATH := $(shell go env GOPATH)
```
guggero, line 8: nit: `LINT_COMMIT` below is now unused.
It would also be nice if we could remove the `GOACC_COMMIT` and `GOIMPORTS_COMMIT` values as well by creating a `go.mod` file in the `tools` directory that pins these versions, then just install them from the `tools` directory as we do here: https://github.com/lightninglabs/taproot-assets/blob/main/Makefile#L79

tools/Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+ENV GOCACHE=/tmp/build/.cache
+ENV GOMODCACHE=/tmp/build/.modcache
+
+COPY . /tmp/tools
+
+RUN cd /tmp \
+  && mkdir -p /tmp/build/.cache \
+  && mkdir -p /tmp/build/.modcache \
+  && cd /tmp/tools \
+  && go install -trimpath github.com/golangci/golangci-lint/cmd/golangci-lint@v1.46.0 \
```
guggero, line 13: Same comment here, re pinning the version in a `go.mod` file instead.

wallet/mock.go
```diff
@@ -42,6 +42,10 @@ func (m *mockChainClient) GetBlockHeader(*chainhash.Hash) (*wire.BlockHeader,
 	return nil, nil
 }
 
+func (m *mockChainClient) GetBlockHeight(*chainhash.Hash) (int32, error) {
```
guggero, line 45: The same method needs to be added to the `RPCClient`.

guggero (commented):
Nice! Compilation seems to fail, but this will be very useful, thanks for adding!

Roasbeef:
Closing in favor of https://github.com/btcsuite/btcwallet/pull/1055

PR #913, opened 2024-03-07
Wrapp all errors

guggero:
In this PR we wrap all errors so we can properly use `errors.Is()` to find the cause for any error. This is especially useful for database related problems, so the underlying database error can be unwrapped.

### DIFF
--- a/waddrmgr/error.go
+++ b/waddrmgr/error.go
@@ -184,7 +184,7 @@ func (e ErrorCode) String() string {
 // manager operation.  It is used to indicate several types of failures
 // including errors with caller requests such as invalid accounts or requesting
 // private keys against a locked address manager, errors with the database
-// (ErrDatabase), errors with key chain derivation (ErrKeyChain), and errors
+// (ErrDatabase), errors with keychain derivation (ErrKeyChain), and errors
 // related to crypto (ErrCrypto).
 //
 // The caller can use type assertions to determine if an error is a ManagerError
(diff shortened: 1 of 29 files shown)


waddrmgr/error.go
```diff
@@ -207,6 +207,11 @@ func (e ManagerError) Error() string {
 	return e.Description
 }
 
+// Unwrap returns the underlying error, if any.
+func (e ManagerError) Unwrap() error {
```
Roasbeef, line 211: 👍

wtxmgr/tx.go
```diff
@@ -834,7 +834,7 @@ func (s *Store) UnspentOutputs(ns walletdb.ReadBucket) ([]Credit, error) {
 		rec, err := fetchTxRecord(ns, &op.Hash, &block)
 		if err != nil {
 			return fmt.Errorf("unable to retrieve transaction %v: "+
-				"%v", op.Hash, err)
+				"%", op.Hash, err)
```
Roasbeef, line 837: Should be `%w`. 

We also need to tag this module after merge to then update the main module.

Roasbeef (changes requested):

wtxmgr/tx.go
guggero: Oops, nice catch. Fixed.

guggero:
Did some more regex foo and found a few more instances. Now there should really be none left (that aren't in test code).

Roasbeef (approved):
LGTM 🧩

PR #918, opened 2024-03-14
chain+wallet: make use of the new errors from `rpcclient`

yyforyongyu:
This PR removes the string matching and makes use of the newly defined errors inside `rpcclient`.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -11,7 +11,6 @@ import (
 	"errors"
 	"fmt"
 	"sort"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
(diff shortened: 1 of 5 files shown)


wallet/wallet.go
```diff
@@ -3760,27 +3741,23 @@ func (w *Wallet) publishTransaction(tx *wire.MsgTx) (*chainhash.Hash, error) {
...
 	_, err = chainClient.SendRawTransaction(tx, false)
 	if err == nil {
 		return &txid, nil
 	}
 
-	returnErr = MapBroadcastBackendError(err)
-
-	var errInMempool *ErrInMempool
-	var errAlreadyConfirmed *ErrAlreadyConfirmed
+	// Map the error to an RPC-specific error type.
```
ziggie1984, line 3750: Nit: Maybe a more detailed comment that all errors are mapped to bitcoind errors no matter the used backend

ziggie1984 (approved):
LGTM thank you for the cleanup makes definitely sense to put the error mapping into the btcd package 🙏

guggero (approved):
Very nice! LGTM pending existing comments (and dependent PR).

wallet/wallet.go
yyforyongyu: updated!

guggero:
Can be updated with final version of `btcd`.

PR #919, opened 2024-03-24
Add Filter Utxo Filter Function

ziggie1984:
Needed for the new design to fix https://github.com/lightningnetwork/lnd/pull/8545

### DIFF
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -140,7 +140,9 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 	coinSelectKeyScope, changeKeyScope *waddrmgr.KeyScope,
 	account uint32, minconf int32, feeSatPerKb btcutil.Amount,
 	strategy CoinSelectionStrategy, dryRun bool,
-	selectedUtxos []wire.OutPoint) (*txauthor.AuthoredTx, error) {
+	selectedUtxos []wire.OutPoint,
+	allowUtxo func(utxo wtxmgr.Credit) bool) (
+	*txauthor.AuthoredTx, error) {
 
 	chainClient, err := w.requireChainClient()
 	if err != nil {
@@ -168,7 +170,8 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 		}
 
 		eligible, err := w.findEligibleOutputs(
-			dbtx, coinSelectKeyScope, account, minconf, bs,
+			dbtx, coinSelectKeyScope, account, minconf,
+			bs, allowUtxo,
 		)
 		if err != nil {
 			return err
@@ -322,7 +325,8 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 
 func (w *Wallet) findEligibleOutputs(dbtx walletdb.ReadTx,
 	keyScope *waddrmgr.KeyScope, account uint32, minconf int32,
-	bs *waddrmgr.BlockStamp) ([]wtxmgr.Credit, error) {
+	bs *waddrmgr.BlockStamp,
+	allowUtxo func(utxo wtxmgr.Credit) bool) ([]wtxmgr.Credit, error) {
 
 	addrmgrNs := dbtx.ReadBucket(waddrmgrNamespaceKey)
 	txmgrNs := dbtx.ReadBucket(wtxmgrNamespaceKey)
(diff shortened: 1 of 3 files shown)


wallet/wallet.go
```diff
@@ -1259,6 +1261,7 @@ out:
 type txCreateOptions struct {
 	changeKeyScope *waddrmgr.KeyScope
 	selectUtxos    []wire.OutPoint
+	allowUtxo      func(wtxmgr.Credit, int32) bool
```
ProofOfKeags, line 1264: I still haven't been able to convince myself why this needs to take block height as an argument.

wallet/createtx.go
```diff
@@ -181,6 +183,13 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 			)
 
 			for _, e := range eligible {
+				// Restrict the selected utxos if a filter
+				// function is provided.
+				if allowUtxo != nil &&
```
ProofOfKeags, line 188: nowhere in the API docs do we state that a nil predicate has "always return true" semantics. I am not sure how and where to communicate this, but I may suggest that we either add it to the api docs or do away with it entirely and just pass in

```go
always := func(_ A) bool { return true }
```

wallet/wallet.go
```diff
@@ -1289,6 +1292,14 @@ func WithCustomSelectUtxos(utxos []wire.OutPoint) TxCreateOption {
 	}
 }
 
+// WithAllowUtxo is used to restrict the selection of the internal wallet inputs
+// by further external conditions.
+func WithAllowUtxo(allowUtxo func(utxo wtxmgr.Credit) bool) TxCreateOption {
```
ProofOfKeags, line 1297: suggestion: rename this to `UtxoFilter` as it is fundamentally being used for a Filter operation

ProofOfKeags (approved):
I'm good to merge as is but would invite you to consider the suggestions

wallet/createtx.go
```diff
@@ -206,6 +215,14 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
...
 			wrappedEligible := make([]Coin, len(eligible))
 			for i := range eligible {
+				// Restrict the selected utxos if a filter
+				// function is provided.
+				if allowUtxo != nil &&
+					!allowUtxo(eligible[i]) {
+
+					continue
+				}
+
```
Chinwendu20, line 225: Maybe we can move this check to `findEligibleOutputs` function?  Seems inline with the function definition and prevents this duplicity?

wallet/createtx.go
ziggie1984: ahh you mean keeping as is but just add `always` in tests where we would prior use `nil` ?

I think we need go 1.21 for this at least thats what my vs-code linter says:

Even if I do it like this:

```
// always defines a generic filter function which always returns true.
func always[A any](A) bool {
	return true
}
```

and supply it to the function, it tells me:

```
implicitly instantiated function as argument requires go1.21 or latercompiler[UnsupportedFeature](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#UnsupportedFeature)
```

So I just added documentation for now.

ProofOfKeags: Yeah the monomorphic one is fine by me. Also if you really think that leaving nil in place is preferred I can be convinced, but I do think that it means that we need to be consistent about the semantics of `nil` predicates (functions from some A to bool). If we do this, we should commit to a convention that `nil` predicates are equivalent to the `always` I defined above.

ziggie1984: I like the design choice to only using `nil` where we conventionally defined it otherwise it's not good practice. So for sure will stay with the `always` function approach.

wallet/wallet.go
```diff
@@ -1289,6 +1292,15 @@ func WithCustomSelectUtxos(utxos []wire.OutPoint) TxCreateOption {
 	}
 }
 
+// WithFilterUtxo is used to restrict the selection of the internal wallet
+// inputs by further external conditions. Utxos which pass the filter are
+// considered when creating the transaction.
+func WithUtxoFilter(allowUtxo func(utxo wtxmgr.Credit) bool) TxCreateOption {
```
Chinwendu20, line 1298: nit: comment

guggero (approved):
Nice, LGTM :tada:

PR #921, opened 2024-03-27
fix pruned node query

ziggie1984:
- **mod: update btcd version.**
- **chain: add getmempoolentry rpc call.**
- **mod: bump neutrino version.**


go.mod
```diff
@@ -52,3 +52,5 @@ require (
 )
 
 go 1.18
+
+replace github.com/lightninglabs/neutrino v0.16.0 => github.com/ziggie1984/neutrino v0.15.1-0.20240403162326-1ef869ffea1a
```
Roasbeef, line 56: I think we can remove this now.

PR #925, opened 2024-04-08
waddrmgr: add MasterKeyFingerprint to new addresses

guggero:
Related to https://github.com/lightningnetwork/lnd/issues/8626.

When deriving a new address, we'll want to store the correct master key fingerprint from the account it's derived from.
This makes sure PSBT signing succeeds even for imported accounts and their change addresses.


ellemouton (approved):
🔍

Roasbeef (approved):
LGTM 🧞‍♀️

PR #927, opened 2024-04-25
add globalhome root flag to set home root directory

### LINKED ISSUE #926: config: allow changing the default root directory
Add a global option in btcwallet to override the default home directory
This issue is realted to issue [#2106](https://github.com/btcsuite/btcd/issues/2106) in bctd

theedtron:
This pull request adds a flag called global home root which adds all btcwallet data into one specified path. I've also added code to check for the flag and reassigns config, datadir, rpc & rpccert file paths.
Ideally when using the flag globalhomeroot you have the ability to keep all your data in btcwallet, btcd and btcctl in one folder.

Expected outcome
btcwallet --globalhomeroot=/User/Specified/Path
This flag will create a Btcwallet folder in the user specified path and use the path as the default for all data and configs
Fixes #926


Onyekachukwu-Nweke:
Hello @theedtron  

I see no issues in the PR, but it may need a little git surgery like rebase

Cc: @guggero

PR #928, opened 2024-05-13
wallet: Return an error if select utxos in `txtoOutputs` contain duplicates.

Chinwendu20:
As detailed here, the sendcoins command in lnd does not fail if a user supplies duplicate select utxos: 
https://github.com/lightningnetwork/lnd/pull/8516#discussion_r1598296346

This means that we could potentially end up crafting invalid transactions as a result of this bug. This PR fixes that by returning an error in such scenarios. Testcases are also added to test for this.

### DIFF
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -20,6 +20,7 @@ import (
 	"github.com/btcsuite/btcwallet/wallet/txsizes"
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/wtxmgr"
+	"github.com/lightningnetwork/lnd/fn"
 )
 
 func makeInputSource(eligible []Coin) txauthor.InputSource {
@@ -179,6 +180,11 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 
 		var inputSource txauthor.InputSource
 		if len(selectedUtxos) > 0 {
+			dedupUtxos := fn.NewSet(selectedUtxos...)
+			if len(dedupUtxos) != len(selectedUtxos) {
+				return errors.New("selected UTXOs contain " +
+					"duplicate values")
+			}
 			eligibleByOutpoint := make(
 				map[wire.OutPoint]wtxmgr.Credit,
 			)
--- a/wallet/createtx_test.go
+++ b/wallet/createtx_test.go
@@ -449,42 +449,110 @@ func TestSelectUtxosTxoToOutpoint(t *testing.T) {
 	require.Len(t, unspent, 4, "expected 4 unspent "+
 		"utxos")
 
-	selectUtxos := []wire.OutPoint{
+	tCases := []struct {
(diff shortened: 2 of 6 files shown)


Chinwendu20:
This PR is dependent on a new release of the fn package. To be updated

wallet/createtx.go
```diff
@@ -179,6 +180,10 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 
 		var inputSource txauthor.InputSource
 		if len(selectedUtxos) > 0 {
+			if !fn.IsUniqueVals(selectedUtxos) {
```
guggero, line 183: I think it makes sense to just inline this function here, since it's literally a single line, instead of waiting for another PR to be merged and tag to be pushed.

wallet/createtx.go
```diff
@@ -179,6 +180,12 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 
 		var inputSource txauthor.InputSource
 		if len(selectedUtxos) > 0 {
+			if len(fn.NewSet(selectedUtxos...)) !=
```
guggero, line 183: nit: (if line length permits)
```
			dedupUtxos := fn.NewSet(selectedUtxos...)
			if len(dedupUtxos) != len(selectedUtxos) {
```

wallet/createtx_test.go
```diff
@@ -449,42 +449,105 @@ func TestSelectUtxosTxoToOutpoint(t *testing.T) {
...
+	for _, tc := range tCases {
+		t.Run(tc.name, func(t *testing.T) {
+			// Test by sending 200_000.
+			targetTxOut := &wire.TxOut{
+				Value:    200_000,
+				PkScript: p2trScript,
+			}
+			tx1, err := w.txToOutputs(
+				[]*wire.TxOut{targetTxOut}, nil, nil, 0, 1, 1000,
+				CoinSelectionLargest, true, tc.selectUtxos, alwaysAllowUtxo,
```
guggero, line 523: Can you please set up your IDE in a way that you get a visual indication of where the 80 characters cutoff limit is (and setting the tabulator width correctly too)?

guggero (changes requested):

wallet/createtx_test.go
Chinwendu20: Apologies about this, as I just copied and put it in a for block, I did not recheck for the effect on the line length.

Chinwendu20: Please I would like to ask, could you tell by just looking at it or you pulled it into your IDE, or you are using a different platform for review?

guggero (approved):
LGTM :tada:

guggero:
cc @sputn1ck, can't request review through GitHub.

guggero:
Replaced by https://github.com/btcsuite/btcwallet/pull/969 (commit author attribution preserved).

PR #937, opened 2024-07-03
Add err match in the package `chain`

yyforyongyu:
Implemented the idea suggested by @Roasbeef [here](https://github.com/btcsuite/btcd/pull/2209#pullrequestreview-2155124809), we now handle the error mapping separately.

I really like this approach, as the chain backend handling is now independent of `btcd`. Image a case where `bitcoind` updates its error str someday, instead of going through the update chain `btcd -> btcwallet -> lnd`, we can now just update via `btcwallet -> lnd`. Think it's also nice to move `rpcclient` to `btcwallet` since it's a chain backend multiplexer, future PR tho.

The original PR when this was added to btcd: https://github.com/btcsuite/btcd/pull/2138

### DIFF
--- a/chain/bitcoind_client.go
+++ b/chain/bitcoind_client.go
@@ -214,7 +214,27 @@ func (c *BitcoindClient) GetTxOut(txHash *chainhash.Hash, index uint32,
 func (c *BitcoindClient) SendRawTransaction(tx *wire.MsgTx,
 	allowHighFees bool) (*chainhash.Hash, error) {
 
-	return c.chainConn.client.SendRawTransaction(tx, allowHighFees)
+	txid, err := c.chainConn.client.SendRawTransaction(tx, allowHighFees)
+	if err != nil {
+		return nil, c.MapRPCErr(err)
+	}
+
+	return txid, nil
+}
+
+// MapRPCErr takes an error returned from calling RPC methods from various
+// chain backends and maps it to an defined error here.
+func (c *BitcoindClient) MapRPCErr(rpcErr error) error {
+	// Try to match it against bitcoind's error.
+	for i := uint32(0); i < uint32(errSentinel); i++ {
+		err := RPCErr(i)
+		if matchErrStr(rpcErr, err.Error()) {
+			return err
+		}
+	}
+
+	// If not matched, return the original error wrapped.
+	return fmt.Errorf("%w: %v", ErrUndefined, rpcErr)
 }
 
 // TestMempoolAcceptCmd returns result of mempool acceptance tests indicating
--- a/chain/btcd.go
+++ b/chain/btcd.go
@@ -6,6 +6,7 @@ package chain
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 	"time"
 
@@ -577,4 +578,33 @@ func (c *RPCClient) LookupInputMempoolSpend(op wire.OutPoint) (
 	chainhash.Hash, bool) {
 
 	return getTxSpendingPrevOut(op, c.Client)
+}
+
+// MapRPCErr takes an error returned from calling RPC methods from various
+// chain backends and maps it to an defined error here. It uses the
+// `BtcdErrMap`, whose keys are btcd error strings and values are errors made
+// from bitcoind error strings.
+func (c *RPCClient) MapRPCErr(rpcErr error) error {
+	// Iterate the map and find the matching error.
+	for btcdErr, matchedErr := range BtcdErrMap {
+		// Match it against btcd's error.
+		if matchErrStr(rpcErr, btcdErr) {
+			return matchedErr
+		}
+	}
+
+	// If not matched, return the original error wrapped.
+	return fmt.Errorf("%w: %v", ErrUndefined, rpcErr)
+}
+
+// SendRawTransaction sends a raw transaction via btcd.
+func (c *RPCClient) SendRawTransaction(tx *wire.MsgTx,
+	allowHighFees bool) (*chainhash.Hash, error) {
+
+	txid, err := c.Client.SendRawTransaction(tx, allowHighFees)
+	if err != nil {
+		return nil, c.MapRPCErr(err)
+	}
+
+	return txid, nil
 }
--- a/chain/btcd.go
+++ b/chain/btcd.go
@@ -593,6 +593,36 @@ func (c *RPCClient) MapRPCErr(rpcErr error) error {
 		}
 	}
 
+	// If no matching error is found, we try to match it to an older
+	// version of `btcd`.
+	//
+	// Get the backend's version.
+	backend, bErr := c.BackendVersion()
+	if bErr != nil {
+		// If there's an error getting the backend version, we return
+		// the original error and the backend error.
+		return fmt.Errorf("%w: %v, failed to get backend version %v",
+			ErrUndefined, rpcErr, bErr)
+	}
+
+	// If this version doesn't support `testmempoolaccept`, it must be
+	// below v0.24.2. In this case, we will match the errors defined in
+	// pre-v0.24.2.
+	//
+	// NOTE: `testmempoolaccept` is implemented in v0.24.1, but this
+	// version was never tagged, which means it must be v0.24.2 when it's
+	// supported.
+	if !backend.SupportTestMempoolAccept() {
+		// If the backend is older than v0.24.2, we will try to match
+		// the error to the older version of `btcd`.
+		for btcdErr, matchedErr := range BtcdErrMapPre2402 {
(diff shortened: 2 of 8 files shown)


saubyk:
cc @bhandras for review

saubyk:
cc: @ziggie1984 for review

chain/btcd.go
```diff
@@ -593,6 +593,38 @@ func (c *RPCClient) MapRPCErr(rpcErr error) error {
...
+	// version was never tagged, which means it must be v0.24.2 when it's
+	// supported.
+	if backend.SupportTestMempoolAccept() {
+		// Not matched, return the original error wrapped.
+		return fmt.Errorf("%w: %v", ErrUndefined, rpcErr)
+	}
+
+	// If the backend is older than v0.24.2, we will try to match the error
+	// to the older version of `btcd`.
+	for btcdErr, matchedErr := range BtcdErrMapPre2402 {
```
bhandras, line 621: nit: it could be slightly cleaner if we move this loop inside `if !backend.SupportTestMempoolAccept()` so that if the version is >= v0.24.2 we just return the error at the end.

bhandras (approved):
LGTM 🧇

chain/btcd.go
yyforyongyu: yeah was thinking about not nesting `if`s, but I guess it's cleaner so now changed.

chain/errors.go
```diff
@@ -0,0 +1,495 @@
...
+//
+// references:
+// - https://github.com/bitcoin/bitcoin/blob/master/test/functional/rpc_rawtransaction.py#L342
+// - https://github.com/bitcoin/bitcoin/blob/master/test/functional/data/invalid_txs.py
+// - https://github.com/bitcoin/bitcoin/blob/master/test/functional/mempool_accept.py
+// - https://github.com/bitcoin/bitcoin/blob/master/test/functional/mempool_accept_wtxid.py
+// - https://github.com/bitcoin/bitcoin/blob/master/test/functional/mempool_dust.py
+// - https://github.com/bitcoin/bitcoin/blob/master/test/functional/mempool_limit.py
+// - https://github.com/bitcoin/bitcoin/blob/master/src/validation.cpp
+func (r RPCErr) Error() string {
```
ziggie1984, line 217: Q: I wonder why we don't call this error in particular bitcoind error, because it reflects the errors of this particular backend ? Calling it rpcerr, I always think this is a generalized error ?

chain/errors.go
```diff
@@ -0,0 +1,495 @@
...
+
+	return "unknown error"
+}
+
+// BtcdErrMap takes the errors returned from btcd's `testmempoolaccept` and
+// `sendrawtransaction` RPCs and map them to the errors defined above, which
+// are results from calling either `testmempoolaccept` or `sendrawtransaction`
+// in `bitcoind`.
+//
+// Errors not mapped in `btcd`:
```
ziggie1984, line 345: Q: Do you mean these are particular btcd errros which exist only in btcd or the other way around ?

chain/errors.go
```diff
@@ -0,0 +1,495 @@
...
+//   - deployment error from `validateSegWitDeployment`.
+//   - the error when total inputs is higher than max allowed value from
+//     `CheckTransactionInputs`.
+//   - the error when total outputs is higher than total inputs from
+//     `CheckTransactionInputs`.
+//   - errors from `CalcSequenceLock`.
+//
+// NOTE: This is not an exhaustive list of errors, but it covers the
```
ziggie1984, line 353: Hmm I am still a bit surprised why we are moving the error definition to the btcwallet pkg, imo they fit better into the btcd pkg, especially the btcd errors should be defined there anyways because its the official client?

chain/neutrino.go
```diff
@@ -815,3 +815,22 @@ out:
...
+	// Iterate the map and find the matching error.
+	for btcdErr, matchedErr := range BtcdErrMap {
+		// Match it against btcd's error.
+		if matchErrStr(rpcErr, btcdErr) {
+			return matchedErr
+		}
+	}
+
+	// If not matched, return the original error wrapped.
+	return fmt.Errorf("%w: %v", ErrUndefined, rpcErr)
```
ziggie1984, line 835: Q: Shouldn't we wrap it the other way around or even both ?

ziggie1984 (commented):
Looks good thank you for this fix, had some questions.

chain/errors.go
```diff
@@ -477,6 +477,52 @@ var BtcdErrMap = map[string]error{
 	"max-fee-exceeded": ErrMaxFeeExceeded,
 }
 
+// BtcdErrMapPre2402 defines the error mapping for btcd versions prior to
+// 0.24.2 - all the errors changed in this commit have been defined here to
```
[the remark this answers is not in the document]
yyforyongyu: they are not removed, instead, in this [commit](https://github.com/btcsuite/btcd/commit/ef54c49df443815d50765e8c4f31a87944d950a6#diff-37e0f6f3ccf82f3fcc49edc58fdace00678d70956ff412c13c64b764dc58a6f4L1353) they are changed so it's easier to do error matching, as they are more "unique".

chain/errors.go
yyforyongyu: yeah originally it was named `BitcoindRPCErr`, but in reality this functions as the generalized error - when we use it in `lnd`, we don't check for backends, but just check against, say, `chain.ErrMempoolConflict`.

Ideally, we can define another set of errors and put them in a map similar to `BtcdErrMap`, keyed by the error string returned from `bitcoind` and valued by the `RPCErr`, which brings more symmetry. However, unless we have a very different error string returned from the `chain`, there's no point in doing so as in the current design, we always use the error string defined in `bitcoind`. In other words, this `Error()` needs to return something that's human-readable, and we decide to use the error string from `bitcoind`.

chain/errors.go
yyforyongyu: so these errors can be returned from `btcd`, but they are not mapped. so yeah they only exist in `btcd`.

chain/errors.go
yyforyongyu: IMO it's better here. First of all, this set of newly defined error types are not `btcd` errors, they are RPC errors, returned by either calling `sendrawtransaction` or in the `reject-reason` field in a response returned by `testmempoolaccept`. When in `lnd`, we don't care about the specific backend, rather, we just want to get a defined error and act on it.

This means whether it's `btcd`, `bitcoind`, or `neutrino`, we should have a place to catch the implementation-specific errors, and "translate" them into a uniformized RPC error. `chain` is a good place to perform this task as it acts like a backend-multiplexer - it provides a standardized interface `chain.Interface`, and handles the implementation-specific logic here.

On the other hand, `btcd` should just be, well, `btcd`, an independent implementation, that's oblivious about how `neutrino` or `bitcoind` handles stuff, which is why in addition to this error handling, I also think the `btcd/rpcclient` should be moved here. The overall design is, that `btcwallet` provides multi-backend handling, and `btcd` just being a protocol implementation.

chain/bitcoind_client.go
```diff
@@ -214,7 +214,27 @@ func (c *BitcoindClient) GetTxOut(txHash *chainhash.Hash, index uint32,
...
+	txid, err := c.chainConn.client.SendRawTransaction(tx, allowHighFees)
+	if err != nil {
+		return nil, c.MapRPCErr(err)
+	}
+
+	return txid, nil
+}
+
+// MapRPCErr takes an error returned from calling RPC methods from various
+// chain backend and map it to an defined error here.
```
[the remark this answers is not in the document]
yyforyongyu: cool fixed

chain/neutrino.go
```diff
@@ -815,3 +815,22 @@ out:
 	close(s.dequeueNotification)
 	s.wg.Done()
 }
+
+// MapRPCErr takes an error returned from calling RPC methods from various
+// chain backend and map it to an defined error here. It uses the `BtcdErrMap`,
```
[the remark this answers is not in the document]
yyforyongyu: Yea totally, now fixed!

chain/errors.go
ziggie1984: Great explanation thank you

chain/neutrino.go
yyforyongyu: `ErrUndefined` is the defined type tho, so in `lnd` when we see `chain.ErrUndefined` we know it's an undefined error. On the other hand `rpcErr` won't have a type in this case, also we cannot use more than one `%w` in `fmt.Errorf` unfortunately.

ziggie1984 (approved):
LGTM 🔥

PR #941, opened 2024-07-16
wallet: Avoid duplicate address creation

guggero:
Fixes https://github.com/lightningnetwork/lnd/issues/8697.

Because the way the scoped manager uses the OnCommit callback of the
database transaction mechanism, the manager's lock is released first,
then re-acquired in the OnCommit. But during that short time of
releasing the lock, another goroutine might already have updated the
in-memory state.
We add another mutex around the whole database transaction to address
the problem.

### DIFF
--- a/wallet/import.go
+++ b/wallet/import.go
@@ -305,6 +305,15 @@ func (w *Wallet) ImportAccountDryRun(name string,
 	*waddrmgr.AccountProperties, []waddrmgr.ManagedAddress,
 	[]waddrmgr.ManagedAddress, error) {
 
+	// The address manager uses OnCommit on the walletdb tx to update the
+	// in-memory state of the account state. But because the commit happens
+	// _after_ the account manager internal lock has been released, there
+	// is a chance for the address index to be accessed concurrently, even
+	// though the closure in OnCommit re-acquires the lock. To avoid this
+	// issue, we surround the whole address creation process with a lock.
+	w.newAddrMtx.Lock()
+	defer w.newAddrMtx.Unlock()
+
 	var (
 		accountProps  *waddrmgr.AccountProperties
 		externalAddrs []waddrmgr.ManagedAddress
--- a/go.mod
+++ b/go.mod
@@ -26,6 +26,7 @@ require (
 	github.com/stretchr/testify v1.8.4
 	golang.org/x/crypto v0.7.0
 	golang.org/x/net v0.10.0
+	golang.org/x/sync v0.0.0-20180314180146-1d60e4601c6f
 	golang.org/x/term v0.8.0
 	google.golang.org/grpc v1.53.0
 )
@@ -33,10 +34,8 @@ require (
 require (
 	github.com/aead/siphash v1.0.1 // indirect
 	github.com/btcsuite/go-socks v0.0.0-20170105172521-4720035b7bfd // indirect
-	github.com/btcsuite/winsvc v1.0.0 // indirect
 	github.com/decred/dcrd/crypto/blake256 v1.0.0 // indirect
 	github.com/decred/dcrd/lru v1.0.0 // indirect
-	github.com/golang/snappy v0.0.4 // indirect
 	github.com/kkdai/bstream v1.0.0 // indirect
 	github.com/kr/pretty v0.3.0 // indirect
 	github.com/lightningnetwork/lnd/clock v1.0.1 // indirect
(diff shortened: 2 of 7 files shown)


yyforyongyu (commented):
Thanks for the fix!
Tried to run the test after the fix but got a timeout,
```
go test -v -timeout=10s -run=TestDuplicateAddressDerivation
```
Looks like there's a dead lock?

guggero:
> Thanks for the fix! Tried to run the test after the fix but got a timeout,
> 
> ```
> go test -v -timeout=10s -run=TestDuplicateAddressDerivation
> ```
> 
> Looks like there's a dead lock?

No, it just takes quite a while to derive 10x10x10 addresses... Can you try again with something like 30 seconds or more?

wallet/import.go
```diff
@@ -305,6 +305,15 @@ func (w *Wallet) ImportAccountDryRun(name string,
 	*waddrmgr.AccountProperties, []waddrmgr.ManagedAddress,
 	[]waddrmgr.ManagedAddress, error) {
 
+	// The address manager uses OnCommit on the walletdb tx to update the
+	// in-memory state of the account state. But because the commit happens
+	// _after_ the account manager internal lock has been released, there
+	// is a chance for the address index to be accessed concurrently, even
+	// though the closure in OnCommit re-acquires the lock. To avoid this
+	// issue, we surround the whole address creation process with a lock.
```
Roasbeef, line 313: In commit text: `mutext` -> `mutex`

Roasbeef:
If I run the test w/o the second commit, I hit it pretty quickly: 
```
=== RUN   TestDuplicateAddressDerivation
    wallet_test.go:359:
                Error Trace:    /Users/roasbeef/gocode/src/github.com/btcsuite/btcwallet/wallet/wallet_test.go:359
                Error:          Received unexpected error:
                                duplicate address! already have tb1qagh3rua975yy0shaw9nhgnm6avaect9te6np4v, want to add tb1qagh3rua975yy0shaw9nhgnm6avaect9te6np4v
                Test:           TestDuplicateAddressDerivation
--- FAIL: TestDuplicateAddressDerivation (3.39s)
FAIL
exit status 1
FAIL    github.com/btcsuite/btcwallet/wallet    3.699s
```

yyforyongyu (commented):
Cool the test passed,
```
=== RUN   TestDuplicateAddressDerivation
--- PASS: TestDuplicateAddressDerivation (20.86s)
PASS
ok      github.com/btcsuite/btcwallet/wallet    21.450s
```

Since deep down they are all calling `nextAddresses`, also noticed that,
- `NewChangeAddress -> newChangeAddress -> NextInternalAddresses`
- `CurrentAddress -> newAddress -> NextExternalAddresses`

Should we also update these calls?

The alternative, instead of locking over locks, we can make these indices atomic, not sure if it helps,
https://github.com/btcsuite/btcwallet/blob/e391a1c31df220939102b26724f863925fbbcde6/waddrmgr/manager.go#L157-L165

Roasbeef:
@yyforyongyu I don't think that's enough as we have other operations in the `OnCommit` call back: https://github.com/btcsuite/btcwallet/blob/e391a1c31df220939102b26724f863925fbbcde6/waddrmgr/scoped_manager.go#L1199-L1209

guggero:
Thanks for taking a closer look, @yyforyongyu !

You are correct, I missed a couple of paths. I now traced back the `s.nextAddresses` call sites to any place where we start a transaction `walletdb.Update()` and wrapped it with an additional lock.

Here are the ones I found and now added the new mutex to:

```
w.ImportAccountDryRun                                            -> s.NextInternalAddresses -> s.nextAddresses
w.NewChangeAddress -> w.newChangeAddress                         -> s.NextInternalAddresses -> s.nextAddresses
w.txToOutputs -> w.addrMgrWithChangeSource -> w.newChangeAddress -> s.NextInternalAddresses -> s.nextAddresses
w.FundPsbt    -> w.addrMgrWithChangeSource -> w.newChangeAddress -> s.NextInternalAddresses -> s.nextAddresses

w.ImportAccountDryRun                 -> s.NextExternalAddresses -> s.nextAddresses
w.NewAddress     -> w.newAddress -> s.NextExternalAddresses -> s.nextAddresses
w.CurrentAddress -> w.newAddress -> s.NextExternalAddresses -> s.nextAddresses
```

I hope that's all of them.

yyforyongyu (approved):
LGTM🙏 I think before merging this we can create a PR in `lnd` first to see if the CI passes.

Roasbeef:
CI passing on the `lnd` itests, so moving to merge this.

Roasbeef (approved):
LGTM 💰

PR #944, opened 2024-08-07
end recovery on shutdown

buck54321:
The `(*Wallet).recovery` loop is unmonitored, and shutting down the wallet during recovery without locking the wallet first will hang. This change ensures that the recovery loop is ended when `(*Wallet).Stop` is called.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -280,6 +280,8 @@ func (w *Wallet) quitChan() <-chan struct{} {
 
 // Stop signals all wallet goroutines to shutdown.
 func (w *Wallet) Stop() {
+	<-w.endRecovery()
+
 	w.quitMu.Lock()
 	quit := w.quit
 	w.quitMu.Unlock()
@@ -1380,6 +1382,23 @@ type (
 	heldUnlock chan struct{}
 )
 
+// endRecovery tells (*Wallet).recovery to stop, if running, and returns a
+// channel that will be closed when the recovery routine exits.
+func (w *Wallet) endRecovery() <-chan struct{} {
+	if recoverySyncI := w.recovering.Load(); recoverySyncI != nil {
+		recoverySync := recoverySyncI.(*recoverySyncer)
+
+		// If recovery is still running, it will end early with an error
+		// once we set the quit flag.
+		atomic.StoreUint32(&recoverySync.quit, 1)
+
+		return recoverySync.done
+	}
+	c := make(chan struct{})
+	close(c)
+	return c
+}
+
 // walletLocker manages the locked/unlocked state of a wallet.
 func (w *Wallet) walletLocker() {
 	var timeout <-chan time.Time
@@ -1472,19 +1491,7 @@ out:
 
 		// We can't lock the manager if recovery is active because we use
 		// cryptoKeyPriv and cryptoKeyScript in recovery.
-		if recoverySyncI := w.recovering.Load(); recoverySyncI != nil {
-			recoverySync := recoverySyncI.(*recoverySyncer)
-			// If recovery is still running, it will end early with an error
-			// once we set the quit flag.
-			atomic.StoreUint32(&recoverySync.quit, 1)
-
-			select {
-			case <-recoverySync.done:
-			case <-quit:
-				break out
-			}
-
-		}
+		<-w.endRecovery()
 
 		timeout = nil
 		err := w.Manager.Lock()
--- a/wallet/wallet_test.go
+++ b/wallet/wallet_test.go
@@ -3,12 +3,16 @@ package wallet
 import (
 	"encoding/hex"
 	"fmt"
+	"math"
+	"strings"
 	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/wtxmgr"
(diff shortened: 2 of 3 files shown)


wallet/wallet.go
```diff
@@ -1380,6 +1382,22 @@ type (
 	heldUnlock chan struct{}
 )
 
+// endRecoveryAndWait tells (*Wallet).recovery to stop, if running, and waits
+// for it to exit.
+func (w *Wallet) endRecoveryAndWait() {
+	if recoverySyncI := w.recovering.Load(); recoverySyncI != nil {
+		recoverySync := recoverySyncI.(*recoverySyncer)
+		// If recovery is still running, it will end early with an error
```
guggero, line 1390: nit: (pre-existing): Add newline before comment?

guggero (approved):
tACK, LGTM :tada:

martonp (approved):

Roasbeef (changes requested):
I think we should add a new unit test that shows the deadlock that can happen re improper shutdown. Then this patch could be applied above, demonstrating concrete resolution.

wallet/wallet_test.go
```diff
@@ -359,3 +363,125 @@ func TestDuplicateAddressDerivation(t *testing.T) {
...
+			case <-time.After(time.Second):
+				t.Fatal("expected BlockHash to be called")
+			}
+			if i == expCalls {
+				break
+			}
+		}
+	}
+
+	// Recovery is running
```
guggero, line 420: nit: missing full stop, here and below.

guggero (approved):
Commits could be squashed, otherwise looks good. Thanks for the test.

Roasbeef (commented):
So I tried to get the test to fail with the fix to call `endRecovery` in the `Stop` method with this patch: 
```diff
diff --git a/wallet/wallet.go b/wallet/wallet.go
index 4bde6225..7d93d153 100644
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -280,8 +280,6 @@ func (w *Wallet) quitChan() <-chan struct{} {
 
 // Stop signals all wallet goroutines to shutdown.
 func (w *Wallet) Stop() {
-	<-w.endRecovery()
-
 	w.quitMu.Lock()
 	quit := w.quit
 	w.quitMu.Unlock()
diff --git a/wallet/wallet_test.go b/wallet/wallet_test.go
index 4c1efd7f..019958e0 100644
--- a/wallet/wallet_test.go
+++ b/wallet/wallet_test.go
@@ -420,11 +420,9 @@ func TestEndRecovery(t *testing.T) {
 	// Recovery is running
 	getBlockHashCalls(3)
 
-	// Closing the quit channel, e.g. Stop() without endRecovery, alone will not
-	// end the recovery loop.
-	w.quitMu.Lock()
-	close(w.quit)
-	w.quitMu.Unlock()
+	// Call stop directly simulating a normal shutdown.
+	w.Stop()
+
 	// Continues scanning.
 	getBlockHashCalls(3)
 
@@ -471,9 +469,7 @@ func TestEndRecovery(t *testing.T) {
 
 	// testWallet starts a couple of other unrelated goroutines that need to be
 	// killed, so we still need to close the quit channel.
-	w.quitMu.Lock()
-	close(w.quit)
-	w.quitMu.Unlock()
+	w.Stop()
 
 	select {
 	case <-waitedForShutdown:
```

With that applied locally, when running both with and without the race condition detector the test still passes. I think the fix itself is sound (now actually signal the recovery loop for a force exit when `Stop` is called), but I think the test needs a bit more tuning. 

Perhaps what we want to do instead, is introspect a bit more into the call to `recovery`? 

https://github.com/btcsuite/btcwallet/blob/aba7c3563bfbfd50cefaf9d4f4c7043ef20ec713/wallet/wallet.go#L468-L475

Roasbeef:
Scratch the comment above, I needed another change to the test, stopping it from explicitly calling `endRecovery`:
```diff
diff --git a/wallet/wallet.go b/wallet/wallet.go
index 4bde6225..7d93d153 100644
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -280,8 +280,6 @@ func (w *Wallet) quitChan() <-chan struct{} {
 
 // Stop signals all wallet goroutines to shutdown.
 func (w *Wallet) Stop() {
-	<-w.endRecovery()
-
 	w.quitMu.Lock()
 	quit := w.quit
 	w.quitMu.Unlock()
diff --git a/wallet/wallet_test.go b/wallet/wallet_test.go
index 4c1efd7f..06ebb45c 100644
--- a/wallet/wallet_test.go
+++ b/wallet/wallet_test.go
@@ -420,11 +420,9 @@ func TestEndRecovery(t *testing.T) {
 	// Recovery is running
 	getBlockHashCalls(3)
 
-	// Closing the quit channel, e.g. Stop() without endRecovery, alone will not
-	// end the recovery loop.
-	w.quitMu.Lock()
-	close(w.quit)
-	w.quitMu.Unlock()
+	// Call stop directly simulating a normal shutdown.
+	w.Stop()
+
 	// Continues scanning.
 	getBlockHashCalls(3)
 
@@ -461,19 +459,9 @@ func TestEndRecovery(t *testing.T) {
 	// Recovery is running
 	getBlockHashCalls(3)
 
-	// endRecovery is required to exit the unmonitored goroutine.
-	end := w.endRecovery()
-	select {
-	case <-blockHashCalled:
-	case <-recoveryDone:
-	}
-	<-end
-
 	// testWallet starts a couple of other unrelated goroutines that need to be
 	// killed, so we still need to close the quit channel.
-	w.quitMu.Lock()
-	close(w.quit)
-	w.quitMu.Unlock()
+	w.Stop()
 
 	select {
 	case <-waitedForShutdown:
@@ -481,6 +469,7 @@ func TestEndRecovery(t *testing.T) {
 		t.Fatal("WaitForShutdown never returned")
 	}
 
 	if !strings.EqualFold(err.Error(), "recovery: forced shutdown") {
 		t.Fatal("wrong error")
 	}
```

Roasbeef (approved):
LGTM 🐊

I think this is g2g once the commits are squashed with some of the minor comment style comments addressed.

wallet/wallet.go
```diff
@@ -138,7 +138,7 @@ type Wallet struct {
 	lockedOutpoints    map[wire.OutPoint]struct{}
 	lockedOutpointsMtx sync.Mutex
 
-	recovering     atomic.Value
+	recovering     atomic.Value // *recoverySyncer
```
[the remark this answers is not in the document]
buck54321: Just something we usually do for our `atomic.Value` fields for documentation of intention, but removing.

PR #945, opened 2024-08-08
wallet: break `FetchInputInfo` into two methods

yyforyongyu:
This PR breaks `FetchInputInfo` into two methods, `FetchOutpointInfo` and `FetchDerivationInfo` so the caller can better control what to fetch from the wallet db.

### DIFF
--- a/wallet/utxos.go
+++ b/wallet/utxos.go
@@ -104,66 +104,42 @@ func (w *Wallet) UnspentOutputs(policy OutputSelectionPolicy) ([]*TransactionOut
 
 // FetchInputInfo queries for the wallet's knowledge of the passed outpoint. If
 // the wallet determines this output is under its control, then the original
-// full transaction, the target txout and the number of confirmations are
-// returned. Otherwise, a non-nil error value of ErrNotMine is returned instead.
+// full transaction, the target txout, the derivation info and the number of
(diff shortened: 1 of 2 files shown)


wallet/utxos.go
```diff
@@ -104,66 +104,42 @@ func (w *Wallet) UnspentOutputs(policy OutputSelectionPolicy) ([]*TransactionOut
...
+// full transaction, the target txout, the derivation info and the number of
+// confirmations are returned. Otherwise, a non-nil error value of ErrNotMine
+// is returned instead.
 func (w *Wallet) FetchInputInfo(prevOut *wire.OutPoint) (*wire.MsgTx,
 	*wire.TxOut, *psbt.Bip32Derivation, int64, error) {
 
-	// We manually look up the output within the tx store.
-	txid := &prevOut.Hash
-	txDetail, err := UnstableAPI(w).TxDetails(txid)
+	tx, txOut, value, err := w.FetchOutpointInfo(prevOut)
```
guggero, line 115: I think `value` as a variable name here is very misleading. Isn't this number of confirmations?

guggero (approved):
Very nice, makes a lot of sense to make the two queries accessible independently. LGTM pending one variable rename.

wallet/utxos.go
yyforyongyu: fixed!

PR #946, opened 2024-08-14
gomod: update go mod using `go work sync`

yyforyongyu:
Sync versions across repos.

Depends on
- #947


guggero (approved):

Roasbeef:
Hmm, I think we might need to do this in stages so we can tag new versions of the various sub-modules incrementally.

Makefile
```diff
@@ -8,8 +8,9 @@ GO_BIN := ${GOPATH}/bin
 LINT_BIN := $(GO_BIN)/golangci-lint
 GOACC_BIN := $(GO_BIN)/go-acc
 
-LINT_COMMIT := v1.46.0
-GOACC_COMMIT := 80342ae2e0fcf265e99e76bcc4efd022c7c3811b
+LINT_VERSION := v1.60.1
```
guggero, line 11: If we're at it, could we also use the Dockerized linter that we use everywhere else, so we don't have to have multiple versions installed?

.golangci.yml
```diff
@@ -11,31 +11,89 @@ linters-settings:
...
+    # exactly a test failed in the generated failure stack trace.
+    - thelper
+    
+    # The linter is too aggressive and doesn't add much value since reviewers
+    # will also catch magic numbers that make sense to extract.
+    - gomnd
 
 issues:
+  # Only check issues in the new code.
+  new-from-rev: HEAD~1
```
guggero, line 95: Wouldn't this cause to only look at code in the very last commit? Shouldn't we pin this to a fixed version and update every once in a while?

guggero (commented):
Yayy, green CI. Thanks for the update!

guggero:
Oops, I just realized that all my comments should've been in https://github.com/btcsuite/btcwallet/pull/947...

guggero:
Can be rebased. Pushed the following tags:

```
To github.com:btcsuite/btcwallet.git
 * [new tag]           wallet/txauthor/v1.3.5 -> wallet/txauthor/v1.3.5
 * [new tag]           wallet/txrules/v1.2.2 -> wallet/txrules/v1.2.2
 * [new tag]           wallet/txsizes/v1.2.5 -> wallet/txsizes/v1.2.5
 * [new tag]           walletdb/v1.4.3 -> walletdb/v1.4.3
```

.golangci.yml
yyforyongyu: yeah was experimenting sth...basically I was trying to use `new-from-rev: master` - the idea is to only lint the diff between the current PR and the master, but couldn't get it to work due to this error,
```
level=warning msg="[runner] Can't process result by diff processor: can't prepare diff by revgrep: could not read git repo: error executing \"git diff --color=never --no-ext-diff --default-prefix --relative master --\": exit status 128: fatal: bad revision 'master'\n"
```
seems to be a `revgrep.go` [issue](https://github.com/golangci/golangci-lint/issues/161).

Changed to commit instead.

PR #948, opened 2024-09-03
chain: match on new bitcoind v28.0 errors

guggero:
An error message was renamed in https://github.com/bitcoin/bitcoin/pull/30212.

We should probably start matching on the [`reject-reason`](https://github.com/bitcoin/bitcoin/blob/4c526f575cde43cc49c0236f4d2b15a5fbb6d7ab/src/rpc/mempool.cpp#L146) returned by `testmempoolaccept` RPC instead, as that's less likely to change in the future.

### DIFF
--- a/chain/bitcoind_client.go
+++ b/chain/bitcoind_client.go
@@ -233,6 +233,15 @@ func (c *BitcoindClient) MapRPCErr(rpcErr error) error {
 		}
 	}
 
+	// Perhaps the backend is a newer version of bitcoind, try to match it
+	// against the v28.0 and later errors.
+	for btcdErr, matchedErr := range Bitcoind28ErrMap {
+		// Match it against btcd's error.
+		if matchErrStr(rpcErr, btcdErr) {
+			return matchedErr
+		}
+	}
+
 	// If not matched, return the original error wrapped.
 	return fmt.Errorf("%w: %v", ErrUndefined, rpcErr)
 }
--- a/chain/errors.go
+++ b/chain/errors.go
@@ -337,6 +337,14 @@ func (r RPCErr) Error() string {
 	return "unknown error"
 }
 
+// Bitcoind28ErrMap contains error messages from bitcoind version v28.0 (and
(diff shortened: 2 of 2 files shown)


Roasbeef (commented):
Re using reject reason instead, we _do_ use that, as we pass it in to this `MapRpcErr` function: https://github.com/lightningnetwork/lnd/blob/e8c5e7d5ce4b79117ef9ab497745fea3a1bffb36/lnwallet/btcwallet/btcwallet.go#L1280

yyforyongyu:
I think this works, but like laolu said we already do the mapping over `reject-reason`. If we believe this field is less likely to be changed, what we can do instead is to always do a `testmempoolaccept` check in `SendRawTransaction` here since we can assume once it's the `bitcoind` backend this RPC must be supported.

guggero:
I was confused about this too... So I did some digging and we _do_ match on `reject-reason`. _BUT_ we have a special case where we still publish the transaction, even when `testmempoolaccept` returns `false`: https://github.com/lightningnetwork/lnd/blob/cc9e2b783ec983fbe3995d47d31898aa122b11c8/lnwallet/btcwallet/btcwallet.go#L1302

And in that case we then get the new error message, which we need to map as well.
So I think we still need this, cc @yyforyongyu @Roasbeef.

yyforyongyu:
> I was confused about this too... So I did some digging and we _do_ match on `reject-reason`. _BUT_ we have a special case where we still publish the transaction, even when `testmempoolaccept` returns `false`: https://github.com/lightningnetwork/lnd/blob/cc9e2b783ec983fbe3995d47d31898aa122b11c8/lnwallet/btcwallet/btcwallet.go#L1302

Yeah we do call publish tx even if there's an error in `lnd`,  to put the tx label in the db...we may need to fix it there, meanwhile this should be good to go.

Roasbeef:
Ahh yeh, we call again just to insert it as a label...

Roasbeef (approved):
LGTM 🎻

PR #950, opened 2024-09-11
Update bbolt to latest release

### LINKED ISSUE #949: Unable to create a new wallet
Running master branch:

```
$ btcwallet --usespv -u jrick -P banana --testnet --create
Enter the private passphrase for your new wallet: 
Confirm passphrase: 
Do you want to add an additional layer of encryption for public data? (n/no/y/yes) [no]: 
Do you have an existing wallet seed you want to use? (n/no/y/yes) [no]: 
Your wallet generation seed is:
378610a42da8f3339155aed68c00f889c35fd6f21a62193bced79704797eabf6
IMPORTANT: Keep the seed in a safe place as you
will NOT be able to restore your wallet without it.
Please keep in mind that anyone who has access
to the seed can also restore your wallet thereby
giving them access to all your funds, so it is
imperative that you keep it in a secure location.
Once you have stored the seed in a safe and secure location, enter "OK" to continue: OK
Creating the wallet...
Unable to create wallet: function not implemented
```

Surely I'm not the first to notice this...?

jrick:
This fixes btcwallet on OpenBSD.  Previously, any bbolt methods would error with "function not implemented".

Fixes #949.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -44,7 +44,7 @@ require (
 	github.com/pmezard/go-difflib v1.0.0 // indirect
 	github.com/rogpeppe/go-internal v1.12.0 // indirect
 	github.com/stretchr/objx v0.5.2 // indirect
-	go.etcd.io/bbolt v1.3.7 // indirect
+	go.etcd.io/bbolt v1.3.11 // indirect
 	golang.org/x/sys v0.19.0 // indirect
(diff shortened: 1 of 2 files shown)


Roasbeef (approved):
LGTM 🦜

PR #958, opened 2024-10-07
waddrmgr: fix lock order in importScriptAddress

JssDWt:
First of all, the code between the scoped manager and the root manager needs a full rewrite. The lock orders are unmanageable. This PR fixes a potential deadlock (seen on Breez), but it doesn't fix the underlying issue here. The main problem is the locks between scoped manager and the root manager are taken in different orders. Sometimes root manager first, then scoped manager, sometimes the other way around. 

For this PR the decision is to never lock the root manager inside a scoped manager lock. This PR only fixes a single occurence of this issue, there are more occurences elsewhere.

This PR has been in production for a swap server for Breez for a few months now and fixes our deadlock occurrences.

### DIFF
--- a/waddrmgr/scoped_manager.go
+++ b/waddrmgr/scoped_manager.go
@@ -2200,7 +2200,12 @@ func (s *ScopedKeyManager) importScriptAddress(ns walletdb.ReadWriteBucket,
 	error) {
 
 	s.mtx.Lock()
-	defer s.mtx.Unlock()
+	unlockNeeded := true
+	defer func() {
+		if unlockNeeded {
+			s.mtx.Unlock()
+		}
+	}()
 
 	// The manager must be unlocked to encrypt the imported script.
 	if isSecretScript && s.rootManager.IsLocked() {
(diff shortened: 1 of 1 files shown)


waddrmgr/scoped_manager.go
```diff
@@ -2324,6 +2321,20 @@ func (s *ScopedKeyManager) importScriptAddress(ns walletdb.ReadWriteBucket,
 	// Add the new managed address to the cache of recent addresses and
 	// return it.
 	s.addrs[addrKey(scriptIdent)] = managedAddr
+
+	if updateStartBlock {
+		// Now that the database has been updated, update the start block in
```
guggero, line 2326: nit: line is slightly over 80 characters if a tabulator size/width of 8 spaces is used.

waddrmgr/scoped_manager.go
```diff
@@ -2324,6 +2321,20 @@ func (s *ScopedKeyManager) importScriptAddress(ns walletdb.ReadWriteBucket,
...
+		// memory too if needed. Ensure the manager lock goes outside the scoped
+		// manager lock.
+		s.mtx.Unlock()
```
guggero, line 2329: Since we're releasing the lock here, it's possible for a second/parallel call to `importScriptAddress` to come in and potentially progress to the `if bs.Height < s.rootManager.syncState.startBlock.Height` check before the below is updated, similar to how we could end up with duplicated addresses in https://github.com/lightningnetwork/lnd/issues/8697. BUT, since we're just increasing the start block, the worst that could happen is that this write would happen twice. So IMO an acceptable solution until we do a full rewrite of this logic (which I agree is definitely necessary at some point).

guggero (approved):
LGTM :tada:

PR #959, opened 2024-10-28
Fix:nil pointer

shnpd:
Error when spending an already used UTXO: **“runtime error: invalid memory address or nil pointer dereference”**

According to my observation, in **btcwallet/rpc/legacyrpc/methods.go:1716** the call to hex.DecodeString(result.ScriptPubKey.Hex) with the parameter **“result” is nil**. This is because when calling resp.Receive() on line 1712, if you use a spent UTXO, it will return (nil,nil), so need to handle the error if result is nil.


guggero:
Replaced by https://github.com/btcsuite/btcwallet/pull/969 (commit author attribution preserved).

PR #964, opened 2024-11-12
chain: make sure to `WaitForShutdown`

yyforyongyu:
Simple fix - looks like we forgot to call `WaitForShutdown` and, add a nil bucket check which was found panic in one of `lnd`'s itests.

### DIFF
--- a/chain/btcd.go
+++ b/chain/btcd.go
@@ -242,6 +242,7 @@ func (c *RPCClient) Stop() {
 	default:
 		close(c.quit)
 		c.Client.Shutdown()
+		c.Client.WaitForShutdown()
 
 		if !c.started {
 			close(c.dequeueNotification)


guggero (approved):
Thanks for the fixes, LGTM :tada:

wallet/chainntfns.go
```diff
@@ -242,6 +247,9 @@ func (w *Wallet) handleChainNotifications() {
 // the passed block.
 func (w *Wallet) connectBlock(dbtx walletdb.ReadWriteTx, b wtxmgr.BlockMeta) error {
 	addrmgrNs := dbtx.ReadWriteBucket(waddrmgrNamespaceKey)
+	if addrmgrNs == nil {
+		return fmt.Errorf("%w: empty waddrmgr bucket", ErrEmptyBucket)
+	}
```
ellemouton, line 252: hmmm wondering how this can be nil if it is always created on DB construction via the `create` function. It's also used in many other places without first doing a nil check. So if we can determine why it can be nil here, then those would also need to be updated?

ellemouton: maybe the lnd itest panic reveals some info re the path taken here?

yyforyongyu: good question - should have kept the logs back then. This was seen in postgres, that somehow there's a goroutine accessing this data during shutdown, yet I guess the data was already GCed at the moment. Now that you mention it I think we should just let it happen and then decide what was the cause there, removed.

ellemouton (approved):
lgtm! 🤖

PR #967, opened 2024-11-25
waddrmgr: fix deadlock

aakselrod:
This PR fixes a deadlock caused by methods of `ScopedManager` read-locking the root manager for `WatchOnly()` and `Locked()` calls while the `ScopedManager` itself is locked. This is the reverse order from the normal way of taking locks, but has been hidden by the fact that DB transactions have never happened in parallel until the upcoming postgres changes in LND.

See discussion [here](https://github.com/lightningnetwork/lnd/pull/9242#issuecomment-2495544838) for more details.

The first commit updates a test to cause this failure. The following commits, pushed after CI completes/tests fail, contain the fix.

### DIFF
--- a/waddrmgr/scoped_manager.go
+++ b/waddrmgr/scoped_manager.go
@@ -447,7 +447,7 @@ func (s *ScopedKeyManager) loadAccountInfo(ns walletdb.ReadBucket,
 
 	// The wallet will only contain private keys for default accounts if the
 	// wallet's not set up as watch-only and it's been unlocked.
-	watchOnly := s.rootManager.watchOnly()
+	watchOnly := s.rootManager.WatchOnly()
 	hasPrivateKey := !s.rootManager.isLocked() && !watchOnly
 
 	// Create the new account info with the known information. The rest of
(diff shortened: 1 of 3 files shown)


waddrmgr/scoped_manager.go
```diff
@@ -775,9 +775,7 @@ func (s *ScopedKeyManager) deriveKeyFromPath(ns walletdb.ReadBucket,
 func (s *ScopedKeyManager) chainAddressRowToManaged(ns walletdb.ReadBucket,
 	row *dbChainAddressRow) (ManagedAddress, error) {
 
-	// Since the manger's mutex is assumed to held when invoking this
```
aakselrod, line 778: I feel like using `atomic.Bool` is safer than just not using the mutexes at all but if you disagree, please let me know and I'll change the rest of the code to do it this way. @guggero @Roasbeef

yyforyongyu: yeah I like `atomic.Bool` - unless there is a set of states must be updated together I always prefer using it as it's way cheaper. And I think it makes sense to use it here.

guggero (approved):
Nice find, LGTM :tada:

losh11:
We found some other places where locking was broken, see:
https://github.com/ltcsuite/ltcwallet/commit/69d6e9a93b5a61625c19043367fe4a4ad2acffb9
https://github.com/ltcsuite/ltcwallet/commit/6570a36a5fa0d5c0490364b4b5867942e42a031b

guggero:
@aakselrod would you have time to create a PR from the two commits linked above?

guggero:
Thanks for the hint, @losh11. Though I took a quick look at the two commits and it looks like the way @aakselrod fixed the code in this PR (turning `watchingOnly` and `locked` of the root manager into `atomic.Bool` values instead of locking the whole root manager), that should eliminate a whole class of potential deadlocks.
Didn't have time to look into things more deeply though.

I merged the two commits and rebased/fixed them for `btcwallet` here: https://github.com/guggero/btcwallet/tree/more-locking (in case anyone wants to take a closer look).

PR #969, opened 2024-12-09
Collect small fixes from stale contributor PRs

guggero:
Replaces https://github.com/btcsuite/btcwallet/pull/959.
Replaces https://github.com/btcsuite/btcwallet/pull/824.
Replaces https://github.com/btcsuite/btcwallet/pull/928.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,6 +21,7 @@ require (
 	github.com/lightninglabs/gozmq v0.0.0-20191113021534-d20a764486bf
 	github.com/lightninglabs/neutrino v0.16.0
 	github.com/lightninglabs/neutrino/cache v1.1.2
+	github.com/lightningnetwork/lnd/fn/v2 v2.0.2
 	github.com/lightningnetwork/lnd/ticker v1.0.0
 	github.com/lightningnetwork/lnd/tlv v1.0.2
 	github.com/stretchr/testify v1.9.0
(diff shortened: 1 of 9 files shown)


rpc/legacyrpc/methods.go
```diff
@@ -1713,6 +1719,9 @@ func signRawTransaction(icmd interface{}, w *wallet.Wallet, chainClient *chain.R
 		if err != nil {
 			return nil, err
 		}
+		if result == nil {
+			return nil, errUtxoSpent
+		}
```
ellemouton, line 1724: ok cool - i guess ideally we could change Receive to return an error if the result is nil and then dont need to explicitly handle this. but i see we already do this special case handling in lnd for all calls to `Receive` done via `GetTxOut`.

ellemouton (approved):
lgtm! thanks 🙏

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	github.com/lightninglabs/gozmq v0.0.0-20191113021534-d20a764486bf
 	github.com/lightninglabs/neutrino v0.16.0
 	github.com/lightninglabs/neutrino/cache v1.1.2
+	github.com/lightningnetwork/lnd/fn v1.2.5
```
yyforyongyu, line 24: didn't realize we were importing modules from `lnd` before😮‍💨

yyforyongyu (commented):
Looks like the liner is complaining - also can we open a PR in `lnd` to test this change?

guggero:
`lnd` PR is here: https://github.com/lightningnetwork/lnd/pull/9346.

go.mod
guggero: Yeah. But those are all submodules, so it shouldn't really matter. But noticed this actually needs to be `github.com/lightningnetwork/lnd/fn/v2`, so I changed that.

yyforyongyu:
hmm think `lnd` CI is broken...

guggero:
Yeah, I'll take a look at some point. Seems like the early rescan abort shortcut breaks things. Perhaps because we re-use existing rescans, but then it's already aborted?

yyforyongyu:
Cool let's also bump https://github.com/lightningnetwork/lnd/pull/9346 - if the CI passes we can merge

guggero:
The CI doesn't pass... The whole "skip rescan early" thing is more involved than initially thought. I'm wondering if we should just drop it for now... I'll invest just a bit more time and then decide.

ellemouton (approved):
🙏

PR #974, opened 2025-02-03
Update scanning state of NeutrinoClient

ziggie1984:


guggero (approved):
Thanks for investigating! Makes sense as a short-term fix. But looks like we really need to refactor the re-scan logic at some point.

saubyk:
cc @yyforyongyu for review

chain/neutrino.go
```diff
@@ -801,9 +801,18 @@ out:
 
 		case err := <-rescanErr:
 			if err != nil {
-				log.Errorf("Neutrino rescan ended with error: %s", err)
+				log.Errorf("Neutrino rescan ended with "+
```
yyforyongyu, line 804: Do we know what the error is? What is the root cause?

chain/neutrino.go
```diff
@@ -801,9 +801,18 @@ out:
...
+					"error: %s", err)
 			}
 
+			// We need to signal to the client that the rescan
+			// has finished otherwise updates to the rescan
+			// object will fail.
+			log.Info("Neutrino rescan finished")
+			s.clientMtx.Lock()
+			s.scanning = false
```
yyforyongyu, line 813: If we don't know what caused the rescan error, this will just create a new scan and then getting the same error?

chain/neutrino.go
ziggie1984: I am trying to obtain the logs from the user, but I think his node is generally working fine, my next step will be to let the user run the patch to see if that fixes the problem. But as Oliver and Laolu pointed out, this code part needs a refactor.

chain/neutrino.go
ziggie1984: I am not sure, basically the link failed because the `NewAddress` (NotifiedReceived) returned an error when registering the wtclient for a specific link:

https://github.com/lightningnetwork/lnd/blob/6bf895aeb929fb64265bfb226f796392921d8d3a/htlcswitch/link.go#L544-L550

user at least reported that now the link is active, so seems like the backend still works correctly maybe some error failed the rescan and now we need to just restart it, but yeah let's hope we get more information from the user.

PR #979, opened 2025-02-25
waddrmgr: Put client request first in recovery.

### LINKED ISSUE #978: sync: failed to fetch block hash for height ####: block not found
This seems to be a random sync error. Its been around forever but I don't see an issue for it. The particular block seems to differ, but sync will get stuck with this error and never fix itself, even restarting:

```
[ERR] BTCW: Unable to synchronize wallet to chain, trying again in 5s: unable to perform wallet recovery: failed to store sync information 000000000000000000007101d8cb8f46c7c17c7c12f873eb0f717d8751628c08: failed to fetch block hash for height 858852: block not found
```

The block exists but is reported as block 858,853 here, so I guess its not finding the parent block hash? https://www.blockchain.com/explorer/blocks/btc/000000000000000000007101d8cb8f46c7c17c7c12f873eb0f717d8751628c08

JoeGruffins:
closes #978

~I'm unsure if this is the best solution but it gets me past the block in my issue. It looks to me like this should always hit when restoring with a birthday, but it does not. Does btcwallet not download all headers anyway?~

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -790,16 +790,25 @@ func (w *Wallet) recovery(chainClient chain.Interface,
 		if len(recoveryBatch) == recoveryBatchSize || height == bestHeight {
 			err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
 				ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+				if err := w.recoverScopedAddresses(
+					chainClient, tx, ns, recoveryBatch,
+					recoveryMgr.State(), scopedMgrs,
+				); err != nil {
+					return err
+				}
+
+				// TODO: Any error here will roll back this
+				// entire tx. This may cause the in memory sync
+				// point to become desyncronized. Refactor so
+				// that this cannot happen.
 				for _, block := range blocks {
 					err := w.Manager.SetSyncedTo(ns, block)
 					if err != nil {
 						return err
 					}
 				}
-				return w.recoverScopedAddresses(
-					chainClient, tx, ns, recoveryBatch,
-					recoveryMgr.State(), scopedMgrs,
-				)
+
+				return nil
 			})
 			if err != nil {
 				return err


JoeGruffins:
If there was an error connecting to a client inside an update in recovery, the tx is rolled back but the in memory sync point is not. this fixes that https://github.com/btcsuite/btcwallet/compare/40d5f25cd330609e350818655632b18fb1a360c8..6935a129d7e68f106895c7dd0c761d8e09b1fdd9

JoeGruffins:
The birthday thing was apparently a coincidence.

dev-warrior777 (approved):
Nice one

wallet/wallet.go
```diff
@@ -802,6 +803,7 @@ func (w *Wallet) recovery(chainClient chain.Interface,
 				)
 			})
 			if err != nil {
+				w.Manager.SetSyncedToMemory(&rollbackStamp)
```
guggero, line 806: Hmm, so what was the initial error in the first place that got this out of sync? Sounds like all the `failed to fetch block hash for height 870852: block not found` errors are follow-up errors.
Would be nice to know that. But this fix definitely makes sense.

JoeGruffins: I have the causing error in the issue for one instance.

guggero: Ah, I see, fetching a block failed in a Neutrino setup.

Great catch, I'm quite certain this is indeed the cause for the errors you saw.

I wonder how we should deal with all the other cases where `SetSyncedTo()` is called within a database transaction that might be rolled back and cause the DB and in-memory state to de-synced...

yyforyongyu: I think the out-of-synced state will make things worse. It looks like `recoverScopedAddresses` should be performed first before calling `SetSyncedTo`. Also it needs some refactor to move the non-db operations out so we don't need to wrap them in one giant tx `walletdb.Update`.

yyforyongyu (changes requested):
I don't think this approach works - it's hacky and creates inconsistent states.

JoeGruffins:
> I think the out-of-synced state will make things worse.

I'll look again but I think all of the db functions never happen, and this was the only thing not rolled back. I'll try do whatever though because this bug is affecting a few of our projects.

JoeGruffins:
@yyforyongyu I've looked through, and there is alot, but I can't find any values that are not rolled back. Could you point them out?

Also, with no fix, if the bug is hit, then the db would be borked anyway. Even if a huge refactor is needed, a fix for the moment would be good to have. I can try to write a test I guess?

JoeGruffins:
@yyforyongyu Is this current version ok for a patch? It only swaps the order as you suggested. Because the client request for headers happens in the first part, and this is where a connection error can happen, it won't continue to update the internal sync point. This also fixes the immediate problem for us. https://github.com/btcsuite/btcwallet/compare/469713a564568f734e51e8bd87e7a2ac2885b754..34a13fb13457f7c3a14dd97d0ecb97be8136ce3e

wallet/wallet.go
```diff
@@ -790,16 +790,25 @@ func (w *Wallet) recovery(chainClient chain.Interface,
 		if len(recoveryBatch) == recoveryBatchSize || height == bestHeight {
 			err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
 				ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+				err := w.recoverScopedAddresses(
+					chainClient, tx, ns, recoveryBatch,
+					recoveryMgr.State(), scopedMgrs,
+				)
+				if err != nil {
+					return err
+				}
```
JoeGruffins, line 799: Will combine this into one if statement.

JoeGruffins: done

JoeGruffins:
Combined the if statement with what was above https://github.com/btcsuite/btcwallet/compare/06463724d4c69dc3e09f2191de442f5635e9c606..7509f6b4b9e3aba479cd0f75b171f3607aaf4132

wallet/wallet.go
```diff
@@ -790,16 +790,24 @@ func (w *Wallet) recovery(chainClient chain.Interface,
 		if len(recoveryBatch) == recoveryBatchSize || height == bestHeight {
 			err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
 				ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+				if err := w.recoverScopedAddresses(
+					chainClient, tx, ns, recoveryBatch,
+					recoveryMgr.State(), scopedMgrs,
+				); err != nil {
+					return err
+				}
+				// TODO: Any error here will roll back this
```
yyforyongyu, line 799: nit: add a newline above

yyforyongyu (approved):
Thanks for the update! LGTM 🙏

wallet/wallet.go
JoeGruffins: above the TODO?

JoeGruffins:
New line above the TODO https://github.com/btcsuite/btcwallet/compare/7509f6b4b9e3aba479cd0f75b171f3607aaf4132..fe45f818f77fd014ff11971e8689cfd274a2df17

guggero:
I think this change should be safe, thanks a lot for the fix.
Would be great if you could open an `lnd` PR that references this PR in the `go.mod` (with a `replace` directive to point to your fork/commit) to see that all integration tests still pass.

JoeGruffins:
> Would be great if you could open an lnd PR that references this PR in the go.mod (with a replace directive to point to your fork/commit) to see that all integration tests still pass.

Attempting to do this but having a hell of a time with module tags. Erros like `server response: not found: github.com/joegruffins/btcwallet@v0.16.10-0.20241127094224-93c858b2ad63: invalid pseudo-version: preceding tag (v0.16.9) not found`

JoeGruffins:
go test is passing locally https://github.com/JoeGruffins/lnd/pull/1

guggero:
You just add `replace github.com/btcsuite/btcwallet => github.com/joegruffins/btcwallet fe45f818f77fd014ff11971e8689cfd274a2df17` into the `go.mod` file, then run `go mod tidy` for it to resolve the pseudo tag.

JoeGruffins:
> You just add `replace github.com/btcsuite/btcwallet => github.com/joegruffins/btcwallet fe45f818f77fd014ff11971e8689cfd274a2df17` into the `go.mod` file, then run `go mod tidy` for it to resolve the pseudo tag.

THX! that was easy.. its updated now. same tests are passing

yyforyongyu:
@JoeGruffins Thanks for putting up the tests! Could you open a PR in `lnd` so other devs can check the CI results there?

guggero:
@yyforyongyu I think we might need to roll this change back and approach it in a different way.
I just tried to recover a testnet (testnet4 to verify the changes in `btcd`) node and noticed that if there's any error during the wallet recovery, it will start over completely at block 1.
So on Neutrino, if there's a timeout fetching a block it will roll back and start from scratch, which can take forever...

PR #992, opened 2025-03-11
fix 404 link clientusage.md

futreall:
hi, i looked at the documentation and found a non-working link, so i decided to replace it, thanks


saubyk:
spam pr

saubyk:
> @saubyk can I tell you the term spam ?

No you may not. We consider any pr with simple typo fixes from private GitHub profiles as spam. thank you!

PR #993, opened 2025-03-11
Move Sequence func to the read interface

ziggie1984:
We do not need a write log to read the sequence number in bbolt so we change the interface and make it less strict.

### DIFF
--- a/walletdb/bdb/db.go
+++ b/walletdb/bdb/db.go
@@ -224,6 +224,11 @@ func (b *bucket) ReadCursor() walletdb.ReadCursor {
 	return b.ReadWriteCursor()
 }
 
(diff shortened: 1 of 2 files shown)


ziggie1984:
cc @guggero @bhandras

ziggie1984:
Needed for: https://github.com/lightninglabs/lndinit/pull/21

bhandras (approved):
LGTM 🎉

guggero (approved):
Makes sense. I actually don't see why this wasn't the case from the beginning. But perhaps there's some rationale behind it? Do you remember, @bhandras?

guggero:
Discussed offline. No blocking reason, this will just cause us to need to update the `kvdb` code in `lnd`, move the corresponding method over there too.

PR #995, opened 2025-03-18
Fix inaccurate `listunspent` result

yyforyongyu:
This PR fixes a long-standing issue found in `lnd` where the `listunspent` gives inaccurate results about the wallet utxos. The fix is achieved through four steps,
1. add more logs to help us find the bug. 
2. move the wallet initialization methods from `lnd` to `btcwallet` so the namespace of the `walletdb` is not exported.
3. update the tx filters synchronously and make sure they are updated before processing the blocks.
4. shorten the db's write tx, which has an impact on postgres.


saubyk:
cc: @Roasbeef for review

wallet/wallet.go
```diff
@@ -1622,6 +1622,38 @@ func (w *Wallet) AccountAddresses(account uint32) (addrs []btcutil.Address, err
 	return
 }
 
+// AccountManagedAddresses returns the managed addresses for every created
+// address for an account.
+func (w *Wallet) AccountManagedAddresses(scope waddrmgr.KeyScope,
```
Roasbeef, line 1627: Solid change! It just so happens that lnd is using the exact same `walletdb` abstraction for its database here. In the future, this won't be the case, so this is also a prep for when we move away from bolt in this package.

wallet/wallet.go
```diff
@@ -3926,6 +3958,220 @@ func (w *Wallet) Database() walletdb.DB {
...
+// from 1 to 255.
+func (w *Wallet) InitAccounts(scope *waddrmgr.ScopedKeyManager,
+	watchOnly bool) error {
+
+	return walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+		addrmgrNs := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+
+		// Generate all accounts that we could ever need. This includes
+		// all key families.
+		for keyFam := uint32(1); keyFam <= 255; keyFam++ {
```
Roasbeef, line 4035: Perhaps just use the term account here since key family is an lnd term? May also want to consider accepting the upper limit of this loop as an arg.

chain/bitcoind_client.go
```diff
@@ -52,14 +52,12 @@ type BitcoindClient struct {
...
-	// rescanUpdate is a channel will be sent items that we should match
-	// transactions against while processing a chain rescan to determine if
-	// they are relevant to the client.
-	rescanUpdate chan interface{}
-
 	// watchedAddresses, watchedOutPoints, and watchedTxs are the set of
 	// items we should match transactions against while processing a chain
 	// rescan to determine if they are relevant to the client.
+	//
+	// TODO(yy): make sure to GC these maps once the spending tx has
```
Roasbeef, line 59: 👍 

Let's just roll this out into an issue now?

Roasbeef: I actually don't think we can ever remote entries in this map. We essentially keep an in memory map of every addr/utxo we might ever want to notify off of. If someone ends to an addr we used last 1 year ago, we still want to generate a notification. 

So one way forward here would be to spin this out into an interface, so we have more flexibility here w.r.t caching and/or implementation.

wallet/notifications.go
```diff
@@ -196,20 +196,38 @@ func relevantAccounts(_ *Wallet, m map[uint32]btcutil.Amount, txs []TransactionS
...
-	defer s.mu.Unlock()
 	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	// Exit early if there are no clients.
 	clients := s.transactions
 	if len(clients) == 0 {
 		return
 	}
+	// It's possible that the transaction was not found within the wallet's
```
Roasbeef, line 210: Style nit: missing new line above.

Roasbeef (approved):
Great set of changes! A good example of an area where removing unnecessary concurrency can help us to better reason about code paths and eliminate potential consistency issues. 

We could've also made the hand off syncronous as well (ack channel), but this is a bit simpler.

wallet/notifications.go
yyforyongyu: fixed!

wallet/wallet.go
yyforyongyu: Updated.

saubyk:
cc: @ziggie1984 for review

chain/bitcoind_client.go
Roasbeef: > This got me to think about the restart behavior, I think watchedAddresses should be updated during restart?

During restart, we'll fetch all the relevant addr/UTXOs using `activeData`: https://github.com/btcsuite/btcwallet/blob/b2cb575cc6ccc0c41cf24ea834df580414b7acac/wallet/wallet.go#L557-L572

chain/bitcoind_client.go
```diff
@@ -1450,3 +1413,58 @@ func (c *BitcoindClient) LookupInputMempoolSpend(op wire.OutPoint) (
...
+			c.watchedAddresses[addr.String()] = struct{}{}
+		}
+
+	// We're adding the outpoints to our filter.
+	case []wire.OutPoint:
+		for _, op := range update {
+			c.watchedOutPoints[op] = struct{}{}
+		}
+
+	case []*wire.OutPoint:
```
ziggie1984, line 1365: seems like a design flaw to have `*wire.OutPoint` and `wire.OutPoint`

chain/bitcoind_client.go
ziggie1984: Q: An active address is bascially every address in our addr db ?

waddrmgr/address.go
```diff
@@ -82,6 +82,30 @@ const (
...
 
+// String returns a human-readbale format of the address type.
+func (a AddressType) String() string {
+	switch a {
+	case PubKeyHash:
+		return "PubKeyHash"
+	case Script:
+		return "Script"
+	case RawPubKey:
+		return "RawPubKey"
```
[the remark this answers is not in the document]
yyforyongyu: I think it can be both - it depends on how you serialize it later.

waddrmgr/manager.go
```diff
@@ -423,6 +423,8 @@ func (m *Manager) IsWatchOnlyAccount(ns walletdb.ReadBucket, keyScope KeyScope,
 // with the address manager.
 //
 // This function MUST be called with the manager lock held for writes.
+//
+// TODO(yy): Rename this to wipe memory for priv keys.
```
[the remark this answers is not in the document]
yyforyongyu: yeah the function name should be renamed if you just look at what it does.

chain/bitcoind_client.go
```diff
@@ -1359,14 +1360,18 @@ func (c *BitcoindClient) filterTx(txDetails *btcutil.Tx,
...
 		// Otherwise, we'll check whether it matches a pkScript in our
 		// watch list encoded as an address. To do so, we'll re-derive
 		// the pkScript of the output the input is attempting to spend.
-		pkScript, err := txscript.ComputePkScript(
-			txIn.SignatureScript, txIn.Witness,
-		)
+		pkScript, err := txscript.ComputePkScript(sig, witness)
 		if err != nil {
 			// Non-standard outputs can be safely skipped.
+			log.Warnf("Received non-standard input sig=%x, "+
```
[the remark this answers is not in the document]
yyforyongyu: I think once we fail to parse the sig or witness we consider it non-standard?

chain/bitcoind_client.go
yyforyongyu: yeah, the code is pretty old, which will be fixed in an upcoming refactor.

wallet/wallet.go
```diff
@@ -3926,6 +3958,220 @@ func (w *Wallet) Database() walletdb.DB {
 	return w.db
 }
 
+// RemoveDescendants attempts to remove any transaction from the wallet's tx
+// store (that may be unconfirmed) that spends outputs created by the passed
```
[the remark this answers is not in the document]
yyforyongyu: This is a pure refactor to move a method from `lnd` to `btcwallet`.

wallet/wallet.go
```diff
@@ -3926,6 +3958,220 @@ func (w *Wallet) Database() walletdb.DB {
...
+		wtxmgrNs := tx.ReadWriteBucket(wtxmgrNamespaceKey)
+
+		return w.TxStore.RemoveUnminedTx(wtxmgrNs, txRecord)
+	})
+}
+
+// BirthdayBlock returns the birthday block of the wallet.
+//
+// NOTE: The wallet won't start until the backend is synced, thus the birthday
+// block won't be set and `ErrBirthdayBlockNotSet` will be returned.
```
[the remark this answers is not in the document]
yyforyongyu: This is a pure refactor to move a method from `lnd` to `btcwallet` - the comment may be updated, will look into improvement once I start the SQL PR.

wallet/wallet.go
```diff
@@ -3926,6 +3958,220 @@ func (w *Wallet) Database() walletdb.DB {
...
+func (w *Wallet) AddScopeManager(scope waddrmgr.KeyScope,
+	addrSchema waddrmgr.ScopeAddrSchema) (
+	*waddrmgr.ScopedKeyManager, error) {
+
+	var scopedManager *waddrmgr.ScopedKeyManager
+
+	err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
+		addrmgrNs := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+
+		manager, err := w.Manager.NewScopedKeyManager(
```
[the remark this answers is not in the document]
yyforyongyu: yep

chain/bitcoind_client.go
yyforyongyu: > Q: An active address is bascially every address in our addr db ?

I think so.

PR #997, opened 2025-04-06
walletdb: add readonly mode to bbolt db implementation

ziggie1984:
Allow opening the bbolt database in read-only mode, so we can safe time when opening the db because we do not need to sync the free-list which takes a lot of time.


guggero (approved):
LGTM :tada: 

Note to mostly myself: We should remember pushing a tag that reflects the breaking change nature of this PR.

bhandras (approved):
LGTM 🎉

PR #1003, opened 2025-04-09
wtxmgr: add method `OutputsToWatch`

yyforyongyu:
This commit adds a new method `OutputsToWatch`, which returns a list of outpoints the rescan process should be watching for during the startup.


Roasbeef (approved):
LGTM 🫛

PR #1011, opened 2025-05-23
wallet+wtxmgr: check credit in `FetchOutpointInfo`

yyforyongyu:
To decide whether an outpoint belongs to the wallet or not, we need to further check that the specified output index is indeed a credit paid to us.


wallet/utxos_test.go
```diff
@@ -129,6 +130,84 @@ func TestFetchOutpointInfo(t *testing.T) {
...
+
+	// Create an address we can use to send some coins to.
+	addr, err := w.CurrentAddress(0, waddrmgr.KeyScopeBIP0084)
+	require.NoError(t, err)
+	p2shAddr, err := txscript.PayToAddrScript(addr)
+	require.NoError(t, err)
+
+	// Create a tx that has two outputs - output1 belongs to the wallet,
+	// output2 is external.
+	output1 := wire.NewTxOut(100000, p2shAddr)
```
ziggie1984, line 149: Hmm why do they they pay to the same addr thought therefore we created addr above ?

ziggie1984: doesn't matter but still better to have a proper logic in.

wallet/utxos_test.go
```diff
@@ -129,6 +130,84 @@ func TestFetchOutpointInfo(t *testing.T) {
...
+				Hash:  tx.TxHash(),
+				Index: 1000,
+			},
+			errExpected: "invalid output index",
+		},
+		{
+			name: "no credit found",
+			prevOut: &wire.OutPoint{
+				Hash:  tx.TxHash(),
+				Index: 1000,
```
ziggie1984, line 190: wrong index probably you wanted to use 1 here ?

wallet/utxos_test.go
```diff
@@ -129,6 +130,84 @@ func TestFetchOutpointInfo(t *testing.T) {
...
+			},
+			errExpected: "invalid output index",
```
ziggie1984, line 192: wrong error ?

ziggie1984: missing the success case as well ?

wallet/utxos_test.go
yyforyongyu: they are dummy values - techinically they should be the same so we know the result doesn't depend on the pkscript, but again I don't think it matters.

wallet/utxos_test.go
yyforyongyu: see `TestFetchOutpointInfo`

wallet/utxos.go
```diff
@@ -224,3 +231,19 @@ func (w *Wallet) FetchDerivationInfo(pkScript []byte) (*psbt.Bip32Derivation,
 
 	return derivation, nil
 }
+
+// hasOutpoint takes an output identified by its output index and determines
+// whether the TxDetails contains this output. If the TxDetails doesn't have
+// this output, it means this output doesn't belong to our wallet.
+//
+// TODO(yy): implement this method on `TxDetails` and update the package
```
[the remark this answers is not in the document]
yyforyongyu: feels a bit wasteful to create a new version based on this simple addition, gonna keep it here unless you have good reasons.

ziggie1984 (approved):
LGTM

guggero (approved):
LGTM, thanks for the fix!

PR #1012, opened 2025-06-03
fix(chain): add missing spaces on ZMQ connection logs.

luisschwab:
## Change Description

Add missing spaces on ZMQ connection log.

Currently: 
```
2025-06-03 14:13:38.555 [INF] BTWL: Started listening for bitcoind block notifications via ZMQ on10.99.99.99:28332
2025-06-03 14:13:38.555 [INF] BTWL: Started listening for bitcoind transaction notifications via ZMQ on10.99.99.99:28333
```

### Testing

### DIFF
--- a/chain/bitcoind_zmq_events.go
+++ b/chain/bitcoind_zmq_events.go
@@ -243,8 +243,8 @@ func (b *bitcoindZMQEvents) LookupInputSpend(
 func (b *bitcoindZMQEvents) blockEventHandler() {
 	defer b.wg.Done()
 
-	log.Info("Started listening for bitcoind block notifications via ZMQ "+
-		"on", b.blockConn.RemoteAddr())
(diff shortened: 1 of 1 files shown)


chain/bitcoind_zmq_events.go
```diff
@@ -244,7 +244,7 @@ func (b *bitcoindZMQEvents) blockEventHandler() {
 	defer b.wg.Done()
 
 	log.Info("Started listening for bitcoind block notifications via ZMQ "+
-		"on", b.blockConn.RemoteAddr())
+		"on ", b.blockConn.RemoteAddr())
```
guggero, line 247: Let's change it to an `Infof` with a placeholder instead? Then it's more apparent that there's a space missing.

luisschwab: Agree, done.

PR #1022, opened 2025-08-01
waddrmgr: make sure to create `imported` account for watchonly

yyforyongyu:
The `NewScopedKeyManager` will skip creating the `imported` account when the watch is watch-only. We now fix it by always creating it even if it's watch-only.

### DIFF
--- a/waddrmgr/manager.go
+++ b/waddrmgr/manager.go
@@ -590,6 +590,14 @@ func (m *Manager) NewScopedKeyManager(ns walletdb.ReadWriteBucket,
 		if err != nil {
 			return nil, err
 		}
+	} else {
+		err := putDefaultAccountInfo(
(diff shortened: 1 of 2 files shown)


guggero:
Oh, it's as easy a fix as that? Thanks a lot for looking into it, will review properly next week!

yyforyongyu:
yeah somehow the `imported` account was not created, tho based on the test here I think it's meant to be there,
https://github.com/btcsuite/btcwallet/blob/038b57795c1f66f99c4ef2d9bd5fc203aea5ae4b/wallet/import_test.go#L210-L220

guggero (approved):
tACK, LGTM 🎉 

Thanks a lot for finding this!

Roasbeef (approved):
LGTM 🥮

PR #1026, opened 2025-08-06
wallet: add interface for `Wallet`

yyforyongyu:
In this commit, a new interface is introduced so when using the package as a library, the caller can use `wallet.Interface` to access the methods provided by `Wallet`.


Roasbeef (approved):
Coulda sworn we already had this, but the last big step hwere was the creation of the `wallet` package in the first place. 

LGTM 🧶

wallet/interface.go
```diff
@@ -0,0 +1,323 @@
...
+// Interface defines the public API for a wallet.
+//
+// This interface is intended to be the primary way that other packages and
+// applications interact with a wallet. It defines a stable, high-level
+// contract that abstracts away the underlying implementation details, making
+// it easier to build applications on top of the wallet and to create mock
+// implementations for testing.
+//
+//nolint:interfacebloat
+type Interface interface {
```
GustavoStingelin, line 31: Is it a good idea to call this `Interface`? I have a list of probably better names:
* Wallet
* IWallet
* BitcoinWallet
* WalletProvider

wallet/interface.go
```diff
@@ -0,0 +1,323 @@
...
+	// expiration time.
+	LockOutpoint(op wire.OutPoint)
+
+	// UnlockOutpoint unlocks a previously locked UTXO, making it available
+	// for coin selection again.
+	UnlockOutpoint(op wire.OutPoint)
+
+	// LockedOutpoint returns whether an outpoint has been marked as locked
+	// and should not be used as an input for created transactions.
+	LockedOutpoint(op wire.OutPoint) bool
```
GustavoStingelin, line 208: I assume these operations interact with the database, right? If so, they can fail and should probably return an error

GustavoStingelin (changes requested):

wallet/interface.go
Roasbeef: I think was an attempt to [avoid](https://go.dev/doc/effective_go#package-names) `wallet.Wallet`? Of the list, I think `wallet.WalletProvider` isn't too bad.

wallet/interface.go
GustavoStingelin: ok

GustavoStingelin (approved):

wallet/interface.go
yyforyongyu: This is done to mimic the `chain.Interface` here.

> In the future, we could benefit from splitting it into smaller interfaces like wallet.Locker, Unlocker, Balancer, Spender, Addresser, Broadcaster, etc.

Def this is just the first step, we def will fix `interfacebloat` and make the interface smaller!

yyforyongyu:
> Are those all commonly used methods mentioned in task 3 regards benchmark tests

Yeah I think so, tho maybe we could categorize them even further.

yyforyongyu:
Will hold off merging until the CI passes in https://github.com/lightningnetwork/lnd/pull/10136

PR #1029, opened 2025-08-12
Add design docs for reference

yyforyongyu:
This PR adds architecture design docs so other tasks can use them as references.

### DIFF
--- a/chain/ARCHITECTURE.md
+++ b/chain/ARCHITECTURE.md
@@ -1,0 +1,106 @@
+# Chain Package Architecture
+
+This document outlines the architecture for the `chain` package, which serves as the foundational layer for interacting with the Bitcoin blockchain.
+
+## Core Principle: A Two-Tiered Architecture
+
+The `btcwallet`'s interaction with the blockchain is defined by a two-tiered architecture that creates a clean separation of concerns:
+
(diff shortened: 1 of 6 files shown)


chain/ARCHITECTURE.md
```diff
@@ -0,0 +1,106 @@
...
+	OnRelevantTx(tx *btcutil.Tx, details *btcjson.BlockDetails)
+}
+```
+
+### `MempoolObserver` (Optional) Interface
+*For full-node-specific mempool functionality.*
+```go
+// MempoolObserver provides methods for observing the mempool of a full node.
+// This is an optional interface that a ChainDriver may implement.
+type MempoolObserver interface {
```
GustavoStingelin, line 89: The observer name here reminds me of the [Observer Pattern](https://refactoring.guru/design-patterns/observer), but the purpose seems to be a bit different, right?

GustavoStingelin (approved):
just one comment, but LGTM!

chain/ARCHITECTURE.md
yyforyongyu: correct - this is more like providing a mempool IO, tho the whole `ntft` uses the [publish-subscribe pattern](https://en.wikipedia.org/wiki/Publish%E2%80%93subscribe_pattern), which I believe is the same as the observer pattern you linked.

yyforyongyu:
Turning this into a draft as this PR serves as a reference point atm - we will update the design docs once the whole series is done, as I'm working on the wallet interfaces, I realized there are a few places that need to be fine-tuned.

PR #1058, opened 2025-09-09
chain: update `Start` method to take `context`

yyforyongyu:
This PR https://github.com/lightninglabs/neutrino/pull/317 changed the signature of `Start`, so we now need to pass a context to it.

### DIFF
--- a/chain/chainservice.go
+++ b/chain/chainservice.go
@@ -1,6 +1,8 @@
 package chain
 
 import (
+	"context"
+
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/btcutil/gcs"
 	"github.com/btcsuite/btcd/chaincfg"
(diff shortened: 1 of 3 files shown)


chain/chainservice.go
```diff
@@ -14,7 +16,7 @@ import (
 // NeutrinoChainService is an interface that encapsulates all the public
 // methods of a *neutrino.ChainService
 type NeutrinoChainService interface {
-	Start() error
+	Start(context.Context) error
```
GustavoStingelin, line 19: It looks like it is causing a compilation error in the pipeline and in my local env

```
➜  btcwallet git:(fix-neutrino) make
 Compiling btcwallet.
GO111MODULE=on go build -v github.com/btcsuite/btcwallet/...
github.com/btcsuite/btcwallet/cmd/sweepaccount
github.com/btcsuite/btcwallet/chain
# github.com/btcsuite/btcwallet/chain
chain/chainservice.go:42:30: cannot use (*neutrino.ChainService)(nil) (value of type *neutrino.ChainService) as NeutrinoChainService value in variable declaration: *neutrino.ChainService does not implement NeutrinoChainService (wrong type for method Start)
                have Start() error
                want Start(context.Context) error
chain/neutrino.go:93:16: cannot use chainService (variable of type *neutrino.ChainService) as NeutrinoChainService value in struct literal: *neutrino.ChainService does not implement NeutrinoChainService (wrong type for method Start)
                have Start() error
                want Start(context.Context) error
make: *** [Makefile:48: build] Error 1
```

Roasbeef:
New neutrino tag pushed: https://github.com/lightninglabs/neutrino/tags

PR #1061, opened 2025-09-15
Separate tools from main project

ziggie1984:
separate tools from the main project to avoid dependency conflicts in projects which importing this library but also in general its way more convenient to keep tool dependencies and  lib dependencies separate.


yyforyongyu (changes requested):
Agree we should keep them seperate, however the current fix will break the `go.work` as it's specifying a different go mod path.
```
> make lint
 Linting source.
go tool --modfile ./tools/go.mod github.com/golangci/golangci-lint/v2/cmd/golangci-lint run -v 
go: -modfile cannot be used in workspace mode
make: *** [lint] Error 1
```

It looks like this cannot be achieved using `go tool` so I think we should just go back to the `build and cache` strategy, sth like we've had before.

ziggie1984:
Golang has a way how to make it work with: 

`GOWORK=off go tool --modfile=$(TOOLS_MODFILE) $(GOIMPORTS_PKG) -w $(GOFILES_NOVENDOR) `

Issue #1036, opened 2025-08-12
Interface Design for `TxReader`


yyforyongyu:
The ["Transaction History"](https://gist.github.com/yyforyongyu/c0a6ec78cedcbf5fb342be412dc591ea) group, which contains methods used for querying and managing the wallet's transaction history. We now introduce `TxReader` interface, which is responsible for all "read" operations related to transactions: querying history and subscribing to new events.

### Rationale for Changes

1.  **Clear Naming (`TxHistoryManager` -> `TxReader`)**: The name `TxReader` is more explicit and consistent with the CQRS pattern, clearly indicating that this interface performs read-only operations.

2.  **Remove Internal Methods**: `RemoveDescendants` is a low-level utility for managing transaction store integrity (e.g., during re-orgs). This should be handled internally by the wallet and not exposed in the high-level public API.

3.  **Standardize Naming**: Method names are standardized to use the `Tx` prefix for consistency (e.g., `GetTransactionDetails` becomes `GetTxDetails`).

4.  **Re-categorization**: `SubscribeTransactions` (renamed to `SubscribeTxes`) is moved here from the `WalletController` interface, as it is fundamentally a read/query operation.

### Final `TxReader` Interface Design

This results in a clean, focused, and safe read-only interface for transaction history.

```go
// TxReader provides an interface for querying transaction history.
type TxReader interface {
    // GetTxDetails returns a detailed description of a transaction
    // given its transaction hash.
    GetTxDetails(ctx context.Context, txHash *chainhash.Hash) (*TransactionDetail, error)

    // ListTxDetails returns a list of all transactions which are
    // relevant to the wallet over a given block range.
    ListTxDetails(ctx context.Context, startHeight, endHeight int32, accountFilter string) ([]*TransactionDetail, error)

    // LabelTx adds a label to a transaction.
    LabelTx(ctx context.Context, hash chainhash.Hash, label string, overwrite bool) error

    // FetchTx attempts to fetch a transaction in the wallet's database.
    FetchTx(ctx context.Context, txHash *chainhash.Hash) (*wire.MsgTx, error)

    // TODO: This is a legacy method that should be removed in favor
    // of a more robust notification system.
    // SubscribeTxes returns a TransactionSubscription client which
    // is capable of receiving async notifications as new transactions
    // related to the wallet are seen within the network, or found in
    // blocks.
    SubscribeTxes() (TransactionSubscription, error)
}
```

PR #1091, opened 2025-09-29
 wallet: add `TxReader` and `TxWriter` interfaces

yyforyongyu:
This PR fixes https://github.com/btcsuite/btcwallet/issues/1036 - the old design has evolved here, detailed below.

# Rationale for the `TxReader` and `TxWriter` API Refinement

This document outlines the design decisions that led to the separation of transaction history querying and transaction modification into two distinct, focused interfaces: `TxReader` and `TxWriter`.

## Initial State: A Monolithic and Inefficient API

Previously, methods for querying transaction history were part of the main `wallet.Wallet` struct. This approach had several significant drawbacks:

1.  **No Separation of Concerns**: The `Wallet` struct was a monolithic entity responsible for everything from signing and broadcasting to address management and history querying. There was no clear boundary between read and write operations, violating the Command Query Responsibility Segregation (CQRS) principle.
2.  **Inefficiency and the N+1 Query Problem**: Older methods like `GetTransactions` were highly inefficient. They would fetch a list of transaction details and then, for each transaction in the list, call a helper function (`makeTxSummary`) that would perform *additional* database lookups to gather account and address information. This resulted in a large number of database queries, creating a performance bottleneck.
3.  **Inconsistent and Low-Level Return Types**: The methods returned raw or minimally processed internal structures (like `wtxmgr.TxDetails` or a basic `TransactionSummary`). This forced the client to perform additional processing and understand the wallet's internal data models.
4.  **Poor Ergonomics**: The API was not designed with the client's needs in mind, leading to cumbersome filtering and the need for boilerplate code in client applications to get the required information.

## The Design Evolution: Separating Reads from Writes

The key insight was to apply the CQRS principle strictly, leading to the creation of two new, more focused interfaces.

### Insight 1: A Pure `TxReader` for Efficient, High-Level Querying

The first step was to create a dedicated `TxReader` interface responsible solely for querying transaction history.

-   **From `makeTxSummary` to `TxDetail`**: The new `GetTx` and `ListTxns` methods return a `wallet.TxDetail` struct. This is a high-level, information-dense Data Transfer Object (DTO) designed specifically for client consumption. It combines all necessary information (confirmations, net value, fee, etc.) into a single object, calculated from a single, efficient database fetch.
-   **Solving the N+1 Problem**: The implementation of `ListTxns` uses `txStore.RangeTransactions` to efficiently stream data from the database and then transforms it in memory using `buildTxDetail`. This completely eliminates the N+1 query problem, as no additional database lookups are performed inside the loop.
-   **Focused Responsibility**: The `TxReader` interface is strictly read-only, making it a safe dependency for any component that only needs to display or analyze transaction history.

### Insight 2: A Dedicated `TxWriter` for Modifying Transactions

During the design process, it became clear that operations that *modify* transaction metadata, such as `LabelTx`, did not belong in a "Reader" interface. This led to the creation of the `TxWriter`.

-   **Strict CQRS**: By moving `LabelTx` into its own `TxWriter` interface, we achieve a perfect separation of reads and writes. This makes the architecture cleaner and the API safer. A client that holds a `TxReader` cannot accidentally modify a transaction's label.
-   **Clear Intent**: The existence of two separate interfaces makes the purpose of each method and the overall design of the system much clearer to developers.

## The Final API: Two Focused, Composable Interfaces

The final design results in two clean, composable interfaces that adhere to the Single Responsibility Principle and CQRS.

### `TxReader`

```go
// TxReader provides an interface for querying tx history.
type TxReader interface {
	// GetTx returns a detailed description of a tx given its tx hash.
	GetTx(ctx context..Context, txHash chainhash.Hash) (
		*TxDetail, error)

	// ListTxns returns a list of all txns which are relevant to the wallet
	// over a given block range.
	ListTxns(ctx context.Context, startHeight, endHeight int32) (
		[]*TxDetail, error)
}
```

### `TxWriter`

```go
// TxWriter provides an interface for updating wallet txns.
type TxWriter interface {
	// LabelTx adds a label to a tx. If a label already exists, it will be
	// overwritten.
	LabelTx(ctx context.Context, hash chainhash.Hash, label string) error
}
```

### Why This Design is Superior:

1.  **Architecturally Sound**: It correctly implements the CQRS pattern by strictly separating read (`TxReader`) and write (`TxWriter`) operations.
2.  **Highly Efficient**: It eliminates the N+1 query problem of the previous implementation, leading to significantly better performance when listing transactions.
3.  **Ergonomic and Client-Focused**: The `TxDetail` DTO provides a rich, high-level, and consistent data structure that is easy for client applications to consume without needing to understand the wallet's internal models.
4.  **Safe and Explicit**: The separation of interfaces makes the API safer and the intent of the code clearer. It prevents misuse and makes the system easier to reason about.

### DIFF
--- a/waddrmgr/manager.go
+++ b/waddrmgr/manager.go
@@ -431,17 +431,16 @@ func (m *Manager) IsWatchOnlyAccount(ns walletdb.ReadBucket, keyScope KeyScope,
 func (m *Manager) lock() {
 	for _, manager := range m.scopedManagers {
 		// Clear all of the account private keys.
-		for _, acctInfo := range manager.acctInfo {
+		for _, acctInfo := range manager.accountInfo() {
 			if acctInfo.acctKeyPriv != nil {
 				acctInfo.acctKeyPriv.Zero()
 			}
 			acctInfo.acctKeyPriv = nil
 		}
-	}
 
-	// Remove clear text private keys and scripts from all address entries.
-	for _, manager := range m.scopedManagers {
-		for _, ma := range manager.addrs {
+		// Remove clear text private keys and scripts from all address
+		// entries.
+		for _, ma := range manager.addresses() {
 			switch addr := ma.(type) {
 			case *managedAddress:
 				addr.lock()
--- a/waddrmgr/scoped_manager.go
+++ b/waddrmgr/scoped_manager.go
@@ -4,6 +4,7 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"fmt"
+	"maps"
 	"sync"
 
 	"github.com/btcsuite/btcd/btcec/v2"
@@ -2653,4 +2654,28 @@ func (s *ScopedKeyManager) NewAddress(addrmgrNs walletdb.ReadWriteBucket,
 	addr := addrs[0].Address()
 
 	return addr, nil
+}
+
+// accountInfo returns a copy of the account info map.
+func (s *ScopedKeyManager) accountInfo() map[uint32]*accountInfo {
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+
+	acctInfoCopy := make(map[uint32]*accountInfo, len(s.acctInfo))
+	maps.Copy(acctInfoCopy, s.acctInfo)
+
+	return acctInfoCopy
+}
+
+// addresses returns a slice of all managed addresses.
+func (s *ScopedKeyManager) addresses() []ManagedAddress {
+	s.mtx.RLock()
+	defer s.mtx.RUnlock()
+
+	addrs := make([]ManagedAddress, 0, len(s.addrs))
+	for _, ma := range s.addrs {
+		addrs = append(addrs, ma)
+	}
+
+	return addrs
 }


wallet/tx_reader.go
```diff
@@ -0,0 +1,328 @@
...
+type TxReader interface {
+	// GetTx returns a detailed description of a tx given its tx hash.
+	GetTx(ctx context.Context, txHash chainhash.Hash) (
+		*TxDetail, error)
+
+	// ListTxns returns a list of all txns which are relevant to the wallet
+	// over a given block range.
+	ListTxns(ctx context.Context, startHeight, endHeight int32) (
+		[]*TxDetail, error)
+}
```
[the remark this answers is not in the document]
yyforyongyu: I think if it's used like `wallet.ListTxns` it should be fine, guess we will see later on, maybe add sth like `Wallet` in the name.

unit/rates.go
```diff
@@ -0,0 +1,289 @@
...
+
+// LessThanOrEqual returns true if the fee rate is less than or equal to the
+// other fee rate.
+func (s SatPerKWeight) LessThanOrEqual(other SatPerKWeight) bool {
+	return s.Cmp(other.Rat) <= 0
+}
+
+// round rounds a big.Rat to the nearest btcutil.Amount (int64).
+
+func round(r *big.Rat) btcutil.Amount {
```
GustavoStingelin:
```suggestion
// roundToAmount rounds a big.Rat to the nearest btcutil.Amount (int64).
func roundToAmount(r *big.Rat) btcutil.Amount {
```

Not a blocker, but I think renaming the function to `roundToAmount` makes it clearer when reading the code, since it specifies the target type. Also, there’s a blank line below the comment.

unit/rates.go
```diff
@@ -0,0 +1,289 @@
+// Copyright (c) 2025 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+// Package unit provides a set of types for dealing with bitcoin units.
+package unit
```
GustavoStingelin: Maybe we could place this unit package inside `pkg`? It might help if we ever want to make the module more generic or move it to `btcutil` later. But I’m also fine keeping it here.

GustavoStingelin (commented):
Really nice use of `big.Rat`, I like it. I also did some quick checks mentally and with a calculator, and the results look correct. I just left a few additional comments.

unit/rates.go
yyforyongyu: done

unit/rates.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// safeUint64ToInt64 converts a uint64 to an int64, capping at math.MaxInt64.
+// This is used to silence gosec warnings about integer overflows. In practice,
+// the values being converted are transaction weights or sizes, which are
+// limited by consensus rules and are not expected to overflow an int64.
+func safeUint64ToInt64(u uint64) int64 {
+	if u > math.MaxInt64 {
+		return math.MaxInt64
+	}
```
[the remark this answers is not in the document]
yyforyongyu: nice updated!

wallet/tx_reader.go
```diff
@@ -0,0 +1,407 @@
...
+
+// TxReader provides an interface for querying tx history.
+type TxReader interface {
+	// GetTx returns a detailed description of a tx given its tx hash.
+	GetTx(ctx context.Context, txHash chainhash.Hash) (
+		*TxDetail, error)
+
+	// ListTxns returns a list of all txns which are relevant to the wallet
+	// over a given block range.
+	ListTxns(ctx context.Context, startHeight, endHeight int32) (
```
mohamedawnallah: **nitpick (feel free to ignore):** `ListTxs` instead of `ListTxns` (without `n` in the plural form). This mirrors the existing API `UnminedTxs` naming in the codebase

wallet/tx_reader.go
```diff
@@ -0,0 +1,407 @@
...
+
+	for _, credit := range txDetails.Credits {
+		balanceDelta += credit.Amount
+	}
+
+	details.Value = balanceDelta
+
+	if len(txDetails.Debits) != len(txDetails.MsgTx.TxIn) {
+		return
+	}
```
mohamedawnallah: **nitpick:** It can be helpful adding comment above this on why we return here. Something along those lines:
```
// If not all inputs are ours, we can't calculate the total fee.
// txDetails.Debits contains only our inputs, while txDetails.MsgTx.TxIn
// contains all inputs. If they differ, some inputs belong to external
// wallets and we don't know their input values.
```

wallet/tx_reader.go
```diff
@@ -0,0 +1,407 @@
...
+// buildTxDetail builds a TxDetail from the given wtxmgr.TxDetails.
+func (w *Wallet) buildTxDetail(txDetails *wtxmgr.TxDetails,
+	currentHeight int32) *TxDetail {
+
+	details := w.buildBasicTxDetail(txDetails)
+
+	w.populateBlockDetails(details, txDetails, currentHeight)
+	w.calculateValueAndFee(details, txDetails)
+	w.populateOutputs(details, txDetails)
+	w.populatePrevOuts(details, txDetails)
```
mohamedawnallah: **thought:** Currently there are redundant iterations over the same data:

- `txDetails.Debits` is iterated **3 times**: twice in `calculateValueAndFee` and once in `populatePrevOuts`
- `txDetails.Credits` is iterated **2 times**: once in `calculateValueAndFee` and once in `populateOutputs`
- `txDetails.MsgTx.TxOut` is iterated **2 times**: once in `calculateValueAndFee` and once in `populateOutputs`

Thinking if there is a way to iterate each dataset only once while combining/maintaining the same refactored functionality somehow. That way we can save a bit computing iteration work here and there

wallet/tx_reader.go
```diff
@@ -0,0 +1,407 @@
...
+					detail := &d[i]
+
+					txDetail := w.buildTxDetail(
+						detail, currentHeight,
+					)
+					details = append(details, txDetail)
+				}
+
+				return false, nil
+			},
```
mohamedawnallah: **thought:**  It can be helpful if we put this anonymous callback function in a variable above. Something along those:
```golang
processTxBatch := func(batch []wtxmgr.TxDetails) (shouldBreak  bool, err error) {
    for i := range batch {
        detail := &batch[i]
        txDetail := w.buildTxDetail(detail, currentHeight)
        details = append(details, txDetail)
    }
    // Continue processing all transactions.
    return false, nil
}
```

Then we call it here:
```golang
err := w.txStore.RangeTransactions(txmgrNs, startHeight, endHeight, processTxBatch)
...
```

wallet/tx_reader.go
```diff
@@ -0,0 +1,407 @@
...
+
+// buildBasicTxDetail builds the basic TxDetail from the given wtxmgr.TxDetails.
+func (w *Wallet) buildBasicTxDetail(txDetails *wtxmgr.TxDetails) *TxDetail {
+	txWeight := blockchain.GetTransactionWeight(
+		btcutil.NewTx(&txDetails.MsgTx),
+	)
+
+	return &TxDetail{
+		Hash:         txDetails.Hash,
+		RawTx:        txDetails.SerializedTx,
```
mohamedawnallah: There is a bug here where `SerializedTx` is always be set to nil and never initialized/set, unlike the old deprecated `GetTransaction` which conditionally initializes `serializedTx` (https://github.com/btcsuite/btcwallet/blob/d4b5cc965ab15086f11cbeab1d0726d833825a1e/wallet/notifications.go#L100-L108). We need to handle this either by mirroring the conditional check in the new API, or potentially a better approach would be something along those lines:
```diff
diff --git a/wtxmgr/db.go b/wtxmgr/db.go
index c7e3619..112aa7c 100644
--- a/wtxmgr/db.go
+++ b/wtxmgr/db.go
@@ -428,7 +428,8 @@ func readRawTxRecord(txHash *chainhash.Hash, v []byte, rec *TxRecord) error {
 	}
 	rec.Hash = *txHash
 	rec.Received = time.Unix(int64(byteOrder.Uint64(v)), 0)
-	err := rec.MsgTx.Deserialize(bytes.NewReader(v[8:]))
+	rec.SerializedTx = v[8:]
+	err := rec.MsgTx.Deserialize(bytes.NewReader(rec.SerializedTx))
 	if err != nil {
 		str := fmt.Sprintf("%s: failed to deserialize transaction %v",
 			bucketTxRecords, txHash)
```

wallet/tx_reader.go
```diff
@@ -0,0 +1,407 @@
...
+
+// GetTx returns a detailed description of a tx given its tx hash.
+//
+// NOTE: This method is part of the TxReader interface.
+//
+// Time complexity: O(1) amortized. The lookup is dominated by a key-based
+// B-tree lookup in the database, which is effectively constant time for any
+// realistic number of transactions.
+func (w *Wallet) GetTx(_ context.Context, txHash chainhash.Hash) (
+	*TxDetail, error) {
```
mohamedawnallah: **question(blocking/non-blocking):** It sounds like `GetTx` API has no amortization in terms of time complexity? It is a read operation with consistent upper/tight bound cost every time. The time complexity here is asymptotically `O(log n + I + O)` not `O(1)` amortized

Where:
- `n` is the number of transactions in the database (B-tree lookup)
- `I` is the number of inputs in the transaction
- `O` is the number of outputs in the transaction

wallet/tx_reader.go
mohamedawnallah: There is a bug here where `SerializedTx` is always be set to nil and never initialized, unlike the old deprecated `GetTransaction` which conditionally initializes `serializedTx` (https://github.com/btcsuite/btcwallet/blob/d4b5cc965ab15086f11cbeab1d0726d833825a1e/wallet/notifications.go#L100-L108). We need to handle this either by mirroring the conditional check in the new API, or potentially a better approach would be something along those lines:
```diff
diff --git a/wtxmgr/db.go b/wtxmgr/db.go
index c7e3619..112aa7c 100644
--- a/wtxmgr/db.go
+++ b/wtxmgr/db.go
@@ -428,7 +428,8 @@ func readRawTxRecord(txHash *chainhash.Hash, v []byte, rec *TxRecord) error {
 	}
 	rec.Hash = *txHash
 	rec.Received = time.Unix(int64(byteOrder.Uint64(v)), 0)
-	err := rec.MsgTx.Deserialize(bytes.NewReader(v[8:]))
+	rec.SerializedTx = v[8:]
+	err := rec.MsgTx.Deserialize(bytes.NewReader(rec.SerializedTx))
 	if err != nil {
 		str := fmt.Sprintf("%s: failed to deserialize transaction %v",
 			bucketTxRecords, txHash)
```

wallet/tx_reader.go
mohamedawnallah: **thought:** looks like we can put this earlier in the beginning of `calculateValueAndFee`. That way we can make most use of the early return

mohamedawnallah: **nitpick:** It can be helpful adding comment above this on why we return here. Something along those lines:
```
// If not all inputs are ours, we can't calculate the total fee.
// txDetails.Debits contains only our inputs, while txDetails.MsgTx.TxIn
// contains all inputs. If they differ, some inputs belong to external
// wallets and we don't know their input values.
if len(txDetails.Debits) != len(txDetails.MsgTx.TxIn) {
	return
}
```

wallet/tx_reader.go
```diff
@@ -0,0 +1,131 @@
...
+// TxReader provides an interface for querying tx history.
+type TxReader interface {
+	// GetTx returns a detailed description of a tx given its tx hash.
+	GetTx(ctx context.Context, txHash chainhash.Hash) (
+		*TxDetail, error)
+
+	// ListTxns returns a list of all txns which are relevant to the wallet
+	// over a given block range.
+	ListTxns(ctx context.Context, startHeight, endHeight int32) (
+		[]*TxDetail, error)
```
mohamedawnallah: Much simpler API design contract compared to the old one `GetTransactions` 👍

wallet/tx_reader.go
```diff
@@ -160,6 +164,52 @@ func (w *Wallet) GetTx(_ context.Context, txHash chainhash.Hash) (
 	return w.buildTxDetail(txDetails, currentHeight), nil
 }
 
+// ListTxns returns a list of all txns which are relevant to the
+// wallet over a given block range.
+//
+// NOTE: This method is part of the TxReader interface.
+//
+// Time complexity: O(B + T), where B is the number of blocks in the range and T
+// is the number of transactions in those blocks.
```
mohamedawnallah: **question(blocking):** It seems the time complexity for `ListTxns` API is `O(B * T * (I+O))` not `O(B + T)`? 

Where:
- `B` is blocks in range – https://github.com/btcsuite/btcwallet/blob/master/wtxmgr/query.go#L324
- `T` is transactions per block – https://github.com/btcsuite/btcwallet/blob/master/wtxmgr/query.go#L333
- `I+O` is per-transaction processing (inputs + outputs) – In the callback function `buildTxDetail` i.e `populatePrevOuts` and `populateOutputs` respectively

We can boil down time complexity asymptotically to `O(N)` where N = total inputs + outputs across all transactions in blocks range

wallet/tx_reader.go
```diff
@@ -160,6 +164,52 @@ func (w *Wallet) GetTx(_ context.Context, txHash chainhash.Hash) (
...
+func (w *Wallet) ListTxns(_ context.Context, startHeight,
+	endHeight int32) ([]*TxDetail, error) {
```
mohamedawnallah: **question(blocking):** Any sanity check to ensure `startHeight` ≤ `endHeight` boundary range? Probably we can also return public error to the user if that early check has been violated

wallet/tx_reader.go
mohamedawnallah: **thought:**  It can be helpful if we put this anonymous callback function in a variable above. Something along those:
```golang
processTxBatch := func(batch []wtxmgr.TxDetails) (bool, error) {
    for i := range batch {
        detail := &batch[i]
        txDetail := w.buildTxDetail(detail, currentHeight)
        details = append(details, txDetail)
    }
    // Continue processing all transactions.
    return false, nil
}
```

Then we call it here:
```golang
err := w.txStore.RangeTransactions(txmgrNs, startHeight, endHeight, processTxBatch)
...
```

wallet/tx_writer.go
```diff
@@ -0,0 +1,55 @@
...
+import (
+	"context"
+	"fmt"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcwallet/walletdb"
+)
+
+// TxWriter provides an interface for updating wallet txns.
+type TxWriter interface {
```
mohamedawnallah: **question(non-blocking):** Perhaps better to be `TxLabeler` instead of `TxWriter`? Currently, the only method it has is `LabelTx`, which suggests the name `TxLabeler` would be more reflective. Not sure though if there is an ongoing intention to extend this interface with other methods in future

wallet/tx_reader.go
mohamedawnallah: **thought:** Relatedly, why we return all txns  when `startHeight=0` and `endHeight=-1` in case of the new API `ListTxns`? When `startHeight=-1` and `endHeight=-1` perceived to me more familiar. That also would make enforcing the constraint above `startHeight`  ≤ `endHeight` easier since the latter satisfies it

pkg/unit/rates.go
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2025 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+// Package unit provides a set of types for dealing with bitcoin units.
+package unit
```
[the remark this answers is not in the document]
yyforyongyu: Agree - updated.

wallet/tx_reader.go
yyforyongyu: yeah I like the second approach, updated.

wallet/tx_reader.go
yyforyongyu: nice added

wallet/tx_reader.go
```diff
@@ -160,6 +164,52 @@ func (w *Wallet) GetTx(_ context.Context, txHash chainhash.Hash) (
...
+
+	bestBlock := w.SyncedTo()
+	currentHeight := bestBlock.Height
+
+	var details []*TxDetail
+
+	err := walletdb.View(w.db, func(dbtx walletdb.ReadTx) error {
+		txmgrNs := dbtx.ReadBucket(wtxmgrNamespaceKey)
+
+		err := w.txStore.RangeTransactions(
```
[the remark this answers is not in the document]
yyforyongyu: yeah it will be done in the `Store` interface series.

wallet/tx_reader.go
yyforyongyu: This is no longer relevant as I realized the db tx can be made even smaller, and updated

yyforyongyu:
needs a reACK on the last commit 071e0fd3167ec71808d929510738a3a6d7b46d4a

GustavoStingelin:
ACK

Issue #1037, opened 2025-08-12
Interface Design for `Signer`


yyforyongyu:
The ["Low-Level Signing & Key Derivation"](https://gist.github.com/yyforyongyu/c0a6ec78cedcbf5fb342be412dc591ea) group provides direct access to the wallet's cryptographic capabilities. This interface is critical for consumers like `lnd` that need to build complex, non-standard transactions. The design prioritizes clarity, consistency, and security.

### Rationale for Changes

1.  **Consolidate Key Derivation**:
    *   `DeriveFromKeyPath` and `DeriveFromKeyPathAddAccount` are combined into a single `DerivePrivKey` method. The `AddAccount` variant mixes account management with key derivation, which violates the single responsibility principle. The caller should use the `AccountManager` to ensure an account exists before attempting to derive keys from it.

2.  **Preserve Distinct Signing Primitives**:
    *   `SignOutputRaw` (returning a raw signature) and `ComputeInputScript` (returning a full script) are both kept. They serve different but equally important use cases for `lnd`. `SignOutputRaw` is essential for building custom witness stacks (e.g., for commitment transactions), while `ComputeInputScript` is a convenience for spending standard outputs.

3.  **Highlight Security-Sensitive Methods**:
    *   `PrivKeyForAddress` exports raw private key material, which is a significant security risk if used improperly. While necessary for certain advanced protocols like `musig2`, its presence in this low-level interface is appropriate, and its documentation will include a strong warning.

### Final `Signer` Interface Design

This refinement results in a clean, focused interface for all low-level cryptographic operations, with clear documentation about its powerful and sensitive nature.

```go
// Signer provides an interface for low-level cryptographic operations,
// including signing and key derivation. This interface should be used with
// caution as it provides access to sensitive key material.
type Signer interface {
    // SignOutputRaw generates a raw signature for a single transaction input.
    SignOutputRaw(ctx context.Context, tx *wire.MsgTx, signDesc *input.SignDescriptor) (input.Signature, error)

    // ComputeInputScript generates the full sigScript and witness required to
    // spend a UTXO.
    ComputeInputScript(ctx context.Context, tx *wire.MsgTx, signDesc *input.SignDescriptor) (*input.Script, error)

    // SignMessage signs an arbitrary message with a key from the wallet.
    SignMessage(ctx context.Context, keyLoc KeyLocator, msg []byte, doubleHash bool) (*ecdsa.Signature, error)

[...]

PR #1092, opened 2025-09-30
wallet: add `Signer` interface

yyforyongyu:
This PR fixes https://github.com/btcsuite/btcwallet/issues/1037 - the old design has evolved here, detailed below.

## Initial State: A Powerful but Risky Interface

The initial design, outlined in #1037, proposed a single, powerful `Signer` interface. It combined routine cryptographic operations (signing transactions, deriving public keys) with highly sensitive, security-critical functions (exporting raw private keys).

While functional, this design presented several significant issues:

1.  **High Security Risk**: Mixing "safe" and "unsafe" methods in a single interface makes it too easy for a consumer of the API to accidentally call a dangerous method like `GetPrivKeyForAddress` when they intended to use a safer alternative. Security relied solely on a `DANGER` comment, which is insufficient for preventing misuse.
2.  **Lack of Extensibility**: The methods used primitive arguments (e.g., `*wire.MsgTx`, `*input.SignDescriptor`). This made the method signatures rigid and hard to extend. Adding a new feature, like support for Taproot signing, would likely require a breaking API change.
3.  **Inflexible `SignMessage`**: The message signing function was basic, only supporting ECDSA. It lacked the flexibility to support different signature schemes (like Schnorr) or formats (like compact, recoverable signatures) required by modern applications.

## The Design Evolution: Security, Clarity, and Extensibility

The final API was shaped by a strong focus on creating a safer, more expressive, and future-proof interface for cryptographic operations.

### Insight 1: Security Through Interface Segregation

The most critical design decision was to split the original `Signer` into two distinct interfaces:

-   **`Signer`**: This interface contains only the "safe" methods that do not export raw private key material. It is intended for the majority of use cases.
-   **`UnsafeSigner`**: This interface embeds the `Signer` interface and adds the two dangerous methods: `DerivePrivKey` and `GetPrivKeyForAddress`.

This separation forces developers to make a conscious, explicit decision to use the `UnsafeSigner`, dramatically reducing the risk of accidental private key exposure. It is a compile-time safeguard that makes the API inherently safer.

### Insight 2: Clarity and Extensibility with Parameter Objects

Instead of passing long, complex lists of arguments, the new design uses dedicated parameter objects for its primary methods:

-   `ComputeUnlockingScript` uses `UnlockingScriptParams`.
-   `ComputeRawSig` uses `RawSigParams`.
-   `SignMessage` uses `SignMessageIntent`.

This approach has two major benefits:

1.  **Clarity**: It makes the code self-documenting. The parameters for each operation are grouped into a single, coherent structure.
2.  **Extensibility**: It allows for non-breaking API changes. New options can be added to the parameter structs without changing the method signatures, making the API much more adaptable to future protocol changes.

### Insight 3: Explicit Support for Spend Types via the Strategy Pattern

To handle the complexity of signing different types of inputs (Legacy, SegWit v0, Taproot), the `RawSigParams` struct includes a `SpendDetails` field. This is a sealed interface with concrete implementations for each spend type (`LegacySpendDetails`, `SegwitV0SpendDetails`, `TaprootSpendDetails`).

This is an application of the **Strategy Pattern**. It cleanly separates the signing logic for each input type, making the code easier to understand, test, and extend. Adding support for a new input type in the future would simply require a new `SpendDetails` implementation, with no changes to the core `Signer` interface.

## The Final API: Safe, Expressive, and Future-Proof

The final design provides a robust and secure foundation for all cryptographic operations in the wallet.

```go
// Signer provides an interface for common, safe cryptographic operations.
type Signer interface {
	DerivePubKey(ctx context.Context, path BIP32Path) (*btcec.PublicKey, error)
	ECDH(ctx context.Context, path BIP32Path, pub *btcec.PublicKey) ([32]byte, error)
	SignMessage(ctx context.Context, path BIP32Path, intent *SignMessageIntent) (Signature, error)
	ComputeUnlockingScript(ctx context.Context, params *UnlockingScriptParams) (*UnlockingScript, error)
	ComputeRawSig(ctx context.Context, params *RawSigParams) (RawSignature, error)
}

// UnsafeSigner provides an interface for security-sensitive cryptographic
// operations that export raw private key material.
type UnsafeSigner interface {
	Signer
	DerivePrivKey(ctx context.Context, path BIP32Path) (*btcec.PrivateKey, error)
	GetPrivKeyForAddress(ctx context.Context, a btcutil.Address) (*btcec.PrivateKey, error)
}
```

### Why This Design is Superior:

1.  **Inherently Safer**: The separation of `Signer` and `UnsafeSigner` is a powerful, compile-time security feature that prevents the accidental exposure of private keys.
2.  **More Extensible**: The use of parameter objects and the strategy pattern (`SpendDetails`) makes the API highly adaptable to future changes in the Bitcoin protocol.
3.  **More Expressive and Clear**: The API is more self-documenting. The intent of the caller is made explicit through the use of structs like `SignMessageIntent` and `RawSigParams`.
4.  **More Powerful**: The new `SignMessage` and `ComputeRawSig` methods provide the flexibility needed to support advanced, multi-party protocols like those used in Lightning.

### DIFF
--- a/wallet/tx_creator_test.go
+++ b/wallet/tx_creator_test.go
@@ -33,11 +33,13 @@ var (
 func TestValidateTxIntent(t *testing.T) {
 	t.Parallel()
 
+	const defaultAccountName = "default"
+
 	// Define a set of valid outputs and inputs to be reused across test
 	// cases.
 	validOutput := wire.TxOut{Value: 10000, PkScript: []byte{}}
 	validUTXO := wire.OutPoint{Hash: [32]byte{1}, Index: 0}
-	validAccountName := "default"
+	validAccountName := defaultAccountName
 	validScopedAccount := &ScopedAccount{
 		AccountName: validAccountName,
 		KeyScope:    waddrmgr.KeyScopeBIP0086,
@@ -58,7 +60,7 @@ func TestValidateTxIntent(t *testing.T) {
 					UTXOs: []wire.OutPoint{validUTXO},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -73,7 +75,7 @@ func TestValidateTxIntent(t *testing.T) {
 					Source: validScopedAccount,
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -91,7 +93,7 @@ func TestValidateTxIntent(t *testing.T) {
 					},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -103,7 +105,7 @@ func TestValidateTxIntent(t *testing.T) {
 				Outputs: []wire.TxOut{validOutput},
 				Inputs:  &InputsPolicy{Source: nil},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -115,7 +117,7 @@ func TestValidateTxIntent(t *testing.T) {
 				Outputs: []wire.TxOut{validOutput},
 				Inputs:  nil,
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -129,7 +131,7 @@ func TestValidateTxIntent(t *testing.T) {
 					UTXOs: []wire.OutPoint{validUTXO},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -143,7 +145,7 @@ func TestValidateTxIntent(t *testing.T) {
 					UTXOs: []wire.OutPoint{validUTXO},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -157,7 +159,7 @@ func TestValidateTxIntent(t *testing.T) {
 					UTXOs: []wire.OutPoint{},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -173,7 +175,7 @@ func TestValidateTxIntent(t *testing.T) {
 					},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -187,7 +189,7 @@ func TestValidateTxIntent(t *testing.T) {
 					Source: &ScopedAccount{AccountName: ""},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -203,7 +205,7 @@ func TestValidateTxIntent(t *testing.T) {
 					},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
@@ -221,7 +223,7 @@ func TestValidateTxIntent(t *testing.T) {
 					},
 				},
 				ChangeSource: &ScopedAccount{
-					AccountName: "default",
+					AccountName: defaultAccountName,
 				},
 				FeeRate: 1000,
 			},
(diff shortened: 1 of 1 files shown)


GustavoStingelin:
Is it already supposed to be reviewed?

wallet/deprecated.go
```diff
@@ -150,3 +154,81 @@ func (w *Wallet) RenameAccountDeprecated(scope waddrmgr.KeyScope,
...
 	return err
 }
+
+// ScriptForOutputDeprecated returns the address, witness program and redeem
+// script for a given UTXO. An error is returned if the UTXO does not
+// belong to our wallet or it is not a managed pubKey address.
+//
+// Deprecated: Use AddressManager.ScriptForOutput instead.
+func (w *Wallet) ScriptForOutputDeprecated(output *wire.TxOut) (
+	waddrmgr.ManagedPubKeyAddress, []byte, []byte, error) {
```
mohamedawnallah: 👍

wallet/signer.go
```diff
@@ -1,92 +1,804 @@
...
+	Msg []byte
 
-	script, err := w.ScriptForOutput(context.Background(), *output)
-	if err != nil {
-		return nil, nil, nil, err
-	}
+	// DoubleHash specifies whether the message should be double-hashed
+	// (SHA256d) before signing. If false, a single SHA256 hash is used.
+	DoubleHash bool
+
```
GustavoStingelin: If this flag is false, then Hash160 is being used, which is very different from a single SHA or a double SHA. Maybe we can add an enum so the caller can choose between double SHA and RIPEMD160 in a clear way. That would make the behavior easier to read.

That said, I do not know any situation in Bitcoin where we sign a digest produced with RIPEMD160. As far as I know, signatures are always made over a double SHA digest. If there is a real reason to support RIPEMD160 here, it would be great to explain that context.

wallet/signer.go
```diff
@@ -1,92 +1,804 @@
...
+	if intent.Schnorr.Tag != nil {
+		taggedHash := chainhash.TaggedHash(
+			intent.Schnorr.Tag, intent.Msg,
+		)
+		digest = taggedHash[:]
+	} else {
+		digest = btcutil.Hash160(intent.Msg)
+	}
+
+	sig, err := schnorr.Sign(privKey, digest)
```
GustavoStingelin: The Schnorr signer only accepts digests that are exactly 32 bytes. RIPEMD160 returns a 20 byte value, so this will fail right away. I am trying to understand the reason for using this hash here. Is there any case where we need this shorter digest for signing?

I also do not see why we cannot sign with Schnorr without a tag. This seems a bit tied to the way Taproot builds its tagged hashing, but Schnorr itself does not require that. This could be an opportunity to improve this design.

wallet/signer.go
mohamedawnallah: > If this flag is false, then Hash160 is being used, which is very different from a single SHA or a double SHA. Maybe we can add an enum so the caller can choose between double SHA and RIPEMD160 in a clear way. That would make the behavior easier to read.

Good point. I have pending review comment which something along those lines. Why aren't we mirroring the LND one (HashB returns 32 bytes when doubleHash is false)?
https://github.com/lightningnetwork/lnd/blob/8c8662c86ae5dfa371941086bb828b0af060f7fd/lnwallet/btcwallet/signer.go#L416-L421

yyforyongyu: oh nice catch! yeah this indeed should be single hash.

wallet/signer.go
yyforyongyu: good question - it's weird that this isn't caught in the unit tests, i think the coverage rate is too low, and this path was never tested. Gonna increase the coverage now!

wallet/tx_creator_test.go
```diff
@@ -33,11 +33,13 @@ var (
 func TestValidateTxIntent(t *testing.T) {
 	t.Parallel()
 
+	const defaultAccountName = "default"
```
yyforyongyu, line 36: This is weird as it sometimes pop up, looks like there's flakeness in the `make lint`, maybe due to caching.

wallet/signer.go
yyforyongyu: I further checked how `bitcoind` handles this, it turns out Schnorr is not supported...I think we can provide a low level method instead, and let the user decide how they wanna hash the msg, appended a new commit to present this new design.

wallet/signer.go
```diff
@@ -523,3 +523,31 @@ func (w *Wallet) fetchManagedPubKeyAddress(path BIP32Path) (
...
+	if err != nil {
+		return [32]byte{}, fmt.Errorf("cannot get private key: %w",
+			err)
+	}
+	defer privKey.Zero()
+
+	// Perform the scalar multiplication and hash the result.
+	secret := btcec.GenerateSharedSecret(privKey, pub)
+
+	var sharedSecret [32]byte
```
mohamedawnallah, line 560: **feel-free-to-ignore:** `32` can be public constant variable

wallet/address_manager.go
```diff
@@ -57,7 +69,7 @@ type AddressProperty struct {
 // Script represents the script information required to spend a UTXO.
 type Script struct {
 	// Addr is the managed address of the UTXO.
```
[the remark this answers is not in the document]
yyforyongyu: Given that the signing process has been delegated to the signer, i don't think we need to mention that the key returned here needs to be a `ManagedPubKeyAddress` as now it's purely providing into.

Issue #1097, opened 2025-10-05
[Documentation]: mapping existing Key-Value (KV) DB schema for future exploration and migration to SQL DB


mohamedawnallah:
## Upstream

https://github.com/btcsuite/btcwallet/issues/1094
https://github.com/btcsuite/btcwallet/issues/1015

## Legend

⬜: Not ready yet
✅: Ready

## Change Description

| Status | Entity                    | Artifact |
|----------|---------------------------|----------|
| ✅ | `b` - Blocks              |      [Mapping Doc #0](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369307295)    |
| ✅ | `t` - Transaction Records |     [Mapping Doc #1](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369307404)     |
| ✅ |`u` - Unspent             |    [Mapping Doc #2](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369307472)      |
| ✅ | `c` - Credits             |    [Mapping Doc #3](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369307678)      |
| ✅ | `d` - Debits              |  [Mapping Doc #4](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369307752)        |
| ✅ | `m` - Unmined             |     [Mapping Doc #5](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369307849)     |
| ✅ | `mc` - UnminedCredits     |   [Mapping Doc #6](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369307911)       |
| ✅ | `mi` - UnminedInputs      |     [Mapping Doc #7](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369308005)     |
| ✅ |`lo` - LockedOutputs      |     [Mapping Doc #8](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369308101)     |
| ✅ | `bal` - Balance           |    [Mapping Doc #9](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369308189)      |
| ✅ | `date` - CreateDate       |     [Mapping Doc #10](https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369308341)     |
[...]

PR #1111, opened 2025-11-06
wallet: introduce db layer interfaces

yyforyongyu:
### Summary

This is a **draft PR** to introduce a new, granular database interface located in the `wallet/internal/db` package. The primary goal of this change is to create a clean abstraction layer that decouples the wallet's business logic from the underlying storage implementation (currently `waddrmgr` and `wtxmgr`).

This PR introduces the interface design only. The `KvdbStore` [[implementation](https://github.com/yyforyongyu/btcwallet/blob/store-draft/wallet/internal/db/driver.go#L134)](https://github.com/yyforyongyu/btcwallet/blob/store-draft/wallet/internal/db/driver.go#L134) is a temporary and incomplete adapter meant to satisfy the interface, but the focus of this review should be on the design of the interfaces themselves.

### Motivation

Currently, the wallet's core logic is tightly coupled to the legacy `waddrmgr` and `wtxmgr` managers. This has several significant drawbacks:

*   **Poor Testability:** It is very difficult to write unit tests for business logic that relies on these complex, stateful managers.
*   **Difficult Maintenance:** The lack of a clear boundary makes the code harder to reason about and refactor safely.
*   **Inflexibility:** It makes it nearly impossible to swap out the database backend. A major long-term goal is to introduce a more robust, multi-wallet SQL backend, which is blocked by the current architecture.

This new `db` interface serves as the foundation for all future database work, providing a clean contract for the rest of the application to build upon.

### The New Interface Design

The new design is split into five granular, role-based interfaces: `WalletStore`, `AccountStore`, `AddressStore`, `TxStore`, and `UTXOStore`. This separation makes the API easier to understand and implement.

The design adheres to several key principles:

*   **Consistent Use of Parameter Structs:** All methods accept a single `params` or `query` struct, which improves readability and makes the API easy to extend without breaking changes.
*   **Idiomatic Handling of Optionals:** Pointers are used for optional fields in update operations (e.g., `UpdateTxParams.Label *string`) to unambiguously signal the caller's intent for partial updates.
*   **Safe and Performant Return Types:** Methods that return a single entity return a pointer (`*AccountInfo`), providing a clear `nil` signal for "not found". Methods that return a list return a slice of values (`[]AccountInfo`), which is more performant (due to cache locality) and safer.
*   **Forward-Looking for Multi-Wallet Support:** All relevant methods accept a `WalletID`, making the interface ready for a true multi-wallet SQL implementation.

### Pros and Cons

**Pros:**

*   **Decoupling:** Provides a clean separation between business logic and data storage.
*   **Testability:** The new interfaces can be easily mocked for robust unit testing.
*   **Extensibility:** Unlocks the ability to implement new database backends (e.g., SQLite/PostgreSQL) without changing the wallet's core logic.
*   **API Safety & Ergonomics:** The consistent patterns make the API safer, more predictable, and easier to use correctly.

**Cons & Trade-offs:**

*   **Potential for "Chatty" API Calls:** The granular nature of the interfaces may require some higher-level operations to make multiple calls to the database. This is a deliberate trade-off for clarity and flexibility.
*   **Derived Data Complexity:** `TxInfo` is now a minimal record. The caller is responsible for deriving related data like credits and debits by querying the UTXO store. This is a positive trade-off for a more normalized and robust database schema.

### Draft Status & Future Work

This interface is the result of careful consideration, but it is still a draft. It is expected that as we begin to integrate it into the wallet's business logic, we may discover areas for improvement or adjustment. Feedback on the overall design and patterns is highly encouraged at this stage.

### TODOs and Next Steps

1.  Gather feedback on this initial interface design.
2.  Complete the `KvdbStore` implementation to fully satisfy the new interfaces, serving as a temporary bridge.
3.  Begin integrating the new `db` interface into the higher-level wallet logic, replacing direct calls to `waddrmgr` and `wtxmgr`.
4.  Based on real-world usage patterns observed during integration, identify any performance hotspots that may require new, specialized batch methods in the interface.

### DIFF
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -1,0 +1,886 @@
+// Package db provides a database-agnostic interface for wallet data storage,
+// defining the core data types and store interfaces for wallets, accounts,
+// addresses, transactions, and UTXOs.
+package db
+
+import (
+	"time"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/hdkeychain"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/wire"
+)
+
+// ============================================================================
+// Data Types & Method Parameters
+// ============================================================================
+
+// KeyScope represents the BIP-44 key scope as defined in BIP-43. It is used
+// to organize keys based on their purpose and coin type, providing a
+// hierarchical structure for key derivation.
+type KeyScope struct {
+	// Purpose is the purpose number for the scope, as defined in BIP-43.
+	Purpose uint32
+
+	// Coin is the coin type number for the scope, as defined in BIP-44.
+	Coin uint32
+}
+
+// AddressType specifies the type of a managed address. This is used to
+// identify the script type of an address, such as P2PKH, P2SH, P2WKH, etc.
+type AddressType uint8
+
+const (
+	// PubKeyHash represents a pay-to-pubkey-hash (P2PKH) address.
+	PubKeyHash AddressType = iota
+
+	// ScriptHash represents a pay-to-script-hash (P2SH) address.
+	ScriptHash
+
+	// WitnessPubKey represents a pay-to-witness-pubkey-hash (P2WKH)
+	// address.
+	WitnessPubKey
+
+	// NestedWitnessPubKey represents a P2WKH output nested within a P2SH
+	// address.
+	NestedWitnessPubKey
+
+	// TaprootPubKey represents a pay-to-taproot (P2TR) address.
+	TaprootPubKey
+)
+
+const (
+	// BIP0044Purpose is the purpose field for BIP0044 derivation.
+	BIP0044Purpose = 44
+
+	// BIP0049Purpose is the purpose field for BIP0049 derivation.
+	BIP0049Purpose = 49
+
+	// BIP0084Purpose is the purpose field for BIP0084 derivation.
+	BIP0084Purpose = 84
+
+	// BIP0086Purpose is the purpose field for BIP0086 derivation.
+	BIP0086Purpose = 86
+)
+
+var (
+	// KeyScopeBIP0049Plus is the key scope of our modified BIP0049
+	// derivation. We say this is BIP0049 "plus", as it acts as an
+	// optimization for fee savings. Standard BIP0049 uses P2SH-wrapped
+	// SegWit for both external (receive) and internal (change) addresses.
+	// This scheme uses P2SH-wrapped SegWit for external addresses (to
+	// ensure backward compatibility with senders using legacy wallets) but
+	// uses Native SegWit (P2WKH) for internal change addresses. Since the
+	// wallet controls its own change, it can use the more efficient Native
+	// SegWit format to reduce transaction weight and save on fees when
+	// spending that change later.
+	KeyScopeBIP0049Plus = KeyScope{
+		Purpose: BIP0049Purpose,
+		Coin:    0,
+	}
+
+	// KeyScopeBIP0084 is the key scope for BIP0084 derivation. BIP0084
+	// will be used to derive all p2wkh addresses.
+	KeyScopeBIP0084 = KeyScope{
+		Purpose: BIP0084Purpose,
+		Coin:    0,
+	}
+
+	// KeyScopeBIP0086 is the key scope for BIP0086 derivation. BIP0086
+	// will be used to derive all p2tr addresses.
+	KeyScopeBIP0086 = KeyScope{
+		Purpose: BIP0086Purpose,
+		Coin:    0,
+	}
+
+	// KeyScopeBIP0044 is the key scope for BIP0044 derivation. Legacy
+	// wallets will only be able to use this key scope, and no keys beyond
+	// it.
+	KeyScopeBIP0044 = KeyScope{
+		Purpose: BIP0044Purpose,
+		Coin:    0,
+	}
+
+	// ScopeAddrMap is a map from the default key scopes to the scope
+	// address schema for each scope type.
+	ScopeAddrMap = map[KeyScope]ScopeAddrSchema{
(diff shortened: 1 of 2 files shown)


wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,799 @@
...
+	// Hash is the 32-byte hash of the block.
+	Hash chainhash.Hash
+
+	// Height is the height of the block in the blockchain.
+	Height int32
+
+	// Timestamp is the timestamp of the block, which is used for wallet
+	// synchronization and rescan operations.
+	Timestamp time.Time
+}
```
mohamedawnallah: It seems this should match the underlying kvdb persisted data even though some persisted data not actually used, right? If so perhaps we can consider adding the transaction count and transaction hash(es) as seen here:
https://github.com/btcsuite/btcwallet/issues/1097#issuecomment-3369307295

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,799 @@
...
+	// Index is the index of the key in the branch.
+	Index uint32
+}
+
+// --------------------
+// TxStore Types
+// --------------------
+
+// TxInfo represents the details of a transaction relevant to the wallet.
+type TxInfo struct {
```
mohamedawnallah: So this Tx represents unmined, mined, or both?

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,799 @@
...
+
+	// SerializedTx is the serialized transaction.
+	SerializedTx []byte
+
+	// Received is the timestamp when the transaction was received.
+	Received time.Time
+
+	// Block contains metadata about the block that includes the
+	// transaction.
+	Block Block
```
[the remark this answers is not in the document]
yyforyongyu: Yeah I like it. Atm we use `Block{Height: -1}` to represent an unmined tx, but using a pointer is more straightforward, updated!

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,800 @@
...
+
+	// PublicPassphrase is the passphrase used to encrypt public data.
+	PublicPassphrase []byte
+
+	// PrivatePassphrase is the passphrase used to encrypt private keys.
+	PrivatePassphrase []byte
+
+	// HDSeed is the Hierarchical Deterministic (HD) seed for the new
+	// wallet.
+	HDSeed []byte
```
GustavoStingelin:
```suggestion
	// HDSeed represents the binary seed used to derive the master private key
    // in an HD wallet as defined by BIP32.
	HDSeed []byte
```
Maybe could clarify a little bit what seed we're talking about.

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,800 @@
...
+
+	// SyncedTo is the new synchronization state for the wallet.
+	SyncedTo *Block
+}
+
+// ChangePassphraseParams contains the parameters for changing a wallet's
+// passphrase.
+type ChangePassphraseParams struct {
+	// WalletID is the ID of the wallet to update.
+	WalletID uint64
```
GustavoStingelin: same from db int size

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,800 @@
...
+// CreateWalletParams contains the parameters required to create a new wallet.
+type CreateWalletParams struct {
+	// Name is the name of the new wallet.
+	Name string
+
+	// PublicPassphrase is the passphrase used to encrypt public data.
+	PublicPassphrase []byte
+
+	// PrivatePassphrase is the passphrase used to encrypt private keys.
+	PrivatePassphrase []byte
```
GustavoStingelin: Since the database does not store the password, only an encrypted encryption key used to decrypt the master private key, how is the passphrase expected to be handled at the database level?

I believe the database layer should not be responsible for decrypting this information, but rather just for storing any new data it receives.

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,800 @@
...
+
+// GetAccountQuery contains the parameters for querying a single account. The
+// query must specify either the account name or the account number. Using
+// pointers for these fields allows the query to be unambiguous, as a nil value
+// indicates that the field should not be used for filtering. This avoids the
+// "zero value" problem, where 0 or an empty string could be valid query
+// targets.
+type GetAccountQuery struct {
+	// WalletID is the ID of the wallet to query.
+	WalletID uint64
```
GustavoStingelin: same from db int size

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,800 @@
...
+	DerivationInfo *DerivationInfo
+
+	// Script is the script associated with the address, if any.
+	Script []byte
+}
+
+// NewAddressParams contains the parameters for creating a new address.
+type NewAddressParams struct {
+	// WalletID is the ID of the wallet to create the address in.
+	WalletID uint64
```
GustavoStingelin: same from db int size

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,800 @@
...
+	Change bool
+}
+
+// ImportAddressParams encapsulates all the data needed to store a new, imported
+// address, script, or private key. All imported addresses are automatically
+// assigned to the wallet's logical "imported" account. The presence of a
+// private key determines whether the address will be spendable or watch-only.
+type ImportAddressParams struct {
+	// WalletID is the ID of the wallet to import the address into.
+	WalletID uint64
```
GustavoStingelin: same from db int size

wallet/internal/db/data_types.go
```diff
@@ -0,0 +1,800 @@
...
+}
+
+// Block defines a block's hash, height, and timestamp. This is used to
+// represent a block's identity and position in the blockchain.
+type Block struct {
+	// Hash is the 32-byte hash of the block.
+	Hash chainhash.Hash
+
+	// Height is the height of the block in the blockchain.
+	Height int32
```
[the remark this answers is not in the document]
yyforyongyu: yeah it was created to stay compatible with other callsite, but agree we should be strict at the db layer, and can convert at a higher layer, updated.

PR #1112, opened 2025-11-07
docs: add ADR for multi wallet support

yyforyongyu:
This document outlines the architectural decision regarding multi-wallet support in `btcwallet`, particularly in the context of implementing a new SQL-based database backend (SQLite/Postgres).

**The final recommendation is to build the architecture to explicitly support multiple wallets.**

While a multi-wallet design introduces a moderate, one-time complexity at the database schema and application level, its benefits are fundamental. It correctly solves the unavoidable problem of network separation, provides a vastly superior user experience, is highly resource-efficient, and future-proofs the application.

Conversely, a single-wallet model, while appearing simpler, is a false economy. It fails to handle multiple networks safely, pushes significant operational complexity onto the end-user, and scales poorly.

## 2. The Core Question

The fundamental question is whether `btcwallet` should be architected as:

*   **A) A single-wallet system:** One daemon process manages exactly one wallet. Managing multiple wallets would require running multiple, separate daemon processes.
*   **B) A multi-wallet system:** A single daemon process can load, manage, and operate on multiple, distinct wallets simultaneously.

This decision is critical as it dictates the fundamental design of the database schema and the application's core logic.

## 3. Arguments FOR a Multi-Wallet Architecture (Pros)

These benefits center on **Correctness, Efficiency, and User Flexibility**.

### 3.1. Solves the Unavoidable Network Separation Problem

A wallet's data (UTXOs, transactions) is only valid on a specific blockchain (`mainnet`, `testnet`, etc.). A design that attempts to switch networks within a single wallet context is inherently unsafe and prone to data corruption. The multi-wallet model correctly frames this reality by making the network an intrinsic property of a wallet. The `wallet_id` becomes the single, unambiguous identifier that scopes all data to a specific chain, preventing any possibility of cross-network contamination.

### 3.2. Superior User Experience for Same-Network Wallets

*   **Centralized Management:** The user manages one daemon, one configuration file, one data directory, and one set of RPC ports for all wallets on a given network.
*   **Seamless Wallet Switching:** A user can seamlessly switch between multiple wallets on the same network (e.g., a "hot" and "cold" mainnet wallet) in consecutive commands to the same daemon, simply by changing a `--wallet` flag. The alternative (managing separate processes) is cumbersome and error-prone. Note that switching between networks (e.g. mainnet to testnet) still requires restarting the daemon against the appropriate chain backend.

### 3.3. Significant System Resource Efficiency

*   **Shared Chain Connection:** A single daemon maintains one connection to the backend node (`btcd`). It downloads and processes each block *once* and distributes the relevant information to all loaded wallets. This is dramatically more efficient than a single-wallet-per-process model, which would amplify the load on the node and network for each wallet.

### 3.4. Enables Powerful Fund Segregation and Security Models

*   **Hot vs. Cold Wallets:** Users can run a low-balance "hot wallet" for frequent use alongside a high-balance, encrypted "cold wallet" that is rarely unlocked, all within the same process.
*   **Compartmentalization:** Users can create separate wallets for different purposes (e.g., "Business Expenses," "Personal Savings") to simplify accounting and improve privacy.

### 3.5. Future-Proofs the Application

This architecture opens the door for future features that are impossible otherwise, such as inter-wallet transfers. The cost of retrofitting a multi-wallet design into a single-wallet database is immense and risky; building it correctly from the start is far cheaper in the long run.

## 4. Arguments AGAINST a Multi-Wallet Architecture (Cons)

These costs center on **Initial Implementation Complexity**.

### 4.1. Increased Database Schema Complexity

*   **Pervasive Scoping:** A `wallet_id` column and its associated `FOREIGN KEY` constraint must be added to nearly every table.
*   **Composite Keys:** Uniqueness constraints must often be composite (e.g., `UNIQUE(wallet_id, txid, vout)`), which requires more careful design.

### 4.2. Increased Application Logic Complexity

*   **State Management:** The application needs a "wallet manager" layer responsible for the lifecycle of each wallet: loading, unloading, locking, and unlocking.
*   **Request Routing:** All incoming requests (e.g., RPC calls) must be parsed to identify the target wallet and then routed to the correct in-memory wallet instance.

### 4.3. More Complex API / RPC Surface

Nearly every RPC call needs a parameter to specify the target wallet (e.g., `wallet_name`). This must be consistently implemented across the entire API.

### 4.4. Performance Implications (Negligible with Proper Design)

While concerns about performance with multiple wallets are valid, a properly normalized SQL schema and indexing mitigate these.

*   **Address Uniqueness:** An address string is not globally unique; its uniqueness is guaranteed by the composite key `(wallet_id, address_string)`. This correctly handles cases where the same seed is used for multiple wallets (e.g., with BIP39 passphrases).
*   **UTXO Performance:** A UTXO belongs to a single wallet. When a transaction moves funds between Wallet A and Wallet B, it's a single transaction affecting two wallets. With appropriate indexing on `(wallet_id, txid, vout)` for UTXOs and a linking table (`wallet_transactions`) for transaction history, performance for both reads and writes remains highly efficient. The slight increase in database size due to linking table entries is a minimal cost for correctness.

## 5. Final Recommendation and Implementation Path

The arguments in favor of the multi-wallet architecture are overwhelming and fundamental, while the arguments against it are challenges of implementation, not of core design.

The recommended path forward is to build out full multi-wallet support from the start of the SQL implementation.

1.  **Build the SQL schema and application logic for a full multi-wallet world.** The database schema must be designed for multiple wallets from day one. The application's core logic, including the in-memory wallet manager and RPC request routing, should also be implemented to handle multiple wallets concurrently.
2.  **Provide a Migration Path for Existing Users.** A followup release will include a migration tool. This tool will be responsible for reading the data from an existing single-wallet `wallet.db` (kvdb) and importing it into the new multi-wallet SQL database. This ensures a smooth transition for current users.
3.  **Deprecate and Remove the KVDB Backend.** Once the SQL backend is stable and the migration path is established, the legacy kvdb implementation can be fully removed from the codebase, completing the transition. During the transition period, it is understood that the kvdb backend will only ever support a single wallet.
4.  **Allow Same-Seed Wallets (with Warning):** The system will allow users to create multiple wallets using the same seed phrase (e.g., for BIP39 passphrases or watch-only purposes). However, a clear and explicit warning will be presented to the user to ensure they understand the implications of this advanced feature.

### DIFF
--- a/docs/developer/adr/0001-multi-wallet-architecture.md
+++ b/docs/developer/adr/0001-multi-wallet-architecture.md
@@ -1,0 +1,62 @@
+# ADR 0001: Multi-Wallet Architecture
+
+## 1. Context
+
+`btcwallet` is undergoing a significant architectural change, transitioning its persistent storage from a key-value database (kvdb) to a SQL-based backend (supporting SQLite and PostgreSQL). During this transition, a fundamental design decision must be made regarding the wallet's ability to manage multiple distinct wallets and networks within a single daemon instance.
+
+## 2. Decision
+
+`btcwallet` will be architected to support multiple wallets, using an **"Explicit DB-per-Network via Configuration"** model.
+
+This design prioritizes safety and clarity by encouraging users to use a separate database for each blockchain network, while retaining the flexibility to manage multiple wallets within each network.
+
+### How It Works:
+
+1.  **Explicit Configuration:** The `btcwallet.conf` file will feature network-specific sections, each with its own `db.dsn` (Data Source Name) setting. This makes the "DB-per-Network" model the default and most intuitive path.
+    ```toml
+    [mainnet]
+    db.dsn = "postgres://user:pass@host/btcwallet_mainnet"
+    
+    [testnet]
+    db.dsn = "postgres://user:pass@host/btcwallet_testnet"
+    ```
+2.  **Application Logic:** On startup, the daemon will read the active network flag (e.g., `--network=testnet`), select the corresponding DSN from the config, and connect to that specific database.
+3.  **Network Binding & Safety Check:** To prevent accidental data corruption (e.g., pointing the testnet DSN to a mainnet database), a `meta` table in each database will bind it to a specific network upon creation. On every startup, the daemon will verify this binding. If a mismatch is detected, `btcwallet` will refuse to start and will output a clear error.
+4.  **Internal Multi-Wallet Schema:** The schema *within* each database will still fully support multiple wallets (e.g., a hot and cold wallet on mainnet). This requires a `wallets` table and `wallet_id` scoping on all relevant tables. The `wallets` table itself will not need a `network` column, as this context is managed at the database level.
+
+## 3. Arguments FOR this Decision (Pros)
+
+### 3.1. Safest by Default
+This model makes it nearly impossible for a user to accidentally mix mainnet and testnet data, preventing a catastrophic class of user errors. The network binding check provides a critical backstop.
+
+### 3.2. Maximum Clarity for the User
+The configuration is explicit and self-documenting. The physical separation of databases (e.g., `mainnet.db`, `testnet.db`) aligns with the user's mental model of these being entirely separate environments.
+
+### 3.3. Retains Full Multi-Wallet Flexibility
+The design fully supports the primary multi-wallet use case: managing multiple wallets (e.g., hot/cold, personal/business) on the *same network*. It also allows for the advanced use case of a unified database for those who explicitly configure it.
+
+### 3.4. Superior Resource Efficiency
+Compared to a "one process per wallet" model, this design remains highly efficient by sharing a single chain connection (`btcd`) among all wallets running on the same network within the single daemon instance.
+
+## 4. Alternatives Considered
+
+### 4.1. Alternative: Unified Database Model
+
+A model where a single database is used for all wallets across all networks was considered. In this design, a `network` column would be added to the `wallets` table to provide logical data separation.
+
+*   **Rationale for Rejection:** While this model simplifies user configuration to a single DSN, it was rejected because it does not provide the same level of default safety. The risk of a novice user unintentionally mixing mainnet and testnet data in a single database was deemed too high. The "Explicit DB-per-Network" model provides a safer default posture while still allowing this behavior for advanced users who opt-in via their configuration.
+
+### 4.2. Alternative: Single-Wallet-per-Database Model
+
+A model where each wallet gets its own database was also considered.
+
+*   **Rationale for Rejection:** This model was rejected due to significant scalability and usability issues, particularly for PostgreSQL, where it would lead to connection limit exhaustion and require users to perform administrative database commands (`CREATE DATABASE`). It also adds significant complexity to the application, which would need to manage a connection pool for every loaded wallet.
+
+## 5. Implementation Plan
+
+1.  **Implement the "Explicit DB-per-Network" configuration** in `btcwallet.conf`.
+2.  **Build the application logic** to select the appropriate DSN based on the active network and manage the corresponding database connection.
+3.  **Implement the network binding safety check** using a `meta` table in the database schema.
+4.  **Build the internal SQL schema to be fully multi-wallet aware** (using `wallet_id` scoping) to support multiple wallets within a single network database.
+5.  **Provide a Migration Path for Existing Users.** A migration tool will be developed to import data from an existing single-wallet `wallet.db` (kvdb) into the new SQL database structure.
+6.  **Deprecate and Remove the KVDB Backend.** Once the SQL backend is stable and the migration path is established, the legacy kvdb implementation will be removed.


mohamedawnallah (commented):
This multi-wallet-single-database architecture looks good. The only concern I have regarding it (already posted previously seen here https://github.com/btcsuite/btcwallet/pull/1051#discussion_r2409266263  and still not concretely answered) is: How can someone guarantee that a database populated with non-mainnet data (testnet/regtest/signet/etc.) – for example, testnet wallets (non-monetary value) doesn't lead to unnecessarily large SQL tables and slow queries that negatively affect mainnet wallets (which hold monetary value on the chain)? In addition to the separation of monetary concerns one.

It looks like the multi-wallet-single-database architecture is more of alternative hypothesis thinking (though the mirroring argument that some libraries currently using that approach)

So we can have hypotheses like this:
**Null Hypothesis (H₀): Single-Wallet-per-Database Architecture (default thinking approach)**

A single-wallet-per-database architecture (where each wallet maintains its own separate database) provides superior performance for mainnet wallets holding monetary value, maintains optimal separation of concerns between monetary (mainnet) and non-monetary (testnet/regtest/signet) data, and prevents non-monetary test data from causing database bloat that could negatively affect mainnet wallet performance, compared to a multi-wallet-single-database architecture.

**Alternative Hypothesis (H₁): Multi-Wallet-Single-Database Architecture**

A multi-wallet-single-database architecture (where all wallets across all networks share a single database instance) does NOT significantly degrade performance for mainnet wallets holding monetary value when sharing the database with non-monetary testnet/regtest/signet wallets, does NOT create unmanageable database bloat from non-monetary test data that impacts mainnet operations, and provides acceptable separation of concerns between monetary and non-monetary wallet data through proper schema design and indexing.

It looks like if someone supports H₁ (the non-default thinking approach) they probably should have a concrete supported reproducible evidence that reject the default thinking approach (H₀) i.e no significant performance impact given above constraints and separation of concerns isn't a concern to worry about.

In addition to this LND currently follows (H₀) approach though it is following it just in kvdb/sqlite but mirroring this approach on PostgreSQL wouldn't be probably a huge dev work and can be solved with intermediary table as level of indirection to state metadata about PostresSQL current chainnet networks with some corresponding code/db layer changes.

If the opposite chosen that means LND need to do not a few changes  i.e change all kvdb/sqlite/PostgreSQL schema/db layer to support multi-wallet and providing migration with bigger surface area. It looks like a very high-resistant path compared to the above with just accommodating PostgreSQL to support single-wallet-per-db.

I may not be currently sure that H₁ would significantly impact performance in non-intended way or that speration of concerns is no longer a concern but I am sure if we adapted the H₀ approach in btcwallet sqlization we wouldn't face something new in terms of performance, user experience, or security concerns given those changes issues that need to constantly monitored and evaluated and would be less resistant path when integrating this with LND

Overall, I currently favor approach H₀ unless an evidence supported that would reject H₀ and I think if we started with single-wallet-per-db and later we wanna move to multi-wallet-per-db is probably gonna be easier since I believe converging (merging) here in that context would be easier than diverging

mohamedawnallah:
> I think the wallet should support multiple wallets, but only one chain per database.

Actually I agree with this. I thought the goal is to support multiple wallets with different chains in the same DB (which I disagreed with in that comment  https://github.com/btcsuite/btcwallet/pull/1112#pullrequestreview-3433429026). In case of multi-wallet-same-chain-per-single DB looks fine to me

docs/developer/adr/0001-multi-wallet-architecture.md
```diff
@@ -0,0 +1,62 @@
...
+
+## 1. Context
+
+`btcwallet` is undergoing a significant architectural change, transitioning its persistent storage from a key-value database (kvdb) to a SQL-based backend (supporting SQLite and PostgreSQL). During this transition, a fundamental design decision must be made regarding the wallet's ability to manage multiple distinct wallets within a single daemon instance.
+
+The existing database interface already hints at multi-wallet support through the pervasive use of `walletID` parameters. This ADR formalizes the decision to fully embrace and implement a multi-wallet architecture.
+
+## 2. Decision
+
+`btcwallet` will be architected to explicitly support multiple wallets within a single daemon instance from the outset of the SQL backend implementation.
```
mohamedawnallah: It looks like it needs to be clarified a bit here. That those wallets need to be from the same chain in a given DB

mohamedawnallah (approved):
LGTM

docs/developer/adr/0001-multi-wallet-architecture.md
```diff
@@ -0,0 +1,62 @@
...
+### How It Works:
+
+1.  **Explicit Configuration:** The `btcwallet.conf` file will feature network-specific sections, each with its own `db.dsn` (Data Source Name) setting. This makes the "DB-per-Network" model the default and most intuitive path.
+    ```toml
+    [mainnet]
+    db.dsn = "postgres://user:pass@host/btcwallet_mainnet"
+    
+    [testnet]
+    db.dsn = "postgres://user:pass@host/btcwallet_testnet"
+    ```
```
mohamedawnallah: 👍

docs/developer/adr/0001-multi-wallet-architecture.md
```diff
@@ -0,0 +1,62 @@
+# ADR 0001: Multi-Wallet Architecture
+
+## 1. Context
+
+`btcwallet` is undergoing a significant architectural change, transitioning its persistent storage from a key-value database (kvdb) to a SQL-based backend (supporting SQLite and PostgreSQL). During this transition, a fundamental design decision must be made regarding the wallet's ability to manage multiple distinct wallets and networks within a single daemon instance.
```
mohamedawnallah: 👍

yyforyongyu:
>  Is this supposed to be merged in master or

Good q - given that this is a doc, will merge to the master instead to keep the side branch smaller.

PR #1121, opened 2025-11-18
wallet: integrate `sqlfluff` metadev helpers

mohamedawnallah:
## Description

Integrate sqlfluff metadev helpers: https://github.com/sqlfluff/sqlfluff

Follow-up on https://github.com/btcsuite/btcwallet/pull/1134#pullrequestreview-3567121546.

### DIFF
--- a/wallet/internal/db/migrations/postgres/000001_blocks.up.sql
+++ b/wallet/internal/db/migrations/postgres/000001_blocks.up.sql
@@ -13,5 +13,5 @@ CREATE TABLE blocks (
     header_hash BYTEA NOT NULL UNIQUE CHECK (length(header_hash) = 32),
 
     -- Unix timestamp - when the block was mined (seconds since epoch).
-    timestamp BIGINT NOT NULL CHECK (timestamp >= 0)
+    block_timestamp BIGINT NOT NULL CHECK (block_timestamp >= 0)
 );
--- a/wallet/internal/db/queries/postgres/wallets.sql
+++ b/wallet/internal/db/queries/postgres/wallets.sql
@@ -1,6 +1,6 @@
 -- name: CreateWallet :one
 INSERT INTO wallets (
-    name,
+    wallet_name,
     is_imported,
     manager_version,
     is_watch_only,
@@ -20,30 +20,30 @@ INSERT INTO wallet_sync_states (
     birthday,
     updated_at
 ) VALUES (
-    $1, $2, $3, $4, CURRENT_TIMESTAMP
+    $1, $2, $3, $4, current_timestamp
 );
 
 -- name: UpdateWalletSyncState :execrows
 UPDATE wallet_sync_states
 SET
     -- If synced_height param is NOT NULL, use it. Otherwise, keep existing value.
-    synced_height = COALESCE(sqlc.narg('synced_height'), synced_height),
+    synced_height = coalesce(sqlc.narg('synced_height'), synced_height),
 
     -- If birthday_height param is NOT NULL, use it. Otherwise, keep existing value.
-    birthday_height = COALESCE(sqlc.narg('birthday_height'), birthday_height),
+    birthday_height = coalesce(sqlc.narg('birthday_height'), birthday_height),
 
     -- If birthday param is NOT NULL, use it. Otherwise, keep existing value.
-    birthday = COALESCE(sqlc.narg('birthday'), birthday),
+    birthday = coalesce(sqlc.narg('birthday'), birthday),
 
     -- Always update timestamp to current database time.
-    updated_at = CURRENT_TIMESTAMP
+    updated_at = current_timestamp
 WHERE
     wallet_id = $1;
 
 -- name: GetWalletByName :one
 SELECT
     w.id,
-    w.name,
+    w.wallet_name,
     w.is_imported,
     w.manager_version,
     w.is_watch_only,
@@ -52,19 +52,19 @@ SELECT
     s.birthday,
     s.updated_at,
     b_synced.header_hash AS synced_block_hash,
-    b_synced.timestamp AS synced_block_timestamp,
+    b_synced.block_timestamp AS synced_block_timestamp,
     b_birthday.header_hash AS birthday_block_hash,
-    b_birthday.timestamp AS birthday_block_timestamp
-FROM wallets w
-LEFT JOIN wallet_sync_states s ON s.wallet_id = w.id
-LEFT JOIN blocks b_synced ON s.synced_height = b_synced.block_height
-LEFT JOIN blocks b_birthday ON s.birthday_height = b_birthday.block_height
-WHERE w.name = $1;
+    b_birthday.block_timestamp AS birthday_block_timestamp
+FROM wallets AS w
+LEFT JOIN wallet_sync_states AS s ON w.id = s.wallet_id
+LEFT JOIN blocks AS b_synced ON s.synced_height = b_synced.block_height
+LEFT JOIN blocks AS b_birthday ON s.birthday_height = b_birthday.block_height
+WHERE w.wallet_name = $1;
 
 -- name: ListWallets :many
 SELECT
     w.id,
-    w.name,
+    w.wallet_name,
     w.is_imported,
     w.manager_version,
     w.is_watch_only,
@@ -73,19 +73,19 @@ SELECT
     s.birthday,
     s.updated_at,
     b_synced.header_hash AS synced_block_hash,
-    b_synced.timestamp AS synced_block_timestamp,
+    b_synced.block_timestamp AS synced_block_timestamp,
     b_birthday.header_hash AS birthday_block_hash,
-    b_birthday.timestamp AS birthday_block_timestamp
-FROM wallets w
-LEFT JOIN wallet_sync_states s ON s.wallet_id = w.id
-LEFT JOIN blocks b_synced ON s.synced_height = b_synced.block_height
-LEFT JOIN blocks b_birthday ON s.birthday_height = b_birthday.block_height
+    b_birthday.block_timestamp AS birthday_block_timestamp
+FROM wallets AS w
+LEFT JOIN wallet_sync_states AS s ON w.id = s.wallet_id
+LEFT JOIN blocks AS b_synced ON s.synced_height = b_synced.block_height
+LEFT JOIN blocks AS b_birthday ON s.birthday_height = b_birthday.block_height
 ORDER BY w.id;
 
 -- name: GetWalletByID :one
 SELECT
     w.id,
-    w.name,
+    w.wallet_name,
     w.is_imported,
     w.manager_version,
     w.is_watch_only,
@@ -94,13 +94,13 @@ SELECT
     s.birthday,
     s.updated_at,
     b_synced.header_hash AS synced_block_hash,
-    b_synced.timestamp AS synced_block_timestamp,
+    b_synced.block_timestamp AS synced_block_timestamp,
     b_birthday.header_hash AS birthday_block_hash,
-    b_birthday.timestamp AS birthday_block_timestamp
-FROM wallets w
-LEFT JOIN wallet_sync_states s ON s.wallet_id = w.id
-LEFT JOIN blocks b_synced ON s.synced_height = b_synced.block_height
-LEFT JOIN blocks b_birthday ON s.birthday_height = b_birthday.block_height
+    b_birthday.block_timestamp AS birthday_block_timestamp
+FROM wallets AS w
+LEFT JOIN wallet_sync_states AS s ON w.id = s.wallet_id
+LEFT JOIN blocks AS b_synced ON s.synced_height = b_synced.block_height
+LEFT JOIN blocks AS b_birthday ON s.birthday_height = b_birthday.block_height
 WHERE w.id = $1;
 
 -- name: InsertWalletSecrets :exec
--- a/wallet/internal/db/itest/fixtures_test.go
+++ b/wallet/internal/db/itest/fixtures_test.go
@@ -59,7 +59,7 @@ func CreateBlockFixture(t *testing.T, dbConn *sql.DB, height uint32) db.Block {
 	// TODO(gustavostingelin): use the block store to insert the block when
 	// available.
 	query := `
-		INSERT INTO blocks (block_height, header_hash, timestamp)
+		INSERT INTO blocks (block_height, header_hash, block_timestamp)
 		VALUES ($1, $2, $3)
 	`
 	_, err := dbConn.ExecContext(t.Context(), query,
--- a/wallet/internal/db/migrations/postgres/000002_wallets.up.sql
+++ b/wallet/internal/db/migrations/postgres/000002_wallets.up.sql
@@ -7,7 +7,7 @@ CREATE TABLE wallets (
     id BIGSERIAL PRIMARY KEY,
 
     -- Human friendly name for the wallet.
-    name TEXT NOT NULL,
+    wallet_name TEXT NOT NULL,
 
     -- Defines if the wallet was imported, all its accounts would also be imported.
     is_imported BOOLEAN NOT NULL,
@@ -30,7 +30,7 @@ CREATE TABLE wallets (
 );
 
 -- Unique index to prevent duplicate wallet names.
-CREATE UNIQUE INDEX uidx_wallets_name ON wallets (name);
+CREATE UNIQUE INDEX uidx_wallets_name ON wallets (wallet_name);
 
 -- Wallet Secrets table to store rarely accessed, highly sensitive encrypted
 -- material with a strict one-to-one relationship with the wallets table.
@@ -58,7 +58,7 @@ CREATE TABLE wallet_secrets (
 
     -- Foreign key constraint to wallet. Using ON DELETE RESTRICT to ensure
     -- that the wallet cannot be deleted if secrets still exist.
-    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT
+    FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT
 );
 
 -- Enforces one-to-one relationship: each wallet has at most one secrets record.
@@ -90,20 +90,20 @@ CREATE TABLE wallet_sync_states (
 
     -- Foreign key constraint to wallet. Using ON DELETE RESTRICT to ensure
     -- that the wallet cannot be deleted if sync state still exists.
-    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT,
+    FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT,
 
     -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to ensure
     -- that the block cannot be deleted if it is referenced by the sync state.
-    FOREIGN KEY (synced_height) REFERENCES blocks(block_height)
-        ON DELETE RESTRICT,
+    FOREIGN KEY (synced_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT,
 
     -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to ensure
     -- that the block cannot be deleted if it is referenced by the sync state.
-    FOREIGN KEY (birthday_height) REFERENCES blocks(block_height)
-        ON DELETE RESTRICT
+    FOREIGN KEY (birthday_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT
 );
 
 -- Enforces one-to-one relationship: each wallet has exactly one sync state record.
 -- Also serves as the effective primary key for this table.
 CREATE UNIQUE INDEX uidx_wallet_sync_states_wallet
-    ON wallet_sync_states (wallet_id);
+ON wallet_sync_states (wallet_id);
--- a/wallet/internal/db/migrations/sqlite/000001_blocks.up.sql
+++ b/wallet/internal/db/migrations/sqlite/000001_blocks.up.sql
@@ -13,5 +13,5 @@ CREATE TABLE blocks (
     header_hash BLOB NOT NULL UNIQUE CHECK (length(header_hash) = 32),
 
     -- Unix timestamp - when the block was mined (seconds since epoch).
-    timestamp INTEGER NOT NULL CHECK (timestamp >= 0)
+    block_timestamp INTEGER NOT NULL CHECK (block_timestamp >= 0)
 );
--- a/wallet/internal/db/migrations/sqlite/000002_wallets.up.sql
+++ b/wallet/internal/db/migrations/sqlite/000002_wallets.up.sql
@@ -7,7 +7,7 @@ CREATE TABLE wallets (
     id INTEGER PRIMARY KEY,
 
     -- Human friendly name for the wallet.
-    name TEXT NOT NULL,
+    wallet_name TEXT NOT NULL,
 
     -- Defines if the wallet was imported, all its accounts would also be imported.
     is_imported BOOLEAN NOT NULL,
@@ -30,7 +30,7 @@ CREATE TABLE wallets (
 );
 
 -- Unique index to prevent duplicate wallet names.
-CREATE UNIQUE INDEX uidx_wallets_name ON wallets (name);
+CREATE UNIQUE INDEX uidx_wallets_name ON wallets (wallet_name);
 
 -- Wallet Secrets table to store rarely accessed, highly sensitive encrypted
 -- material with a strict one-to-one relationship with the wallets table.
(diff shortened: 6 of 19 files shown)


mohamedawnallah:
@GustavoStingelin, Could you please fix the linting issues for SQL files that introduced in this PR https://github.com/btcsuite/btcwallet/pull/1125?

CI: https://github.com/btcsuite/btcwallet/actions/runs/20141579623/job/57810339472?pr=1121

You can push changes to this PR

GustavoStingelin:
@mohamedawnallah pushed the fix

wallet/internal/db/migrations/postgres/000001_blocks.up.sql
```diff
@@ -13,5 +13,5 @@ CREATE TABLE blocks (
     header_hash BYTEA NOT NULL UNIQUE CHECK (length(header_hash) = 32),
 
     -- Unix timestamp - when the block was mined (seconds since epoch).
-    timestamp BIGINT NOT NULL CHECK (timestamp >= 0)
+    created_at BIGINT NOT NULL CHECK (created_at >= 0)
```
GustavoStingelin: As the new linter prevents us from using reserved words like `timestamp`, a reasonable approach in cases like this is to prefix the original column with something tied to the table name, for example `block_timestamp`, or another term that keeps the intent clear without relying on a generic prefix.

I do not think `created_at` is a good fit here. It can confuse the meaning of the field because a block time is not the moment when the block was created or mined. It is only the header time chosen by the miner, and it can even end up lower than the time of the previous block, which shows how loosely this value relates to real time.

config/.golangci.yml
GustavoStingelin, line 1: I appreciate the effort here. From my perspective, it would be tidier to keep this sort of configuration file in the project root. The root is where most developers expect to find these files, and this particular configuration applies to the entire project rather than to a single subpackage. I tend to use a configs directory for runtime settings such as environment variables, dynamic configuration loaders, or credential management. The project layout described in the [golang‑standards](https://github.com/golang-standards/project-layout?tab=readme-ov-file#configs) repository treats configs this way; while we don’t have to follow that layout rigidly, it can be a helpful convention. Regarding SQLC.yaml, it seems sensible to place it near the SQL sources it operates on so the relationship is clear. Let me know if you see it differently or have strong feelings about preferring this PR structure.

GustavoStingelin (changes requested):
Looks amazing, the SQLFluff parts. I just left two comments regarding the refactoring.

config/.golangci.yml
mohamedawnallah: Acknowledged. Let's take @yyforyongyu's take on it as well

wallet/internal/db/migrations/postgres/000001_blocks.up.sql
mohamedawnallah: Acknowledged. Let's take @yyforyongyu's take on it as well

wallet/internal/db/queries/postgres/wallets.sql
```diff
@@ -55,16 +55,16 @@ SELECT
     b_synced.created_at AS synced_block_timestamp,
     b_birthday.header_hash AS birthday_block_hash,
     b_birthday.created_at AS birthday_block_timestamp
-FROM wallets w
-LEFT JOIN wallet_sync_states s ON s.wallet_id = w.id
-LEFT JOIN blocks b_synced ON s.synced_height = b_synced.block_height
-LEFT JOIN blocks b_birthday ON s.birthday_height = b_birthday.block_height
-WHERE w.name = $1;
+FROM wallets AS w
```
yyforyongyu, line 58: Q: is this generated from the new auto fix? like `make sql-lint`?

wallet/internal/db/migrations/postgres/000001_blocks.up.sql
yyforyongyu: yeah agreed - think `block_timestamp` is better

config/.golangci.yml
yyforyongyu: I prefer it this way, rather than having multiple yml files sitting in the root, plus this is not [new](https://gist.github.com/ayoubzulfiqar/9f1a34049332711fddd4d4b2bfd46096). 

> The project layout described in the [golang‑standards](https://github.com/golang-standards/project-layout?tab=readme-ov-file#configs) repository treats configs this way;

I think it's pretty vague about this, as it says,
> /configs
>
>Configuration file templates or default configs.
>
>Put your confd or consul-template template files here.

config/testing_flags.mk
```diff
@@ -72,7 +76,7 @@ UNIT_RACE := $(GOTEST) -tags="$(DEV_TAGS) $(LOG_TAGS)" $(TEST_FLAGS) -race $(UNI
 
 # NONE is a special value which selects no other tests but only executes the
 # benchmark tests here.
-UNIT_BENCH := $(GOTEST) -tags="$(DEV_TAGS) $(LOG_TAGS)" -test.bench=. -test.run=NONE $(UNITPKG)
+UNIT_BENCH := $(GOTEST) -tags="$(DEV_TAGS) $(LOG_TAGS)" $(TEST_FLAGS) -test.bench=. -test.run=NONE $(UNITPKG)
```
yyforyongyu, line 79: 👍

Makefile
```diff
@@ -206,8 +269,12 @@ sqlc-check: sqlc
 	fmt-check \
 	tidy-module \
 	tidy-module-check \
+	sql-parse \
 	sqlc \
 	sqlc-check \
+	sql-format \
+	sql-lint \
+	sql-lint-check \
```
yyforyongyu, line 277: We are still missing `sql-format-check`

Makefile
```diff
@@ -181,16 +195,65 @@ tidy-module:
...
 	$(DOCKER_TOOLS) sqlc generate -f config/sqlc.yaml
 
-#? sqlc-check: Make sure sql models and queries are up to date
+#? sqlc-check: Verify generated Go SQL queries and migrations are up-to-date
 sqlc-check: sqlc
 	@$(call print, "Verifying sql code generation.")
 	if test -n "$$(git status --porcelain '*.go')"; then echo "SQL models not properly generated!"; git status --porcelain '*.go'; exit 1; fi
 
+#? sql-format: Format SQL migration and query files (like 'make fmt')
+sql-format: docker-tools
```
yyforyongyu: we don't need `docker-tools` here?

Makefile
```diff
@@ -181,16 +195,65 @@ tidy-module:
...
+	@$(call print, "Formatting SQL files (postgres migrations).")
+	$(SQLFLUFF) format --config /sql/config/sqlfluff.cfg --dialect postgres $(SQL_POSTGRES_MIGRATIONS)
```
yyforyongyu, line 222: https://github.com/btcsuite/btcwallet/pull/1134#discussion_r2610464334

Makefile
```diff
@@ -181,16 +195,65 @@ tidy-module:
...
+#? sql-check: Verify SQL migration and query files are formatted correctly (like 'make fmt-check')
+sql-format-check: sql-format
+	@$(call print, "Checking SQL formatting.")
+	if test -n "$$(git status --porcelain '$(SQL_MIGRATIONS_DIR)/**/*.sql' '$(SQL_QUERIES_DIR)/**/*.sql')"; then echo "SQL files not formatted correctly, please run 'make sql-format' again!"; git status; git diff; exit 1; fi
+
+#? sql-lint: Lint SQL migration and query files and fix issues (like 'make lint')
+sql-lint: docker-tools
+	@$(call print, "Linting SQL files (postgres migrations).")
+	$(SQLFLUFF) fix --config /sql/config/sqlfluff.cfg --dialect postgres $(SQL_POSTGRES_MIGRATIONS)
+	@$(call print, "Linting SQL files (postgres queries).")
```
yyforyongyu, line 239: https://github.com/btcsuite/btcwallet/pull/1134#discussion_r2610464334

config/.golangci.yml
GustavoStingelin: Even in the [“this is not new”](https://gist.github.com/ayoubzulfiqar/9f1a34049332711fddd4d4b2bfd46096) reference, you can see the configs under configs/ are focused on _runtime environment variables_, not tooling like linters. For example:
```
    ├── configs/                 # Configuration files for different environments
    │   ├── development.yaml
    │   ├── production.yaml
```
I feel like we’d be moving from a well-understood convention to something less meaningful by relocating `.golangci.yml`.

My IDE, and probably yours too, rely on `.golangci.yml` to show real-time lint feedback while coding. Moving it into a subdirectory could introduce friction in the dev setup.

Personally, I rarely use the Docker setup to run the linter. It's too heavy. It takes around 15 seconds to run. A plain `golangci-lint run -v` runs in under 1 second, or about 6 seconds with a cold cache. I would rather manage different build versions myself than deal with that. If we move the config, we’ll need to run it like `golangci-lint run -v --config config/.golangci.yml`.

That said, if we’re assuming users shouldn’t use the linter natively in their editor and instead rely entirely on Docker, wouldn’t it make more sense to move all these tool configs into something like a `/tools `folder? Since they’re being treated as docker-only assets anyway.

Also, just a small note: if we really want to isolate config files, using `.config/ `(dot-prefixed) might be better than `config/` to avoid being understood as a Go package. This is even mentioned in [golangci-lint issue #4554](https://github.com/golangci/golangci-lint/issues/4554)

Lastly, about the `.mk` file. I think it’d make sense to place it in a `scripts/` or `build/` folder. Or we could stick with a `make/` folder like other Lightning Labs projects do, which feels like a good standard to follow, at this time.

Makefile
mohamedawnallah: > #1134 (comment)

`fix` tries to fix both linting (can be non-fixable) and non-linting errors (e.g indentation, etc) while `format` tries fix only non-linting errors which always fixable. A reproducible experiment I did is I run `make sql-format` in the commit where I added sqlfluff metadev helpers (b7b87f0bf2673f370fc51a339f5c463c37ef443d) and before the linting fixing commit someone will see diffs like the following:

<details>
<summary> sql-format-diff (format) </summary>

```diff
diff --git a/wallet/internal/db/migrations/postgres/000002_wallets.up.sql b/wallet/internal/db/migrations/postgres/000002_wallets.up.sql
index 6bf74e4a..a3d17320 100644
--- a/wallet/internal/db/migrations/postgres/000002_wallets.up.sql
+++ b/wallet/internal/db/migrations/postgres/000002_wallets.up.sql
@@ -58,7 +58,7 @@ CREATE TABLE wallet_secrets (

     -- Foreign key constraint to wallet. Using ON DELETE RESTRICT to ensure
     -- that the wallet cannot be deleted if secrets still exist.
-    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT
+    FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT
 );

 -- Enforces one-to-one relationship: each wallet has at most one secrets record.
@@ -90,20 +90,20 @@ CREATE TABLE wallet_sync_states (

     -- Foreign key constraint to wallet. Using ON DELETE RESTRICT to ensure
     -- that the wallet cannot be deleted if sync state still exists.
-    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT,
+    FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT,

     -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to ensure
     -- that the block cannot be deleted if it is referenced by the sync state.
-    FOREIGN KEY (synced_height) REFERENCES blocks(block_height)
-        ON DELETE RESTRICT,
+    FOREIGN KEY (synced_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT,

     -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to ensure
     -- that the block cannot be deleted if it is referenced by the sync state.
-    FOREIGN KEY (birthday_height) REFERENCES blocks(block_height)
-        ON DELETE RESTRICT
+    FOREIGN KEY (birthday_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT
 );

 -- Enforces one-to-one relationship: each wallet has exactly one sync state record.
 -- Also serves as the effective primary key for this table.
 CREATE UNIQUE INDEX uidx_wallet_sync_states_wallet
-    ON wallet_sync_states (wallet_id);
+ON wallet_sync_states (wallet_id);
diff --git a/wallet/internal/db/migrations/sqlite/000002_wallets.up.sql b/wallet/internal/db/migrations/sqlite/000002_wallets.up.sql
index eebcd592..5bc5996d 100644
--- a/wallet/internal/db/migrations/sqlite/000002_wallets.up.sql
+++ b/wallet/internal/db/migrations/sqlite/000002_wallets.up.sql
@@ -58,7 +58,7 @@ CREATE TABLE wallet_secrets (

     -- Foreign key constraint to wallet. Using ON DELETE RESTRICT to ensure
     -- that the wallet cannot be deleted if secrets still exist.
-    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT
+    FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT
 );

 -- Enforces one-to-one relationship: each wallet has at most one secrets record.
@@ -90,20 +90,20 @@ CREATE TABLE wallet_sync_states (

     -- Foreign key constraint to wallet. Using ON DELETE RESTRICT to ensure
     -- that the wallet cannot be deleted if sync state still exists.
-    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT,
+    FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT,

     -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to ensure
     -- that the block cannot be deleted if it is referenced by the sync state.
-    FOREIGN KEY (synced_height) REFERENCES blocks(block_height)
-        ON DELETE RESTRICT,
+    FOREIGN KEY (synced_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT,

     -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to ensure
     -- that the block cannot be deleted if it is referenced by the sync state.
-    FOREIGN KEY (birthday_height) REFERENCES blocks(block_height)
-        ON DELETE RESTRICT
+    FOREIGN KEY (birthday_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT
 );

 -- Enforces one-to-one relationship: each wallet has exactly one sync state record.
 -- Also serves as the effective primary key for this table.
 CREATE UNIQUE INDEX uidx_wallet_sync_states_wallet
-    ON wallet_sync_states (wallet_id);
+ON wallet_sync_states (wallet_id);
diff --git a/wallet/internal/db/queries/postgres/blocks.sql b/wallet/internal/db/queries/postgres/blocks.sql
index dc9884ba..aa772840 100644
--- a/wallet/internal/db/queries/postgres/blocks.sql
+++ b/wallet/internal/db/queries/postgres/blocks.sql
@@ -1,5 +1,8 @@
 -- name: GetBlockByHeight :one
-SELECT block_height, header_hash, timestamp
+SELECT
+    block_height,
+    header_hash,
+    timestamp
 FROM blocks
 WHERE block_height = $1;

diff --git a/wallet/internal/db/queries/postgres/wallets.sql b/wallet/internal/db/queries/postgres/wallets.sql
index 248f3b98..f57e53cd 100644
--- a/wallet/internal/db/queries/postgres/wallets.sql
+++ b/wallet/internal/db/queries/postgres/wallets.sql
@@ -20,23 +20,23 @@ INSERT INTO wallet_sync_states (
     birthday,
     updated_at
 ) VALUES (
-    $1, $2, $3, $4, CURRENT_TIMESTAMP
+    $1, $2, $3, $4, current_timestamp
 );

 -- name: UpdateWalletSyncState :execrows
 UPDATE wallet_sync_states
 SET
     -- If synced_height param is NOT NULL, use it. Otherwise, keep existing value.
-    synced_height = COALESCE(sqlc.narg('synced_height'), synced_height),
+    synced_height = coalesce(sqlc.narg('synced_height'), synced_height),

     -- If birthday_height param is NOT NULL, use it. Otherwise, keep existing value.
-    birthday_height = COALESCE(sqlc.narg('birthday_height'), birthday_height),
+    birthday_height = coalesce(sqlc.narg('birthday_height'), birthday_height),

     -- If birthday param is NOT NULL, use it. Otherwise, keep existing value.
-    birthday = COALESCE(sqlc.narg('birthday'), birthday),
+    birthday = coalesce(sqlc.narg('birthday'), birthday),

     -- Always update timestamp to current database time.
-    updated_at = CURRENT_TIMESTAMP
+    updated_at = current_timestamp
 WHERE
     wallet_id = $1;

diff --git a/wallet/internal/db/queries/sqlite/blocks.sql b/wallet/internal/db/queries/sqlite/blocks.sql
index f37744df..c0d574b9 100644
--- a/wallet/internal/db/queries/sqlite/blocks.sql
+++ b/wallet/internal/db/queries/sqlite/blocks.sql
@@ -1,5 +1,8 @@
 -- name: GetBlockByHeight :one
-SELECT block_height, header_hash, timestamp
+SELECT
+    block_height,
+    header_hash,
+    timestamp
 FROM blocks
 WHERE block_height = ?;

diff --git a/wallet/internal/db/queries/sqlite/wallets.sql b/wallet/internal/db/queries/sqlite/wallets.sql
index 0259e880..356112f9 100644
--- a/wallet/internal/db/queries/sqlite/wallets.sql
+++ b/wallet/internal/db/queries/sqlite/wallets.sql
@@ -20,23 +20,23 @@ INSERT INTO wallet_sync_states (
     birthday,
     updated_at
 ) VALUES (
-    ?, ?, ?, ?, CURRENT_TIMESTAMP
+    ?, ?, ?, ?, current_timestamp
 );

 -- name: UpdateWalletSyncState :execrows
 UPDATE wallet_sync_states
 SET
     -- If synced_height param is NOT NULL, use it. Otherwise, keep existing value.
-    synced_height = COALESCE(sqlc.narg('synced_height'), synced_height),
+    synced_height = coalesce(sqlc.narg('synced_height'), synced_height),

     -- If birthday_height param is NOT NULL, use it. Otherwise, keep existing value.
-    birthday_height = COALESCE(sqlc.narg('birthday_height'), birthday_height),
+    birthday_height = coalesce(sqlc.narg('birthday_height'), birthday_height),

     -- If birthday param is NOT NULL, use it. Otherwise, keep existing value.
-    birthday = COALESCE(sqlc.narg('birthday'), birthday),
+    birthday = coalesce(sqlc.narg('birthday'), birthday),

     -- Always update timestamp to current database time.
-    updated_at = CURRENT_TIMESTAMP
+    updated_at = current_timestamp
 WHERE
     wallet_id = sqlc.arg('wallet_id');
```
</details>
<details>
<summary> sql-lint-diff (fix) </summary>

```diff
diff --git a/wallet/internal/db/migrations/postgres/000002_wallets.up.sql b/wallet/internal/db/migrations/postgres/000002_wallets.up.sql
index 6bf74e4a..a3d17320 100644
--- a/wallet/internal/db/migrations/postgres/000002_wallets.up.sql
+++ b/wallet/internal/db/migrations/postgres/000002_wallets.up.sql
@@ -58,7 +58,7 @@ CREATE TABLE wallet_secrets (

     -- Foreign key constraint to wallet. Using ON DELETE RESTRICT to ensure
     -- that the wallet cannot be deleted if secrets still exist.
-    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT
+    FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT
 );

 -- Enforces one-to-one relationship: each wallet has at most one secrets record.
@@ -90,20 +90,20 @@ CREATE TABLE wallet_sync_states (

     -- Foreign key constraint to wallet. Using ON DELETE RESTRICT to ensure
     -- that the wallet cannot be deleted if sync state still exists.
-    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT,
+    FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT,

     -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to ensure
     -- that the block cannot be deleted if it is referenced by the sync state.
-    FOREIGN KEY (synced_height) REFERENCES blocks(block_height)
-        ON DELETE RESTRICT,
+    FOREIGN KEY (synced_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT,

     -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to ensure
     -- that the block cannot be deleted if it is referenced by the sync state.
-    FOREIGN KEY (birthday_height) REFERENCES blocks(block_height)
-        ON DELETE RESTRICT
+    FOREIGN KEY (birthday_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT
 );

 -- Enforces one-to-one relationship: each wallet has exactly one sync state record.
 -- Also serves as the effective primary key for this table.
 CREATE UNIQUE INDEX uidx_wallet_sync_states_wallet
-    ON wallet_sync_states (wallet_id);
+ON wallet_sync_states (wallet_id);
```

</details>


As can be seen `make sql-format` always succeed but `make sql-lint` fails on the first non-fixable lint.

It would be nice to have both like existing workflows `make lint` and `make fmt`

Makefile
mohamedawnallah: Replied in the above comment

wallet/internal/db/queries/postgres/wallets.sql
mohamedawnallah: > Q: is this generated from the new auto fix? like `make sql-lint`?

Yes, that's right. It is generated from `make sql-lint` and also can be generated from `make sql-format`

Makefile
mohamedawnallah: Addressed

Makefile
mohamedawnallah: Yes, that's right. Removed

wallet/internal/db/migrations/postgres/000001_blocks.up.sql
mohamedawnallah: Addressed

mohamedawnallah (commented):
> what is the order to run these commands? and how can we enforce that? If we run make sqlc first, then make sql-lint, does it mean we need to run make sqlc again to include the changes by the linter?

There are currently to sequentail phases:
- Making sure the SQL queries/migrations are compilable. This is ensured by parsing the AST the one provided out of the box by sqlfluff  and then those SQL code can be converted to the equivalent Golang code. This phase is achieved by `make sqlc`
- Making sure the SQL queries/migrations are properly well-formated/linted. This is achieved by `make sql-lint` (better for CI can fail for non-fixable lints) or `make sql-format` (better for local development since it is tied only to formatting issues and always pass)

This is currently enforced. Since if the user applied the linting issues proposed by sqlfluff e.g changing column name, the CI would fail since the new compiled code by `make sqlc` will no longer have that old column name that is currently referenced in the codebase.

The lack of one tool that handles linting/formatting/compilation resulted from the fact that Golang doesn't have native support for SQL files. it kinda depends on other third-party community tools like `sqlc`

config/.golangci.yml
yyforyongyu: > Personally, I rarely use the Docker setup to run the linter. It's too heavy. It takes around 15 seconds to run.

hmm so maybe this is why this happened? https://github.com/btcsuite/btcwallet/pull/1125#discussion_r2607641452

> My IDE, and probably yours too, rely on .golangci.yml to show real-time lint feedback while coding. 

Make sense - LSPs do look for these files, like `.golangci.yml` or `.protolint.yml`, tho I really like the idea of having one place to host all the `.yml` files, which are used to configure tools...

> if we really want to isolate config files, using .config/ (dot-prefixed) might be better than config/ to avoid being understood as a Go package. 

Yeah think it works - I prolly would move `.golangci.yml` and `.protolint.yml` back to root, and move everything into `.config`; Or, we can just keep the files at their existing locations. How's that @GustavoStingelin ?

Makefile
yyforyongyu: > As can be seen make sql-format always succeed but make sql-lint fails on the first non-fixable lint.

Cool yeah make sense to keep both.

Makefile
```diff
@@ -24,6 +35,13 @@ DOCKER_TOOLS = docker run \
   -v $(shell bash -c "mkdir -p /tmp/go-build-cache; echo /tmp/go-build-cache"):/root/.cache/go-build \
   -v $$(pwd):/build btcwallet-tools
 
+SQLFLUFF = docker run \
+	--rm \
+	--user $$(id -u):$$(id -g) \
+    -v $$(pwd):/sql \
+    -w /sql \
+    sqlfluff/sqlfluff
```
yyforyongyu, line 43: We prolly should ping the version here

config/.golangci.yml
GustavoStingelin: > hmm so maybe this is why this happened? https://github.com/btcsuite/btcwallet/pull/1125#discussion_r2607641452

No, in that case it was just a misunderstanding of what the new line represented there. I saw the same behavior in Docker as well. 

> Yeah think it works - I prolly would move .golangci.yml and .protolint.yml back to root, and move everything into .config; Or, we can just keep the files at their existing locations. How's that @GustavoStingelin ?

I am approving the PR so we can merge it quickly, but that does not mean I agree with this change.

Makefile
mohamedawnallah: Addressed in 6b342fddb9fc2a59ffe8f191d6ab0d78568efd8a

Issue #1103, opened 2025-10-20
[sqldb/address]: Implement wallet and sync state support


GustavoStingelin:
Create migrations for wallets, wallet_secrets, and wallet_sync_states (ref #1051)

Provide SQL queries for inserting and reading these tables

Run SQLC to generate Go models and query wrappers

Implement the methods defined in the interfaces

PR #1125, opened 2025-11-25
wallet: implement WalletStore interface across PG and SQLite

GustavoStingelin:
## Change Description
This PR aims to implement the WalletStore interface across PG and SQLite.
#1103 

## Steps to Test
```
make unit
```

```
 btcwallet git:(sqldb/add-wallet) make itest-db db=sqlite nocache=1 verbose=1
 Running SQLite integration tests.
go list -tags="itest dev " -deps github.com/btcsuite/btcwallet/... | grep 'github.com/btcsuite/btcwallet' | grep '/itest$' | xargs -L 1 env GO111MODULE=on go test -tags="itest dev debug stdlog "  -test.timeout=20m -test.v -test.count=1
=== RUN   TestCreateWallet
=== PAUSE TestCreateWallet
=== RUN   TestCreateWallet_DuplicateName
...
ok      github.com/btcsuite/btcwallet/wallet/internal/db/itest  0.255s
```

```
 btcwallet git:(sqldb/add-wallet) make itest-db db=postgres nocache=1 verbose=1
 Running PostgreSQL integration tests.
go list -tags="itest dev  test_db_postgres" -deps github.com/btcsuite/btcwallet/... | grep 'github.com/btcsuite/btcwallet' | grep '/itest$' | xargs -L 1 env GO111MODULE=on go test -tags="itest dev debug stdlog  test_db_postgres"  -test.timeout=20m -test.v -test.count=1
=== RUN   TestCreateWallet
...
=== CONT  TestCreateWallet_Variants/imported_wallet
2025/12/01 17:50:33 github.com/testcontainers/testcontainers-go - Connected to docker: 
  Server Version: 28.5.2
  API Version: 1.51
  Operating System: OrbStack
  Total Memory: 12006 MB
  Testcontainers for Go Version: v0.40.0
  Resolved Docker Host: unix:///var/run/docker.sock
  Resolved Docker Socket Path: /var/run/docker.sock
  Test SessionID: 6de082e185166b7f31532ce8f1ba4b776ba33a83ad003b6e7e5f2ca6c060bc67
  Test ProcessID: 9835bcf9-e5a1-4971-986b-0bc8838a509e
2025/12/01 17:50:33 🐳 Creating container for image postgres:18-alpine
2025/12/01 17:50:34 🐳 Creating container for image testcontainers/ryuk:0.13.0
2025/12/01 17:50:34 ✅ Container created: 1513089a5bab
2025/12/01 17:50:34 🐳 Starting container: 1513089a5bab
2025/12/01 17:50:34 ✅ Container started: 1513089a5bab
2025/12/01 17:50:34 ⏳ Waiting for container id 1513089a5bab image: testcontainers/ryuk:0.13.0. Waiting for: port 8080/tcp to be listening
2025/12/01 17:50:34 Shell not found in container
2025/12/01 17:50:34 🔔 Container is ready: 1513089a5bab
2025/12/01 17:50:34 ✅ Container created: b7254208795e
2025/12/01 17:50:34 🐳 Starting container: b7254208795e
2025/12/01 17:50:34 ✅ Container started: b7254208795e
2025/12/01 17:50:34 ⏳ Waiting for container id b7254208795e image: postgres:18-alpine. Waiting for: all of: [port 5432/tcp to be listening]
2025/12/01 17:50:35 🔔 Container is ready: b7254208795e
--- PASS: TestGetEncryptedHDSeed (2.05s)
=== CONT  TestCreateWallet_Variants/watch-only_wallet
--- PASS: TestGetWallet_NotFound (2.05s)
=== CONT  TestCreateWallet_Variants/standard_wallet
--- PASS: TestUpdateWalletSecrets (2.06s)
...
--- PASS: TestCreateWallet_Variants (0.00s)
    --- PASS: TestCreateWallet_Variants/imported_wallet (2.07s)
    --- PASS: TestCreateWallet_Variants/watch-only_wallet (0.06s)
    --- PASS: TestCreateWallet_Variants/standard_wallet (0.06s)
PASS
2025/12/01 17:50:35 🐳 Stopping container: b7254208795e
2025/12/01 17:50:36 ✅ Container stopped: b7254208795e
2025/12/01 17:50:36 🐳 Terminating container: b7254208795e
2025/12/01 17:50:36 🚫 Container terminated: b7254208795e
ok      github.com/btcsuite/btcwallet/wallet/internal/db/itest  2.730s
```


mohamedawnallah (commented):
The PR approach looks good. If there are only three comments, they would be:
- Currently we do repeat implementation N times where is N is the number of unique SQL databases supported. This is not ideal nor necessary desired. Would it be impossible to have a abstraction layer on serializing/deserializing for the slight differences between different SQL databases? That way we only write it once
- How can we make sure that the current SQLized version in that PR would be equivalent to KV store and work properly in whole context of the application?
- Would you mind putting the itests in the same PR? That way the PR would be self-contained and gives us +1 of confidence that those functionally can be well-integrated

GustavoStingelin:
Hey @mohamedawnallah tks for taking a look!

>Currently we do repeat implementation N times where is N is the number of unique SQL databases supported. This is not ideal nor necessary desired. Would it be impossible to have a abstraction layer on serializing/deserializing for the slight differences between different SQL databases? That way we only write it once

Yeahhh, I also was not very happy with the repetition, but I could not find a way to keep things simple and avoid confusion at the same time. Every attempt to reduce the repetition turned into abstractions that were harder to follow. I think we can still look for a cleaner approach, but for now this version stays readable.

>How can we make sure that the current SQLized version in that PR would be equivalent to KV store and work properly in whole context of the application?

I think the next stages will reveal that. We will probably need to adjust some parts or extend the API as we integrate the pieces.


> Would you mind putting the itests in the same PR? That way the PR would be self-contained and gives us +1 of confidence that those functionally can be well-integrated

Added. It does increase the review work a bit, although it also helps understand the overall behavior and confirm that the code holds up.

wallet/internal/db/wallet_sqlite.go
```diff
@@ -0,0 +1,392 @@
+package db
+
```
mohamedawnallah, line 2: > Yeahhh, I also was not very happy with the repetition, but I could not find a way to keep things simple and avoid confusion at the same time. Every attempt to reduce the repetition turned into abstractions that were harder to follow. I think we can still look for a cleaner approach, but for now this version stays readable.

We can use adapter interface. Here is a concrete example for the blocks schema (seeing block_store.go):
https://github.com/btcsuite/btcwallet/pull/1134/files#diff-9ebd23f8506ae291b28acb39488af16d77bbe0029db94999f513aa8825334aedR51

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
```diff
@@ -0,0 +1,69 @@
+-- Wallet metadata and related state tables.
+--
+-- Migration note: Intentionally NOT idempotent (no "IF NOT EXISTS").
+-- This ensures migration tracking stays accurate and fails loudly if run twice.
+CREATE TABLE wallets (
+    -- DB ID of the wallet, primary key. Only used for DB level relations.
+    id BIGSERIAL PRIMARY KEY,
+    -- Human friendly name for the wallet.
+    name TEXT NOT NULL,
+    -- Defines if the wallet was imported, so all accounts should be imported.
```
yyforyongyu: You mean when the wallet is imported, all its accounts are also considered imported right?


```suggestion
    -- Defines if the wallet was imported, all its accounts would also be imported.
```

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
```diff
@@ -0,0 +1,69 @@
+-- Wallet metadata and related state tables.
+--
+-- Migration note: Intentionally NOT idempotent (no "IF NOT EXISTS").
+-- This ensures migration tracking stays accurate and fails loudly if run twice.
+CREATE TABLE wallets (
+    -- DB ID of the wallet, primary key. Only used for DB level relations.
+    id BIGSERIAL PRIMARY KEY,
```
yyforyongyu, line 7: nit: I think we should add a newline after every field, to align the format we've been using in other projects.

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
```diff
@@ -0,0 +1,69 @@
...
+    wallet_id BIGINT NOT NULL,
+    -- Params to derive the private master key.
+    master_priv_params BYTEA,
+    -- Encrypted key used to encrypt/decrypt wallet data related to private keys.
+    encrypted_crypto_priv_key BYTEA,
+    -- Encrypted key used to encrypt/decrypt wallet data related to scripts.
+    encrypted_crypto_script_key BYTEA,
+    -- Encrypted HD private key of the wallet.
+    encrypted_master_hd_priv_key BYTEA,
+    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT
```
yyforyongyu, line 61: Nice - add a few sentence to explain why `RESTRICT`? we need more verbose comments in this file such that a future dev can easily catch up and understand the design rationales.

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
```diff
@@ -0,0 +1,69 @@
...
+    wallet_id BIGINT NOT NULL,
+    -- Current sync status of the wallet (references blocks table).
+    synced_height INTEGER,
+    -- Birthday block height of the wallet (references blocks table).
+    birthday_height INTEGER,
+    -- Indicates if the birthday block has been verified.
+    birthday_verified BOOLEAN NOT NULL,
+    -- Last updated timestamp stored in UTC without timezone info.
+    updated_at TIMESTAMP NOT NULL,
+    FOREIGN KEY (wallet_id) REFERENCES wallets(id) ON DELETE RESTRICT,
```
yyforyongyu, line 93: same here regarding more verbose comments

wallet/internal/db/migrations/sqlite/000002_wallets.up.sql
```diff
@@ -0,0 +1,69 @@
+-- Wallet metadata and related state tables.
+--
+-- Migration note: Intentionally NOT idempotent (no "IF NOT EXISTS").
+-- This ensures migration tracking stays accurate and fails loudly if run twice.
+CREATE TABLE wallets (
+    -- DB ID of the wallet, primary key. Only used for DB level relations.
+    id INTEGER PRIMARY KEY,
+    -- Human friendly name for the wallet.
+    name TEXT NOT NULL,
+    -- Defines if the wallet was imported, so all accounts should be imported.
```
yyforyongyu: same in this file re the comments made in postgres

wallet/internal/db/block_pg.go
```diff
@@ -0,0 +1,32 @@
...
+	"time"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+)
+
+// buildPgBlock constructs a Block from the given PostgreSQL block
+// fields.
+func buildPgBlock(
+	height sql.NullInt32, hash []byte, timestamp sql.NullInt64,
+) (*Block, error) {
```
yyforyongyu: nit: I guess this is undefined format [here](https://github.com/btcsuite/btcwallet/blob/master/docs/developer/code_formatting_rules.md#wrapping-long-function-definitions), maybe we should update the docs if we decide to allow/forbid this format. And normally we would do, 
```go
func buildPgBlock(height sql.NullInt32, hash []byte, 
	timestamp sql.NullInt64) (*Block, error) {
```
same for the rest

wallet/internal/db/block_sqlite.go
```diff
@@ -0,0 +1,32 @@
...
+	"fmt"
+	"time"
+
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+)
+
+// buildSqliteBlock constructs a Block from the given SQLite block
+// fields.
+func buildSqliteBlock(
+	height sql.NullInt64, hash []byte, timestamp sql.NullInt64,
```
yyforyongyu: looks like the only diff from `buildPgBlock` is the `height` type? I think we can just use type constraint, sth like [this](https://github.com/lightningnetwork/lnd/blob/20473482d45f6cb5ef45ee4b80b6505555d9cc1f/tlv/primitive.go#L379).

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,392 @@
...
+func (w *PostgresWalletDB) GetWallet(
+	ctx context.Context, name string,
+) (*WalletInfo, error) {
+
+	row, err := w.queries.GetWalletByName(ctx, name)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, fmt.Errorf(
+				"wallet %q: %w",
+				name, ErrWalletNotFound,
```
yyforyongyu: 👍

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,392 @@
...
+			return nil, fmt.Errorf(
+				"secrets for wallet %d: %w",
+				walletID, ErrWalletNotFound,
+			)
+		}
+
+		return nil, fmt.Errorf("get wallet secrets: %w", err)
+	}
+
+	return secrets.EncryptedMasterHdPrivKey, nil
```
yyforyongyu, line 258: will this value be `[]byte` or `nil` for imported wallets?

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,392 @@
...
+	syncedBlockHash        []byte
+	syncedBlockTimestamp   sql.NullInt64
+	birthdayHeight         sql.NullInt32
+	birthdayBlockHash      []byte
+	birthdayBlockTimestamp sql.NullInt64
+}
+
+// buildPgWalletInfo constructs a WalletInfo from the given wallet row
+// parameters.
+func buildPgWalletInfo(params pgWalletRowParams) (*WalletInfo, error) {
```
yyforyongyu: nitty-nit: maybe name it `rows` so it's even clearer that this is converting db data to struct data here?

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,392 @@
...
+		birthdayVerified = true
+	}
+
+	// Upsert the sync state.
+	return sqlcpg.UpsertWalletSyncStateParams{
+		WalletID:         int64(params.WalletID),
+		SyncedHeight:     syncedHeight,
+		BirthdayHeight:   birthdayHeight,
+		BirthdayVerified: birthdayVerified,
+		UpdatedAt:        time.Now().UTC(),
```
yyforyongyu: Given that `UpdatedAt` is always the current time, this is where the splitting sql schema approaches shines - we can just define the updated_at in the query instead, let the db handle the auto-updating timestamps, which is not feasible previously due to the syntax are different.

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,392 @@
...
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return fmt.Errorf(
+				"wallet %d: %w",
+				params.WalletID, ErrWalletNotFound,
+			)
+		}
+
+		return fmt.Errorf(
+			"get wallet: %w", err,
```
yyforyongyu: can fit in oneline

wallet/internal/db/queries/postgres/wallets.sql
```diff
@@ -0,0 +1,122 @@
...
+    master_pub_params,
+    encrypted_crypto_pub_key,
+    encrypted_master_hd_pub_key
+) VALUES (
+    $1, $2, $3, $4, $5, $6, $7
+)
+RETURNING id;
+
+-- name: UpsertWalletSyncState :exec
+INSERT INTO wallet_sync_states (
```
yyforyongyu, line 16: After reviewing the PR I think we can use this query instead,
```sql
1 -- name: UpdateWalletSyncState :exec
UPDATE wallet_sync_states
SET
    -- If input $2 is NOT NULL, use it. Otherwise, keep the existing 'synced_height'.
    synced_height = COALESCE($2, synced_height),

    -- If input $3 is NOT NULL, use it. Otherwise, keep the existing 'birthday_height'.
    birthday_height = COALESCE($3, birthday_height),

    -- If input $4 is NOT NULL, use it. Otherwise, keep the existing 'birthday_verified'.
    birthday_verified = COALESCE($4, birthday_verified),

    -- Always update the timestamp to the current database time.
    updated_at = NOW()
WHERE
    wallet_id = $1;
```

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,392 @@
...
+		)
+	}
+
+	syncParams, err := buildPgWalletSyncStateUpsert(params, current)
```
yyforyongyu: I think we hitting the same issue we are facing in `lnd`. Say two goroutines are trying to update the same wallet,
  - GA: wants to update the Synced Height to 101.
  - GB: wants to update the Birthday Height to 50.

We would have this possible race:
  1. Time 0: DB has: Synced=100, Birthday=40.
  2. Time 1 (GA): Reads DB. Gets Synced=100, Birthday=40.
  3. Time 2 (GB): Reads DB. Gets Synced=100, Birthday=40.
  4. Time 3 (GA): Merges its change.
       * New Synced = 101 (its change)
       * New Birthday = 40 (from its read at Time 1)
  5. Time 4 (GA): Writes to DB: Synced=101, Birthday=40. DB is now: 101, 40.
  6. Time 5 (GB): Merges its change.
       * New Synced = 100 (FROM ITS READ AT TIME 2!) <--- STALE DATA
       * New Birthday = 50 (its change)
  7. Time 6 (GB): Writes to DB: Synced=100, Birthday=50.

then we would have a DB State: Synced=100, Birthday=50, because GA's update to Synced=101 was overwritten and lost by GB because GB was working with old information.

This method, `buildPgWalletSyncStateUpsert`, while it tries to overwrite the params field when they are not set, cannot handle this race.

To fix it, we can use `SELECT FOR UPDATE` for postgres, and for sqlite we need to dig more. Or we can use `COALESCE` to achieve it - this way we don't need to read the wallet row first, we can just do the update, it should make the tx smaller and faster.

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,392 @@
...
+
+// UpdateWallet updates various properties of a wallet, such as its
+// birthday, birthday block, or sync state. The specific fields to
+// update are provided in the UpdateWalletParams struct. It returns an
+// error if the update fails.
+func (w *PostgresWalletDB) UpdateWallet(
+	ctx context.Context, params UpdateWalletParams,
+) error {
+
+	tx, err := w.db.BeginTx(ctx, nil)
```
yyforyongyu: Looks like this is a repeating pattern, begin tx -> with tx -> commit/rollback, i think we should build a `ExecuteTx` to wrap the query? sth like,
```go
func (s *BaseDB) ExecuteTx(ctx context.Context, fn func(*Queries) error) error {
```

wallet/internal/db/wallet_sqlite.go
yyforyongyu: While this approach is clever, we should not go with that route, at least for this milestone. If we find that we indeed need to de-duplicate things, we can then maybe do that.

The large duplication here comes from lots of the code which can otherwise be made into private helpers, and be called by both implementations, that should eliminate quite some repeating pattern.

With the generic approach, it tries to hide the difference between the different types generated by `sqlc`, by increasing the cognitive load and readability, which increases the maintenance burden. For instance, one change in sqlite now needs to be reflected on the interface, which can further affect the postgres driver. In the long run, we always prefer seemingly "dumb" and simple solution(KISS principle).

wallet/internal/db/block_sqlite.go
yyforyongyu: or build a helper method that can be used by both impls,
```go
func buildBlockCommon(height uint32, hash []byte, timestampSecs int64) (*Block, error) {
     h, err := chainhash.NewHash(hash)
     if err != nil {
         return nil, fmt.Errorf("block hash: %w", err)
     }
 
     return &Block{
         Hash:      *h,
         Height:    height,
         Timestamp: time.Unix(timestampSecs, 0),
     }, nil
 }
```

.github/workflows/main.yml
```diff
@@ -152,3 +152,67 @@ jobs:
         with:
           path-to-profile: coverage.txt
           parallel: true
+
+  ########################
+  # run integration tests (SQLite)
+  ########################
+  itest-sqlite:
+    name: integration tests (sqlite)
+    runs-on: ubuntu-latest
```
yyforyongyu: I think we should use `fail-fast: false`, and what about using `matrix` so we don't need to duplicate the code here? Need to double check whether they will be running in its own instance or not tho, as we want to use independent machine to speed up the tests.

wallet/internal/db/itest/fixtures_test.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// RandomBytes generates random bytes for test data.
+func RandomBytes(n int) []byte {
+	b := make([]byte, n)
+
+	_, err := rand.Read(b)
+	if err != nil {
+		// This should never happen.
+		panic(fmt.Sprintf("failed to generate random bytes: %v", err))
```
yyforyongyu, line 83: Nice! This is exactly what `design error out of existence` is for.

wallet/internal/db/itest/fixtures_test.go
```diff
@@ -0,0 +1,100 @@
...
+	t.Helper()
+
+	hash := RandomHash()
+	timestamp := time.Now().UTC()
+
+	// TODO(gustavostingelin): use the block store to insert the block when
+	// available.
+	query := `
+		INSERT INTO blocks (block_height, header_hash, timestamp)
+		VALUES ($1, $2, $3)
```
yyforyongyu, line 63: So this syntax is for postgres? maybe mention why it's ok to use it here?

wallet/internal/db/itest/pg_test.go
```diff
@@ -0,0 +1,191 @@
...
+// The container is created once and reused across all tests for performance.
+func GetPostgresContainer(ctx context.Context) (*postgres.PostgresContainer, error) {
+	pgContainerOnce.Do(func() {
+		cfg := DefaultPostgresConfig()
+
+		pgContainer, pgContainerErr = postgres.RunContainer(ctx,
+			testcontainers.WithImage(cfg.Image),
+			postgres.WithDatabase(cfg.Database),
+			postgres.WithUsername(cfg.Username),
+			postgres.WithPassword(cfg.Password),
```
yyforyongyu, line 101: What is the default max db conns? We need to document that here so it's easier to debug in the future.

wallet/internal/db/itest/pg_test.go
```diff
@@ -0,0 +1,191 @@
...
+	dbName := sanitizedPgDBName(t)
+
+	var dbConn *sql.DB
+
+	// Cleanup: drop the database after the test.
+	t.Cleanup(func() {
+		if dbConn != nil {
+			_ = dbConn.Close()
+		}
+		if adminDB != nil {
```
yyforyongyu: I think this is unnecessary? The container will be destroyed anyway.

wallet/internal/db/itest/sqlite_test.go
```diff
@@ -0,0 +1,53 @@
...
+	tmpDir := t.TempDir()
+	dbPath := filepath.Join(tmpDir, "test.db")
+
+	// TODO(gustavostingelin): replace with the real SQLite database
+	// connection constructor when available.
+	dbConn, err := sql.Open("sqlite", dbPath)
+	require.NoError(t, err, "failed to open sqlite database")
+
+	// Enable foreign keys (required for proper constraint enforcement).
+	_, err = dbConn.Exec("PRAGMA foreign_keys = ON")
```
yyforyongyu: I'm not sure if it works, given that `Open` maintains its own pool of idle conns, so conn A may have this enabled, but conn B doesn't.
https://pkg.go.dev/database/sql#OpenDB

Makefile
```diff
@@ -84,6 +84,15 @@ unit-bench:
 	@$(call print, "Running benchmark tests.")
 	$(UNIT_BENCH)
 
+#? itest: Run integration tests for wallet database
+itest:
```
yyforyongyu: Let's name it `itest-db`? I think the current test is more like component tests, but it's also checking the integrating part, so ok to keep it. `make itest` can be saved later when we build the full integration tests for `wallet + chain + db`.

yyforyongyu (changes requested):
Very clean PR! My comments are,
- we need to fix the race condition from the read-write pattern.
- we can create a smaller PR to include the itest code, merge it to the base branch so others can use.
- we need to create more private helper methods to dedup the code, and add unit tests for these private helpers.
- we need to think about how to run race tests, maybe run the itest with the race flag?

Also checked the coverage rate,
Sqlite:
```
make itest db=sqlite nocache=1 TEST_FLAGS="-coverpkg=github.com/btcsuite/btcwallet/wallet/internal/db/... -coverprofile=itest.out" && go tool cover -func=itest.out | grep total

github.com/btcsuite/btcwallet/wallet/internal/db/block_sqlite.go:13:			buildSqliteBlock			71.4%
github.com/btcsuite/btcwallet/wallet/internal/db/migrations.go:38:			applyMigrations				69.2%
github.com/btcsuite/btcwallet/wallet/internal/db/migrations.go:72:			ApplySQLiteMigrations			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/migrations.go:87:			ApplyPostgresMigrations			0.0%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:22:			int64ToUint32				66.7%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:35:			int64ToInt32				66.7%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:48:			uint32ToInt32				0.0%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:61:			uint32ToNullInt32			0.0%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:72:			nullInt32ToUint32			0.0%
github.com/btcsuite/btcwallet/wallet/internal/db/sqlite.go:16:				NewSQLiteWalletDB			66.7%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_sqlite.go:19:			CreateWallet				78.6%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_sqlite.go:115:			GetWallet				83.3%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_sqlite.go:149:			ListWallets				80.0%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_sqlite.go:187:			UpdateWallet				78.9%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_sqlite.go:241:			GetEncryptedHDSeed			50.0%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_sqlite.go:261:			UpdateWalletSecrets			80.0%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_sqlite.go:299:			buildSqliteWalletInfo			78.9%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_sqlite.go:354:			buildSqliteWalletSyncStateUpsert	100.0%
```

Postgres:
```
make itest db=postgres nocache=1 TEST_FLAGS="-coverpkg=github.com/btcsuite/btcwallet/wallet/internal/db/... -coverprofile=itest-pg.out" && go tool cover -func=itest-pg.out | grep total

github.com/btcsuite/btcwallet/wallet/internal/db/block_pg.go:13:			buildPgBlock				71.4%
github.com/btcsuite/btcwallet/wallet/internal/db/migrations.go:38:			applyMigrations				69.2%
github.com/btcsuite/btcwallet/wallet/internal/db/migrations.go:72:			ApplySQLiteMigrations			0.0%
github.com/btcsuite/btcwallet/wallet/internal/db/migrations.go:87:			ApplyPostgresMigrations			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/pg.go:17:				NewPostgresWalletDB			66.7%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:22:			int64ToUint32				66.7%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:35:			int64ToInt32				0.0%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:48:			uint32ToInt32				66.7%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:61:			uint32ToNullInt32			75.0%
github.com/btcsuite/btcwallet/wallet/internal/db/safecasting.go:72:			nullInt32ToUint32			60.0%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_pg.go:19:			CreateWallet				78.6%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_pg.go:115:			GetWallet				83.3%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_pg.go:149:			ListWallets				80.0%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_pg.go:187:			UpdateWallet				76.2%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_pg.go:244:			GetEncryptedHDSeed			50.0%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_pg.go:264:			UpdateWalletSecrets			80.0%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_pg.go:302:			buildPgWalletInfo			81.2%
github.com/btcsuite/btcwallet/wallet/internal/db/wallet_pg.go:349:			buildPgWalletSyncStateUpsert		87.5%
```

I think we can use another PR to include `itest-db-cover` to run the coverage tests.

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
GustavoStingelin: yes, changed

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
GustavoStingelin: done

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
GustavoStingelin: added

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
GustavoStingelin: sure

wallet/internal/db/migrations/sqlite/000002_wallets.up.sql
GustavoStingelin: added

wallet/internal/db/block_pg.go
GustavoStingelin: reformated

wallet/internal/db/block_sqlite.go
GustavoStingelin: added the common function

wallet/internal/db/wallet_pg.go
GustavoStingelin: changed to db side

wallet/internal/db/queries/postgres/wallets.sql
GustavoStingelin: applied

wallet/internal/db/wallet_pg.go
GustavoStingelin: nice catch, changed to the coalesce approach

wallet/internal/db/wallet_pg.go
GustavoStingelin: added

wallet/internal/db/wallet_pg.go
GustavoStingelin: I think both are equivalents for the golang side, ref https://go.dev/wiki/CodeReviewComments#declaring-empty-slices

wallet/internal/db/wallet_pg.go
GustavoStingelin: changed, take a look

.github/workflows/main.yml
GustavoStingelin: added

wallet/internal/db/itest/fixtures_test.go
GustavoStingelin: it's working on both databases and is just temporary, as the todo above, do you think it's really needed?

wallet/internal/db/itest/pg_test.go
GustavoStingelin: added a comment

wallet/internal/db/itest/pg_test.go
GustavoStingelin: removed

Makefile
GustavoStingelin: renamed

GustavoStingelin:
@yyforyongyu, I added the race tests and caught the https://gitlab.com/cznic/sqlite/-/issues/180 issue, fixed by upgrading the lib.

GustavoStingelin:
I will split the itests pr later, but we will have nothing to run

wallet/internal/db/itest/sqlite_test.go
GustavoStingelin: moved to the connection string

GustavoStingelin:
splited itests in #1136, I will need rebase after merge

GustavoStingelin:
rebased

wallet/internal/db/tx.go
```diff
@@ -0,0 +1,36 @@
...
+// PostgresWalletDB and SQLiteWalletDB. It guarantees that the transaction
+// will be either committed (on success) or rolled back (on error or panic).
+func execInTx(ctx context.Context, db *sql.DB, fn func(*sql.Tx) error) error {
+	tx, err := db.BeginTx(ctx, nil)
+	if err != nil {
+		return fmt.Errorf("begin tx: %w", err)
+	}
+
+	defer func() {
+		_ = tx.Rollback()
```
yyforyongyu, line 27: nit: explain why it's ok to ignore the error here?

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,333 @@
...
+// error if the wallet is not found.
+func (w *PostgresWalletDB) GetWallet(ctx context.Context,
+	name string) (*WalletInfo, error) {
+
+	row, err := w.queries.GetWalletByName(ctx, name)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, fmt.Errorf(
+				"wallet %q: %w",
+				name, ErrWalletNotFound,
```
yyforyongyu: nit: can fit in oneline

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,333 @@
...
+	return w.ExecuteTx(ctx, func(qtx *sqlcpg.Queries) error {
+		// Build sync state update params directly from input.
+		// Only set fields that are being updated (leave others as nil).
+		syncParams := sqlcpg.UpdateWalletSyncStateParams{
+			WalletID: int64(params.WalletID),
+		}
+
+		if params.SyncedTo != nil {
+			syncedHeight, err := uint32ToNullInt32(
+				params.SyncedTo.Height)
```
yyforyongyu, line 196: missing a newline

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,333 @@
...
+			if err != nil {
+				return err
+			}
+
+			syncParams.SyncedHeight = syncedHeight
+		}
+
+		if params.BirthdayBlock != nil {
+			birthdayHeight, err := uint32ToNullInt32(
+				params.BirthdayBlock.Height)
```
yyforyongyu, line 213: same here re newline

wallet/internal/db/wallet_pg.go
```diff
@@ -0,0 +1,333 @@
...
+			if err != nil {
+				return err
+			}
+
+			syncParams.BirthdayHeight = birthdayHeight
+
+			// Set birthday verified flag to true if the birthday
+			// block is set.
+			syncParams.BirthdayVerified = sql.NullBool{
```
yyforyongyu: Q: in other words this field is always true when the `BirthdayHeight` is set, which means this field is redundant?

wallet/internal/db/wallet_sqlite.go
```diff
@@ -0,0 +1,332 @@
...
+			syncedHeight:           row.SyncedHeight,
+			syncedBlockHash:        row.SyncedBlockHash,
+			syncedBlockTimestamp:   row.SyncedBlockTimestamp,
+			birthdayHeight:         row.BirthdayHeight,
+			birthdayBlockHash:      row.BirthdayBlockHash,
+			birthdayBlockTimestamp: row.BirthdayBlockTimestamp,
+		})
+		if err != nil {
+			return fmt.Errorf(
+				"convert wallet row to info: %w", err,
```
yyforyongyu: nit: feels like this can be fit in one line tho

wallet/internal/db/itest/fixtures_test.go
yyforyongyu: Ok I thought `$1` syntax is for postgres only, but forgot that `sqlc` will perform the translation for sqlite so nvm.

wallet/internal/db/itest/wallet_store_test.go
```diff
@@ -0,0 +1,277 @@
...
+	require.Equal(t, params.ManagerVersion, info.ManagerVersion)
+	require.Equal(t, params.IsWatchOnly, info.IsWatchOnly)
+
+	require.Nil(t, info.SyncedTo)
+	require.Equal(t, uint32(0), info.BirthdayBlock.Height)
+}
+
+// TestCreateWallet_DuplicateName verifies that creating a wallet with a
+// duplicate name fails with an appropriate error.
+func TestCreateWallet_DuplicateName(t *testing.T) {
```
yyforyongyu, line 56: hmm the `_` is a new format used in our project, but I guess it's fine given that it's a common use in the standard lib, e.g., https://cs.opensource.google/go/go/+/master:src/net/http/httputil/reverseproxy_test.go;l=345

wallet/internal/db/itest/wallet_store_test.go
```diff
@@ -0,0 +1,277 @@
...
+
+	store, _ := NewTestStore(t)
+	params := CreateWalletParamsFixture("duplicate-wallet")
+
+	_, err := store.CreateWallet(t.Context(), params)
+	require.NoError(t, err)
+
+	// Attempt to create second wallet with same name.
+	_, err = store.CreateWallet(t.Context(), params)
+	require.Error(t, err, "expected error creating duplicate wallet")
```
yyforyongyu, line 67: We should perform more rigorous check here, so check the error type via `require.ErrorIs`?

wallet/internal/db/itest/wallet_store_test.go
```diff
@@ -0,0 +1,277 @@
...
+		WalletID: created.ID,
+		SyncedTo: &block,
+	}
+	err = store.UpdateWallet(t.Context(), updateParams)
+	require.NoError(t, err)
+
+	retrieved, err := store.GetWallet(t.Context(), created.Name)
+	require.NoError(t, err)
+	require.NotNil(t, retrieved.SyncedTo)
+	require.Equal(t, block.Height, retrieved.SyncedTo.Height)
```
yyforyongyu, line 202: let's also assert other fields are NOT updated?

wallet/internal/db/itest/wallet_store_test.go
```diff
@@ -0,0 +1,277 @@
...
+		WalletID:      created.ID,
+		BirthdayBlock: &block,
+	}
+	err = store.UpdateWallet(t.Context(), updateParams)
+	require.NoError(t, err)
+
+	retrieved, err := store.GetWallet(t.Context(), created.Name)
+	require.NoError(t, err)
+	require.Equal(t, block.Height, retrieved.BirthdayBlock.Height)
+	require.Equal(t, block.Hash, retrieved.BirthdayBlock.Hash)
```
yyforyongyu, line 240: same here re asserting other fields like `SyncedTo` is nil

wallet/internal/db/wallet_pg.go
yyforyongyu: yeah my concern is this impl doesn't follow the API contract - that we either return `nil, err`, or `value, nil`, as the caller may forget to check it against nil. To be safer, I think we should check it against nil and return an error when empty.

yyforyongyu (commented):
Very close! Left a few nits and questions.

wallet/internal/db/wallet_pg.go
GustavoStingelin: I'm trying to add, but `make lint` is removing this new line... should we try to fix the linter?

wallet/internal/db/wallet_pg.go
GustavoStingelin: same issue with the linter here

wallet/internal/db/tx.go
GustavoStingelin: added the comment

wallet/internal/db/wallet_pg.go
GustavoStingelin: fixed

wallet/internal/db/wallet_sqlite.go
GustavoStingelin: fixed

wallet/internal/db/itest/wallet_store_test.go
GustavoStingelin: It is a bit tricky because SQLite returns `constraint failed: UNIQUE constraint failed: wallets.name (2067)` while Postgres returns `ERROR: duplicate key value violates unique constraint "uidx_wallets_name" (SQLSTATE 23505)`. I added a check that asserts the error contains the common parts of these messages, please see if that feels reliable enough. If not, we will probably need to handle and map these errors in a more structured way.

wallet/internal/db/itest/wallet_store_test.go
GustavoStingelin: added

wallet/internal/db/itest/wallet_store_test.go
GustavoStingelin: added

wallet/internal/db/wallet_pg.go
GustavoStingelin: ok, now we return an error when someone tries to retrieve the seed from a wallet that has no seed info or has an empty value. please take a look.

wallet/internal/db/itest/fixtures_test.go
GustavoStingelin: The Go standard library or the SQLite driver is handling this, not SQLC. I will replace this with the SQLC implementation once Mohamed’s PR is merged.

wallet/internal/db/wallet_pg.go
GustavoStingelin: I noticed that we were probably missing some pieces in the design, so the last commit adds these rules:

- A wallet can be created with a birthday timestamp that does not match any entry in the blocks table yet, since the node may still be syncing. Because of that, the birthday height alone is not enough.

- If the birthday timestamp is not set, the wallet should rescan starting from the genesis block. If it is set, the rescan should begin from the user supplied timestamp.

- When the wallet finds transactions and identifies the corresponding block, it should record the birthday height. From that moment on, the birthday block is considered confirmed, and future rescans should begin from there.

I used the `SetBirthdayBlock` function usage as a guide.

What do you think of this approach?

wallet/internal/db/queries/postgres/wallets.sql
```diff
@@ -0,0 +1,134 @@
...
+    updated_at
+) VALUES (
+    $1, $2, $3, $4, CURRENT_TIMESTAMP
+);
+
+-- name: UpdateWalletSyncState :execrows
+UPDATE wallet_sync_states
+SET
+    -- If synced_height param is NOT NULL, use it. Otherwise, keep existing value.
+    synced_height = COALESCE(sqlc.narg('synced_height'), synced_height),
```
mohamedawnallah, line 30: https://github.com/mohamedawnallah/btcwallet/pull/39#discussion_r2609411428

mohamedawnallah (approved):

wallet/internal/db/wallet_pg.go
yyforyongyu: Weird, cause it worked for me locally, so I did,
```diff
diff --git a/wallet/internal/db/wallet_pg.go b/wallet/internal/db/wallet_pg.go
index c50f16d8..31e70fd2 100644
--- a/wallet/internal/db/wallet_pg.go
+++ b/wallet/internal/db/wallet_pg.go
@@ -193,7 +193,8 @@ func (w *PostgresWalletDB) UpdateWallet(ctx context.Context,
 
 		if params.SyncedTo != nil {
 			syncedHeight, err := uint32ToNullInt32(
-				params.SyncedTo.Height)
+				params.SyncedTo.Height,
+			)
 			if err != nil {
 				return err
 			}
```
then run `make lint`, and it didn't change anything or report any errors. Can you run `make lint-check` to see which linter is complaining?

wallet/internal/db/queries/postgres/wallets.sql
GustavoStingelin: We need to use narg so sqlc generates the correct optional type in the patch query. Without that, sqlc uses the original column type and we cannot apply the coalesce approach.

wallet/internal/db/itest/wallet_store_test.go
yyforyongyu: yeah ideally we don't want to return the raw error from sql here, and we should wrap it so the upper layer understands how it should be handled - but that can be done in future improvement PRs.

wallet/internal/db/wallet_pg.go
yyforyongyu: When a wallet is created, we perform a binary search to find the birthday block,
https://github.com/btcsuite/btcwallet/blob/81113d4438d6fb6bbd7510b0df5cc40642eda3f7/wallet/wallet.go#L628-L630

So yeah that block is likely not existed in the blocks table, nor should we care about it. We will always locate the birthday block first before rescaning, otherwise it will take forever for the rescan to finish if we were to start from the genesis block. From the db layer, it should not care about this business logic tho, so I think it makes sense to records a birthday ts here.

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
```diff
@@ -0,0 +1,109 @@
...
+    -- that haven't synced any blocks yet.
+    synced_height INTEGER,
+
+    -- Birthday block height of the wallet (references blocks table). NULL if the
+    -- wallet has no known birthday block. When set, indicates the block has been
+    -- verified.
+    birthday_height INTEGER,
+
+    -- User-provided birthday timestamp for wallet rescan. NULL if not set.
+    birthday TIMESTAMP,
```
yyforyongyu, line 86: nit: would be more explicit to call it `birthday_ts` or `birthday_timestamp`.

yyforyongyu (approved):
Looking good, feel free to ignore/address the nits in another PR.

wallet/internal/db/wallet_pg.go
GustavoStingelin: Got it, your diff works fine with the linter. I was mistakenly assuming there had to be a blank line after the closing parenthesis.

```diff
diff --git a/wallet/internal/db/wallet_pg.go b/wallet/internal/db/wallet_pg.go
index 1ac42a3d..e8bbf3f6 100644
--- a/wallet/internal/db/wallet_pg.go
+++ b/wallet/internal/db/wallet_pg.go
@@ -194,6 +194,7 @@ func (w *PostgresWalletDB) UpdateWallet(ctx context.Context,
                if params.SyncedTo != nil {
                        syncedHeight, err := uint32ToNullInt32(
                                params.SyncedTo.Height)
+
                        if err != nil {
                                return err
                        }
```

GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1139

wallet/internal/db/wallet_pg.go
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1139

wallet/internal/db/migrations/postgres/000002_wallets.up.sql
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1139

PR #1127, opened 2025-11-27
unit+wallet: add missing units and base units

yyforyongyu:
Introduced from #1091, I realized we are missing a few units in this package. In this PR,
- add missing `KVByte` and `KWeightUnit` and `SatPerWeight` units
- unify all calculations - we will always use `weight` as the base size and `sat/kw` as the base fee rate
- rename methods for clarity
- use the new fee rate unit in `TxIntent`

### DIFF
--- a/pkg/btcunit/rates.go
+++ b/pkg/btcunit/rates.go
@@ -34,6 +34,9 @@ var (
 
 	// ZeroSatPerKWeight is a fee rate of 0 sat/kw.
 	ZeroSatPerKWeight = NewSatPerKWeight(0, NewKWeightUnit(1))
+
+	// ZeroSatPerWeight is a fee rate of 0 sat/wu.
+	ZeroSatPerWeight = NewSatPerWeight(0, NewWeightUnit(1))
 )
 
 // baseFeeRate stores the canonical representation of a fee rate, which is
@@ -74,6 +77,11 @@ func (f baseFeeRate) ToSatPerKVByte() SatPerKVByte {
 // ToSatPerKWeight converts the fee rate to sat/kw.
 func (f baseFeeRate) ToSatPerKWeight() SatPerKWeight {
 	return SatPerKWeight{f}
+}
+
+// ToSatPerWeight converts the fee rate to sat/wu.
+func (f baseFeeRate) ToSatPerWeight() SatPerWeight {
+	return SatPerWeight{f}
 }
 
 // FeeForWeight calculates the fee resulting from this fee rate and the given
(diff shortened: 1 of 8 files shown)


pkg/btcunit/rates.go
```diff
@@ -15,268 +15,471 @@ import (
...
+	// calculations and to minimize rounding errors.
+	satsPerKWU *big.Rat
+}
+
+// newBaseFeeRate creates a new baseFeeRate with the given numerator and
+// denominator. It handles the zero denominator case by returning a zero fee
+// rate.
+func newBaseFeeRate(numerator btcutil.Amount, denominator uint64) baseFeeRate {
+	if denominator == 0 {
+		return baseFeeRate{satsPerKWU: big.NewRat(0, 1)}
```
mohamedawnallah: When the denominator can be zero? We can raise and propagate an error instead of swallowing it

mohamedawnallah (commented):
Thanks, Looks good. I posted one comment and gemini posted a few comments – seems valid. Gemini comments can be accessed here: https://github.com/mohamedawnallah/btcwallet/pull/35#pullrequestreview-3517808308

pkg/btcunit/rates.go
yyforyongyu: This method takes an input, so it can be anything. Previously the mental model was, if the tx has 0 weight, we will pay 0 fee, so fee rate is also 0. Added a commit to panic instead.

yyforyongyu:
> Thanks, Looks good. I posted one comment and gemini posted a few comments – seems valid. Gemini comments can be accessed here: [mohamedawnallah#35 (review)](https://github.com/mohamedawnallah/btcwallet/pull/35#pullrequestreview-3517808308)

Cool addressed

mohamedawnallah (approved):
Thanks, LGTM!

pkg/btcunit/rates.go
```diff
@@ -15,268 +15,455 @@ import (
...
+func newBaseFeeRate(numerator btcutil.Amount, denominator uint64) baseFeeRate {
+	if denominator == 0 {
+		panic("fee rate calculation: denominator cannot be zero")
 	}
+
+	return baseFeeRate{satsPerKWU: big.NewRat(
+		int64(numerator),
+		safeUint64ToInt64(denominator),
+	)}
+}
```
GustavoStingelin, line 64: Q: Could this panic be used as an attack vector?

GustavoStingelin (approved):
🚀

pkg/btcunit/rates.go
yyforyongyu: I don't think so - it's up to the upper layer to validate that the tx has a size greater than 0, and we don't have a way to build a tx that has 0 size I think - unless this is used for calculating fee rate directly, without asking the size of a tx first. In that case an error is probably better, then again it should be handled at the callsite.

PR #1128, opened 2025-11-27
wallet: add `PsbtManager` interface

yyforyongyu:
Check the final PR for reference: #1130


wallet/psbt_manager.go
```diff
@@ -405,3 +430,274 @@ func findCredit(txDetail *wtxmgr.TxDetails, outputIndex uint32) bool {
...
+	// We'll also re-initialize the input and output slices to match the
+	// dimensions of the new transaction.
+	packet.Inputs = make([]psbt.PInput, len(authoredTx.Tx.TxIn))
+	packet.Outputs = make([]psbt.POutput, len(authoredTx.Tx.TxOut))
+
+	// With the new inputs in place, we'll decorate them with UTXO and
+	// derivation information from the wallet. We set `skipUnknown` to
+	// false because all inputs in the `authoredTx` must be known to the
+	// wallet.
+	_, err := w.DecorateInputs(ctx, packet, false)
```
mohamedawnallah, line 500: I have checked in the current codebase where `DecorateInputs` is called `skipUnknown` parameter is always `false` excluding the tests. Why is it needed in `DecorateInputs` signature?

wallet/psbt_manager.go
```diff
@@ -405,3 +430,274 @@ func findCredit(txDetail *wtxmgr.TxDetails, outputIndex uint32) bool {
...
+
+	for i, txOut := range packet.UnsignedTx.TxOut {
+		if psbt.TxOutsEqual(
+			changeOutput,
+			txOut,
+		) {
+
+			if i > math.MaxInt32 {
+				return 0, ErrChangeIndexOutOfRange
+			}
```
mohamedawnallah: Having this check outside the for loop for earlier termination?

mohamedawnallah (approved):
Thanks, LGTM! Left non-blocking comments

GustavoStingelin:
@yyforyongyu missing a rebase

wallet/psbt_manager.go
yyforyongyu: So it's used in `lnd` here, https://github.com/lightningnetwork/lnd/blob/a76f22da9d167bb456c0828c3c723c6e41e1f01d/lnrpc/walletrpc/walletkit_server.go#L1947

Note that we flipped the semantics here, previously we use `failOnUnknown`, now we use `skipUnknown`.

wallet/psbt_manager.go
yyforyongyu: Done!

wallet/psbt_manager.go
```diff
@@ -0,0 +1,699 @@
...
+//  3. Population: Calling `populatePsbtPacket` to apply the selected inputs and
+//     change output to the PSBT structure and sort it according to BIP 69.
+func (w *Wallet) FundPsbt(ctx context.Context, intent *FundIntent) (
+	*psbt.Packet, int32, error) {
+
+	// Validate the funding intent before proceeding.
+	err := w.validateFundIntent(intent)
+	if err != nil {
+		return nil, 0, err
+	}
```
GustavoStingelin, line 453: This function will panic if someone accidentally passes a nil intent

wallet/psbt_manager.go
```diff
@@ -0,0 +1,699 @@
...
+
+	// The authored transaction contains the selected inputs and the change
+	// output (if any). We'll update the PSBT packet with this new
+	// unsigned transaction.
+	packet.UnsignedTx = authoredTx.Tx
+
+	// We'll also re-initialize the input and output slices to match the
+	// dimensions of the new transaction.
+	packet.Inputs = make([]psbt.PInput, len(authoredTx.Tx.TxIn))
+	packet.Outputs = make([]psbt.POutput, len(authoredTx.Tx.TxOut))
```
GustavoStingelin, line 494: Q: Why is it safe to wipe the Outputs from the PSBT and trust only on authoredTx?

GustavoStingelin (approved):
Left two comments, but can be addressed in the next PR if needed.

wallet/psbt_manager.go
yyforyongyu: Cool added a check in the last commit!

wallet/psbt_manager.go
yyforyongyu: Added some comments there, in short we treat the `authoredTx` as the source of truth, and we also want to maintain the same outputs order as it may be different from the original outputs order.

PR #1129, opened 2025-11-27
wallet: implement `SignPsbt`

yyforyongyu:
Check the final PR for reference: #1130

### DIFF
--- a/wallet/psbt_manager.go
+++ b/wallet/psbt_manager.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"math"
 
+	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/btcutil/psbt"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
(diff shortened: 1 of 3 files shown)


wallet/psbt_manager.go
```diff
@@ -43,8 +46,101 @@ var (
...
+	// ErrInvalidBip32DerivationCoinType is returned when the coin type in
+	// a BIP32 derivation path does not match the wallet's chain parameters.
+	ErrInvalidBip32DerivationCoinType = errors.New(
+		"invalid BIP32 derivation path, coin type mismatch",
+	)
+)
+
+const (
+	// BIP32PathLength is the expected length of a BIP32 derivation path.
+	BIP32PathLength = 5
```
mohamedawnallah, line 224: Perhaps better for the godoc comment to have what 5 elements are. I see it is mentioned down below in some of code paths:
```
	// The BIP32 path must have exactly 5 elements:
	// m / purpose' / coin_type' / account' / branch / index
	if len(path) != BIP32PathLength {
```

wallet/psbt_manager.go
```diff
@@ -697,3 +793,717 @@ func (w *Wallet) createTxIntent(intent *FundIntent) *TxIntent {
...
+	TaprootSpendDetails, error) {
+
+	var details TaprootSpendDetails
+
+	switch {
+	// Case 1: Key Path Spend.
+	// A non-empty merkle root means we committed to a taproot hash
+	// that we need to use in the tap tweak. If LeafHashes is empty, it
+	// means we are signing for the internal key (Key Path).
+	case len(tapDerivation.LeafHashes) == 0:
```
mohamedawnallah: The operation `len(tapDerivation.LeafHashes)` repeated twice in switch case statements here and down below. Though it seems O(1), can be as variable above

wallet/psbt_manager.go
```diff
@@ -697,3 +793,717 @@ func (w *Wallet) createTxIntent(intent *FundIntent) *TxIntent {
...
+	return bip32Path, nil
+}
+
+// addressTypeFromPurpose maps a BIP purpose to a wallet address type.
+func addressTypeFromPurpose(purpose uint32) (waddrmgr.AddressType, error) {
+	// TODO(yy): Currently, we hardcode the supported BIP purposes.
+	// A more robust solution would dynamically query the `waddrmgr` to
+	// determine supported key scopes configured in the database, allowing
+	// for custom purposes (e.g., LND's 1017 purpose key) to be seamlessly
+	// supported without code changes here.
```
mohamedawnallah, line 1089: 👍

wallet/psbt_manager.go
```diff
@@ -43,8 +46,101 @@ var (
...
+		"psbts do not refer to the same transaction",
+	)
+
+	// ErrInputCountMismatch is returned when PSBTs have different input
+	// counts.
+	ErrInputCountMismatch = errors.New("input count mismatch")
+
+	// ErrOutputCountMismatch is returned when PSBTs have different output
+	// counts.
+	ErrOutputCountMismatch = errors.New("output count mismatch")
```
mohamedawnallah, line 68: Those errors defined here and not used

mohamedawnallah (approved):

wallet/psbt_manager.go
GustavoStingelin: I agree

wallet/psbt_manager.go
```diff
@@ -43,11 +47,104 @@ var (
...
+	ErrTaprootLeafHashMismatch = errors.New("specified leaf hash in " +
+		"taproot BIP0032 derivation but corresponding taproot leaf " +
+		"script was not found")
+
+	// errAlreadySigned is returned when an input is already signed.
+	errAlreadySigned = errors.New("input already signed")
+
+	// errComputeRawSig is returned when the wallet cannot produce a
+	// signature for the input (e.g. key not found, signing error).
+	errComputeRawSig = errors.New("cannot compute raw signature")
```
GustavoStingelin: NON-BLOCKING: could explain why these errors are private

wallet/psbt_manager.go
```diff
@@ -710,3 +807,717 @@ func (w *Wallet) createTxIntent(intent *FundIntent) *TxIntent {
...
+//     wallet and ready for signing.
+//  3. Derivation Validation: Enforcing strict rules on derivation paths (one
+//     path per input) to ensure deterministic key selection.
+//  4. Signing: dispatching to `signTaprootPsbtInput` or `signBip32PsbtInput` to
+//     generate the raw ECDSA or Schnorr signature using the underlying
+//     `Signer`.
+func (w *Wallet) SignPsbt(ctx context.Context, params *SignPsbtParams) (
+	*SignPsbtResult, error) {
+
+	packet := params.Packet
```
GustavoStingelin, line 907: could panic nil pointer if `params == nil`

wallet/psbt_manager.go
```diff
@@ -710,3 +807,717 @@ func (w *Wallet) createTxIntent(intent *FundIntent) *TxIntent {
...
+
+			return nil, fmt.Errorf("input %d: %w", i, err)
+		}
+
+		// If signing was successful (or the error was gracefully
+		// skipped), we record the index of this input as one that the
+		// wallet has contributed a signature to. This provides
+		// valuable feedback to the caller about the progress of the
+		// signing operation.
+		signedInputs = append(signedInputs, uint32(i)) //nolint:gosec
```
GustavoStingelin: Q: What guarantees that we will not receive a slice larger than uint32?

wallet/psbt_manager.go
```diff
@@ -710,3 +807,717 @@ func (w *Wallet) createTxIntent(intent *FundIntent) *TxIntent {
...
+	// The BIP32 path must have exactly 5 elements:
+	// m / purpose' / coin_type' / account' / branch / index
+	if len(path) != BIP32PathLength {
+		return BIP32Path{}, fmt.Errorf("%w: %d",
+			ErrInvalidBip32PathLength, len(path))
+	}
+
+	// The first 3 elements (Purpose, CoinType, Account) must be hardened.
+	// We check this by verifying they are >= HardenedKeyStart.
+	for i := range BIP32PathLength - 2 {
```
GustavoStingelin: Q: Why not just `3`?
```suggestion
	for i := range 3 {
```

wallet/psbt_manager.go
```diff
@@ -710,3 +807,717 @@ func (w *Wallet) createTxIntent(intent *FundIntent) *TxIntent {
...
+				ErrInvalidBip32PathElementHardened, i)
+		}
+	}
+
+	// Helper to extract values (remove hardened flag).
+	purpose := path[0] - hdkeychain.HardenedKeyStart
+	coinType := path[1] - hdkeychain.HardenedKeyStart
+	account := path[2] - hdkeychain.HardenedKeyStart
+	branch := path[3]
+	index := path[4]
```
GustavoStingelin, line 1057: I am curious about why we are handling hardened indices this way. Is this behavior inherited from previous wallet logic or is it LND specific?

IIUC, [BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#extended-keys) describes the subtraction and addition of 2^31 purely as a notation convenience to make hardened indices more human-readable.

> "Each extended key has 2^31 normal child keys, and 2^31 hardened child keys. Each of these child keys has an index. The normal child keys use indices 0 through (2^31)-1. The hardened child keys use indices 2^31 through (2^32)-1. To ease notation for hardened key indices, a number iH represents i+2^31."

Also, it is worth clarifying that the requirement for the first three derivation levels to be hardened comes from BIP44 style paths, not from BIP32. BIP32 includes test vectors where those levels are not hardened.

wallet/psbt_manager_test.go
```diff
@@ -1774,3 +1782,1501 @@ func TestFundPsbtDecorateFailure(t *testing.T) {
...
+			err: ErrAmbiguousDerivation,
+		},
+		{
+			name: "no derivation info (valid state for " +
+				"validation check itself, returns false/nil)",
+			// Arrange: Input with no derivation information.
+			pInput:    &psbt.PInput{},
+			isTaproot: false,
+			err:       nil,
+		},
```
GustavoStingelin: Maybe these comments are redundant?

GustavoStingelin: In the other tests too, given that the test name is self-explanatory.

wallet/psbt_manager.go
```diff
@@ -710,3 +807,717 @@ func (w *Wallet) createTxIntent(intent *FundIntent) *TxIntent {
...
+	idx int, sigHashes *txscript.TxSigHashes,
+	tweaker PrivKeyTweaker) error {
+
+	// It is safe to access packet.Inputs[idx] directly here because
+	// SignPsbt calls psbt.InputsReadyToSign before this method, which
+	// ensures that the Inputs slice corresponds to the UnsignedTx inputs.
+	pInput := &packet.Inputs[idx]
+
+	// Fetch the UTXO (Witness or NonWitness) needed for signing.
+	utxo := fetchPsbtUtxo(packet, idx)
```
GustavoStingelin: Q: Will it be a problem if this returns nil?

GustavoStingelin (changes requested):
Left some comments

wallet/psbt_manager.go
yyforyongyu: correct they are used in the following PR #1130

wallet/psbt_manager.go
yyforyongyu: updated!

wallet/psbt_manager.go
yyforyongyu: updated!

wallet/psbt_manager.go
yyforyongyu: cool added a check

wallet/psbt_manager.go
```diff
@@ -710,3 +807,717 @@ func (w *Wallet) createTxIntent(intent *FundIntent) *TxIntent {
...
+			err = w.signTaprootPsbtInput(
+				ctx, packet, i, sigHashes,
+				params.InputTweakers[i],
+			)
+		} else {
+			err = w.signBip32PsbtInput(
+				ctx, packet, i, sigHashes,
+				params.InputTweakers[i],
+			)
+		}
```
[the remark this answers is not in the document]
yyforyongyu: `InputTweakers` is not a slice

wallet/psbt_manager.go
yyforyongyu: updated the comments - it's impractical to have this many inputs

wallet/psbt_manager.go
yyforyongyu: switched

wallet/psbt_manager.go
yyforyongyu: Make sense - updated.

wallet/psbt_manager.go
yyforyongyu: good q - refactored a bit, now `fetchPsbtUtxo` returns an error instead when it's nil, so the caller can handle it.

wallet/psbt_manager.go
yyforyongyu: > I am curious about why we are handling hardened indices this way. Is this behavior inherited from previous wallet logic or is it LND specific?

yeah here: https://github.com/lightningnetwork/lnd/blob/04f61080519930116a7cc1447d5da6124e341f3d/lnwallet/btcwallet/signer.go#L98

BIP32 defines the algorithm, and other BIPs, like 43/44/84/86 make use of it, so I think it makes sense to name it a BIP32 path, although we are only using the subset and require it to be BIP44-ish like, which are wallet-context that we enforce here. I think we can consider renaming it to BIP44 path, however, due to names like `KeyScopeBIP0044` first, as it will also be confusing, as by renaming it will look like it's parsing path for this specific key scope only.

GustavoStingelin (approved):
LGTM!

PR #1132, opened 2025-11-27
wallet: add store db constructors and safe-casting

GustavoStingelin:
## Change Description
  - Add PostgresWalletDB and SQLiteWalletDB structures that will implement the interface methods.
  - Add overflow-safe casting helpers for int64/int32/uint32/sql.NullInt32 plus explicit errors for overflow and invalid nulls.
  
## test
`go test ./wallet/internal/db/...`

### DIFF
--- a/wallet/internal/db/safecasting_test.go
+++ b/wallet/internal/db/safecasting_test.go
@@ -1,0 +1,236 @@
+package db
+
+import (
+	"database/sql"
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestInt64ToUint32 checks that an int64 value is converted to uint32 only
+// when it is non-negative and fits within the uint32 range. It should fail
+// loudly for any value outside those bounds.
+func TestInt64ToUint32(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		val     int64
+		want    uint32
+		wantErr bool
+	}{
+		{name: "zero", val: 0, want: 0},
+		{
+			name: "max uint32",
+			val:  int64(math.MaxUint32),
+			want: math.MaxUint32,
+		},
+		{name: "negative", val: -1, wantErr: true},
+		{
+			name:    "too large",
+			val:     int64(math.MaxUint32) + 1,
+			wantErr: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			got, err := int64ToUint32(tc.val)
+			if tc.wantErr {
+				require.ErrorIs(t, err, ErrCastingOverflow)
+				return
+			}
+
+			require.NoError(t, err)
+			require.Equal(t, tc.want, got)
+		})
+	}
+}
+
+// TestInt64ToInt32 checks that an int64 value is converted to int32 only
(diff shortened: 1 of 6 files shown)


wallet/internal/db/safecasting_test.go
```diff
@@ -0,0 +1,221 @@
...
+				return
+			}
+
+			require.NoError(t, err)
+			require.Equal(t, tc.want, got)
+		})
+	}
+}
+
+func TestInt64ToInt32(t *testing.T) {
```
mohamedawnallah, line 56: missing go doc comments but The test func name and scope is sufficiently clear to me. Adding godoc comments may not be necessary

wallet/internal/db/safecasting_test.go
mohamedawnallah, line 1: Nice comprehensive tests :+1:

mohamedawnallah (approved):
Thanks, this LGTM!

wallet/internal/db/safecasting_test.go
GustavoStingelin: added for compliance

GustavoStingelin:
@mohamedawnallah I added a new test file to guarantee that we never end up with a nil database connection. Please take a look to ack.

wallet/internal/db/safecasting_test.go
```diff
@@ -0,0 +1,236 @@
...
+// when it is non-negative and fits within the uint32 range. It should fail
+// loudly for any value outside those bounds.
+func TestInt64ToUint32(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		val     int64
+		want    uint32
+		wantErr bool
```
yyforyongyu, line 21: sth for future dev: we can define `errExpected error` or `errWanted error`, then pass nil for cases where errors are not expected, in the assertion we can use a single line `require.ErrorIs` without the logical branch.

wallet/internal/db/safecasting_test.go
```diff
@@ -0,0 +1,236 @@
...
+	}{
+		{name: "zero", val: 0, want: 0},
```
yyforyongyu, line 23: also style nit: we should use multi lines here, so it's easier to follow each setup

wallet/internal/db/safecasting_test.go
```diff
@@ -0,0 +1,236 @@
...
+	t.Parallel()
+
+	tests := []struct {
+		name    string
+		val     uint32
+		want    int32
+		wantErr bool
+	}{
+		{name: "zero", val: 0, want: 0},
+		{name: "max int32", val: math.MaxInt32, want: math.MaxInt32},
```
yyforyongyu, line 116: was wondering why we can just pass this to `val` directly, then realized this is just a const

wallet/internal/db/safecasting_test.go
```diff
@@ -0,0 +1,236 @@
...
+			if tc.wantErr != nil {
+				require.ErrorIs(t, err, tc.wantErr)
+				return
+			}
+
+			require.NoError(t, err)
+			require.Equal(t, tc.want, got)
+		})
+	}
+}
```
yyforyongyu, line 236: very cool - and this file is 100% covered!

yyforyongyu (approved):
💟

PR #1096, opened 2025-10-05
wtxmgr: design transactions manager SQL schema

mohamedawnallah:
## Upstream

https://github.com/btcsuite/btcwallet/issues/1094
https://github.com/btcsuite/btcwallet/issues/1015

## SQL Schema Design

Wallet Transactions Schema

https://dbdiagram.io/d/Btcwallet-Transactions-Schema-68e2c917d2b621e42259f15a

_Note: Indices will be mindfully added later where necessary (if any)_

## What is the original txs mapping for KV DB schema?

Seeing https://github.com/btcsuite/btcwallet/issues/1097.

### Testing
- [ ] Your PR passes all CI checks.
- [ ] Tests covering the positive and negative (error paths) are included.
- [ ] Bug fixes contain tests triggering the bug to prevent regressions.

### Code Style and Documentation
- [ ] The change is not [insubstantial](https://github.com/btcsuite/btcwallet/blob/master/docs/contribution_guidelines.md#substantial-contributions-only). Typo fixes are not accepted to fight bot spam.
- [ ] The change obeys the [Code Documentation and Commenting](https://github.com/btcsuite/btcwallet/blob/master/docs/contribution_guidelines.md#code-documentation-and-commenting) guidelines, and lines wrap at 80.
- [ ] Commits follow the [Ideal Git Commit Structure](https://github.com/btcsuite/btcwallet/blob/master/docs/contribution_guidelines.md#ideal-git-commit-structure).
- [ ] Any new logging statements use an appropriate subsystem and logging level.

📝 Please see our [Contribution Guidelines](https://github.com/btcsuite/btcwallet/blob/master/docs/contribution_guidelines.md) for further guidance.

PR #1134, opened 2025-12-01
[1/5] wallet: SQLize blocks table

mohamedawnallah:
## Change Description
SQLize blocks as part of https://github.com/btcsuite/btcwallet/pull/1096.

## Steps to Test
Steps for reviewers to follow to test the change.

### Testing

### DIFF
--- a/wallet/internal/db/block_pg.go
+++ b/wallet/internal/db/block_pg.go
@@ -1,8 +1,11 @@
 package db
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
+
+	sqlcpg "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/postgres"
 )
 
 // buildPgBlock constructs a Block from the given PostgreSQL block
@@ -16,4 +19,27 @@ func buildPgBlock(height sql.NullInt32, hash []byte,
 	}
 
 	return buildBlock(hash, height32, timestamp.Int64)
+}
+
+// ensureBlockExistsPg ensures that a block exists in the database.
+func ensureBlockExistsPg(ctx context.Context, qtx *sqlcpg.Queries,
+	block *Block) error {
+
+	height, err := uint32ToInt32(block.Height)
+	if err != nil {
+		return fmt.Errorf("convert block height: %w", err)
+	}
+
+	blockParams := sqlcpg.InsertBlockParams{
+		BlockHeight:    height,
+		HeaderHash:     block.Hash[:],
+		BlockTimestamp: block.Timestamp.Unix(),
+	}
+
+	err = qtx.InsertBlock(ctx, blockParams)
+	if err != nil {
+		return fmt.Errorf("insert block: %w", err)
+	}
+
+	return nil
 }
--- a/wallet/internal/db/queries/postgres/blocks.sql
+++ b/wallet/internal/db/queries/postgres/blocks.sql
@@ -8,7 +8,8 @@ WHERE block_height = $1;
 
 -- name: InsertBlock :exec
 INSERT INTO blocks (block_height, header_hash, block_timestamp)
-VALUES ($1, $2, $3);
+VALUES ($1, $2, $3)
+ON CONFLICT (block_height) DO NOTHING;
 
 -- name: DeleteBlock :exec
 DELETE FROM blocks
--- a/wallet/internal/db/queries/sqlite/blocks.sql
+++ b/wallet/internal/db/queries/sqlite/blocks.sql
@@ -7,7 +7,7 @@ FROM blocks
 WHERE block_height = ?;
 
 -- name: InsertBlock :exec
-INSERT INTO blocks (block_height, header_hash, block_timestamp)
+INSERT OR IGNORE INTO blocks (block_height, header_hash, block_timestamp)
 VALUES (?, ?, ?);
 
 -- name: DeleteBlock :exec
--- a/wallet/internal/db/wallet_pg.go
+++ b/wallet/internal/db/wallet_pg.go
@@ -185,6 +185,25 @@ func (w *PostgresWalletDB) UpdateWallet(ctx context.Context,
 	params UpdateWalletParams) error {
 
 	return w.ExecuteTx(ctx, func(qtx *sqlcpg.Queries) error {
+		// Insert blocks if needed.
+		if params.SyncedTo != nil {
+			err := ensureBlockExistsPg(ctx, qtx, params.SyncedTo)
+			if err != nil {
+				return fmt.Errorf("ensure synced block: %w",
+					err)
+			}
+		}
+
+		if params.BirthdayBlock != nil {
+			err := ensureBlockExistsPg(
+				ctx, qtx, params.BirthdayBlock,
+			)
+			if err != nil {
+				return fmt.Errorf("ensure birthday block: %w",
+					err)
+			}
+		}
+
 		// Build sync state update params directly from input.
 		// Only set fields that are being updated (leave others as nil).
 		syncParams := sqlcpg.UpdateWalletSyncStateParams{
--- a/wallet/internal/db/wallet_pg.go
+++ b/wallet/internal/db/wallet_pg.go
@@ -204,39 +204,9 @@ func (w *PostgresWalletDB) UpdateWallet(ctx context.Context,
 			}
 		}
 
-		// Build sync state update params directly from input.
-		// Only set fields that are being updated (leave others as nil).
-		syncParams := sqlcpg.UpdateWalletSyncStateParams{
-			WalletID: int64(params.WalletID),
-		}
-
-		if params.SyncedTo != nil {
-			syncedHeight, err := uint32ToNullInt32(
-				params.SyncedTo.Height,
-			)
-			if err != nil {
-				return err
-			}
-
-			syncParams.SyncedHeight = syncedHeight
-		}
-
-		if params.Birthday != nil {
-			syncParams.BirthdayTimestamp = sql.NullTime{
-				Time:  *params.Birthday,
-				Valid: true,
-			}
-		}
-
-		if params.BirthdayBlock != nil {
-			birthdayHeight, err := uint32ToNullInt32(
-				params.BirthdayBlock.Height,
-			)
-			if err != nil {
-				return err
-			}
-
-			syncParams.BirthdayHeight = birthdayHeight
+		syncParams, err := buildUpdateSyncParamsPg(params)
+		if err != nil {
+			return err
 		}
 
 		rowsAffected, err := qtx.UpdateWalletSyncState(ctx, syncParams)
@@ -368,4 +338,43 @@ func buildPgWalletInfo(row pgWalletRowParams) (*WalletInfo, error) {
 	}
 
 	return info, nil
+}
+
+// buildUpdateSyncParamsPg constructs the UpdateWalletSyncStateParams from
+// the given UpdateWalletParams.
+func buildUpdateSyncParamsPg(params UpdateWalletParams) (
+	sqlcpg.UpdateWalletSyncStateParams, error) {
+
+	syncParams := sqlcpg.UpdateWalletSyncStateParams{
+		WalletID: int64(params.WalletID),
+	}
+
+	if params.SyncedTo != nil {
+		syncedHeight, err := uint32ToNullInt32(params.SyncedTo.Height)
+		if err != nil {
+			return syncParams, err
+		}
+
+		syncParams.SyncedHeight = syncedHeight
+	}
+
+	if params.Birthday != nil {
+		syncParams.BirthdayTimestamp = sql.NullTime{
+			Time:  *params.Birthday,
+			Valid: true,
+		}
+	}
+
+	if params.BirthdayBlock != nil {
+		birthdayHeight, err := uint32ToNullInt32(
+			params.BirthdayBlock.Height,
+		)
+		if err != nil {
+			return syncParams, err
+		}
+
+		syncParams.BirthdayHeight = birthdayHeight
+	}
+
+	return syncParams, nil
 }
--- a/wallet/internal/db/wallet_sqlite.go
+++ b/wallet/internal/db/wallet_sqlite.go
@@ -206,32 +206,7 @@ func (w *SQLiteWalletDB) UpdateWallet(ctx context.Context,
 			}
 		}
 
-		// Build sync state update params directly from input.
-		// Only set fields that are being updated (leave others as nil).
-		syncParams := sqlcsqlite.UpdateWalletSyncStateParams{
-			WalletID: int64(params.WalletID),
-		}
-
-		if params.SyncedTo != nil {
-			syncParams.SyncedHeight = sql.NullInt64{
-				Int64: int64(params.SyncedTo.Height),
-				Valid: true,
-			}
-		}
-
-		if params.Birthday != nil {
-			syncParams.BirthdayTimestamp = sql.NullTime{
-				Time:  *params.Birthday,
-				Valid: true,
-			}
-		}
-
-		if params.BirthdayBlock != nil {
-			syncParams.BirthdayHeight = sql.NullInt64{
-				Int64: int64(params.BirthdayBlock.Height),
-				Valid: true,
-			}
-		}
+		syncParams := buildUpdateSyncParamsSqlite(params)
 
 		rowsAffected, err := qtx.UpdateWalletSyncState(ctx, syncParams)
 		if err != nil {
@@ -367,4 +342,37 @@ func buildSqliteWalletInfo(row sqliteWalletRowParams) (*WalletInfo, error) {
 	}
 
 	return info, nil
+}
+
+// buildUpdateSyncParamsSqlite constructs the UpdateWalletSyncStateParams from
+// the given UpdateWalletParams.
+func buildUpdateSyncParamsSqlite(
+	params UpdateWalletParams) sqlcsqlite.UpdateWalletSyncStateParams {
+
+	syncParams := sqlcsqlite.UpdateWalletSyncStateParams{
+		WalletID: int64(params.WalletID),
+	}
+
+	if params.SyncedTo != nil {
+		syncParams.SyncedHeight = sql.NullInt64{
+			Int64: int64(params.SyncedTo.Height),
+			Valid: true,
+		}
+	}
+
+	if params.Birthday != nil {
+		syncParams.BirthdayTimestamp = sql.NullTime{
+			Time:  *params.Birthday,
+			Valid: true,
+		}
+	}
+
+	if params.BirthdayBlock != nil {
+		syncParams.BirthdayHeight = sql.NullInt64{
+			Int64: int64(params.BirthdayBlock.Height),
+			Valid: true,
+		}
+	}
+
+	return syncParams
 }
--- a/wallet/internal/db/wallet_sqlite.go
+++ b/wallet/internal/db/wallet_sqlite.go
@@ -185,6 +185,27 @@ func (w *SQLiteWalletDB) UpdateWallet(ctx context.Context,
 	params UpdateWalletParams) error {
 
 	return w.ExecuteTx(ctx, func(qtx *sqlcsqlite.Queries) error {
+		// Insert blocks if needed.
+		if params.SyncedTo != nil {
+			err := ensureBlockExistsSqlite(
+				ctx, qtx, params.SyncedTo,
+			)
+			if err != nil {
+				return fmt.Errorf("ensure synced block: %w",
+					err)
+			}
+		}
+
+		if params.BirthdayBlock != nil {
+			err := ensureBlockExistsSqlite(
+				ctx, qtx, params.BirthdayBlock,
+			)
+			if err != nil {
+				return fmt.Errorf("ensure birthday block: %w",
+					err)
+			}
+		}
+
 		// Build sync state update params directly from input.
 		// Only set fields that are being updated (leave others as nil).
 		syncParams := sqlcsqlite.UpdateWalletSyncStateParams{
--- a/wallet/internal/db/block_sqlite.go
+++ b/wallet/internal/db/block_sqlite.go
@@ -1,8 +1,11 @@
 package db
 
 import (
+	"context"
 	"database/sql"
 	"fmt"
+
+	sqlcsqlite "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/sqlite"
 )
 
 // buildSqliteBlock constructs a Block from the given SQLite block
@@ -16,4 +19,25 @@ func buildSqliteBlock(height sql.NullInt64, hash []byte,
 	}
 
 	return buildBlock(hash, height32, timestamp.Int64)
+}
+
+// ensureBlockExistsSqlite ensures that a block exists in the database. If it
+// doesn't exist, it inserts it.
+func ensureBlockExistsSqlite(ctx context.Context, qtx *sqlcsqlite.Queries,
+	block *Block) error {
+
+	height := int64(block.Height)
+
+	blockParams := sqlcsqlite.InsertBlockParams{
+		BlockHeight:    height,
+		HeaderHash:     block.Hash[:],
+		BlockTimestamp: block.Timestamp.Unix(),
+	}
+
+	err := qtx.InsertBlock(ctx, blockParams)
+	if err != nil {
+		return fmt.Errorf("insert block: %w", err)
+	}
+
+	return nil
 }
--- a/wallet/internal/db/itest/wallet_store_test.go
+++ b/wallet/internal/db/itest/wallet_store_test.go
@@ -358,3 +358,64 @@ func TestUpdateWalletSecrets(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, newSecrets.EncryptedMasterHdPrivKey, seed)
 }
+
+// TestUpdateWallet_AutoBlockInsertion verifies that UpdateWallet automatically
+// inserts blocks when updating SyncedTo or BirthdayBlock.
+func TestUpdateWallet_AutoBlockInsertion(t *testing.T) {
+	t.Parallel()
+
+	store, _ := NewTestStore(t)
+
+	params := CreateWalletParamsFixture("auto-block-wallet")
+	created, err := store.CreateWallet(t.Context(), params)
+	require.NoError(t, err)
+
+	// Create a block WITHOUT pre-inserting it into the blocks table.
+	block := db.Block{
+		Height:    uint32(100),
+		Hash:      RandomHash(),
+		Timestamp: time.Now().UTC(),
+	}
+
+	// Update wallet with SyncedTo - should automatically insert the block.
+	updateParams := db.UpdateWalletParams{
+		WalletID: created.ID,
+		SyncedTo: &block,
+	}
+	err = store.UpdateWallet(t.Context(), updateParams)
+	require.NoError(t, err)
+
+	// Verify the wallet was updated.
+	retrieved, err := store.GetWallet(t.Context(), created.Name)
+	require.NoError(t, err)
+	require.NotNil(t, retrieved.SyncedTo)
+	require.Equal(t, block.Height, retrieved.SyncedTo.Height)
+	require.Equal(t, block.Hash, retrieved.SyncedTo.Hash)
+
+	// Update again with the same block - should be idempotent.
+	err = store.UpdateWallet(t.Context(), updateParams)
+	require.NoError(t, err, "updating with same block should be idempotent")
+
+	// Create another block for BirthdayBlock.
+	birthdayBlock := db.Block{
+		Height:    uint32(50),
+		Hash:      RandomHash(),
+		Timestamp: time.Now().UTC().Add(-time.Hour),
+	}
+
+	// Update wallet with BirthdayBlock - should automatically insert it.
+	updateParams = db.UpdateWalletParams{
+		WalletID:      created.ID,
+		BirthdayBlock: &birthdayBlock,
+	}
+	err = store.UpdateWallet(t.Context(), updateParams)
+	require.NoError(t, err)
+
+	// Verify both blocks are set.
+	retrieved, err = store.GetWallet(t.Context(), created.Name)
+	require.NoError(t, err)
+	require.NotNil(t, retrieved.SyncedTo)
+	require.NotNil(t, retrieved.BirthdayBlock)
+	require.Equal(t, block.Height, retrieved.SyncedTo.Height)
+	require.Equal(t, birthdayBlock.Height, retrieved.BirthdayBlock.Height)
+}
--- a/wallet/internal/db/sqlc/postgres/blocks.sql.go
+++ b/wallet/internal/db/sqlc/postgres/blocks.sql.go
@@ -38,6 +38,7 @@ func (q *Queries) GetBlockByHeight(ctx context.Context, blockHeight int32) (Bloc
 const InsertBlock = `-- name: InsertBlock :exec
 INSERT INTO blocks (block_height, header_hash, block_timestamp)
 VALUES ($1, $2, $3)
+ON CONFLICT (block_height) DO NOTHING
 `
 
 type InsertBlockParams struct {
--- a/wallet/internal/db/sqlc/sqlite/blocks.sql.go
+++ b/wallet/internal/db/sqlc/sqlite/blocks.sql.go
@@ -36,7 +36,7 @@ func (q *Queries) GetBlockByHeight(ctx context.Context, blockHeight int64) (Bloc
 }
 
 const InsertBlock = `-- name: InsertBlock :exec
-INSERT INTO blocks (block_height, header_hash, block_timestamp)
+INSERT OR IGNORE INTO blocks (block_height, header_hash, block_timestamp)
 VALUES (?, ?, ?)
 `
 


itest/block_store_test.go
```diff
@@ -0,0 +1,66 @@
+package itest
+
+import (
+	"context"
+	"time"
+
+	"github.com/btcsuite/btcwallet/wallet/db"
+	"github.com/btcsuite/btcwallet/walletdb"
+	. "github.com/onsi/ginkgo/v2"
```
yyforyongyu: I thought we won't use this for the current milestone? From previous [discussion](https://github.com/btcsuite/btcwallet/pull/1027#discussion_r2320751758).

wallet/db/block_store_kvdb.go
```diff
@@ -0,0 +1,378 @@
...
+	BlockHeight      int32
+	HeaderHash       []byte
+	Timestamp        int64
+	TransactionHashs [][]byte // Transaction hashes stored with this block
+}
+
+// kvdbBlockStore is an adapter that makes the legacy key-value database
+// (walletdb) implement the BlockStore interface. This allows gradual migration
+// from kvdb to SQL while maintaining a consistent interface.
+type kvdbBlockStore struct {
```
yyforyongyu: The scope should be sqlite and postgres only in this PR. These changes can be saved tho, as later we will implement the kvdb store too.

wallet/internal/db/data_types.go
```diff
@@ -180,20 +180,6 @@ type WalletInfo struct {
 	SyncedTo *Block
 }
 
-// Block defines a block's hash, height, and timestamp. This is used to
```
yyforyongyu, line 210: Why this change?

GustavoStingelin (changes requested):
Why move the database code out of the internal package? We placed it inside the internal exactly to prevent any usage outside the wallet package. This was meant to keep the boundary clear and avoid unintended imports.

GustavoStingelin:
just merged the #1136

wallet/internal/db/data_types.go
mohamedawnallah: Moved to relevant file instead of having bloated data_types.go file

Makefile
```diff
@@ -126,17 +126,17 @@ rpc-format:
...
 lint-check: lint-config-check
 	@$(call print, "Linting source.")
-	$(DOCKER_TOOLS) golangci-lint run -v $(LINT_WORKERS)
+	$(DOCKER_TOOLS) golangci-lint run -v --config config/.golangci.yml $(LINT_WORKERS)
 
 #? lint: Lint source and fix
 lint: lint-config-check
 	@$(call print, "Linting source.")
-	$(DOCKER_TOOLS) golangci-lint run -v --fix $(LINT_WORKERS)
+	$(DOCKER_TOOLS) golangci-lint run -v --fix --config config/.golangci.yml $(LINT_WORKERS)
```
yyforyongyu, line 157: Nice cleanup!

Makefile
```diff
@@ -6,6 +6,27 @@ GOINSTALL := GO111MODULE=on go install -v
...
+SQL_MIGRATIONS_DIR := wallet/internal/db/migrations
+SQL_QUERIES_DIR := wallet/internal/db/queries
+
+# SQL file paths.
+SQL_POSTGRES_MIGRATIONS := $(SQL_MIGRATIONS_DIR)/postgres/*.sql
+SQL_POSTGRES_QUERIES := $(SQL_QUERIES_DIR)/postgres/*.sql
+SQL_SQLITE_MIGRATIONS := $(SQL_MIGRATIONS_DIR)/sqlite/*.sql
+SQL_SQLITE_QUERIES := $(SQL_QUERIES_DIR)/sqlite/*.sql
+
+# SQL directories.
```
yyforyongyu: Duplicates?

Makefile
```diff
@@ -196,12 +258,19 @@ sqlc-check: sqlc
 	unit-race \
 	unit-debug \
 	unit-bench \
-	itest-db \
```
yyforyongyu, line 266: looks like it's deleted from rebase

config/sqlfluff.cfg
```diff
@@ -0,0 +1,41 @@
+# SQLFluff configuration
```
yyforyongyu, line 1: 👍

tools/Dockerfile
```diff
@@ -2,21 +2,26 @@ FROM golang:1.24.6-alpine
 
 ENV GOFLAGS="-buildvcs=false"
 
-RUN apk update && apk add --no-cache git bash
+RUN apk update && apk add --no-cache git bash python3 py3-pip py3-virtualenv
```
yyforyongyu: I don't think we should bloat this image, given that it's supposed to be go env. An easier and faster way is to use their image directly: https://hub.docker.com/r/sqlfluff/sqlfluff

Makefile
```diff
@@ -177,16 +198,57 @@ tidy-module:
...
+#? sql-compile: Generate sql models and queries in Go
+sql-compile: docker-tools sql-parse
 	@$(call print, "Generating sql models and queries in Go")
-	$(DOCKER_TOOLS) sqlc generate -f wallet/internal/db/sqlc/sqlc.yaml
+	$(DOCKER_TOOLS) sqlc generate -f config/sqlc.yaml
 
-#? sqlc-check: Make sure sql models and queries are up to date
-sqlc-check: sqlc
+#? sqlc: Alias for sql-compile
+sqlc: sql-compile
```
yyforyongyu: I think we need to find better names here, as `sql-xxx` and `sqlc-xxx` can easily be mixed up. We have the tool name `sqlc`, then we sort have a domain name `sql`, yet the `make sqlc` is calling `sql-compile` under the hood.

Makefile
```diff
@@ -177,16 +198,57 @@ tidy-module:
...
 
+#? sqlc-check: Alias for sql-compile-check
+sqlc-check: sql-compile-check
+
+#? sql-format: Format SQL migration and query files
+sql-format: docker-tools
+	@$(call print, "Formatting SQL files (postgres).")
+	$(DOCKER_TOOLS) bash -c "sqlfluff format --config config/sqlfluff.cfg -d postgres $(SQL_POSTGRES_MIGRATIONS) $(SQL_POSTGRES_QUERIES)"
+	@$(call print, "Formatting SQL files (sqlite).")
+	$(DOCKER_TOOLS) bash -c "sqlfluff format --config config/sqlfluff.cfg -d sqlite $(SQL_SQLITE_MIGRATIONS) $(SQL_SQLITE_QUERIES)"
```
yyforyongyu: What are the differences between `fix` and `format`?

Makefile
```diff
@@ -196,12 +258,19 @@ sqlc-check: sqlc
 	unit-race \
 	unit-debug \
 	unit-bench \
-	itest-db \
-	itest-db-race \
 	fmt \
 	fmt-check \
 	tidy-module \
 	tidy-module-check \
+	sql-parse \
```
yyforyongyu, line 272: It seems we have too many commands here, which are fragile and overlapping. After reading https://docs.sqlfluff.com/en/stable/reference/cli.html I think all we need is,
```make
#? sql-lint: Lint SQL files and fix issues (matches 'make lint' behavior)
sql-lint:
    @$(call print, "Fixing SQL files.")
    $(DOCKER_SQLFLUFF) fix --config config/sqlfluff.cfg $(SQL_MIGRATIONS_DIR) $(SQL_QUERIES_DIR)

#? sql-lint-check: Lint SQL files and report errors (matches 'make lint-check' behavior)
sql-lint-check:
    @$(call print, "Linting SQL files.")
    $(DOCKER_SQLFLUFF) lint --config config/sqlfluff.cfg $(SQL_MIGRATIONS_DIR) $(SQL_QUERIES_DIR)

#? sqlc: Generate Go code from SQL
sqlc:
    @$(call print, "Generating SQL models.")
    $(DOCKER_TOOLS) sqlc generate -f config/sqlc.yaml

#? sqlc-check: Verify generated SQL models are up-to-date
# This depends on sqlc (generate) and checks if git status is clean.
sqlc-check: sqlc
    @$(call print, "Verifying SQL code generation.")
    ...
    
```

yyforyongyu (changes requested):
Nice cleanups! Also like that we are using `sqlfluff` - my comments are,
- we should move the last 4 or 5 commits to a new PR, as that is actually the code for `[1/5] wallet: SQLize blocks table#1134`
- then we can keep reviewing this PR to get the `sqlfluff` quickly merged, such that others can also use it.

Makefile
mohamedawnallah: Addressed

Makefile
mohamedawnallah: Addressed in https://github.com/btcsuite/btcwallet/pull/1134#discussion_r2610504946

Makefile
mohamedawnallah: Addressed in https://github.com/btcsuite/btcwallet/pull/1134#discussion_r2610504946

tools/Dockerfile
mohamedawnallah: Addressed by using sqlfluff image

Makefile
mohamedawnallah: Addressed

Makefile
mohamedawnallah: Addressed

mohamedawnallah:
Will rebase after https://github.com/btcsuite/btcwallet/pull/1121 is merged

wallet/internal/db/block_common.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+// Block represents the block data returned from the database. This struct
+// normalizes the differences between PostgreSQL (which uses int32 for block
+// height) and SQLite (which uses int64) by always using int32.
+type Block struct {
+	BlockHeight int32
+	HeaderHash  []byte
+	CreatedAt   int64
+}
```
GustavoStingelin: I think we should reuse the existing `Block` struct defined in `wallet/internal/db/data_types.go`. Otherwise, we end up with two separate block representations that do not interoperate. If we feel strongly that the current structure is not ideal, then we should refactor the existing one rather than introducing a second type that serves the same purpose.

wallet/internal/db/block_common.go
```diff
@@ -0,0 +1,73 @@
...
+
+// BlockStore defines the interface for database block operations.
+type BlockStore interface {
+	// GetBlock retrieves a block by its height from the database.
+	GetBlock(ctx context.Context, height int32) (Block, error)
```
GustavoStingelin: Why not keep this in `wallet/internal/db/interface.go`? We can centralize everything in that file, making the API contract easier to find. If necessary, we can refactor and simplify the file later to keep it manageable.

wallet/internal/db/block_common.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// InsertBlock inserts a new block into the database.
+	InsertBlock(ctx context.Context, block Block) error
+
+	// DeleteBlock removes a block from the database by its height.
+	DeleteBlock(ctx context.Context, height int32) error
+
+	// DeleteBlocksFromHeight removes all blocks at the specified height and
+	// above.
+	DeleteBlocksFromHeight(ctx context.Context, height int32) error
```
GustavoStingelin: I'm not sure if this is already a standard naming pattern, but I think we could make it a bit more expressive to avoid misunderstandings. Something like `DeleteBlocksAtOrAboveHeight` or even `PruneAtHeight` might communicate the intent more clearly.

wallet/internal/db/block_pg.go
```diff
@@ -0,0 +1,128 @@
...
+type PostgresBlockStore struct {
+	queries *sqlcpg.Queries
+}
+
+// NewPostgresBlockStore creates a new PostgreSQL block store.
+func NewPostgresBlockStore(queries *sqlcpg.Queries) *PostgresBlockStore {
+	return &PostgresBlockStore{
+		queries: queries,
+	}
+}
```
GustavoStingelin: I think we could reuse the existing `PostgresWalletDB` and `SQLiteWalletDB` structs for all interface implementations. Splitting this into separate structs doesn’t seem to add much value, since the interface already provides that abstraction.

wallet/internal/db/block_pg.go
```diff
@@ -0,0 +1,128 @@
...
+func (s *PostgresBlockStore) GetBlock(ctx context.Context,
+	height int32) (Block, error) {
+
+	pgBlock, err := s.queries.GetBlockByHeight(ctx, height)
+	if err != nil {
+		return Block{}, fmt.Errorf("%w: %w", ErrBlockQuery, err)
+	}
+
+	return pgBlockToBlock(pgBlock), nil
+}
```
GustavoStingelin: I think the caller of `GetBlock` already understands that any returned error indicates a failure while retrieving the data. Because of that, wrapping the original error into an additional `ErrBlockQuery` type doesn’t seem to add meaningful context. In practice, checking `err != nil` versus checking `errors.Is(err, ErrBlockQuery)` doesn’t provide extra signal.

What _would_ add value is handling `sql.ErrNoRows` explicitly and returning a custom “block not found” error. That gives the caller more actionable and semantically clear information. This is similar to what I implemented here: https://github.com/btcsuite/btcwallet/blob/sql-wallet/wallet/internal/db/wallet_pg.go#L119

Applies to the other methods too.

wallet/internal/db/block_pg.go
```diff
@@ -0,0 +1,128 @@
...
+func (s *PostgresBlockStore) DeleteBlock(ctx context.Context,
+	height int32) error {
+
+	err := s.queries.DeleteBlock(ctx, height)
+	if err != nil {
+		return fmt.Errorf("%w: %w", ErrBlockDelete, err)
+	}
+
+	return nil
+}
```
GustavoStingelin: Shouldn’t we return an error when attempting to delete a non-existing block? For example, similar to what I did here: https://github.com/btcsuite/btcwallet/blob/sql-wallet/wallet/internal/db/wallet_pg.go#L226

This can help surface potential race conditions and allow the caller to handle them at a higher layer.

wallet/internal/db/block_common.go
```diff
@@ -0,0 +1,73 @@
...
+// BlockStore defines the interface for database block operations.
+type BlockStore interface {
+	// GetBlock retrieves a block by its height from the database.
+	GetBlock(ctx context.Context, height int32) (Block, error)
+
+	// GetBlockTime retrieves just the timestamp of a block by its height.
+	GetBlockTime(ctx context.Context, height int32) (int64, error)
+
+	// InsertBlock inserts a new block into the database.
+	InsertBlock(ctx context.Context, block Block) error
```
GustavoStingelin: What do you think about standardizing the insert methods to use the prefix `Create`? That way we avoid mixing different verbs in the interface for operations that mean the same thing.

wallet/internal/db/block_pg.go
```diff
@@ -0,0 +1,128 @@
...
+func (s *PostgresBlockStore) DeleteBlocksFromHeight(ctx context.Context,
+	height int32) error {
+
+	err := s.queries.DeleteBlocksFromHeight(ctx, height)
+	if err != nil {
+		return fmt.Errorf("%w: %w", ErrBlockDelete, err)
+	}
+
+	return nil
+}
```
GustavoStingelin: It could be useful to return the affected rows in this function.

wallet/internal/db/block_pg.go
```diff
@@ -0,0 +1,128 @@
...
+		}
+
+		if err != nil {
+			yield(Block{}, err)
+			return
+		}
+
+		yieldBlocks(pgBlocks, yield, pgBlockToBlock)
+	}
+}
```
GustavoStingelin: I think this still ends up loading all block rows into memory and then wrapping them in an iterator. Since the conversion inside the iterator is very cheap, just a simple mapping, there doesn’t seem to be a real performance benefit here. If a caller prefers the iterator pattern, they could easily build one from the returned slice themselves. For that reason, it might be simpler to return the slice directly.

wallet/internal/db/data_types.go
GustavoStingelin: Added a comment related to this. I think the original YY isn't about the location.

wallet/internal/db/itest/block_pg_test.go
```diff
@@ -0,0 +1,12 @@
...
+package itest
+
+import (
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
+)
+
+// newBlockStoreFromTestStore creates a BlockStore from a PostgresWalletDB.
+func newBlockStoreFromTestStore(store *db.PostgresWalletDB) db.BlockStore {
+	return db.NewPostgresBlockStore(store.Queries())
+}
```
GustavoStingelin: wouldn't need this if we use only the `PostgresWalletDB.`

wallet/internal/db/itest/utils_test.go
```diff
@@ -0,0 +1,28 @@
...
+	hash := make([]byte, n)
+	_, _ = rand.Read(hash)
+
+	return hash
+}
+
+// randomHash32 generates a random 32-byte hash.
+func randomHash32() []byte {
+	return randomHashN(32)
+}
```
GustavoStingelin: already implemented at wallet/internal/db/itest/fixtures_test.go

wallet/internal/db/queries/postgres/blocks.sql
```diff
@@ -1,12 +1,46 @@
...
 
 -- name: InsertBlock :exec
-INSERT INTO blocks (block_height, header_hash, timestamp)
-VALUES ($1, $2, $3);
+INSERT INTO blocks (block_height, header_hash, created_at)
+VALUES ($1, $2, $3)
+ON CONFLICT (block_height) DO UPDATE
+    SET
+        header_hash = excluded.header_hash,
+        created_at = excluded.created_at;
```
GustavoStingelin: Is this intended to be an insert or an upsert? I wonder if it would be better to keep these operations separate and let an insert fail loudly when attempting to insert the same block twice. Ideally, the application should avoid this kind of race condition at the database layer, since allowing it could lead to an inconsistent state across goroutines.

wallet/internal/db/queries/sqlite/blocks.sql
```diff
@@ -1,12 +1,46 @@
...
 
 -- name: InsertBlock :exec
-INSERT INTO blocks (block_height, header_hash, timestamp)
-VALUES (?, ?, ?);
+INSERT INTO blocks (block_height, header_hash, created_at)
+VALUES (?, ?, ?)
+ON CONFLICT (block_height) DO UPDATE
+SET
+    header_hash = excluded.header_hash,
+    created_at = excluded.created_at;
```
GustavoStingelin: same here

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+	store, _ := NewTestStore(t)
+
+	// Access the queries field to create a BlockStore. The implementation
+	// differs based on build tags.
+	blockStore := newBlockStoreFromTestStore(store)
+
+	// Clean up all blocks after test completes.
+	t.Cleanup(func() {
+		_ = blockStore.DeleteBlocksFromHeight(t.Context(), 0)
+	})
```
GustavoStingelin: Why do we need this, given that we create a fresh database per test?

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+// TestBlockInsertAndGet tests inserting and retrieving blocks.
+func TestBlockInsertAndGet(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	s := getBlockStore(t)
+
+	testCases := AllBlockTestCases()
+
+	for _, tc := range testCases {
+		tc := tc
```
GustavoStingelin:
```suggestion
```
This is an old hack that is not needed in the new Go version.

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+		_ = blockStore.DeleteBlocksFromHeight(t.Context(), 0)
+	})
+
+	return blockStore
+}
+
+// TestBlockInsertAndGet tests inserting and retrieving blocks.
+func TestBlockInsertAndGet(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
GustavoStingelin:
```suggestion
```
should use `t.Context()`

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+			if tc.ExpectedError != nil {
+				require.Error(
+					t, err, "Expected error for: %s",
+					tc.Name,
+				)
+				require.True(
+					t, errors.Is(err, tc.ExpectedError),
+					"Expected error type %v, got %v",
+					tc.ExpectedError, err,
+				)
```
GustavoStingelin: `require` pkg has an API for `ErrorIs`

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+				require.Equal(t, tc.Block, retrieved)
+			}
+		})
+	}
+}
+
+// TestBlockGetBlockTime tests retrieving block timestamps.
+func TestBlockGetBlockTime(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
GustavoStingelin: `t.Context()`

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// TestBlockDelete tests deleting a specific block.
+func TestBlockDelete(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
+	s := getBlockStore(t)
+
+	// Clean up before test.
+	_ = s.DeleteBlocksFromHeight(ctx, 0)
```
GustavoStingelin: I think this is not necessary.

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+
+		// Verify it's deleted.
+		_, err := s.GetBlock(ctx, heightToDelete)
+		require.Error(t, err)
+		require.True(t, errors.Is(err, db.ErrBlockQuery))
+
+		// Verify other blocks still exist.
+		_, err = s.GetBlock(ctx, testCases[0].Block.BlockHeight)
+		require.NoError(t, err)
+	}
```
GustavoStingelin: I think we should require the number of test cases to be greater than 5 to make the assumption explicit and avoid the test being skipped unintentionally. Also will simplify the logic.

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// TestBlockDeleteFromHeight tests deleting blocks from a specified height.
+func TestBlockDeleteFromHeight(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
GustavoStingelin: `t.Context()`

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+	require.True(t, errors.Is(err, db.ErrBlockQuery))
+	_, err = s.GetBlock(ctx, 500)
+	require.Error(t, err)
+	require.True(t, errors.Is(err, db.ErrBlockQuery))
+}
+
+// TestBlockIterateAscending tests iterating blocks in ascending order.
+func TestBlockIterateAscending(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
GustavoStingelin: `t.Context()`

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+			CreatedAt:   b.time,
+		})
+		require.NoError(t, err)
+	}
+
+	var retrieved []db.Block
+	for block, err := range s.IterateBlocks(ctx, 200, false) {
+		require.NoError(t, err)
+		retrieved = append(retrieved, block)
+	}
```
GustavoStingelin: Q: How many real calls will really use the iterator, and how many will need to transform to a slice after?

wallet/internal/db/itest/block_test.go
```diff
@@ -0,0 +1,297 @@
...
+	require.Equal(t, int32(200), retrieved[0].BlockHeight)
+	require.Equal(t, int32(300), retrieved[1].BlockHeight)
+	require.Equal(t, int32(400), retrieved[2].BlockHeight)
+	require.Equal(t, int32(500), retrieved[3].BlockHeight)
+}
+
+// TestBlockIterateDescending tests iterating blocks in descending order.
+func TestBlockIterateDescending(t *testing.T) {
+	t.Parallel()
+	ctx := context.Background()
```
GustavoStingelin: `t.Context()`

GustavoStingelin (changes requested):
left some comments, plus the #1121

wallet/internal/db/queries/postgres/blocks.sql
mohamedawnallah: Removed the `ON CONFLICT` part

wallet/internal/db/queries/sqlite/blocks.sql
mohamedawnallah: Addressed

wallet/internal/db/block_common.go
mohamedawnallah: Addressed

wallet/internal/db/block_common.go
mohamedawnallah: Renamed it to `DeleteBlocksFromHeightOnwards`

wallet/internal/db/data_types.go
mohamedawnallah: Addressed

wallet/internal/db/block_pg.go
mohamedawnallah: It doesn't look like the deleted entries will be used in the current codebase + it contributes to memory usage unnecessarily. Will address the comment, given concrete evidence on what makes it necessary

wallet/internal/db/itest/block_test.go
mohamedawnallah: Removed

wallet/internal/db/itest/block_test.go
mohamedawnallah: Replied in relevant comment

wallet/internal/db/block_common.go
mohamedawnallah: Addressed

wallet/internal/db/block_common.go
mohamedawnallah: Addressed

wallet/internal/db/block_common.go
mohamedawnallah: Updated

wallet/internal/db/block_common.go
mohamedawnallah: Renamed it to `DeleteBlocksFromHeightOnwards`

wallet/internal/db/block_pg.go
mohamedawnallah: Addressed

wallet/internal/db/block_pg.go
mohamedawnallah: > What would add value is handling sql.ErrNoRows explicitly and returning a custom “block not found” error. That gives the caller more actionable and semantically clear information. This is similar to what I implemented here: https://github.com/btcsuite/btcwallet/blob/sql-wallet/wallet/internal/db/wallet_pg.go#L119

Not much agree with this. Returning low-level error to the application layer doesn't seem handy. Let's take @yyforyongyu's take on it as well

wallet/internal/db/block_pg.go
mohamedawnallah: Hesitant to do so. The delete operation supposed to be idempotent. It mirrors many software behaviors like Databases when someone remove non-existing entity it returns success. Let's take @yyforyongyu's take on it as well

wallet/internal/db/block_pg.go
mohamedawnallah: It doesn't look like the deleted entries will be used in the current codebase + it contributes to memory usage unnecessarily. Will address the comment, given concrete evidence on what makes it necessary

wallet/internal/db/block_pg.go
mohamedawnallah: It mirrors the access patterns here:
https://github.com/btcsuite/btcwallet/blob/81113d4438d6fb6bbd7510b0df5cc40642eda3f7/wtxmgr/query.go#L324

GustavoStingelin: @yyforyongyu, just to give your view here too, later

wallet/internal/db/block_pg.go
mohamedawnallah: Handled both

wallet/internal/db/interface.go
```diff
@@ -221,3 +222,31 @@ type UTXOStore interface {
 	Balance(ctx context.Context, params BalanceParams) (
 		btcutil.Amount, error)
 }
+
+// BlockStore defines the interface for database block operations.
+type BlockStore interface {
```
yyforyongyu: Why do we need this interface? I think the `Wallet` doesn't interact with blocks table directly? The blocks are created when we sync the wallet, and deleted when we rollback, which are already defined.

wallet/internal/db/queries/postgres/blocks.sql
```diff
@@ -13,3 +13,30 @@ VALUES ($1, $2, $3);
...
+-- name: GetBlocksFromHeight :many
+SELECT
+    block_height,
+    header_hash,
+    block_timestamp
+FROM blocks
+WHERE block_height >= $1
+ORDER BY block_height ASC;
+
+-- name: GetBlocksFromHeightDesc :many
```
yyforyongyu: 👍

wallet/internal/db/queries/postgres/blocks.sql
```diff
@@ -13,3 +13,30 @@ VALUES ($1, $2, $3);
...
+-- name: GetBlockTime :one
+SELECT block_timestamp
+FROM blocks
+WHERE block_height = $1;
+
+-- name: DeleteBlocksFromHeight :exec
+DELETE FROM blocks
+WHERE block_height >= $1;
+
+-- name: GetBlocksFromHeight :many
```
yyforyongyu: We need to pass a `LIMIT` here to avoid OOM?

wallet/internal/db/block_pg.go
yyforyongyu: Yeah I think deletion is usually idempotent, but also case-dependent. I think here it should be idempotent to keep things simple - if for w/e reason the caller needs to make sure the block is there before deleting, they can always perform a get first.

wallet/internal/db/block_pg.go
yyforyongyu: Iterator is only helpful when combined with pagination so we can perform lazy loading without loading everything into memory first. Also just practically speaking I don't think we need this interface, nor this iterating method.

yyforyongyu (changes requested):
Main comments are we should add a limit in queries and the wallet doesn't need to interact with the `blocks`.

wallet/internal/db/block_pg.go
mohamedawnallah: > Iterator is only helpful when combined with pagination so we can perform lazy loading without loading everything into memory first. Also just practically speaking I don't think we need this interface, nor this iterating method.

@yyforyongyu, you mean the iteration design pattern would be removed from the current kvdb code (`rangeBlockTransactions` - https://github.com/btcsuite/btcwallet/blob/81113d4438d6fb6bbd7510b0df5cc40642eda3f7/wtxmgr/query.go#L286-L324)?

If that iterator pattern would be removed from the current codebase, we would not need to have this iterator pattern in `IterateBlocks` for sure

If it wouldn't be removed from the current codebase, we need that `IterateBlocks` otherwise it would make the job of interfacing the `rangeBlockTransactions` API difficult down the line (to adapt for SQL and KVDB stores)

mohamedawnallah: I have tried to do initial refactoring  for `rangeBlockTransactions` to adapt for KVDB and SQL without that iterator design pattern in SQL it would be very difficult to have unified interface across both (since KVDB currently uses that iterator pattern)

mohamedawnallah: We need this iterator pattern for https://github.com/btcsuite/btcwallet/pull/1134#discussion_r2622603089 otherwise we can get potential OOM error

yyforyongyu: > If that iterator pattern would be removed from the current codebase, we would not need to have this iterator pattern in IterateBlocks for sure

Nah I was referring to the iterator in general, in case we want to use it in this milestone. It should provide lazy loading; otherwise, it's not worth the increased complexity.

mohamedawnallah: Yep, I am working locally now on changes for lazy loading so we can support batch size and limit query you mentioned here https://github.com/btcsuite/btcwallet/pull/1134#discussion_r2622603089

wallet/internal/db/interface.go
mohamedawnallah: > Why do we need this interface? I think the `Wallet` doesn't interact with blocks table directly? The blocks are created when we sync the wallet, and deleted when we rollback, which are already defined.

Can you expand? The block store has methods that would be mainly used in wtxmgr e.g:
https://github.com/btcsuite/btcwallet/blob/7bd08f6cc395e8e2c5289ce8214d2e838b49cbd7/wtxmgr/tx.go#L419 

My understanding as well that later when we work on store interface we implement the corresponding ones for KVDB. With this interface, we can just implement the equivalent for KVDB.

Also I am not sure without having BlockStore interface how are we supposed to proceed in sqlization/testing across multiple backends? Since currently the patterns looks like:
```
// Ensure PostgresWalletDB satisfies the WalletStore interface.
var _ WalletStore = (*PostgresWalletDB)(nil)

// Ensure PostgresWalletDB satisfies the BlockStore interface.
var _ BlockStore = (*PostgresWalletDB)(nil)

....
````

File: wallet/internal/db/queries/postgres/blocks.sql (hunk 38 above)
Comment (mohamedawnallah) [reply wallet/internal/db/queries/postgres/blocks.sql]: Great point, addressed

File: wallet/internal/db/block_common.go [hunk 39]
```diff
@@ -1,23 +1,59 @@
...
+	//         - Actual hash data:        32 bytes
+	//     - BlockTimestamp (int64):       8 bytes
+	//   ------------------------------------------------
+	//   Total per block:                 72 bytes
+	//   Total per batch:                 72 × 4096 = ~295 KB
+	//
+	// For Bitcoin mainnet (~875,000 blocks), this results in ~214 database
+	// queries to iterate the entire chain, with a peak memory usage of
+	// ~295 KB per batch instead of ~63 MB for loading everything at once.
+	BlockIteratorBatchSize = 4096 // 2^12
```
mohamedawnallah: @yyforyongyu, This is introduced to mainly address this comment https://github.com/btcsuite/btcwallet/pull/1134#discussion_r2622603089 and with that batch size it makes the iterator pattern comes to live. Having batch size for blocks is important since e.g mainnet has ~1million blocks and without lazy/batch loading this can negatively impact the peak memory usage

mohamedawnallah (commented):
@yyforyongyu – PTAL if there is something need to be changed/clarified

wallet/internal/db/block_pg.go
```diff
@@ -1,19 +1,164 @@
...
+	}
+
+	return timestamp, nil
+}
+
+// CreateBlock creates a new block into the database.
+func (w *PostgresWalletDB) CreateBlock(ctx context.Context, block Block) error {
+	params := sqlcpg.InsertBlockParams{
+		//nolint:gosec // Block height is within int32 range.
+		BlockHeight:    int32(block.Height),
```
GustavoStingelin: Why not use the safe casting helpers here? That would guarantee we avoid a panic and make the behavior more defensive. Applies to all `gosec`.

wallet/internal/db/block_common.go
GustavoStingelin: Maybe we should perform the calculation using the Block struct from interface.go? That would help normalize behavior across databases and make it explicit that the underlying representation can vary by backend. Also, that Block struct is only 64 bytes and fits neatly within a single CPU cache line without the slice pointer to the hash.

wallet/internal/db/itest/block_store_test.go
```diff
@@ -0,0 +1,466 @@
...
+	t.Parallel()
+	ctx := t.Context()
+	store, _ := NewTestStore(t, nil)
+
+	testCases := AllValidBlockTestCases()
+
+	// Insert a few blocks.
+	for i := 0; i < 3 && i < len(testCases); i++ {
+		require.NoError(t, store.CreateBlock(ctx, testCases[i].Block))
+	}
```
GustavoStingelin: I think we should explicitly require at least three test cases, rather than silently skipping when they are missing.

```suggestion
    require.GreaterOrEqual(t, len(testCases), 3)
    
	// Insert a few blocks.
	for i := 0; i < 3; i++ {
		require.NoError(t, store.CreateBlock(ctx, testCases[i].Block))
	}
```

wallet/internal/db/block_pg.go
```diff
@@ -1,19 +1,164 @@
...
+
+	return func(yield func(Block, error) bool) {
+		currentH := startHeight
+
+		for {
+			var (
+				blocks []sqlcpg.Block
+				err    error
+				//nolint:gosec
+				batchSize = int32(w.BlockIteratorBatchSize)
```
GustavoStingelin: Maybe the `BlockIteratorBatchSize` could already be an int32

wallet/internal/db/block_pg.go
```diff
@@ -1,19 +1,164 @@
...
+// If reverse is false, iterates in ascending order from startHeight upward.
+func (w *PostgresWalletDB) IterateBlocks(ctx context.Context, startHeight int32,
+	reverse bool) iter.Seq2[Block, error] {
+
+	return func(yield func(Block, error) bool) {
+		currentH := startHeight
+
+		for {
+			var (
+				blocks []sqlcpg.Block
```
GustavoStingelin: Cool, this is the correct way to use the iterator. Now we avoid bloating memory by fetching the entire database at once.

wallet/internal/db/itest/pg_test.go
```diff
@@ -167,12 +167,20 @@ func NewPostgresDB(t *testing.T) *sql.DB {
...
+func NewTestStore(t *testing.T,
+	params *db.WalletDBParams) (*db.PostgresWalletDB, *sql.DB) {
+
 	t.Helper()
 
+	if params == nil {
+		params = &db.WalletDBParams{
+			BlockIteratorBatchSize: db.BlockIteratorBatchSize,
+		}
+	}
```
GustavoStingelin: We could use the optional pattern here, but I think this is fine for now.

GustavoStingelin (changes requested):
Looks almost done! Left a few comments

wallet/internal/db/block_pg.go
GustavoStingelin: ok, let's see in the future if we will need this

wallet/internal/db/block_pg.go
GustavoStingelin: ok, let's see in the future if we will need this

wallet/internal/db/block_pg.go
GustavoStingelin: I also noticed that this reference requires the transaction as well. That means we will likely need to combine or refactor this query later to keep it both performant and complete.

wallet/internal/db/block_pg.go
mohamedawnallah: Updated

mohamedawnallah: Updated

wallet/internal/db/block_pg.go
mohamedawnallah: Updated

wallet/internal/db/block_common.go
mohamedawnallah: Made the comment concise by saying less words

wallet/internal/db/itest/block_store_test.go
mohamedawnallah: Updated

wallet/internal/db/itest/block_store_test.go
```diff
@@ -0,0 +1,468 @@
...
+// TestBlockInsertAndGet tests inserting and retrieving blocks.
+func TestBlockInsertAndGet(t *testing.T) {
+	t.Parallel()
+	ctx := t.Context()
+	store, _ := NewTestStore(t, nil)
+
+	testCases := AllBlockTestCases()
+
+	for _, tc := range testCases {
+		tc := tc
```
GustavoStingelin:
```suggestion
```
nit: This line is not needed

wallet/internal/db/itest/block_store_test.go
GustavoStingelin: nit: could remove the && condition in the for loop, as now it's redundant

GustavoStingelin (approved):
LGTM!

wallet/internal/db/itest/block_store_test.go
mohamedawnallah: Removed

wallet/internal/db/itest/block_store_test.go
mohamedawnallah: Removed

wallet/internal/db/itest/block_store_test.go
```diff
@@ -0,0 +1,467 @@
...
+					ctx, int32(tc.Block.Height),
+				)
+				require.NoError(t, err)
+				require.Equal(t, tc.Block, retrieved)
+			}
+		})
+	}
+}
+
+// TestGetBlock_NotFound tests the case where a block is not found.
```
yyforyongyu:
```suggestion
// TestGetBlockNotFound tests the case where a block is not found.
```

wallet/internal/db/itest/block_data_test.go
```diff
@@ -0,0 +1,180 @@
...
+			},
+			ExpectedError: db.ErrBlockInsert,
+		},
+		{
+			Name: "Minimum values (height 10, epoch timestamp)",
+			Block: db.Block{
+				Height:    10,
+				Hash:      RandomHash(),
+				Timestamp: time.Unix(0, 0),
+			},
```
yyforyongyu: Why is there no `ExpectedError`?

wallet/internal/db/itest/block_store_test.go
```diff
@@ -0,0 +1,467 @@
...
+	t.Parallel()
+	ctx := t.Context()
+	store, _ := NewTestStore(t, nil)
+
+	testCases := AllBlockTestCases()
+
+	for _, tc := range testCases {
+		t.Run(tc.Name, func(t *testing.T) {
+			err := store.CreateBlock(ctx, tc.Block)
+
```
yyforyongyu: nits:
- `require.ErrorIs(t, err, tc.ExpectedError)` can be done without the if `tc.ExpectedError` check
- we can call `continue` inside `if tc.ExpectedError != nil {`, so we save some indentations.

wallet/internal/db/itest/block_store_test.go
```diff
@@ -0,0 +1,467 @@
...
+// TestGetBlock_NotFound tests the case where a block is not found.
+func TestGetBlockNotFound(t *testing.T) {
+	t.Parallel()
+	ctx := t.Context()
+	store, _ := NewTestStore(t, nil)
+
+	// Attempt to retrieve a block that does not exist.
+	_, err := store.GetBlock(ctx, 999999)
+	require.Error(t, err)
```
yyforyongyu: redundant

wallet/internal/db/itest/block_store_test.go
```diff
@@ -0,0 +1,467 @@
...
+	t.Parallel()
+	ctx := t.Context()
+	store, dbConn := NewTestStore(t, nil)
+
+	// Close the database connection to force a query error.
+	require.NoError(t, dbConn.Close())
+
+	// Attempt to retrieve a block, which should now fail.
+	_, err := store.GetBlock(ctx, 1)
+	require.Error(t, err)
```
yyforyongyu: same here - redundant line

wallet/internal/db/itest/block_store_test.go
```diff
@@ -0,0 +1,467 @@
...
+	const numBlocks = 5
+	require.GreaterOrEqual(t, len(testCases), numBlocks,
+		"Need at least %d test cases for this test", numBlocks)
+
+	// Insert first numBlocks blocks.
+	for i := 0; i < numBlocks; i++ {
+		require.NoError(t, store.CreateBlock(ctx, testCases[i].Block))
+	}
+
+	// Delete the middle block.
```
yyforyongyu: Q: Why this approach? I think for unit tests, we can just call delete and check it's deleted, that's all we need

wallet/internal/db/itest/block_store_test.go
```diff
@@ -0,0 +1,467 @@
...
+	middleIdx := numBlocks / 2
+	heightToDelete := int32(testCases[middleIdx].Block.Height)
+	require.NoError(t, store.DeleteBlock(ctx, heightToDelete))
+
+	// Verify it's deleted.
+	_, err := store.GetBlock(ctx, heightToDelete)
+	require.Error(t, err)
+	require.ErrorIs(t, err, db.ErrBlockNotFound)
```
yyforyongyu: All these can be removed and keep only the `require.ErrorsIs`
```go
require.Error(t, err)
require.ErrorIs(t, err, db.ErrBlockNotFound)
```

wallet/internal/db/block_pg.go
```diff
@@ -1,19 +1,179 @@
...
+
+// GetBlock retrieves a block by its height from the database.
+func (w *PostgresWalletDB) GetBlock(ctx context.Context,
+	height int32) (Block, error) {
+
+	pgBlock, err := w.queries.GetBlockByHeight(ctx, height)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return Block{}, fmt.Errorf("%w: height=%d",
+				ErrBlockNotFound, height)
```
yyforyongyu: 👍

wallet/internal/db/block_pg.go
```diff
@@ -1,19 +1,179 @@
...
+		return fmt.Errorf("%w: %w", ErrBlockDelete, err)
+	}
+
+	return nil
+}
+
+// IterateBlocks returns an iterator over blocks starting from the given height.
+// If reverse is true, iterates in descending order from startHeight downward.
+// If reverse is false, iterates in ascending order from startHeight upward.
+func (w *PostgresWalletDB) IterateBlocks(ctx context.Context, startHeight int32,
```
yyforyongyu: What happens if two goroutines access the `blocks` at the same time? will there be a risk of a phantom read?

wallet/internal/db/interface.go
```diff
@@ -221,3 +222,31 @@ type UTXOStore interface {
...
+	GetBlock(ctx context.Context, height int32) (Block, error)
+
+	// GetBlockTime retrieves just the timestamp of a block by its height.
+	GetBlockTime(ctx context.Context, height int32) (int64, error)
+
+	// CreateBlock creates a new block into the database.
+	CreateBlock(ctx context.Context, block Block) error
+
+	// DeleteBlock removes a block from the database by its height.
+	DeleteBlock(ctx context.Context, height int32) error
```
yyforyongyu: If we want to have this interface, can we rework the methods here to respect the `xxxParams` and `xxxQueries` pattern? I think we can consolidate these methods, and we are missing a "list blocks" so the caller can decide which methods to use.

wallet/internal/db/interface.go
yyforyongyu: > The block store has methods that would be mainly used in wtxmgr e.g:

This is the `w.queries` level method, not the `Store` interface level methods which are used by the `wallet` - think of it like this, whenever you see a method used in `wtxmgr`, it is a new SQL query method, not the `Store` method here, which are used by the wallet only.

yyforyongyu (changes requested):
We don't need a store interface for every table that we build in SQL. I mean, the `wallets` table uses tables like `wallet_sync_states` and `wallet_secrets`, but we don't expose a store for these tables. This is also why the `blocks` table is a blocker for @GustavoStingelin before - the `wallets` table also uses the `blocks` table, or specifically, the interface method `WalletStore` here,
https://github.com/btcsuite/btcwallet/blob/7bd08f6cc395e8e2c5289ce8214d2e838b49cbd7/wallet/internal/db/interface.go#L49
needs to access the `w.queries.GetBlockByHeight` or `w.queries.InsertBlock` in this method - sort like what the old method does here(tho it should be made better)
https://github.com/btcsuite/btcwallet/blob/7bd08f6cc395e8e2c5289ce8214d2e838b49cbd7/waddrmgr/db.go#L2207-L2208

then in the call, we can just call this `UpdateWallet` to perform two tasks for this scenario,
- insert the block into the table `blocks`
- update the wallet's sync info to point to this block height

In code, it will look sth like this (this is an old draft, so the names were not updated, but you get the idea),
https://github.com/yyforyongyu/btcwallet/blob/06cdd0a2e979fec11e6a3650e28a2c6e2a5f79fa/wallet/wallet.go#L573-L581

And we do NOT, say, let the wallet,
- insert the block in one db tx
- update the wallet sync in another db tx

As we can leverage the atomic db operation here - this is also the essence of the `blocks` table, as the block data will not be useful unless the wallet's sync operation is successful.

So what it should be looking like,
- at the bottom layer, we have all these atomic CRUD operations, generated from the sqlc, and accessed via `w.queries`
- at the up layer, we have the defined `Store` interfaces, and provide high layer DB operations by using the CRUD operations.

However, the addition of the `BlockStore` interface seems to leak implementation details, like the existence of a separate `blocks` table, to the wallet application layer. As I'm about to finish the `Controller` interface, I've gained more knowledge about which complex db queries we may need, and we need to build these queries in one db tx for correctness and performance.

Re the testing across multiple backends without this interface - testing strategy should NOT dictate the public architecture. There are many ways we can verify the correctness, like building unit tests inside the package - imagine we build the `iterateBlocks` as a common helper method, then the helper method can have unit tests. We can also verify it through behavioral testing of the public interfaces (e.g., verifying that `RollbackToBlock` correctly cleans up the underlying table).


So my verdict is that we should not include this as an interface - the wallet should never micro-manage the `blocks` table. **However**, given that the agreement is, we will ensure changes very soon, due to the `Store` interface is a draft version, I'm ok with including it for now, and see if it's indeed the case that we don't need it later.

Meanwhile, still have a few comments for the interface, which should be fixed before merging.

wallet/internal/db/block_common.go
```diff
@@ -1,23 +0,0 @@
...
-import (
-	"fmt"
-	"time"
-
-	"github.com/btcsuite/btcd/chaincfg/chainhash"
-)
-
-// buildBlock constructs a Block from the provided components that are common
-// across different database backends.
-func buildBlock(hash []byte, height uint32, timestamp int64) (*Block, error) {
```
yyforyongyu, line 12: it's nice to have this helper, why deleting it tho?

wallet/internal/db/block_pg.go
```diff
@@ -1,19 +1,28 @@
...
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	sqlcpg "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/postgres"
 )
 
-// buildPgBlock constructs a Block from the given PostgreSQL block
-// fields.
-func buildPgBlock(height sql.NullInt32, hash []byte,
-	timestamp sql.NullInt64) (*Block, error) {
+// pgBlockToBlock converts a PostgreSQL block to a Block.
+func pgBlockToBlock(pgBlock sqlcpg.Block) (Block, error) {
```
yyforyongyu: Unnecessary renaming? These methods were recently introduced and approved. Changing them now creates unnecessary churn and diff noise without providing functional benefit. We should aim for minimal changes and respect existing conventions unless they are broken.

wallet/internal/db/wallet_sqlite.go
```diff
@@ -319,30 +319,38 @@ func buildSqliteWalletInfo(row sqliteWalletRowParams) (*WalletInfo, error) {
...
 		if err != nil {
-			return nil, fmt.Errorf("birthday block: %w", err)
+			return nil, fmt.Errorf(
+				"failed to convert birthday block: %w", err,
+			)
 		}
 
-		info.BirthdayBlock = block
+		info.BirthdayBlock = &block
 	}
```
yyforyongyu, line 375: It's unclear to me about this diff - can you update the commit to explain it a bit?

wallet/internal/db/block_pg.go
```diff
@@ -26,3 +29,36 @@ func pgBlockToBlock(pgBlock sqlcpg.Block) (Block, error) {
...
+		}
+
+		// Block doesn't exist, insert it.
+		blockParams := sqlcpg.InsertBlockParams{
+			BlockHeight:    height,
+			HeaderHash:     block.Hash[:],
+			BlockTimestamp: block.Timestamp.Unix(),
+		}
+
+		err := qtx.InsertBlock(ctx, blockParams)
```
yyforyongyu: This is suffering from the same race mentioned [here](https://github.com/btcsuite/btcwallet/pull/1125#discussion_r2588261942). Generally, whenever we perform check-then-act (read->write), there's always this race - two concurrent goroutines can perform read->write, then either we get a phantom read, or we get an overwrite on the write. And in this case, `InsertBlock` will throw an error if we write it twice.

I think it needs to be fixed at the query level, maybe sth like `ON CONFLICT (block_height) DO NOTHING;` for postgres then `IGNORE` for sqlite.

wallet/internal/db/wallet_pg.go
```diff
@@ -185,6 +185,25 @@ func (w *PostgresWalletDB) UpdateWallet(ctx context.Context,
 	params UpdateWalletParams) error {
 
 	return w.ExecuteTx(ctx, func(qtx *sqlcpg.Queries) error {
+		// Insert blocks if needed.
```
yyforyongyu, line 188: 👍

yyforyongyu (changes requested):
Nice work - just FYI what I meant [here](https://github.com/btcsuite/btcwallet/pull/1134#pullrequestreview-3585587523) was that @GustavoStingelin should add the block when updating the wallet, but it's nice that it gets done here.

There's the race we need to fix, and while the `BlockStore` interface is gone, we still need to have SQL queries to list blocks? This will be needed when we handle reorg - we will perform a list blocks to find the fork point, delete all the blocks, then perform a write.

Also given these changes are significantly different from the previous ACK, I'm re-requesting a review from @GustavoStingelin

wallet/internal/db/block_pg.go
GustavoStingelin: Maybe we should also consider the case where the block height already exists, but the hash is different? For example, a wallet sync might try to insert height 10 with hash A while the database already has height 10 with hash B. That would create an unreliable source of truth.

Because of that, I think we should explicitly return an error when a block is found at the same height but with a mismatching hash. This error can then be used to trigger the reorg flow, if it has not already started.

wallet/internal/db/block_pg.go
```diff
@@ -1,19 +1,64 @@
...
 
-// buildPgBlock constructs a Block from the given PostgreSQL block
-// fields.
-func buildPgBlock(height sql.NullInt32, hash []byte,
-	timestamp sql.NullInt64) (*Block, error) {
+// pgBlockToBlock converts a PostgreSQL block to a Block.
+func pgBlockToBlock(pgBlock sqlcpg.Block) (Block, error) {
+	// Header hash is guaranteed to be 32 bytes (validated on insertion).
+	var hash chainhash.Hash
+	copy(hash[:], pgBlock.HeaderHash)
```
GustavoStingelin: If an incorrect SELECT or JOIN causes this to return an invalid hash column, we would silently end up with an empty hash. Because of that, I think it is safer to use `chainhash.NewHash()`, since it makes this behavior explicit and avoids masking data issues. WDYT?

GustavoStingelin (changes requested):
I left two comments about edge cases; it also needs a rebase. 

@yyforyongyu, I would like to hear your thoughts on how we should handle block sync given the context below.

If we have multiple wallet syncers sharing the responsibility of updating the blocks table, I am concerned we could run into race conditions, especially during a reorg. An alternative could be to have a single process responsible for keeping the blocks table up to date, while separate processes handle wallet syncing.

In this model, the wallet syncers would not insert or update blocks at all. Only the block syncer would be allowed to modify the blocks table. When a reorg happens, the block syncer would notify the wallet syncers about the rescan point. Each wallet would then be responsible for validating transactions based on the already persisted block history maintained by the block syncer.

How does this approach sound to you? I am raising this now because it could be a blocker for the direction we are taking in this PR.

wallet/internal/db/block_pg.go
yyforyongyu: > Maybe we should also consider the case where the block height already exists, but the hash is different? 

Yeah that's indeed a case to consider, but not here at the db layer, as it implies reorgs, and we have dedicated process to handle reorgs - in this case, we will find the fork point, and rewrite the blocks there.

yyforyongyu:
> If we have multiple wallet syncers sharing the responsibility of updating the blocks table, I am concerned we could run into race conditions, especially during a reorg. An alternative could be to have a single process responsible for keeping the blocks table up to date, while separate processes handle wallet syncing.

The short answer is no - we either won't support multi wallets, or we will only support a minimal version due to its complexity involved. The complexity of multi wallets arises when we decide to scan - either we perform a **resync**, which scans the blockchain with wallet data; or we perform a **rescan** with targets, to search the blockchain for certain imported addresses/accounts. We cannot allow each wallet to manage its own synced state - that would create a very complex state for the system to manage, not to mention we need to care about reorgs and perform rollbacks.

The design I have so far is to make sure all wallets share the same sync height, and when a new wallet is imported/created, the existing wallets won't be operational until the new wallet is synced. In other words, in our system, we only track a single synced block, and let it shared by all the wallets. When updating the wallets' sync info, we will update all of them in a single db tx, and the same when we rollback. By supporting this minimal version, we can greatly limit our scope and focus on bringing the sqlization.

I think for this PR, addressing the existing comments will be enough - we will make some changes to the `Store` interface, and @GustavoStingelin may need to update the wallet queries given that's related to wallet/addr SQL scope. The current `Store` interface had some uncertainties before, however, as the `Controller` is about to finish, we are very confident about its final shape, e.g., we may miss three or four methods, based on this [draft](https://github.com/yyforyongyu/btcwallet/blob/controller-interface/wallet/db_ops.go).

wallet/internal/db/block_common.go
mohamedawnallah: Kept it as it is currently

wallet/internal/db/block_pg.go
mohamedawnallah: > This is suffering from the same race mentioned [here](https://github.com/btcsuite/btcwallet/pull/1125#discussion_r2588261942). Generally, whenever we perform check-then-act (read->write), there's always this race - two concurrent goroutines can perform read->write, then either we get a phantom read, or we get an overwrite on the write. And in this case, `InsertBlock` will throw an error if we write it twice.
> 
> I think it needs to be fixed at the query level, maybe sth like `ON CONFLICT (block_height) DO NOTHING;` for postgres then `IGNORE` for sqlite.

Updated

wallet/internal/db/block_pg.go
```diff
@@ -17,3 +20,27 @@ func buildPgBlock(height sql.NullInt32, hash []byte,
 
 	return buildBlock(hash, height32, timestamp.Int64)
 }
+
+// ensureBlockExistsPg ensures that a block exists in the database. If it
+// doesn't exist, it inserts it.
```
yyforyongyu: I think this is a pure overwrite now so the comments can be updated

yyforyongyu (approved):
LGTM⛵

GustavoStingelin (approved):
LGTM!

PR #1135, opened 2025-12-01
[2/5] wallet: SQLize transactions table

mohamedawnallah:
## Change Description

SQLize transactions as part of https://github.com/btcsuite/btcwallet/pull/1096.

## Steps to Test
Steps for reviewers to follow to test the change.

### Testing


saubyk:
@mohamedawnallah please update the description on this pr to something more explanatory. Thanks

mohamedawnallah:
Referenced the relevant issue/PR

mohamedawnallah:
The Itest failures are related to https://github.com/btcsuite/btcwallet/issues/1140

wallet/internal/db/interface.go
```diff
@@ -138,50 +138,73 @@ type AddressStore interface {
...
-	// inconsistent state if an error occurred mid-process. The current
-	// approach guarantees that the rollback is either fully completed or
-	// not at all.
-	//
-	// TODO(yy): explore performance improvement for this method.
-	RollbackToBlock(ctx context.Context, height uint32) error
+	// CreateTransaction inserts a new transaction record into the database.
+	// It returns the ID of the newly created transaction or an error if
+	// the creation fails.
+	CreateTransaction(ctx context.Context, tx TxInfo) (int64, error)
```
yyforyongyu: Why this change?

wallet/internal/db/interface.go
```diff
@@ -138,50 +138,73 @@ type AddressStore interface {
...
+
+	// UnconfirmedTransactionExists checks if an unconfirmed transaction
+	// with the given hash exists. It returns true if found, false
+	// otherwise.
+	UnconfirmedTransactionExists(ctx context.Context, txHash []byte) (
+		bool, error)
+
+	// DeleteTransaction deletes a specific confirmed transaction by hash
+	// and block height. It returns an error if the deletion fails.
+	DeleteTransaction(ctx context.Context, txHash []byte,
```
yyforyongyu: I also don't understand why we removed the usage of the params like `DeleteTxParams`?

mohamedawnallah: > I also don't understand why we removed the usage of the params like `DeleteTxParams`?

As currently in the codebase we can either delete using txHash or blockHeight. But `DeleteTxParams` has:
```golang
// DeleteTxParams contains the parameters for the DeleteTx method.
type DeleteTxParams struct {
	// WalletID is the ID of the wallet containing the transaction.
	//
	// NOTE: uint32 is used to ensure compatibility with standard SQL
	// databases (signed 64-bit integers).
	WalletID uint32

	// Txid is the hash of the transaction to delete.
	Txid chainhash.Hash
}
```

wallet/internal/db/interface.go
mohamedawnallah: > Why this change?

Before updating interfaces, I performed initial refactoring to ensure they're meaningful and align with the proposed interface methods for SQLization. This approach will make interfacing with the store a simple plug-in process (to adapt for KVDB and SQL). I can confirm that the current interface methods represent the minimal set we need before implementing the store interface and proceeding with SQLization in the current milestone

mohamedawnallah:
Unrequesting review till I restore a variant of [DeleteTxParams](https://github.com/btcsuite/btcwallet/pull/1135#discussion_r2622704818) and adapt the current interface methods for multi-wallet support

wallet/internal/db/interface.go
yyforyongyu: hmmm what i meant was, why change the name from `CreateTx` to `CreateTransaction`, seems unnecessary?

wallet/internal/db/interface.go
yyforyongyu: Cool then we should update the params instead - there's a clear pattern in this new design that we try to use xxxParam and xxxQuery. We also don't need these many different interface methods. By sticking as closely to a simple set of CRUDs as possible, and leveraging the params, we can pretty much support any needs from the wallet. We may still need some complex SQL queries for performance tho, but that will come later.

wallet/internal/db/interface.go
mohamedawnallah: > hmmm what i meant was, why change the name from `CreateTx` to `CreateTransaction`, seems unnecessary?

Ah yeah the name can be restored back, I thought you were referring to the whole changed diff

mohamedawnallah:
Supposed to be ready for review after https://github.com/btcsuite/btcwallet/pull/1134 is merged

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
+-- Transaction records - stores tx data for both confirmed and unconfirmed
+--
+-- Migration note: Intentionally NOT idempotent (no "IF NOT EXISTS").
+-- This ensures migration tracking stays accurate and fails loudly if run twice.
+CREATE TABLE transactions (
+    -- Surrogate key for space efficiency in foreign key references.
+    -- Auto-incrementing primary key.
+    id SERIAL PRIMARY KEY,
```
GustavoStingelin, line 8: Q: Why not bigserial?

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+
+    -- Transaction hash - unique identifier for this transaction (32 bytes).
+    -- This is the double SHA256 of the serialized transaction.
+    tx_hash BYTEA NOT NULL CHECK (length(tx_hash) = 32),
+
+    -- Block height - NULL for unconfirmed (mempool) transactions.
+    -- Foreign key to blocks table.
+    block_height INTEGER REFERENCES blocks (block_height) ON DELETE CASCADE CHECK (block_height >= 0),
```
GustavoStingelin, line 16: Should we explicitly delete the transactions before deleting the block? Also, is this check redundant given the foreign key constraint, or does it cover a case the constraint does not?

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+
+    -- Coinbase flag - true if this is a coinbase transaction (block reward).
+    is_coinbase BOOLEAN NOT NULL DEFAULT FALSE,
```
GustavoStingelin, line 19: If we are explicitly passing this value on insert, what is the benefit of having a default of false at the database layer?

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+
+    -- Timestamp when wallet became aware of the transaction (Unix timestamp).
+    -- For unconfirmed: time.Now() when first seen in mempool.
+    -- For confirmed: block header tx_hash (during sync) or time.Now() (during creation).
+    received_timestamp BIGINT NOT NULL CHECK (received_timestamp >= 0),
```
GustavoStingelin, line 24: Why not use the database type `TIMESTAMP`?

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+
+    -- Serialized transaction data (variable length).
+    -- This is the full Bitcoin transaction in wire format.
+    serialized_tx BYTEA NOT NULL CHECK (length(serialized_tx) > 0),
+
+    -- User-defined label for the transaction.
+    -- Optional descriptive text to help identify the transaction's purpose.
+    tx_label TEXT NOT NULL DEFAULT '',
```
GustavoStingelin, line 32: If we are explicitly passing this value on insert, what is the benefit of having a default empty at the database layer?

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+
+    -- In a reorg, a tx hash can appear in multiple blocks. A hash is only
+    -- unique within a given block.
+    UNIQUE (tx_hash, block_height)
```
GustavoStingelin, line 36: Should we persist the same transaction twice during a reorg? My understanding is that during a reorg we should first prune the invalid entries and then reinsert them later, instead of allowing an inconsistent state. If the intention is to allow duplication between mempool and blocks, would it make sense to enforce this with two partial indexes like:
```
ON transactions (tx_hash)
WHERE block_height IS NOT NULL;
+
ON transactions (tx_hash)
WHERE block_height IS NULL;
```

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
+-- name: InsertTransaction :one
+-- Inserts a new transaction record.
+-- For confirmed transactions, block_height must be set.
+-- For unconfirmed transactions, block_height should be NULL.
```
GustavoStingelin, line 4:
```suggestion
-- For confirmed transactions, block_height MUST be set.
-- For unconfirmed transactions, block_height MUST be NULL.
```

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+-- Returns the auto-generated ID.
+INSERT INTO transactions (tx_hash, block_height, is_coinbase, received_timestamp, serialized_tx, tx_label)
+VALUES ($1, $2, $3, $4, $5, $6)
+ON CONFLICT (tx_hash, block_height) DO NOTHING
+RETURNING id;
```
GustavoStingelin, line 9: 👍🏼

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
GustavoStingelin: But I am still thinking this should just be a single unique constraint on `transactions (tx_hash)`.

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+    t.received_timestamp,
+    t.serialized_tx,
+    t.tx_label,
+    b.header_hash,
+    b.block_timestamp
+FROM transactions AS t
+LEFT JOIN blocks AS b ON t.block_height = b.block_height
+WHERE t.tx_hash = $1
+ORDER BY t.block_height DESC NULLS FIRST
+LIMIT 1;
```
GustavoStingelin, line 27: With this, and given the previous unique index, the caller would not be able to tell whether the transaction exists both in the mempool and in a block, it would only see the mempool version.

wallet/internal/db/queries/sqlite/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+WHERE tx_hash = ? AND block_height IS NOT NULL
+ORDER BY block_height DESC
+LIMIT 1;
+
+-- name: TransactionExists :one
+-- Checks if a transaction with the given hash and block exists.
+SELECT CASE WHEN EXISTS (
+    SELECT 1 FROM transactions
+    WHERE tx_hash = ? AND block_height = ?
+) THEN 1 ELSE 0 END;
```
GustavoStingelin, line 76: Why not just like this?

```suggestion
SELECT EXISTS (
    SELECT 1 FROM transactions
    WHERE tx_hash = ? AND block_height = ?
);
```


> If executing the SELECT statement specified as the right-hand operand of the EXISTS operator would return one or more rows, then the EXISTS operator evaluates to 1. If executing the SELECT would return no rows at all, then the EXISTS operator evaluates to 0.

wallet/internal/db/queries/sqlite/transactions.sql
```diff
@@ -0,0 +1,169 @@
+-- name: InsertTransaction :one
+-- Inserts a new transaction record.
+-- For confirmed transactions, block_height must be set.
+-- For unconfirmed transactions, block_height should be NULL.
+-- Returns the auto-generated ID.
+INSERT INTO transactions (tx_hash, block_height, is_coinbase, received_timestamp, serialized_tx, tx_label)
+VALUES (?, ?, ?, ?, ?, ?)
+ON CONFLICT (tx_hash, block_height) DO NOTHING
+RETURNING id;
```
GustavoStingelin, line 9: If this works for SQLite, why are we previously using `INSERT OR IGNORE INTO` in `blocks.sql`?

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+-- name: UnconfirmTransactionsFromHeight :exec
+-- Moves all transactions at or after a given height back to unconfirmed.
+-- Used during blockchain reorganizations.
+UPDATE transactions
+SET block_height = NULL
+WHERE block_height >= $1;
+
+-- name: ListTransactionsByHeightRange :many
+-- Retrieves transactions within a block height range, with block metadata.
+-- Returns transactions ordered by block height and received timestamp.
```
GustavoStingelin, line 155: Writing the comments as full sentences, without splitting, could make them easier to understand.

```suggestion
-- name: ListTransactionsByHeightRange :many
-- Returns transactions within the given block height range, including block metadata,
-- ordered by block height and received timestamp.
```

wallet/internal/db/tx_pg.go
```diff
@@ -0,0 +1,272 @@
...
+	"time"
+
+	sqlcpg "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/postgres"
+)
+
+// Ensure PostgresWalletDB satisfies the TxStore interface.
+var _ TxStore = (*PostgresWalletDB)(nil)
+
+// CreateTx atomically records a transaction and its credits.
+func (w *PostgresWalletDB) CreateTx(ctx context.Context,
```
GustavoStingelin, line 17: We are not adding the credits yet, right? So, should we adjust the comment and add a TODO to reflect that?

wallet/internal/db/tx_pg.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+
+		txInfos = append(txInfos, *txInfo)
+	}
+
+	return txInfos, nil
+}
+
+// DeleteTx removes an unmined transaction from the store.
+func (w *PostgresWalletDB) DeleteTx(ctx context.Context,
```
GustavoStingelin, line 185: Maybe we could be more explicit in the interface that is only for unmined Tx, like renaming the method

GustavoStingelin (changes requested):
Left some comments

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
yyforyongyu: yeah let's keep it aligned with other places and use `BIGSERIAL`?

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+-- Migration note: Intentionally NOT idempotent (no "IF NOT EXISTS").
+-- This ensures migration tracking stays accurate and fails loudly if run twice.
+CREATE TABLE transactions (
+    -- Surrogate key for space efficiency in foreign key references.
+    -- Auto-incrementing primary key.
+    id SERIAL PRIMARY KEY,
+
+    -- Transaction hash - unique identifier for this transaction (32 bytes).
+    -- This is the double SHA256 of the serialized transaction.
+    tx_hash BYTEA NOT NULL CHECK (length(tx_hash) = 32),
```
yyforyongyu, line 12: nice `CHECK`!

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
yyforyongyu: yeah `TIMESTAMP` is better

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
yyforyongyu: There's a mismatch in the current design, as we already deleted the tx on reorg, which means we can never have `txid1+block1` and `txid1+block2` existed at the same time.

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
yyforyongyu: Was thinking it's nice to have default, sort like documenting that most txns are non-coinbase. But I think being explicit is safer - i think we should enforce every insert to be strict on the values so prefer not having the defaults.

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
yyforyongyu: The `ON DELETE CASCADE` has quite some implications during reorgs - this approach is nice as we can now wipe the data and re-insert the txns found in the new block, which is simple. However, the critical info such as `received_timestamp` will be reset.

Given that the `transactions` table should record the history of the wallet txns, we should instead do `ON DELETE SET NULL`, this means when we handle reorg, we will,
- update the tx's block to be the new block if it's found there, or,
- update the tx's block to be null if not, which means the tx is now unconfirmed.

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
yyforyongyu: I think either way is fine, but we should add a limit on the length of the label to prevent inserting massive strings.

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+    -- Block height - NULL for unconfirmed (mempool) transactions.
+    -- Foreign key to blocks table.
+    block_height INTEGER REFERENCES blocks (block_height) ON DELETE CASCADE CHECK (block_height >= 0),
+
+    -- Coinbase flag - true if this is a coinbase transaction (block reward).
+    is_coinbase BOOLEAN NOT NULL DEFAULT FALSE,
+
+    -- Timestamp when wallet became aware of the transaction (Unix timestamp).
+    -- For unconfirmed: time.Now() when first seen in mempool.
+    -- For confirmed: block header tx_hash (during sync) or time.Now() (during creation).
```
yyforyongyu, line 23: Q: why do we need to set this value to be the blocks' ts? i think that part can be retrieved from the block table?

Actually I just realized this is not [addressed](https://github.com/btcsuite/btcwallet/pull/1096/changes#r2412348935).

wallet/internal/db/migrations/postgres/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+    tx_label TEXT NOT NULL DEFAULT '',
+
+    -- In a reorg, a tx hash can appear in multiple blocks. A hash is only
+    -- unique within a given block.
+    UNIQUE (tx_hash, block_height)
+);
+
+-- Unconfirmed transactions must be unique. A partial index on tx_hash where
+-- block_height is NULL enforces this.
+CREATE UNIQUE INDEX transactions_unconfirmed_hash_idx ON transactions (tx_hash)
```
yyforyongyu, line 41: I think the `tx_hash` should be unique on its own. The extremely unlikely case is when outputs from BIP30 got stored here, in that case we will fix it at the app layer, similar to [this check](https://github.com/bitcoin/bitcoin/commit/ab91bf39b7c11e9c86bb2043c24f0f377f1cf514).

wallet/internal/db/migrations/sqlite/000005_transactions.up.sql
```diff
@@ -0,0 +1,42 @@
...
+    -- Auto-incrementing primary key.
+    id INTEGER PRIMARY KEY AUTOINCREMENT,
+
+    -- Transaction hash - unique identifier for this transaction (32 bytes).
+    -- This is the double SHA256 of the serialized transaction.
+    tx_hash BLOB NOT NULL CHECK (length(tx_hash) = 32),
+
+    -- Block height - NULL for unconfirmed (mempool) transactions.
+    -- Foreign key to blocks table.
+    block_height INTEGER REFERENCES blocks (block_height) ON DELETE CASCADE CHECK (block_height >= 0),
```
yyforyongyu, line 16: same here re `ON DELETE`, the `received_timestamp`, `tx_label` and `UNIQUE`

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
+-- name: InsertTransaction :one
+-- Inserts a new transaction record.
+-- For confirmed transactions, block_height must be set.
+-- For unconfirmed transactions, block_height should be NULL.
+-- Returns the auto-generated ID.
+INSERT INTO transactions (tx_hash, block_height, is_coinbase, received_timestamp, serialized_tx, tx_label)
+VALUES ($1, $2, $3, $4, $5, $6)
+ON CONFLICT (tx_hash, block_height) DO NOTHING
```
yyforyongyu, line 8: Q: why perform `DO NOTHING`?

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+RETURNING id;
+
+-- name: GetTransactionByHash :one
+-- Gets a transaction by hash with priority: unconfirmed first, then highest confirmed block.
```
yyforyongyu, line 12: This complexity will be dropped if we use `UNIQUE (tx_hash)`

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+    t.tx_label,
+    b.header_hash,
+    b.block_timestamp
+FROM transactions AS t
+LEFT JOIN blocks AS b ON t.block_height = b.block_height
+WHERE t.tx_hash = $1
+ORDER BY t.block_height DESC NULLS FIRST
+LIMIT 1;
+
+-- name: GetTransactionByHashAndBlock :one
```
yyforyongyu, line 29: Q: when will this be used?

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+    is_coinbase,
+    received_timestamp,
+    serialized_tx,
+    tx_label
+FROM transactions
+WHERE tx_hash = $1 AND block_height IS NOT NULL
+ORDER BY block_height DESC
+LIMIT 1;
+
+-- name: TransactionExists :one
```
yyforyongyu, line 71: Redundant?

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+    block_height,
+    is_coinbase,
+    received_timestamp,
+    serialized_tx,
+    tx_label
+FROM transactions
+WHERE block_height IS NULL
+ORDER BY received_timestamp ASC;
+
+-- name: GetAllUnconfirmedTransactionHashes :many
```
yyforyongyu, line 112: also why do we need this method?

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+SET block_height = $2
+WHERE tx_hash = $1 AND block_height IS NULL;
+
+-- name: UpdateTransactionLabel :exec
+-- Updates the label of a transaction.
+UPDATE transactions
+SET tx_label = $2
+WHERE tx_hash = $1;
+
+-- name: UnconfirmTransactionsFromHeight :exec
```
yyforyongyu, line 146: The name should be `UpdateXXX`

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+-- Moves all transactions at or after a given height back to unconfirmed.
+-- Used during blockchain reorganizations.
+UPDATE transactions
+SET block_height = NULL
+WHERE block_height >= $1;
+
+-- name: ListTransactionsByHeightRange :many
```
yyforyongyu, line 153: I see lots of methods using `ByHeight`, but the `block_height` is not indexed.

wallet/internal/db/queries/postgres/transactions.sql
```diff
@@ -0,0 +1,169 @@
...
+FROM transactions
+WHERE block_height IS NULL
+ORDER BY received_timestamp ASC;
+
+-- name: GetAllUnconfirmedTransactionHashes :many
+-- Retrieves all unconfirmed transaction hashes only.
+SELECT tx_hash
+FROM transactions
+WHERE block_height IS NULL
+ORDER BY received_timestamp ASC;
```
yyforyongyu, line 117: We may need a partial index on `received_timestamp` otherwise this can be slow

wallet/internal/db/tx_pg.go
yyforyongyu: Q: why deleting is limited to unmined txns?

wallet/internal/db/tx_pg.go
```diff
@@ -0,0 +1,272 @@
...
+	if err != nil {
+		return fmt.Errorf("delete unmined transaction: %w", err)
+	}
+
+	return nil
+}
+
+// RollbackToBlock removes all blocks at and after a given height, moving
+// transactions back to the unconfirmed pool.
+func (w *PostgresWalletDB) RollbackToBlock(ctx context.Context,
```
yyforyongyu, line 198: We need to rethink about this method, maybe we could drop this method for the current PR - we also need to have the UTXOs table ready first before we can properly implement this method.

yyforyongyu (changes requested):

PR #1136, opened 2025-12-09
wallet: add sqldb itest

GustavoStingelin:
This PR split the itest code that was introduced in #1125. The goal is to add the SQL DB itest code first so that other contributors can start using it.


Makefile
```diff
@@ -84,6 +84,24 @@ unit-bench:
 	@$(call print, "Running benchmark tests.")
 	$(UNIT_BENCH)
 
+#? itest-db: Run integration tests for wallet database
+itest-db:
```
yyforyongyu, line 88: A minor point - can we add some default behavior like what we did in `lnd`? so we can default to sqlite.

.github/workflows/main.yml
```diff
@@ -152,3 +152,40 @@ jobs:
...
+  # run integration tests (SQLite + Postgres)
+  ########################
+  itest-db:
+    name: integration tests (${{ matrix.db }}, ${{ matrix.target }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        db: [sqlite, postgres]
+        target: [itest-db, itest-db-race]
```
yyforyongyu, line 166: Q: I think `itest-db-race` will cover `itest-db`? so we don't need to run both?

Makefile
```diff
@@ -178,6 +196,7 @@ sqlc-check: sqlc
 	unit-race \
 	unit-debug \
 	unit-bench \
+	itest \
```
yyforyongyu: itest-db and itest-db-race?

GustavoStingelin: fixed

Makefile
GustavoStingelin: added sqlite as default

.github/workflows/main.yml
GustavoStingelin: I think keeping both is useful because:

- Race mode is much slower, so it delays feedback when a regular integration test fails
- Race mode needs CGO enabled, and that setup can hide problems that only appear in pure Go builds

yyforyongyu (approved):
LGTM 🚢

mohamedawnallah (approved):

PR #1137, opened 2025-12-11
Add helper methods for `WalletController`

yyforyongyu:
In this PR,
- added `DeriveAddr` and `DeriveAddrs` for in-memory derivation
- exposed exsiting `GetCFilter` on `chain.Interface`, which will be used for fast querying
- copied `port.go` from `lnd` into `chain/port` to fix port collision in unit tests

### DIFF
--- a/chain/bitcoind_events_test.go
+++ b/chain/bitcoind_events_test.go
@@ -2,7 +2,6 @@ package chain
 
 import (
 	"fmt"
-	"math/rand"
 	"os/exec"
 	"testing"
 	"time"
@@ -15,6 +14,7 @@ import (
 	"github.com/btcsuite/btcd/rpcclient"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/chain/port"
 	"github.com/stretchr/testify/require"
 )
 
@@ -460,10 +460,13 @@ func setupBitcoind(t *testing.T, minerAddr string,
 	// Start a bitcoind instance and connect it to miner1.
 	tempBitcoindDir := t.TempDir()
 
-	zmqBlockHost := "ipc:///" + tempBitcoindDir + "/blocks.socket"
-	zmqTxHost := "ipc:///" + tempBitcoindDir + "/tx.socket"
+	zmqBlockPort := port.NextAvailablePort()
+	zmqTxPort := port.NextAvailablePort()
 
-	rpcPort := rand.Int()%(65536-1024) + 1024
+	zmqBlockHost := fmt.Sprintf("tcp://127.0.0.1:%d", zmqBlockPort)
+	zmqTxHost := fmt.Sprintf("tcp://127.0.0.1:%d", zmqTxPort)
+
+	rpcPort := port.NextAvailablePort()
 	bitcoind := exec.Command(
 		"bitcoind",
 		"-datadir="+tempBitcoindDir,
--- a/chain/bitcoind_events_test.go
+++ b/chain/bitcoind_events_test.go
@@ -8,6 +8,8 @@ import (
 
 	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/btcutil/gcs"
+	"github.com/btcsuite/btcd/btcutil/gcs/builder"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/integration/rpctest"
@@ -68,7 +70,12 @@ func TestBitcoindEvents(t *testing.T) {
 			testNotifySpentMempool(t, miner1, btcClient)
 
 			// Test looking up mempool for input spent.
+			btcClient = setupBitcoind(t, addr, test.rpcPolling)
 			testLookupInputMempoolSpend(t, miner1, btcClient)
+
+			// Test GetCFilter method.
+			btcClient = setupBitcoind(t, addr, test.rpcPolling)
+			testBitcoindClientGetCFilter(t, miner1, btcClient)
 		})
 	}
 }
@@ -480,6 +487,7 @@ func setupBitcoind(t *testing.T, minerAddr string,
 		"-disablewallet",
 		"-zmqpubrawblock="+zmqBlockHost,
 		"-zmqpubrawtx="+zmqTxHost,
+		"-blockfilterindex=1",
 	)
 	require.NoError(t, bitcoind.Start())
 
@@ -558,4 +566,59 @@ func randPubKeyHashScript() ([]byte, *btcec.PrivateKey, error) {
 	}
 
 	return pkScript, privKey, nil
+}
+
+// testBitcoindClientGetCFilter verifies the BitcoindClient's GetCFilter
+// implementation by interacting with a live bitcoind node.
+func testBitcoindClientGetCFilter(t *testing.T, miner *rpctest.Harness,
+	client *BitcoindClient) {
+
+	t.Helper()
+
+	require := require.New(t)
+
+	// Generate a block to have something to query a filter for.
+	hashes, err := miner.Client.Generate(1)
+	require.NoError(err)
+
+	blockHash := hashes[0]
+
+	// Get the CFilter using the BitcoindClient. This might take a few
+	// attempts as the filter index might not be immediately available.
+	var gcsFilter *gcs.Filter
+	require.Eventually(func() bool {
+		gcsFilter, err = client.GetCFilter(
+			blockHash, wire.GCSFilterRegular,
+		)
+
+		return err == nil
+	}, 5*time.Second, 100*time.Millisecond, "GetCFilter should succeed")
+	require.NotNil(gcsFilter, "GCS filter should not be nil")
+	require.IsType(&gcs.Filter{}, gcsFilter)
+
+	// Verify the filter matches the block data.
+	block, err := client.GetBlock(blockHash)
+	require.NoError(err)
+
+	// Use the first transaction's first output script.
+	script := block.Transactions[0].TxOut[0].PkScript
+
+	// Derive the filter key.
+	key := builder.DeriveKey(blockHash)
+
+	// Check match.
+	matched, err := gcsFilter.Match(key, script)
+	require.NoError(err)
+	require.True(matched, "Filter should match script from block")
+
+	// Test with an unsupported filter type.
+	_, err = client.GetCFilter(blockHash, wire.FilterType(99))
+	require.ErrorContains(err, "only basic filters are supported",
+		"Unsupported filter type should return an error")
+
+	// Test GetCFilter for a non-existent block.
+	dummyHash := &chainhash.Hash{0x01, 0x02, 0x03}
+	_, err = client.GetCFilter(dummyHash, wire.GCSFilterRegular)
+	require.ErrorContains(err, "Block not found",
+		"Non-existent block should return an error")
 }
--- a/chain/bitcoind_events_test.go
+++ b/chain/bitcoind_events_test.go
@@ -20,62 +20,83 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-// TestBitcoindEvents ensures that the BitcoindClient correctly delivers tx and
-// block notifications for both the case where a ZMQ subscription is used and
-// for the case where RPC polling is used.
-func TestBitcoindEvents(t *testing.T) {
+const (
+	// defaultTestTimeout is the default timeout used for tests in this
+	// file. It is set to 30 seconds to allow for slow test environments.
+	defaultTestTimeout = 30 * time.Second
+)
+
+// TestBitcoindEventsZMQ runs all bitcoind event tests using ZMQ subscriptions.
+//
+// We cannot run these tests in parallel as it involves running multiple
+// bitcoind servers and btcd servers in the background. While running multiple
+// bitcoind servers is fine, the current integration test setup in `btcd`
+// doesn't allow it as the created RPC client will share the same ports.
+//
+//nolint:paralleltest
+func TestBitcoindEventsZMQ(t *testing.T) {
+	runBitcoindEventsTests(t, false)
+}
+
+// TestBitcoindEventsRPC runs all bitcoind event tests using RPC polling.
+//
+// We cannot run these tests in parallel as it involves running multiple
+// bitcoind servers and btcd servers in the background. While running multiple
+// bitcoind servers is fine, the current integration test setup in `btcd`
+// doesn't allow it as the created RPC client will share the same ports.
+//
+//nolint:paralleltest
+func TestBitcoindEventsRPC(t *testing.T) {
+	runBitcoindEventsTests(t, true)
+}
+
+// runBitcoindEventsTests runs the suite of bitcoind event tests with the
+// specified polling mode.
+func runBitcoindEventsTests(t *testing.T, rpcPolling bool) {
+	t.Helper()
+
 	tests := []struct {
-		name       string
-		rpcPolling bool
+		name   string
+		testFn func(*testing.T, *rpctest.Harness, *BitcoindClient)
 	}{
 		{
-			name:       "Events via ZMQ subscriptions",
-			rpcPolling: false,
+			name:   "Reorg",
+			testFn: testReorg,
 		},
 		{
-			name:       "Events via RPC Polling",
-			rpcPolling: true,
+			name:   "NotifyBlocks",
+			testFn: testNotifyBlocks,
+		},
+		{
+			name:   "NotifyTx",
+			testFn: testNotifyTx,
+		},
+		{
+			name:   "NotifySpentMempool",
+			testFn: testNotifySpentMempool,
+		},
+		{
+			name:   "LookupInputMempoolSpend",
+			testFn: testLookupInputMempoolSpend,
+		},
+		{
+			name:   "GetCFilter",
+			testFn: testBitcoindClientGetCFilter,
 		},
 	}
 
 	for _, test := range tests {
 		test := test
-
-		// Set up 2 btcd miners.
-		miner1, miner2 := setupMiners(t)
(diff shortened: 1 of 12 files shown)


chain/bitcoind_events_test.go
```diff
@@ -2,22 +2,30 @@ package chain
...
 	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/chain/port"
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	// defaultTestTimeout is the default timeout used for tests in this
+	// file. It is set to 10 seconds to allow for slow test environments.
+	defaultTestTimeout = 10 * time.Second
+)
```
mohamedawnallah, line 27: This doesn't fix the flaky tests for `TestPrunedBlockDispatcherMultipleQueryPeers` neither `TestBitcoindEvents`. It can be reproduced by something along those lines (-count field):
```bash
$ go test -v -run TestPrunedBlockDispatcherMultipleQueryPeers -count=500 ./chain
....
=== CONT  TestPrunedBlockDispatcherMultipleQueryPeers
    pruned_block_dispatcher_test.go:635: expected nil err to signal completion
    pruned_block_dispatcher_test.go:186: did not consume all queriedPeer signals
--- FAIL: TestPrunedBlockDispatcherMultipleQueryPeers (10.01s)
....
```

```bash
$ go test -v -run TestBitcoindEvents -count=500 ./chain
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
--- PASS: TestBitcoindEvents (17.20s)
    --- PASS: TestBitcoindEvents/Events_via_ZMQ_subscriptions (7.14s)
    --- PASS: TestBitcoindEvents/Events_via_RPC_Polling (7.46s)
=== RUN   TestBitcoindEvents
    bitcoind_events_test.go:449:
        	Error Trace:	/home/dev/btcwallet/chain/bitcoind_events_test.go:449
        	            				/home/dev/btcwallet/chain/bitcoind_events_test.go:51
        	Error:      	Received unexpected error:
        	            	error connecting RPC client: connection timeout, tried 20 times with timeout 50ms, last err: dial tcp 127.0.0.1:8348: connect: connection refused
        	Test:       	TestBitcoindEvents
--- FAIL: TestBitcoindEvents (9.68s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
--- PASS: TestBitcoindEvents (16.83s)
    --- PASS: TestBitcoindEvents/Events_via_ZMQ_subscriptions (7.13s)
    --- PASS: TestBitcoindEvents/Events_via_RPC_Polling (7.44s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
    bitcoind_events_test.go:319: received a notification of type chain.FilteredBlockConnected but expected, none
--- FAIL: TestBitcoindEvents (10.51s)
    --- PASS: TestBitcoindEvents/Events_via_ZMQ_subscriptions (7.12s)
    --- FAIL: TestBitcoindEvents/Events_via_RPC_Polling (1.14s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
--- PASS: TestBitcoindEvents (16.86s)
    --- PASS: TestBitcoindEvents/Events_via_ZMQ_subscriptions (7.15s)
    --- PASS: TestBitcoindEvents/Events_via_RPC_Polling (7.40s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
--- PASS: TestBitcoindEvents (16.77s)
    --- PASS: TestBitcoindEvents/Events_via_ZMQ_subscriptions (7.13s)
    --- PASS: TestBitcoindEvents/Events_via_RPC_Polling (7.41s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
--- PASS: TestBitcoindEvents (16.81s)
    --- PASS: TestBitcoindEvents/Events_via_ZMQ_subscriptions (7.13s)
    --- PASS: TestBitcoindEvents/Events_via_RPC_Polling (7.43s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
--- PASS: TestBitcoindEvents (16.77s)
    --- PASS: TestBitcoindEvents/Events_via_ZMQ_subscriptions (7.12s)
    --- PASS: TestBitcoindEvents/Events_via_RPC_Polling (7.39s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
--- PASS: TestBitcoindEvents (16.81s)
    --- PASS: TestBitcoindEvents/Events_via_ZMQ_subscriptions (7.14s)
    --- PASS: TestBitcoindEvents/Events_via_RPC_Polling (7.42s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
    bitcoind_events_test.go:84:
        	Error Trace:	/home/dev/btcwallet/chain/bitcoind_events_test.go:600
        	            				/home/dev/btcwallet/chain/bitcoind_events_test.go:84
        	Error:      	Condition never satisfied
        	Test:       	TestBitcoindEvents/Events_via_ZMQ_subscriptions
        	Messages:   	GetCFilter should succeed
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
--- FAIL: TestBitcoindEvents (26.73s)
    --- FAIL: TestBitcoindEvents/Events_via_ZMQ_subscriptions (17.06s)
    --- PASS: TestBitcoindEvents/Events_via_RPC_Polling (7.40s)
=== RUN   TestBitcoindEvents
=== RUN   TestBitcoindEvents/Events_via_ZMQ_subscriptions
=== RUN   TestBitcoindEvents/Events_via_RPC_Polling
```

mohamedawnallah (changes requested):

chain/bitcoind_events_test.go
yyforyongyu: 500 seems too much - I suspect there is some cleanup process for `bitcoind` missing; however, I used `count=10`, it was failing before, and now it's working under `count=10`, so I think it's an improvement. Btw did you observe anything from the `count=500`, like any useful logs to further debug this issue?

mohamedawnallah: >  Btw did you observe anything from the count=500, like any useful logs to further debug this issue?

Haven't put more investigation on why it was failing but you can try the command locally. I think it is reproducible

yyforyongyu: Ran it 1000 times and can't reproduce,
```
> └─[0] <git:(prep-controller e1b17e2f) > go test -v -run TestPrunedBlockDispatcherMultipleQueryPeers -count=1000 ./chain
...
--- PASS: TestPrunedBlockDispatcherMultipleQueryPeers (0.00s)
PASS
ok  	github.com/btcsuite/btcwallet/chain	6.518s
```

Also rewrote the `TestBitcoindEvents` to make it more verbose, can you run it again and show me the errors? I think it's related to your local env.

mohamedawnallah: Passing on first run for count=500:
https://gist.github.com/mohamedawnallah/825f2372e82832f9ff8fed1a5499df93#file-testprunedblockdispatchermultiplequerypeers_1_passing-log

Failing on second run for count=500:
https://gist.github.com/mohamedawnallah/825f2372e82832f9ff8fed1a5499df93#file-testprunedblockdispatchermultiplequerypeers_2_failing-log

`TestPrunedBlockDispatcherMultipleQueryPeers` seems not deterministic. Not sure though if this is env issue related

mohamedawnallah (approved):

chain/bitcoind_events_test.go
yyforyongyu: What about `TestBitcoindEvents`?

mohamedawnallah: It was same. Resolving this for now. Can be iterated later if there are issues found in a typical CI environment

waddrmgr/scoped_manager.go
```diff
@@ -2014,6 +2014,74 @@ func (s *ScopedKeyManager) ImportPublicKey(ns walletdb.ReadWriteBucket,
...
+	addrs := make([]btcutil.Address, 0, count)
+	scripts := make([][]byte, 0, count)
+
+	// Iterate through the requested range of child indexes (startIndex to
+	// startIndex+count). For each index, we derive the corresponding
+	// extended key and convert it into a payment address and script.
+	//
+	// TODO(yy): Optimize by deriving the branch key once outside the loop
+	// instead of re-deriving it for every index via s.deriveKey.
+	endIndex := startIndex + count
```
GustavoStingelin, line 2077: Should we add a guard to prevent `uint32` overflow? For example, if the current index is close to the maximum value, say max minus 10, and we try to derive 20 entries, we would end up with an invalid derivation. Adding an explicit check could help fail early and avoid subtle bugs.

waddrmgr/scoped_manager.go
```diff
@@ -2656,6 +2724,64 @@ func (s *ScopedKeyManager) NewAddress(addrmgrNs walletdb.ReadWriteBucket,
...
+// the provided account info. It assumes the manager lock is held.
+func (s *ScopedKeyManager) deriveAddr(acctInfo *accountInfo, account, branch,
+	index uint32) (btcutil.Address, []byte, error) {
+
+	// Determine the address type (schema) for this account and branch.
+	// This tells us whether to generate P2PKH (BIP44), P2WPKH (BIP84),
+	// Nested P2WPKH (BIP49), or Taproot (BIP86) addresses.
+	// Internal branch usually implies change addresses.
+	internal := branch == InternalBranch
+	addrType := s.accountAddrType(acctInfo, internal)
```
GustavoStingelin, line 2745: 👍🏼

waddrmgr/scoped_manager.go
```diff
@@ -2656,6 +2724,64 @@ func (s *ScopedKeyManager) NewAddress(addrmgrNs walletdb.ReadWriteBucket,
...
+	key, err := s.deriveKey(acctInfo, branch, index, false)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	pubKey, err := key.ECPubKey()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to derive public key: %w",
+			err)
+	}
```
GustavoStingelin, line 2760: Maybe the first error should be reserved for derivation failures, while the second one should cover errors when extracting or parsing the public key from the extended key, since it will only contains public key material.

waddrmgr/scoped_manager_test.go
```diff
@@ -0,0 +1,280 @@
...
+			branch:     ExternalBranch,
+			startIndex: 1000,
+			count:      1,
+		},
+		{
+			name:       "Zero Addresses",
+			branch:     ExternalBranch,
+			startIndex: 0,
+			count:      0,
+		},
```
GustavoStingelin, line 96: could also add a uint32 overflow case

chain/port/port.go
```diff
@@ -0,0 +1,149 @@
...
+
+// NextAvailablePort returns the first port that is available for listening by a
+// new node, using a lock file to make sure concurrent access for parallel tasks
+// on the same system don't re-use the same port.
+func NextAvailablePort() int {
+	portFileMutex.Lock()
+	defer portFileMutex.Unlock()
+
+	lockFile := filepath.Join(os.TempDir(), uniquePortFile+".lock")
+	timeout := time.After(defaultTimeout)
```
GustavoStingelin: IIUC, this will be used only for tests, right? So, how can we prevent being used in production? Maybe adding some reference in the name, or in the package, or a comment note?

chain/bitcoind_events_test.go
```diff
@@ -559,3 +567,58 @@ func randPubKeyHashScript() ([]byte, *btcec.PrivateKey, error) {
...
 }
+
+// testBitcoindClientGetCFilter verifies the BitcoindClient's GetCFilter
+// implementation by interacting with a live bitcoind node.
+func testBitcoindClientGetCFilter(t *testing.T, miner *rpctest.Harness,
+	client *BitcoindClient) {
+
+	t.Helper()
+
+	require := require.New(t)
```
GustavoStingelin, line 603: Q: Since this is effectively the test itself, just wrapped to be included in the harness, what is the benefit of marking it with `t.Helper()`?

chain/port/port.go
```diff
@@ -139,11 +135,38 @@ func NextAvailablePort() int {
...
+
+	var (
+		lockFileHandle *os.File
+		err            error
+	)
+	for {
+		// Attempt to acquire the lock file. If it already exists,
+		// wait for a bit and retry.
+		lockFileHandle, err = os.OpenFile( //nolint:gosec
```
GustavoStingelin: could add a comment about the //nolint:gosec

chain/bitcoind_events_test.go
```diff
@@ -20,62 +20,84 @@ import (
...
 		},
 	}
 
 	for _, test := range tests {
 		test := test
+		t.Run(test.name, func(t *testing.T) {
+			// Initialize a fresh set of miners for EVERY test
+			// case.
+			miner1 := setupMiner(t)
+			addr := miner1.P2PAddress()
```
GustavoStingelin: The comment says a set of miners, but we're creating just one

chain/bitcoind_events_test.go
GustavoStingelin: same for others that are being used in "real" tests

chain/port/port.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+
+	lastPort, err := strconv.Atoi(string(port))
+	if err != nil {
+		panic(fmt.Errorf("error parsing port: %w", err))
+	}
+
+	// We take the next one.
+	lastPort++
+	for lastPort < maxPort {
```
GustavoStingelin: Looks like when the _`port file`_ reaches `65534`, the next call increments to `65535`, skips the loop (lastPort < maxPort), and panics instead of wrapping to `defaultNodePort`. Also, would need a backport for LND and unit tests, but not a serious problem asit  would take a long time to happen and only in developer environments.

chain/port/port.go
```diff
@@ -0,0 +1,172 @@
...
+		)
+		if err == nil {
+			// Lock acquired.
+			return lockFileHandle
+		}
+
+		// Wait for a bit and retry.
+		select {
+		case <-timeout:
+			panic("timeout waiting for lock file")
```
GustavoStingelin: If the process is killed before the deferred cleanup runs, the `.lock` file will remain. In that case, `acquireLockFile` will keep spinning until it hits the timeout and then panic. I do not think this is a major issue, but it is a possible failure. We could document this behavior and also make the panic message more explicit to help with debugging.

GustavoStingelin (changes requested):
Looking good, just left some comments.

chain/port/port.go
yyforyongyu: Correct - this is mostly just a copy from `lnd`'s `port.go` - i think we could update the `lnd` to use this one instead, tho not urgent.

yyforyongyu: also fixed the max port case so we can readh `65535`

chain/port/port.go
yyforyongyu: updated

chain/bitcoind_events_test.go
yyforyongyu: fixed

waddrmgr/scoped_manager.go
yyforyongyu: nice catch - added!

waddrmgr/scoped_manager_test.go
yyforyongyu: added!

chain/port/port.go
yyforyongyu: yeah added the filename and docs

chain/bitcoind_events_test.go
yyforyongyu: yeah the linter complains without this so added

waddrmgr/scoped_manager.go
yyforyongyu: Not sure I follow, can you elaborate? what is the first or second error?

chain/port/port.go
yyforyongyu: yeah only for tests - I think I will move this to a different package, like `btctest` (similar to `lntest`) since the itest needs to use this too.

waddrmgr/scoped_manager.go
GustavoStingelin: I meant that `deriveKey` can fail during the derivation step itself, and in that case returning a derivation related error makes sense. However, once `deriveKey` succeeds and returns an extended key with `isPrivate = false`, the key already contains only public key material. Calling `key.ECPubKey()` does not perform any derivation, it only extracts or parses the public key from the extended key.

Because of that, an error coming from `key.ECPubKey()` is not a derivation failure. It is a failure to parse or extract the public key from the extended key. Using an error message like “failed to derive public key” in this second case feels ambiguous, since no derivation is happening at that point. A more specific error message would make it clearer where the failure actually occurred.

GustavoStingelin (approved):
ACK

waddrmgr/scoped_manager.go
yyforyongyu: oh i see - yea make sense, updated it to be 
```go
		return nil, nil, fmt.Errorf("failed to parse public key: %w",
			err)
```

PR #1138, opened 2025-12-15
wallet: Implement address type store support

### LINKED ISSUE #1104: [sqldb/address]: Implement address type support
Create migrations for address_types (ref #1051)

GustavoStingelin:
This change adds SQL support for address types, which is required to introduce follow up tables such as key scopes and addresses.

Closes #1104.


GustavoStingelin:
Guys, I tried to generalize this a bit. I aimed to balance DRY and KISS. Please take a look and let me know if this feels right to you.

wallet/internal/db/migrations/postgres/000003_address_types.up.sql
```diff
@@ -0,0 +1,32 @@
...
+-- Seed reference data matching the Go AddressType enum constants.
+-- These values are static and represent the Bitcoin address types.
+-- IDs MUST match the iota values in wallet/internal/db/data_types.go.
+INSERT INTO address_types (id, description) VALUES
+(0, 'P2PKH'),          -- Pay-to-PubKey-Hash
+(1, 'P2SH'),           -- Pay-to-Script-Hash
+(2, 'P2WPKH'),         -- Pay-to-Witness-PubKey-Hash
+(3, 'P2WSH'),          -- Pay-to-Witness-Script-Hash
+(4, 'P2SH-P2WPKH'),    -- Nested Witness PubKey
+(5, 'P2TR');           -- Pay-to-Taproot
```
mohamedawnallah, line 32: 👍

mohamedawnallah (approved):

mohamedawnallah (changes requested):
~~Should be itested against SQLite AND PostgreSQL~~ It is indeed itested

mohamedawnallah (approved):
re-ACK

GustavoStingelin:
just rebased

wallet/internal/db/migrations/postgres/000003_address_types.up.sql
yyforyongyu, line 32: Nice list - just a bit forward thinking, we should also add [P2A](https://github.com/bitcoin/bips/pull/1982) and [`SilentPayment`](https://github.com/bitcoin/bips/blob/master/bip-0352.mediawiki) addr types, also we should add `P2PK`.

None of these types will be supported yet, but having in here first saves us from running a migration in the future.

yyforyongyu (approved):
LGTM ☘️ Again feel free to address the comments in a future PR.

wallet/internal/db/migrations/postgres/000003_address_types.up.sql
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1143

Issue #1140, opened 2025-12-16
[bug]: PostgreSQL integration tests fail with (FATAL: sorry, too many clients already (SQLSTATE 53300))


mohamedawnallah:
### Background

```bash
$ make itest-db db=postgres
 Running PostgreSQL integration tests.
GO111MODULE=on go test -tags="itest dev debug stdlog test_db_postgres" -test.timeout=20m github.com/btcsuite/btcwallet/wallet/internal/db/itest
--- FAIL: TestCreateWallet_Variants (0.00s)
    --- FAIL: TestCreateWallet_Variants/standard_wallet (0.25s)
        wallet_store_test.go:107:
            	Error Trace:	/home/dev/btcwallet/wallet/internal/db/itest/wallet_store_test.go:107
            	Error:      	Received unexpected error:
            	            	begin tx: failed to connect to `host=localhost user=postgres database=testcreatewallet_variants_standard_wallet`: server error (FATAL: sorry, too many clients already (SQLSTATE 53300))
            	Test:       	TestCreateWallet_Variants/standard_wallet
    --- FAIL: TestCreateWallet_Variants/watch-only_wallet (0.25s)
        wallet_store_test.go:107:
            	Error Trace:	/home/dev/btcwallet/wallet/internal/db/itest/wallet_store_test.go:107
            	Error:      	Received unexpected error:
            	            	begin tx: failed to connect to `host=localhost user=postgres database=testcreatewallet_variants_watch_only_wallet`: server error (FATAL: sorry, too many clients already (SQLSTATE 53300))
            	Test:       	TestCreateWallet_Variants/watch-only_wallet
FAIL
FAIL	github.com/btcsuite/btcwallet/wallet/internal/db/itest	7.779s
FAIL
make: *** [Makefile:109: itest-db] Error 1
```

mohamedawnallah:
PTAL @GustavoStingelin

GustavoStingelin:
In which branch is it happening? On CI too or only local?

mohamedawnallah:
> In which branch is it happening? On CI too or only local?

Both, CI: https://github.com/btcsuite/btcwallet/actions/runs/20255289509/job/58155797182?pr=1135

GustavoStingelin:
Fixed in #1141.

[...]

PR #1141, opened 2025-12-16
wallet: fix pg itest "too many clients"

GustavoStingelin:
Solves #1140

### Issue
pg itest hit `FATAL: sorry, too many clients already` because the shared container was reused across tests while admin/test connections stayed open, leaking idle clients.

### Fix
add `t.Cleanup` to close both the admin connection (used to create per-test DBs) and each test DB handle, preventing connection leaks in the shared container.
  
### Test
`make itest-db db=postgres` and `make itest-db-race db=postgres`


GustavoStingelin:
@mohamedawnallah I tested this fix in your PR #1135 and all tests passed on my side. Could you please test it as well?

mohamedawnallah (approved):
tACK, Thanks!

yyforyongyu (approved):
🔥

Issue #1106, opened 2025-10-20
[sqldb/address]: Implement account support


GustavoStingelin:
Create migrations for accounts, account_origins, and account_secrets (ref #1051)

Provide SQL queries for inserting and reading these tables

Run SQLC to generate Go models and query wrappers

Implement the methods defined in the interfaces

PR #1142, opened 2025-12-17
wallet: add key scope SQL migrations and queries

### LINKED ISSUE #1105: [sqldb/address]: Implement key scope support
Create migrations for key_scopes and key_scope_secrets (ref #1051)

GustavoStingelin:
This PR adds the SQL migrations and database level queries for the key scope tables. It does not implement any interfaces, since those will live behind `AccountStore`. The key scope does not expose an external interface.

This PR depends on #1138, closes #1105, and the next work will be #1106.


GustavoStingelin:
just rebased

wallet/internal/db/migrations/sqlite/000004_key_scopes.up.sql
```diff
@@ -0,0 +1,69 @@
...
+    -- Foreign key constraints to address types. Using ON DELETE RESTRICT to ensure
+    -- that the address types cannot be deleted if key scopes still exist.
+    FOREIGN KEY (internal_type_id) REFERENCES address_types (id) ON DELETE RESTRICT,
+    FOREIGN KEY (external_type_id) REFERENCES address_types (id) ON DELETE RESTRICT
+);
+
+-- Index on foreign wallet_id for faster lookups and joins. Probably not needed
+-- because with just one or a few wallets the cardinality is low, but added for
+-- completeness.
+CREATE INDEX idx_key_scopes_wallet ON key_scopes (wallet_id);
```
yyforyongyu: Q: Given that we have `uidx_key_scopes_wallet_purpose_coin` i think this is not needed as `wallet_id` is already indexed?

yyforyongyu (commented):
Only a question, otherwise good to go!

mohamedawnallah (approved):

wallet/internal/db/migrations/sqlite/000004_key_scopes.up.sql
GustavoStingelin: Good question. I just learned about this PostgreSQL behavior after reading the docs here https://www.postgresql.org/docs/18/indexes-multicolumn.html. I could not find clear confirmation that SQLite behaves the same way. I think we should properly benchmark and tune the index and query later, based on real behavior in each engine. For now, I removed it.

>A multicolumn B tree index can be used with query conditions that involve any subset of the index columns, but it is most efficient when the constraints apply to the leading columns.

yyforyongyu (approved):

PR #1143, opened 2025-12-17
wallet: add P2A and P2PK address types

GustavoStingelin:
This PR adds two additional address types to avoid future schema migrations. The new types are `P2A` and `P2PK`, as suggested by YY in this review comment https://github.com/btcsuite/btcwallet/pull/1138#pullrequestreview-3587295552.

I intentionally did not include Silent Payment addresses for now. I think this deserves a more careful design later, since the `sp1...` address itself never appears on chain as a standard output. Instead, the sender derives a regular `bc1...` address from the silent payment address during transaction construction, which has different semantics compared to other address types stored today.

### DIFF
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -36,11 +36,18 @@ type KeyScope struct {
 type AddressType uint8
 
 const (
+	// RawPubKey represents a pay-to-pubkey (P2PK) address.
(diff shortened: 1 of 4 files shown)


yyforyongyu (approved):
LGTM 🧷

mohamedawnallah (approved):

PR #1144, opened 2025-12-19
Add RPC methods for `WalletController`

yyforyongyu:
This PR prepares the next `WalletController` PR, it,
- adds batch RPC methods to be used, the underlying methods are already built, we just need to expose them on the interface.
- adds several methods on `wtxmgr` and `waddrmgr` which are to be used in the next PR.

### DIFF
--- a/chain/btcd.go
+++ b/chain/btcd.go
@@ -38,6 +38,8 @@ type RPCClient struct {
 	wg      sync.WaitGroup
 	started bool
 	quitMtx sync.Mutex
+
+	batchClient *rpcclient.Client
 }
 
 // A compile-time check to ensure that RPCClient satisfies the chain.Interface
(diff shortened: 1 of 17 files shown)


chain/neutrino.go
```diff
@@ -176,12 +176,15 @@ func (s *NeutrinoClient) GetBlockHeight(hash *chainhash.Hash) (int32, error) {
 
 // GetBestBlock replicates the RPC client's GetBestBlock command.
 func (s *NeutrinoClient) GetBestBlock() (*chainhash.Hash, int32, error) {
-	chainTip, err := s.CS.BestBlock()
+	s.clientMtx.Lock()
+	defer s.clientMtx.Unlock()
```
GustavoStingelin: Q: why needs mutex?

chain/btcd.go
```diff
@@ -38,6 +38,8 @@ type RPCClient struct {
 	wg      sync.WaitGroup
 	started bool
 	quitMtx sync.Mutex
+
+	batchClient *rpcclient.Client
```
GustavoStingelin, line 42: Q: Missing shutdown of this client in `Stop()` function?

GustavoStingelin (approved):
LGTM! Comments can be addressed later.

mohamedawnallah (approved):

chain/neutrino.go
yyforyongyu: hmm removed

chain/btcd.go
yyforyongyu: nice catch - fixed!

GustavoStingelin:
Ack

PR #1145, opened 2025-12-22
Clean up deprecated code for upcoming db impl

yyforyongyu:
This PR does three things,
- move dead code into `deprecated.go` for clarity
- move code snippets to more appropriate files from legacy files
- fix linter errors - some of the existing code was not linted due to they are not touched, we now fix it

Most of the changes are pure code move, and it paves the way for upcoming controller and db impl.


mohamedawnallah (approved):
> move dead code into `deprecated.go` for clarity

👍

GustavoStingelin (approved):
ACK

PR #1147, opened 2025-12-31
wallet: add AccountStore implementation

### LINKED ISSUE #1106: [sqldb/address]: Implement account support

GustavoStingelin:
Closes #1106

This PR implements the AccountStore interface for both PG and SQLite.

It is a bit large, but I spent time keeping the functions concise and readable. The goal was to make the review as simple and smooth as possible.


wallet/internal/db/migrations/postgres/000004_key_scopes.up.sql
```diff
@@ -17,15 +17,24 @@ CREATE TABLE key_scopes (
...
     -- Reference to the address type used for external/receiving addresses.
     external_type_id SMALLINT NOT NULL,
 
+    -- Counter used to allocate sequential account numbers within this scope.
+    -- This avoids scanning the accounts table to compute MAX(account_number).
+    -- The value is updated atomically via UPDATE with RETURNING, which allows
+    -- concurrent account creation without additional locking logic.
+    -- The counter starts at minus one. Each new account consumes the current
+    -- value plus one, then stores the updated value for the next allocation.
+    last_account_number BIGINT NOT NULL DEFAULT -1,
```
yyforyongyu: I think this violates the Single Source of Truth principle, also seems like a premature optimization given that the accounts table is usually small. Plus `MAX(account_number)` can be fast if it's indexed, which is the case here?

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
```diff
@@ -0,0 +1,104 @@
+-- Migration note: Intentionally NOT idempotent (no "IF NOT EXISTS").
+-- This ensures migration tracking stays accurate and fails loudly if run twice.
+
+-- Account origin lookup table - provides standardized descriptions for wallet
+-- account origins. This is a reference table that maps the AccountOrigin enum
+-- values used in Go code to their human-readable names.
+CREATE TABLE account_origins (
+    -- Primary key matching the Go AccountOrigin enum values.
+    -- Using explicit IDs rather than auto-increment to ensure consistency
```
yyforyongyu, line 9: 👍

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
```diff
@@ -0,0 +1,104 @@
...
+    -- Master fingerprint is the fingerprint of the master pub key that created
+    -- this account.
+    master_fingerprint BIGINT,
+
+    -- Encrypted public key for the account.
+    encrypted_public_key BYTEA,
+
+    -- Foreign key constraints to key scope. Using ON DELETE RESTRICT to ensure
+    -- that the key scope cannot be deleted if accounts still exist.
+    FOREIGN KEY (scope_id) REFERENCES key_scopes (id) ON DELETE RESTRICT,
```
yyforyongyu, line 69: 👍

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
```diff
@@ -0,0 +1,104 @@
...
+
+    -- Defines if the account is watch-only.
+    is_watch_only BOOLEAN NOT NULL,
+
+    -- Master fingerprint is the fingerprint of the master pub key that created
+    -- this account.
+    master_fingerprint BIGINT,
+
+    -- Encrypted public key for the account.
+    encrypted_public_key BYTEA,
```
yyforyongyu, line 62: Q: can this field be null?

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
```diff
@@ -0,0 +1,104 @@
...
+
+    -- Reference to the origin of the account.
+    origin_id SMALLINT NOT NULL,
+
+    -- Defines if the account is watch-only.
+    is_watch_only BOOLEAN NOT NULL,
+
+    -- Master fingerprint is the fingerprint of the master pub key that created
+    -- this account.
+    master_fingerprint BIGINT,
```
yyforyongyu, line 59: Q: should this be BYTE?

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
```diff
@@ -0,0 +1,104 @@
...
+    -- Encrypted public key for the account.
+    encrypted_public_key BYTEA,
+
+    -- Foreign key constraints to key scope. Using ON DELETE RESTRICT to ensure
+    -- that the key scope cannot be deleted if accounts still exist.
+    FOREIGN KEY (scope_id) REFERENCES key_scopes (id) ON DELETE RESTRICT,
+
+    -- Foreign key constraint to account origins. Using ON DELETE RESTRICT to
+    -- ensure that the origin cannot be deleted if accounts still exist.
+    FOREIGN KEY (origin_id) REFERENCES account_origins (id) ON DELETE RESTRICT
```
yyforyongyu, line 73: what about a `created_at` field? like auto recording when the account is created

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
```diff
@@ -0,0 +1,104 @@
...
+ON accounts (scope_id, account_name);
+
+-- Account Secrets table to hold encrypted account-level secrets.
+CREATE TABLE account_secrets (
+    -- Reference to the account these keys belong to.
+    account_id BIGINT NOT NULL,
+
+    -- Encrypted private key for the account. Watch-only accounts may have
+    -- no row in this table.
+    encrypted_private_key BYTEA,
```
yyforyongyu: Q: should this be `NOT NULL`?

wallet/internal/db/migrations/sqlite/000004_key_scopes.up.sql
```diff
@@ -17,15 +17,24 @@ CREATE TABLE key_scopes (
...
     -- Reference to the address type used for external/receiving addresses.
     external_type_id INTEGER NOT NULL,
 
+    -- Counter used to allocate sequential account numbers within this scope.
+    -- This avoids scanning the accounts table to compute MAX(account_number).
+    -- The value is updated atomically via UPDATE with RETURNING, which allows
+    -- concurrent account creation without additional locking logic.
+    -- The counter starts at minus one. Each new account consumes the current
+    -- value plus one, then stores the updated value for the next allocation.
+    last_account_number INTEGER NOT NULL DEFAULT -1,
```
yyforyongyu: same re this field

wallet/internal/db/queries/postgres/accounts.sql
```diff
@@ -0,0 +1,226 @@
+-- name: CreateDerivedAccount :one
+-- Creates a new derived account under the given scope, allocating a fresh
+-- sequential account number from key_scopes.last_account_number.
+-- The allocation is atomic: the UPDATE takes the row lock on the scope row,
+-- returns the allocated number, and updates the counter for the next call.
+WITH allocated_number AS (
+    UPDATE key_scopes
+    SET last_account_number = last_account_number + 1
```
yyforyongyu: this we can do this instead, to remove the usage of the `last_account_number`,
```sql
INSERT INTO accounts (...)
SELECT $1, COALESCE(MAX(account_number), -1) + 1, ...
FROM accounts WHERE scope_id = $1
RETURNING id, account_number;
```

wallet/internal/db/queries/sqlite/accounts.sql
```diff
@@ -0,0 +1,220 @@
...
+--
+-- NOTE: Unlike Postgres, SQLite can't combine an UPDATE...RETURNING allocation
+-- step with an INSERT in a single CTE.
+--
+-- We instead:
+--  1) call AllocateAccountNumber (key_scopes.sql)
+--  2) call CreateDerivedAccount with the returned number
+--
+-- Both statements run within the same SQL transaction.
+INSERT INTO accounts (
```
yyforyongyu, line 5: Looks like there's the `BEGIN EXCLUSIVE TRANSACTION` we can use in sqlite

wallet/internal/db/queries/postgres/accounts.sql
```diff
@@ -0,0 +1,226 @@
...
+    a.origin_id,
+    a.is_watch_only,
+    ks.purpose,
+    ks.coin_type
+FROM accounts AS a
+INNER JOIN key_scopes AS ks ON a.scope_id = ks.id
+WHERE ks.wallet_id = $1
+ORDER BY a.account_number NULLS LAST;
+
+-- name: UpdateAccountNameByWalletScopeAndNumber :execrows
```
yyforyongyu, line 227: TIL - nice usage of `execrows`

wallet/internal/db/data_types.go
```diff
@@ -368,8 +367,9 @@ type ScopeAddrSchema struct {
 	InternalAddrType AddressType
 }
 
-// CreateAccountParams contains the parameters for creating a new account.
-type CreateAccountParams struct {
+// CreateDerivedAccountParams contains the parameters for creating a new derived
+// account.
+type CreateDerivedAccountParams struct {
```
yyforyongyu, line 400: 👍

wallet/internal/db/data_types.go
```diff
@@ -441,11 +434,11 @@ type AccountProperties struct {
...
-	// AccountPubKey is the account's public key that can be used to
-	// derive any address relevant to said account.
-	//
-	// NOTE: This may be nil for imported accounts.
-	AccountPubKey *hdkeychain.ExtendedKey
+	// EncryptedPublicKey is the encrypted account public key. This is the
+	// encrypted form of the extended public key that can be used to derive
+	// addresses for the account. The caller must decrypt this using the
+	// appropriate crypto key to use it.
+	EncryptedPublicKey []byte
```
yyforyongyu, line 476: nice!

wallet/internal/db/data_types.go
```diff
@@ -521,7 +514,7 @@ type RenameAccountParams struct {
 
 	// OldName is the current name of the account. This is used to identify
 	// the account if AccountNumber is not provided.
-	OldName string
+	OldName *string
```
yyforyongyu: Q: can account name be empty string?

wallet/internal/db/address_types_common.go
```diff
@@ -23,22 +48,12 @@ func addressTypeInfosFromRows[T any](rows []T,
...
-
-func getAddressTypeByID[T any, ID any](ctx context.Context,
-	get func(context.Context, ID) (T, error), queryID ID,
-	id AddressType, toInfo func(T) (AddressTypeInfo, error)) (
-	AddressTypeInfo, error) {
+// getAddressTypeByID is a generic helper that retrieves a single address type
+// by its ID and converts it to an AddressTypeInfo struct. It returns
+// ErrAddressTypeNotFound if no matching type is found.
+func getAddressTypeByID[R any, ID any](ctx context.Context,
+	get func(context.Context, ID) (R, error), queryID ID, id AddressType,
```
yyforyongyu: nit: get -> `getter`

wallet/internal/db/accounts_common.go
```diff
@@ -0,0 +1,475 @@
...
+	// possible if the database is modified incorrectly or the query is
+	// incorrect.
+	errNilDBAccountNumber = errors.New(
+		"database returned nil account number")
+
+	// errInvalidAccountOrigin is returned when an account origin ID from
+	// the database does not correspond to a known AccountOrigin value. In
+	// practice, this should never happen, but it's possible if the database
+	// is modified incorrectly or the query is incorrect.
+	errInvalidAccountOrigin = errors.New("invalid account origin")
```
yyforyongyu, line 21: Love all these defensive checks!

wallet/internal/db/accounts_common.go
```diff
@@ -0,0 +1,475 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("master fingerprint: %w", err)
+		}
+	}
+
+	return &AccountProperties{
+		AccountNumber:        accountNum,
+		AccountName:          accountName,
+		Origin:               origin,
+		ExternalKeyCount:     0,
```
yyforyongyu, line 126: hmm why are these keys 0? my guess is we will populate these fields once address table is done?

wallet/internal/db/accounts_common.go
```diff
@@ -0,0 +1,475 @@
...
+	if err == nil {
+		return id, nil
+	}
+
+	if !errors.Is(err, sql.ErrNoRows) {
+		// A real database error occurred (not a conflict).
+		return 0, fmt.Errorf("create key scope: %w", err)
+	}
+
+	// ErrNoRows means the scope was created concurrently by another process
```
yyforyongyu, line 314: 👍

wallet/internal/db/accounts_common.go
```diff
@@ -0,0 +1,475 @@
...
+
+// accountPropsRowToProps converts a database row containing full account
+// properties into an AccountProperties struct. The idToAddrType function is
+// used to convert the internal and external address type IDs to AddressType
+// values.
+//
+// TODO(stingelin): Add address counting support after address management is
+// implemented.
+func accountPropsRowToProps[AddrTypeId any](
+	accountNumber sql.NullInt64, accountName string, originID int64,
```
yyforyongyu: nit: It looks like some of the methods are taking lots of params, think we can define a `rowAccountProps` here instead? this is purely aesthetic tho, as it doesn't reduce complexity.

wallet/internal/db/accounts_common.go
```diff
@@ -0,0 +1,475 @@
...
+// getAccount is a generic helper that retrieves an account using the provided
+// query function. It handles error mapping and delegates conversion to the
+// toInfo function.
+func getAccount[T any, Args any](ctx context.Context,
+	get func(context.Context, Args) (T, error), args Args,
+	query GetAccountQuery, toInfo func(T) (*AccountInfo, error)) (
+	*AccountInfo, error) {
+
+	row, err := get(ctx, args)
+	if err != nil {
```
yyforyongyu: can perform `if err == nil` check instead to save some indentations, a common technique we use here

wallet/internal/db/accounts_common.go
```diff
@@ -0,0 +1,475 @@
...
+func renameAccount[Args any](ctx context.Context,
+	update func(context.Context, Args) (int64, error), args Args,
+	params RenameAccountParams) error {
+
+	rowsAffected, err := update(ctx, args)
+	if err != nil {
+		return fmt.Errorf("rename account: %w", err)
+	}
+
+	if rowsAffected == 0 {
```
yyforyongyu: similarly, can check if `rowsAffected != 0` first

wallet/internal/db/accounts_pg.go
```diff
@@ -0,0 +1,365 @@
...
+var _ AccountStore = (*PostgresWalletDB)(nil)
+
+// GetAccount retrieves information about a specific account, identified by its
+// name or account number within a given key scope.
+func (w *PostgresWalletDB) GetAccount(ctx context.Context,
+	query GetAccountQuery) (*AccountInfo, error) {
+
+	getQueries := pgAccountGetQueries{q: w.queries}
+
+	return getAccountByQuery(ctx, query,
```
yyforyongyu: nit: formatting for multi lines, this should be,
```go
getAccountByQuery(
    ctx, query,...,
)
```

same for other places

wallet/internal/db/interface.go
```diff
@@ -25,6 +25,41 @@ var (
...
+	ErrAccountNotFound = errors.New("account not found")
+
+	// ErrKeyScopeNotFound is returned when a key scope is not found in the
+	// database.
+	ErrKeyScopeNotFound = errors.New("key scope not found")
+
+	// ErrUnknownKeyScope is returned when a key scope is not found
+	// in ScopeAddrMap.
+	ErrUnknownKeyScope = errors.New(
+		"unknown scope in ScopeAddrMap",
```
yyforyongyu: nitty-nit: looks like it can fit in oneline

wallet/internal/db/itest/fixtures_common_test.go
```diff
@@ -0,0 +1,263 @@
...
+// CreateImportedWalletParams creates test parameters for an imported wallet.
+func CreateImportedWalletParams(name string) db.CreateWalletParams {
+	params := CreateWalletParamsFixture(name)
+	params.IsImported = true
+
+	return params
+}
+
+// CreateWatchOnlyWalletParams creates test parameters for a watch-only wallet.
+func CreateWatchOnlyWalletParams(name string) db.CreateWalletParams {
```
yyforyongyu, line 38: nice breakdown!

wallet/internal/db/itest/fixtures_common_test.go
```diff
@@ -0,0 +1,263 @@
...
+	_, err := rand.Read(h[:])
+	if err != nil {
+		// This should never happen.
+		panic(fmt.Sprintf("failed to generate random hash: %v", err))
+	}
+
+	return h
+}
+
+func NewBlockFixture(height uint32) db.Block {
```
yyforyongyu, line 77: nit: missing godoc

wallet/internal/db/itest/fixtures_common_test.go
```diff
@@ -0,0 +1,263 @@
...
+	// IsWatchOnly indicates whether the account has no private key.
+	// Only relevant for imported accounts.
+	IsWatchOnly bool
+}
+
+// DerivedAccountCases contains derived account fixtures across all standard
+// key scopes, with multiple accounts per scope.
+var DerivedAccountCases = []AccountTestCase{
+	{
+		Name:   "derived-bip84-default",
```
yyforyongyu, line 110: nice list - I wonder if we should define this in the store interface instead, such that we have a list of default scopes, which can be used everywhere, like initializing the wallet, loading accounts during startup, etc.

wallet/internal/db/itest/account_store_test.go
```diff
@@ -0,0 +1,729 @@
...
+
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCreateAccounts verifies that Create**Account correctly creates accounts
+// across all standard key scopes between multiple wallets.
+func TestCreateAccounts(t *testing.T) {
+	t.Parallel()
+	store, _ := NewTestStore(t)
```
yyforyongyu, line 24: perform a require.NoError here?

wallet/internal/db/itest/account_store_test.go
```diff
@@ -0,0 +1,729 @@
...
+
+	for i := range 3 {
```
yyforyongyu, line 27: add some comments explaining the number `3`?

yyforyongyu (commented):
Very nice!! I still need to review the last commits which added the tests, meanwhile, my main comment is that we can remove the `last_account_number`, the rest are nits. I think it's very close!

wallet/internal/db/queries/postgres/accounts.sql
GustavoStingelin: Yeah, but the downside is that we lose the locking capability. If two inserts happen at the same time, we can end up with a conflict error on the second query. To handle that, we would need to do one of the following.

1. Implement a retry mechanism, which can still get exhausted and fail.
2. Add a mutex on the Go side to prevent concurrent inserts.
3. Use a channel based approach to guarantee that only one call happens at a time.
4. Write more complex code to lock rows while computing the max and inserting. That was actually my first attempt, but I moved away from it because the code was getting messy.

Given this context, do you still think removing the max account makes sense? If so, how would you suggest handling concurrent inserts safely?

yyforyongyu: Correct - we always have this kind issue when dealing with concurrent writing. However, when thinking about this kinda issues, IMO we should always favor data integrity, as there are ways to circumvent the concurrency issues. 

For instance, one way to do this is to use advisory locks or explicit locks, e.g., we can use `SELECT ... FOR UPDATE` to lock the `key_scopes`, and for sqlite we could use `BEGIN IMMEDIATE` (which is already handled automatically via the connection setting `_txlock=immediate` IIUC, plus it only allows one writer anyway)

so sth like this should prevent concurrent writing?
```sql
WITH lock_scope AS (
    SELECT id
    FROM key_scopes
    WHERE id = $1
    FOR UPDATE
)
INSERT INTO accounts (
    scope_id,
    account_number,
    account_name,
    origin_id,
    encrypted_public_key,
    master_fingerprint,
    is_watch_only
)
SELECT
    lock_scope.id,
    (SELECT COALESCE(MAX(a.account_number), -1) + 1 FROM accounts a WHERE a.scope_id = lock_scope.id),
    $2, $3, $4, $5, $6
FROM lock_scope
```

And pragmatically speaking, I don't think we will have high throughput in the accounts table, concurrent access is rare, and we can use the above lock to handle it, at the cost of increasing db lock contention.

GustavoStingelin: Cool, will do this way

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
GustavoStingelin: For derived accounts, we should persist only the derivation information. The key itself should be derived in memory, starting from the master key down to the account.

Otherwise, we cannot create ad-hoc accounts only in database side. The application would first need to discover the next account ID, derive the corresponding public key, and only then persist the complete data.

At the moment, the implementation assumes that dbDefaultAccountRow already contains the public key. If we want to keep this behavior, we would need to change the db.interface to behave as a pure storage layer and move all derivation logic to a higher layer, which would be responsible for passing the account ID and keys.

What do you think?

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
GustavoStingelin: IDK, the original kvdb was storing it as bytes, but every call treats it as an int. That is why I persisted it as an int. Do you think we should change it?

wallet/internal/db/accounts_common.go
GustavoStingelin: yes, for now we do not have this info

wallet/internal/db/migrations/postgres/000004_key_scopes.up.sql
GustavoStingelin: changed in the last commit, take a look

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
GustavoStingelin: added

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
GustavoStingelin: fixed

wallet/internal/db/migrations/sqlite/000004_key_scopes.up.sql
GustavoStingelin: changed in the last commit, take a look

wallet/internal/db/queries/postgres/accounts.sql
GustavoStingelin: changed in the last commit, take a look

wallet/internal/db/queries/sqlite/accounts.sql
GustavoStingelin: `txlock=immediate` already handles this, and we will need this parameter anyway to allow concurrent queries

wallet/internal/db/data_types.go
GustavoStingelin: I felt that the pointer here looks more ergonomic, but I changed to the empty string approach

wallet/internal/db/address_types_common.go
GustavoStingelin: fixed

wallet/internal/db/accounts_common.go
GustavoStingelin: really cool ideia, fixed

wallet/internal/db/accounts_common.go
GustavoStingelin: really cool idea, fixed

wallet/internal/db/accounts_pg.go
GustavoStingelin: fixed

wallet/internal/db/itest/fixtures_common_test.go
GustavoStingelin: added

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: That is not needed because the underlying `NewTestStore` already tests this error. Also, the unused returned parameter is not an error, it is the querier, which can be useful in some tests.

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: added

wallet/internal/db/accounts_common.go
GustavoStingelin: added the structs

wallet/internal/db/interface.go
GustavoStingelin: fixed

wallet/internal/db/itest/fixtures_common_test.go
GustavoStingelin: I think the idea here is to include all common configs for tests, but in real usage, a user may create only a small number of accounts, or just one. Since we will need to load data from the database anyway, this feels more test-focused to me.

config/.golangci.yml
```diff
@@ -48,7 +48,12 @@ linters:
     funlen:
       # Checks the number of lines in a function.
       # If lower than 0, disable the check.
-      lines: 100
+      # Increased from 100 to 200 to allow more verbose parameters were still
```
yyforyongyu: Typo: "were still keep" -> "while still keeping"

wallet/internal/db/accounts_pg.go
```diff
@@ -0,0 +1,398 @@
...
+// accounts have NULL account_number since they don't follow BIP44 derivation.
+func (w *PostgresWalletDB) CreateImportedAccount(ctx context.Context,
+	params CreateImportedAccountParams) (*AccountProperties, error) {
+
+	var props *AccountProperties
+
+	err := w.ExecuteTx(ctx, func(qtx *sqlcpg.Queries) error {
+		var err error
+
+		props, err = createImportedAccount(
```
yyforyongyu, line 128: 😿 guess that's the price we pay for this split db schema approach, the code block looks quite dense - not a priority tho, we can always fix it in the future.

wallet/internal/db/accounts_pg.go
```diff
@@ -0,0 +1,398 @@
...
+			ctx, qtx, params.WalletID, params.Scope,
+		)
+		if err != nil {
+			return err
+		}
+
+		// Acquire an advisory lock for this scope to serialize account creation
+		// and prevent race conditions when computing MAX(account_number). This
+		// MUST be a separate statement that completes before
+		// CreateDerivedAccount runs. See the LockAccountScope comments for why
```
yyforyongyu: nit: `CreateDerivedAccount` -> `qtx.CreateDerivedAccount` as for a moment I mixed `PostgresWalletDB.CreateDerivedAccount` with this one...

wallet/internal/db/accounts_sqlite.go
```diff
@@ -0,0 +1,393 @@
...
+	var info *AccountInfo
+
+	err := w.ExecuteTx(ctx, func(qtx *sqlcsqlite.Queries) error {
+		scopeID, err := sqliteEnsureKeyScope(
+			ctx, qtx, params.WalletID, params.Scope,
+		)
+		if err != nil {
+			return err
+		}
+
```
yyforyongyu, line 68: Now the independent db schema approach shines! We can easily leverage db-specific properties to fix the races - will keep this as an example to show others why this approach is superior!

wallet/internal/db/address_types_common.go
```diff
@@ -7,10 +7,35 @@ import (
...
+	if v < 0 || v > T(Anchor) {
+		return 0, fmt.Errorf("%w: %d", errInvalidAddressType, v)
+	}
+
+	return AddressType(v), nil
+}
+
+// listAddressTypes is a generic helper that retrieves all address types from
+// the database and converts them to AddressTypeInfo structs.
+func listAddressTypes[R any](ctx context.Context,
```
yyforyongyu: Q: i thought we usually use `T any` in generics? so `R` stands for `return` or `row`?

wallet/internal/db/itest/account_store_test.go
```diff
@@ -0,0 +1,821 @@
...
+		t.Context(), db.CreateDerivedAccountParams{
+			WalletID: walletID,
+			Scope:    db.KeyScopeBIP0084,
+			Name:     "account-overflow",
+		},
+	)
+	require.ErrorIs(t, err, db.ErrMaxAccountNumberReached)
+}
+
+func newWallet(t *testing.T, store db.WalletStore, name string) uint32 {
```
yyforyongyu, line 701: nit: we are missing godocs for these methods

wallet/internal/db/itest/fixtures_pg_test.go
```diff
@@ -0,0 +1,65 @@
...
+
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	sqlcpg "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/postgres"
+	"github.com/stretchr/testify/require"
+)
+
+// CreateBlockFixture inserts a test block into the database and returns it.
+func CreateBlockFixture(t *testing.T, queries *sqlcpg.Queries,
+	height uint32) db.Block {
+	t.Helper()
```
yyforyongyu, line 17: weird linter didn't pick this up - but we should add a newline after `t.Helper` - also I think it's fine to **not** fix it in this PR, as we should figure out whether this can be validated by linter instead and automate it.

wallet/internal/db/itest/sqlite_test.go
```diff
@@ -23,6 +24,17 @@ func NewSQLiteDB(t *testing.T) *sql.DB {
 	// Enable foreign keys (required for proper constraint enforcement).
 	dsn := dbPath + "?_pragma=foreign_keys=on"
 
+	// Enable WAL mode for better concurrency. WAL allows multiple readers and
+	// reduces lock contention for concurrent writers.
+	dsn = dsn + "&_pragma=journal_mode=WAL"
```
yyforyongyu, line 29: We should fix this in the next PR - we should ensure that when in production, the settings like `_txlock=immediate` are explicitly configured, to ensure the test env matches as much as possible.

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
yyforyongyu: yeah classic chicken egg problem - let's keep it null and handle it in the code instead

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
yyforyongyu: in that case let's use int!

wallet/internal/db/queries/postgres/accounts.sql
```diff
@@ -0,0 +1,268 @@
+-- name: LockAccountScope :exec
+-- Acquires a transaction-level advisory lock to serialize account creation within a scope.
```
yyforyongyu, line 2: 👍

yyforyongyu (approved):
Left some questions and nits, feel free to address them in a follow-up PR, nice work!

GustavoStingelin:
### Coverage reports:

<details open><summary>accounts_common</summary>
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:25:			validate				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:34:			validate				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:47:			isWatchOnly				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:76:			accountPropsRowToProps			63.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:146:		accountInfosFromRows			85.7%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:164:		listAccounts				75.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:178:		getAddrSchemaForScope			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:195:		buildAccountInfo			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:216:		idToAccountOrigin			66.7%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:239:		accountRowToInfo			75.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:278:		ensureKeyScope				58.8%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:330:		getAccountByQuery			75.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:353:		listAccountsByQuery			80.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:377:		renameAccountByQuery			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:399:		getAccount				87.5%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:426:		renameAccount				75.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_common.go:452:		createImportedAccount			94.4%
</details> 


<details><summary>accounts_sqlite</summary>
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:16:			GetAccount				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:26:			ListAccounts				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:38:			RenameAccount				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:51:			CreateDerivedAccount			95.2%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:108:		sqliteEnsureKeyScope			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:140:		CreateImportedAccount			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:177:		sqliteBuildCreateImportedAccountArgs	100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:200:		sqliteBuildCreateAccountSecretArgs	100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:214:		sqliteGetAccountProps			75.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:255:		sqliteAccountRowToInfo			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:280:		byScope					100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:294:		byName					100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:307:		all					100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:322:		byNumber				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:340:		byName					100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:360:		byNumber				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_sqlite.go:380:		byName					100.0%
</details> 

<details><summary>accounts_pg</summary>
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:16:			GetAccount				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:26:			ListAccounts				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:38:			RenameAccount				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:51:			CreateDerivedAccount			91.7%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:120:			CreateImportedAccount			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:151:			pgBuildCreateImportedAccountArgs	100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:174:			pgBuildCreateAccountSecretArgs		100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:188:			pgGetAccountProps			75.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:215:			pgEnsureKeyScope			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:261:			pgAccountRowToInfo			100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:284:			byScope					100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:298:			byName					100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:311:			all					100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:326:			byNumber				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:344:			byName					100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:365:			byNumber				100.0%
github.com/btcsuite/btcwallet/wallet/internal/db/accounts_pg.go:385:			byName					100.0%
</details>

PR #1148, opened 2025-12-31
Add docs and `Controller` interface

yyforyongyu:
In this PR,
- add ADRs for the controller implementation
- add the interface definition
- add the `syncer` skeleton and wallet state

### DIFF
--- a/docs/developer/scanning_sync_architecture.md
+++ b/docs/developer/scanning_sync_architecture.md
@@ -1,0 +1,167 @@
+# Wallet Synchronization and Scanning Architecture
+
+This document details the architecture of the `btcwallet` synchronization subsystem. It explains how the wallet maintains consensus with the blockchain, discovers relevant transactions, and manages the recovery of funds.
+
+## 1. High-Level Architecture
+
+The synchronization system is designed around a **Controller-Worker-State** pattern, separating the public API from the background work and the core logic.
+
+```mermaid
+graph TD
+    User[User / RPC] -->|Calls Start/Rescan/Unlock| Controller
+    
+    subgraph "Wallet Package"
+        Controller[Controller] -->|Manages| State[Wallet State]
+        Controller -->|Sends Req| Syncer[Syncer]
+        
+        Syncer -->|Maintains| RecoveryState[Recovery State]
+        Syncer -->|Reads/Writes| DB[(Wallet DB)]
+        Syncer -->|Fetches Data| Chain[Chain Backend]
+    end
+```
+
+### 1.1 Key Components
+
+*   **Controller (`wallet/controller.go`)**: The public face of the wallet. It manages the wallet's lifecycle (`Start`, `Stop`), handles authentication (`Lock`, `Unlock`), and acts as the gatekeeper for state transitions. It does *not* perform blocking chain operations directly.
+*   **Syncer (`wallet/syncer.go`)**: A dedicated background worker responsible for the main synchronization loop. It communicates with the chain backend (e.g., `bitcoind`, `neutrino`), orchestrates batch scanning, and handles blockchain reorganizations (rollbacks).
+*   **RecoveryState (`wallet/recovery.go`)**: A specialized state machine that encapsulates the logic for *what* to scan for. It manages BIP32 derivation horizons, address lookahead windows, and the set of watched outpoints. It is purely logic and memory-based, decoupled from the I/O mechanisms of the Syncer.
+
+---
+
+## 2. State Management: The Orthogonal Model
+
+To manage concurrency and API availability safely, the wallet employs an **Orthogonal State Model**. Instead of a single monolithic status (e.g., "Syncing"), we track three independent dimensions of state. This decoupling allows for precise representation of complex conditions (e.g., a wallet can be "Started" AND "Syncing" AND "Locked") without state explosion.
+
+### 2.1 Lifecycle (System State)
+Tracks the runtime status of the wallet's main event loop and background processes.
+*   **Stopped**: The wallet is idle. No background routines are running.
+*   **Starting**: The wallet is in the middle of its synchronous startup sequence (e.g., loading accounts, verifying birthday).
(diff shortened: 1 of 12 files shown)


docs/developer/scanning_sync_architecture.md
```diff
@@ -0,0 +1,167 @@
...
+## 7. Strategy Comparison & Selection
+
+The wallet automatically selects the best strategy based on the environment and state.
+
+| Feature | Full Block Scanning | CFilter Scanning |
+| :--- | :--- | :--- |
+| **Bandwidth** | High (All data) | Low (Headers + Filters + Matches) |
+| **CPU Usage** | Low (Hash map lookups) | High (Elliptic Curve ops + SIPHash matching) |
+| **Latency** | Low (Local) / High (Remote) | Low (Parallel Fetch) |
+| **Privacy** | Poor (if remote RPC) | Good (Neutrino) |
```
GustavoStingelin: Q: Why is privacy considered poor if we need to scan all blocks in the full block scanning approach?
From the remote node perspective, it only sees that you are requesting many blocks. How would it learn which transactions belong to you?

docs/developer/scanning_sync_architecture.md
```diff
@@ -0,0 +1,167 @@
...
+        Syncer -->|Reads/Writes| DB[(Wallet DB)]
+        Syncer -->|Fetches Data| Chain[Chain Backend]
+    end
+```
+
+### 1.1 Key Components
+
+*   **Controller (`wallet/controller.go`)**: The public face of the wallet. It manages the wallet's lifecycle (`Start`, `Stop`), handles authentication (`Lock`, `Unlock`), and acts as the gatekeeper for state transitions. It does *not* perform blocking chain operations directly.
+*   **Syncer (`wallet/syncer.go`)**: A dedicated background worker responsible for the main synchronization loop. It communicates with the chain backend (e.g., `bitcoind`, `neutrino`), orchestrates batch scanning, and handles blockchain reorganizations (rollbacks).
+*   **RecoveryState (`wallet/recovery.go`)**: A specialized state machine that encapsulates the logic for *what* to scan for. It manages BIP32 derivation horizons, address lookahead windows, and the set of watched outpoints. It is purely logic and memory-based, decoupled from the I/O mechanisms of the Syncer.
```
GustavoStingelin, line 27: Maybe it would be better to call this RescanState? To stay aligned with the previous ADRs? It is easier to understand that we are talking about finding the funds. To me, “Recovery” sounds more destructive and is usually associated with data loss scenarios, like recovering after a power outage that left the disk in an inconsistent state.

wallet/state.go
```diff
@@ -0,0 +1,274 @@
...
+	// the wallet always starts in a secure state.
+	s.unlocked.Store(false)
+
+	return nil
+}
+
+// toStarted marks the wallet as fully started. This should be called only
+// after all resource initialization is complete.
+func (s *walletState) toStarted() {
+	s.lifecycle.Store(uint32(lifecycleStarted))
```
GustavoStingelin: Q: Why not compare and swap here?

wallet/state.go
```diff
@@ -0,0 +1,274 @@
...
+	if !s.lifecycle.CompareAndSwap(
+		uint32(lifecycleStopped), uint32(lifecycleStarting)) {
+
+		return fmt.Errorf("%w: current state is %v",
+			ErrWalletAlreadyStarted, lifecycle(s.lifecycle.Load()))
+	}
+
+	// 2. Authentication (Security State): Reset to Locked. This ensures
+	// the wallet always starts in a secure state.
+	s.unlocked.Store(false)
```
GustavoStingelin, line 132: Maybe we also need to ensure that private keys are zeroed from memory? Otherwise, we could end up in a situation where the wallet appears locked, but all private data could still be extracted from a memory dump.

The Locker could be a specialized manager that handles all private data internally and exposes whether it is locked or unlocked.

wallet/wallet.go
```diff
@@ -87,6 +88,109 @@ var (
...
+// Config holds the configuration options for creating a new
+// WalletController.
+type Config struct {
+	// DB is the underlying database for the wallet.
+	DB walletdb.DB
+
+	// Chain is the interface to the blockchain (e.g. bitcoind,
+	// neutrino). If set, the wallet will automatically synchronize with
+	// the chain upon Start.
+	Chain chain.Interface
```
GustavoStingelin, line 157: Maybe handling the underlying DB and Chain makes it more than just a config holder, and closer to a manager or controller?

GustavoStingelin (commented):
Looking really good! Just left some comments.

docs/developer/scanning_sync_architecture.md
yyforyongyu: totally - atm it's called `RecoveryState` because we are using the existing struct to build the rescan, I think we can use a small PR in the end to perform the rename.

docs/developer/scanning_sync_architecture.md
yyforyongyu: yeah it actually should be reversed as the cfilters actually tells the remote RPC which blocks are of interests, updated!

wallet/state.go
yyforyongyu: Good question - this is currently done in the `Unlock/Lock` on the controller, which will be clearer in later PRs. A good point is also to do this when the wallet is stopped, will see how to add that, meanwhile, for `walletState`, we will keep it simple and only manage the state variables.

wallet/state.go
yyforyongyu: the original thought was, "CAS to enter a transient state, Store to finalize it", but it does introduce some edgy races. Changed to `CompareAndSwap` instead.

wallet/wallet.go
yyforyongyu: For a single-wallet design, I think the DB and Chain are config values as they stay static across the lifetime. There will be a final PR tho, that we move these values to the wallet manager instead, which again will be put on the manager's config struct.

GustavoStingelin (approved):
Ok

PR #1152, opened 2026-01-13
 Prepare upcoming `Controller`

yyforyongyu:
Accidentally deleted #1146, reopen it here.

We did a cleanup in #1145, however there are still a few more cleanups needed. In addition, 
- we fixed the case where duplicated tx notifications are sent - we don't need it as we already listen for `case chain.FilteredBlockConnected:`, in which the relevant txns are sent in batch inside the notification already: https://github.com/btcsuite/btcwallet/blob/f358d00bcfc3b77c64252d09ec36b811ed530649/wallet/chainntfns.go#L176-L178
- added dedup logic for PSBT's unknown fields
- fixed the mockChain impl

### DIFF
--- a/.gemini/config.yaml
+++ b/.gemini/config.yaml
@@ -33,4 +33,4 @@ code_review:
 
 # List of glob patterns to ignore (files and directories).
 # Type: array of string, default: [].
-ignore_patterns: []
+ignore_patterns: ["deprecated.go"]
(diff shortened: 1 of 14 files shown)


GustavoStingelin (approved):

yyforyongyu:
Also reviewed by LLMs.

PR #1155, opened 2026-01-15
Separate DB operations from rescan logic

yyforyongyu:
We now introduce a few preliminary methods to help us separate the DB logic from business logic, and expand the `RecoveryState` to handle the ephemeral state to further isolate the DB operations.

### DIFF
--- a/wallet/db_ops.go
+++ b/wallet/db_ops.go
@@ -1,0 +1,737 @@
+// Package wallet provides the implementation of a Bitcoin wallet.
+//
+// TODO(yy): This file will be removed once the Store implementation is
+// finished.
+package wallet
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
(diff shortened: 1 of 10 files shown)


wallet/db_ops.go
```diff
@@ -0,0 +1,737 @@
...
+
+// DBGetAllAccounts ensures all account properties are loaded into the address
+// manager's cache.
+//
+// TODO(yy): Refactor this in the `Store` implementation - we can call
+// `ListAccounts` instead, without the balance info.
+func (w *Wallet) DBGetAllAccounts(_ context.Context) error {
+	scopes := w.addrStore.ActiveScopedKeyManagers()
+
+	err := walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
```
GustavoStingelin: Maybe it could be View instead of Update?

GustavoStingelin (approved):
LGMT, just left one comment

wallet/db_ops.go
yyforyongyu: updated!

PR #1156, opened 2026-01-15
Use new db methods in wallet controller

yyforyongyu:
We now apply the new db methods and implement the `Controller` interface.

### DIFF
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -123,6 +123,100 @@ type Controller interface {
 		targets []waddrmgr.AccountScope) error
 }
 
+// Start starts the background processes necessary to manage the wallet.
+//
+// This is part of the Controller interface.
+func (w *Wallet) Start(startCtx context.Context) error {
+	// 1. Attempt to transition from Stopped to Starting.
+	err := w.state.toStarting()
+	if err != nil {
+		return err
+	}
+
+	// 2. Setup background resources.
+	//
+	// w.lifetimeCtx governs the lifecycle of all background goroutines.
+	// It is canceled when stop() is called.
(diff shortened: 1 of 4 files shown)


wallet/controller.go
```diff
@@ -122,3 +123,517 @@ type Controller interface {
...
+
+	// Report the result back to the caller.
+	req.resp <- err
+}
+
+// sendReq sends an operation request to the main loop or handles cancellation.
+func (w *Wallet) sendReq(ctx context.Context, req any) error {
+	select {
+	case w.requestChan <- req:
+		return nil
```
GustavoStingelin, line 616: Q: We are only initializing this in the test scenario with a buffer size of 1. Will this be the same in production?

If so, maybe the behavior is effectively the same as not using channels at all? Since the sender still has to wait for the single buffer slot to become available or hit a timeout.

If production uses a larger buffer, for example, size 10, should we not test with the same buffer size to better reflect real behavior?

GustavoStingelin (approved):
Looking good, just left one comment that can be addressed later.

Jason-123-cyber (approved):
Kk

wallet/controller.go
yyforyongyu: Correct - we use a buffer size of 1 to enforce that the requests are handled sequentially in the main loop, this is also a technique we use here a lot to centralize the lifecycle management of a system - by moving them into one single main loop, it's easier to reason about the flow, and add other concurrent actions, in this case check `<-w.lockTimer.C` and the quit signal.

Btw we will have another PR to add more tests to make sure this loop is robust.

PR #1159, opened 2026-01-23
Controller: patch more unit tests to increase the coverage rate

yyforyongyu:
We now add more unit tests to increase the coverage rate - also found a bug and fixed in the last commit.

This PR has been reviewed [here](https://github.com/yyforyongyu/btcwallet/pull/8) already.


GustavoStingelin (approved):
LGTM!

PR #1160, opened 2026-01-26
Cleanup deprecated code post `Controller` implementation

yyforyongyu:
We now perform a cleanup given that the `Controller` is now implemented, in this PR,
- all the new implementations now use the new `w.cfg` fields
- properly validate the wallet state when calling public RPCs
- update the old tests to use the new `Controller` 
- move deprecated code and tests into dedicated files for maintenance.

The code has been reviewed [here](https://github.com/yyforyongyu/btcwallet/pull/9).


wallet/account_manager_test.go
```diff
@@ -29,33 +28,83 @@ func TestNewAccount(t *testing.T) {
...
+	w, deps := createStartedWalletWithMocks(t)
 
 	// We'll start by creating a new account under the BIP0084 scope. We
 	// expect this to succeed.
 	scope := waddrmgr.KeyScopeBIP0084
-	account, err := w.NewAccount(
-		t.Context(), scope, testAccountName,
-	)
+
+	deps.addrStore.On("FetchScopedKeyManager", scope).
```
yyforyongyu, line 90: In the end, all these will be removed and replaced by a mock on the `Store` interface.

.gemini/styleguide.md
```diff
@@ -9,7 +9,7 @@
 - Unit tests must always use the `require` library. Either table driven unit 
   tests or tests using the `rapid` library are preferred. 
 - The line length MUST NOT exceed 80 characters, this is very important. 
-  You must count the Golang indentation (tabulator character) as 8 spaces when 
+  You must count the Golang indentation (tabulator character) as 4 spaces when
```
GustavoStingelin, line 12: so, here was that weird bug, hahaha

wallet/signer.go
```diff
@@ -541,6 +546,11 @@ func (w *Wallet) fetchManagedPubKeyAddress(path BIP32Path) (
 func (w *Wallet) ECDH(_ context.Context, path BIP32Path,
 	pub *btcec.PublicKey) ([32]byte, error) {
 
+	err := w.state.canSign()
```
GustavoStingelin, line 549: 👍🏻

GustavoStingelin (approved):
🚢 🚢 🚢

PR #1162, opened 2026-01-26
wallet: implement AddressStore

### LINKED ISSUE #1107: [sqldb/address]: Implement address support
Create migrations for addresses, address_secrets, and address_sync_statuses (ref #1051)

GustavoStingelin:
This PR aims to implement the SQL parts of the AddressStore interface. It is probably not final, but it is a solid starting point, and we can iterate on it as needed.

Closes #1107.


wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,420 @@
...
+		return nil, err
+	}
+
+	witnessVersion, err := witnessVersionToPtr(row.WitnessVersion)
+	if err != nil {
+		return nil, err
+	}
+
+	isWatchOnly := origin == ImportedAccount && !row.HasPrivateKey &&
+		!row.HasScript
```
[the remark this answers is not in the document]
GustavoStingelin: script is the private script, not the script pub key.

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,420 @@
...
+		return true
+	}
+
+	if len(p.EncryptedScript) > 0 {
+		return true
+	}
+
+	return false
+}
+
```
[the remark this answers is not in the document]
GustavoStingelin: no wayyyy, fixed.

wallet/internal/db/addresses_sqlite.go
```diff
@@ -0,0 +1,369 @@
...
+	}, nil
+}
+
+// insertAddressSecretIfNeededSQlite inserts an address secret row if the
+// address is not watch-only.
+func insertAddressSecretIfNeededSQlite(ctx context.Context,
+	qtx *sqlcsqlite.Queries, addressID int64,
+	params NewImportedAddressParams) error {
+
+	if params.isWatchOnly() {
```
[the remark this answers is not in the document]
GustavoStingelin: fixed.

wallet/internal/db/itest/address_store_test.go
```diff
@@ -0,0 +1,984 @@
...
+			require.NotZero(t, info.ID)
+			require.NotZero(t, info.AccountID)
+			require.Equal(t, db.ImportedAccount, info.Origin)
+			require.NotZero(t, info.CreatedAt)
+			require.Equal(t, uint32(0), info.Branch)
+			require.Equal(t, uint32(0), info.Index)
+			require.NotNil(t, info.EncryptedPubKey)
+			require.NotNil(t, info.EncryptedScriptPubKey)
+			require.Equal(t, tc.expectedAddrType, info.AddrType)
+			require.Equal(t, tc.providePrivateKey, info.IsWatchOnly)
```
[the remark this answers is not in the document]
GustavoStingelin: fixed.

wallet/internal/db/addresses_sqlite.go
```diff
@@ -0,0 +1,369 @@
...
+		Origin:      DerivedAccount,
+		Branch:      branch,
+		Index:       index,
+		IsWatchOnly: false,
+	}, nil
+}
+
+// insertAddressSecretIfNeededSQlite inserts an address secret row if the
+// address is not watch-only.
+func insertAddressSecretIfNeededSQlite(ctx context.Context,
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/data_types.go
```diff
@@ -167,17 +166,22 @@ const (
...
-// outputs.
-type Tapscript struct {
-	// ControlBlock is the control block for the Taproot script, which is
-	// required to reveal the script path during spending.
-	ControlBlock []byte
+// AddressSyncStatus represents the synchronization status of an address.
+//
+// The enum values MUST match the IDs in the address_sync_statuses database
+// table. See migration 000006_addresses for the canonical descriptions.
+type AddressSyncStatus uint8
```
yyforyongyu: I don't think this is needed - based on this [design decision](https://github.com/btcsuite/btcwallet/blob/sql-wallet/docs/developer/adr/0005-no-auto-rescan-on-import.md), users now have to call rescan to tell the wallet to trigger a rescan.

More importantly, even when an address is in `SyncStatusFull`, it doesn't provide utility - say it reaches this status on block X, but there's no guarantee that it won't receive another utxo at block X+1. I think this is largely from the legacy code, and can now be removed.

wallet/internal/db/data_types.go
```diff
@@ -570,36 +574,85 @@ type RenameAccountParams struct {
...
+	// NOTE: uint32 is used to ensure compatibility with standard SQL
+	// databases (signed 64-bit integers).
+	AccountID uint32
+
+	// AddrType is the type of address (P2PKH, P2WPKH, P2TR, etc.).
+	AddrType AddressType
+
+	// Dirty indicates the address is used in a transaction that hasn't
+	// been broadcast/confirmed yet.
+	Dirty bool
```
yyforyongyu: nit: I think it's called `Used` previously, why this rename tho

wallet/internal/db/data_types.go
```diff
@@ -570,36 +574,85 @@ type RenameAccountParams struct {
...
+	// address. Reuses the AccountOrigin enum.
+	Origin AccountOrigin
+
+	// Branch is the BIP44 branch number (0=external, 1=internal/change).
+	// Zero value for imported addresses.
+	Branch uint32
+
+	// Index is the BIP44 index within the branch. Zero value for imported
+	// addresses.
+	Index uint32
```
yyforyongyu, line 589: Q: why this flatten approach other than, say, create a `DerivationPath` to wrap the indices?

wallet/internal/db/data_types.go
```diff
@@ -570,36 +574,85 @@ type RenameAccountParams struct {
...
-	Compressed bool
+	// EncryptedPubKey is the encrypted public key. Zero value for derived
+	// addresses.
+	EncryptedPubKey []byte
 
-	// Used indicates whether the address has been used in a transaction.
-	Used bool
+	// WitnessVersion is the SegWit version (0 for P2WPKH/P2WSH, 1 for P2TR).
+	// nil for non-witness address types or derived addresses.
+	WitnessVersion *int16
```
yyforyongyu: This seems redundant? the version info has been captured in the `AddressType` - whenever a new version is created, it must have a new address type?

wallet/internal/db/data_types.go
```diff
@@ -570,36 +574,85 @@ type RenameAccountParams struct {
...
-	// this address.
+	// IsWatchOnly indicates whether the wallet has the private key for this
+	// address. Convenience field.
 	IsWatchOnly bool
+}
 
-	// AddrType is the type of the address (P2PKH, P2SH, etc.).
-	AddrType AddressType
+// AddressSecret contains sensitive encrypted material for an address.
+type AddressSecret struct {
```
yyforyongyu, line 604: 👍

wallet/internal/db/data_types.go
```diff
@@ -570,36 +574,85 @@ type RenameAccountParams struct {
...
 
-	// Script is the script associated with the address, if any.
-	Script []byte
+	// EncryptedScript is the encrypted redeem or witness script for
+	// P2SH/P2WSH addresses. For Taproot, this is the TLV-encoded Tapscript.
+	EncryptedScript []byte
+
+	// IsSecretScript indicates which key encrypted EncryptedScript:
+	// true = the script crypto key, false = the public crypto key.
+	IsSecretScript bool
```
yyforyongyu: ok this is indeed a rabbit hole😅

The current dual-encryption mode led me to dig into this topic deeper, and I think we should simplify it by removing the public phrase, which seems to be the modern wallet approach. More analysis [here](https://gist.github.com/yyforyongyu/edfeee0f84cf79851735bcc3e740a871).

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -0,0 +1,152 @@
+-- Migration note: Intentionally NOT idempotent (no "IF NOT EXISTS").
+-- This ensures migration tracking stays accurate and fails loudly if run twice.
+
+-- Address sync status lookup table - provides standardized descriptions for
+-- address synchronization states. This is a reference table that maps the
+-- AddressSyncStatus enum values used in Go code to their human-readable names.
+CREATE TABLE address_sync_statuses (
```
yyforyongyu: Think this can be dropped 🛴

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    -- branch). NULL for imported addresses.
+    address_index BIGINT,
+
+    -- Encrypted public key for imported addresses. NULL for HD-derived
+    -- addresses since their public keys are derived from the account key.
+    encrypted_pub_key BYTEA,
+
+    -- Witness version for SegWit addresses (0 for P2WPKH/P2WSH, 1 for P2TR).
+    -- NULL for non-witness address types.
+    witness_version SMALLINT,
```
yyforyongyu: can be removed

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -0,0 +1,152 @@
...
+    created_at TIMESTAMP NOT NULL DEFAULT current_timestamp,
+
+    -- Foreign key constraint to accounts. Using ON DELETE RESTRICT to ensure
+    -- that the account cannot be deleted if addresses still exist.
+    FOREIGN KEY (account_id) REFERENCES accounts (id) ON DELETE RESTRICT,
+
+    -- Foreign key constraint to blocks. Using ON DELETE RESTRICT to prevent
+    -- block deletion if addresses reference that height.
+    FOREIGN KEY (first_seen_height) REFERENCES blocks (block_height)
+    ON DELETE RESTRICT,
```
yyforyongyu: what about reorgs?

yyforyongyu (changes requested):
Cool - I left two critical design decisions we need to make, think we need to sort them out first.

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,420 @@
...
+	return nil, fmt.Errorf("get address secret: %w", err)
+}
+
+// validate validates the required fields for creating an imported address.
+// Returns sentinel errors on failure.
+func (p NewImportedAddressParams) validate() error {
+	if len(p.EncryptedScriptPubKey) == 0 {
+		return ErrMissingEncryptedScriptPubKey
+	}
```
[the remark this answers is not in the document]
GustavoStingelin: should not require pub key bc user can import a script address without a pub key.

wallet/internal/db/data_types.go
GustavoStingelin: good, removed.

wallet/internal/db/data_types.go
GustavoStingelin: good, removed.

wallet/internal/db/data_types.go
GustavoStingelin: Cool, I agree this will be simpler to manage. We could suggest that node runners use LUKS disk encryption if they want this extra layer of security. I removed encryption from the address code for public data. We will need a follow up PR to remove it from wallet, keyscope, account, and related areas.

Do we need an additional ack from anyone else to move forward with this decision?

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: dropped.

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: removed

wallet/internal/db/data_types.go
GustavoStingelin: Yes, the issue shows up once we have this boolean alongside the tx or credits tables. At that point we end up with two sources claiming whether an address is used. Trying to remove this ambiguity, I renamed it to `Dirty` to make it distinct from `Used`.

Dirty means the address has appeared in some transaction output or was seen in the mempool, but it may not be confirmed in a block yet. That is different from what we get from tx or credit relations, or from the `first_seen_block` column.

Does this distinction make sense, or do you think we should keep the `Used` approach instead?

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: During reorgs we would need to clear this field if the target height is earlier than this one. That said, given the new rescan architecture, do you think this field still makes sense? If not, we can drop it.

wallet/internal/db/data_types.go
GustavoStingelin: Because this `AddressInfo` is a bit weird in nature, we will end up with missing fields depending on the `Origin`, which makes it hard to represent cleanly in Go. I chose to flatten it to keep the struct simple and easier to understand, instead of introducing interfaces with different structs for derived and imported data.

That said, at a higher layer we may still want some kind of abstraction. Also, for me, at this point, simply splitting the data into another struct does not seem to improve the clarity.

GustavoStingelin:
related to #1166

wallet/internal/db/data_types.go
yyforyongyu: For the previous `Used` flag, i think we set it when the address is seen in a tx that's found in the mempool or block, tho we may want to move it earlier like when we decide to publish the tx.

Here, based on the inline docs, it seems to me the `Dirty` flag is redundant - when will we use it?

This, however, is indeed a very sharp insight - there are two sources of truth in the old design where we rely on the `Used` flag, but now we should only rely on the tx table to decide whether an address is indeed used or not.

I think the boils down to how we balance privacy and address gap, will create an issue. Meanwhile, I think you are right - we should stick to a single source of truth to decide whether an address is used or not.

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -0,0 +1,109 @@
...
+    -- how the address is encoded and how funds can be spent.
+    type_id SMALLINT NOT NULL,
+
+    -- Block height when this address was first seen on-chain. NULL if the
+    -- address has never appeared in the blockchain. References the blocks table.
+    first_seen_height INTEGER,
+
+    -- This column is used to mark addresses that have been used in a
+    -- transaction, but the transaction hasn't been broadcast or confirmed yet.
+    dirty BOOLEAN NOT NULL,
```
yyforyongyu: If we remove this, we could build sth like,
```sql
-- Check if address ID 123 is used
SELECT EXISTS (
    -- Check for received funds (UTXOs created for this address)
    SELECT 1 FROM utxos WHERE address_id = 123
    UNION ALL
    -- Check for spent funds (Inputs spending UTXOs of this address)
    -- Note: In the ADR 0006 schema, spent inputs are tracked by updating the 
    -- 'spent_by_tx_id' on the utxos table itself.
    -- So checking 'utxos' covers both received (row exists) and spent (row exists).
    -- If we had a separate 'inputs' table, we'd check that too.
    SELECT 1 FROM utxos WHERE address_id = 123 AND spent_by_tx_id IS NOT NULL
);
```
Once the tx tables are built.

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -0,0 +1,109 @@
...
+    -- addresses this must be stored.
+    script_pub_key BYTEA,
+
+    -- Reference to the address type (e.g., P2PKH, P2WPKH, P2TR). Determines
+    -- how the address is encoded and how funds can be spent.
+    type_id SMALLINT NOT NULL,
+
+    -- Block height when this address was first seen on-chain. NULL if the
+    -- address has never appeared in the blockchain. References the blocks table.
+    first_seen_height INTEGER,
```
yyforyongyu: It seems this field is not exported in the go struct, also that we can now query the tx table to get it,
```sql
SELECT MIN(t.block_height)
FROM utxos u
JOIN transactions t ON u.tx_id = t.id
WHERE u.address_id = $1
  AND t.block_height IS NOT NULL;
```

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
yyforyongyu: yeah think we can drop it

wallet/internal/db/addresses_common.go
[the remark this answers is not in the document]
yyforyongyu: But why would having a redeem script or witness script makes it watch-only? We can import an address and its witness script, but we don't have the private key to generate the witness, so this should also be considered as a watch-only? I think watch-only means the wallet cannot spend.

wallet/internal/db/queries/postgres/addresses.sql
```diff
@@ -0,0 +1,169 @@
+-- name: GetAddressByID :one
```
yyforyongyu: Looks like this query is not used?

wallet/internal/db/queries/postgres/accounts.sql
```diff
@@ -85,10 +85,15 @@ SELECT
     a.is_watch_only,
     a.created_at,
     ks.purpose,
-    ks.coin_type
+    ks.coin_type,
+    count(*) FILTER (WHERE addr.address_branch = 0) AS external_key_count,
```
yyforyongyu: I think when the address table grows, this query will be O(N), which can be slow for lots of addresses. Think we need to fix it, and found two options,

Option A: Application Logic (Simpler for Go devs)
Update the counter atomically when inserting the address.
```sql
// In NewDerivedAddress transaction:
qtx.CreateDerivedAddress(...)
qtx.IncrementAccountKeyCount(...) // UPDATE accounts SET ... +1
```

Option B: Database Triggers (Robust)
Use a trigger to enforce consistency automatically.
Postgres:
```sql
CREATE OR REPLACE FUNCTION update_account_key_count() RETURNS TRIGGER AS $$
BEGIN
    UPDATE accounts SET 
        external_key_count = external_key_count + CASE WHEN NEW.address_branch = 0 THEN 1 ELSE 0 END,
        internal_key_count = internal_key_count + CASE WHEN NEW.address_branch = 1 THEN 1 ELSE 0 END,
        imported_key_count = imported_key_count + CASE WHEN NEW.address_branch IS NULL THEN 1 ELSE 0 END
    WHERE id = NEW.account_id;
    RETURN NEW;
END;
$$ LANGUAGE plpgsql;
CREATE TRIGGER trigger_update_key_count AFTER INSERT ON addresses FOR EACH ROW EXECUTE FUNCTION update_account_key_count();
```

SQLite:
```sql
CREATE TRIGGER update_account_key_count AFTER INSERT ON addresses
BEGIN
    UPDATE accounts SET 
        external_key_count = external_key_count + CASE WHEN NEW.address_branch = 0 THEN 1 ELSE 0 END,
        -- ... (same logic)
    WHERE id = NEW.account_id;
END;
```

Option B seems more sql native, wdyt?

wallet/internal/db/addresses_pg.go
```diff
@@ -0,0 +1,318 @@
...
+				WalletID:     int64(q.WalletID),
+			}, pgAddressRowToInfo,
+		)
+	}
+
+	return getAddressByQuery(ctx, query, getByScript)
+}
+
+// ListAddresses returns all addresses for a given account.
+func (w *PostgresWalletDB) ListAddresses(ctx context.Context,
```
yyforyongyu, line 34: I think we should add pagination.

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,394 @@
...
+}
+
+// getAddressFunc defines a function signature for retrieving a single address.
+type getAddressFunc func(context.Context, GetAddressQuery) (*AddressInfo, error)
+
+// getAddressByQuery validates the query and executes the script-based lookup.
+func getAddressByQuery(ctx context.Context, query GetAddressQuery,
+	getter getAddressFunc) (*AddressInfo, error) {
+
+	if len(query.ScriptPubKey) == 0 {
```
yyforyongyu: This will always be null for derived addresses?

yyforyongyu (changes requested):
Found a few issues, think we are very close!

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -0,0 +1,109 @@
...
+    -- DB ID of the address, primary key.
+    id BIGSERIAL PRIMARY KEY,
+
+    -- Reference to the account this address belongs to.
+    account_id BIGINT NOT NULL,
+
+    -- Script pubkey that locks funds on-chain (stored in plaintext). For
+    -- derived addresses this is NULL (can be re-derived). For imported
+    -- addresses this must be stored.
+    script_pub_key BYTEA,
```
yyforyongyu: Actually i think we need to store this for the address, so the utxo can reference the address to get the pk script.

yyforyongyu: But I guess it can always be derived - as long as it guarantees that it can be derived.

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: Hmm, so the table will also include cases where a transaction was constructed but not broadcast yet?

wallet/internal/db/queries/postgres/accounts.sql
GustavoStingelin: What do you think the option C:

This one, first brings a reflection, "why we need the address counting in the account query?" maybe looks like just a statistical value? So, maybe we could make the account query simplier, bringing no count data, and create another query that will be responsible to get this statiscal counting when needed, making the code simpler while not looping the database on every request.


Also, if this is an issue for `count`, it will also affect `max`, which we use to derive a new address. In this case, for consistency, I think we should move the logic for `counting`, `next_address`, and `next_account` to cached fields. This would also make it easier to ensure that derived addresses have the script_pub_key available at row insertion time, enforced as a not null field.

wdyt?

wallet/internal/db/addresses_pg.go
GustavoStingelin: Totally agree, but pagination is not easy by default. I think we should use cursor based pagination and maybe return a page struct to embed metadata. It would also be nice to have a way to expose this as a sequence for callers that do not want to deal with pagination details.

Given that, what do you think about keeping this as is for now and opening an issue to introduce pagination across all similar cases? Then we can iterate later with focus only on this topic and converge on a solid solution.

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: Yes, I also think it makes sense to store this. We will probably need to pass a callback to the insert function, so `db.Store` can derive this information, or do it in separate transactions and accept some risk of inconsistent state if a transaction fails, or expose transaction rollback to layers above the database.

I will work on this once we define the counting and max approach.

wallet/internal/db/addresses_common.go
GustavoStingelin: replied in the next review

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
yyforyongyu: Correct - so to break down case by case,
1. user calls `NewAddress`, always gives a new addr as guaranteed by the index
2. user uses this addr to construct a tx and broadcast, `utxos` will have this record
3. users use this addr to construct a tx without broadcasting, `utxos` won't have this record

The third case is challenging - the user could make lots of txns without broadcasting, then we will have lots of unused addresses, making the gap large.

I think for now we should stick to the existing behavior - atm we only mark the addr as used when it's found in the mempool or blockchain, which is what the query on the utxos mimics - we don't have the tables yet, so we can add it as a TODO there. 

To properly fix it, turn to #1167

wallet/internal/db/queries/postgres/accounts.sql
yyforyongyu: >  I think we should move the logic for counting, next_address, and next_account to cached fields. 

What do you mean by saying cached fields? You mean new columns or in-memory tracking?

If new columns, yeah, we could use a dedicated counter column on accounts allows us to:
1.  `UPDATE accounts SET next_index = next_index + 1 RETURNING next_index`
2.  Derive keys.
3.  `INSERT INTO addresses ...`

In that case, think we can just return the count from the account query since that's cheap. If you are talking about building separate, dedicated queries, like get account address count, without adding the columns, then yeah it also makes sense. Tho at this stage I'd perfer the new columns approach as it looks simpler.

wallet/internal/db/addresses_pg.go
yyforyongyu: Def - we can defer this to a new PR, as long as it's shipped eventually!

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
yyforyongyu: If we were to save it, I don't think it should be derived again - once we have the `pub_key` derived, we will create the `script_pub_key` based on its address type and save it there, most likely be done when calling `NewAddress`.

In theory, for derived addresses, we don't need to store anything as long as we have the `type_id`, we can always derive on the fly to get the pub key, then use it to construct the script. For imported however, we need to store the `pub_key`. This ofc has the downside of CPU computations.

My overall assessment is, storage is cheap, and we should store the script - for derived keys, no need to store `pub_key`, as that's not used frequently; for imported keys, we should construct the script based on the imported public key and store it.

wallet/internal/db/queries/postgres/accounts.sql
GustavoStingelin: Sorry for the confusion around the term cached. What I meant is use the counter column approach.

wallet/internal/db/queries/postgres/addresses.sql
GustavoStingelin: Actually, we are using this in tests to make sure the update query for an address saved the info correctly.

wallet/internal/db/addresses_common.go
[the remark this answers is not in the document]
GustavoStingelin: My understanding is that if we have the redeem script required to satisfy a P2SH output, then we are able to spend it, so it should not be considered watch only.

In the older kvdb implementation, we did not store public or private keys for P2SH outputs, for example.

Based on that, it seems that an output should be considered watch only when we do not have any private key and also do not have a spending script available.

Does this match your understanding, or am I missing a relevant case here?

wallet/internal/db/queries/postgres/accounts.sql
yyforyongyu: yeah no worries - i think the counter approach makes sense - easy and straightforward.

wallet/internal/db/data_types.go
GustavoStingelin: So, removed it to use the table that will store the outputs.

wallet/internal/db/data_types.go
GustavoStingelin: removed the encryption

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: dropped

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: just dropped the dirty/used flag to use the table that will store the outputs later

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: dropped

wallet/internal/db/queries/postgres/addresses.sql
GustavoStingelin: As we changed some approaches, this became unused, so I removed it now

wallet/internal/db/queries/postgres/accounts.sql
GustavoStingelin: Added the approach for derived address, do u think it will also be relevant for the imported?

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: Changed it to not null, and the create method now receives a callback function to derive this info.

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -0,0 +1,94 @@
...
+    encrypted_priv_key BYTEA,
+
+    -- Encrypted script for script-based addresses (P2SH, P2WSH). Contains the
+    -- redeem script or witness script needed to spend the output.
+    encrypted_script BYTEA,
+
+    -- Foreign key constraint to addresses. Using ON DELETE RESTRICT to ensure
+    -- that the address cannot be deleted if secrets still exist.
+    FOREIGN KEY (address_id) REFERENCES addresses (id) ON DELETE RESTRICT
+);
```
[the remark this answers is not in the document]
GustavoStingelin: Using a primary key here would implicitly introduce auto increment semantics through the foreign key relationship, which is unnecessary and intentionally avoided.

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
```diff
@@ -0,0 +1,95 @@
...
+    encrypted_priv_key BLOB,
+
+    -- Encrypted script for script-based addresses (P2SH, P2WSH). Contains the
+    -- redeem script or witness script needed to spend the output.
+    encrypted_script BLOB,
+
+    -- Foreign key constraint to addresses. Using ON DELETE RESTRICT to ensure
+    -- that the address cannot be deleted if secrets still exist.
+    FOREIGN KEY (address_id) REFERENCES addresses (id) ON DELETE RESTRICT
+);
```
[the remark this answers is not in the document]
GustavoStingelin: Using a primary key here would implicitly introduce auto increment semantics through the foreign key relationship, which is unnecessary and intentionally avoided.

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,619 @@
...
+// handling.
+func convertAddressPath(branch, index sql.NullInt64) (uint32, uint32, error) {
+	addrBranch, err := int64ToUint32(branch.Int64)
+	if err != nil {
+		return 0, 0, fmt.Errorf("address branch: %w", err)
+	}
+
+	addrIndex, err := int64ToUint32(index.Int64)
+	if err != nil {
+		return 0, 0, fmt.Errorf("address index: %w", err)
```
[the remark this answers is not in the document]
GustavoStingelin: added a check

wallet/internal/db/interface.go
```diff
@@ -130,53 +129,61 @@ type AccountStore interface {
 	RenameAccount(ctx context.Context, params RenameAccountParams) error
 }
 
+// AddressDerivationFunc is called by the database layer after allocating an
+// address index to derive the actual address data (script_pub_key). As the
+// database should not know about how to derive an address, we pass this as a
+// callback.
+type AddressDerivationFunc func(ctx context.Context, accountID uint32,
```
yyforyongyu, line 153: cool - also like that this is stateless so it's safe for callback.

wallet/internal/db/data_types.go
```diff
@@ -666,8 +680,10 @@ type GetAddressQuery struct {
 	// databases (signed 64-bit integers).
 	WalletID uint32
 
-	// Address is the address to query.
-	Address btcutil.Address
+	// ScriptPubKey is the script pubkey. If provided, the query will be
+	// performed using this value. Only applicable to imported addresses,
+	// as derived addresses have NULL script pubkeys.
```
yyforyongyu: This is no longer correct.

wallet/internal/db/data_types.go
```diff
@@ -570,36 +557,69 @@ type RenameAccountParams struct {
...
+	// Branch is the BIP44 branch number (0=external, 1=internal/change).
+	// Zero value for imported addresses.
+	Branch uint32
+
+	// Index is the BIP44 index within the branch. Zero value for imported
+	// addresses.
+	Index uint32
+
+	// ScriptPubKey is the script pubkey (plaintext). Zero value for
+	// derived addresses.
```
yyforyongyu: This is no longer true

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
```diff
@@ -0,0 +1,95 @@
...
+WHERE
+    address_branch IS NOT NULL
+    AND address_index IS NOT NULL;
+
+-- Unique partial index to prevent duplicate script_pub_key within the same
+-- account. Only enforced when script_pub_key is non-NULL (imported addresses).
+-- Derived addresses are excluded from this constraint.
+CREATE UNIQUE INDEX uidx_addresses_account_script_pub_key
+ON addresses (account_id, script_pub_key)
+WHERE script_pub_key IS NOT NULL;
```
yyforyongyu: `script_pub_key` is always not null now so can drop the `WHERE`

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
```diff
@@ -0,0 +1,95 @@
...
+
+-- Index on script_pub_key for efficient lookups by script pubkey.
+-- Used by GetAddressByScriptPubKey which joins through accounts to key_scopes.
+CREATE INDEX idx_addresses_script_pub_key
+ON addresses (script_pub_key)
+WHERE script_pub_key IS NOT NULL;
```
yyforyongyu: `script_pub_key` is always not null now.

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
```diff
@@ -0,0 +1,95 @@
...
+    -- Script pubkey that locks funds on-chain (stored in plaintext).
+    script_pub_key BLOB NOT NULL,
+
+    -- Reference to the address type (e.g., P2PKH, P2WPKH, P2TR). Determines
+    -- how the address is encoded and how funds can be spent.
+    type_id INTEGER NOT NULL,
+
+    -- Branch number in BIP44 derivation path (typically 0 for external, 1 for
+    -- internal/change). NULL for imported addresses.
+    address_branch INTEGER,
```
yyforyongyu, line 27: Add a constraint such that `address_branch` is 0 or 1, that `branch` and `index` are both `NULL` or both non-NULL?

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
```diff
@@ -0,0 +1,95 @@
...
+CREATE INDEX idx_addresses_script_pub_key
+ON addresses (script_pub_key)
+WHERE script_pub_key IS NOT NULL;
+
+-- Address Secrets table stores sensitive encrypted material needed to spend
+-- from an address. This table has a one-to-one relationship with addresses.
+-- Watch-only addresses may have no row in this table.
+CREATE TABLE address_secrets (
+    -- Reference to the address these secrets belong to.
+    address_id INTEGER NOT NULL,
```
yyforyongyu: We can do `address_id BIGINT PRIMARY KEY` and drop the separate unique index?

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -0,0 +1,94 @@
...
+    -- how the address is encoded and how funds can be spent.
+    type_id SMALLINT NOT NULL,
+
+    -- Branch number in BIP44 derivation path (typically 0 for external, 1 for
+    -- internal/change). NULL for imported addresses.
+    address_branch BIGINT,
+
+    -- Index number in BIP44 derivation path (sequential counter within each
+    -- branch). NULL for imported addresses.
+    address_index BIGINT,
```
yyforyongyu, line 33: nit: can use `SMALLINT`?

wallet/internal/db/queries/sqlite/addresses.sql
```diff
@@ -0,0 +1,94 @@
...
+    a.id AS address_id,
+    s.encrypted_priv_key,
+    s.encrypted_script
+FROM addresses AS a
+LEFT JOIN address_secrets AS s ON a.id = s.address_id
+WHERE a.id = ?;
+
+-- name: CreateDerivedAddress :one
+-- Creates a derived address with the given index and derived data.
+-- The index is allocated separately via GetAndIncrementNextAddressIndex.
```
yyforyongyu: looks like a stale comment, should be `GetAndIncrementNextExternalIndex / GetAndIncrementNextInternalIndex`?

wallet/internal/db/queries/sqlite/addresses.sql
```diff
@@ -0,0 +1,94 @@
...
+    type_id,
+    address_branch,
+    address_index,
+    pub_key
+) VALUES (
+    ?1, ?2, ?3, NULL, NULL, ?4
+)
+RETURNING id, created_at;
+
+-- name: ListAddressesByAccount :many
```
yyforyongyu, line 73: I think we should do an `ORDER BY` to make sure the returned results are always the same for two quries.

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
```diff
@@ -64,6 +64,12 @@ CREATE TABLE accounts (
     -- Timestamp when the account was created. Automatically set by the database.
     created_at TIMESTAMP NOT NULL DEFAULT current_timestamp,
 
+    -- Next index to use for external addresses (branch 0)
+    next_external_index BIGINT NOT NULL DEFAULT 0,
```
yyforyongyu, line 68: Would be nice to have `CHECK (next_external_index >= 0)` to prevent corruption.

wallet/internal/db/queries/sqlite/accounts.sql
```diff
@@ -245,3 +245,33 @@ INSERT INTO accounts (
...
+-- name: GetAndIncrementNextInternalIndex :one
+-- Atomically gets the next internal/change address index and increments the counter.
+-- Returns the current index value (before incrementing) for the address derivation.
+UPDATE accounts
+SET next_internal_index = next_internal_index + 1
+WHERE id = ?
+RETURNING next_internal_index - 1 AS address_index;
+
+-- name: UpdateAccountNextExternalIndex :exec
+-- Updates the next_external_index counter for an account. Used in tests
```
yyforyongyu: IF it's just for testing, i think we can use `fixtures_sqlite_test.go` and exec raw sql directly, adding this to production code is less clean.

wallet/internal/db/safecasting.go
```diff
@@ -43,6 +43,17 @@ func int64ToInt32(v int64) (int32, error) {
 	return int32(v), nil
 }
 
+// int64ToInt16 safely casts an int64 to an int16, returning an error
+// if the value is out of range.
+func int64ToInt16(v int64) (int16, error) {
```
yyforyongyu: looks like deadcode?

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,633 @@
...
+	return infos, nil
+}
+
+// derivedAddressAdapters groups the functions needed to create a
+// derived address across different database backends.
+type derivedAddressAdapters[QTX any, AccountRow any, AccountParams any,
+	AddrRow any] struct {
+
+	// getAccount retrieves an account by the provided parameters.
+	getAccount func(context.Context, AccountParams) (AccountRow, error)
```
yyforyongyu, line 400: this should accept an `QTX` too?

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,633 @@
...
+	}
+
+	rowIDVal := rowID(row)
+
+	id, convertedAcctID, err := convertAddressIDs(rowIDVal, accountID)
+	if err != nil {
+		return nil, err
+	}
+
+	return &AddressInfo{
```
yyforyongyu, line 506: `scriptPubKey` is missing here?

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,633 @@
...
+		return 0, 0, fmt.Errorf("address origin: %w", err)
+	}
+
+	return addrType, origin, nil
+}
+
+// convertAddressPath converts BIP44 branch and index values with error
+// handling.
+func convertAddressPath(branch, index sql.NullInt64) (uint32, uint32, error) {
+	addrBranch, err := int64ToUint32(branch.Int64)
```
yyforyongyu, line 291: we should check `Valid` on this `Int64`? for imported ones, we don't have them set.

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,633 @@
...
+	if err != nil {
+		return 0, 0, fmt.Errorf("address branch: %w", err)
+	}
+
+	addrIndex, err := int64ToUint32(index.Int64)
```
yyforyongyu, line 296: same here re`Valid`

wallet/internal/db/addresses_pg.go
```diff
@@ -0,0 +1,304 @@
...
+func importedAddressRowCreatedAtPg(
+	row sqlcpg.CreateImportedAddressRow) time.Time {
+
+	return row.CreatedAt
+}
+
+// pgAddressSecretRowToSecret converts a sqlc GetAddressSecretRow row to the
+// db.AddressSecret type used by the wallet, handling null value conversions.
+// Returns ErrSecretNotFound if the address exists but has no secret.
+func pgAddressSecretRowToSecret(row sqlcpg.GetAddressSecretRow) (*AddressSecret,
```
yyforyongyu: nit: can use the same generic union pattern used by `pgAddressRowToInfo/sqliteAddressRowToInfo`.

wallet/internal/db/queries/postgres/accounts.sql
yyforyongyu: yeah i think so

wallet/internal/db/queries/postgres/accounts.sql
```diff
@@ -85,10 +85,15 @@ SELECT
...
     ks.purpose,
-    ks.coin_type
+    ks.coin_type,
+    a.next_external_index AS external_key_count,
+    a.next_internal_index AS internal_key_count,
+    count(*) FILTER (WHERE addr.address_branch IS NULL AND addr.id IS NOT NULL) AS imported_key_count
 FROM accounts AS a
 INNER JOIN key_scopes AS ks ON a.scope_id = ks.id
-WHERE a.scope_id = $1 AND a.account_name = $2;
+LEFT JOIN addresses AS addr ON a.id = addr.account_id
```
yyforyongyu: we should drop all these left join, otherwise we still suffer the O(N) issue - think we can just add a column `imported_key_count` there, then an atomic increment query called during `NewImportedAddress`:
```sql
  -- name: IncrementImportedKeyCount :exec
  UPDATE accounts SET imported_key_count = imported_key_count + 1 WHERE id = $1;
```

wallet/internal/db/itest/address_store_test.go
```diff
@@ -0,0 +1,1120 @@
...
+	// uint32.
+	_, err := store.NewDerivedAddress(
+		t.Context(), db.NewDerivedAddressParams{
+			WalletID:    walletID,
+			Scope:       db.KeyScopeBIP0084,
+			AccountName: "max-acct",
+			Change:      false,
+		}, mockDeriveFunc(),
+	)
+	require.Error(t, err)
```
yyforyongyu: should be `require.ErrorIs(t, err, db.ErrMaxAddressIndexReached)`?

wallet/internal/db/itest/address_store_test.go
```diff
@@ -0,0 +1,1120 @@
...
+			)
+
+			// Verify AddressInfo fields.
+			require.NotZero(t, info.ID)
+			require.NotZero(t, info.AccountID)
+			require.Equal(t, db.DerivedAccount, info.Origin)
+			require.NotZero(t, info.CreatedAt)
+			require.Equal(t, tc.expectedBranch, info.Branch)
+			require.GreaterOrEqual(t, info.Index, uint32(0))
+			require.Nil(t, info.ScriptPubKey)
```
yyforyongyu: This field should not be nil

yyforyongyu (changes requested):
My suggestion is we address them in a new PR, meanwhile keep this PR open, and open that PR against this branch.

wallet/internal/db/queries/postgres/accounts.sql
GustavoStingelin: added via triggers

wallet/internal/db/data_types.go
GustavoStingelin: fixed

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
GustavoStingelin: fixed

wallet/internal/db/data_types.go
GustavoStingelin: fixed

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
GustavoStingelin: fixed

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
GustavoStingelin: added

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
GustavoStingelin: cool, changed, just realized that just `pk` does not create the sequence, but the `serial` yes.

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
GustavoStingelin: You are referring to the branch field, right?

Yes, this change saves 6 bytes per address row, but it introduces some extra complexity in the conversion logic. You can see the details in the commit.

The slightly awkward part is that we model branch as an integer, following BIP44, even though in practice we only use two values, zero and one. Conceptually, this could be represented as a boolean. That leaves us with a trade off.

We can keep a uint32, which is more expressive and aligned with the specification, and leaves room for future extensions. Or we can optimize for storage and simplicity by using a boolean, which is faster and smaller but less flexible if the model ever expands.

I opted to follow SMALLINT as a middle ground. It reduces storage, keeps the representation numeric, and makes a future ALTER TABLE trivial if we ever need to expand the domain.

wallet/internal/db/queries/sqlite/addresses.sql
GustavoStingelin: fixed

wallet/internal/db/queries/sqlite/addresses.sql
GustavoStingelin: Added by id to use the simpler index

wallet/internal/db/migrations/postgres/000005_accounts.up.sql
GustavoStingelin: added

wallet/internal/db/queries/sqlite/accounts.sql
GustavoStingelin: I have no preferred approach for now, changed to raw approach

wallet/internal/db/safecasting.go
GustavoStingelin: yes, removed

wallet/internal/db/addresses_common.go
GustavoStingelin: It is already using the qtx, but in this case it is passed through `pgGetAccountFromKey`. Because of that, these callback functions have become a bit messy to manage without duplication.

The root issue is that a single create function is doing too many things at once. There is clear room for improvement here.

A possible next step is to introduce converter functions near the sqlc package. These functions would be responsible only for conversions between sqlc types and db types. With that separation, this layer could focus purely on the actual business logic, reducing complexity and making the code easier to reason about.

wallet/internal/db/addresses_common.go
GustavoStingelin: Good catch, fixed

wallet/internal/db/addresses_common.go
GustavoStingelin: This validation already exists, but I moved it into the convert function to make it more explicit.

wallet/internal/db/addresses_common.go
GustavoStingelin: same

wallet/internal/db/addresses_pg.go
GustavoStingelin: cool, added

wallet/internal/db/queries/postgres/accounts.sql
GustavoStingelin: just removed that join because we now have the count column

wallet/internal/db/itest/address_store_test.go
GustavoStingelin: added

wallet/internal/db/itest/address_store_test.go
GustavoStingelin: fixed

GustavoStingelin:
@yyforyongyu, I addressed the comments and added new commits starting at 61b269c to make the review easier. I can also squash or absorb them later to keep the history clean if needed.

wallet/internal/db/addresses_common.go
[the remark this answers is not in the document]
yyforyongyu: hmm forgot to reply here, but should we care if the script is there? i mean, even if it's watch only, if the user imports the witness, it's still considered watchonly?

wallet/internal/db/queries/postgres/addresses.sql
```diff
@@ -0,0 +1,96 @@
...
+-- - Address exists with secret: returns full row
+-- - Address exists without secret (watch-only/derived): returns row with NULL secret fields
+-- - Address does not exist: returns no rows (sql.ErrNoRows)
+SELECT
+    a.id AS address_id,
+    s.encrypted_priv_key,
+    s.encrypted_script
+FROM addresses AS a
+LEFT JOIN address_secrets AS s ON a.id = s.address_id
+WHERE a.id = $1;
```
yyforyongyu, line 43: `GetAddressSecret` scopes only by `address_id` — since these are sequential surrogates, a caller that can guess an ID could read secrets from a different wallet. The other address methods already enforce `wallet_id` via JOINs through `accounts → key_scopes`. Could we add the same filter here?

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,677 @@
...
+			"address index: %w", err)
+	}
+
+	acctID, err := int64ToUint32(accountID)
+	if err != nil {
+		return 0, 0, 0, nil, fmt.Errorf(
+			"account ID: %w", err)
+	}
+
+	derivedData, err := deriveFn(ctx, acctID, branch, index)
```
yyforyongyu, line 571: A nil `deriveFn` panics here, and a nil `derivedData` return (without error) panics at the dereference on line 577. Since this is a long-lived wallet process, guarding both with a typed error would be safer than an unrecoverable panic.

wallet/internal/db/addresses_common.go
```diff
@@ -0,0 +1,677 @@
...
+// Returns sentinel errors on failure.
+func (p NewImportedAddressParams) validate() error {
+	if len(p.ScriptPubKey) == 0 {
+		return ErrMissingScriptPubKey
+	}
+
+	return nil
+}
+
+// isWatchOnly returns true if the params include neither a private key nor
```
yyforyongyu, line 91: Coming back to the watch-only question — the interface doc (`interface.go:173`) and the `IsWatchOnly` field doc (`data_types.go`) both define watch-only as *no private key*. But this method returns `false` for script-only imports (no priv key, has script), and `addressRowToInfo` at line 326 mirrors that. Think we should align the code to the documented contract — if there's no priv key, it's watch-only, regardless of whether a script is present.

nit: also, `return noPrivKey && noScript` would be equivalent and shorter here.

wallet/internal/db/queries/postgres/addresses.sql
```diff
@@ -0,0 +1,96 @@
...
+-- Not used for derived addresses (their keys are derived from account key).
+INSERT INTO address_secrets (
+    address_id,
+    encrypted_priv_key,
+    encrypted_script
+) VALUES (
+    $1, $2, $3
+);
+
+-- name: GetAddressByScriptPubKey :one
```
yyforyongyu, line 12: This is `:one`, but uniqueness is per `(account_id, script_pub_key)` (migration line 71). If the same `script_pub_key` gets imported into two scope-specific imported accounts in one wallet (e.g. BIP44/imported and BIP84/imported — both valid since validation only checks non-empty script), `QueryRow` silently returns whichever row comes first. Worth either tightening uniqueness to wallet-level, or adding account/scope to the query contract?

wallet/internal/db/migrations/postgres/000006_addresses.down.sql
```diff
@@ -0,0 +1,4 @@
+-- Rollback note: Idempotent by design (using "IF EXISTS").
+-- Must succeed even if tables are already dropped or database is in unexpected state.
+DROP TABLE IF EXISTS address_secrets;
```
yyforyongyu, line 3: The two PL/pgSQL functions (`sync_account_imported_key_count_insert`, `sync_account_imported_key_count_delete`) survive a `DROP TABLE` — triggers are dropped with their table but standalone functions aren't. Adding `DROP FUNCTION IF EXISTS` for both before the table drops keeps rollback/reapply clean.

yyforyongyu (approved):
Found a few more items — mostly around wallet isolation and the watch-only semantics. Think we can address them in a follow up PR!

wallet/internal/db/migrations/postgres/000006_addresses.down.sql
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1210

wallet/internal/db/queries/postgres/addresses.sql
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1219

wallet/internal/db/queries/postgres/addresses.sql
GustavoStingelin: now, addressed in https://github.com/btcsuite/btcwallet/pull/1219 too

wallet/internal/db/addresses_common.go
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1220

wallet/internal/db/addresses_common.go
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1221

PR #1164, opened 2026-01-28
Add benchmark for `Controller`

yyforyongyu:
This PR adds benchmark tests to check the new syncer performance.

## Executive Summary
The New Controller API demonstrates significant performance advantages over the Legacy implementation across all scenarios. 
- **Full Block Mode** (`SyncMethodFullBlocks`) provides a **~2x speedup** for typical wallets by utilizing batched RPCs to amortize network latency.
- **Compact Filter Mode** (`SyncMethodCFilters`) provides a **~4x speedup** for sparse wallets by minimizing data transfer and avoiding block parsing for irrelevant data.

---

## 1. Syncing Empty Blocks (`BenchmarkSyncEmpty`)
This benchmark measures the fixed overhead of scanning blocks that contain no transactions relevant to the wallet.

| Block Count | Legacy (Bitcoind) | `SyncMethodFullBlocks` | `SyncMethodCFilters` | Improvement (Full vs Legacy) |
| :--- | :--- | :--- | :--- | :--- |
| 10 | 229 ms | 164 ms | 162 ms | ~28% |
| 100 | 294 ms | 160 ms | 167 ms | ~45% |
| 1,000 | 1,079 ms | 355 ms | 422 ms | **~3.0x Faster** |

**Insight**: Legacy scaling is linear with network latency because it fetches blocks sequentially. The New Controller uses batched RPCs to fetch the entire range in a single round-trip, drastically reducing wall-clock time.

---

## 2. Syncing Data Blocks (`BenchmarkSyncData`)
Measured over a fixed range of **1,000 blocks** with 100 transactions per block at varying hit densities.

| Hit Density | Legacy (Bitcoind) | `SyncMethodFullBlocks` | `SyncMethodCFilters` | Recommended Strategy |
| :--- | :--- | :--- | :--- | :--- |
| **0.1% (Sparse)** | 1,203 ms | 626 ms | **325 ms** | **`SyncMethodCFilters` (~4x gain)** |
| **1.0% (Periodic)** | 1,102 ms | 728 ms | **527 ms** | **`SyncMethodCFilters` (~2x gain)** |
| **10.0% (Dense)** | 1,302 ms | **729 ms** | 2,242 ms | **`SyncMethodFullBlocks` (~1.8x gain)** |

### Analysis:
1.  **`SyncMethodCFilters` (Winner for Sparse Data)**: At 0.1% density, the wallet avoids downloading 99.9% of the block data. The overhead of fetching filters is negligible compared to the bandwidth and CPU savings.
2.  **`SyncMethodFullBlocks` (Winner for Dense Data)**: As density increases (10% hits), the `SyncMethodCFilters` strategy of "Match -> Fetch Block" becomes a bottleneck due to serialized block requests. `SyncMethodFullBlocks` mode remains consistent (~730ms) because it always fetches and parses everything in one optimized batch.

---

### Legacy API
- **Bottleneck**: **Network I/O & System Calls**.
- **Observation**: Profiles are dominated by `syscall.syscall` and `pthread_cond_signal`. The CPU spends most of its time idle, waiting for sequential `GetBlock` responses from the node.

### New Controller (`SyncMethodFullBlocks`)
- **Bottleneck**: **JSON Parsing & Memory Management**.
- **Observation**: ~30% of CPU time is spent in `encoding/json.Unmarshal` and ~30% in `runtime.madvise`. 
- **Root Cause**: Fetching 1000 blocks in one JSON-RPC batch creates a massive JSON array payload. Parsing this payload into Go objects causes high memory churn and GC pressure.
- **Scaling**: Despite this CPU bottleneck, it is still 2x faster than Legacy because it eliminates network round-trip stalls.

### New Controller (`SyncMethodCFilters`)
- **Bottleneck**: **Memory Allocator**.
- **Observation**: Dominated by `runtime.madvise` (~80%). 
- **Root Cause**: Churning through many small filter and header objects. However, because the total data volume is low, it remains the fastest method for non-dense wallets.

---

## 4. Final Recommendations
- **Default to `SyncMethodCFilters`**: For the vast majority of user wallets (sparse history), Compact Filters provide the best UX and lowest bandwidth consumption.
- **Fallback to `SyncMethodFullBlocks`**: For heavy-duty indexing or known dense wallets, Full Block mode provides stable, high-throughput performance.
- **Deprecate Legacy Sync**: The Legacy sequential sync path is significantly outperformed by the New Controller when talking to standard `bitcoind` nodes.


GustavoStingelin (approved):
Cool, LGTM. The current approach makes sense, and we can revisit the memory model later to reduce allocations and speed things up further.

ACK.

PR #1165, opened 2026-01-28
Add design docs for transaction table schema

yyforyongyu:
Replaces,
- #1096 
- #1135

### DIFF
--- a/docs/developer/adr/0006-wtxmgr-sql-schema.md
+++ b/docs/developer/adr/0006-wtxmgr-sql-schema.md
@@ -1,0 +1,400 @@
+# ADR 0006: Wallet Transaction Manager SQL Schema
+
+## 1. Context
+
+As part of the migration from a Key-Value store to a relational SQL backend, the Wallet Transaction Manager (`wtxmgr`) requires a new schema design. The `wtxmgr` is responsible for tracking:
+1.  **Transactions:** Immutable blockchain data and user intent.
+2.  **UTXOs (Credits):** Outputs owned by the wallet (the primary operational unit).
+3.  **Spends (Debits):** Inputs that spend those outputs.
+4.  **Metadata:** User labels and transient locks (leases).
+
+For a detailed theoretical analysis of the data model, see [UTXO Data Model and Lifecycle](../utxo_data_model.md).
+
+Note on txid uniqueness: Bitcoin has historical txid-duplicate edge cases (primarily coinbase). However, since BIP30, duplicates cannot create concurrently-unspent outpoints. This schema therefore treats `tx_hash` as unique per wallet for balance/coin selection purposes.
+
+## 2. Decision
+
+We will adopt a **UTXO-Centered, Soft-Deletion Schema**.
+
+### 2.1 Core Principles
+1.  **UTXO-Centered Operations:** The schema is optimized for `Balance()` and `CoinSelection()` queries, which focus on the `utxos` table.
+2.  **Transaction-Centered Integrity:** Validity flows from Parent to Child. A UTXO is only valid if its parent Transaction is valid (`status='published'`) or is explicitly allowed for chaining (`status='pending'`).
+3.  **Immutable History (Soft Deletion):** We **NEVER** automatically `DELETE` rows.
+    *   Failed/RBF'd transactions are marked with a `status` field (e.g., `replaced`, `failed`).
+    *   They remain in the database for audit history but are excluded from balance queries.
+    *   Foreign Keys use `ON DELETE RESTRICT` for creation relationships to prevent accidental data loss.
+4.  **Wallet-Scoped Rows:** All `wtxmgr` tables are scoped by `wallet_id` to support multiple wallets sharing a single database without row-level conflicts.
+
+### 2.1.1 Wallet-Scoped vs Global Transactions
+Two designs were considered:
+
+*   **Wallet-scoped tables (chosen):**
+    *   Pros: Simple schema; no join tables; queries stay local to a wallet; avoids cross-wallet coordination.
+    *   Cons: Duplicates transaction rows across wallets that observe the same global tx.
+*   **Global transactions table (alternative):**
+    *   Pros: Storage efficiency; one canonical row per `tx_hash`.
+    *   Cons: Requires additional mapping tables (e.g., `wallet_transactions`) for per-wallet ownership/metadata; more complex queries and constraints; more careful concurrency semantics.
+
+This ADR chooses wallet-scoped tables for simplicity and implementability. A future migration to a global transactions table is possible but is considered a separate architectural decision.
+
+### 2.1.2 Explicit Status vs. Derived Status
+Two designs were considered for tracking transaction validity:
+
+*   **Explicit `status` column (chosen):**
+    *   The `status` field is a pre-computed materialization of the transaction's validity state, set atomically at write time when the invalidating event occurs (RBF, double-spend, reorg).
+    *   Pros: Balance and coin-selection queries filter on a single status predicate (`status IN ('published', 'pending')`) with no joins (and can be indexed if profiling justifies it); cascading invalidation is performed once at write time and never re-derived; audit states (`failed`, `replaced`, `orphaned`) are directly queryable.
+    *   Cons: Introduces a field that could theoretically drift from the underlying facts. This is partially mitigated by `CHECK` constraints (`check_confirmed_published`, `check_coinbase_confirmation_state`) and coinbase reorg triggers; transition correctness for `failed`/`replaced` remains a write-path responsibility validated by tests.
+
+*   **Derived status from other columns (alternative):**
+    *   At coarse granularity (pending vs active vs invalid), most states are derivable: `orphaned` = `is_coinbase AND block_height IS NULL`; `replaced` and `failed` (direct victim) = existence in `tx_replacements.replaced_tx_id`.
+    *   However, a full replacement requires preserving the distinct invalid states (`replaced`, `failed`, `orphaned`), which have different operational semantics: RBF (`replaced`) allows re-spending the same inputs with a new tx; `orphaned` coinbase can recover on reconfirmation; `failed` (double-spend) is permanent. Collapsing these into a single boolean loses information needed for recovery logic and user-facing audit.
+    *   A boolean decomposition (e.g., `is_broadcast` + `is_invalid`) also introduces problems: `is_broadcast` does not equal `published` — a tx can be valid/published because it was received from the network, not broadcast by this wallet; boolean pairs create ambiguous combinations (e.g., `is_broadcast=false, is_invalid=true`) requiring additional constraints.
+    *   A fully-derived alternative that preserves the same information would require at least three columns (`is_broadcast`/source marker, `invalidation_reason` enum, optional `invalidated_by_tx_id`) — strictly more schema surface than a single `status` column.
+    *   Cascading invalidation adds further complexity: downstream txs whose parent became invalid were never directly "replaced." Without a pre-computed status, every balance/coin-selection query would need a recursive CTE walking up the ancestor chain — O(depth) per tx on the hot path.
+
+This ADR chooses explicit status as the minimal representation that captures all lifecycle states without ambiguity. The `status` column is a pre-computed materialization set once at write time; database constraints/triggers enforce key invariants (confirmation and coinbase semantics), while write-path logic is responsible for setting `failed`/`replaced` transitions correctly.
+
+### 2.2 Consistency & Concurrency Model
+This design assumes standard SQL ACID guarantees.
+
+*   **Atomic updates:** Reorg disconnects, status transitions (RBF/failure/orphaning), and lease acquisition MUST be performed using explicit SQL transactions.
+*   **Leases are the concurrency primitive:** `utxo_leases` provides an application-level lock that prevents concurrent coin selection from choosing the same UTXO.
+*   **Isolation:** Wallet implementations SHOULD treat coin selection + lease acquisition as a single atomic unit. If multiple processes share one database, they must rely on database-enforced constraints (primary keys/uniques) and transactional semantics rather than best-effort in-memory coordination.
+
+Recommended operational defaults:
+*   Prefer running write paths under `SERIALIZABLE` and retry on serialization failures.
+*   If using weaker isolation (e.g., `READ COMMITTED`), rely on unique constraints and single-statement lease acquisition (no read-then-write without conflict handling).
+
+### 2.3 Reorg & RBF Strategy
+*   **Reorgs:** The `blocks` table represents the current best chain. If a block is disconnected, any referencing transaction will have `block_height` set to `NULL` via `ON DELETE SET NULL`.
+    *   **Effect (non-coinbase):** Regular transactions become unconfirmed (but remain `published`).
+    *   **Coinbase special case:** Coinbase transactions from the disconnected block are marked `orphaned` (they cannot exist outside the block that created them).
+    *   **Atomicity requirement:** The coinbase status update MUST occur atomically with the disconnect.
+        *   PostgreSQL evaluates `CHECK` constraints immediately, including updates performed by foreign-key actions such as `ON DELETE SET NULL`.
+        *   If you enforce the coinbase invariant at the database level, a simple `DELETE FROM blocks ...` will fail unless the coinbase row's `status` is updated to `orphaned` as part of the same statement.
+        *   Recommended: use a trigger to rewrite coinbase `status` during the `block_height -> NULL` update (see 3.6).
+    *   **Reconfirmation:** If an orphaned coinbase transaction re-enters the best chain, restoring it requires setting `block_height` and `status='published'` atomically.
+*   **RBF:** Handled by updating the `utxos.spent_by_tx_id` pointer to the new transaction and marking the old transaction as `replaced`.
+
+## 3. Reference Schema
+
+### 3.1 Wallet: `transactions`
+Stores the provenance of funds for a specific wallet. Acts as the source of truth for Validity (`status`) and Confirmation (`block_height`).
+
+```sql
+CREATE TABLE transactions (
+    wallet_id BIGINT NOT NULL REFERENCES wallets(id) ON DELETE RESTRICT,
+    id BIGSERIAL NOT NULL,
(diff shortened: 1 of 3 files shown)


docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+    
+    -- Creation (OutPoint):
+    -- ON DELETE RESTRICT: We NEVER delete confirmed transactions or their UTXOs.
+    -- To remove history, the user must explicitly Prune (manual op).
+    tx_id BIGINT NOT NULL REFERENCES transactions(id) ON DELETE RESTRICT,
+    output_index INTEGER NOT NULL CHECK (output_index >= 0),
+    
+    amount BIGINT NOT NULL CHECK (amount >= 0),
+    pk_script BYTEA NOT NULL,
+    address_id BIGINT NOT NULL REFERENCES addresses(id) ON DELETE RESTRICT,
```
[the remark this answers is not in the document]
yyforyongyu: the addresses table already exists, so we won't redo the work here.

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+CREATE TABLE transactions (
+    id BIGSERIAL PRIMARY KEY,
+    tx_hash BYTEA NOT NULL CHECK (length(tx_hash) = 32),
+    raw_tx BYTEA NOT NULL,
+    
+    -- Confirmation State:
+    -- NULL = Unconfirmed (Mempool)
+    -- INT  = Confirmed (Mined)
+    -- ON DELETE SET NULL: If block is reorged, tx becomes unconfirmed.
+    block_height INTEGER REFERENCES blocks(block_height) ON DELETE SET NULL,
```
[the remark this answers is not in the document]
yyforyongyu: again `blocks` table already exists

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+
+For a detailed theoretical analysis of the data model, see [UTXO Data Model and Lifecycle](../utxo_data_model.md).
+
+## 2. Decision
+
+We will adopt a **UTXO-Centered, Soft-Deletion Schema**.
+
+### 2.1 Core Principles
+1.  **UTXO-Centered Operations:** The schema is optimized for `Balance()` and `CoinSelection()` queries, which focus on the `utxos` table.
+2.  **Transaction-Centered Integrity:** Validity flows from Parent to Child. A UTXO is only valid if its parent Transaction is `published`.
```
GustavoStingelin: Q: `published` means `mined` or `broadcasted`?

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+3.  **Immutable History (Soft Deletion):** We **NEVER** automatically `DELETE` rows.
+    *   Failed/RBF'd transactions are marked with a `status` field (e.g., `replaced`, `failed`).
+    *   They remain in the database for audit history but are excluded from balance queries.
+    *   Foreign Keys use `ON DELETE RESTRICT` for creation relationships to prevent accidental data loss.
+
+### 2.2 Reorg & RBF Strategy
+*   **Reorgs:** Handled automatically via the `block_height` column. If a block is disconnected, `block_height` becomes `NULL`, and the transaction reverts to "Unconfirmed" (but remains `published`).
+*   **RBF:** Handled by updating the `utxos.spent_by_tx_id` pointer to the new transaction and marking the old transaction as `replaced`.
```
GustavoStingelin, line 77: Q: Does the presence of this field already imply that the item was replaced? If so, what extra information does the replaced status actually add?

Could we simplify this by deriving the status from existing fields instead of storing it explicitly? For example, using the pointer together with is_broadcasted and block_height seems enough to infer all possible states.

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
+# ADR 0006: Wallet Transaction Manager SQL Schema
+
+## 1. Context
+
+As part of the migration from a Key-Value store to a relational SQL backend, the Wallet Transaction Manager (`wtxmgr`) requires a new schema design. The `wtxmgr` is responsible for tracking:
+1.  **Transactions:** Immutable blockchain data and user intent.
+2.  **UTXOs (Credits):** Outputs owned by the wallet (the primary operational unit).
+3.  **Spends (Debits):** Inputs that spend those outputs.
+4.  **Metadata:** User labels and transient locks (leases).
```
GustavoStingelin, line 9: What about parent-child relations? like cluster mempool, P2A...

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+    -- INT  = Confirmed (Mined)
+    -- ON DELETE SET NULL: If block is reorged, tx becomes unconfirmed.
+    block_height INTEGER REFERENCES blocks(block_height) ON DELETE SET NULL,
+    
+    -- Validity State (Soft Deletion):
+    -- pending:   Created locally, not yet broadcast.
+    -- published: Active in mempool or blockchain (Valid).
+    -- replaced:  RBF'd by another transaction (Invalid).
+    -- failed:    Double-spent by a competitor (Invalid).
+    -- orphaned:  Coinbase tx that was reorged out (Invalid).
```
GustavoStingelin, line 104: This is a case where removing the status field reads cleaner, because it eliminates redundant information and duplicated state transitions. 

Here, we need to make sure we correctly revert the state for `orphaned` entries when `block_height` is set to `NULL` via `ON DELETE SET NULL`.

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+    CONSTRAINT valid_status CHECK (status IN ('pending', 'published', 'replaced', 'failed', 'orphaned')),
+    -- Invariant: If a transaction is confirmed (mined), it must be 'published'.
+    CONSTRAINT check_confirmed_published CHECK (block_height IS NULL OR status = 'published')
+);
+
+-- Optimization for Mempool lookups
+CREATE INDEX idx_transactions_unconfirmed ON transactions (block_height) WHERE block_height IS NULL;
+```
+
+### 3.2 Local: `utxos`
```
GustavoStingelin, line 152: Since this table has a `spent_by_tx_id` field, and we will often store outputs that are *already spent*, I think it makes more sense to rename it to something like `outputs`, `transaction_outputs`, or `txos`. Using `utxo` can be misleading, because the `U` implies we are only talking about *unspent outputs*, not both *spent and unspent*.

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+    output_index INTEGER NOT NULL CHECK (output_index >= 0),
+    
+    amount BIGINT NOT NULL CHECK (amount >= 0),
+    pk_script BYTEA NOT NULL,
+    address_id BIGINT NOT NULL REFERENCES addresses(id) ON DELETE RESTRICT,
+    
+    -- Spending (Input):
+    -- ON DELETE SET NULL: If the spending tx is manually pruned, the UTXO becomes unspent.
+    spent_by_tx_id BIGINT REFERENCES transactions(id) ON DELETE SET NULL,
+    spent_input_index INTEGER CHECK (spent_input_index >= 0),
```
GustavoStingelin: Q: What is the purpose of this field? Could it violate the transaction double-entry model? How does it behave with `ANYONECANPAY` or with cases where multiple inputs contribute to a single output?

docs/developer/adr/0006-wtxmgr-sql-schema.md
GustavoStingelin: Are we talking about RBF or spent tx?

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+
+-- Optimization for finding inputs (debits) of a transaction
+CREATE INDEX idx_utxos_spent_by ON utxos(spent_by_tx_id);
+```
+
+### 3.3 Audit: `tx_replacements`
+Tracks the history of RBF and Double-Spends.
+
+```sql
+CREATE TABLE tx_replacements (
```
GustavoStingelin, line 229: What about adding a `replaced_by_tx_id` field directly on the tx table? It would achieve the same goal, and it would also be pruned automatically when the referenced transaction row is deleted, assuming the current NOT NULL constraint + FK reference + DELETE CASCADE.

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+);
+```
+
+### 3.4 Local: `utxo_leases`
+Transient application locks.
+
+```sql
+CREATE TABLE utxo_leases (
+    utxo_id BIGINT PRIMARY KEY REFERENCES utxos(id) ON DELETE CASCADE,
+    lock_id BIGINT NOT NULL,
```
GustavoStingelin: What about renaming it to `external_lock_id` to emphasize that this ID comes from an external request, for example from LND?

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,158 @@
...
+    expiration TIMESTAMP NOT NULL
```
GustavoStingelin: `expiration` sounds like a duration. What about renaming it to `expires_at`?

GustavoStingelin (changes requested):
Left some comments.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: In this design `published` is the "valid" state — the tx is either in the mempool or mined, so it's eligible for balance/coin-selection queries. We leaned into that meaning in the status enum (`pending → published → failed/replaced/orphaned`) and by keeping `published` distinct from `block_height` (the view can still return `NULL` for unmined txs).

If the name is confusing, we can document it as "valid/mempool/mined" or consider a synonym such as `active`, but the core invariant is that `published` means "not invalidated yet" rather than "mined only."

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: Went through a few iterations on this — here's where we landed:

**Coarse derivation works** (pending vs active vs invalid): `orphaned` = `is_coinbase AND block_height IS NULL`; `replaced`/`failed` (direct) = existence in `tx_replacements`.

**Full replacement of the status enum does not**, for three reasons:

1. **`is_broadcast` ≠ `published`** — a tx can be valid/published because it was received from the network, not broadcast by this wallet. `published` means "active/not-invalidated," which is a broader concept than "I sent it."

2. **Distinct invalid states matter operationally** — `replaced` (RBF: can re-spend same inputs), `failed` (double-spend: permanent), and `orphaned` (coinbase reorg: recoverable on reconfirmation) have different recovery and audit semantics. A single boolean collapses all three and requires joins to reconstruct the reason.

3. **A fully-derived alternative needs _more_ columns** — to preserve the same information you'd need at minimum `is_broadcast`/source marker + `invalidation_reason` enum + optional `invalidated_by_tx_id`. That's strictly more schema surface than a single `status` column for the same information.

4. **Cascading invalidation** — downstream txs whose parent became invalid were never directly "replaced." Without pre-computed status, balance/coin-selection queries need a recursive CTE walking up the ancestor chain (O(depth) per tx on the hot path).

The `status` enum is the minimal representation that captures all lifecycle states without ambiguity — a pre-computed materialization set once at write time. Added a `2.1.2 Explicit Status vs. Derived Status` section to the ADR documenting this full trade-off analysis.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: Dependency chains are already represented by the `utxos` table's `(tx_id, output_index)` and the `spent_by_tx_id` link; the `spendable_utxos` view explicitly joins in parents (including `pending` parents) so that zero-latency chaining and cluster-mempool scenarios are supported.

Anything beyond serial UTXO dependencies (e.g., P2A anchor graphs) still reduces to the same parent → child edges: every dependent tx points to a UTXO in `utxos`. If additional metadata is needed for P2A, we can layer it on top of this UTXO graph.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: See the detailed analysis in the thread above. Summary: coarse derivation (valid vs invalid) works, but fully replacing the status enum requires preserving the distinct invalid states (`replaced`, `failed`, `orphaned`) which have different operational semantics (RBF re-spend, permanent double-spend, recoverable coinbase reorg). A boolean decomposition either loses that information or requires more columns than the single `status` field.

The `status` column is the minimal representation — a pre-computed materialization set at write time. Constraints/triggers enforce consistency with the underlying facts (confirmation, coinbase semantics); `failed`/`replaced` transition correctness is a write-path responsibility validated by tests.

Documented in the new ADR section `2.1.2 Explicit Status vs. Derived Status`.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: The table remains `utxos` because it is the canonical output set for the wallet — every row represents an output the wallet once owned. Spent rows are distinguished by their `spent_by_tx_id` pointer (non-NULL = spent), not physically deleted.

Calling it `outputs` or `txos` would lose the connection to existing wallet abstractions (`ListUnspent`, coin selection, etc.) that are UTXO-centric. The name signals that the table's primary operational purpose is serving the unspent subset, even though it retains spent history for audit and reconstruction.

We can add a short note clarifying that the table stores both spent and unspent outputs but retains the `utxos` name for consistency with existing wallet interfaces.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: `spent_input_index` records the exact input index in the spending transaction that consumed this UTXO. The primary purpose is deterministic reconstruction of the spending transaction's input set (mapping each input back to the UTXO it consumed without re-parsing `raw_tx`).

The `check_spent_tx_and_index_pair` constraint ensures we either have both `spent_by_tx_id` and `spent_input_index`, or neither — so the state is always consistent.

Re: `ANYONECANPAY` / multi-input flows — each input in a transaction has its own index and consumes exactly one UTXO. Each such UTXO gets its own `spent_by_tx_id` + `spent_input_index` pair. Nothing here enforces any relationship between inputs and outputs beyond that one-to-one mapping (one input index → one UTXO consumed), so the semantics stay consistent regardless of sighash flags or how many inputs contribute to a transaction's outputs.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: A single `replaced_by_tx_id` pointer on the transactions table can represent a linear chain (A→B→C) but cannot represent the full many-to-many graph: one tx can be replaced by multiple competitors simultaneously (e.g., conflicting double-spends from different peers), and one replacement can invalidate multiple prior txs that shared inputs. The separate `tx_replacements` table stores every directed edge (`replaced_tx_id → replacement_tx_id`) with `ON DELETE CASCADE`, unique constraints, and timestamps.

This gives clients the full audit trail — every replacement attempt, not just the latest successor — which is useful for debugging RBF chains and understanding conflict history in production.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: Good point. The current schema stores a caller-provided 32-byte token, so adding a doc note that it's an external lease identifier (e.g., from LND) would clarify intent. We can either rename to \`external_lock_id\` to emphasize that it originates outside the store, or keep \`lock_id\` and add a comment documenting it as an "opaque caller-supplied lock handle." Will add the clarifying note.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: Agreed — this column stores an absolute timestamp, so \`expires_at\` is more precise than the abstract noun \`expiration\`. Will update the column name and comment.

docs/developer/adr/0006-wtxmgr-sql-schema.md
```diff
@@ -0,0 +1,400 @@
...
+    received_time TIMESTAMPTZ NOT NULL,
+    is_coinbase BOOLEAN NOT NULL DEFAULT FALSE,
+    
+    -- Composite primary key is intentional: it allows foreign keys to enforce
+    -- wallet scoping (preventing cross-wallet references). Do not replace this
+    -- with a single-column PK on `id` unless all referencing FKs are also
+    -- reworked to maintain the same invariant.
+    CONSTRAINT pidx_transactions PRIMARY KEY (wallet_id, id),
+    CONSTRAINT uidx_transactions_hash UNIQUE (wallet_id, tx_hash),
+    CONSTRAINT valid_status CHECK (status IN ('pending', 'published', 'replaced', 'failed', 'orphaned')),
```
GustavoStingelin, line 117: Since we are keeping the status column to carry additional information, what do you think about moving statuses into a separate table?

That would align this design with the other tables we have already implemented. If we believe this approach is clearly better, it might also make sense to refactor the other tables to follow the same pattern, so we keep the schema consistent and easier to reason about over time.

GustavoStingelin (approved):
OK, let’s proceed with this approach. Once we have the concrete implementation in place, I can revisit some of the details.

docs/developer/adr/0006-wtxmgr-sql-schema.md
yyforyongyu: will revisit when implementing it!

PR #1166, opened 2026-02-02
docs: add keyvault ADR

GustavoStingelin:
This PR proposes a set of standards for upcoming implementations, documented through an ADR, and is explicitly open for discussion. The current proposal reflects my initial thinking and is meant as a starting point.

I would really appreciate feedback, alternative approaches, or refutations backed by prior experience or concrete evidence. Please feel free to challenge the design and suggest improvements.

### DIFF
--- a/docs/developer/adr/0010-keyvault-encryption-layer.md
+++ b/docs/developer/adr/0010-keyvault-encryption-layer.md
@@ -1,0 +1,108 @@
+# ADR 0010: Keyvault Encryption Layer
+
+## 1. Context
+
+The updated encryption model defined in ADR 0009 and the planned
+cryptographic primitive migration proposed in ADR 0007 require a clear
+boundary between domain logic and the SQL database layer. The
+legacy `waddrmgr` design tightly couples storage, locking, and key derivation,
+which complicates the SQL migration and makes encryption behavior hard to test
+in isolation.
+
+To address this, we need a dedicated component that owns lock state, key
+derivation, and encryption, while keeping the database layer strictly
+encryption agnostic.
+
+The `db.Store` remains available to other callers for non-cryptographic
+queries and updates.
+
+## 2. Decision
+
+We will introduce a dedicated **`wallet/internal/keyvault`** package that
+defines the encryption boundary between domain code and the SQL store layer.
+Keyvault accesses the database through `db.Store`, not by talking directly
+to the SQL backend.
+
+```mermaid
+flowchart TD
+    A[Domain Code] -->|crypto ops| B[keyvault]
+    A -->|non-crypto ops| E[db.Store]
+    B -->|derive| C[btcutil/hdkeychain]
+    B -->|decrypt/encrypt| F[XChaCha20Poly1305</br>XSalsa20Poly1305]
+    B -->|read/write| E
+    B -->|read/write| H[(memory cache)]
+    E --> D[(SQL DB)]
+    E --> G[(kvdb)]
+    classDef store fill:#f8f9fb,stroke:#9aa0a6,stroke-width:1px
+    class G,D store
+```
+
+### Responsibilities
+
+1. **Own lock state and key lifecycle**
+   Centralized management of unlock state, key derivation, key material
+   lifetime, and secure memory zeroing.
+
+2. **Expose typed domain interfaces**
+   Methods return `*btcec.PrivateKey`, `*btcec.PublicKey`, and
+   `btcutil.Address` instead of encrypted `[]byte` values.
+
+3. **Handle HD derivation**
+   Use `btcutil/hdkeychain` for BIP32 and BIP44 derivation and return or
+   persist derived keys as needed.
+
+4. **Maintain an in-memory cache**
+   Cache account level and derived keys to avoid repeated derivation and
+   database reads.
+
+5. **Support multi-wallet operation**
+   A single keyvault instance manages multiple wallets via `wallet_id`
+   parameters.
+
+6. **Track current and planned cryptographic primitives**
+   Encryption follows the accepted single-passphrase model in ADR 0009 and
+   adopts ADR 0007 once the XChaCha20-Poly1305 migration is implemented.
+
+7. **Coexist with `waddrmgr` during migration**
+   New code uses keyvault while legacy code continues to rely on
+   `waddrmgr` until the migration is complete.
+
+## 3. Consequences
+
+### Pros
+
+* **Separation of concerns**
+   Database code stores opaque bytes without knowledge of cryptography or lock
+   state.
+
+* **Type safety**
+  Callers work with strongly typed keys rather than raw blobs.
+
+* **Centralized lock management**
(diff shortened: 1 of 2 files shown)


docs/developer/adr/0006-xchacha20-poly1305-encryption.md
```diff
@@ -0,0 +1,96 @@
+# ADR 0006: XChaCha20-Poly1305 Encryption
```
yyforyongyu: A second thought on this - we can create a new PR for this ADR, and move this out as it feels like it's out of scope, given that we are focusing on sqlization.

docs/developer/adr/0006-xchacha20-poly1305-encryption.md
```diff
@@ -0,0 +1,96 @@
...
+material.
+
+While XSalsa20-Poly1305 remains secure, XChaCha20-Poly1305 is now the
+preferred successor in modern systems. It preserves the large nonce model
+while offering better library support and alignment with current AEAD
+standards.
+
+Key motivations for this change include:
+
+* Broad adoption of XChaCha20-Poly1305 in modern cryptographic systems
```
yyforyongyu, line 17: we an also mention https://github.com/lightning/bolts/blob/master/08-transport.md on why chacha is chosen here.

docs/developer/adr/0006-xchacha20-poly1305-encryption.md
```diff
@@ -0,0 +1,96 @@
...
+  Smaller and clearer code paths improve reviewability and long-term
+  maintenance.
+
+* **Nonce safety**
+   The 192-bit nonce space allows safe random nonce generation and preserves
+   compatibility with existing designs.
+
+### Cons
+
+* **Migration cost**
```
yyforyongyu, line 85: Yeah given this, i think a good opportunity to implement this is for the next release, where we implement the migration from kvdb to sql, that would be a perfect moment since we will encrypt/decrypt there.

docs/developer/adr/0007-single-passphrase-encryption.md
```diff
@@ -0,0 +1,63 @@
...
+    D --> F
+```
+
+### Key points
+
+1. **One master encryption key** derived from the user passphrase
+2. **Private data encrypted** (private keys, HD seeds, xprivs, scripts)
+3. **Public data plaintext** (addresses, pubkeys, transactions, balances)
+4. **Simplified hierarchy** by removing `CryptoPubKey` and `CryptoScriptKey`
+5. **Modern primitive** used for encryption is defined in
```
yyforyongyu: Given this, maybe we should implement ADR6, but only use it when the db is in native sql? this way, we don't need to touch kvdb migration for now.

docs/developer/adr/0008-cryptostore-encryption-layer.md
```diff
@@ -0,0 +1,110 @@
...
+derivation, and encryption, while keeping the database layer strictly
+encryption agnostic.
+
+The `db.Store` remains available to other callers for non-cryptographic
+queries and updates.
+
+## 2. Decision
+
+We will introduce a dedicated **`wallet/internal/cryptostore`** package that
+defines the encryption boundary between domain code and the SQL store layer.
```
yyforyongyu, line 22: Naming it `cryptostore` can be a bit misleading given that there's no store in this package. IIUC its only job is to encrypt and decrypt, sitting like a middleware between the upper layer and the db layer. Sth like `crypter` may be more fit.

yyforyongyu (changes requested):
Cool stuff - my overall assessment is, let's bring `ADR 0007` in, but leave `ADR 0006` and especially `ADR 0008` open for now - I think things will become more clear once we start working on the kvdb impl of the `Store`, which involves actually using the `Store` methods in the managers. Once that's started, we will have a better understanding on how `ADR 0008` should be like.

docs/developer/adr/0006-xchacha20-poly1305-encryption.md
GustavoStingelin: exactely

docs/developer/adr/0007-single-passphrase-encryption.md
GustavoStingelin: That's the way

docs/developer/adr/0008-cryptostore-encryption-layer.md
GustavoStingelin: Hmm, the intention is also to manage the lock state and in-memory caches for keys and crypto data, so we can centralize this logic. This also covers concerns like memory zeroing, to reduce the risk of memory dump attacks when the wallet is locked.

For pure crypto functions, we already have the `snacl` package at the root.

GustavoStingelin:
I created separate PRs for each ADR to discuss individually, will put this one as a draft until the others became merged

single passphrase https://github.com/btcsuite/btcwallet/pull/1170
xchacha20 https://github.com/btcsuite/btcwallet/pull/1171

docs/developer/adr/0006-xchacha20-poly1305-encryption.md
GustavoStingelin: addressed in the separate PR

docs/developer/adr/0008-cryptostore-encryption-layer.md
yyforyongyu: yeah i still find the naming confusing - at a glance people would treat it as a store, given that word is largely used in this repo so the contextual meaning of `store` is imprinted as storage, but in reality it's more like a decrypt/encrypt/caching center. Can you come up with a few other naming options?

Meanwhile just to be clear, this new layer won't perform key derivation, it's pure encrypt/decrypt right? Asking cause i saw this line,
> legacy `waddrmgr` design tightly couples storage, locking, and key derivation,

GustavoStingelin: I was thinking of keeping derivation in scope at the orchestration level: this layer would not implement HD derivation logic itself, but it would call the existing derivation functions. The goal is to centralize private-key usage in one place, avoid returning raw private key material to external callers, and make lock/unlock + memory zeroing safer and easier to enforce. That is why I designed it this way (the Mermaid diagram tries to show this boundary as well).
Given that, this component is conceptually close to a scoped version of the current address manager. For naming, I think `keyring` (or possibly `keymanager`) could communicate this better than `cryptostore`; `keyvault` could also work, but it may sound more like opaque storage than lifecycle/orchestration.

I’m open to other alternatives if they make the boundaries and responsibilities clearer.
WDYT?

yyforyongyu: yeah `keyvault` seems good!

GustavoStingelin: renamed!

yyforyongyu (approved):

PR #1168, opened 2026-02-06
kvdb: start implementing `Store` interface

yyforyongyu:
We build the skeleton to refactor the existing kvdb to meet the new `Store` interface.

As a starting point we add a simple method from the utxo store.

This has been review [here](https://github.com/yyforyongyu/btcwallet/pull/18) already.


GustavoStingelin (approved):
LGTM!

PR #1172, opened 2026-02-09
itest: implement `lnd` style itest framework

yyforyongyu:
We now start the framework, mostly just mimicking the `lnd`'s `itest` and `lntest` pkg.

In addition we also fix the git worktree, using a similar approach from [here](https://github.com/lightningnetwork/lnd/pull/10547).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -30,10 +30,22 @@ ifneq ($(workers),)
 LINT_WORKERS = --concurrency=$(workers)
 endif
 
+# Detect if we're in a git worktree. Use git rev-parse --git-common-dir to get
+# the path to the main git directory for the linter's diff processor to work
+# correctly with the new-from-rev setting.
+GIT_COMMON_DIR := $(shell \
+  common_dir="$$(git rev-parse --git-common-dir 2>/dev/null)"; \
+  if [ "$$common_dir" != ".git" ] && [ -n "$$common_dir" ]; then \
+    echo "$$common_dir"; \
+  fi)
+GIT_VOLUME := $(if $(GIT_COMMON_DIR),-v "$(GIT_COMMON_DIR):$(GIT_COMMON_DIR):ro",)
+
 DOCKER_TOOLS = docker run \
   --rm \
   -v $(shell bash -c "mkdir -p /tmp/go-build-cache; echo /tmp/go-build-cache"):/root/.cache/go-build \
-  -v $$(pwd):/build btcwallet-tools
+  -v $$(pwd):/build \
+  $(GIT_VOLUME) \
+  btcwallet-tools
 
 SQLFLUFF = docker run \
 	--rm \
--- a/Makefile
+++ b/Makefile
@@ -136,6 +136,22 @@ itest-db-race:
 	@$(call print, "Running $(IT_DB_LABEL) integration tests (race).")
 	env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(ITEST_DB_RACE)
 
+#? itest: Run integration tests
+#? itest (vars): chain=btcd|bitcoind|neutrino db=kvdb|sqlite|postgres
+#? itest (vars): icase=<regex> (filter itest cases)
+#? itest (vars): timeout=<duration> verbose=1 nocache=1
+#? itest (ex): make itest icase=manager
+#? itest (ex): make itest chain=btcd db=kvdb
+itest:
+	@$(call print, "Running integration tests.")
+	@$(GOTEST) -v ./itest \
+		-tags="itest $(DEV_TAGS) nolog" \
+		$(if $(icase),-test.run="TestBtcWallet/.*/$(icase)",) \
+		$(filter-out -test.run=%,$(TEST_FLAGS)) \
+		-args \
+		-chain="$(if $(chain),$(chain),btcd)" \
+		-db="$(if $(db),$(db),kvdb)"
+
 # =========
 # UTILITIES
 # =========
@@ -279,6 +295,7 @@ sql-lint-check:
 	unit-race \
 	unit-debug \
 	unit-bench \
+	itest \
 	itest-db \
 	itest-db-race \
 	fmt \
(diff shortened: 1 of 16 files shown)


itest/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+var (
+	// chainBackend defines the blockchain backend to be used for the
+	// integration tests.
+	// Options: "btcd" (default), "bitcoind", "neutrino".
+	chainBackend = flag.String(
+		"chain", "btcd",
+		"chain backend to use (btcd, bitcoind, neutrino)",
+	)
```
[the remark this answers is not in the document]
yyforyongyu: this is implemented in the followup pr

itest/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+	// dbBackend defines the database backend to be used for the wallet
+	// storage.
+	// Options: "kvdb" (default), "sqlite", "postgres".
+	//
+	// This flag allows verifying that the wallet functions correctly across all
+	// supported database drivers.
+	dbBackend = flag.String(
+		"db", "kvdb",
+		"database backend to use (kvdb, sqlite, postgres)",
+	)
```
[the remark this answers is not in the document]
yyforyongyu: this is implemented in the followup pr

Makefile
```diff
@@ -124,6 +136,22 @@ itest-db-race:
...
+#? itest (ex): make itest chain=btcd db=kvdb
+itest:
+	@$(call print, "Running integration tests.")
+	@$(GOTEST) -v ./itest \
+		-tags="integration $(DEV_TAGS) nolog" \
+		$(if $(icase),-test.run="TestBtcWallet/.*/$(icase)",) \
+		$(filter-out -test.run=%,$(TEST_FLAGS)) \
+		-args \
+		-chain="$(if $(chain),$(chain),btcd)" \
+		-db="$(if $(db),$(db),kvdb)"
```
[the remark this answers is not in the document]
yyforyongyu: this is implemented in the followup pr

itest/main_test.go
```diff
@@ -0,0 +1,86 @@
...
+// TestBtcWallet runs the btcwallet integration test suite.
+func TestBtcWallet(t *testing.T) {
+	if len(allTestCases) == 0 {
+		t.Skip("no integration test cases registered")
+	}
+
+	harness := bwtest.SetupHarness(t, *chainBackend, *dbBackend)
+
+	for _, tc := range allTestCases {
+		tc := tc
```
GustavoStingelin:
```suggestion
```
This line is not needed bc we are using a recent go version

itest/manager_test.go
```diff
@@ -0,0 +1,56 @@
...
+	require.NoError(h, err, "failed to create wallet")
+
+	err = w.Start(h.Context())
+	require.NoError(h, err, "failed to start wallet")
+	h.Cleanup(func() {
+		// We use a background context here because h.Context() might be
+		// cancelled already.
+		require.NoError(
+			h, w.Stop(context.Background()),
+			"failed to stop wallet",
```
GustavoStingelin: Looks like this fits in one line

bwtest/wait/wait.go
```diff
@@ -0,0 +1,54 @@
...
+		select {
+		case <-deadline.C:
+			if lastErr == nil {
+				return ErrNoResponse
+			}
+
+			return lastErr
+
+		case <-ticker.C:
+			err := f()
```
GustavoStingelin, line 50: The first call to `f` only happens after the first ticker tick (200ms). This means every single polling call in the harness has an unnecessary 200ms cold-start delay. We could call `f()` once before entering the select loop.

bwtest/harness_miner.go
```diff
@@ -0,0 +1,506 @@
...
+		h.Fatalf("nil tx")
+	}
+
+	txid := tx.TxHash()
+	h.AssertTxInMempool(txid)
+
+	// Ensure the mempool only contains our transaction so the mined block
+	// contains only the coinbase and this transaction.
+	mempoolTxids := h.AssertNumTxnsInMempool(1)
+	require.Len(h, mempoolTxids, 1, "expected exactly 1 txn in mempool")
```
GustavoStingelin:
```suggestion
```

the `AssertNumTxnsInMempool` already does the assertion

bwtest/chain_backend.go
```diff
@@ -0,0 +1,133 @@
+// Package bwtest provides the integration test framework for btcwallet.
+package bwtest
```
GustavoStingelin, line 2: Maybe this package should also use the `integration` build tag? since it is not intended for production use and does not follow the `_test` naming convention.

We could also move it inside the `itest/` directory? as it will only be used by those tests.

This would make its purpose explicit and ensure it is not included in regular builds.

GustavoStingelin (approved):
Looking good, just left a few potential improvements.

itest/list_on_test.go
```diff
@@ -0,0 +1,22 @@
+//go:build integration
```
GustavoStingelin: I just noticed that we are using the `itest` tag in the SQL DB integration tests. I think we should standardize on a single build tag to keep things consistent.

itest/main_test.go
yyforyongyu: Fixed!

itest/manager_test.go
yyforyongyu: Updated!

bwtest/wait/wait.go
yyforyongyu: Fixed!

bwtest/harness_miner.go
yyforyongyu: Fixed!

bwtest/chain_backend.go
yyforyongyu: > We could also move it inside the itest/ directory? as it will only be used by those tests.

Nah we will keep it here so this can be reuse - similar to how people are using `lntest` as a lib to build their tests, we can already use `bwtest` to replace the chain setup in controller benchmark test already.

itest/list_on_test.go
yyforyongyu: Updated!

PR #1173, opened 2026-02-09
Finish itest framework

yyforyongyu:
We now support,
- multiple chain backends: `neutrino`, `bitcoind` and `btcd`
- two versions of `bitcoind`
- add github workflows
- a few other `lnd`-style setups, like parallelizing test runners and uploading failed logs.

Once this is merged, we can add test cases without concerning the specific setups.


.github/workflows/main.yml
```diff
@@ -213,6 +215,176 @@ jobs:
...
+        uses: actions/cache@v4
+        with:
+          path: /home/runner/work/go
+          key: btcwallet-${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ github.job }}-${{ matrix.bitcoind_version }}-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            btcwallet-${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ github.job }}-${{ matrix.bitcoind_version }}-${{ hashFiles('**/go.sum') }}
+            btcwallet-${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ github.job }}-${{ matrix.bitcoind_version }}-
+            btcwallet-${{ runner.os }}-go-${{ env.GO_VERSION }}-${{ github.job }}-
+            btcwallet-${{ runner.os }}-go-${{ env.GO_VERSION }}-
+            btcwallet-${{ runner.os }}-go-
```
GustavoStingelin: Why does only the bitcoind job have this cache? Maybe we could add the same caching strategy to the other jobs as well.

We could also consider using a single matrix job with conditionals to install the appropriate backend node, which would reduce duplication and keep the workflow more consistent.

GustavoStingelin (changes requested):
Looking good, missing a rebase.

.github/workflows/main.yml
yyforyongyu: the cache is only needed for bitcoind since it's an external binary we download - btcd is just a go module dep (already covered by the go module cache) and neutrino doesn't need a separate binary, so there's nothing to cache for those.

re matrix job - i'd keep them separate, the setup steps are different enough that a matrix with conditionals would just add complexity. plus they already run in parallel so we're not losing anything.

GustavoStingelin: I am still not sure the current cache step is actually helping for `bitcoind`. The configured cache path is `/home/runner/work/go`, which corresponds to the Go workspace and build directory. However, the bitcoind binary is downloaded from a Docker image and extracted into `/usr/local/bin/bitcoind`, so it would not live under that cached path.

TIL that `actions/setup-go@v5` already provides module and build caching.

Perhaps we could either remove it or adjust the cache path to explicitly target where the binary is stored. If this starts to expand beyond the scope of this PR, we could also open a follow up issue to revisit the caching strategy and avoid scope creep here.

GustavoStingelin (approved):

.github/workflows/main.yml
yyforyongyu: agreed - yeah you are right this is not useful, gonna remove it

GustavoStingelin (approved):
Ack

PR #1182, opened 2026-02-17
wallet: add real db constructors out of itest

GustavoStingelin:
Adds real database constructors outside the itest package so integration tests exercise the same constructor code paths used in production, including connection setup and pool configuration.

### DIFF
--- a/wallet/internal/db/config.go
+++ b/wallet/internal/db/config.go
@@ -1,0 +1,90 @@
+package db
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+)
+
+const (
+	// DefaultMaxConnections is the default maximum number of permitted
+	// connections (both active and idle) to the database. We want to limit
+	// this so it isn't unlimited. The same value is used for the maximum
+	// number of idle connections, which can improve performance by avoiding
+	// the overhead of establishing a new connection for each query.
+	DefaultMaxConnections = 25
+
+	// DefaultConnIdleLifetime is the default amount of time a connection
+	// can be idle before being closed.
+	DefaultConnIdleLifetime = 5 * time.Minute
+
+	// DefaultConnectionTimeout is the default timeout for establishing
+	// a new database connection.
+	DefaultConnectionTimeout = 5 * time.Second
+)
+
+var (
+	// ErrEmptyDBPath is returned when an empty database path is provided.
+	ErrEmptyDBPath = errors.New("database path cannot be empty")
+
+	// ErrNegativeMaxConns is returned when MaxConnections is negative.
+	ErrNegativeMaxConns = errors.New("max connections must be non-negative")
+
+	// ErrEmptyDSN is returned when the DSN string is empty.
+	ErrEmptyDSN = errors.New("DSN is required")
+)
+
+// SqliteConfig holds the configuration for the SQLite database.
+type SqliteConfig struct {
+	// DBPath is the filesystem path to the SQLite database file.
+	DBPath string
+
+	// MaxConnections is the maximum number of open connections to the
+	// database. Set to zero to use DefaultMaxConnections.
+	MaxConnections int
+}
+
+// Validate checks that the SqliteConfig values are valid.
+func (c *SqliteConfig) Validate() error {
+	if c.DBPath == "" {
+		return ErrEmptyDBPath
+	}
+
+	if c.MaxConnections < 0 {
+		return ErrNegativeMaxConns
+	}
+
+	return nil
+}
+
+// PostgresConfig holds the configuration for the PostgreSQL database.
+type PostgresConfig struct {
+	// Dsn is the database connection string.
+	Dsn string
+
+	// MaxConnections is the maximum number of open connections to the
+	// database. Set to zero to use DefaultMaxConnections.
+	MaxConnections int
+}
+
+// Validate checks that the PostgresConfig values are valid.
+func (c *PostgresConfig) Validate() error {
+	if c.Dsn == "" {
+		return ErrEmptyDSN
+	}
+
+	// Parse the DSN using pgx to ensure it's a valid PostgreSQL
+	// connection string.
+	_, err := pgx.ParseConfig(c.Dsn)
+	if err != nil {
+		return fmt.Errorf("invalid DSN: %w", err)
+	}
+
+	if c.MaxConnections < 0 {
+		return ErrNegativeMaxConns
+	}
+
+	return nil
+}
--- a/wallet/internal/db/db_connectors_test.go
+++ b/wallet/internal/db/db_connectors_test.go
@@ -42,16 +42,16 @@ func newMockedTestDB(t *testing.T) *sql.DB {
 	return db
 }
 
-// TestNewPostgresWalletDB checks that the PostgresWalletDB constructor
+// TestNewPostgresStore checks that the PostgresStore constructor
 // properly guards against nil *sql.DB inputs and wires up the queries
 // correctly.
-func TestNewPostgresWalletDB(t *testing.T) {
+func TestNewPostgresStore(t *testing.T) {
 	t.Parallel()
 
 	t.Run("nil db", func(t *testing.T) {
 		t.Parallel()
 
-		db, err := NewPostgresWalletDB(nil)
+		db, err := NewPostgresStore(nil)
 		require.ErrorIs(t, err, ErrNilDB)
 		require.Nil(t, db)
 	})
@@ -61,7 +61,7 @@ func TestNewPostgresWalletDB(t *testing.T) {
 
 		sqlDB := newMockedTestDB(t)
 
-		db, err := NewPostgresWalletDB(sqlDB)
+		db, err := NewPostgresStore(sqlDB)
 		require.NoError(t, err)
 		require.NotNil(t, db)
 		require.Equal(t, sqlDB, db.db)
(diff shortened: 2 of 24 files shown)


wallet/internal/db/itest/sqlite_test.go
```diff
@@ -3,75 +3,33 @@
...
-	sqlcsqlite "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/sqlite"
 	"github.com/stretchr/testify/require"
 	_ "modernc.org/sqlite"
 )
 
-// NewSQLiteDB creates a new SQLite database for testing with migrations
-// applied. Each test gets its own temporary database file.
-func NewSQLiteDB(t *testing.T) *sql.DB {
+// NewTestStore creates a new SQLite database for testing with migrations
+// // applied. Each test gets its own temporary database file.
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/config.go
```diff
@@ -0,0 +1,84 @@
...
+	"fmt"
+	"time"
+
+	"github.com/jackc/pgx/v5"
+)
+
+const (
+	// DefaultMaxConnections is the default maximum number of permitted
+	// active and idle connections. We want to limit this so it isn't
+	// unlimited. We use the same value for the number of idle connections
```
yyforyongyu: What do you mean that we use the same number of idle connections?

wallet/internal/db/config.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// ErrEmptyDSN is returned when the DSN string is empty.
+	ErrEmptyDSN = errors.New("DSN is required")
+)
+
+// SqliteConfig holds the configuration for the SQLite database.
+type SqliteConfig struct {
+	// SkipMigrations determines whether to skip applying migrations on
+	// startup.
+	SkipMigrations bool
```
yyforyongyu: Q: when will we need to skip?

wallet/internal/db/sqlite.go
```diff
@@ -13,27 +15,82 @@ type SqliteStore struct {
...
 
-// NewSqliteStore creates a new SQLite-based WalletStore.
-func NewSqliteStore(db *sql.DB) (*SqliteStore, error) {
-	if db == nil {
-		return nil, ErrNilDB
+// NewSqliteStore creates a new SQLite-based WalletStore. It handles the full
+// connection setup including DSN construction with pragmas, connection
+// opening, health checks, connection pool configuration, and migration
+// application.
+func NewSqliteStore(cfg SqliteConfig, dbPath string) (
```
yyforyongyu: I think `dbPath` should be a part of `SqliteConfig`?

wallet/internal/db/config.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/jackc/pgx/v5"
+)
+
+const (
+	// DefaultMaxConnections is the default maximum number of permitted
+	// active and idle connections. We want to limit this so it isn't
+	// unlimited. We use the same value for the number of idle connections
+	// as this can speed up queries given a new connection doesn't need to
+	// be established each time.
+	DefaultMaxConnections = 25
```
yyforyongyu, line 17: I think this would be a bit high for sqlite - we should use different vars for the two dbs?

wallet/internal/db/pg.go
```diff
@@ -3,38 +3,83 @@ package db
...
+		return nil, fmt.Errorf("invalid config: %w", err)
 	}
 
-	return &PostgresWalletDB{
+	db, err := sql.Open("pgx", cfg.Dsn)
+	if err != nil {
+		return nil, fmt.Errorf("open database: %w", err)
+	}
+
+	err = db.PingContext(context.Background())
```
yyforyongyu: Q: should we use a timeout context?

wallet/internal/db/itest/fixtures_pg_test.go
```diff
@@ -182,3 +183,15 @@ func deleteAddress(ctx context.Context, dbConn *sql.DB,
 
 	return nil
 }
+
+func setupMaxAccountNumberTest(t *testing.T, store db.AccountStore,
+	walletID uint32) {
+
+	t.Helper()
+
+	pgStore := store.(*db.PostgresStore)
```
yyforyongyu, line 194: should assert the type

wallet/internal/db/db_connectors_test.go
```diff
@@ -1,116 +1,66 @@
...
+		cfg := PostgresConfig{
+			Dsn: "",
+		}
+
+		db, err := NewPostgresStore(cfg)
+		require.ErrorIs(t, err, ErrEmptyDSN)
 		require.Nil(t, db)
 	})
 
-	t.Run("valid db", func(t *testing.T) {
```
yyforyongyu, line 59: looks like this test case is removed?

wallet/internal/db/itest/sqlite_test.go
```diff
@@ -3,75 +3,33 @@
 package itest
 
 import (
-	"database/sql"
 	"path/filepath"
 	"testing"
 
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
-	sqlcsqlite "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/sqlite"
```
yyforyongyu, line 11: think we can also remove the `_ "modernc.org/sqlite"` below

yyforyongyu (changes requested):
Straightforward PR - left a few comments.

wallet/internal/db/sqlite.go
GustavoStingelin: added

wallet/internal/db/pg.go
GustavoStingelin: added, also receiving context from parent now

wallet/internal/db/itest/fixtures_pg_test.go
GustavoStingelin: asserted

wallet/internal/db/db_connectors_test.go
GustavoStingelin: Yes, the issue here is that in this file we are not able to connect to a Postgres instance, and also we do not have a way to inject the mocked database driver anymore.

I refactored the test names to make them more concise and created a new file, `store_test.go`, to properly isolate and handle these tests.

wallet/internal/db/itest/sqlite_test.go
GustavoStingelin: cool, removed

wallet/internal/db/config.go
GustavoStingelin: For now we are not using it, I copied it from LND. It might be useful later for some test related scenarios.

Do you think it would be better to remove it for now, since it is unused and the intent is not clear?

wallet/internal/db/config.go
GustavoStingelin: I clarified the comment, but the strategy itself was copied from LND.

We use the same `MaxConnections` value for both the maximum open connections and the maximum idle connections. This allows flexible distribution depending on load. For example, if it is set to 25:

* 10 active, 15 idle, total 25
* 25 active, 0 idle, total 25
* 0 active, 25 idle, total 25
* 5 active, 5 idle, total 10
* ...

wallet/internal/db/config.go
GustavoStingelin: Maybe the key point is that we use `MaxConnections` as an upper bound for how many operations can run concurrently against the database.

In practice, it behaves more like a cap on concurrency than a fixed number of connections that will always be used. The actual number of connections depends on how many goroutines are active. If managers only spawn three goroutines, the driver will only use three connections. If they spawn more, usage grows up to the configured limit.

So it is more accurate to think of it as a ceiling for concurrent database work, not a guaranteed number of active connections.

We also have test cases that intentionally stress concurrency to verify idempotency, such as `TestNewDerivedAddressConcurrent`, which spawns 20 goroutines at the same time. If the database is not configured with a sufficient connection limit, we hit `SQLITE_BUSY` errors.

At that point, the options would be:

* Reduce the number of concurrent workers in the tests to match SQLite limits.
* Increase the max connections only for tests, which could hide production concurrency issues.
* Add an extra pooling layer in front of SQLite, which feels unnecessary since SQLite already provides pooling behavior.

Given this, keeping a reasonable upper bound at the connection level seems like the most balanced approach for me.

wallet/internal/db/config.go
yyforyongyu: yea would like to remove it, mostly a safety concern as I couldn't think of any cases we need to skip.

wallet/internal/db/config.go
yyforyongyu: Guess my question is:
- why do we choose 25? Do we expect a lot of concurrent read and write?
- Can we use different values for Postgres vs SQLite?
- Can we make it configurable?

> If the database is not configured with a sufficient connection limit, we hit SQLITE_BUSY errors.

This is great! We definitely need to handle this error `SQLITE_BUSY`. In fact, it's been a long standing issue in `lnd`. I think we should catch this error and handle it gracefully.

None of this is blocking though. We can address them in the follow-up PR. But it would be nice to have some answers here first.

yyforyongyu (commented):
Nice - only comment is the `SkipMigrations`.

wallet/internal/db/config.go
GustavoStingelin: removed

wallet/internal/db/config.go
GustavoStingelin: These are good questions. Sorry for the long response.

For now, 25 is mostly an empirical default. It is not meant to be a carefully tuned value yet. The goal was simply to have a reasonable upper bound that avoids hitting concurrency issues too early. To find a better number, we will probably need dedicated stress tests in the wallet layer that simulate concurrent operations and help us observe where contention starts to appear.

Regarding configurability, the value is already configurable. If a value greater than zero is provided in the config, that value is used. The default of 25 is only applied when the config value is zero.

About using different limits for Postgres and SQLite, that is definitely possible. My intuition is that SQLite might actually benefit from a higher limit rather than a lower one.

The reason is that SQLite effectively serializes writes internally. Even if multiple connections are active, write operations will still execute sequentially because of the database lock. In practice, this means the connection pool behaves more like a queue of pending operations.

Postgres behaves differently. Since it supports true concurrent transactions, operations can run in parallel as long as they do not conflict on locks. Because of that, Postgres may complete work faster and reuse connections more efficiently under normal load.

However, under heavy load, Postgres can also start behaving more sequentially. In that scenario, having a similar upper bound still helps ensure there are enough available connections.

In practice, I expect most concurrency to be controlled at the manager level. If each manager performs only one or two database operations at a time while processing events, the connection pressure becomes fairly predictable. In that case, we can estimate the required pool size more reliably and avoid hitting `SQLITE_BUSY`.

The main risk appears when a large number of goroutines perform database operations without coordination. In that situation it is easy to run into contention problems regardless of the backend.

yyforyongyu (approved):

PR #1189, opened 2026-03-12
wallet: implement store pagination

GustavoStingelin:
This PR adds pagination support to list queries that may return very large result sets, such as addresses.

- List queries are now paginated by default. They require a query struct and return a single page of results, instead of returning the full dataset as before. This changes the behavior of the existing `List*` API.

- New interface methods named `IterSomething` were added. These provide seamless iteration over all results without requiring the caller to manage pagination, using the `iter.Seq2` API, and can be used as a replacement for the previous full-result queries.

### DIFF
--- a/wallet/internal/db/addresses_common.go
+++ b/wallet/internal/db/addresses_common.go
@@ -359,36 +359,14 @@ func getAddress[T any, Args any](ctx context.Context,
 	return nil, ErrAddressNotFound
 }
 
-// listAddresses is a generic helper that retrieves addresses using the provided
-// lister function and converts the results to AddressInfo structs.
-func listAddresses[T any, Args any](ctx context.Context,
-	lister func(context.Context, Args) ([]T, error), args Args,
-	toInfo func(T) (*AddressInfo, error)) ([]AddressInfo, error) {
+// nextListAddressesQuery returns a query with its pagination cursor advanced to
+// the provided value.
+func nextListAddressesQuery(q ListAddressesQuery,
+	cursor uint32) ListAddressesQuery {
 
-	rows, err := lister(ctx, args)
-	if err != nil {
-		return nil, fmt.Errorf("list addresses: %w", err)
-	}
+	q.Page = q.Page.WithAfter(cursor)
 
-	return addressInfosFromRows(rows, toInfo)
-}
-
-// addressInfosFromRows converts a slice of row types to AddressInfo structs
-// using the provided converter function.
-func addressInfosFromRows[T any](rows []T,
-	toInfo func(T) (*AddressInfo, error)) ([]AddressInfo, error) {
-
-	infos := make([]AddressInfo, len(rows))
-	for i, row := range rows {
-		info, err := toInfo(row)
-		if err != nil {
-			return nil, err
-		}
-
-		infos[i] = *info
-	}
-
-	return infos, nil
+	return q
 }
 
 // derivedAddressAdapters groups the functions needed to create a
--- a/wallet/internal/db/addresses_pg.go
+++ b/wallet/internal/db/addresses_pg.go
@@ -9,6 +9,8 @@ import (
 	sqlcpg "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/postgres"
 )
 
+var _ AddressStore = (*PostgresStore)(nil)
+
 // GetAddress retrieves information about a specific address, identified by
 // its script pubkey.
 func (s *PostgresStore) GetAddress(ctx context.Context,
--- a/wallet/internal/db/addresses_pg.go
+++ b/wallet/internal/db/addresses_pg.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"database/sql"
 	"fmt"
+	"iter"
 	"time"
 
+	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
 	sqlcpg "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/postgres"
 )
 
@@ -31,19 +33,31 @@ func (s *PostgresStore) GetAddress(ctx context.Context,
 	return getAddressByQuery(ctx, query, getByScript)
 }
 
-// ListAddresses returns a slice of AddressInfo for all addresses in a given
-// account.
+// ListAddresses returns a page of addresses matching the given query.
 func (s *PostgresStore) ListAddresses(ctx context.Context,
-	query ListAddressesQuery) ([]AddressInfo, error) {
+	query ListAddressesQuery) (page.Result[AddressInfo, uint32], error) {
 
-	return listAddresses(
-		ctx, s.queries.ListAddressesByAccount,
-		sqlcpg.ListAddressesByAccountParams{
-			WalletID:    int64(query.WalletID),
-			Purpose:     int64(query.Scope.Purpose),
-			CoinType:    int64(query.Scope.Coin),
-			AccountName: query.AccountName,
-		}, pgAddressRowToInfo,
+	items, err := pgListAddressesByAccount(ctx, s.queries, query)
+	if err != nil {
+		return page.Result[AddressInfo, uint32]{}, err
+	}
+
+	result := page.BuildResult(
+		query.Page, items,
+		func(item AddressInfo) uint32 {
+			return item.ID
+		},
+	)
+
+	return result, nil
+}
+
+// IterAddresses returns an iterator over paginated address results.
+func (s *PostgresStore) IterAddresses(ctx context.Context,
+	query ListAddressesQuery) iter.Seq2[AddressInfo, error] {
+
+	return page.Iter(
+		ctx, query, s.ListAddresses, nextListAddressesQuery,
 	)
 }
 
@@ -299,4 +313,54 @@ func pgAddressRowToInfo[T pgAddressInfoRow](row T) (*AddressInfo, error) {
 	}
 
 	return info, nil
+}
+
+// pgListAddressesByAccount lists addresses filtered by wallet ID, key scope,
+// and account name, with pagination support.
+func pgListAddressesByAccount(ctx context.Context, q *sqlcpg.Queries,
+	query ListAddressesQuery) ([]AddressInfo, error) {
+
+	rows, err := q.ListAddressesByAccount(
+		ctx, pgBuildAddressPageParams(query),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list addresses by account: %w", err)
+	}
+
+	items := make([]AddressInfo, len(rows))
+	for i, row := range rows {
+		item, err := pgAddressRowToInfo(row)
+		if err != nil {
+			return nil,
+				fmt.Errorf("list addresses by account: map address row: %w",
+					err)
+		}
+
+		items[i] = *item
+	}
+
+	return items, nil
+}
+
+// pgBuildAddressPageParams translates a ListAddresses query to
+// ListAddressesByAccount parameters, handling pagination cursors.
+func pgBuildAddressPageParams(
+	q ListAddressesQuery) sqlcpg.ListAddressesByAccountParams {
+
+	params := sqlcpg.ListAddressesByAccountParams{
+		WalletID:    int64(q.WalletID),
+		Purpose:     int64(q.Scope.Purpose),
+		CoinType:    int64(q.Scope.Coin),
+		AccountName: q.AccountName,
+		PageLimit:   int64(q.Page.QueryLimit()),
+	}
+
+	if cursor, ok := q.Page.After(); ok {
+		params.CursorID = sql.NullInt64{
+			Int64: int64(cursor),
+			Valid: true,
+		}
+	}
+
+	return params
 }
--- a/wallet/internal/db/addresses_sqlite.go
+++ b/wallet/internal/db/addresses_sqlite.go
@@ -3,8 +3,11 @@ package db
 import (
 	"context"
 	"database/sql"
+	"fmt"
+	"iter"
 	"time"
 
+	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
 	sqlcsqlite "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/sqlite"
 )
 
@@ -30,19 +33,31 @@ func (s *SqliteStore) GetAddress(ctx context.Context,
 	return getAddressByQuery(ctx, query, getByScript)
 }
 
-// ListAddresses returns a slice of AddressInfo for all addresses in a given
-// account.
+// ListAddresses returns a page of addresses matching the given query.
 func (s *SqliteStore) ListAddresses(ctx context.Context,
-	query ListAddressesQuery) ([]AddressInfo, error) {
+	query ListAddressesQuery) (page.Result[AddressInfo, uint32], error) {
 
-	return listAddresses(
-		ctx, s.queries.ListAddressesByAccount,
-		sqlcsqlite.ListAddressesByAccountParams{
-			WalletID:    int64(query.WalletID),
-			Purpose:     int64(query.Scope.Purpose),
-			CoinType:    int64(query.Scope.Coin),
-			AccountName: query.AccountName,
-		}, sqliteAddressRowToInfo,
+	items, err := sqliteListAddressesByAccount(ctx, s.queries, query)
+	if err != nil {
+		return page.Result[AddressInfo, uint32]{}, err
+	}
+
+	result := page.BuildResult(
+		query.Page, items,
+		func(item AddressInfo) uint32 {
+			return item.ID
+		},
+	)
+
+	return result, nil
+}
+
+// IterAddresses returns an iterator over paginated address results.
+func (s *SqliteStore) IterAddresses(ctx context.Context,
+	query ListAddressesQuery) iter.Seq2[AddressInfo, error] {
+
+	return page.Iter(
+		ctx, query, s.ListAddresses, nextListAddressesQuery,
 	)
 }
 
@@ -292,4 +307,51 @@ func sqliteAddressRowToInfo[T sqliteAddressInfoRow](row T) (*AddressInfo,
 	}
 
 	return info, nil
+}
+
+// sqliteListAddressesByAccount lists addresses filtered by wallet ID, key
+// scope, and account name, with pagination support.
+func sqliteListAddressesByAccount(ctx context.Context, q *sqlcsqlite.Queries,
+	query ListAddressesQuery) ([]AddressInfo, error) {
+
+	rows, err := q.ListAddressesByAccount(
+		ctx, sqliteBuildAddressPageParams(query),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list addresses by account: %w", err)
+	}
+
+	items := make([]AddressInfo, len(rows))
+	for i, row := range rows {
+		item, err := sqliteAddressRowToInfo(row)
+		if err != nil {
+			return nil,
+				fmt.Errorf("list addresses by account: map address row: %w",
+					err)
+		}
+
+		items[i] = *item
+	}
+
+	return items, nil
+}
+
+// sqliteBuildAddressPageParams translates a ListAddresses query to
+// ListAddressesByAccount parameters, handling pagination cursors.
+func sqliteBuildAddressPageParams(
+	q ListAddressesQuery) sqlcsqlite.ListAddressesByAccountParams {
+
+	params := sqlcsqlite.ListAddressesByAccountParams{
+		WalletID:    int64(q.WalletID),
+		Purpose:     int64(q.Scope.Purpose),
+		CoinType:    int64(q.Scope.Coin),
+		AccountName: q.AccountName,
+		PageLimit:   int64(q.Page.QueryLimit()),
+	}
+
+	if cursor, ok := q.Page.After(); ok {
+		params.CursorID = int64(cursor)
+	}
+
+	return params
 }
--- a/wallet/internal/db/addresses_sqlite.go
+++ b/wallet/internal/db/addresses_sqlite.go
@@ -8,6 +8,8 @@ import (
 	sqlcsqlite "github.com/btcsuite/btcwallet/wallet/internal/db/sqlc/sqlite"
 )
 
+var _ AddressStore = (*SqliteStore)(nil)
+
 // GetAddress retrieves information about a specific address, identified by
 // its script pubkey.
 func (s *SqliteStore) GetAddress(ctx context.Context,
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -240,6 +240,12 @@ type Block struct {
 	Timestamp time.Time
 }
 
+// ListWalletsQuery contains the parameters for listing wallets.
+type ListWalletsQuery struct {
+	// Page holds the pagination parameters for this query.
+	Page page.Request[uint32]
+}
+
 // CreateWalletParams contains the parameters required to create a new wallet.
 type CreateWalletParams struct {
 	// Name is the name of the new wallet.
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -10,6 +10,7 @@ import (
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
 )
 
 const (
@@ -710,6 +711,9 @@ type ListAddressesQuery struct {
 
 	// Scope is the key scope of the account.
 	Scope KeyScope
+
+	// Page holds the pagination parameters for this query.
+	Page page.Request[uint32]
 }
 
 // --------------------
--- a/wallet/internal/db/interface.go
+++ b/wallet/internal/db/interface.go
@@ -3,6 +3,9 @@ package db
 import (
 	"context"
 	"errors"
+	"iter"
+
+	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
 )
 
 var (
@@ -235,10 +238,15 @@ type AddressStore interface {
 	// an error if the address is not found.
 	GetAddress(ctx context.Context, query GetAddressQuery) (*AddressInfo, error)
 
-	// ListAddresses returns a slice of AddressInfo for all addresses in a
-	// given account. It returns an empty slice if no addresses are found.
-	ListAddresses(ctx context.Context, query ListAddressesQuery) ([]AddressInfo,
-		error)
+	// ListAddresses returns one page of addresses for the given query,
+	// including a next-cursor for the following page.
+	ListAddresses(ctx context.Context, query ListAddressesQuery) (
+		page.Result[AddressInfo, uint32], error)
+
+	// IterAddresses returns an iterator that fetches pages transparently and
+	// yields addresses one by one until exhaustion or error.
+	IterAddresses(ctx context.Context,
+		query ListAddressesQuery) iter.Seq2[AddressInfo, error]
 
 	// GetAddressSecret retrieves the encrypted secret material for a given
 	// address. Returns the AddressSecret containing encrypted private key
--- a/wallet/internal/db/interface.go
+++ b/wallet/internal/db/interface.go
@@ -142,10 +142,15 @@ type WalletStore interface {
 	// error if the wallet is not found.
 	GetWallet(ctx context.Context, name string) (*WalletInfo, error)
 
-	// ListWallets returns a slice of WalletInfo for all wallets stored in
-	// the database. It returns an empty slice if no wallets are found, or
-	// an error if the retrieval fails.
-	ListWallets(ctx context.Context) ([]WalletInfo, error)
+	// ListWallets returns one page of wallets for the given query, including a
+	// next-cursor for the following page.
+	ListWallets(ctx context.Context, query ListWalletsQuery) (
+		page.Result[WalletInfo, uint32], error)
+
+	// IterWallets returns an iterator that fetches pages transparently and
+	// yields wallets one by one until exhaustion or error.
+	IterWallets(ctx context.Context,
+		query ListWalletsQuery) iter.Seq2[WalletInfo, error]
 
 	// UpdateWallet updates various properties of a wallet, such as its
 	// birthday, birthday block, or sync state. The specific fields to
--- a/wallet/internal/db/itest/address_store_test.go
+++ b/wallet/internal/db/itest/address_store_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
 	"github.com/stretchr/testify/require"
 )
 
@@ -35,9 +36,9 @@ func mockDeriveFunc() db.AddressDerivationFunc {
 	}
 }
 
+// newDerivedAddress creates and returns a derived address for testing.
 func newDerivedAddress(t *testing.T, store db.AddressStore, walletID uint32,
 	scope db.KeyScope, accountName string, change bool) *db.AddressInfo {
-
 	t.Helper()
 
 	info, err := store.NewDerivedAddress(
@@ -53,10 +54,11 @@ func newDerivedAddress(t *testing.T, store db.AddressStore, walletID uint32,
 	return info
 }
 
+// createDerivedAddresses creates and returns a slice of derived addresses for
+// testing.
 func createDerivedAddresses(t *testing.T, store db.AddressStore,
 	walletID uint32, scope db.KeyScope, accountName string, change bool,
 	count int) []db.AddressInfo {
-
 	t.Helper()
 
 	addresses := make([]db.AddressInfo, 0, count)
@@ -70,9 +72,9 @@ func createDerivedAddresses(t *testing.T, store db.AddressStore,
 	return addresses
 }
 
+// getAccountByName retrieves an account by wallet, scope, and name for testing.
 func getAccountByName(t *testing.T, store db.AccountStore, walletID uint32,
 	scope db.KeyScope, accountName string) *db.AccountInfo {
-
 	t.Helper()
 
 	account, err := store.GetAccount(
@@ -81,6 +83,45 @@ func getAccountByName(t *testing.T, store db.AccountStore, walletID uint32,
 	require.NoError(t, err)
 
 	return account
+}
+
+// collectAddressPages collects paginated address results by iterating
+// through all pages from ListAddresses, using cursor pagination until
+// Next is nil.
+func collectAddressPages(t *testing.T, store db.AddressStore,
+	query db.ListAddressesQuery) []page.Result[db.AddressInfo, uint32] {
+	t.Helper()
+
+	pages := make([]page.Result[db.AddressInfo, uint32], 0)
+	for {
+		pageResult, err := store.ListAddresses(t.Context(), query)
+		require.NoError(t, err)
+		pages = append(pages, pageResult)
+
+		if pageResult.Next == nil {
+			return pages
+		}
+
+		query.Page = query.Page.WithAfter(*pageResult.Next)
+	}
+}
+
+// flattenAddressPages flattens paginated address results into a single
+// slice containing all addresses from all pages.
+func flattenAddressPages(
+	pages []page.Result[db.AddressInfo, uint32]) []db.AddressInfo {
+
+	count := 0
+	for i := range pages {
+		count += len(pages[i].Items)
+	}
+
+	addresses := make([]db.AddressInfo, 0, count)
+	for i := range pages {
+		addresses = append(addresses, pages[i].Items...)
+	}
+
+	return addresses
 }
 
 // TestNewImportedAddress verifies that NewImportedAddress correctly imports
@@ -236,7 +277,9 @@ func TestNewImportedAddressWithEncryptedScript(t *testing.T) {
 	queries := store.Queries()
 	walletID := newWallet(t, store, "wallet-encrypted-script")
 	createImportedAccount(t, store, walletID, db.KeyScopeBIP0044, "imported")
-	createImportedAccount(t, store, walletID, db.KeyScopeBIP0049Plus, "imported")
+	createImportedAccount(
+		t, store, walletID, db.KeyScopeBIP0049Plus, "imported",
+	)
 
 	redeemScript := RandomBytes(32)
 	witnessScript := RandomBytes(48)
@@ -661,9 +704,9 @@ func TestGetAddress(t *testing.T) {
 	}
 }
 
-// TestListAddresses verifies that ListAddresses correctly returns all
-// addresses for an account in a specified scope, filters by scope
-// appropriately, and handles empty results without error.
+// TestListAddresses verifies that ListAddresses correctly returns addresses
+// with page-contract behavior, filters by scope appropriately, and handles
+// empty results without error.
 func TestListAddresses(t *testing.T) {
 	t.Parallel()
 
@@ -780,7 +823,7 @@ func TestListAddresses(t *testing.T) {
 			walletID := newWallet(t, store, tc.name+"-wallet")
 
 			query := tc.setupFunc(t, store, store, walletID)
-			addrs, err := store.ListAddresses(t.Context(), query)
+			pageResult, err := store.ListAddresses(t.Context(), query)
 
 			if tc.wantErr != nil {
 				require.ErrorIs(t, err, tc.wantErr)
@@ -788,6 +831,7 @@ func TestListAddresses(t *testing.T) {
 			}
 
 			require.NoError(t, err)
+			addrs := pageResult.Items
 			require.Len(t, addrs, tc.wantCount)
 
 			if tc.validate != nil {
@@ -946,7 +990,7 @@ func TestListAddressesOrdering(t *testing.T) {
 		t, store, walletID, db.KeyScopeBIP0084, "ordering-account", true, 3,
 	)
 
-	addresses, err := store.ListAddresses(
+	pageResult, err := store.ListAddresses(
 		t.Context(),
 		db.ListAddressesQuery{
 			WalletID:    walletID,
@@ -956,6 +1000,7 @@ func TestListAddressesOrdering(t *testing.T) {
 	)
 
 	require.NoError(t, err)
+	addresses := pageResult.Items
 	require.Len(t, addresses, 6)
 
 	// Separate addresses by branch for verification.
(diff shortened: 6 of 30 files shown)


wallet/internal/db/data_types.go
```diff
@@ -508,6 +519,27 @@ type GetAccountQuery struct {
...
+	// the derived group.
+	Imported bool
+
+	// AccountNumber is the BIP44 account number of the last derived
+	// account. Zero when Imported is true.
+	AccountNumber uint32
+
+	// RowID is the internal account row id used as the final stable
+	// tie-breaker when multiple rows share the same visible sort fields.
+	RowID int64
```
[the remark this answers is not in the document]
GustavoStingelin: Added as public ID

wallet/internal/db/itest/account_store_test.go
```diff
@@ -840,3 +1115,67 @@ func findAccountInList(t *testing.T, accounts []db.AccountInfo,
...
+	err := store.DB().QueryRowContext(
+		t.Context(),
+		"SELECT a.id FROM accounts a JOIN key_scopes ks ON a.scope_id = ks.id "+
+			"WHERE ks.wallet_id = $1 AND ks.purpose = $2 "+
+			"AND ks.coin_type = $3 AND a.account_name = $4",
+		int64(walletID),
+		int64(info.KeyScope.Purpose),
+		int64(info.KeyScope.Coin),
+		info.AccountName,
+	).Scan(&rowID)
```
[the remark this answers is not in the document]
GustavoStingelin: removed

wallet/internal/db/interface.go
```diff
@@ -109,10 +110,15 @@ type WalletStore interface {
...
-	ListWallets(ctx context.Context) ([]WalletInfo, error)
+	// ListWallets returns one page of wallets for the given query. It
+	// returns an empty slice if no wallets are found.
+	ListWallets(ctx context.Context, query ListWalletsQuery) ([]WalletInfo,
+		error)
+
+	// IterWallets returns an iterator that fetches pages transparently and
+	// yields wallets one by one until exhaustion or error.
+	IterWallets(ctx context.Context,
+		query ListWalletsQuery) iter.Seq2[WalletInfo, error]
```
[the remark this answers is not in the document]
GustavoStingelin: Yeah, that's near the same.

wallet/internal/db/interface.go
```diff
@@ -162,6 +168,11 @@ type AccountStore interface {
 	ListAccounts(ctx context.Context, query ListAccountsQuery) (
 		[]AccountInfo, error)
 
+	// IterAccounts returns an iterator that fetches pages transparently and
+	// yields accounts one by one until exhaustion or error.
+	IterAccounts(ctx context.Context,
+		query ListAccountsQuery) iter.Seq2[AccountInfo, error]
```
GustavoStingelin: I am thinking that, since we will always rely on the automatic pagination provided by the iterator, it may be better to drop the list method entirely, or change its signature to return `iter.Seq2`.

This would expose a single way to query these results and help prevent misuse.

wallet/internal/db/page/iter_test.go
```diff
@@ -0,0 +1,347 @@
...
+	"context"
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+var errTest = errors.New("test error")
+
+func TestIterTraversal(t *testing.T) {
```
yyforyongyu, line 22: None of these tests have docs?

wallet/internal/db/queries/postgres/accounts.sql
```diff
@@ -217,11 +220,67 @@ WHERE
...
+    ks.wallet_id = $1
+    AND ks.purpose = $2
+    AND ks.coin_type = $3
+    -- sqlc.arg() calls are bind parameters, not column references; the
+    -- RF02 suppression below silences a false-positive from sqlfluff,
+    -- which cannot distinguish sqlc pseudo-functions from column names
+    -- in a multi-table JOIN context.
+    AND (
+        (
+            NOT sqlc.arg(cursor_imported)::BOOL -- noqa: RF02
```
yyforyongyu: Add some inline docs to explain the `noqa`?

wallet/internal/db/addresses_pg.go
```diff
@@ -300,3 +303,57 @@ func pgAddressRowToInfo[T pgAddressInfoRow](row T) (*AddressInfo, error) {
...
+		pgBuildFirstPageParams(query),
+		p.q.ListAddressesByAccountNextPage,
+		pgBuildPagedParams(query),
+		query.Page,
+		pgAddressRowToInfo[sqlcpg.ListAddressesByAccountFirstPageRow],
+		pgAddressRowToInfo[sqlcpg.ListAddressesByAccountNextPageRow],
+	)
+}
+
+func pgBuildFirstPageParams(q ListAddressesQuery) func(
```
yyforyongyu: it seems to me that a lot of the new methods are not documented, maybe we could point out the docs `@docs/developer/ENGINEERING_GUIDE.md` and  `@docs/developer/contribution_guidelines.md ` so agents will follow.

wallet/internal/db/page/iter.go
```diff
@@ -0,0 +1,48 @@
...
+// Iter iterates through paginated results by repeatedly fetching pages
+// and computing the next query from the last returned item.
+func Iter[Q, R any](ctx context.Context, query Q,
+	fetchPage func(context.Context, Q) ([]R, error),
+	nextQuery func(Q, []R) Q) iter.Seq2[R, error] {
+
+	return func(yield func(R, error) bool) {
+		for {
+			select {
+			case <-ctx.Done():
```
yyforyongyu: This check is redundant - I have also seen it multiple times that LLMs turn to have this pattern, weird...but `fetchPage` below will already handle it if the context is done. Plus once we are in the `for _, item := range items {` loop, the cancellation won't stop the yield emission.

wallet/internal/db/page/iter.go
```diff
@@ -0,0 +1,48 @@
+package page
+
+import (
+	"context"
+	"iter"
+)
+
+// Iter iterates through paginated results by repeatedly fetching pages
+// and computing the next query from the last returned item.
+func Iter[Q, R any](ctx context.Context, query Q,
```
yyforyongyu: I think the current naming is a bit hard to follow, so suggest sth like,
```go
func Iter[Query, Item any](
	ctx context.Context,
	initial Query,
	fetchPage func(context.Context, Query) ([]Item, error),
	nextQuery func(Query, []Item) Query,
) iter.Seq2[Item, error]
```

wallet/internal/db/page/options.go
```diff
@@ -0,0 +1,49 @@
...
+	// Cursor is the pagination cursor from the previous page.
+	// Nil means the first page.
+	Cursor *C
+}
+
+// Size returns the normalized page size for this option set. A size of zero
+// returns DefaultPageSize. A size greater than MaxPageSize is clamped to
+// MaxPageSize.
+func (o Options[C]) Size() uint32 {
+	if o.size == 0 {
```
yyforyongyu: What if the caller intentionally doesn't want to set the size, hence skipping the cursor? so these two quries:
- `ListAccounts(ctx, ListAccountsQuery{WalletID: id})`, means “caller never asked for pagination” in old code.
- `ListAccounts(ctx, ListAccountsQuery{WalletID: id, Page: page.Options[Cursor]{}.WithSize(100)})`, means “caller explicitly wants paginated results, 100 at a time”.

But I guess given we are in a side branch, we can change all public APIs behaviors here to always use the pagination.

wallet/internal/db/itest/address_store_test.go
```diff
@@ -83,6 +84,16 @@ func getAccountByName(t *testing.T, store db.AccountStore, walletID uint32,
 	return account
 }
 
+func nextAddressCursor(t *testing.T, items []db.AddressInfo) uint32 {
+
```
yyforyongyu: unnecessary newline

wallet/internal/db/page/iter.go
```diff
@@ -0,0 +1,48 @@
...
+			if err != nil {
+				var zero R
+
+				yield(zero, err)
+
+				return
+			}
+
+			for _, item := range items {
+				if !yield(item, nil) {
```
yyforyongyu, line 43: I guess if we wanna catch the cancel signal, we need to perform it here.

wallet/internal/db/page/options.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+// Options holds the pagination parameters used in queries.
+type Options[C any] struct {
+	// Size is the maximum number of items to return per page.
+	size uint32
+
+	// Cursor is the pagination cursor from the previous page.
+	// Nil means the first page.
+	Cursor *C
```
yyforyongyu: Does it mean this is mutable?

wallet/internal/db/page/options.go
```diff
@@ -0,0 +1,49 @@
...
+// WithSize returns a copy of options with size replaced.
+func (o Options[C]) WithSize(size uint32) Options[C] {
+	o.size = size
+
+	return o
+}
+
+// WithCursor returns a copy of options with cursor replaced.
+func (o Options[C]) WithCursor(cursor C) Options[C] {
+	o.Cursor = &cursor
```
yyforyongyu: to keep it immutable, i think sth like,
```go
c := cursor
o.cursor = &c
```
This is not a deep copy, but it should prevent direct mutation through `opts.Cursor`.

wallet/internal/db/page/options_test.go
```diff
@@ -0,0 +1,138 @@
...
+				return
+			}
+
+			require.NotNil(t, tc.options.Cursor)
+			require.Equal(t, tc.wantCursor, *tc.options.Cursor)
+		})
+	}
+}
+
+func TestOptionsWithSizeImmutability(t *testing.T) {
```
yyforyongyu: Good that we have these immutability tests, but it seems like we only test against `uint32` while we support `Options[C any]`. I think we need to add more tests.

yyforyongyu: That said, I think it's better to keep the `cursor` unexported - maybe add a new method `Cursor` and return the copy for reading?

wallet/internal/db/itest/address_store_test.go
```diff
@@ -960,6 +971,111 @@ func TestListAddressesOrdering(t *testing.T) {
...
+		Page:        page.Options[uint32]{}.WithSize(uint32(pageSize)),
+	}
+
+	items1, err := store.ListAddresses(t.Context(), query)
+	require.NoError(t, err)
+	require.Len(t, items1, 2)
+	require.Equal(t, accountA[:2], items1)
+
+	query.Page = query.Page.WithCursor(nextAddressCursor(t, items1))
+	items2, err := store.ListAddresses(t.Context(), query)
```
yyforyongyu: If we adopt the new design, the call will instead be,
```go
res, err := store.ListAccounts(ctx, q)
if err != nil { ... }
for _, acc := range res.Items {
	...
}
if res.Next != nil {
	q.Page.After = res.Next
}
```

yyforyongyu (changes requested):
Hmm I think the new API exposes too many details to the caller, so ask a few LLMs on how to improve it plus a few of my own thoughts, I think the design can be improved.

yyforyongyu:
Yes — here’s the redesign I’d recommend.

The short version:

- `List*` should return a page object, not a bare slice
- the store should return the next cursor explicitly
- callers should never derive the next cursor from the last item
- `Iter*` should be a thin convenience wrapper over `List*`
- use `LIMIT n+1` internally so you know whether another page exists without an extra empty fetch
- simplify SQL where it is safe, but do not let SQL unification block the API redesign

**Goals**

The redesigned API should make these things true:

- the caller asks for a page and gets both `Items` and `Next`
- the store owns sort order and cursor derivation
- page requests are simple data, not builder objects
- iteration over all results is easy
- implementation stays boring and local, not spread across several generic helper layers

**What’s wrong with the current shape**

The current design has a few structural problems:

- `List*` returns `[]Item`, so the caller must know how to derive the next cursor
- that leaks ordering logic into item types like `AccountInfo.Cursor()`
- `page.Options` adds ceremony without really making the API safer
- `page.Iter` is generic, but the stores still have to do a lot of setup around it
- splitting SQL into `FirstPage` and `NextPage` doubles row/param types and causes a lot of boilerplate
- the iterator needs an extra empty fetch to know it is done

The main design mistake is ownership: the store knows pagination semantics, but the caller is currently forced to participate in them.

**Proposed API**

I would replace `page.Options` with a plain request type, and replace bare slice returns with a page result.

```go
package page

const (
  DefaultLimit = 100
  MaxLimit     = 1000
)

type Request[C any] struct {
  Limit uint32
  After *C
}

func (r Request[C]) EffectiveLimit() uint32 {
  switch {
  case r.Limit == 0:
    return DefaultLimit
  case r.Limit > MaxLimit:
    return MaxLimit
  default:
    return r.Limit
  }
}

type Result[T any, C any] struct {
  Items []T
  Next  *C
}
```

This is a much better fit:

- `Request` is just input data
- `Result` is just output data
- `Next` is owned by the store, not reconstructed by the caller

**Query types**

Then the store query types become simple and obvious.

```go
type ListWalletsQuery struct {
  Page page.Request[uint32]
}

type AccountCursor struct {
  Imported      bool
  AccountNumber uint32
  RowID         uint32
}

type ListAccountsQuery struct {
  WalletID uint32
  Scope    *KeyScope
  Name     *string
  Page     page.Request[AccountCursor]
}

type ListAddressesQuery struct {
  WalletID    uint32
  Scope       KeyScope
  AccountName string
  Page        page.Request[uint32]
}
```

This reads better than:

```go
page.Options[uint32]{}.WithSize(50).WithCursor(99)
```

and it is easier to debug because you can just print the struct.

**Store interfaces**

The store interfaces should return `page.Result[...]`.

```go
type WalletStore interface {
  ListWallets(ctx context.Context, query ListWalletsQuery) (
    page.Result[WalletInfo, uint32], error,
  )

  IterWallets(ctx context.Context, query ListWalletsQuery) iter.Seq2[WalletInfo, error]
}

type AccountStore interface {
  ListAccounts(ctx context.Context, query ListAccountsQuery) (
    page.Result[AccountInfo, AccountCursor], error,
  )

  IterAccounts(ctx context.Context, query ListAccountsQuery) iter.Seq2[AccountInfo, error]
}

type AddressStore interface {
  ListAddresses(ctx context.Context, query ListAddressesQuery) (
    page.Result[AddressInfo, uint32], error,
  )

  IterAddresses(ctx context.Context, query ListAddressesQuery) iter.Seq2[AddressInfo, error]
}
```

This is the key public API improvement.

**Caller experience**

Current style:

```go
items, err := store.ListAccounts(ctx, q)
if err != nil {
  return err
}

for _, item := range items {
  use(item)
}

if len(items) > 0 {
  q.Page = q.Page.WithCursor(items[len(items)-1].Cursor())
}
```

Redesigned style:

```go
res, err := store.ListAccounts(ctx, q)
if err != nil {
  return err
}

for _, item := range res.Items {
  use(item)
}

if res.Next != nil {
  q.Page.After = res.Next
}
```

That is much easier to use, and much harder to misuse.

The store computes `Next`. The caller just forwards it.

**Why `Result.Next` is such a big improvement**

This removes pagination knowledge from item types.

Right now something like `AccountInfo.Cursor()` leaks internal ordering assumptions:

- imported accounts sort after derived
- imported state is inferred from `Origin`
- `RowID` is part of cursor semantics

That logic should not live in the public `AccountInfo` API at all. It belongs inside `ListAccounts`.

With `page.Result.Next`, you can keep cursor derivation private:

```go
func accountCursorFromInfo(info AccountInfo) AccountCursor {
  return AccountCursor{
    Imported:      info.Origin == ImportedAccount,
    AccountNumber: info.AccountNumber,
    RowID:         info.ID,
  }
}
```

That helper stays private to the store package.

**Store implementation pattern**

For each `List*` method, use this flow:

1. normalize page size
2. fetch `limit + 1` rows
3. convert rows to infos
4. if more than `limit`, compute `Next` from the last retained item and trim
5. return `page.Result`

That removes the need for generic `nextQuery` plumbing and avoids the extra terminal fetch.

**Example: `ListWallets`**

This is the ideal simple case because the cursor is just `wallet.id`.

```go
func (s *SqliteStore) ListWallets(
  ctx context.Context,
  query ListWalletsQuery,
) (page.Result[WalletInfo, uint32], error) {

  limit := query.Page.EffectiveLimit()
  fetchLimit := int64(limit) + 1

  rows, err := s.queries.ListWallets(ctx, sqlc.ListWalletsParams{
    AfterID: nullUint32(query.Page.After),
    Limit:   fetchLimit,
  })
  if err != nil {
    return page.Result[WalletInfo, uint32]{}, fmt.Errorf("list wallets: %w", err)
  }

  items, err := walletInfosFromRows(rows)
  if err != nil {
    return page.Result[WalletInfo, uint32]{}, fmt.Errorf("convert wallets: %w", err)
  }

  var next *uint32
  n := int(limit)
  if len(items) > n {
    cursor := items[n-1].ID
    next = &cursor
    items = items[:n]
  }

  return page.Result[WalletInfo, uint32]{
    Items: items,
    Next:  next,
  }, nil
}
```

The important detail is this line:

```go
cursor := items[n-1].ID
```

When you fetch `limit+1`, the cursor should point to the last returned item, not the discarded extra row.

**Example: `IterWallets`**

With the new `ListWallets` shape, the iterator becomes trivial.

```go
func (s *SqliteStore) IterWallets(
  ctx context.Context,
  query ListWalletsQuery,
) iter.Seq2[WalletInfo, error] {
  return func(yield func(WalletInfo, error) bool) {
    for {
      res, err := s.ListWallets(ctx, query)
      if err != nil {
        var zero WalletInfo
        _ = yield(zero, err)
        return
      }

      for _, item := range res.Items {
        select {
        case <-ctx.Done():
          var zero WalletInfo
          _ = yield(zero, ctx.Err())
          return
        default:
        }

        if !yield(item, nil) {
          return
        }
      }

      if res.Next == nil {
        return
      }

      next := *res.Next
      query.Page.After = &next
    }
  }
}
```

I would handwrite `IterWallets`, `IterAccounts`, and `IterAddresses` instead of keeping a generic `page.Iter`. With only three call sites, that is clearer and easier to maintain.

If you still want a generic helper, it should operate on `page.Result`, not on `(fetchPage, nextQuery)`.

**A better generic iterator, if you keep one**

If you really want a helper, make it work with explicit page metadata.

```go
func Iter[Q, T, C any](
  ctx context.Context,
  query Q,
  fetch func(context.Context, Q) (page.Result[T, C], error),
  withAfter func(Q, *C) Q,
) iter.Seq2[T, error] {
  return func(yield func(T, error) bool) {
    for {
      res, err := fetch(ctx, query)
      if err != nil {
        var zero T
        _ = yield(zero, err)
        return
      }

      for _, item := range res.Items {
        select {
        case <-ctx.Done():
          var zero T
          _ = yield(zero, ctx.Err())
          return
        default:
        }

        if !yield(item, nil) {
          return
        }
      }

      if res.Next == nil {
        return
      }

      query = withAfter(query, res.Next)
    }
  }
}
```

That is much better than the current `nextQuery func(Q, []R) Q` design.

But again, I would still prefer handwritten iterators here.

**SQL design**

There are two parts here:

- simple cursors: wallets, addresses
- composite cursors: accounts

The API redesign should happen regardless of whether all SQL is unified.

**Wallets and addresses: use one query**

These are the easy cases. Use one query with a nullable cursor and `LIMIT n+1`.

Pseudo-SQL:

```sql
SELECT ...
FROM wallets w
WHERE (:after_id IS NULL OR w.id > :after_id)
ORDER BY w.id
LIMIT :limit_plus_one;
```

and similarly for addresses:

```sql
SELECT ...
FROM addresses a
WHERE
   a.account_id = :account_id
   AND (:after_id IS NULL OR a.id > :after_id)
ORDER BY a.id
LIMIT :limit_plus_one;
```

This gives you:

- one SQL query
- one sqlc row type
- one sqlc params type
- much less Go boilerplate

**Accounts: two acceptable options**

Accounts are more complicated because the sort key is composite:

- derived accounts first, ordered by `(account_number, id)`
- imported accounts after, ordered by `id`

You have two reasonable choices.

**Option A: one unified query**

Use a single query with nullable cursor fields.

Pseudo-SQL:

```sql
SELECT ...
FROM accounts a
WHERE
   a.wallet_id = :wallet_id
   AND (:scope_purpose IS NULL OR a.purpose = :scope_purpose)
   AND (:scope_coin_type IS NULL OR a.coin_type = :scope_coin_type)
   AND (:name IS NULL OR a.account_name = :name)
   AND (
     :after_id IS NULL
     OR (
       (
         NOT :after_imported
         AND a.account_number IS NOT NULL
         AND (
           a.account_number > :after_account_number
           OR (
             a.account_number = :after_account_number
             AND a.id > :after_id
           )
         )
       )
       OR (
         a.account_number IS NULL
         AND (
           NOT :after_imported
           OR a.id > :after_id
         )
       )
     )
   )
ORDER BY a.account_number NULLS LAST, a.id
LIMIT :limit_plus_one;
```

That is still complicated SQL, but the complexity is in one place instead of duplicated across first-page and next-page versions.

This is the cleanest overall design if the query planner behaves well.

**Option B: keep split SQL for accounts only**

If `EXPLAIN` shows that the unified query produces worse plans, keep the first-page / next-page split for accounts only.

That is fine.

The important point is: the Go API does not depend on SQL unification.

You can still ship the better `page.Request` / `page.Result` design even if accounts continue using two separate SQL statements internally.

That is probably the safest rollout path.

**Query planner caution**

This is the one place where I would be pragmatic instead of ideological.

For wallets and addresses, one nullable-cursor query is obviously worth it.

For accounts, I would require plan verification:

- PostgreSQL: `EXPLAIN (ANALYZE, BUFFERS)` for first-page and next-page cases
- SQLite: `EXPLAIN QUERY PLAN`
- test with realistic data and the actual ordering/index strategy

If the unified account query regresses, do not force it. Keep two queries there.

**Optional tiny helper for `limit+1` trimming**

If the trimming logic repeats too much, add one very small helper.

```go
func BuildResult[T any, C any](
  items []T,
  limit uint32,
  cursorOf func(T) C,
) page.Result[T, C] {
  n := int(limit)
  if len(items) <= n {
    return page.Result[T, C]{Items: items}
  }

  cursor := cursorOf(items[n-1])
  return page.Result[T, C]{
    Items: items[:n],
    Next:  &cursor,
  }
}
```

That is the kind of generic helper that helps without taking over the design.

This is very different from the current multi-layer generic abstraction.

**What the API should not do**

I would explicitly avoid these patterns:

- do not make callers compute the next cursor from the last item
- do not put `Cursor()` methods on public info types
- do not hide plain data behind builder methods like `WithSize().WithCursor()`
- do not introduce a generic pagination helper that is more complicated than the call sites
- do not require an extra empty fetch just to detect completion
- do not let SQLC-generated first/next row types dictate the public API

**Testing strategy**

Once redesigned, the tests should validate the actual contract.

At the page type level:

- `Limit == 0` uses `DefaultLimit`
- `Limit > MaxLimit` clamps to `MaxLimit`

At the store level:

- zero-value query returns the first page with `DefaultLimit`
- explicit `Limit` is honored
- `Next == nil` when the result fits in one page
- `Next != nil` when more items exist
- following `Next` yields the remaining items without duplicates or gaps
- accounts cross the derived/imported boundary correctly
- scope and name filters still page correctly
- iteration matches concatenated page traversal

At the iterator level:

- stop on consumer break
- surface errors once
- if you support it, stop promptly on mid-page cancellation
- do not do the old extra empty fetch

That last one is important: with `LIMIT+1`, exact-multiple traversals should not need one final empty-page round trip.

**Migration plan**

I would do this in stages.

1. Introduce `page.Request` and `page.Result`
2. Convert `ListWallets` first, because it is the simplest case
3. Convert `IterWallets` to use the new result type
4. Repeat for addresses
5. Convert accounts last
6. Remove `AccountInfo.Cursor()` and similar cursor-derivation helpers from public types
7. Delete `page.Options` and the current generic `page.Iter`
8. Optionally unify SQL where planner behavior is acceptable

That gives you a clean incremental path.

**Recommended final shape**

If I had to boil it down to one target design, it would be this:

```go
package page

type Request[C any] struct {
  Limit uint32
  After *C
}

type Result[T any, C any] struct {
  Items []T
  Next  *C
}
```

```go
type AccountStore interface {
  ListAccounts(ctx context.Context, query ListAccountsQuery) (
    page.Result[AccountInfo, AccountCursor], error,
  )
  IterAccounts(ctx context.Context, query ListAccountsQuery) iter.Seq2[AccountInfo, error]
}
```

```go
res, err := store.ListAccounts(ctx, query)
if err != nil {
  return err
}

for _, item := range res.Items {
  ...
}

if res.Next != nil {
  query.Page.After = res.Next
}
```

That is the core redesign.

It fixes the ergonomics, keeps pagination ownership in the right layer, and lets you simplify the implementation without over-abstracting it.

GustavoStingelin:
Regarding returning a page struct with the list inside, that was my initial approach. However, after looking at LND, I noticed it returns only the slice, with the cursor handled externally. Since termination can be inferred from an empty list, this seemed simpler at first.

That said, I missed an important point you raised: this approach shifts the responsibility to the caller to understand how to assemble the cursor correctly. Because of that, I am now including the page struct to make the contract more explicit and reduce the risk of misuse.

saubyk:
is this pr still in draft?

GustavoStingelin:
> is this pr still in draft?

yes, will be ready soon

GustavoStingelin:
I ran benchmarks comparing the split queries, with and without cursor, against the unified approach. The results show the split version is generally about 7% faster. For PostgreSQL, the results are mixed, sometimes slightly faster, sometimes slightly slower, within a similar range.

Full benchmark details are available here: https://github.com/GustavoStingelin/btcwallet/blob/6eb7ab030a1899aba0a92e5f12ea2c8ad1763caf/wallet/internal/db/itest/PAGINATION_REPORT.md

Given that the performance difference is relatively small, I agree it is worth simplifying the implementation by using a single conditional cursored query, and I will do this.

wallet/internal/db/migrations/sqlite/000006_addresses.up.sql
```diff
@@ -74,6 +74,10 @@ ON addresses (account_id, script_pub_key);
 -- Used by GetAddressByScriptPubKey.
 CREATE INDEX idx_addresses_script_pub_key ON addresses (script_pub_key);
 
+-- Index on (account_id, id) for efficient pagination of addresses by account.
+-- Used by GetAddressesByAccount for cursor-based pagination.
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/queries/sqlite/wallets.sql
```diff
@@ -80,7 +83,12 @@ FROM wallets AS w
 LEFT JOIN wallet_sync_states AS s ON w.id = s.wallet_id
 LEFT JOIN blocks AS b_synced ON s.synced_height = b_synced.block_height
 LEFT JOIN blocks AS b_birthday ON s.birthday_height = b_birthday.block_height
-ORDER BY w.id;
+WHERE (
+    sqlc.narg('cursor_id') IS NULL
+    OR w.id > sqlc.narg('cursor_id')
+)
+ORDER BY w.id
+LIMIT sqlc.arg(page_limit);
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/itest/wallet_store_test.go
```diff
@@ -152,9 +154,13 @@ func TestListWallets(t *testing.T) {
...
 	// Initially empty.
-	wallets, err := store.ListWallets(t.Context())
-	require.NoError(t, err)
-	require.Empty(t, wallets)
+	query := db.ListWalletsQuery{
+		Page: page.Request[uint32]{}.WithSize(2),
+	}
+
+	for w, err := range store.IterWallets(t.Context(), query) {
+		require.Failf(t, "unexpected wallet", "wallet=%v err=%v", w, err)
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/queries/postgres/wallets.sql
```diff
@@ -80,7 +83,12 @@ FROM wallets AS w
 LEFT JOIN wallet_sync_states AS s ON w.id = s.wallet_id
 LEFT JOIN blocks AS b_synced ON s.synced_height = b_synced.block_height
 LEFT JOIN blocks AS b_birthday ON s.birthday_height = b_birthday.block_height
-ORDER BY w.id;
+WHERE (
+    sqlc.narg('cursor_id')::BIGINT IS NULL
+    OR w.id > sqlc.narg('cursor_id')::BIGINT
+)
+ORDER BY w.id
+LIMIT sqlc.arg(page_limit)::BIGINT;
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/migrations/postgres/000006_addresses.up.sql
```diff
@@ -76,6 +76,10 @@ ON addresses (account_id, script_pub_key);
 -- Used by GetAddressByScriptPubKey.
 CREATE INDEX idx_addresses_script_pub_key ON addresses (script_pub_key);
 
+-- Index on (account_id, id) for efficient pagination of addresses by account.
+-- Used by GetAddressesByAccount for cursor-based pagination.
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/itest/wallet_store_test.go
```diff
@@ -152,9 +154,14 @@ func TestListWallets(t *testing.T) {
...
-	require.NoError(t, err)
-	require.Empty(t, wallets)
+	query := db.ListWalletsQuery{
+		Page: page.Request[uint32]{}.WithSize(2),
+	}
+
+	for w, err := range store.IterWallets(t.Context(), query) {
+		require.NoError(t, err)
+		require.Failf(t, "unexpected wallet", "wallet=%v", w)
+	}
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/interface.go
GustavoStingelin: It will block the use case where we need to expose the cursor via rpc

wallet/internal/db/page/iter_test.go
GustavoStingelin: added

wallet/internal/db/queries/postgres/accounts.sql
GustavoStingelin: added

wallet/internal/db/addresses_pg.go
GustavoStingelin: added

wallet/internal/db/page/iter.go
GustavoStingelin: removed, caller should handle the ctx.Done()

GustavoStingelin: btw, the first version was without this, and in a review, the LLM said that it should have this, and I did not review carefully.

wallet/internal/db/page/iter.go
GustavoStingelin: With the refactor, Iter changed a bit, take a look at the new version

wallet/internal/db/page/options.go
GustavoStingelin: Yeah, this slightly adjusts the API behavior. If the caller prefers not to handle pagination explicitly, they can use the Iter* methods instead.

wallet/internal/db/itest/address_store_test.go
GustavoStingelin: line refactored, btw, think we could activate more aggressive linters on tests later

wallet/internal/db/page/iter.go
GustavoStingelin: removed to be handled by the caller in the yield function if wanted.

wallet/internal/db/page/options.go
GustavoStingelin: yes... turned into a private field

wallet/internal/db/page/options.go
GustavoStingelin: This struct was refactored, but since it uses value receivers rather than pointer receivers, a copy is already being made implicitly.

wallet/internal/db/page/options_test.go
GustavoStingelin: Turned the cursor unexported

GustavoStingelin:
Think it's finally ready for review again.

### Main changes

* Replaced the split cursor queries with a unified approach:
  [https://github.com/btcsuite/btcwallet/pull/1189#issuecomment-4156521943](https://github.com/btcsuite/btcwallet/pull/1189#issuecomment-4156521943)

* Introduced `page.Response` as the return type for paginated queries instead of plain slices:
  [https://github.com/btcsuite/btcwallet/pull/1189#issuecomment-4098281345](https://github.com/btcsuite/btcwallet/pull/1189#issuecomment-4098281345)

* Pagination strategy:

  * Implemented the `pageSize + 1` approach to detect if there is a next page, which avoids an extra round trip and is more efficient
  * Kept the previous approach, which performs an additional round trip but provides stronger completeness guarantees
  * The choice is configurable via `page.Request`
   * The trade off between the extra round trip and completeness guarantees is explained in the docs for the structs and functions.

GustavoStingelin:
also, removed the pagination of accounts in this PR, will open it later to be easier to review

GustavoStingelin:
reduced from 5k lines to 2.8k

wallet/internal/db/page/iter.go
```diff
@@ -0,0 +1,84 @@
...
+// The entity parameter is a human-readable resource name (e.g., "account",
+// "transaction"), used only in error messages.
+//
+// The fetchPage function must honour ctx and return an error if cancelled.
+//
+// Iter stops when HasMore is false, when yield returns false (caller break),
+// or when fetchPage returns an error. Errors are yielded as the second value;
+// the iterator terminates after the first error. Cancellation is only observed
+// between page fetches, and only if fetchPage checks ctx.
+func Iter[Query, Item, Cursor any](ctx context.Context, query Query,
```
yyforyongyu, line 13: This `entity` parameter seems to exist only to decorate the invariant-violation error below. Since every current call site passes a string literal, would it be simpler to drop it from `Iter` and just return `errPageNoCursor` (or a generic wrapper) here? That would shrink the helper API a bit without losing much diagnostic value.

wallet/internal/db/addresses_pg.go
```diff
@@ -298,3 +316,57 @@ func pgAddressRowToInfo[T pgAddressInfoRow](row T) (*AddressInfo, error) {
 
 	return info, nil
 }
+
+// pgAddressListPagedQueries holds the queries for paginated address listing
+// in PostgreSQL.
+type pgAddressListPagedQueries struct {
```
yyforyongyu: Do we need a wrapper struct here? `pgAddressListPagedQueries` currently has a single `byAccount` method and is only instantiated once in `ListAddresses`. A plain helper like `pgListAddressesByAccount(ctx, s.queries, query)` would be a little more direct. Same thought applies to the SQLite variant.

wallet/internal/db/page/request.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// MaxPageSize is the maximum number of items that can be returned in a
+	// page. SQL queries that implement early-exhaustion lookahead may receive
+	// MaxPageSize+1 as the page limit.
+	MaxPageSize = 1000
+)
+
+// Request holds the parameters for a paginated list query. The zero value is
+// valid and requests the first page at DefaultPageSize. All With* methods
+// return a modified copy (value-type immutability).
```
yyforyongyu: Minor doc nit: `Request` carries `cursor *C`, so `WithSize` and `WithEarlyExhaustion` are only doing a shallow copy. The behavior here is still safe, but `value-type immutability` reads a bit stronger than what the struct layout literally guarantees. Maybe soften this to `return a modified copy` or similar.

wallet/internal/db/addresses_common.go
```diff
@@ -359,36 +359,15 @@ func getAddress[T any, Args any](ctx context.Context,
 	return nil, ErrAddressNotFound
 }
 
-// listAddresses is a generic helper that retrieves addresses using the provided
-// lister function and converts the results to AddressInfo structs.
-func listAddresses[T any, Args any](ctx context.Context,
-	lister func(context.Context, Args) ([]T, error), args Args,
-	toInfo func(T) (*AddressInfo, error)) ([]AddressInfo, error) {
+// nextListAddressesQuery returns the query for the next page by advancing the
```
yyforyongyu: Nit: this comment describes the current call pattern of `page.Iter` rather than what `nextListAddressesQuery` itself does. I think it would read a little more cleanly if it just said that the helper advances the cursor for the next page fetch.

wallet/internal/db/page/iter.go
```diff
@@ -0,0 +1,84 @@
...
+				return
+			}
+
+			// HasMore=true with LastCursor=nil violates the Result invariant
+			// and indicates a bug in the fetchPage implementation. This is
+			// treated as a fatal invariant violation rather than a silent
+			// error, ensuring the caller is alerted to the programming mistake
+			// in the fetch implementation.
+			if result.LastCursor == nil {
+				yield(zero, fmt.Errorf("page for %s: %w", entity,
```
yyforyongyu: i think we need to either wrap more context here, or define a new error for this special case where we have a conflict here, as the table is in an inconsitent state.

wallet/internal/db/itest/account_store_test.go
```diff
@@ -182,7 +182,13 @@ func TestCreateDerivedAccountConcurrent(t *testing.T) {
 	scope := db.KeyScopeBIP0084
 
 	const workers = 20
-	results := make([]uint32, workers)
+
```
yyforyongyu, line 185: since we focus on address/wallets pagination here, the last two commits can be moved to a new PR and merged as is

wallet/internal/db/page/request.go
```diff
@@ -0,0 +1,179 @@
...
+// page operations.
+type Request[C any] struct {
+	// size is the maximum number of items to return per page.
+	size uint32
+
+	// cursor is the pagination cursor from the previous page.
+	// Nil means the first page.
+	cursor *C
+
+	// earlyExhaustion controls whether LastCursor and HasMore use deferred
```
yyforyongyu: I don't quite get this design - it seems we have dual exhaustion modes, which makes the pkg heavier than it needs?

wallet/internal/db/page/iter.go
```diff
@@ -0,0 +1,84 @@
...
+// between page fetches, and only if fetchPage checks ctx.
+func Iter[Query, Item, Cursor any](ctx context.Context, query Query,
+	entity string,
+	fetchPage func(context.Context, Query) (Result[Item, Cursor], error),
+	setCursor func(Query, Cursor) Query) iter.Seq2[Item, error] {
+
+	return func(yield func(Item, error) bool) {
+		var zero Item
+
+		for {
```
yyforyongyu, line 20: we should check the cancellation of the `ctx`.

wallet/internal/db/page/result.go
```diff
@@ -0,0 +1,28 @@
+package page
+
+// Result holds one page of items returned by a paginated list query. It
+// maintains an invariant: LastCursor is nil if and only if Items is empty. The
+// meaning of HasMore depends on the Request's early-exhaustion mode; see
+// Request.WithEarlyExhaustion for details.
+type Result[T any, C any] struct {
```
yyforyongyu: Following the above design, we can have sth like,
```go
type Result[T any, C any] struct {
	// Items contains the results for this page.
	Items []T

	// Next is the cursor for the next page.
	Next C

	// HasNext reports whether another page may be fetched.
	HasNext bool
}
```

wallet/internal/db/page/iter.go
yyforyongyu: actually i think based on the improved design we can eliminate this error completely

wallet/internal/db/page/iter.go
```diff
@@ -0,0 +1,84 @@
...
+			}
+
+			// Advance the query cursor only after the current page is fully
+			// yielded to the caller, and we've verified a valid cursor exists
+			// for the next fetch. This ensures the query advances one page at
+			// a time in lockstep with item consumption.
+			query = setCursor(query, *result.LastCursor)
+		}
+	}
+}
```
yyforyongyu, line 55: I've tried a simplified version [here](https://github.com/yyforyongyu/btcwallet/tree/cr-1189-260403-pagination-final-design/wallet/internal/db/page) to help me understand the shape.Given that this is a brand-new foundational package and only wallets/addresses consume it so far, I think we can simplify it more, like using the classical `LIMIT n+1` model with plain value-semantic types like `Request{Limit, After, HasAfter}` and `Result{Items, Next, HasNext}`.

This way we can removes `earlyExhaustion`, `LastCursor`, pointer optionality, `QueryLimit()`, and the `HasMore=true` but nil cursor invariant/error path in Iter, while keeping cursor derivation inside the store. The resulting API ended up materially smaller and easier to reason about when I tried it locally.

yyforyongyu (changes requested):
Much cleaner!

I still think the pkg leaks abstraction in both directions. `entity` and `withAfter` bring store-specific details into `page`: `entity` is just a resource label for error messages, and `withAfter` exists because each caller has its own query type that needs to be rebuilt with a new cursor. Then the result side pushes those mechanics back out again via `LastCursor` and `HasMore`, which means one thing in deferred mode and another in early-exhaustion mode

As mentioned before, when designing pagination, we should think in terms of `Next`, not `Last`. The caller should think "what do I pass to fetch the next page?", not "what was the last cursor on the current page?"

I played out the design [here](https://github.com/yyforyongyu/btcwallet/tree/cr-1189-260403-pagination-final-design/wallet/internal/db/page) to strengthen the final shape, and I think it can be further simplified and improved like,
```go
func Iter[Q, T, C any](ctx context.Context, query Q,
	fetch func(context.Context, Q) (Result[T, C], error),
	withAfter func(Q, C) Q) iter.Seq2[T, error] {

	...
}

type Request[C any] struct {
	// Limit is the requested number of items to return.
	Limit uint32

	// After is the cursor from the previous page.
	After C

	// HasAfter reports whether After should be applied.
	HasAfter bool
}

type Result[T any, C any] struct {
	// Items contains the results for this page.
	Items []T

	// Next is the cursor for the next page.
	Next C

	// HasNext reports whether another page may be fetched.
	HasNext bool
}
```
Which gives us a shape that's easier to reason about.

wallet/internal/db/interface.go
```diff
@@ -139,10 +142,15 @@ type WalletStore interface {
...
-	ListWallets(ctx context.Context) ([]WalletInfo, error)
+	// ListWallets returns one page of wallets for the given query, including a
+	// next-cursor for the following page.
+	ListWallets(ctx context.Context, query ListWalletsQuery) (
+		page.Result[WalletInfo, uint32], error)
+
+	// IterWallets returns an iterator that fetches pages transparently and
+	// yields wallets one by one until exhaustion or error.
+	IterWallets(ctx context.Context,
+		query ListWalletsQuery) iter.Seq2[WalletInfo, error]
```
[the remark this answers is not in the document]
GustavoStingelin: desc updated

wallet/internal/db/addresses_pg.go
```diff
@@ -298,3 +316,57 @@ func pgAddressRowToInfo[T pgAddressInfoRow](row T) (*AddressInfo, error) {
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list addresses by account: %w", err)
+	}
+
+	items := make([]AddressInfo, len(rows))
+	for i, row := range rows {
+		item, err := pgAddressRowToInfo(row)
+		if err != nil {
+			return nil, fmt.Errorf("map address row: %w", err)
```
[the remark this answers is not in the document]
GustavoStingelin: applied

wallet/internal/db/addresses_sqlite.go
```diff
@@ -291,3 +310,54 @@ func sqliteAddressRowToInfo[T sqliteAddressInfoRow](row T) (*AddressInfo,
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list addresses by account: %w", err)
+	}
+
+	items := make([]AddressInfo, len(rows))
+	for i, row := range rows {
+		item, err := sqliteAddressRowToInfo(row)
+		if err != nil {
+			return nil, fmt.Errorf("map address row: %w", err)
```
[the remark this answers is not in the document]
GustavoStingelin: applied

wallet/internal/db/page/request.go
```diff
@@ -0,0 +1,113 @@
...
+func (r Request[Cursor]) WithLimit(limit uint32) Request[Cursor] {
+	r.limit = limit
+
+	return r
+}
+
+// After returns the pagination after from the previous page.
+// A nil return value means the first page is being requested.
+func (r Request[Cursor]) After() *Cursor {
+	return r.after
```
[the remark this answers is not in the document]
GustavoStingelin: applied

wallet/internal/db/page/iter.go
GustavoStingelin: added

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: moved

wallet/internal/db/page/iter.go
GustavoStingelin: yeah, removed

wallet/internal/db/addresses_common.go
GustavoStingelin: updated

wallet/internal/db/page/request.go
GustavoStingelin: updated

wallet/internal/db/addresses_pg.go
GustavoStingelin: removed, much cleaner

wallet/internal/db/page/iter.go
GustavoStingelin: removed since we do not have the invariant anymore

wallet/internal/db/page/request.go
GustavoStingelin: Basically, the approach of using `limit + 1` to detect early whether there is a next page has a drawback. Consider the example below (early exhaustion):

1. A → executes the query and receives `< limit + 1` rows  
2. A → starts processing the page  
3. B → inserts a new row  
4. A → finishes processing the page  
5. A → does not see the inserted row, because it already "knows" this was the last page  

The previous approach (deferred exhaustion) solves this by adding an extra fetch:

1. A → executes the query and receives a page  
2. A → starts processing the page  
3. B → inserts a new row  
4. A → finishes processing the page  
5. A → executes one more query to ensure all results were processed  
6. A → discovers the new row and processes it  
7. A → executes one more query to ensure all results were processed  
8. A → receives zero rows and now has a guarantee that all results were processed  

That said, the current refactor removes the deferred exhaustion approach. We can reintroduce it later if this behavior becomes a real issue. The tricky part is that this is hard to predict, since it depends on concurrent operations.

wallet/internal/db/page/iter.go
GustavoStingelin: Applied some of it, but I kept the Request internals private. My main concern is that, if we make limit public, we lose the compile time guarantee that callers will only use a normalized value. Someone could read it directly and pass a raw zero value into a query in a way that is not intended.

That said, zero is also a valid use case for callers that do not want to set a custom config, so I think the API still needs to preserve that behavior clearly and safely.

I made a few changes based on the comments, for example, I removed HasNext and left the nullable cursor as the only signal, since that already seems enough to determine whether there is another page.

I did not add value semantics. With the field unexported, that would likely require two getters, one for the cursor itself and another to indicate whether a cursor is present, and I was not sure that tradeoff was worth it.

What do you think about this version given the comments?

wallet/internal/db/page/request.go
```diff
@@ -0,0 +1,119 @@
...
+type Request[Cursor any] struct {
+	// limit is the maximum number of items to return per page.
+	limit uint32
+
+	// after is the pagination cursor that marks where the next page
+	// starts after. Nil means the first page.
+	after *Cursor
+}
+
+// QueryLimit returns the number of rows the SQL query should fetch. Queries
```
yyforyongyu, line 31: `QueryLimit()` still exposes the `LIMIT n+1` lookahead strategy from the store layer through the shared `page` API. Since callers only care about `limit` and `after`, would it be cleaner to keep query-limit calculation inside the store-specific param builders and let `page` model only the caller-facing pagination contract?

wallet/internal/db/page/iter.go
```diff
@@ -0,0 +1,55 @@
...
+	"context"
+	"iter"
+)
+
+// Iter iterates through paginated results by repeatedly fetching pages and
+// yielding items until exhaustion, caller break, error, or cancellation.
+//
+// Errors are yielded as the second value in the iterator pair. Callers must
+// check the error on every iteration before using the yielded item.
+func Iter[Query, Item, Cursor any](ctx context.Context, query Query,
```
yyforyongyu, line 13: Related non-blocking thought: `withAfter` is reasonable, but it still means the generic helper knows each caller has some query struct that needs cursor plumbing. If you want the cleanest end-state, I think there is still an argument for keeping the shared `page` package focused on request/result semantics and leaving more of this wiring in store code.

wallet/internal/db/page/request.go
yyforyongyu: This is a classic moving-dataset read race. I don’t think pagination mode can solve it without DB transaction/snapshot semantics. Deferred exhaustion only shifts the race window; it doesn’t eliminate it, since another insertion can still happen after the final read.

wallet/internal/db/page/iter.go
yyforyongyu: > we lose the compile time guarantee that callers will only use a normalized value. 

Can you elaborate a bit? What is the compile-time guarantee here?

And if it's 0 we just use the default?

Guess this is why I still prefer separating the caller-facing pagination contract from the SQL fetch strategy. `QueryLimit()` pushes the `LIMIT n+1` implementation detail into the shared page API, while the thing callers actually care about is just `limit + after`.

yyforyongyu (approved):
Much cleaner! The PR is largely good to go💯 

There is still room for improvement, tho, given that this is a low-level API and is likely to be used for other projects, we need to pay more attention here. Will address it in a follow-up PR instead.

wallet/internal/db/page/iter.go
GustavoStingelin: By “compile-time guarantee” I was referring to encapsulation of the raw field: because `limit` is unexported, code outside `page` cannot directly observe or reuse an unnormalized value. The public API keeps `0` as a valid caller input and consistently interprets it as the default, so normalization stays inside the package instead of being pushed out to every caller site.

If `limit` were public, each caller would need to remember to validate or normalize it before using it, which spreads that responsibility across the codebase. At that point, `Request` is closer to a passive data carrier than a type that owns its own pagination invariants.

wallet/internal/db/page/iter.go
GustavoStingelin: hmm, yeah, think it could be better moving to another package

wallet/internal/db/page/request.go
GustavoStingelin: I agree this would make the API look cleaner, but I also think it may become more error-prone in practice. Right now the lookahead calculation is centralized, and `BuildResult` depends on the same normalization rule. If we move that logic into each caller/store builder, we make the surface API clearer, but we also spread that responsibility across multiple call sites.

PR #1192, opened 2026-03-24
wallet: fix pg itest connection exhaustion under parallelism

GustavoStingelin:
This fixes a bug that had been haunting me for a while in the integration tests.

#### Error
The random error:
```
failed to connect to host=localhost user=postgres database=postgres: server error (FATAL: sorry, too many clients already (SQLSTATE 53300))
```

#### Root cause
We were opening more PostgreSQL connections than the limit.

This happened because:
- Calling `t.Parallel()` inside a `t.Run()` releases the parent goroutine early, so more top-level tests start before the parent store is closed.
- Each `NewTestStore` creates its own database with a pool of up to 25 connections.
- `max_connections` was fixed and did not match the actual connection budget. The semaphore only limited database creation, but not the total number of open connections.

#### Changes
- Store per subtest — each parallel subtest now calls NewTestStore(t) itself so the store lifecycle is tied to the subtest's parallel slot.
- Sized max_connections to parallelism — replaced the fixed limit with testParallelism() * DefaultMaxConnections, matching the actual number of concurrent stores the test runner can open.
- Removed redundant caching — pgMaxConnections() and its package variable pgMaxConns were called once inside a sync.Once; inlined the computation at the call site.
- Table-driven error tests use plain struct params with WalletID injected after wallet creation. Setup funcs in TestGetAddress and TestListAddresses accept store and walletID parameters so each subtest operates on its own isolated database.

#### Result
- No more connection exhaustion errors.
- Tests still run in parallel.

### DIFF
--- a/wallet/internal/db/itest/account_store_test.go
+++ b/wallet/internal/db/itest/account_store_test.go
@@ -51,10 +51,6 @@ func TestCreateAccounts(t *testing.T) {
 func TestCreateDerivedAccountErrors(t *testing.T) {
 	t.Parallel()
 
-	store := NewTestStore(t)
-
-	walletID := newWallet(t, store, "wallet-create-derived-account-errors")
-
 	tests := []struct {
 		name    string
 		params  db.CreateDerivedAccountParams
@@ -63,18 +59,16 @@ func TestCreateDerivedAccountErrors(t *testing.T) {
 		{
 			name: "missing name",
 			params: db.CreateDerivedAccountParams{
-				WalletID: walletID,
-				Scope:    db.KeyScopeBIP0084,
-				Name:     "",
+				Scope: db.KeyScopeBIP0084,
+				Name:  "",
 			},
 			wantErr: db.ErrMissingAccountName,
 		},
 		{
 			name: "unknown scope",
 			params: db.CreateDerivedAccountParams{
-				WalletID: walletID,
-				Scope:    db.KeyScope{Purpose: 999, Coin: 999},
-				Name:     "unknown-scope-account",
+				Scope: db.KeyScope{Purpose: 999, Coin: 999},
+				Name:  "unknown-scope-account",
 			},
 			wantErr: db.ErrUnknownKeyScope,
 		},
@@ -83,6 +77,10 @@ func TestCreateDerivedAccountErrors(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
+
+			store := NewTestStore(t)
+			walletID := newWallet(t, store, tc.name+"-wallet")
+			tc.params.WalletID = walletID
 
 			info, err := store.CreateDerivedAccount(t.Context(), tc.params)
 			require.ErrorIs(t, err, tc.wantErr)
@@ -222,10 +220,6 @@ func TestCreateDerivedAccountConcurrent(t *testing.T) {
 func TestCreateImportedAccountErrors(t *testing.T) {
 	t.Parallel()
 
-	store := NewTestStore(t)
-
-	walletID := newWallet(t, store, "wallet-create-imported-account-errors")
-
 	tests := []struct {
 		name    string
 		params  db.CreateImportedAccountParams
@@ -234,7 +228,6 @@ func TestCreateImportedAccountErrors(t *testing.T) {
 		{
 			name: "missing name",
 			params: db.CreateImportedAccountParams{
-				WalletID:           walletID,
 				Name:               "",
 				Scope:              db.KeyScopeBIP0084,
 				EncryptedPublicKey: RandomBytes(32),
@@ -244,7 +237,6 @@ func TestCreateImportedAccountErrors(t *testing.T) {
 		{
 			name: "missing public key",
 			params: db.CreateImportedAccountParams{
-				WalletID:           walletID,
 				Name:               "missing-pubkey",
 				Scope:              db.KeyScopeBIP0084,
 				EncryptedPublicKey: nil,
@@ -254,7 +246,6 @@ func TestCreateImportedAccountErrors(t *testing.T) {
 		{
 			name: "unknown scope",
 			params: db.CreateImportedAccountParams{
-				WalletID:           walletID,
 				Name:               "unknown-scope",
 				Scope:              db.KeyScope{Purpose: 999, Coin: 999},
 				EncryptedPublicKey: RandomBytes(32),
@@ -266,6 +257,10 @@ func TestCreateImportedAccountErrors(t *testing.T) {
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
+
+			store := NewTestStore(t)
+			walletID := newWallet(t, store, tc.name+"-wallet")
+			tc.params.WalletID = walletID
 
 			props, err := store.CreateImportedAccount(t.Context(), tc.params)
 			require.ErrorIs(t, err, tc.wantErr)
@@ -345,16 +340,14 @@ func TestGetAccount(t *testing.T) {
 func TestGetAccountNotFound(t *testing.T) {
 	t.Parallel()
 
-	store := NewTestStore(t)
-
-	walletID := newWallet(t, store, "wallet-get-account-not-found")
-
-	createAllAccounts(t, store, walletID)
-
 	scope := db.KeyScopeBIP0084
 
 	t.Run("by name", func(t *testing.T) {
 		t.Parallel()
+
+		store := NewTestStore(t)
+		walletID := newWallet(t, store, "wallet-get-account-not-found-name")
+		createAllAccounts(t, store, walletID)
 
 		query := getAccountQueryByName(walletID, scope, "non-existent")
 		info, err := store.GetAccount(t.Context(), query)
@@ -364,6 +357,10 @@ func TestGetAccountNotFound(t *testing.T) {
 
 	t.Run("by number", func(t *testing.T) {
 		t.Parallel()
+
+		store := NewTestStore(t)
+		walletID := newWallet(t, store, "wallet-get-account-not-found-number")
+		createAllAccounts(t, store, walletID)
 
 		query := getAccountQueryByNumber(walletID, scope, 99999)
 		info, err := store.GetAccount(t.Context(), query)
@@ -380,14 +377,12 @@ func TestListAccounts(t *testing.T) {
 	// Ensure that has at least 3 accounts to be tested.
 	require.GreaterOrEqual(t, len(AllAccountCases), 3)
 
-	store := NewTestStore(t)
-
-	walletID := newWallet(t, store, "wallet-list-accounts")
-
-	createAllAccounts(t, store, walletID)
-
 	t.Run("all accounts", func(t *testing.T) {
 		t.Parallel()
+
+		store := NewTestStore(t)
+		walletID := newWallet(t, store, "wallet-list-accounts-all")
+		createAllAccounts(t, store, walletID)
 
 		query := db.ListAccountsQuery{WalletID: walletID}
 		accounts, err := store.ListAccounts(t.Context(), query)
@@ -401,6 +396,10 @@ func TestListAccounts(t *testing.T) {
 
 	t.Run("filter by scope", func(t *testing.T) {
 		t.Parallel()
+
+		store := NewTestStore(t)
+		walletID := newWallet(t, store, "wallet-list-accounts-scope")
+		createAllAccounts(t, store, walletID)
 
 		scope := db.KeyScopeBIP0084
 		query := db.ListAccountsQuery{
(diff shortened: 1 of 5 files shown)


wallet/internal/db/itest/pg_test.go
```diff
@@ -22,8 +24,24 @@ import (
...
+	// entire test suite run.
+	storesTotal atomic.Int64
+
+	// pgMaxConns sets the max_connections for the test PostgreSQL container.
+	//
+	// Tests may run in parallel, and each test can open up to
+	// db.DefaultMaxConnections connections. We size the container limit as
+	// NumCPU * DefaultMaxConnections to match the expected level of parallelism
+	// and avoid connection exhaustion during test execution.
+	pgMaxConns = runtime.NumCPU() * db.DefaultMaxConnections
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/itest/pg_test.go
```diff
@@ -61,6 +98,14 @@ func TestMain(m *testing.M) {
 		}
 	}
 
+	maxStores := int64(runtime.NumCPU())
+	if storesPeak.Load() > maxStores {
+		debugLogf(nil, "WARNING: peak open stores (%d) exceeded safe limit "+
+			"(%d); each store uses up to %d connections, container "+
+			"max_connections=%d\n", storesPeak.Load(), maxStores,
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/itest/pg_test.go
```diff
@@ -22,8 +24,24 @@ import (
...
+	// storesOpen tracks the number of currently open test stores.
+	storesOpen atomic.Int64
+
+	// storesPeak tracks the highest observed concurrent open store count across
+	// the entire test suite run.
+	storesPeak atomic.Int64
+
+	// storesTotal tracks the cumulative number of stores opened across the
+	// entire test suite run.
+	storesTotal atomic.Int64
```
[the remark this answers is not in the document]
GustavoStingelin: moved the logic after store creation to avoid that issue.

wallet/internal/db/itest/pg_test.go
```diff
@@ -190,8 +237,24 @@ func NewTestStore(t *testing.T) *db.PostgresStore {
...
+
+	debugLogf(t, "STORE OPEN: %s | open=%d peak=%d total=%d", t.Name(),
+		open, storesPeak.Load(), storesTotal.Load())
+
 	t.Cleanup(func() {
 		_ = store.Close()
+
+		cur := storesOpen.Add(-1)
+		debugLogf(t, "STORE CLOSE: %s | open=%d peak=%d total=%d",
+			t.Name(), cur, storesPeak.Load(), storesTotal.Load())
```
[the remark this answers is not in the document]
GustavoStingelin: I think this case is not really a problem

GustavoStingelin:
I can also just remove the debug logs if u guys think that doesn't add up.

wallet/internal/db/itest/pg_test.go
```diff
@@ -145,16 +196,16 @@ func sanitizedPgDBName(t *testing.T) string {
...
+// before calling NewTestStore.
 func NewTestStore(t *testing.T) *db.PostgresStore {
 	t.Helper()
 	ctx := t.Context()
-
-	// Acquire a semaphore slot to limit concurrent database creation and
-	// parallel test execution that depends on it.
-	pgDBSemaphore <- struct{}{}
-	defer func() {
-		<-pgDBSemaphore
```
yyforyongyu, line 156: looks like all we need is to release this in `t.Cleanup`? so we just add a line
```go
t.Cleanup(func() { <-pgDBSemaphore })  // registered first → runs last (LIFO)

File: wallet/internal/db/itest/account_store_test.go [hunk 6]
```diff
@@ -82,8 +82,6 @@ func TestCreateDerivedAccountErrors(t *testing.T) {
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
-			t.Parallel()
```
yyforyongyu, line 85: I don't think this can be enforced, and can take more effort as LLMs tend to write it automatically, so another round needed to tell them to remove it, which can cost more work.

yyforyongyu (changes requested):
Agree with the analysis, but the requirement of removing `t.Parallel` is error-prone - there's no check, and it's easy to add it back unless we ask future devs to bear this cognitive load.

I think we can instead fix it by moving the release of that semaphore in `t.Cleanup`, which is simpler.

wallet/internal/db/itest/pg_test.go
GustavoStingelin: This was one of the approaches I tried, but because of this behavior:

> Calling t.Parallel() inside a t.Run() releases the parent goroutine early, so more top level tests start before the parent store is closed.

it ends up causing the channel to deadlock.

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: Agreed, this is not easy, that is why I added the log `if storesPeak.Load() > maxStores`. We could also turn this into a failure condition.

yyforyongyu: I think there are other ways we can do this without removing `t.Parallel`.

wallet/internal/db/itest/pg_test.go
yyforyongyu: I couldn't reproduce the deadlock,
```diff
diff --git a/wallet/internal/db/itest/pg_test.go b/wallet/internal/db/itest/pg_test.go
index d8975fa32..8b8204170 100644
--- a/wallet/internal/db/itest/pg_test.go
+++ b/wallet/internal/db/itest/pg_test.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"os"
 	"regexp"
+	"runtime"
 	"strings"
 	"sync"
 	"testing"
@@ -22,8 +23,8 @@ import (
 )
 
 var (
-	// Limit concurrent database creation to avoid exhausting connections.
-	pgDBSemaphore = make(chan struct{}, 4)
+	// Limit the number of concurrently open stores to GOMAXPROCS.
+	pgDBSemaphore = make(chan struct{}, runtime.GOMAXPROCS(0))
 
 	// Shared container instance, reused across tests for performance.
 	// This is safe to use concurrently because we only share the container
@@ -91,8 +92,6 @@ func DefaultPostgresConfig() PostgresConfig {
 
 // GetPostgresContainer returns the shared PostgreSQL container instance.
 // The container is created once and reused across all tests for performance.
-//
-// Note: postgres:18-alpine defaults max_connections to 100.
 func GetPostgresContainer(ctx context.Context) (*postgres.PostgresContainer, error) {
 	pgContainerOnce.Do(func() {
 		cfg := DefaultPostgresConfig()
@@ -115,6 +114,12 @@ func GetPostgresContainer(ctx context.Context) (*postgres.PostgresContainer, err
 			postgres.WithDatabase(cfg.Database),
 			postgres.WithUsername(cfg.Username),
 			postgres.WithPassword(cfg.Password),
+			testcontainers.WithCmd(
+				"-c", fmt.Sprintf(
+					"max_connections=%d",
+					runtime.GOMAXPROCS(0)*db.DefaultMaxConnections,
+				),
+			),
 			testcontainers.WithWaitStrategyAndDeadline(
 				pgInitTimeout, waitForSQL,
 			),
@@ -149,12 +154,11 @@ func NewTestStore(t *testing.T) *db.PostgresStore {
 	t.Helper()
 	ctx := t.Context()
 
-	// Acquire a semaphore slot to limit concurrent database creation and
-	// parallel test execution that depends on it.
+	// Acquire a semaphore slot to limit the number of concurrently open
+	// stores. Registered first so it runs last (t.Cleanup is LIFO),
+	// ensuring the slot is released only after the store is closed.
 	pgDBSemaphore <- struct{}{}
-	defer func() {
-		<-pgDBSemaphore
-	}()
+	t.Cleanup(func() { <-pgDBSemaphore })
 
 	container, err := GetPostgresContainer(ctx)
 	require.NoError(t, err, "failed to get postgres container")
```

wallet/internal/db/itest/account_store_test.go
```diff
@@ -265,8 +263,6 @@ func TestCreateImportedAccountErrors(t *testing.T) {
...
```
yyforyongyu, line 268: i think another way to do this is to create `store := NewTestStore(t)` inside the subtest, sth like,
```go
func TestFoo(t *testing.T) {
	t.Parallel()
	for _, tc := range tests {
		tc := tc
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			store := NewTestStore(t)
			walletID := newWallet(t, store, tc.name)
			// subtest-specific setup and assertions
			_ = walletID
		})
	}
}
```

Techinically this is better as each subtest should be using its own store.

wallet/internal/db/itest/pg_test.go
```diff
@@ -22,8 +24,24 @@ import (
 )
 
 var (
-	// Limit concurrent database creation to avoid exhausting connections.
-	pgDBSemaphore = make(chan struct{}, 4)
```
yyforyongyu, line 26: either way agree that we can remove this sema.

yyforyongyu (commented):
Feel like we are overcomplicating a capacity/configuration issue, given that the suite already has a clear concurrency knob (`-parallel`) and we should size the Postgres container from that knob, with headroom, instead of rewriting tests around scheduler behavior

yyforyongyu (changes requested):
To move forward, I think we should,
1. drop the newly added control states like `storesPeak`
2. add `t.Parallel()` back
3. create `NewTestStore(t)` inside the subtest
4. keep `pgMaxConns`

GustavoStingelin:
> create NewTestStore(t) inside the subtest

This one looks good to me, will test

GustavoStingelin:
Note, that `storesPeak` is just for debug logs

wallet/internal/db/itest/pg_test.go
```diff
@@ -145,16 +196,16 @@ func sanitizedPgDBName(t *testing.T) string {
 
 // NewTestStore creates a new PostgreSQL database connection with migrations
 // applied. Each test gets its own database for isolation.
+//
+// Callers must NOT call t.Parallel() inside t.Run() subtests after obtaining a
```
yyforyongyu: This doc needs to be updated.

wallet/internal/db/itest/pg_test.go
```diff
@@ -22,8 +24,24 @@ import (
...
+	// entire test suite run.
+	storesTotal atomic.Int64
+
+	// pgMaxConns sets the max_connections for the test PostgreSQL container.
+	//
+	// Tests may run in parallel, and each test can open up to
+	// db.DefaultMaxConnections connections. We size the container limit as
+	// GOMAXPROCS * DefaultMaxConnections to match the configured level of
+	// parallelism and avoid connection exhaustion during test execution.
+	pgMaxConns = runtime.GOMAXPROCS(0) * db.DefaultMaxConnections
```
yyforyongyu: ok this is not quite right yet - we want to catch `-parallel`, not how many cores the machine has, as `-parallel` is what determines the num of postgres conns will be used.

yyforyongyu: sth like this,
```go
func testParallelism() int {
	f := flag.Lookup("test.parallel")
	if f == nil {
		return runtime.GOMAXPROCS(0)
	}
	n, err := strconv.Atoi(f.Value.String())
	if err != nil || n <= 0 {
		return runtime.GOMAXPROCS(0)
	}
	return n
}
```

yyforyongyu (changes requested):
We need to fix the `-parallel` counting, i ran locally and got this warning:
```
> env GOWORK=off go test -count=1 -v -tags="itest dev debug stdlog test_db_postgres" -test.parallel=32 ./wallet/internal/db/itest

...
[18:22:11.588] WARNING: peak open stores (13) exceeded safe limit (10); each store uses up to 25 connections, container max_connections=250
```


> Note, that storesPeak is just for debug logs

Can we remove them entirely?  The warning path adds a lot of noise for very little long-term value. It also makes the test helper more complex than the actual problem requires.

GustavoStingelin:
Cool, think now it's ready

wallet/internal/db/itest/pg_test.go
GustavoStingelin: added

wallet/internal/db/itest/pg_test.go
GustavoStingelin: updated

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: changed to this way

GustavoStingelin:
liked the new solution

wallet/internal/db/itest/account_store_test.go
```diff
@@ -665,6 +656,16 @@ func TestRenameAccountErrors(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 
+			store := NewTestStore(t)
+			walletID := newWallet(t, store, "wallet-rename-account-errors")
+			createAllAccounts(t, store, walletID)
+			tc.params.WalletID = walletID
+
+			if tc.name == "invalid - both set" {
```
yyforyongyu, line 664: this can be fragile tho, a better way is to use a flag to toggle this behavior.

yyforyongyu (approved):
LGTM 🛡️

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: cool, nice catch!

https://github.com/btcsuite/btcwallet/pull/1198

PR #1186, opened 2026-03-09
wallet: add SQL tx stores

yyforyongyu:
This PR builds on #1197
and adds the SQL-backed `TxStore` implementations for postgres
and sqlite, along with the shared helpers and tests needed to exercise the new
store read/write paths.

Already reviewed [here](https://github.com/yyforyongyu/btcwallet/pull/31).

PR #1193, opened 2026-03-24
wallet: add SQL UTXO store

yyforyongyu:
This PR builds on https://github.com/btcsuite/btcwallet/pull/1186
and adds the SQL-backed `UTXOStore` implementations for postgres
and sqlite, along with the shared helpers and tests needed to exercise the new
store read/write paths.

### DIFF
--- a/wallet/internal/db/README.md
+++ b/wallet/internal/db/README.md
@@ -1,0 +1,109 @@
+# wallet/internal/db
+
+This package keeps the public wallet-store APIs in one place while supporting
+multiple SQL backends.
+
+## Ops interfaces
+
+Some store methods use a small backend-specific `ops` interface together with a
+shared backend-independent helper.
+
+Here, `ops` is short for the backend-specific database operations that one
+shared workflow needs.
+
+We use this pattern because postgres and sqlite usually agree on the high-level
+wallet workflow, but they still differ in important low-level ways:
+
+- sqlc generates different parameter and row types for each backend
+- nullable arguments often have different shapes, such as typed `sql.Null*`
+  values in postgres versus sqlite-specific generated forms
+- integer widths and conversion helpers differ across backend bindings
+- some statements need backend-specific SQL or binding workarounds
+- we still want compile-time checking against each backend's concrete query set
+
+Without a small adapter layer, each backend file tends to duplicate the same
+business workflow with only query-shape differences. That makes reviews noisy
+and makes later invariant changes easy to miss in one backend.
+
+Use this pattern when a method has both of these properties:
+
+- the high-level workflow is the same for postgres and sqlite
+- the SQL query types, nullable argument shapes, or row handling differ by
+  backend
+
+In that case:
+
+- keep the shared workflow in one backend-independent helper
+- keep the backend adapters close to the concrete backend methods
+- keep shared unit tests near the shared workflow
+- keep backend-visible behavior in integration tests
+- document the algorithm in detail on both the `ops` interface and the shared
+  helper so reviewers can see the intended sequencing and ownership boundaries
+
+The abstraction boundary is intentionally narrow:
+
+- one backend-specific `ops` interface per shared method workflow
+- one shared backend-independent workflow helper for that method
+- backend-specific preparation, row-count handling, sqlc binding shapes, and
+  query error mapping stay in backend files
+
+If a helper is only called from backend-specific files and never from the
+shared workflow helper, it should stay in backend-specific files rather than in
+the common workflow file.
+
+This approach is meant to keep one copy of the domain workflow while still
+leaving backend-specific SQL details explicit and close to the actual queries.
+The shared helper owns the sequencing and invariants. The backend `ops`
+implementation owns query calls, generated binding types, backend-specific
+conversions, and any helper code that only those backend methods need.
+
+Examples:
+
+- tx store: `CreateTx`, `UpdateTx`, `DeleteTx`, `RollbackToBlock`
+- utxo store: `LeaseOutput`, `ReleaseOutput`
+
+Minimal example:
+
+```go
+type someMethodOps interface {
+	load(ctx context.Context, id int64) (row, error)
+	write(ctx context.Context, req request) error
+}
+
+func someMethodWithOps(ctx context.Context, req request,
+	ops someMethodOps) error {
+	loaded, err := ops.load(ctx, req.id)
+	if err != nil {
+		return err
+	}
+
+	return ops.write(ctx, mergeRequest(req, loaded))
+}
+```
+
+The shared helper owns the ordering and invariants. Each backend `ops`
+implementation only adapts query calls, generated sqlc types, and backend-
+specific conversions needed by that shared method.
+
+The algorithm docs should be explicit. They should describe the ordered stages
+of the shared workflow, what each stage is responsible for, and why the shared
+helper owns that sequencing instead of leaving it implicit in backend files.
+
+Do not introduce an `ops` interface for thin read methods or simple wrappers.
+If a method is mostly one query plus row conversion, prefer direct backend
+implementations. Examples include `GetTx`, `ListTxns`, `GetUtxo`,
+`ListUTXOs`, `ListLeasedOutputs`, and `Balance`.
+
+Likewise, do not add extra common-file helper layers for backend-only prep
+work. If a helper only exists to support postgres or sqlite methods and is not
+called by the shared workflow helper itself, keep it in backend-specific files.
+
+## File layout
+
+- shared helpers and shared workflows should stay backend-independent
+- backend-specific implementations should stay close to the concrete backend
+- method-specific files are preferred over large backend files that mix many
+  unrelated methods
+
+This layout keeps commit boundaries small, makes review easier, and lets shared
+logic evolve without hiding backend-specific SQL details.
--- a/wallet/internal/db/itest/tx_utxo_store_test.go
+++ b/wallet/internal/db/itest/tx_utxo_store_test.go
@@ -1419,6 +1419,106 @@ func TestLeaseOutputRejectsNonPositiveDuration(t *testing.T) {
 	require.ErrorIs(t, err, db.ErrInvalidParam)
 }
 
+// TestReleaseOutputUnlocksMatchingLease verifies that ReleaseOutput removes the
+// active lease when the caller presents the matching lock ID.
+func TestReleaseOutputUnlocksMatchingLease(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-release-output")
+	createDerivedAccount(t, store, walletID, db.KeyScopeBIP0084, "default")
+
+	addr := newDerivedAddress(
+		t, store, walletID, db.KeyScopeBIP0084, "default", false,
+	)
+
+	tx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 20000, PkScript: addr.ScriptPubKey}},
+	)
+
+	err := store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       tx,
+		Received: time.Unix(1710001900, 0),
+		Status:   db.TxStatusPending,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
+	require.NoError(t, err)
+
+	leaseID := RandomHash()
+	_, err = store.LeaseOutput(t.Context(), db.LeaseOutputParams{
+		WalletID: walletID,
+		ID:       leaseID,
+		OutPoint: wire.OutPoint{Hash: tx.TxHash(), Index: 0},
+		Duration: time.Minute,
+	})
+	require.NoError(t, err)
+
+	err = store.ReleaseOutput(t.Context(), db.ReleaseOutputParams{
+		WalletID: walletID,
+		ID:       leaseID,
+		OutPoint: wire.OutPoint{Hash: tx.TxHash(), Index: 0},
+	})
+
+	require.NoError(t, err)
+
+	otherID := RandomHash()
+	_, err = store.LeaseOutput(t.Context(), db.LeaseOutputParams{
+		WalletID: walletID,
+		ID:       otherID,
+		OutPoint: wire.OutPoint{Hash: tx.TxHash(), Index: 0},
+		Duration: time.Minute,
+	})
+	require.NoError(t, err)
+}
+
+// TestReleaseOutputRejectsWrongLockID verifies that ReleaseOutput reports the
+// public unlock error when another active lock still owns the output.
+func TestReleaseOutputRejectsWrongLockID(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-release-conflict")
+	createDerivedAccount(t, store, walletID, db.KeyScopeBIP0084, "default")
+
+	addr := newDerivedAddress(
+		t, store, walletID, db.KeyScopeBIP0084, "default", false,
+	)
+
+	tx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 21000, PkScript: addr.ScriptPubKey}},
+	)
+
+	err := store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       tx,
+		Received: time.Unix(1710002000, 0),
+		Status:   db.TxStatusPending,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
+	require.NoError(t, err)
+
+	leaseID := RandomHash()
+	_, err = store.LeaseOutput(t.Context(), db.LeaseOutputParams{
+		WalletID: walletID,
+		ID:       leaseID,
+		OutPoint: wire.OutPoint{Hash: tx.TxHash(), Index: 0},
+		Duration: time.Minute,
+	})
+	require.NoError(t, err)
+
+	wrongID := RandomHash()
+	err = store.ReleaseOutput(t.Context(), db.ReleaseOutputParams{
+		WalletID: walletID,
+		ID:       wrongID,
+		OutPoint: wire.OutPoint{Hash: tx.TxHash(), Index: 0},
+	})
+
+	require.ErrorIs(t, err, db.ErrOutputUnlockNotAllowed)
+}
+
 // newCoinbaseTx builds a simple coinbase fixture transaction.
 func newCoinbaseTx(pkScript []byte) *wire.MsgTx {
 	tx := wire.NewMsgTx(2)
--- a/wallet/internal/db/itest/tx_utxo_store_test.go
+++ b/wallet/internal/db/itest/tx_utxo_store_test.go
@@ -1316,6 +1316,109 @@ func TestListUTXOsFiltersByAccount(t *testing.T) {
 	require.Equal(t, txSavings.TxHash(), utxos[0].OutPoint.Hash)
 }
 
+// TestLeaseOutputLocksCurrentUtxo verifies that LeaseOutput returns the active
+// lease metadata for a current wallet-owned output.
+func TestLeaseOutputLocksCurrentUtxo(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-lease-output")
+	createDerivedAccount(t, store, walletID, db.KeyScopeBIP0084, "default")
+
+	addr := newDerivedAddress(
+		t, store, walletID, db.KeyScopeBIP0084, "default", false,
+	)
+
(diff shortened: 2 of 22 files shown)


wallet/internal/db/utxo_store_common.go
```diff
@@ -0,0 +1,127 @@
+package db
```
yyforyongyu, line 1: Missing documentation at the top of the file explaining the purpose and scope of the UTXO store common helpers. Add package-level doc comment. (`gemini-3.1-pro-preview`)

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

wallet/internal/db/utxo_store_pg.go
```diff
@@ -0,0 +1,323 @@
...
+		}
+
+		leases[i] = *lease
+	}
+
+	return leases, nil
+}
+
+// Balance returns the sum of wallet-owned current UTXOs after optional filters.
+func (s *PostgresStore) Balance(ctx context.Context,
```
yyforyongyu: `NowUtc` is missing from the `BalanceParams` struct, so it defaults to Go's zero `time.Time` (year 0001). The Balance SQL uses `l.expires_at > sqlc.arg('now_utc')` in the locked-balance subquery — with a zero timestamp, every lease row (including expired ones not yet garbage-collected) counts toward `locked_balance`. Same issue in `SqliteStore.Balance`. Should be `NowUtc: time.Now().UTC()` matching the pattern used by `LeaseOutput`, `ReleaseOutput`, and `ListLeasedOutputs`. **(`claude-opus-4-6`)**

yyforyongyu:
Reviewed the split PR - no issues found. LGTM ✨ `(qwen3.5-397b-a17b)`

wallet/internal/db/utxo_store_pg.go
```diff
@@ -0,0 +1,323 @@
...
+	params BalanceParams) (BalanceResult, error) {
+
+	balance, err := s.queries.Balance(ctx, sqlcpg.BalanceParams{
```
yyforyongyu: The balance SQL now depends on `now_utc` to decide whether a lease is still active, but this call never populates that first query parameter. The generated bindings therefore compare `expires_at` against Go's zero time and count expired leases as locked forever. Please thread a real current UTC timestamp through this path, and mirror the fix in SQLite. (`gpt-5.4`)

yyforyongyu:
LGTM ✨ (`glm5`)

wallet/internal/db/utxo_store_pg.go
yyforyongyu: The balance SQL now depends on now_utc to decide whether a lease is still active, but this call never populates that first query parameter. The generated bindings therefore compare expires_at against Go zero time and count expired leases as locked forever. Please thread a real current UTC timestamp through this path, and mirror the fix in SQLite. (`gpt-5.4`)

wallet/internal/db/utxo_store_pg.go
```diff
@@ -0,0 +1,323 @@
...
+// LeaseOutput atomically acquires or renews a lease for one current UTXO.
+//
+// The lease lookup and acquisition run in one transaction so competing calls
+// cannot observe a partially-written lease. Expiration timestamps are
+// normalized to UTC before insert.
+func (s *PostgresStore) LeaseOutput(ctx context.Context,
+	params LeaseOutputParams) (*LeasedOutput, error) {
+
+	nowUTC := time.Now().UTC()
+	expiresAt := nowUTC.Add(params.Duration)
```
yyforyongyu: This accepts zero or negative lease durations, so the method can return success while writing a lease whose expiration is already in the past. The rest of the SQL only treats expires_at > now_utc as active, so such a lease never actually reserves the UTXO and can be stolen immediately. Please reject non-positive durations here, and mirror the fix in SQLite. (`gpt-5.4`)

wallet/internal/db/utxo_store_pg.go
yyforyongyu: Addressed locally on `adr-0006-utxo` in `00cf76fab`: both backends now thread `time.Now().UTC()` into `Balance`, so `locked_balance` only counts still-active leases. Verified with `GOWORK=off make itest-db db=sqlite` and `make lint`. I will include it in the next push.

wallet/internal/db/utxo_store_pg.go
yyforyongyu: Addressed locally on `adr-0006-utxo` in `00cf76fab`: the balance path now threads `time.Now().UTC()` into the query in both backends, which fixes the locked-balance calculation for expired leases. Verified with `GOWORK=off make itest-db db=sqlite` and `make lint`. I will include it in the next push.

yyforyongyu: Addressed locally on `adr-0006-utxo` in `00cf76fab`: `Balance` now passes the current UTC time through the generated params in both postgres and sqlite, so expired leases no longer contribute to `locked_balance`. Verified with `GOWORK=off make itest-db db=sqlite` and `make lint`. I will include it in the next push.

wallet/internal/db/utxo_store_pg.go
yyforyongyu: Addressed locally on `adr-0006-utxo` in `9759780b4`: `LeaseOutput` now rejects zero or negative durations in both backends before attempting the lease write, and the LeaseOutput integration tests cover the non-positive duration cases. Verified with `GOWORK=off make itest-db db=sqlite` and `make lint`. I will include it in the next push.

yyforyongyu (commented):
LGTM 🦊 **(`claude-opus-4-6`)**

yyforyongyu (commented):
LGTM 🌱 (`gpt-5.4`)

yyforyongyu (commented):
LGTM 🌿 (`gpt-5.4`)

yyforyongyu (commented):
LGTM 🌿 (`gpt-5.4`)

yyforyongyu:
No issues from my review. LGTM ✨ `(qwen3.5-397b-a17b)`

wallet/internal/db/itest/tx_utxo_store_test.go
```diff
@@ -691,6 +691,671 @@ func TestDeleteTxRejectsNonLeafTx(t *testing.T) {
...
+		WalletID: walletID,
+		Txid:     tx.TxHash(),
+	})
+	require.NoError(t, err)
+	require.Nil(t, info.Block)
+	require.Equal(t, db.TxStatusPublished, info.Status)
+}
+
+// TestRollbackToBlockOrphansCoinbaseAndFailsDescendants verifies that
+// RollbackToBlock lets the block delete orphan rolled-back coinbase rows and
```
yyforyongyu: Agree with copilot: the comment says "lets the block delete orphan rolled-back coinbase rows" but the test asserts `TxStatusOrphaned` — the row is kept, not deleted. Should read something like "marks rolled-back coinbase rows as orphaned and marks any unmined descendants of those coinbase roots as failed." **(`claude-opus-4-6`)**

yyforyongyu: This comment still states that RollbackToBlock "lets the block delete orphan rolled-back coinbase rows", but the assertions below expect the coinbase tx to still be returned by GetTx with status `TxStatusOrphaned`. Please adjust the comment to explicitly state that it "marks rolled-back coinbase rows as orphaned" rather than letting them be deleted. (`gemini-3.1-pro-preview`)

wallet/internal/db/queries/sqlite/utxo_leases.sql
```diff
@@ -49,7 +49,7 @@ SET
 WHERE
 utxo_leases.wallet_id = excluded.wallet_id
 AND (
-    utxo_leases.expires_at <= sqlc.arg('now_utc')
+    utxo_leases.expires_at <= ?6
     OR utxo_leases.lock_id = excluded.lock_id
 )
```
yyforyongyu, line 57: In the SQLite `AcquireUtxoLease` query, the hard-coded positional parameter `?6` couples the conflict predicate to the current sqlc argument ordering. This breaks easily if the query is refactored. Consider rewriting this query to avoid referencing a specific parameter index, perhaps by introducing a CTE (`WITH params AS (SELECT sqlc.arg('now_utc') AS now_utc)`), or adding an explicit inline comment explaining why `?6` maps directly to `now_utc`. (`gemini-3.1-pro-preview`)

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

yyforyongyu (commented):
LGTM 🌻 (`gpt-5.4`)

wallet/internal/db/queries/sqlite/utxo_leases.sql
yyforyongyu, line 57: Agreed with the reviewer - the hard-coded `?6` positional parameter couples the conflict predicate to the sqlc argument ordering. This is fragile if the query is refactored. Consider adding an explicit inline comment explaining that `?6` must map to `now_utc` to prevent accidental breakage during refactoring. (`glm5`)

wallet/internal/db/itest/tx_utxo_store_test.go
yyforyongyu: Agreed with the reviewers - the comment says "lets the block delete orphan rolled-back coinbase rows" but the test asserts TxStatusOrphaned. The row is kept and marked orphaned, not deleted. The comment should be updated to: "marks rolled-back coinbase rows as orphaned and marks any unmined descendants of those coinbase roots as failed." (`glm5`)

yyforyongyu (commented):
LGTM 🌻 (`gpt-5.4`)

wallet/internal/db/queries/sqlite/utxo_leases.sql
yyforyongyu: Addressed locally on `adr-0006-utxo`: I kept `?6` because the earlier `sqlc.arg('now_utc')` form generated invalid SQLite here, and added an inline note explaining that `?6` is the generated bind slot for `now_utc` and must stay aligned with the arg ordering above. Reverified with `GOWORK=off make itest-db db=sqlite` and `make lint`.

wallet/internal/db/itest/tx_utxo_store_test.go
yyforyongyu: Addressed locally on `adr-0006-utxo`: the test comment now says rollback marks rolled-back coinbase rows as orphaned and marks any unmined descendants as failed, which matches the assertions below. Reverified with `GOWORK=off make itest-db db=sqlite` and `make lint`.

wallet/internal/db/queries/sqlite/utxo_leases.sql
yyforyongyu: Addressed locally on `adr-0006-utxo`: I kept `?6` because the earlier `sqlc.arg('now_utc')` form generated invalid SQLite here, and added an inline note explaining that `?6` is the generated bind slot for `now_utc` and must stay aligned with the arg ordering above. Reverified with `GOWORK=off make itest-db db=sqlite` and `make lint`.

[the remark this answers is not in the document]
yyforyongyu: Addressed locally on `adr-0006-utxo`: I kept `?6` because the earlier `sqlc.arg('now_utc')` form generated invalid SQLite here, and added an inline note explaining that `?6` is the generated bind slot for `now_utc` and must stay aligned with the arg ordering above. Reverified with `GOWORK=off make itest-db db=sqlite` and `make lint`.

wallet/internal/db/itest/tx_utxo_store_test.go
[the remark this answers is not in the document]
yyforyongyu: Addressed locally on `adr-0006-utxo`: the test comment now says rollback marks rolled-back coinbase rows as orphaned and marks any unmined descendants as failed, which matches the assertions below. Reverified with `GOWORK=off make itest-db db=sqlite` and `make lint`.

yyforyongyu: Addressed locally on `adr-0006-utxo`: the test comment now says rollback marks rolled-back coinbase rows as orphaned and marks any unmined descendants as failed, which matches the assertions below. Reverified with `GOWORK=off make itest-db db=sqlite` and `make lint`.

yyforyongyu: Addressed locally on `adr-0006-utxo`: the test comment now says rollback marks rolled-back coinbase rows as orphaned and marks any unmined descendants as failed, which matches the assertions below. Reverified with `GOWORK=off make itest-db db=sqlite` and `make lint`.

yyforyongyu (commented):
LGTM 🔥 **(`claude-opus-4-6`)**

yyforyongyu:

yyforyongyu:
LGTM 🎉 (`glm5`)

wallet/internal/db/utxo_store_common.go
```diff
@@ -23,6 +24,8 @@ var (
 	// ErrOutputUnlockNotAllowed reports that a UTXO release request used a lock
 	// ID different from the active lease.
 	ErrOutputUnlockNotAllowed = errors.New("output unlock not allowed")
+
+	errLeaseOutputNoRow = errors.New("lease output no row")
```
yyforyongyu, line 30: missing godoc

wallet/internal/db/utxo_store_common.go
```diff
@@ -26,6 +27,14 @@ var (
 	ErrOutputUnlockNotAllowed = errors.New("output unlock not allowed")
 
 	errLeaseOutputNoRow = errors.New("lease output no row")
+
+	errReleaseOutputUtxoNotFound = errors.New(
```
yyforyongyu, line 34: missing godocs

wallet/internal/db/README.md
```diff
@@ -0,0 +1,65 @@
...
+- the high-level workflow is the same for postgres and sqlite
+- the SQL query types, nullable argument shapes, or row handling differ by
+  backend
+
+In that case:
+
+- keep the shared workflow in `*_common.go`
+- keep the backend adapters in `postgres_*` and `sqlite_*`
+- keep shared unit tests in `*_common_test.go`
+- keep backend-visible behavior in `itest/tx_utxo_store_test.go`
```
yyforyongyu: we should make this more general, not specific to certain files?

wallet/internal/db/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+This approach is meant to keep one copy of the domain workflow while still
+leaving backend-specific SQL details explicit and close to the actual queries.
+The shared helper owns the sequencing and invariants. The backend `ops`
+implementation owns query calls, generated binding types, and backend-specific
+conversions.
+
+Examples:
```
yyforyongyu, line 60: should give some minimal coding examples?

yyforyongyu: Added a small generic code example showing the shared helper plus backend ops adapter shape.

wallet/internal/db/README.md
yyforyongyu: Updated the README wording to make the guidance more package-level and less tied to specific filenames.

wallet/internal/db/utxo_store_common.go
yyforyongyu: Added godocs for the internal ReleaseOutput sentinels as well.

wallet/internal/db/utxo_store_common.go
yyforyongyu: Added godoc for this internal sentinel in the shared UTXO helper layer.

yyforyongyu:

yyforyongyu (commented):
LGTM 🎉 **(`claude-opus-4-6`)**

yyforyongyu:
LGTM 🎉 (`glm5`)

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

yyforyongyu (commented):
LGTM 🌻 (`gpt-5.4`)

yyforyongyu (commented):
LGTM 🌻 (`gpt-5.4`)

yyforyongyu:
No issues from my review. LGTM ✨ `(qwen3.5-397b-a17b)`

yyforyongyu (commented):
LGTM 🌻 (`gpt-5.4`)

wallet/internal/db/postgres_utxostore_leaseoutput.go
```diff
@@ -0,0 +1,108 @@
...
+			NowUtc:      nowUTC,
+		},
+	)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return time.Time{}, errLeaseOutputNoRow
+		}
+
+		return time.Time{}, fmt.Errorf("acquire lease row: %w", err)
+	}
```
GustavoStingelin: nit: could use that trick to early return `err == nil { return expiration, nil }` and avoid one if nested to check error type. Apply for other methods too.

wallet/internal/db/utxo_store_common.go
```diff
@@ -0,0 +1,298 @@
...
+
+// optionalUint32Int64 converts an optional uint32 filter into the nullable any
+// form used by sqlite sqlc queries.
+func optionalUint32Int64(value *uint32) any {
+	if value == nil {
+		return nil
+	}
+
+	return int64(*value)
+}
```
GustavoStingelin: I missed this in the previous review, but some of the SQLC-generated code is producing weakly typed fields as `interface{}`, which significantly reduces our type safety.

I think we should restructure the queries that lead to those interfaces to enforce strong typing, similar to what you did in `ListUtxos` with `sqlc.narg('min_confirms')::INTEGER`.

For context, `sqlc.narg` indicates a nullable argument, while `sqlc.arg` is non-nullable.

wallet/internal/db/utxo_store_common.go
```diff
@@ -0,0 +1,298 @@
...
+
+// nullableInt32 converts an optional int32 filter into the typed null form
+// used by postgres sqlc queries.
+func nullableInt32(value *int32) sql.NullInt32 {
+	if value == nil {
+		return sql.NullInt32{}
+	}
+
+	return sql.NullInt32{Int32: *value, Valid: true}
+}
```
GustavoStingelin: nit: these functions could be moved to a more general file to better reflect their broader scope

wallet/internal/db/postgres_utxostore_leaseoutput.go
```diff
@@ -0,0 +1,108 @@
...
+// acquire attempts to write or renew one postgres lease row for the requested
+// outpoint.
+func (o *pgLeaseOutputOps) acquire(ctx context.Context,
+	params LeaseOutputParams, nowUTC time.Time,
+	expiresAt time.Time) (time.Time, error) {
+
+	outputIndex, err := uint32ToInt32(params.OutPoint.Index)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("convert output index: %w", err)
+	}
```
GustavoStingelin, line 59: What about a uint32 that comes with a value (2³¹ + n)? Will not support storing in the int32? Is this an expected behavior?

wallet/internal/db/utxo_store_common.go
```diff
@@ -0,0 +1,298 @@
...
+// questions in the same order: did the lease write succeed, and if not, was the
+// target output missing or merely already leased by a different lock?
+func leaseOutputWithOps(ctx context.Context, params LeaseOutputParams,
+	ops leaseOutputOps) (*LeasedOutput, error) {
+
+	if params.Duration <= 0 {
+		return nil, fmt.Errorf(
+			"%w: lease duration must be positive",
+			ErrInvalidParam,
+		)
```
GustavoStingelin:
```suggestion
		return nil, fmt.Errorf("%w: lease duration must be positive",
			ErrInvalidParam)
```

GustavoStingelin (changes requested):
Looking good, just a major concern about queries that use `interface` types in the struct.

wallet/internal/db/postgres_utxostore_leaseoutput.go
yyforyongyu: yeah, and i don't think it's realistic to think the index can be greater than `int32`.

wallet/internal/db/utxo_store_common.go
yyforyongyu: Good catch! this is now fixed in a new commit (the second last one)

wallet/internal/db/postgres_utxostore_leaseoutput.go
yyforyongyu: nice! updated

wallet/internal/db/utxo_store_common.go
yyforyongyu: make sense - added a new commit, also updated account to use it.

GustavoStingelin (approved):
🚀 🚀 🚀

PR #1199, opened 2026-03-30
wallet: simplify postgres itest headroom and add repo guide

yyforyongyu:
## Summary
- simplify the Postgres itest `max_connections` budget in `wallet/internal/db/itest/pg_test.go` to `2 * testParallelism() * db.DefaultMaxConnections`
- add `AGENTS.md` with btcwallet-specific build, test, and style guidance for coding agents
- ignore nested worktree directories in `.gitignore` so repo-local worktrees do not pollute `git status`

## Why
The previous Postgres itest budget still proved too tight in CI and was getting increasingly complicated to reason about.

This change takes the simpler test-infra approach: give the shared Postgres container 2x the steady-state pool budget instead of trying to precisely model each transient connection source during parallel startup and teardown. Since this is only for integration tests, the larger buffer is easier to understand, more robust to future itest growth, and still scales automatically with `-test.parallel`.

The repo guide documents local workflow, build, test, and style conventions for coding agents so future changes follow the same repository expectations.

## Testing
- `env GOWORK=off make itest-db db=postgres cover=1 verbose=1`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,160 @@
+# AGENTS.md
+
+This file is for coding agents working in `btcwallet`.
+
+## Scope and Priority
+
+- Follow this file first for repo-specific workflow and style.
+- Then follow `docs/developer/` for deeper rationale.
+- If guidance conflicts, prefer the stricter rule and match nearby code.
+
+## Repo Workflow
+
+- Keep the main checkout at the repo root.
+- Place auxiliary worktrees under `<repo-root>/.worktrees/<name>`.
+- Use short issue- or task-oriented worktree names.
+- Do not create sibling worktrees outside the repo root unless asked.
+- Remove finished worktrees with `git worktree remove <path>`.
+- Clean stale worktree metadata with `git worktree prune`.
+
+## Tooling and Environment
+
+- The authoritative Go version is `1.24.6`.
+- Sources: `go.mod`, `.golangci.yml`, and `.github/workflows/main.yml`.
+- Some older docs mention older Go versions; ignore them.
+- Many maintenance targets use Docker-backed tooling.
+- `make fmt`, `make lint*`, `make sqlc`, and `make protolint` use Docker.
+- PostgreSQL DB integration tests also require Docker via testcontainers.
+- No Cursor rules were found in `.cursor/rules/` or `.cursorrules`.
+- No Copilot instructions were found in `.github/copilot-instructions.md`.
+- There is an additional style summary in `.gemini/styleguide.md`.
+
+## Primary Build, Format, and Codegen Commands
+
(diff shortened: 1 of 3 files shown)


wallet/internal/db/itest/pg_test.go
```diff
@@ -136,21 +136,11 @@ func GetPostgresContainer(ctx context.Context) (*postgres.PostgresContainer,
...
-		//          3 transient connections while a new test starts —
-		//          one admin connection for CREATE DATABASE, ~1 for
-		//          PingContext, and ~1 for migration setup — so 3*P
-		//          covers all slots transitioning simultaneously.
-		pgMaxConns := p*m + m + 3*p
+		// test container. We budget 2x the steady-state pool size to
+		// absorb connection lifecycle overlap during parallel test
+		// teardown and startup without trying to model each transient
+		// connection source separately.
+		pgMaxConns := 2 * p * m
```
yyforyongyu, line 143: I intentionally simplified the Postgres itest sizing here.

We had already tried a more exact transient-overlap formula, but CI still hit `too many clients already`, e.g., [here](https://github.com/btcsuite/btcwallet/actions/runs/23739923422/job/69154124795?pr=1196), and the reasoning was getting **harder to maintain than the fix itself**.

This switches the shared test container budget to:  `2 * testParallelism() * db.DefaultMaxConnections`.

The idea is just to give the itest container 2x the steady-state store pool budget and stop trying to account for each startup/teardown connection source individually. Since this is test-only infrastructure, the simpler and more conservative rule felt like the better tradeoff.

It also still scales automatically with `-test.parallel`, so anyone who wants a smaller local footprint can lower parallelism and the connection budget follows.

AGENTS.md
```diff
@@ -0,0 +1,160 @@
...
+- If guidance conflicts, prefer the stricter rule and match nearby code.
+
+## Repo Workflow
+
+- Keep the main checkout at the repo root.
+- Place auxiliary worktrees under `<repo-root>/.worktrees/<name>`.
+- Use short issue- or task-oriented worktree names.
+- Do not create sibling worktrees outside the repo root unless asked.
+- Remove finished worktrees with `git worktree remove <path>`.
+- Clean stale worktree metadata with `git worktree prune`.
```
GustavoStingelin, line 18: I usually do not use worktrees, idk if my LLMs will try to use them with this workflow here, but I also don't complain if they start using them. So, let's see.

GustavoStingelin (approved):
LGTM! 🚢 🚢 🚢

AGENTS.md
yyforyongyu: yeah it's extremely helpful, especially when we are working on multiple branches at the same time

PR #1200, opened 2026-03-30
wallet+db: handle SQL backend errors

yyforyongyu:
## Summary
- introduce the shared SQL error framework in `wallet/internal/db/err`, including the shared taxonomy, wrapper, normalization helpers, stats collector, and package README
- add backend-specific SQL error mapper packages in `wallet/internal/db/postgres` and `wallet/internal/db/sqlite`, with same-package tests and private helper details
- add the shared `wallet/internal/db/runtime` helper framework for retry, unhealthy-store checks, transaction execution, and ambiguous commit handling
- keep this PR focused on framework code only; production store wiring and runtime integration remain follow-up work once store code moves into the backend packages

## Follow-up Plan
- move store code into `wallet/internal/db/postgres` and `wallet/internal/db/sqlite`
- integrate production normalization, stats recording, and runtime/store reaction logic in their final package locations
- wire the shared runtime helper package on top of the backend-owned store implementations

## Testing
- make lint
- GOWORK=off go test ./wallet/internal/db/... ./wallet
- GOWORK=off go test ./wallet/internal/db/runtime
- GOWORK=off go test -run '^$' -bench . ./wallet/internal/db/runtime

### DIFF
--- a/wallet/internal/db/err/errors.go
+++ b/wallet/internal/db/err/errors.go
@@ -1,0 +1,441 @@
+// Package dberr contains the shared SQL error taxonomy and normalization logic
+// used by backend-specific db packages.
+package dberr
+
+import (
+	"context"
+	"database/sql"
+	"database/sql/driver"
+	"errors"
+	"fmt"
+	"io"
+	"net"
+)
+
+// unknownString is the fallback text for invalid enum values.
+const unknownString = "unknown"
+
+// Backend identifies which SQL backend produced a classified SQL error.
+//
+// The backend is preserved on SQLError so callers can log raw backend codes,
+// maintain per-backend stats, and keep backend-specific handling separate from
+// the shared error model.
+type Backend string
+
+// Backend constants identify supported SQL backends.
+const (
+	// BackendPostgres identifies the PostgreSQL SQL backend.
+	BackendPostgres Backend = "postgres"
+
+	// BackendSQLite identifies the SQLite SQL backend.
+	BackendSQLite Backend = "sqlite"
+)
+
+// String returns the canonical string form of the backend.
+func (b Backend) String() string {
+	switch b {
+	case BackendPostgres:
+		return string(b)
+
+	case BackendSQLite:
+		return string(b)
+
+	default:
+		return unknownString
+	}
+}
+
+// Valid reports whether the backend is one of the known SQL backends.
+func (b Backend) Valid() bool {
+	switch b {
+	case BackendPostgres:
+		return true
+
+	case BackendSQLite:
+		return true
+
+	default:
+		return false
+	}
+}
+
+// Class describes the caller-facing policy bucket for a classified SQL backend
+// error.
+//
+// The class answers what runtime code should do next:
+//
+//   - Transient: retry the operation when it is safe to retry.
+//   - Permanent: fail the operation immediately without poisoning the store.
+//   - Fatal: fail the operation and mark the store unhealthy so later calls
+//     fail fast until the backend is reopened or repaired.
+type Class uint8
+
+// Class constants identify caller-facing SQL error policy buckets.
+const (
+	// ClassTransient marks failures that may succeed on a later retry.
+	ClassTransient Class = iota
+
+	// ClassPermanent marks failures that should fail the current
+	// operation without retrying or poisoning the store.
+	ClassPermanent
+
+	// ClassFatal marks failures that should fail the current operation
+	// and take the store out of service.
+	ClassFatal
+)
+
+// String returns the canonical string form of the class.
+func (c Class) String() string {
+	switch c {
+	case ClassTransient:
+		return "transient"
+
+	case ClassPermanent:
+		return "permanent"
+
+	case ClassFatal:
+		return "fatal"
+
+	default:
+		return unknownString
+	}
+}
+
+// Valid reports whether the class is recognized.
+func (c Class) Valid() bool {
+	switch c {
+	case ClassTransient:
+		return true
+
+	case ClassPermanent:
+		return true
+
+	case ClassFatal:
+		return true
+
+	default:
+		return false
+	}
+}
+
+// Reason describes the specific backend failure bucket for a classified SQL
+// error.
+//
+// Callers use Reason for diagnosis and metrics, then derive runtime policy
+// through Class(). New or unmapped backend codes should fall back to
+// ReasonUnknown while preserving the raw backend code on SQLError.
+type Reason uint8
+
+// Reason constants identify shared SQL backend failure buckets.
+const (
+	// ReasonUnknown is the fallback reason for uncategorized backend codes.
+	ReasonUnknown Reason = iota
+
+	// ReasonSerialization marks transaction serialization failures
+	// caused by concurrent updates.
+	ReasonSerialization
+
+	// ReasonDeadlock marks deadlock failures where the backend aborts
+	// one participant to break a lock cycle.
+	ReasonDeadlock
+
+	// ReasonBusy marks SQLite busy failures where a lock cannot be
+	// acquired in time.
+	ReasonBusy
+
+	// ReasonLocked marks lock-not-available failures.
+	ReasonLocked
+
+	// ReasonUnavailable marks backend availability and transport
+	// failures.
+	ReasonUnavailable
+
+	// ReasonPoolExhausted marks connection-pool or backend
+	// connection-limit failures.
+	ReasonPoolExhausted
+
+	// ReasonSchemaMismatch marks schema drift or schema version
+	// mismatches.
+	ReasonSchemaMismatch
+
+	// ReasonConstraint marks backend constraint failures.
+	ReasonConstraint
+
+	// ReasonResourceExhausted marks disk, memory, or similar hard
+	// resource failures.
+	ReasonResourceExhausted
+
+	// ReasonReadOnly marks read-only backend failures.
+	ReasonReadOnly
+
+	// ReasonPermission marks backend permission failures.
+	ReasonPermission
+
+	// ReasonCorrupt marks corruption failures.
+	ReasonCorrupt
+)
+
+// Class returns the caller-facing policy bucket derived from the reason.
+func (r Reason) Class() Class {
+	switch r {
+	case ReasonSerialization, ReasonDeadlock, ReasonBusy, ReasonLocked,
+		ReasonUnavailable, ReasonPoolExhausted:
+
+		return ClassTransient
+
+	// Unknown or caller-fixable backend states fail the current operation
+	// without retrying or taking the store out of service.
+	case ReasonSchemaMismatch, ReasonConstraint, ReasonUnknown:
+		return ClassPermanent
+
+	case ReasonResourceExhausted, ReasonReadOnly, ReasonPermission,
+		ReasonCorrupt:
+
+		return ClassFatal
+
+	default:
+		return ClassPermanent
+	}
+}
+
+// String returns the canonical string form of the reason.
+//
+//nolint:cyclop // One explicit enum-to-string switch is the clearest form here.
+func (r Reason) String() string {
+	switch r {
+	case ReasonSerialization:
+		return "serialization"
+
+	case ReasonDeadlock:
+		return "deadlock"
+
+	case ReasonBusy:
+		return "busy"
+
+	case ReasonLocked:
+		return "locked"
+
+	case ReasonUnavailable:
+		return "unavailable"
+
+	case ReasonPoolExhausted:
+		return "pool_exhausted"
+
+	case ReasonSchemaMismatch:
+		return "schema_mismatch"
+
+	case ReasonConstraint:
+		return "constraint"
+
+	case ReasonUnknown:
+		return unknownString
+
+	case ReasonResourceExhausted:
+		return "resource_exhausted"
+
+	case ReasonReadOnly:
+		return "read_only"
+
+	case ReasonPermission:
+		return "permission"
+
+	case ReasonCorrupt:
+		return "corrupt"
+
+	default:
+		return unknownString
+	}
+}
+
+// Valid reports whether the reason is recognized.
+func (r Reason) Valid() bool {
+	return r <= ReasonCorrupt
+}
+
+// SQLError wraps a backend or transport error with stable SQL classification
+// metadata.
+//
+// The wrapper keeps classification data inside ordinary error chains so callers
+// can recover backend, reason, class, and raw codes with errors.As after higher
+// layers add their own context.
+type SQLError struct {
+	// Backend identifies which SQL backend produced the error.
+	Backend Backend
+
+	// Reason identifies the specific backend failure bucket.
+	Reason Reason
+
+	// Code stores the raw backend code used for classification.
+	Code string
+
+	// Err stores the wrapped backend or transport error.
+	Err error
+}
+
+// NewSQLError builds a classified SQL backend error wrapper.
+func NewSQLError(backend Backend, reason Reason, code string,
+	err error) *SQLError {
+
+	return &SQLError{
+		Backend: backend,
+		Reason:  reason,
+		Code:    code,
+		Err:     err,
+	}
+}
+
+// Class returns the caller-facing policy bucket derived from the reason.
+func (e *SQLError) Class() Class {
+	if e == nil {
+		// A missing wrapper is treated like an unknown classified error so
+		// defensive callers do not retry or poison store state based on
+		// absent classification data.
+		return ReasonUnknown.Class()
+	}
+
+	return e.Reason.Class()
+}
+
+// Error returns the printable form of the wrapped backend error.
+func (e *SQLError) Error() string {
+	if e == nil {
+		return "<nil>"
+	}
+
+	// Normal wrappers carry the backend or transport error, so preserve that
+	// message and leave the remaining branches as defensive fallbacks.
+	if e.Err != nil {
+		return e.Err.Error()
+	}
+
+	// Fall back to synthesized text only for malformed or test-built
+	// wrappers that do not carry an underlying cause.
+	return fmt.Sprintf("%s %s sql error", e.Backend.String(), e.Reason)
+}
+
+// Unwrap returns the wrapped backend error.
+func (e *SQLError) Unwrap() error {
+	if e == nil {
+		return nil
+	}
+
+	return e.Err
+}
+
+// Normalize converts one backend error into the shared SQL error model.
+//
+// It handles errors in this order:
+//
+//  1. A nil error stays nil.
+//  2. Context cancellation and deadline errors are returned unchanged,
+//     including any caller-added wrapping context.
+//  3. Existing SQLError wrappers are returned unchanged so Normalize does not
+//     rebuild or strip the original error chain.
+//  4. The mapper gets the first chance to translate backend-native driver
+//     errors into SQLError values.
+//  5. Generic connection and transport failures are classified as
+//     ReasonUnavailable.
+//  6. If the backend is known and no earlier rule matches, err is wrapped as
+//     ReasonUnknown.
+//  7. If the backend is unknown, err is returned unchanged.
+//
+// Normalize is intentionally side-effect free. Callers own follow-up actions
+// such as stats recording or unhealthy-store transitions.
+//
+// The mapper keeps backend-specific matching outside the shared error package
+// so backend packages can return shared SQL errors without introducing import
+// cycles. Callers should pass a no-op mapper when they have no backend-
+// specific codes to inspect.
+func Normalize(backend Backend, mapper func(error) *SQLError, err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// Preserve caller-driven cancellation semantics, including any wrapping
+	// context, instead of reclassifying them as backend errors.
+	contextErr := unwrapContextErr(err)
+	if contextErr != nil {
+		return contextErr
+	}
+
+	// Preserve existing SQL wrappers so caller-added context is not dropped
+	// from the original error chain.
+	if extractSQLError(err) != nil {
+		return err
+	}
+
+	// Ask the backend-specific mapper first so driver-native codes stay as
+	// specific as possible.
+	sqlErr := mapper(err)
+	if sqlErr != nil {
+		return sqlErr
+	}
+
+	// Fall back to transport-level classification for connection-oriented
+	// failures that are not backend-code specific.
+	transportErr := classifyConnErr(backend, err)
+	if transportErr != nil {
+		return transportErr
+	}
+
+	// Unknown backends are left untouched so callers do not accidentally invent
+	// a backend identity the error never had.
+	if !backend.Valid() {
+		return err
+	}
+
+	// Keep backend identity even when the exact backend code is not recognized
+	// so logs and stats retain useful diagnosis data.
+	return NewSQLError(backend, ReasonUnknown, "", err)
+}
+
+// extractSQLError extracts the shared SQL error wrapper from err, if present.
+func extractSQLError(err error) *SQLError {
+	var sqlErr *SQLError
+	if errors.As(err, &sqlErr) {
+		return sqlErr
+	}
+
+	return nil
+}
+
+// classifyConnErr classifies generic connection and transport failures that do
+// not come with backend-specific codes.
+func classifyConnErr(backend Backend, err error) *SQLError {
+	if !backend.Valid() {
+		return nil
+	}
+
+	badConn := errors.Is(err, driver.ErrBadConn)
+	connDone := errors.Is(err, sql.ErrConnDone)
+
+	// Broken-connection sentinels and EOF all mean the caller lost the SQL
+	// transport path before receiving a normal backend result.
+	if badConn || connDone || errors.Is(err, io.EOF) {
+		return NewSQLError(backend, ReasonUnavailable, "", err)
+	}
+
+	var netErr net.Error
+	if errors.As(err, &netErr) {
+		// Timeout-only net errors are still transport failures even when the
+		// backend never returned a backend-specific code.
+		if netErr.Timeout() {
+			return NewSQLError(backend, ReasonUnavailable, "", err)
+		}
+	}
+
+	return nil
+}
+
+// unwrapContextErr preserves caller-driven context cancellation and deadlines.
+func unwrapContextErr(err error) error {
+	if errors.Is(err, context.Canceled) {
+		return err
+	}
+
+	if errors.Is(err, context.DeadlineExceeded) {
+		return err
+	}
+
+	return nil
+}
--- a/wallet/internal/db/err/stats.go
+++ b/wallet/internal/db/err/stats.go
@@ -1,0 +1,226 @@
+package dberr
+
+import (
+	"errors"
+	"sync/atomic"
+)
+
+// StatsSnapshot is a read-only copy of the SQL backend error counters.
+type StatsSnapshot struct {
+	// Backend identifies which SQL backend produced the snapshot.
+	Backend Backend
+
+	// TotalErrs counts all classified SQL backend errors.
+	TotalErrs uint64
+
+	// TransientErrs counts classified transient SQL backend errors.
+	TransientErrs uint64
+
+	// PermanentErrs counts classified permanent SQL backend errors.
+	PermanentErrs uint64
+
+	// FatalErrs counts classified fatal SQL backend errors.
+	FatalErrs uint64
+
+	// Serialization counts serialization failures.
+	Serialization uint64
+
+	// Deadlocks counts deadlock failures.
+	Deadlocks uint64
+
+	// Busy counts SQLite busy failures.
+	Busy uint64
+
+	// Locked counts lock-not-available failures.
+	Locked uint64
+
+	// Unavailable counts backend availability failures.
+	Unavailable uint64
+
+	// PoolExhausted counts connection exhaustion failures.
+	PoolExhausted uint64
+
+	// ResourceExhausted counts disk, memory, or similar failures.
+	ResourceExhausted uint64
+
+	// ReadOnly counts read-only backend failures.
+	ReadOnly uint64
+
+	// Permission counts backend permission failures.
+	Permission uint64
+
+	// Corrupt counts corruption failures.
+	Corrupt uint64
+
+	// SchemaMismatch counts schema mismatch failures.
+	SchemaMismatch uint64
+
+	// Constraint counts backend constraint failures.
+	Constraint uint64
+
+	// Unknown counts classified backend failures with an unknown reason.
+	Unknown uint64
+}
+
+// Stats stores low-overhead atomic counters for classified SQL errors.
+type Stats struct {
+	// totalErrs counts all classified SQL backend errors.
+	totalErrs atomic.Uint64
+
+	// transientErrs counts transient SQL backend errors.
+	transientErrs atomic.Uint64
+
+	// permanentErrs counts permanent SQL backend errors.
+	permanentErrs atomic.Uint64
+
+	// fatalErrs counts fatal SQL backend errors.
+	fatalErrs atomic.Uint64
+
+	// serialization counts serialization failures.
+	serialization atomic.Uint64
+
+	// deadlocks counts deadlock failures.
+	deadlocks atomic.Uint64
+
+	// busy counts SQLite busy failures.
+	busy atomic.Uint64
+
+	// locked counts lock-not-available failures.
+	locked atomic.Uint64
+
+	// unavailable counts backend availability failures.
+	unavailable atomic.Uint64
+
+	// poolExhausted counts connection exhaustion failures.
+	poolExhausted atomic.Uint64
+
+	// resourceExhausted counts resource exhaustion failures.
+	resourceExhausted atomic.Uint64
+
+	// readOnly counts read-only backend failures.
+	readOnly atomic.Uint64
+
+	// permission counts backend permission failures.
+	permission atomic.Uint64
+
+	// corrupt counts corruption failures.
+	corrupt atomic.Uint64
+
+	// schemaMismatch counts schema mismatch failures.
+	schemaMismatch atomic.Uint64
+
+	// constraint counts backend constraint failures.
+	constraint atomic.Uint64
+
+	// unknown counts classified backend failures with an unknown reason.
+	unknown atomic.Uint64
+}
+
+// reasonRecorder updates one per-reason counter on Stats.
+type reasonRecorder func(*Stats)
+
+// reasonRecorders maps each valid reason to its per-reason counter update.
+var reasonRecorders = [...]reasonRecorder{
+	ReasonSerialization: func(s *Stats) {
+		s.serialization.Add(1)
+	},
+	ReasonDeadlock: func(s *Stats) {
+		s.deadlocks.Add(1)
+	},
+	ReasonBusy: func(s *Stats) {
+		s.busy.Add(1)
+	},
+	ReasonLocked: func(s *Stats) {
+		s.locked.Add(1)
+	},
+	ReasonUnavailable: func(s *Stats) {
+		s.unavailable.Add(1)
+	},
+	ReasonPoolExhausted: func(s *Stats) {
+		s.poolExhausted.Add(1)
+	},
+	ReasonSchemaMismatch: func(s *Stats) {
+		s.schemaMismatch.Add(1)
+	},
+	ReasonConstraint: func(s *Stats) {
+		s.constraint.Add(1)
+	},
+	ReasonUnknown: func(s *Stats) {
+		s.unknown.Add(1)
+	},
+	ReasonResourceExhausted: func(s *Stats) {
+		s.resourceExhausted.Add(1)
+	},
+	ReasonReadOnly: func(s *Stats) {
+		s.readOnly.Add(1)
+	},
+	ReasonPermission: func(s *Stats) {
+		s.permission.Add(1)
+	},
+	ReasonCorrupt: func(s *Stats) {
+		s.corrupt.Add(1)
+	},
+}
+
+// Record updates counters from a classified SQL backend error.
+func (s *Stats) Record(err error) {
+	var sqlErr *SQLError
+	if !errors.As(err, &sqlErr) {
+		return
+	}
+
+	s.totalErrs.Add(1)
+
+	switch sqlErr.Class() {
+	case ClassTransient:
+		s.transientErrs.Add(1)
+
+	case ClassPermanent:
+		s.permanentErrs.Add(1)
+
+	case ClassFatal:
+		s.fatalErrs.Add(1)
+	}
+
+	s.recordReason(sqlErr.Reason)
+}
+
+// Snapshot returns a read-only copy of the current SQL error counters.
+func (s *Stats) Snapshot(backend Backend) StatsSnapshot {
+	return StatsSnapshot{
+		Backend:           backend,
+		TotalErrs:         s.totalErrs.Load(),
+		TransientErrs:     s.transientErrs.Load(),
+		PermanentErrs:     s.permanentErrs.Load(),
+		FatalErrs:         s.fatalErrs.Load(),
+		Serialization:     s.serialization.Load(),
+		Deadlocks:         s.deadlocks.Load(),
+		Busy:              s.busy.Load(),
+		Locked:            s.locked.Load(),
+		Unavailable:       s.unavailable.Load(),
+		PoolExhausted:     s.poolExhausted.Load(),
+		ResourceExhausted: s.resourceExhausted.Load(),
+		ReadOnly:          s.readOnly.Load(),
+		Permission:        s.permission.Load(),
+		Corrupt:           s.corrupt.Load(),
+		SchemaMismatch:    s.schemaMismatch.Load(),
+		Constraint:        s.constraint.Load(),
+		Unknown:           s.unknown.Load(),
+	}
+}
+
+// recordReason updates counters for one classified SQL error reason.
+func (s *Stats) recordReason(reason Reason) {
+	if !reason.Valid() {
+		s.unknown.Add(1)
+		return
+	}
+
+	record := reasonRecorders[reason]
+	if record == nil {
+		s.unknown.Add(1)
+		return
+	}
+
+	record(s)
+}
--- a/wallet/internal/db/pg/errors.go
+++ b/wallet/internal/db/pg/errors.go
@@ -1,0 +1,139 @@
+// Package pg contains PostgreSQL-specific SQL helpers.
+package pg
+
+import (
+	"errors"
+
+	dberr "github.com/btcsuite/btcwallet/wallet/internal/db/err"
+	"github.com/jackc/pgx/v5/pgconn"
+)
+
+// SQLSTATE helper constants support PostgreSQL error classification.
+const (
+	// connectionExceptionClass identifies PostgreSQL SQLSTATE class 08,
+	// which covers connection exceptions.
+	connectionExceptionClass = "08"
+
+	// sqlStateClassLen is the length of a SQLSTATE class prefix.
+	sqlStateClassLen = 2
+)
+
+// SQLSTATE code constants capture the PostgreSQL errors mapped here.
+const (
+	// The SQLSTATE codes below follow PostgreSQL's documented error code
+	// appendix. The wallet intentionally collapses many backend-specific
+	// codes into a smaller caller-facing reason set.
+	//
+	// Reference:
+	// https://www.postgresql.org/docs/current/errcodes-appendix.html
+	codeSerializationFailure = "40001"
+	codeDeadlockDetected     = "40P01"
+	codeLockNotAvailable     = "55P03"
+	codeQueryCanceled        = "57014"
+	codeAdminShutdown        = "57P01"
+	codeCrashShutdown        = "57P02"
+	codeCannotConnectNow     = "57P03"
+	codeTooManyConnections   = "53300"
+	codeDiskFull             = "53100"
+	codeOutOfMemory          = "53200"
+	codeConfigLimitExceeded  = "53400"
+	codeReadOnlyTxn          = "25006"
+	codeInsufficientPriv     = "42501"
+	codeCorruptData          = "XX001"
+	codeCorruptIndex         = "XX002"
+	codeUndefinedTable       = "42P01"
+	codeUndefinedColumn      = "42703"
+	codeUniqueViolation      = "23505"
+	codeForeignKeyViolation  = "23503"
+	codeCheckViolation       = "23514"
+	codeNotNullViolation     = "23502"
+	codeExclusionViolation   = "23P01"
+)
+
+// reasonByCode maps PostgreSQL SQLSTATE codes into the shared SQL error model.
+var reasonByCode = map[string]dberr.Reason{
+	codeSerializationFailure: dberr.ReasonSerialization,
+	codeDeadlockDetected:     dberr.ReasonDeadlock,
+	codeLockNotAvailable:     dberr.ReasonLocked,
+	codeQueryCanceled:        dberr.ReasonUnknown,
+	codeAdminShutdown:        dberr.ReasonUnavailable,
+	codeCrashShutdown:        dberr.ReasonUnavailable,
+	codeCannotConnectNow:     dberr.ReasonUnavailable,
+	codeTooManyConnections:   dberr.ReasonPoolExhausted,
+	codeDiskFull:             dberr.ReasonResourceExhausted,
+	codeOutOfMemory:          dberr.ReasonResourceExhausted,
+	codeConfigLimitExceeded:  dberr.ReasonResourceExhausted,
+	codeReadOnlyTxn:          dberr.ReasonReadOnly,
+	codeInsufficientPriv:     dberr.ReasonPermission,
+	codeCorruptData:          dberr.ReasonCorrupt,
+	codeCorruptIndex:         dberr.ReasonCorrupt,
+	codeUndefinedTable:       dberr.ReasonSchemaMismatch,
+	codeUndefinedColumn:      dberr.ReasonSchemaMismatch,
+	codeUniqueViolation:      dberr.ReasonConstraint,
+	codeForeignKeyViolation:  dberr.ReasonConstraint,
+	codeCheckViolation:       dberr.ReasonConstraint,
+	codeNotNullViolation:     dberr.ReasonConstraint,
+	codeExclusionViolation:   dberr.ReasonConstraint,
+}
+
+// mapErr maps PostgreSQL driver and transport errors into SQLError.
+func mapErr(err error) *dberr.SQLError {
+	// Prefer SQLSTATE-based mapping first so a completed PostgreSQL statement
+	// keeps its specific backend code instead of being collapsed into a generic
+	// transport fallback.
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) {
+		return mapCode(pgErr.Code, err)
+	}
+
+	var connectErr *pgconn.ConnectError
+
+	// ConnectError means the driver failed while establishing or maintaining
+	// the connection, so callers see the same transient unavailable bucket used
+	// for other connection-path failures.
+	if errors.As(err, &connectErr) {
+		return dberr.NewSQLError(
+			dberr.BackendPostgres, dberr.ReasonUnavailable, "", err,
+		)
+	}
+
+	// Safe-to-retry and timeout failures happen on the connection or transport
+	// path rather than as a completed SQL statement outcome, so they share the
+	// transient backend-unavailable bucket with SQLSTATE connection exceptions.
+	if pgconn.SafeToRetry(err) || pgconn.Timeout(err) {
+		return dberr.NewSQLError(
+			dberr.BackendPostgres, dberr.ReasonUnavailable, "", err,
+		)
+	}
+
+	return nil
+}
+
+// mapCode maps one PostgreSQL SQLSTATE into SQLError.
+func mapCode(code string, err error) *dberr.SQLError {
+	reason, ok := reasonByCode[code]
+	if ok {
+		return dberr.NewSQLError(dberr.BackendPostgres, reason, code, err)
+	}
+
+	if sqlStateClass(code) == connectionExceptionClass {
+		// SQLSTATE class 08 reports connection exceptions rather than completed
+		// statement outcomes, so it maps to the shared unavailable reason.
+		return dberr.NewSQLError(
+			dberr.BackendPostgres, dberr.ReasonUnavailable, code, err,
+		)
+	}
+
+	return dberr.NewSQLError(
+		dberr.BackendPostgres, dberr.ReasonUnknown, code, err,
+	)
+}
+
+// sqlStateClass returns the SQLSTATE class prefix.
+func sqlStateClass(code string) string {
+	if len(code) < sqlStateClassLen {
+		return ""
+	}
+
+	return code[:sqlStateClassLen]
+}
--- a/wallet/internal/db/runtime/runtime.go
+++ b/wallet/internal/db/runtime/runtime.go
@@ -23,6 +23,12 @@ var (
 	// ErrAmbiguousTxCommit matches commit failures whose final database outcome
 	// is unknown because the commit may already have reached the backend before
 	// the client observed the failure.
+	//
+	// Callers should treat this differently from an ordinary commit failure and
+	// should not blindly retry the transaction. Write returns this through
+	// AmbiguousTxCommitError so callers can use errors.Is(err,
+	// ErrAmbiguousTxCommit) for policy checks while still inspecting the
+	// underlying classified backend error.
 	ErrAmbiguousTxCommit = errors.New("sql ambiguous tx commit")
 )
 
@@ -34,8 +40,8 @@ var (
 	errReadDelayOrder  = errors.New("read base delay exceeds max delay")
 )
 
-// AmbiguousTxCommitError wraps a classified commit error returned when Commit
-// fails after the final database outcome may already be decided by the
+// AmbiguousTxCommitError wraps a classified commit error returned by Write when
+// Commit fails after the final database outcome may already be decided by the
 // backend.
 //
 // This lets callers ask two independent questions about the same failure:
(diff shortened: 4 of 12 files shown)


wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,415 @@
...
+	// ErrSQLConstraint matches backend constraint failures.
+	ErrSQLConstraint = errors.New("sql constraint failure")
+)
+
+const (
+	// unknownString is the shared string form for unknown DB values.
+	unknownString = "unknown"
+
+	// sqlStateClassPrefixLen is the length of a SQLSTATE class prefix.
+	sqlStateClassPrefixLen = 2
```
yyforyongyu: add some example error code here?

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,415 @@
...
+
+	// sqlitePrimaryCodeMask strips a SQLite extended result code to its primary
+	// result code.
+	sqlitePrimaryCodeMask = 0xff
```
yyforyongyu: add some example errors code here?

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,415 @@
...
+
+const (
+	// DBBackendPostgres identifies the PostgreSQL SQL backend.
+	DBBackendPostgres = "postgres"
+
+	// DBBackendSQLite identifies the SQLite SQL backend.
+	DBBackendSQLite = "sqlite"
+)
+
+var (
```
yyforyongyu: like i said, it should be well-documented, so pls add detailed docs explaining why these errors.

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,415 @@
...
+	ErrorClassTransient
+
+	// ErrorClassPermanent marks errors that should fail immediately.
+	ErrorClassPermanent
+
+	// ErrorClassFatal marks errors that should take the store out of service.
+	ErrorClassFatal
+)
+
+var errorClassStrings = [...]string{
```
yyforyongyu: don't do this, instead, add a new `errorClassSentinel` in the end of the `const` above, and we can check against this sentinel

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,415 @@
...
+	ErrorReasonCorrupt
+
+	// ErrorReasonSchemaMismatch marks schema mismatch failures.
+	ErrorReasonSchemaMismatch
+
+	// ErrorReasonConstraint marks backend constraint failures.
+	ErrorReasonConstraint
+)
+
+var errorReasonStrings = [...]string{
```
yyforyongyu: same here, add a `errorReasonSentinel`, remember to keep it private!

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,415 @@
...
+	if contextErr != nil {
+		return contextErr
+	}
+
+	existing := GetDBError(err)
+	if existing != nil {
+		return enrichDBError(existing, backend)
+	}
+
+	postgresErr := classifyPostgresError(backend, err)
```
yyforyongyu: why are we mixing both sqlite and postgres error checking in one method, shouldn't we perform a `if` check first to find out the backend here?

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,415 @@
...
+		Backend: backend,
+		Class:   class,
+		Reason:  reason,
+		Code:    code,
+		Err:     err,
+	}
+}
+
+// enrichDBError fills in any missing backend name on an existing SQL error.
+func enrichDBError(dbErr *Error, backend string) *Error {
```
yyforyongyu: be specific - rename it to patchBackendName? or `padDBName`?

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,415 @@
...
+func sqliteCodeString(code int) string {
+	return strconv.Itoa(code)
+}
+
+// sqlitePrimaryCode strips a SQLite extended result code to its primary code.
+func sqlitePrimaryCode(code int) int {
+	return code & sqlitePrimaryCodeMask
+}
+
+var errorClassesBySentinel = map[error]ErrorClass{
```
yyforyongyu: remove these vars and use the xxxSentinel

wallet/internal/db/dberrors_pg.go
```diff
@@ -0,0 +1,198 @@
...
+// pgErrorSpec defines the wallet-facing class and reason for a SQLSTATE.
+type pgErrorSpec struct {
+	// class identifies how callers should react to the SQLSTATE.
+	class ErrorClass
+
+	// reason identifies the stable cause bucket for the SQLSTATE.
+	reason ErrorReason
+}
+
+var pgErrorSpecs = map[string]pgErrorSpec{
```
yyforyongyu: missing detailed docs on this var

wallet/internal/db/sqlite_dberrors.go
```diff
@@ -0,0 +1,92 @@
+package db
+
+import (
+	"errors"
+
+	"modernc.org/sqlite"
+	sqlite3 "modernc.org/sqlite/lib"
+)
+
+var sqliteErrorSpecs = map[int]errorSpec{
```
yyforyongyu: missing detailed docs on this var

wallet/internal/db/postgres_dberrors.go
```diff
@@ -0,0 +1,198 @@
...
+
+var pgErrorSpecs = map[string]pgErrorSpec{
+	// Serialization and deadlock failures are transient because
+	// retrying the same work may succeed once conflicting activity
+	// completes.
+	pgCodeSerializationFailure: {
+		class:  ErrorClassTransient,
+		reason: ErrorReasonSerialization,
+	},
+	pgCodeDeadlockDetected: {
```
yyforyongyu: each of the errors here should have some docs explain the error and when it can happen, remember to add a newline above.

wallet/internal/db/sqlite_dberrors.go
```diff
@@ -0,0 +1,92 @@
...
+	},
+	sqlite3.SQLITE_LOCKED: {
+		class:  ErrorClassTransient,
+		reason: ErrorReasonLocked,
+	},
+	sqlite3.SQLITE_INTERRUPT: {
+		class:  ErrorClassTransient,
+		reason: ErrorReasonUnknown,
+	},
+	sqlite3.SQLITE_CONSTRAINT: {
```
yyforyongyu: each of the errors here should have some docs explain the error and when it can happen, remember to add a newline above.

wallet/internal/db/dbstats.go
```diff
@@ -0,0 +1,232 @@
...
+	schemaMismatch atomic.Uint64
+
+	// constraint counts backend constraint failures.
+	constraint atomic.Uint64
+
+	// unknown counts classified backend failures with an unknown reason.
+	unknown atomic.Uint64
+}
+
+var reasonRecorders = map[ErrorReason]func(*dbStats){
```
yyforyongyu: missing docs

yyforyongyu: and the `reasonRecorders` should be generated from `errorReasonSentinel` once it's added, we don't wanna hardcode the errors here which is error prone? so we can create a method and using `switch`?

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

wallet/internal/db/postgres_dberrors.go
```diff
@@ -0,0 +1,198 @@
...
+	// Schema and constraint failures are permanent for the current request.
+	pgCodeUndefinedTable: {
+		class:  ErrorClassPermanent,
+		reason: ErrorReasonSchemaMismatch,
+	},
+	pgCodeUndefinedColumn: {
+		class:  ErrorClassPermanent,
+		reason: ErrorReasonSchemaMismatch,
+	},
+	pgCodeUniqueViolation: {
```
yyforyongyu: This constraint mapping still misses several standard integrity-constraint SQLSTATEs, most notably `23502` (`not_null_violation`). Those errors currently fall through to `ErrorReasonUnknown`, so callers and the new stats counters will misclassify real Postgres constraint failures. Please include the full constraint class coverage you want to bucket as `constraint`, not just unique/foreign-key/check. (`gpt-5.4`)

wallet/db_stats.go
```diff
@@ -0,0 +1,18 @@
+package wallet
+
+import db "github.com/btcsuite/btcwallet/wallet/internal/db"
+
+// DBStats returns a read-only snapshot of SQL backend error counters. The
+// boolean result reports whether the active wallet store exposes SQL stats.
+func (w *Wallet) DBStats() (db.StatsSnapshot, bool) {
+	if w == nil || w.store == nil {
```
yyforyongyu: The new startup counters are not observable through this API. Both SQL constructors record ping/migration failures into a temporary store struct and then return `nil, err`, so `w.store` is never populated on those failure paths and `DBStats()` can only report counters from a successfully opened store. That means the new “record startup failures” behavior is invisible to callers. (`gpt-5.4`)

yyforyongyu (commented):
LGTM 🛡️

Error classification is correct — verified all PG SQLSTATE codes and SQLite result codes against official docs. Classification ordering in `classifySQLError` is sound (context errors first, then backend-specific, then generic connection, then permanent fallback). Stats counters are atomic. `Error.Is` correctly matches both class and reason sentinels.

Two minor observations (not bugs):
1. Stats recorded during constructor error paths (`PingContext`/`ApplyMigrations` failure) are discarded since the store is never returned.
2. A few common PG constraint codes (`23502`, `23P01`) are not explicitly mapped but fall through correctly to permanent/unknown.

**(`claude-opus-4-6`)**

wallet/internal/db/dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dberrors_pg.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/sqlite_dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/postgres_dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/sqlite_dberrors.go
yyforyongyu: Addressed in the latest commit.

wallet/internal/db/dbstats.go
yyforyongyu: Addressed in the latest commit.

yyforyongyu: Addressed in the latest commit.

yyforyongyu:
I've addressed both observations in the latest commit:
1. Mapped `23502` (not_null_violation) and `23P01` (exclusion_violation) to `ErrorClassPermanent` and `ErrorReasonConstraint`.
2. Modified the DB constructors (`NewPostgresStore` and `NewSqliteStore`) to return the instantiated store alongside the initialization error during PingContext/ApplyMigrations failures, allowing callers to inspect `store.DBStatsSnapshot()`. Tested via `db_connectors_test.go`.

yyforyongyu (commented):
LGTM 🎲

The PR now includes:
- Detailed docs for error sentinels explaining when they occur
- Added  and  private sentinel types for type checking
- Switched  to a switch-based approach derived from 
- Clear documentation on pg and sqlite error specs
- Backend resolution happens before classification in 

Good work addressing all the feedback! ()

wallet/internal/db/postgres_dberrors.go
yyforyongyu: Addressed in the latest commit. (`gemini-3.1-pro-preview`)

wallet/db_stats.go
yyforyongyu: Addressed in the latest commit. (`gemini-3.1-pro-preview`)

yyforyongyu:

yyforyongyu (commented):
LGTM 🎲

The PR now includes:
- Detailed docs for error sentinels explaining when they occur
- Added errorClassSentinel and errorReasonSentinel private sentinel types for type checking
- Switched reasonRecorders to a switch-based approach derived from errorReasonSentinel
- Clear documentation on pg and sqlite error specs
- Backend resolution happens before classification in classifySQLError

Good work addressing all the feedback! (glm5)

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,468 @@
...
+	// ErrorClassTransient marks errors that may succeed on retry.
+	ErrorClassTransient ErrorClass = iota
+
+	// ErrorClassPermanent marks errors that should fail immediately.
+	ErrorClassPermanent
+
+	// ErrorClassFatal marks errors that should take the store out of service.
+	ErrorClassFatal
+
+	sentinelErrorClass
```
yyforyongyu: missing docs explaining what this is and why do we need it

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,468 @@
...
+)
+
+// String returns the canonical string form of the error class.
+func (c ErrorClass) String() string {
+	switch c {
```
yyforyongyu: every switch should have a newline for each case?

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,468 @@
...
+//
+// The purpose of the reason is to answer the diagnostic question: what kind
+// of backend failure happened once the caller already knows the class-level
+// policy. Reasons are grouped by the class they map to, and callers can
+// derive a stable ErrorClass from any known reason.
+type ErrorReason uint8
+
+const (
+	// Transient reasons.
+
```
yyforyongyu: we should add a `//` here to keep the docs aligned!

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,468 @@
...
+	// ErrorReasonUnavailable marks backend availability failures.
+	ErrorReasonUnavailable
+
+	// ErrorReasonPoolExhausted marks pool exhaustion failures.
+	ErrorReasonPoolExhausted
+
+	sentinelTransientReason
+
+	// Permanent reasons.
+
```
yyforyongyu: missing //

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,468 @@
...
+	// ErrorReasonConstraint marks backend constraint failures.
+	ErrorReasonConstraint
+
+	// ErrorReasonUnknown is the fallback reason for uncategorized failures.
+	ErrorReasonUnknown
+
+	sentinelPermanentReason
+
+	// Fatal reasons.
+
```
yyforyongyu: missing //

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,468 @@
...
+	ErrorReasonReadOnly
+
+	// ErrorReasonPermission marks backend permission failures.
+	ErrorReasonPermission
+
+	// ErrorReasonCorrupt marks corruption failures.
+	ErrorReasonCorrupt
+
+	sentinelFatalReason
+	sentinelErrorReason
```
yyforyongyu: missing docs for each, and why do we need `sentinelErrorReason`?

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,468 @@
...
+		return ErrorClassFatal
+
+	default:
+		return ErrorClassPermanent
+	}
+}
+
+// String returns the canonical string form of the error reason.
+//
+//nolint:cyclop
```
yyforyongyu: should have comments explaining the skip?

wallet/internal/db/dberrors.go
```diff
@@ -0,0 +1,468 @@
...
+func (r ErrorReason) String() string {
+	switch r {
+	case ErrorReasonSerialization:
```
yyforyongyu: each case should have a newline above!!

wallet/internal/db/dbstats.go
```diff
@@ -0,0 +1,220 @@
...
+		// Unknown classes do not increment a class bucket.
+	}
+
+	s.recordReason(err.Reason)
+}
+
+// recordReason updates counters for one classified SQL error reason.
+//
+//nolint:cyclop
+func (s *dbStats) recordReason(reason ErrorReason) {
```
yyforyongyu: again whenever we skip a linter, we must give it a reason, and the switch cases must have newlines

wallet/internal/db/dbstats.go
```diff
@@ -0,0 +1,220 @@
...
+		SchemaMismatch:    s.schemaMismatch.Load(),
+		Constraint:        s.constraint.Load(),
+		Unknown:           s.unknown.Load(),
+	}
+}
+
+// recordStatsError records err when it is a classified SQL backend error.
+func recordStatsError(stats *dbStats, err error) {
+	dbErr := GetDBError(err)
+	if dbErr == nil {
```
yyforyongyu: recordError already checks this? so we should have a single place checking it? decide one

wallet/internal/db/migrations.go
```diff
@@ -76,8 +81,9 @@ func ApplySQLiteMigrations(db *sql.DB) error {
 //
 // NOTE: not ready for production use.
 func ApplyPostgresMigrations(db *sql.DB) error {
-	return applyMigrations(db, postgresFS, "migrations/postgres",
-		"postgres", func(db *sql.DB) (database.Driver, error) {
+	return applyMigrations(
+		db, postgresFS, "migrations/postgres", "postgres",
```
yyforyongyu: we already pass `postgres` here, why do we need yet another var `DBBackendPostgres`?

wallet/internal/db/sqlite.go
```diff
@@ -44,7 +57,12 @@ func NewSqliteStore(ctx context.Context, cfg SqliteConfig) (*SqliteStore,
 	err = db.PingContext(connCtx)
 	if err != nil {
 		_ = db.Close()
-		return nil, fmt.Errorf("ping database: %w", err)
+		classifiedErr := ClassifyError(
+			DBBackendSQLite, fmt.Errorf("ping database: %w", err),
+		)
+		recordStatsError(&store.stats, classifiedErr)
```
yyforyongyu: I think `recordStatsError` should be put into `ClassifyError` instead?

yyforyongyu (commented):
LGTM 🎲

All review comments have been addressed. The PR now has:

- Clean package structure with shared `err` package and backend-specific `postgres`/`sqlite` packages
- Well-documented enums with `Valid()` methods for backend, class, and reason
- Clear `Reason.Class()` method deriving policy from reason
- Proper normalization helper that stays side-effect free
- README documenting package ownership boundaries

Good work on the iterations! (`glm5`)

yyforyongyu:

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

yyforyongyu (commented):
LGTM 🛡️

All 21 PG SQLSTATE codes and 13 SQLite result codes verified against official docs — correctly classified. `Reason.Class()` range partitioning is correct. `Normalize()` ordering is sound (context first, existing wrappers preserved, backend-specific before generic transport, permanent fallback). `pgconn.PgError` is matched before `SafeToRetry`/`Timeout` so SQLSTATE codes are never shadowed. Stats counters atomic throughout. Retry loop boundaries correct with proper overflow handling in `retryDelay`. All 13 review threads resolved.

**(`claude-opus-4-6`)**

yyforyongyu:
LGTM 🛡️ (qwen3.5-397b-a17b)

All review threads resolved. The PR now includes:
- Comprehensive error classification with detailed docs for each sentinel
- Private sentinel types (`errorClassSentinel`, `errorReasonSentinel`) for type checking
- Switch-based error classification with proper backend resolution
- Complete SQLSTATE coverage for constraint violations (23502, 23503, 23504, 23505, 23514, 23P01)
- Stats API exposing startup failures via `DBStats()`
- Proper comment formatting throughout

Error classification logic is sound and well-documented.

yyforyongyu (commented):
LGTM 🛡️ **(`claude-opus-4-6`)**

yyforyongyu:

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,442 @@
...
+	// Preserve caller-driven cancellation semantics instead of wrapping them as
+	// backend errors.
+	contextErr := unwrapContextErr(err)
+	if contextErr != nil {
+		return contextErr
+	}
+
+	// Preserve existing SQL wrappers while filling in a missing backend
+	// when the caller provides one.
+	existing := extractSQLError(err)
```
yyforyongyu: This branch now normalizes any wrapped `*SQLError` by extracting it with `errors.As` and returning the inner SQL error directly. That drops outer context such as `fmt.Errorf("open wallet: %w", sqlErr)`, so callers lose the wrapper text they added before normalization. Please preserve the original error chain when only the backend patching path applies. (`gpt-5.4`)

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+
+	// Commit transport failures are classified and then wrapped so callers know
+	// the final transaction outcome is unknown.
+	err = tx.Commit()
+	if err != nil {
+		classifiedErr := hooks.ClassifyError(err)
+		hooks.RecordError(classifiedErr)
+
+		if isCommitTransportError(err) {
```
yyforyongyu, line 367: Ambiguous commit detection is still keyed only off raw transport-shaped errors from `tx.Commit()`. If commit returns a backend-classified availability error such as PostgreSQL SQLSTATE class `08`, this path records it as unavailable but not ambiguous even though the final commit outcome is still unknown. Please let classified unavailable commit errors trigger the ambiguous-commit path too. (`gpt-5.4`)

yyforyongyu (commented):
LGTM 🎲

The PR introduces a comprehensive SQL error handling framework with:

- Shared error taxonomy in `db/err` package
- Backend-specific mappers for PostgreSQL and SQLite
- Runtime helpers for retry and transaction execution
- Well-documented error classification and stats collection

Minor observations (not blocking):

1. `Reason.Class()` uses range-based classification that depends on enum ordering - consider using a map or explicit switch for better maintainability
2. `retryDelay` uses left shift which could overflow with large attempt values - consider using multiplication or limiting max attempts
3. Some SQLite error codes are not mapped (e.g., SQLITE_PROTOCOL, SQLITE_NOTFOUND) - these are less common but could be added for completeness
4. `isCommitTransportError` declares `netErr` variable that's never used - the `errors.As` call is correct but the variable is confusing

Overall, the code is well-structured, well-tested, and follows the repo's conventions. (`z-ai/glm4.7`)

yyforyongyu:
The `Is` method at `runtime.go:63-65` returns `errors.Is(target, ErrAmbiguousTxCommit)`, which checks if `target` is equivalent to `ErrAmbiguousTxCommit`. This means `AmbiguousTxCommitError{Err: someOtherError}.Is(ErrAmbiguousTxCommit)` would return `true` (since `errors.Is(ErrAmbiguousTxCommit, ErrAmbiguousTxCommit)` is `true`), even though the wrapped error is unrelated to ambiguous commits.

The correct implementation should be:

```go
func (e *AmbiguousTxCommitError) Is(target error) bool {
	return target == ErrAmbiguousTxCommit
}
```

This ensures `AmbiguousTxCommitError` is only considered equivalent to `ErrAmbiguousTxCommit`, not to arbitrary wrapped errors.
`dummy`

yyforyongyu (commented):
Finding 1: AmbiguousTxCommitError.Is implementation issue

Location: wallet/internal/db/runtime/runtime.go:63-65

The Is method returns errors.Is(target, ErrAmbiguousTxCommit), which checks if target is equivalent to ErrAmbiguousTxCommit. This means AmbiguousTxCommitError{Err: someOtherError}.Is(ErrAmbiguousTxCommit) would incorrectly return true (since errors.Is(ErrAmbiguousTxCommit, ErrAmbiguousTxCommit) is true), even though the wrapped error is unrelated to ambiguous commits.

The correct implementation should be:

    func (e *AmbiguousTxCommitError) Is(target error) bool {
        return target == ErrAmbiguousTxCommit
    }

This ensures AmbiguousTxCommitError is only considered equivalent to ErrAmbiguousTxCommit, not to arbitrary wrapped errors.

yyforyongyu (commented):
The `Reason.Class()` method uses range-based classification that depends on the order of enum constants. This is fragile because if someone adds a new reason in the wrong position, it could break the classification. Consider using a map or explicit switch statement instead. (`z-ai/glm4.7`)

yyforyongyu (commented):
The `retryDelay` function uses left shift which could overflow with large attempt values. Consider using multiplication or limiting max attempts to prevent overflow. (`z-ai/glm4.7`)

yyforyongyu (commented):
Some SQLite error codes are not mapped (e.g., SQLITE_PROTOCOL, SQLITE_NOTFOUND). These are less common but could be added for completeness. (`z-ai/glm4.7`)

yyforyongyu (commented):
The `isCommitTransportError` function declares `netErr` variable that's never used. The `errors.As` call is correct but the variable is confusing. Consider removing it or using it explicitly. (`z-ai/glm4.7`)

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,442 @@
...
+	ReasonReadOnly
+
+	// ReasonPermission marks backend permission failures.
+	ReasonPermission
+
+	// ReasonCorrupt marks corruption failures.
+	ReasonCorrupt
+)
+
+// Class returns the caller-facing policy bucket derived from the reason.
```
yyforyongyu, line 178: The `Reason.Class()` method uses range-based classification that depends on the order of enum constants. This is fragile because if someone adds a new reason in the wrong position, it could break the classification. Consider using a map or explicit switch statement instead. (`z-ai/glm4.7`)

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,363 @@
...
+		}
+
+		return fmt.Errorf("commit tx: %w", classifiedErr)
+	}
+
+	return nil
+}
+
+// retryDelay applies bounded exponential backoff for retry attempts.
+func retryDelay(attempt int, baseDelay, maxDelay time.Duration) time.Duration {
```
yyforyongyu, line 404: The `retryDelay` function uses left shift which could overflow with large attempt values. Consider using multiplication or limiting max attempts to prevent overflow. (`z-ai/glm4.7`)

wallet/internal/db/sqlite/errors.go
```diff
@@ -0,0 +1,64 @@
...
+// SQLite result-code helper constants support error classification.
+const (
+	// primaryCodeMask strips a SQLite extended result code to its primary
+	// result code.
+	primaryCodeMask = 0xff
+)
+
+// reasonByCode maps SQLite primary result codes into the shared SQL error
+// model.
+var reasonByCode = map[int]dberr.Reason{
```
yyforyongyu, line 21: Some SQLite error codes are not mapped (e.g., SQLITE_PROTOCOL, SQLITE_NOTFOUND). These are less common but could be added for completeness. (`z-ai/glm4.7`)

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,363 @@
...
+// isCommitTransportError reports whether commit failed after the request may
+// already have reached the backend.
+func isCommitTransportError(err error) bool {
+	if errors.Is(err, driverErrBadConn()) || errors.Is(err, sql.ErrConnDone) ||
+		errors.Is(err, io.EOF) {
+
+		return true
+	}
+
+	var netErr netError
```
yyforyongyu: The `isCommitTransportError` function declares `netErr` variable that's never used. The `errors.As` call is correct but the variable is confusing. Consider removing it or using it explicitly. (`z-ai/glm4.7`)

wallet/internal/db/err/errors.go
yyforyongyu: Fixed. Normalize now only patches a missing backend when the top-level error itself is `*SQLError`; wrapped SQL errors return unchanged so outer context stays in the chain. Folded into rewritten commit `2e1461f5e`.

wallet/internal/db/runtime/runtime.go
yyforyongyu: Fixed. Commit handling now treats classified `ReasonUnavailable` commit failures as ambiguous even when the raw transport-shaped sentinel is gone. Folded into rewritten commit `a08b30a9a`.

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,363 @@
...
+	case config.MaxAttempts <= 0:
+		return readConfig{}, errReadMaxAttempts
+
+	case config.BaseDelay <= 0:
+		return readConfig{}, errReadBaseDelay
+
+	case config.MaxDelay <= 0:
+		return readConfig{}, errReadMaxDelay
+
+	case config.BaseDelay > config.MaxDelay:
```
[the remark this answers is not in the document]
yyforyongyu: Fixed. `MaxAttempts == 1` now allows zero `BaseDelay` and `MaxDelay`, and the docs call out that those values are ignored when retries are disabled. Folded into rewritten commit `a08b30a9a`.

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,363 @@
...
+	timer func(time.Duration) <-chan time.Time, delay time.Duration) error {
+
+	select {
+	case <-timer(delay):
+		return nil
+
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
```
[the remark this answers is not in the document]
yyforyongyu: Fixed. The retry timer hook now uses a stoppable `*time.Timer`, and the helper stops and drains it on cancellation instead of relying on `time.After`. Folded into rewritten commit `a08b30a9a`.

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,442 @@
...
+	contextErr := unwrapContextErr(err)
+	if contextErr != nil {
+		return contextErr
+	}
+
+	// Preserve existing SQL wrappers while filling in a missing backend
+	// when the caller provides one.
+	existing := extractSQLError(err)
+	if existing != nil {
+		return patchBackend(existing, backend)
```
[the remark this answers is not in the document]
yyforyongyu: Fixed. Normalize now preserves the original wrapped error chain and only patches the backend when the top-level error itself is the shared SQL wrapper. Folded into rewritten commit `2e1461f5e`.

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,363 @@
...
+	if e == nil {
+		return nil
+	}
+
+	return e.Err
+}
+
+// Is reports whether target matches the ambiguous commit sentinel.
+func (e *AmbiguousTxCommitError) Is(target error) bool {
+	return errors.Is(target, ErrAmbiguousTxCommit)
```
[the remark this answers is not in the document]
yyforyongyu: Fixed. `AmbiguousTxCommitError.Is` now does a direct sentinel comparison instead of calling `errors.Is` on the target. Folded into rewritten commit `a08b30a9a`.

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,363 @@
...
+func (e *AmbiguousTxCommitError) Unwrap() error {
+	if e == nil {
+		return nil
+	}
+
+	return e.Err
+}
+
+// Is reports whether target matches the ambiguous commit sentinel.
+func (e *AmbiguousTxCommitError) Is(target error) bool {
```
yyforyongyu, line 77: Fixed. `AmbiguousTxCommitError.Is` now returns `target == ErrAmbiguousTxCommit`, so it only matches the sentinel directly. Folded into rewritten commit `a08b30a9a`.

wallet/internal/db/err/errors.go
yyforyongyu: Changed this to an explicit switch, so `Reason.Class()` no longer depends on enum ordering. Folded into rewritten commit `2e1461f5e`.

wallet/internal/db/runtime/runtime.go
yyforyongyu: Fixed. `retryDelay` now uses bounded doubling without a left shift, so it saturates at `MaxDelay` without overflow behavior. Folded into rewritten commit `a08b30a9a`.

wallet/internal/db/runtime/runtime.go
yyforyongyu: Cleaned this up by factoring the `errors.As` check into `extractNetError`, so the transport helper no longer carries a throwaway net error variable. Folded into rewritten commit `a08b30a9a`.

wallet/internal/db/sqlite/errors.go
yyforyongyu: Added explicit mappings for `SQLITE_PROTOCOL` and `SQLITE_NOTFOUND`. Folded into rewritten commit `2e1461f5e`.

yyforyongyu (commented):
LGTM 🔒 **(`claude-opus-4-6`)**

yyforyongyu:

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,418 @@
...
+	return errors.Is(target, ErrAmbiguousTxCommit)
```
yyforyongyu: `AmbiguousTxCommitError.Is` is still using `errors.Is(target, ErrAmbiguousTxCommit)`, which makes `errors.Is(err, fmt.Errorf("wrap: %w", ErrAmbiguousTxCommit))` unexpectedly return true. This is also visible in this branch: `GOWORK=off go test ./wallet/internal/db/runtime -run TestHelperUtilities -count=1` currently fails at the `NotErrorIs` assertion. The method should compare the target directly (`target == ErrAmbiguousTxCommit`). (`gpt-5.3-codex`)

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,451 @@
...
+// package so backend packages can return shared SQL errors without introducing
+// import cycles.
+func Normalize(backend Backend, mapper func(error) *SQLError, err error) error {
+	if err == nil {
+		return nil
+	}
+
+	// Preserve caller-driven cancellation semantics instead of wrapping them as
+	// backend errors.
+	contextErr := unwrapContextErr(err)
```
yyforyongyu, line 356: This still returns the bare cancellation sentinel from `unwrapContextErr(err)` instead of the original wrapped error. A caller that does `fmt.Errorf("query accounts: %w", context.Canceled)` loses its added context here even though the surrounding comments say cancellation semantics should be preserved, not stripped. Please return the original error when it wraps cancellation or deadline expiration. (`gpt-5.4`)

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,418 @@
...
+
+		return fmt.Errorf("commit tx: %w", classifiedErr)
+	}
+
+	return nil
+}
+
+// isAmbiguousCommitError reports whether a commit failure leaves the final
+// transaction outcome unknown.
+func isAmbiguousCommitError(commitErr, classifiedErr error) bool {
```
yyforyongyu: This now treats every classified `ReasonUnavailable` commit error as ambiguous, but the backend mappers assign that reason to explicit server-declared failures too (for example PostgreSQL shutdown codes and SQLite protocol errors). Those are not the same as “transport died after commit may already have landed”, so this can over-report ambiguous commit outcomes. Please narrow this check to truly outcome-unknown commit failures instead of all unavailable classifications. (`gpt-5.4`)

yyforyongyu (commented):
LGTM 🎲 (`z-ai/glm4.7`)

wallet/internal/db/runtime/runtime.go
yyforyongyu: Fixed in rewritten commit `a7e6c4267`. `TestHelperUtilities` now covers the shallow match directly, and wrapped sentinels no longer match through `AmbiguousTxCommitError.Is`.

wallet/internal/db/err/errors.go
yyforyongyu: Fixed in rewritten commit `24193a665`. `unwrapContextErr` now returns the original wrapped cancellation or deadline error, and `Normalize` preserves caller-added context instead of collapsing to the bare sentinel.

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,418 @@
...
+// Unwrap returns the wrapped classified commit error.
+func (e *AmbiguousTxCommitError) Unwrap() error {
+	if e == nil {
+		return nil
+	}
+
+	return e.Err
+}
+
+// Is reports whether target matches the ambiguous commit sentinel.
```
yyforyongyu, line 76: Fixed in rewritten commit `a7e6c4267`. `AmbiguousTxCommitError.Is` now does a shallow sentinel check with `target == ErrAmbiguousTxCommit`, so wrapped versions of the sentinel no longer match.

wallet/internal/db/runtime/runtime.go
yyforyongyu: Fixed in rewritten commit `a7e6c4267`. Ambiguous commit detection is now limited to transport-level commit failures; plain classified `ReasonUnavailable` commit errors return as ordinary commit failures, with regression coverage for that case.

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

yyforyongyu:
LGTM 🚀 (`minimax-m2.7`)

yyforyongyu (commented):
LGTM 🌟 (`gpt-5.4`)

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,478 @@
...
+	ReasonCorrupt
+)
+
+// Class returns the caller-facing policy bucket derived from the reason.
+func (r Reason) Class() Class {
+	switch {
+	case !r.Valid():
+		return ClassPermanent
+
+	case r <= ReasonPoolExhausted:
```
yyforyongyu: we should add a `reasonSentinelTransient` after `ReasonPoolExhausted` so we don't need to mannually remember `ReasonPoolExhausted` is the last one in this `ClassTransient`.

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,478 @@
...
+		return ClassTransient
+
+	case r <= ReasonUnknown:
```
yyforyongyu: similarly, create a `reasonSentinelPermanent`

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,478 @@
...
+		return ClassPermanent
+
+	default:
+		return ClassFatal
```
yyforyongyu: similarly, create a `reasonSentinelFaltal`

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+}
+
+// String returns the canonical string form of the reason.
+func (r Reason) String() string {
```
yyforyongyu: don't overcomplicate it - just switch all reasons here instead, remove all the helpers like `transientReasonString`

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,436 @@
+// Package err contains the shared SQL error taxonomy and normalization logic
+// used by backend-specific db packages.
+package err
```
GustavoStingelin: I think this package name may introduce shadowing issues with common patterns like `err := someThingThatThrows()`.

Using a name like `errors` would also conflict with the standard library package, which makes it easy to misuse or confuse at call sites.

Choosing something like `dberrors` avoids the conflict, but the repetition feels a bit awkward.

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,436 @@
...
+	// caused by concurrent updates.
+	ReasonSerialization Reason = iota
+
+	// ReasonDeadlock marks deadlock failures where the backend aborts
+	// one participant to break a lock cycle.
+	ReasonDeadlock
+
+	// ReasonBusy marks SQLite busy failures where a lock cannot be
+	// acquired in time.
+	ReasonBusy
```
GustavoStingelin, line 144: Could also generalize for [pg max connections](https://www.postgresql.org/docs/current/runtime-config-connection.html#GUC-MAX-CONNECTIONS)

GustavoStingelin: I think `ReasonBusy` and `ReasonPoolExhausted` effectively fall into the same category. Both SQLite busy and Postgres max connection limits reflect temporary capacity contention. It might make sense to merge them.

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,436 @@
...
+
+	default:
+		return ClassPermanent
+	}
+}
+
+// String returns the canonical string form of the reason.
+//
+//nolint:cyclop // One explicit enum-to-string switch is the clearest form here.
+func (r Reason) String() string {
```
GustavoStingelin, line 204: non-blocking: GPT gave me this pattern, which could be good 

```
type Class uint8

const (
	ClassTransient Class = iota
	ClassPermanent
	ClassFatal

	classCount
)

const unknownString = "unknown"

var classStrings = [...]string{
	ClassTransient: "transient",
	ClassPermanent: "permanent",
	ClassFatal:     "fatal",
}

func (c Class) String() string {
	if c >= classCount {
		return unknownString
	}
	return classStrings[c]
}
```

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,436 @@
...
+
+// Error returns the printable form of the wrapped backend error.
+func (e *SQLError) Error() string {
+	if e == nil {
+		return "<nil>"
+	}
+
+	if e.Err != nil {
+		return e.Err.Error()
+	}
```
GustavoStingelin, line 309: Maybe the default should be the format below with the complete info + this Err? Otherwise, we will always unwrap by this path and lose information

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,436 @@
...
+	// ReasonSchemaMismatch marks schema drift or schema version
+	// mismatches.
+	ReasonSchemaMismatch
+
+	// ReasonConstraint marks backend constraint failures.
+	ReasonConstraint
+
+	// ReasonUnknown is the fallback reason for uncategorized backend
+	// codes.
+	ReasonUnknown
```
GustavoStingelin: Unknown could be the first iota value (zero) for the cases where the value is initialized with no value. So the default will be unknown, by e.g., `var r Reason`.

wallet/internal/db/err/stats.go
```diff
@@ -0,0 +1,226 @@
...
+
+	// Constraint counts backend constraint failures.
+	Constraint uint64
+
+	// Unknown counts classified backend failures with an unknown reason.
+	Unknown uint64
+}
+
+// Stats stores low-overhead atomic counters for classified SQL errors.
+type Stats struct {
```
GustavoStingelin, line 66: Q: I think this is a good starting point for adding metrics, but it is still limited to simple total counters. We would also need to handle logging and exposing these metrics ourselves.

There are mature solutions in the OSS ecosystem, such as Prometheus, that already provide these capabilities. They support not only counting errors, but also richer insights like timing via histograms and better dimensionality for analysis.

Have we already evaluated using something like [Prometheus](https://prometheus.io/docs/guides/go-application/#adding-your-own-metrics) or [OpenTelemetry](https://opentelemetry.io/docs/languages/go/instrumentation/#using-observable-async-updown-counters) (all written in Golang btw) for observability in LL projects? I have strong experience with this from enterprise work, especially the paid ones like DataDog.

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,475 @@
...
+
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// newRetryTimer builds the stoppable retry timer used by the runtime helpers.
+func newRetryTimer(delay time.Duration) *time.Timer {
+	return time.NewTimer(delay)
+}
```
GustavoStingelin: Looks like this function has no additional behavior?

GustavoStingelin: Unless we want to use this one to help in the tests https://github.com/lightningnetwork/lnd/tree/7024f36a76e3557bb672fc9264a306a7314d971b/clock

wallet/internal/db/pg/errors.go
```diff
@@ -0,0 +1,136 @@
...
+	// appendix. The wallet intentionally collapses many backend-specific
+	// codes into a smaller caller-facing reason set.
+	//
+	// Reference:
+	// https://www.postgresql.org/docs/current/errcodes-appendix.html
+	codeSerializationFailure = "40001"
+	codeDeadlockDetected     = "40P01"
+	codeLockNotAvailable     = "55P03"
+	codeQueryCanceled        = "57014"
+	codeAdminShutdown        = "57P01"
```
GustavoStingelin, line 33: Unfortunately, these codes are quite difficult to validate without integration tests against the actual backends. Without that, we end up relying a lot on assumptions that the mappings are correct.

I think it would be worth adding at least one or two simple test cases, for example, constraint violations, just to verify that errors are being properly captured and classified from the database side. We could also open a follow-up issue to expand this coverage later as a nice-to-have.

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,436 @@
...
+	// ReasonPoolExhausted marks connection-pool or backend
+	// connection-limit failures.
+	ReasonPoolExhausted
+
+	// ReasonSchemaMismatch marks schema drift or schema version
+	// mismatches.
+	ReasonSchemaMismatch
+
+	// ReasonConstraint marks backend constraint failures.
+	ReasonConstraint
```
GustavoStingelin, line 162: random-thoughts: Wondering whether we should separate `IndexConstraint` errors from this general one, especially for cases where concurrent inserts trigger database computed conflicts, so we could safely retry.

That said, since our insert paths are already designed with this in mind, using database-level locking strategies, I'm not sure if this additional distinction is worthwhile.

GustavoStingelin (changes requested):
I’m still reviewing this, but leaving some comments along the way.

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,436 @@
...
+	ReasonLocked
+
+	// ReasonUnavailable marks backend availability and transport
+	// failures.
+	ReasonUnavailable
+
+	// ReasonPoolExhausted marks connection-pool or backend
+	// connection-limit failures.
+	ReasonPoolExhausted
+
```
yyforyongyu, line 156: again we should define a `sentinelClassTransient` here, otherwise, every time a new error is added, we need to remember to update `Class`?

wallet/internal/db/err/errors.go
```diff
@@ -0,0 +1,436 @@
...
+// not come with backend-specific codes.
+func classifyConnErr(backend Backend, err error) *SQLError {
+	if !backend.Valid() {
+		return nil
+	}
+
+	badConn := errors.Is(err, driver.ErrBadConn)
+	connDone := errors.Is(err, sql.ErrConnDone)
+
+	if badConn || connDone || errors.Is(err, io.EOF) {
```
yyforyongyu, line 414: add inline comments above to explain this case

wallet/internal/db/pg/errors.go
```diff
@@ -0,0 +1,136 @@
...
+	codeForeignKeyViolation:  dberr.ReasonConstraint,
+	codeCheckViolation:       dberr.ReasonConstraint,
+	codeNotNullViolation:     dberr.ReasonConstraint,
+	codeExclusionViolation:   dberr.ReasonConstraint,
+}
+
+// mapErr maps PostgreSQL driver and transport errors into SQLError.
+func mapErr(err error) *dberr.SQLError {
+	var pgErr *pgconn.PgError
+	if errors.As(err, &pgErr) {
```
yyforyongyu, line 85: add inline comments to mention that we always try to use the map code to map against the predefined errors first?

wallet/internal/db/sqlite/errors.go
```diff
@@ -0,0 +1,66 @@
...
+	sqlite3.SQLITE_NOTADB:     dberr.ReasonCorrupt,
+	sqlite3.SQLITE_NOTFOUND:   dberr.ReasonUnknown,
+	sqlite3.SQLITE_SCHEMA:     dberr.ReasonSchemaMismatch,
+	sqlite3.SQLITE_CANTOPEN:   dberr.ReasonUnknown,
+}
+
+// mapErr maps SQLite result codes into SQLError.
+func mapErr(err error) *dberr.SQLError {
+	var sqliteErr *sqlite.Error
+	if !errors.As(err, &sqliteErr) {
```
yyforyongyu, line 44: similar to the `mapErr` in pg, we should also have inline docs to guide through the process here?

wallet/internal/db/runtime/runtime.go
```diff
@@ -0,0 +1,475 @@
...
+
+	// RecordRetrySuccess records a successful retry outcome.
+	RecordRetrySuccess()
+
+	// RecordRetryExhausted records that retry budget was exhausted.
+	RecordRetryExhausted()
+}
+
+// TxHooks defines the runtime hooks needed by the shared transaction helper.
+type TxHooks interface {
```
yyforyongyu: this should now be renamed to `WriteHooks`?

yyforyongyu: actually they should be `ReadHook` and `WriteHook`?

wallet/internal/db/runtime/runtime_bench_test.go
```diff
@@ -0,0 +1,64 @@
+package runtime
+
```
yyforyongyu, line 2: this commit and the previous one, 809a7b9d2590db2b1f735deeeed965449fe97432, should be reworked and be broken into three smaller commits:
- one commit to introduce the ambiguous error, and its tests
- one commit to introduce the read, and its tests
- one commit to introduce the write, and its tests

wallet/internal/db/sqlite/backend_rows_test.go
```diff
@@ -18,7 +17,7 @@ func TestCreateTxOpsAdditionalBranches(t *testing.T) {
 	t.Parallel()
 
 	req := testCreateTxRequest(t)
-	ctx := context.Background()
+	ctx := t.Context()
```
yyforyongyu: Push this to a different cleanup PR so we stay focus here?

wallet/internal/db/err/errors.go
yyforyongyu: funny it gave me the exact same pattern, didn't want to introduce too much deviation from the usual format so i chose to use the old shcool way.

wallet/internal/db/err/errors.go
yyforyongyu: hmm looks like the agent removed my comments again, will ask it to add back. but this is the common case `e.Err`, the rest are just there to be defensive.

wallet/internal/db/err/errors.go
yyforyongyu: i think for an internal pkg it is fine, especially when it's named as `err` and we actually import it via `dberr`. No strong preference here tho, as can just ask the llm to update it.

wallet/internal/db/pg/errors.go
yyforyongyu: Agreed. The current unit tests validate the mapping table itself, but they do not prove that the backend surfaces those failures as expected. I would prefer to keep this PR focused on the taxonomy and runtime split, then add real DB-backed mapping coverage in a follow-up, starting with simple constraint violations.

wallet/internal/db/sqlite/backend_rows_test.go
yyforyongyu: Done in the rewritten local history. I dropped the unrelated `t.Context()` cleanup from this PR so it stays focused on the SQL error and runtime work.

wallet/internal/db/runtime/runtime.go
yyforyongyu: Renamed this to `WriteHooks` in the rewritten local history so the runtime API now matches `Read` and `Write` consistently.

wallet/internal/db/err/errors.go
yyforyongyu: I kept the explicit switch intentionally so the class policy does not depend on enum ordering. A sentinel would move that maintenance burden into declaration order instead of removing it. To guard against omissions, I added a test that enumerates every valid Reason and asserts its expected Class mapping in the rewritten local history.

wallet/internal/db/err/errors.go
yyforyongyu: Makes sense as a possible future refinement. I left constraint failures grouped for now because they currently share the same runtime policy, and I do not have a concrete caller that benefits from splitting index-computed conflicts yet. If we add a real retry path that needs that distinction, I think that would justify splitting it out.

wallet/internal/db/err/errors.go
yyforyongyu: Added the inline comment back in the rewritten local history. `e.Err` is the normal path for real wrappers, and the remaining branches are only defensive fallbacks for malformed or test-built values.

wallet/internal/db/runtime/runtime.go
yyforyongyu: I kept `ReadHooks` and `WriteHooks` plural because each interface groups several hook methods rather than a single callback.

wallet/internal/db/runtime/runtime.go
yyforyongyu: Exactly. I kept `newRetryTimer` as a tiny wrapper intentionally so `Read` has a default production timer hook while tests can swap in deterministic timer behavior through `readConfig.timer` without adding another abstraction to the call sites. If we later want richer time control, I think a clock-style helper would be the next step rather than inlining this one.

wallet/internal/db/err/errors.go
yyforyongyu: I left them separate intentionally. `ReasonPoolExhausted` is the shared bucket for backend or pool connection-limit pressure, including PostgreSQL `too_many_connections`, while `ReasonBusy` stays specific to SQLite lock contention. They are both transient at the class level, but the separate reasons keep the diagnostics and counters more useful.

wallet/internal/db/err/errors.go
yyforyongyu: Agreed. I moved `ReasonUnknown` to the zero value in the rewritten local history so a default-initialized `Reason` no longer looks like `ReasonSerialization`.

wallet/internal/db/err/stats.go
yyforyongyu: Makes sense. I am treating these counters as a very small internal starting point for classified SQL failures rather than a full observability stack. I have not evaluated Prometheus or OpenTelemetry for this PR, and I think that broader metrics/export decision would be better handled as follow-up project-level work instead of coupling it to this package split.

yyforyongyu: ok in addition to gpt's response above, which focuses on scoping, I'd like to mention that we are unlikely to use 3rd pkgs here due to the project being a high-assurance software, and similar in lnd. Tools as such should only be used on the RPC layer, that is, when we expose RPCs, then they can use 3rd party tools to build the metrics.

wallet/internal/db/runtime/runtime.go
yyforyongyu: GPT is apparently hallucinating above - removed the thin wrapper which is useless.



> Unless we want to use this one to help in the tests https://github.com/lightningnetwork/lnd/tree/7024f36a76e3557bb672fc9264a306a7314d971b/clock

This looks weird as the commit referenced is from 7 years ago, that repo no longer looks like this...guess GPT has quite some old data.

wallet/internal/db/pg/errors.go
yyforyongyu: btw we have itests now in #1212

wallet/internal/db/runtime/runtime.go
GustavoStingelin: Holy crap, the tricky part is that I first referenced this file
[https://github.com/lightningnetwork/lnd/blob/master/clock/default_clock.go](https://github.com/lightningnetwork/lnd/blob/master/clock/default_clock.go) in an earlier comment on that commit from about 7 years ago. Later, I updated the reference to point to the package instead, which may have caused the confusion.

That said, both the old and current versions of the package look essentially the same. Also, I did not use GPT to reason about this comment, only to clean up the grammar.

GustavoStingelin (approved):
LGTM!

yyforyongyu (commented):
LGTM 🌟 (`gpt-5.4`)

PR #1201, opened 2026-03-30
wallet: migrate LabelTx to db.Store

yyforyongyu:
## Summary
- add `UpdateTx` support to the transitional `db.Store`
- route wallet transaction label updates through `w.store.UpdateTx`
- keep the remaining legacy tx-manager wrappers isolated from the new writer path

## Why
- wallet label updates are one of the last small tx-writer mutations still going through direct legacy paths
- moving them behind `db.Store` makes the tx-writer migration more complete and keeps wallet code off direct `walletdb` access
- this also establishes the mutation hook later writer/store migrations can reuse

## What Changed
- added `UpdateTx` support to `db.Store`
- updated wallet label mutation flow to use `w.store.UpdateTx`
- kept the migration incremental without changing the external wallet behavior

## Testing
- `GOWORK=off scripts/check-each-commit.sh prep-kvdb-deprecate`
- `GOWORK=off go test ./wallet/internal/db/... ./wallet`
- `GOWORK=off go test ./...` in `wtxmgr`
- `GOWORK=off make lint-check`


yyforyongyu (commented):
LGTM 🏷️ **(`claude-opus-4-6`)**

yyforyongyu:

yyforyongyu:
LGTM 🚀 (`minimax-m2.7`)

wallet/tx_writer.go
```diff
@@ -24,36 +24,21 @@ var _ TxWriter = (*Wallet)(nil)
...
-
-		return nil
+	err = w.store.UpdateTx(ctx, db.UpdateTxParams{
+		WalletID: w.id,
+		Txid:     hash,
+		Label:    &label,
 	})
 	if err != nil {
-		return fmt.Errorf("failed to update wallet db: %w", err)
+		return fmt.Errorf("update tx label: %w", err)
```
yyforyongyu, line 46: This now wraps `db.ErrTxNotFound` from `wallet/internal/db`, but callers outside this module cannot import that internal package, so they lose any way to `errors.Is` the not-found case. Before this migration `LabelTx` surfaced the wallet-level sentinel (`wallet.ErrTxNotFound`). Can we translate `db.ErrTxNotFound` back to `ErrTxNotFound` here before returning, so the public wallet API remains stable? (`gpt-5.3-codex`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -0,0 +1,155 @@
...
+	}
+
+	if params.Label == nil {
+		return nil, fmt.Errorf(
+			"kvdb.Store.UpdateTx: label patch required: %w",
+			db.ErrInvalidParam,
+		)
+	}
+
+	if *params.Label == "" {
```
yyforyongyu: The SQL backends and schema treat the empty string as a valid “no label” value, but this kvdb path now rejects `UpdateTx{Label: &""}` as not implemented. Since `Wallet.LabelTx` now routes through `store.UpdateTx`, the migrated label API behaves differently on the default kvdb backend than on the SQL backends. Please accept empty labels here too so the backends stay contract-compatible. (`gpt-5.4`)

wallet/tx_writer.go
```diff
@@ -24,36 +24,21 @@ var _ TxWriter = (*Wallet)(nil)
...
-			return ErrTxNotFound
-		}
-
-		err = w.txStore.PutTxLabel(txmgrNs, hash, label)
-		if err != nil {
-			return fmt.Errorf("failed to put tx label: %w", err)
-		}
-
-		return nil
+	err = w.store.UpdateTx(ctx, db.UpdateTxParams{
```
yyforyongyu, line 36: This now forwards backend `db.ErrTxNotFound` directly out of `Wallet.LabelTx`, but the wallet package still exposes its own `ErrTxNotFound` sentinel for transaction APIs. Callers checking `errors.Is(err, wallet.ErrTxNotFound)` will stop matching on the migrated path. Please translate the store-layer sentinel back to the wallet-layer one before returning. (`gpt-5.4`)

wallet/internal/db/kvdb/txstore_test.go
```diff
@@ -0,0 +1,110 @@
...
+	})
+	require.ErrorIs(t, err, errNotImplemented)
+
+	empty := ""
+	err = store.UpdateTx(t.Context(), db.UpdateTxParams{
+		WalletID: 0,
+		Txid:     chainhash.Hash{1},
+		Label:    &empty,
+	})
+	require.ErrorIs(t, err, errNotImplemented)
```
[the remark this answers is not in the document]
yyforyongyu: Updated the kvdb tests accordingly. The empty-label case now verifies that a previously stored label is cleared rather than rejected.

wallet/tx_writer.go
```diff
@@ -24,36 +24,21 @@ var _ TxWriter = (*Wallet)(nil)
...
+		WalletID: w.id,
+		Txid:     hash,
+		Label:    &label,
 	})
 	if err != nil {
-		return fmt.Errorf("failed to update wallet db: %w", err)
+		return fmt.Errorf("update tx label: %w", err)
 	}
```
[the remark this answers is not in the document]
yyforyongyu: Handled the wallet boundary translation here as well: `LabelTx` now maps store-layer `db.ErrTxNotFound` back to `wallet.ErrTxNotFound` before returning.

wallet/internal/db/kvdb/txstore.go
```diff
@@ -0,0 +1,155 @@
...
+		)
+	}
+
+	if params.Label == nil {
+		return nil, fmt.Errorf(
+			"kvdb.Store.UpdateTx: label patch required: %w",
+			db.ErrInvalidParam,
+		)
+	}
+
```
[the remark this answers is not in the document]
yyforyongyu: Removed the redundant `params.Label == nil` branch while updating the empty-label handling, so the validation flow is now straight-line.

wallet/tx_writer.go
yyforyongyu: Good catch. I updated `wallet.LabelTx` to translate `db.ErrTxNotFound` back to `wallet.ErrTxNotFound`, and restored the wallet-level assertion in `TestLabelTxNotFound` so the public API contract stays stable.

wallet/internal/db/kvdb/txstore.go
```diff
@@ -0,0 +1,155 @@
...
+func (s *Store) CreateTx(ctx context.Context, _ db.CreateTxParams) error {
+	return notImplemented(ctx, "CreateTx")
+}
+
+// UpdateTx patches mutable transaction metadata for the legacy kvdb backend.
+//
+// The current kvdb implementation only supports the label-only patch path
+// needed by the wallet tx-writer migration.
+func (s *Store) UpdateTx(_ context.Context, params db.UpdateTxParams) error {
+	label, err := validateUpdateTxLabelPatch(params)
```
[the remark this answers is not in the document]
yyforyongyu: Added the `WalletID` note to `kvdb.Store.UpdateTx` for consistency with the other single-wallet kvdb adapters.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Agreed. kvdb `UpdateTx` now accepts empty labels and clears the existing label by deleting the legacy label entry, so it matches the shared `UpdateTxParams` contract.

wallet/tx_writer.go
yyforyongyu: Fixed. `wallet.LabelTx` now translates store-layer `db.ErrTxNotFound` back to the wallet-level sentinel before returning, so existing `errors.Is(err, wallet.ErrTxNotFound)` callers keep working.

wallet/tx_writer_test.go
```diff
@@ -45,15 +41,17 @@ func TestLabelTxNotFound(t *testing.T) {
...
 
 	// Act: Attempt to label a tx that is not known to the wallet.
-	err := w.LabelTx(t.Context(), *TstTxHash, "some label")
+	err := w.LabelTx(t.Context(), *TstTxHash, label)
 
 	// Assert: Check that the correct error is returned.
-	require.ErrorIs(t, err, ErrTxNotFound)
-	mocks.txStore.AssertExpectations(t)
+	require.ErrorIs(t, err, db.ErrTxNotFound)
+	mocks.store.AssertExpectations(t)
```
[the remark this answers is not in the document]
yyforyongyu: Updated `TestLabelTxNotFound` to assert `wallet.ErrTxNotFound` again. The mock still returns `db.ErrTxNotFound`, and `LabelTx` now translates it at the wallet boundary.

wallet/internal/db/kvdb/txstore.go
```diff
@@ -0,0 +1,155 @@
...
+			db.ErrInvalidParam,
+		)
+	}
+
+	if *params.Label == "" {
+		return nil, fmt.Errorf(
+			"kvdb.Store.UpdateTx: empty label: %w",
+			errNotImplemented,
+		)
+	}
```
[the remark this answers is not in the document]
yyforyongyu: Agreed. I updated kvdb `UpdateTx` to accept `Label: &""` and clear by deleting the legacy label entry instead of writing an empty label, since legacy reads treat zero-length stored labels as invalid.

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

wallet/tx_writer.go
```diff
@@ -6,54 +6,44 @@ package wallet
...
-			return ErrTxNotFound
-		}
-
-		err = w.txStore.PutTxLabel(txmgrNs, hash, label)
-		if err != nil {
-			return fmt.Errorf("failed to put tx label: %w", err)
-		}
-
-		return nil
+	err = w.store.UpdateTx(ctx, db.UpdateTxParams{
```
yyforyongyu, line 36: This migration now forwards label writes straight to backend-specific `UpdateTx` implementations, which leaves label validation behavior backend-dependent. The kvdb path still routes through the legacy `PutTxLabel` checks, but the SQL paths just issue `UPDATE ... SET tx_label = ?` and surface generic constraint/query failures for overlong labels. Please keep `LabelTx` validation backend-agnostic by validating label length in the wallet/store layer before dispatching to the backend. (`gpt-5.4`)

yyforyongyu: Agreed. I added label-length validation in `wallet.LabelTx` using the legacy `wtxmgr.TxLabelLimit` before dispatch, so callers now get backend-independent `ErrLabelTooLong` behavior again. I also added a wallet test to verify long labels are rejected before reaching the store.

wallet/tx_writer.go
```diff
@@ -9,7 +9,7 @@ import (
 	"fmt"
 
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
-	"github.com/btcsuite/btcwallet/walletdb"
+	db "github.com/btcsuite/btcwallet/wallet/internal/db"
```
yyforyongyu: no need for the alias?

wallet/internal/db/kvdb/txstore.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+// RollbackToBlock is not yet implemented for kvdb.
+func (s *Store) RollbackToBlock(ctx context.Context, _ uint32) error {
+	return notImplemented(ctx, "RollbackToBlock")
+}
+
+// validateUpdateTxLabelPatch checks whether the requested legacy kvdb update is
+// a supported label-only patch and returns the requested label when it is.
+func validateUpdateTxLabelPatch(params db.UpdateTxParams) (*string, error) {
```
yyforyongyu: rename this to validateUpdateTx params and return an error when `param.State` is set - we should be clear in the godoc - we are re-implementing the current kvdb behavior, we are not changing any of its behaivor here?

wallet/tx_writer.go
```diff
@@ -32,12 +34,20 @@ func (w *Wallet) LabelTx(ctx context.Context,
 		return err
 	}
 
+	if label != "" && len(label) > wtxmgr.TxLabelLimit {
```
yyforyongyu: this check should be placed in the kvdb impl instead?

wallet/internal/db/kvdb/txstore.go
```diff
@@ -39,6 +47,20 @@ func (s *Store) UpdateTx(_ context.Context, params db.UpdateTxParams) error {
 			return db.ErrTxNotFound
 		}
 
+		if *label == "" {
```
yyforyongyu: i don't see why we need this?

wallet/tx_writer.go
yyforyongyu: Agreed. I dropped the alias in the follow-up so this just uses the plain `db` package import path directly.

wallet/tx_writer.go
yyforyongyu: Yes, agreed. I removed the wallet-layer length check in the follow-up. The goal here is to preserve legacy kvdb behavior while routing through `Store`, so label validation stays owned by the kvdb/wtxmgr path rather than being redefined in `wallet.LabelTx`.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Agreed. I removed this branch in the follow-up. Clearing on empty string changed legacy kvdb behavior, but this adapter should preserve the existing `PutTxLabel` validation semantics and only satisfy the transitional `Store` interface.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Agreed. I renamed the helper to `validateUpdateTxParams` and tightened the godoc around `UpdateTx` to make the intent explicit: this adapter is preserving the existing kvdb label-only behavior while satisfying the transitional `Store` interface.

yyforyongyu:
LGTM 🚀 (`minimax-m2.7`)

yyforyongyu:

yyforyongyu (commented):
LGTM 🏷️ **(`claude-opus-4-7`)**

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

wallet/tx_writer.go
```diff
@@ -24,36 +25,25 @@ var _ TxWriter = (*Wallet)(nil)
...
-			return ErrTxNotFound
-		}
-
-		err = w.txStore.PutTxLabel(txmgrNs, hash, label)
-		if err != nil {
-			return fmt.Errorf("failed to put tx label: %w", err)
-		}
-
-		return nil
+	err = w.store.UpdateTx(ctx, db.UpdateTxParams{
```
yyforyongyu, line 36: This path now makes empty-label behavior backend-dependent. The kvdb adapter still returns `wtxmgr.ErrEmptyLabel`, but the SQL backends accept `""` and persist it as “no label” because `UpdateTx` does not validate labels before dispatch. `Wallet.LabelTx("")` therefore succeeds on SQL and fails on kvdb, which is a contract regression for the migrated API. Please enforce one backend-agnostic empty-label policy before calling `store.UpdateTx`. (`gpt-5.4`)

yyforyongyu, line 36: Label length enforcement is now backend-dependent too. kvdb still returns `wtxmgr.ErrLabelTooLong`, while the SQL backends rely on column/check constraints and surface backend-specific SQL errors instead of the legacy sentinel. That means callers no longer get a stable error contract from `Wallet.LabelTx` after this migration. Please validate label length in the wallet/store layer so all backends return the same error. (`gpt-5.4`)

yyforyongyu: I agree the SQL and kvdb paths differ here today, but for this PR we want to preserve the existing kvdb behavior while routing `LabelTx` through the transitional `Store` interface. On kvdb, empty-label validation is still intentionally owned by the legacy `PutTxLabel` path, so `LabelTx("")` continuing to return `wtxmgr.ErrEmptyLabel` is expected in this migration. If we want one backend-agnostic empty-label policy, I think that should happen as a separate follow-up that updates all backends together rather than changing kvdb behavior in this refactor.

yyforyongyu: Same reasoning here: this PR is trying to preserve kvdb behavior while removing the direct `walletdb.Update/View` usage from `wallet.LabelTx`, not redefine the label contract across backends. kvdb still intentionally gets `ErrLabelTooLong` from the legacy `PutTxLabel` validation path, and I would prefer to keep that behavior stable in this migration. If we want `LabelTx` to guarantee one backend-agnostic length-validation error, I think that belongs in a dedicated follow-up that changes kvdb and SQL together with an explicit contract decision.

yyforyongyu (commented):
LGTM 🌼 (`gpt-5.4`)

yyforyongyu (commented):
LGTM 🌼 (`gpt-5.4`)

GustavoStingelin (approved):
LGTM!

PR #1203, opened 2026-03-31
wallet: migrate utxo manager to db.Store (superseded by #1238)

yyforyongyu:
**Superseded by #1238.**

The lease-side wallet routing (`LeaseOutput`, `ListLeasedOutputs`) was folded into PR #1238 (`impl-utxo-kvdb`) so all UTXO-store kvdb impls and wallet routes ship in one reviewable chain.

No content review needed here.


wallet/utxo_manager.go
```diff
@@ -205,189 +169,155 @@ func (w *Wallet) ListUnspent(_ context.Context,
...
 	return &Utxo{
-		OutPoint:      output.OutPoint,
-		Amount:        output.Amount,
-		PkScript:      output.PkScript,
-		Confirmations: confs,
+		OutPoint:      info.OutPoint,
+		Amount:        info.Amount,
+		PkScript:      info.PkScript,
+		Confirmations: confirmations,
 		Spendable:     spendable,
```
yyforyongyu, line 267: Once `ListUTXOs` returns leased rows, this preserves key spendability even when `locked` is true. The old direct `UnspentOutputs` path never returned locked coins, so callers can now see a leased coin with both `Locked=true` and `Spendable=true`; either filter locked rows or fold `!locked` into this flag. (`openai/gpt-5.5`)

wallet/utxo_manager.go
```diff
@@ -205,189 +169,155 @@ func (w *Wallet) ListUnspent(_ context.Context,
...
+		return false, "", 0, err
+	}
 
-		// In a single lookup, get all the required
-		// address-related details: spendability, account name,
-		// and address type. This avoids the N+1 query problem.
-		spendable, account, addrType := w.addrStore.AddressDetails(
-			addrmgrNs, addr,
-		)
+	return !addrInfo.IsWatchOnly, addrInfo.AccountName, walletAddrType, nil
```
yyforyongyu, line 305: This is not equivalent to the old `AddressDetails` policy. That path marked all imported addresses unspendable and also checked watch-only accounts, while `AddressInfo.IsWatchOnly` only captures wallet/private-key watch-only state; imported-key or watch-only-account UTXOs can now be reported as spendable. (`openai/gpt-5.5`)

wallet/utxo_manager.go
```diff
@@ -205,189 +169,155 @@ func (w *Wallet) ListUnspent(_ context.Context,
...
 
-		// Extract the address from the UTXO's public key script.
-		// For multi-address scripts, the first address is used.
-		addr := extractAddrFromPKScript(
-			output.PkScript, w.cfg.ChainParams,
-		)
-		if addr == nil {
-			return wtxmgr.ErrUtxoNotFound
-		}
+	walletAddrType, err := walletAddressType(addrInfo.AddrType)
```
yyforyongyu, line 300: For taproot script imports the store representation is `AddrType: db.TaprootPubKey` plus `HasScript: true` (see `addressInfoFromStoreAddress`). Converting only `AddrType` drops that discriminator, so `GetUtxo`/`ListUnspent` report these UTXOs as `TaprootPubKey` instead of the legacy `TaprootScript` type. (`openai/gpt-5.5`)

wallet/utxo_manager.go
```diff
@@ -471,54 +407,31 @@ func (w *Wallet) ReleaseOutput(ctx context.Context, id wtxmgr.LockID,
 	return w.store.ReleaseOutput(ctx, params)
```
yyforyongyu, line 407: `LeaseOutput` maps store errors back to the legacy public errors, but release now returns the store error directly. Backends that emit `db.ErrUtxoNotFound` or `db.ErrOutputUnlockNotAllowed` change the public API away from `wtxmgr.ErrUnknownOutput`/`ErrOutputUnlockNotAllowed`, so existing callers using `errors.Is` on wtxmgr errors will stop matching. (`openai/gpt-5.5`)

PR #1204, opened 2026-04-01
Prep address manager for store migration

yyforyongyu:
## Summary
- add wallet-owned `AddressInfo` and `OutputScriptInfo` types so wallet-facing lookup paths stop exposing `waddrmgr.ManagedAddress`
- separate signer private-key access from general address inspection, keeping key retrieval on signer-specific paths while preserving existing signing behavior
- centralize address metadata, signing policy, and imported pubkey derivation on `waddrmgr.AddressType` so wallet and manager code can use methods instead of repeating address-type switches

## Testing
- `GOWORK=off go test ./waddrmgr ./wallet ./wallet/internal/db/...`

### DIFF
--- a/waddrmgr/addr_type.go
+++ b/waddrmgr/addr_type.go
@@ -112,6 +112,48 @@ const (
 	DerivationSchemeBIP86
 )
 
+// AddressTypeForScope returns the default address type used by the given key
+// scope.
+func AddressTypeForScope(scope KeyScope) (AddressType, error) {
+	switch scope {
+	case KeyScopeBIP0044:
+		return PubKeyHash, nil
+
+	case KeyScopeBIP0049Plus:
+		return NestedWitnessPubKey, nil
+
+	case KeyScopeBIP0084:
+		return WitnessPubKey, nil
+
+	case KeyScopeBIP0086:
+		return TaprootPubKey, nil
+
+	default:
+		return 0, fmt.Errorf("%w: %v", ErrUnknownAddressType, scope)
+	}
+}
+
+// AddressTypeForPurpose returns the default address type used by the given BIP
+// purpose.
+func AddressTypeForPurpose(purpose uint32) (AddressType, error) {
+	switch purpose {
+	case KeyScopeBIP0044.Purpose:
+		return PubKeyHash, nil
+
+	case KeyScopeBIP0049Plus.Purpose:
+		return NestedWitnessPubKey, nil
+
+	case KeyScopeBIP0084.Purpose:
+		return WitnessPubKey, nil
+
+	case KeyScopeBIP0086.Purpose:
+		return TaprootPubKey, nil
+
+	default:
+		return 0, fmt.Errorf("%w: %d", ErrUnknownAddressType, purpose)
+	}
+}
+
 // SpendType returns the spend behavior implied by the address type. Unknown
 // address types return SpendTypeUnknown.
 func (a AddressType) SpendType() SpendType {
@@ -160,6 +202,51 @@ func (a AddressType) DerivationScheme() DerivationScheme {
 
 	default:
 		return DerivationSchemeNone
+	}
+}
+
+// KeyScope returns the preferred scoped-manager key scope for the address type.
+func (a AddressType) KeyScope() (KeyScope, error) {
+	switch a {
+	case PubKeyHash:
+		return KeyScopeBIP0044, nil
+
+	case NestedWitnessPubKey:
+		return KeyScopeBIP0049Plus, nil
+
+	case WitnessPubKey:
+		return KeyScopeBIP0084, nil
+
+	case TaprootPubKey:
+		return KeyScopeBIP0086, nil
+
+	default:
+		return KeyScope{}, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+	}
+}
+
+// ScopeAddrSchema returns the scoped-manager external/internal schema for the
+// address type.
+func (a AddressType) ScopeAddrSchema() (*ScopeAddrSchema, error) {
+	switch a {
+	// BIP49 keeps nested witness on the external branch but uses native
+	// witness change outputs on the internal branch.
+	case NestedWitnessPubKey:
+		return &ScopeAddrSchema{
+			ExternalAddrType: NestedWitnessPubKey,
+			InternalAddrType: WitnessPubKey,
+		}, nil
+
+	// The other key-derived scopes use the same address type for both
+	// external receives and internal change.
+	case PubKeyHash, WitnessPubKey, TaprootPubKey:
+		return &ScopeAddrSchema{
+			ExternalAddrType: a,
+			InternalAddrType: a,
+		}, nil
+
+	default:
+		return nil, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
 	}
 }
 
--- a/waddrmgr/addr_type.go
+++ b/waddrmgr/addr_type.go
@@ -112,6 +112,24 @@ const (
 	DerivationSchemeBIP86
 )
 
+// SigningMethod identifies the wallet signing flow currently supported for an
+// address type.
+type SigningMethod uint8
+
+const (
+	// SigningMethodLegacy indicates legacy ECDSA signing against a non-witness
+	// pkScript, for example PubKeyHash.
+	SigningMethodLegacy SigningMethod = iota
+
+	// SigningMethodWitnessV0 indicates SegWit v0 ECDSA signing, for example
+	// WitnessPubKey or NestedWitnessPubKey.
+	SigningMethodWitnessV0
+
+	// SigningMethodTaprootKeySpend indicates Taproot key-path Schnorr signing,
+	// for example TaprootPubKey.
+	SigningMethodTaprootKeySpend
+)
+
 // AddressTypeForScope returns the default address type used by the given key
 // scope.
 func AddressTypeForScope(scope KeyScope) (AddressType, error) {
@@ -286,5 +304,24 @@ func (a AddressType) WitnessVersion() (byte, error) {
 
 	default:
 		return 0, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+	}
+}
+
+// SigningMethod returns the wallet signing method for the address type.
+func (a AddressType) SigningMethod() (SigningMethod, error) {
+	switch a {
+	case PubKeyHash:
+		return SigningMethodLegacy, nil
+
+	case NestedWitnessPubKey, WitnessPubKey:
+		return SigningMethodWitnessV0, nil
+
+	case TaprootPubKey:
+		return SigningMethodTaprootKeySpend, nil
+
+	case Script, RawPubKey, WitnessScript, TaprootScript:
+		return 0, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+	default:
+		return 0, fmt.Errorf("%w: %v", ErrUnknownAddressType, a)
 	}
 }
--- a/waddrmgr/addr_type.go
+++ b/waddrmgr/addr_type.go
@@ -1,0 +1,203 @@
+package waddrmgr
+
+import (
+	"errors"
+	"fmt"
+)
+
+const (
+	// p2pkhPkScriptSize is the fixed scriptPubKey size for P2PKH outputs.
+	// 25 bytes = OP_DUP (1) + OP_HASH160 (1) + OP_DATA_20 (1) +
+	// <pubkey hash> (20) + OP_EQUALVERIFY (1) + OP_CHECKSIG (1).
+	p2pkhPkScriptSize = 25
+
+	// p2shPkScriptSize is the fixed scriptPubKey size for P2SH outputs.
+	// 23 bytes = OP_HASH160 (1) + OP_DATA_20 (1) + <script hash> (20) +
+	// OP_EQUAL (1).
+	p2shPkScriptSize = 23
+
+	// p2wpkhPkScriptSize is the fixed scriptPubKey size for native P2WPKH
+	// outputs.
+	// 22 bytes = OP_0 (1) + OP_DATA_20 (1) + <pubkey hash> (20).
+	p2wpkhPkScriptSize = 22
+
+	// p2wshPkScriptSize is the fixed scriptPubKey size for native P2WSH
+	// outputs.
+	// 34 bytes = OP_0 (1) + OP_DATA_32 (1) + <script hash> (32).
+	p2wshPkScriptSize = 34
+
+	// p2trPkScriptSize is the fixed scriptPubKey size for P2TR outputs.
+	// 34 bytes = OP_1 (1) + OP_DATA_32 (1) + <x-only output key> (32).
+	p2trPkScriptSize = 34
+
+	// nestedP2WPKHPkScriptSize is the outer P2SH scriptPubKey size used by
+	// nested P2WPKH outputs.
+	// 23 bytes = OP_HASH160 (1) + OP_DATA_20 (1) +
+	// HASH160(<inner witness program>) (20) + OP_EQUAL (1).
+	nestedP2WPKHPkScriptSize = p2shPkScriptSize
+)
+
+var (
+	// ErrUnknownAddressType is returned when no supported address type exists
+	// for a lookup key.
+	ErrUnknownAddressType = errors.New("unknown address type")
+
+	// ErrUnsupportedAddressType is returned when an address type is known but a
+	// specific API does not support it.
+	ErrUnsupportedAddressType = errors.New("unsupported address type")
+)
+
+// SpendType captures the spend behavior implied by an address type.
+type SpendType uint8
+
+const (
+	// SpendTypeUnknown is used when no standard spend template applies, such as
+	// RawPubKey.
+	SpendTypeUnknown SpendType = iota
+
+	// SpendTypeLegacyKey is a direct legacy key spend, for example PubKeyHash.
+	SpendTypeLegacyKey
+
+	// SpendTypeScriptHash is a generic redeem-script spend, for example Script.
+	SpendTypeScriptHash
+
+	// SpendTypeNestedWitnessKey is a nested P2WPKH-in-P2SH spend, for example
+	// NestedWitnessPubKey.
+	SpendTypeNestedWitnessKey
+
+	// SpendTypeWitnessKey is a native witness key spend, for example
+	// WitnessPubKey.
+	SpendTypeWitnessKey
+
+	// SpendTypeWitnessScript is a witness-script spend, for example
+	// WitnessScript.
+	SpendTypeWitnessScript
+
+	// SpendTypeTaprootKeyPath is a taproot key-path spend, for example
+	// TaprootPubKey.
+	SpendTypeTaprootKeyPath
+
+	// SpendTypeTaprootScriptPath is a taproot script-path spend, for example
+	// TaprootScript.
+	SpendTypeTaprootScriptPath
+
+	// SpendTypeAnyoneCanSpend is an anyone-can-spend policy output, for
+	// example a future anchor-only address type.
+	SpendTypeAnyoneCanSpend
+)
+
+// DerivationScheme captures the wallet derivation policy implied by an address
+// type.
+type DerivationScheme uint8
+
+const (
+	// DerivationSchemeNone indicates that no standard account derivation scheme
+	// applies, for example Script or TaprootScript.
+	DerivationSchemeNone DerivationScheme = iota
+
+	// DerivationSchemeBIP44 indicates BIP44 account derivation, for example the
+	// mainnet path m/44'/0'/0'/0/7.
+	DerivationSchemeBIP44
+
+	// DerivationSchemeBIP49 indicates BIP49 account derivation, for example the
+	// mainnet path m/49'/0'/0'/0/7.
+	DerivationSchemeBIP49
+
+	// DerivationSchemeBIP84 indicates BIP84 account derivation, for example the
+	// mainnet path m/84'/0'/0'/0/7.
+	DerivationSchemeBIP84
+
+	// DerivationSchemeBIP86 indicates BIP86 account derivation, for example the
+	// mainnet path m/86'/0'/0'/0/7.
+	DerivationSchemeBIP86
+)
+
+// SpendType returns the spend behavior implied by the address type. Unknown
+// address types return SpendTypeUnknown.
+func (a AddressType) SpendType() SpendType {
+	switch a {
+	case PubKeyHash:
+		return SpendTypeLegacyKey
+
+	case Script:
+		return SpendTypeScriptHash
+
+	case NestedWitnessPubKey:
+		return SpendTypeNestedWitnessKey
+
+	case WitnessPubKey:
+		return SpendTypeWitnessKey
+
+	case WitnessScript:
+		return SpendTypeWitnessScript
+
+	case TaprootPubKey:
+		return SpendTypeTaprootKeyPath
+
+	case TaprootScript:
+		return SpendTypeTaprootScriptPath
+
+	default:
+		return SpendTypeUnknown
+	}
+}
+
+// DerivationScheme returns the wallet derivation scheme implied by the address
+// type. Unknown address types return DerivationSchemeNone.
+func (a AddressType) DerivationScheme() DerivationScheme {
+	switch a {
+	case PubKeyHash:
+		return DerivationSchemeBIP44
+
+	case NestedWitnessPubKey:
+		return DerivationSchemeBIP49
+
+	case WitnessPubKey:
+		return DerivationSchemeBIP84
+
+	case TaprootPubKey:
+		return DerivationSchemeBIP86
+
+	default:
+		return DerivationSchemeNone
+	}
+}
+
+// ScriptPubKeySize returns the fixed outer scriptPubKey size for the address
+// type, if the outer form has one.
+func (a AddressType) ScriptPubKeySize() (int, error) {
+	switch a {
+	case PubKeyHash:
+		return p2pkhPkScriptSize, nil
+
+	case Script, NestedWitnessPubKey:
+		return p2shPkScriptSize, nil
+
+	case WitnessPubKey:
+		return p2wpkhPkScriptSize, nil
+
+	case WitnessScript:
+		return p2wshPkScriptSize, nil
+
+	case TaprootPubKey, TaprootScript:
+		return p2trPkScriptSize, nil
+
+	default:
+		return 0, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+	}
+}
+
+// WitnessVersion returns the witness version for the address type, if it uses
+// a witness program.
+func (a AddressType) WitnessVersion() (byte, error) {
+	switch a {
+	case NestedWitnessPubKey, WitnessPubKey, WitnessScript:
+		return witnessVersionV0, nil
+
+	case TaprootPubKey, TaprootScript:
+		return witnessVersionV1, nil
+
+	default:
+		return 0, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+	}
+}
--- a/waddrmgr/addr_type.go
+++ b/waddrmgr/addr_type.go
@@ -3,6 +3,12 @@ package waddrmgr
 import (
 	"errors"
 	"fmt"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
 )
 
 const (
@@ -197,6 +203,9 @@ func (a AddressType) SpendType() SpendType {
 	case TaprootScript:
 		return SpendTypeTaprootScriptPath
 
+	case RawPubKey:
+		return SpendTypeUnknown
+
 	default:
 		return SpendTypeUnknown
 	}
@@ -218,6 +227,9 @@ func (a AddressType) DerivationScheme() DerivationScheme {
 	case TaprootPubKey:
 		return DerivationSchemeBIP86
 
+	case Script, RawPubKey, WitnessScript, TaprootScript:
+		return DerivationSchemeNone
+
 	default:
 		return DerivationSchemeNone
 	}
@@ -238,8 +250,11 @@ func (a AddressType) KeyScope() (KeyScope, error) {
 	case TaprootPubKey:
 		return KeyScopeBIP0086, nil
 
-	default:
+	case Script, RawPubKey, WitnessScript, TaprootScript:
 		return KeyScope{}, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+
+	default:
+		return KeyScope{}, fmt.Errorf("%w: %v", ErrUnknownAddressType, a)
 	}
 }
 
@@ -263,8 +278,11 @@ func (a AddressType) ScopeAddrSchema() (*ScopeAddrSchema, error) {
 			InternalAddrType: a,
 		}, nil
 
-	default:
+	case Script, RawPubKey, WitnessScript, TaprootScript:
 		return nil, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+
+	default:
+		return nil, fmt.Errorf("%w: %v", ErrUnknownAddressType, a)
 	}
 }
 
@@ -287,8 +305,11 @@ func (a AddressType) ScriptPubKeySize() (int, error) {
 	case TaprootPubKey, TaprootScript:
 		return p2trPkScriptSize, nil
 
-	default:
+	case RawPubKey:
 		return 0, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+
+	default:
+		return 0, fmt.Errorf("%w: %v", ErrUnknownAddressType, a)
 	}
 }
 
@@ -302,8 +323,11 @@ func (a AddressType) WitnessVersion() (byte, error) {
 	case TaprootPubKey, TaprootScript:
 		return witnessVersionV1, nil
 
-	default:
+	case PubKeyHash, Script, RawPubKey:
 		return 0, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+
+	default:
+		return 0, fmt.Errorf("%w: %v", ErrUnknownAddressType, a)
 	}
 }
 
@@ -321,7 +345,110 @@ func (a AddressType) SigningMethod() (SigningMethod, error) {
 
 	case Script, RawPubKey, WitnessScript, TaprootScript:
 		return 0, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+
 	default:
 		return 0, fmt.Errorf("%w: %v", ErrUnknownAddressType, a)
 	}
+}
+
+// AddrFromPubKeyBytes reconstructs the standard address from the serialized
+// public key form stored by the wallet.
+func (a AddressType) AddrFromPubKeyBytes(pubKeyBytes []byte,
+	net *chaincfg.Params) (btcutil.Address, error) {
+
+	switch a {
+	case PubKeyHash:
+		address, err := btcutil.NewAddressPubKeyHash(
+			btcutil.Hash160(pubKeyBytes), net,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("new pubkey hash address: %w", err)
+		}
+
+		return address, nil
+
+	case NestedWitnessPubKey:
+		return nestedWitnessAddrFromPubKeyBytes(pubKeyBytes, net)
+
+	case WitnessPubKey:
+		address, err := btcutil.NewAddressWitnessPubKeyHash(
+			btcutil.Hash160(pubKeyBytes), net,
+		)
+		if err != nil {
+			return nil, fmt.Errorf("new witness pubkey hash address: %w", err)
+		}
+
+		return address, nil
+
+	case TaprootPubKey:
+		return taprootAddrFromPubKeyBytes(pubKeyBytes, net)
+
+	case Script, RawPubKey, WitnessScript, TaprootScript:
+		return nil, fmt.Errorf("%w: %v", ErrUnsupportedAddressType, a)
+
+	default:
+		return nil, fmt.Errorf("%w: %v", ErrUnknownAddressType, a)
+	}
+}
+
+// nestedWitnessProgramFromPubKeyBytes builds the nested witness redeem program
+// from serialized pubkey bytes.
+func nestedWitnessProgramFromPubKeyBytes(pubKeyBytes []byte,
+	net *chaincfg.Params) ([]byte, error) {
+
+	witnessAddr, err := btcutil.NewAddressWitnessPubKeyHash(
+		btcutil.Hash160(pubKeyBytes), net,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("new witness pubkey hash: %w", err)
+	}
+
+	script, err := txscript.PayToAddrScript(witnessAddr)
+	if err != nil {
+		return nil, fmt.Errorf("pay to witness address: %w", err)
+	}
+
+	return script, nil
+}
+
+// nestedWitnessAddrFromPubKeyBytes builds the outer P2SH address for a nested
+// witness pubkey-hash spend.
+func nestedWitnessAddrFromPubKeyBytes(pubKeyBytes []byte,
+	net *chaincfg.Params) (btcutil.Address, error) {
+
+	witnessProgram, err := nestedWitnessProgramFromPubKeyBytes(
+		pubKeyBytes, net,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("build nested witness program: %w", err)
+	}
+
+	address, err := btcutil.NewAddressScriptHash(witnessProgram, net)
+	if err != nil {
+		return nil, fmt.Errorf("new nested witness address: %w", err)
+	}
+
+	return address, nil
+}
+
+// taprootAddrFromPubKeyBytes builds a P2TR address from serialized internal
+// pubkey bytes.
+func taprootAddrFromPubKeyBytes(pubKeyBytes []byte,
+	net *chaincfg.Params) (btcutil.Address, error) {
+
+	internalPubKey, err := btcec.ParsePubKey(pubKeyBytes)
+	if err != nil {
+		return nil, fmt.Errorf("parse taproot internal pubkey: %w", err)
+	}
+
+	outputKey := schnorr.SerializePubKey(
+		txscript.ComputeTaprootKeyNoScript(internalPubKey),
+	)
+
+	address, err := btcutil.NewAddressTaproot(outputKey, net)
+	if err != nil {
+		return nil, fmt.Errorf("new taproot address: %w", err)
+	}
+
+	return address, nil
 }
--- a/waddrmgr/addr_type_test.go
+++ b/waddrmgr/addr_type_test.go
@@ -6,6 +6,34 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestAddressTypeLookups verifies the supported address-type lookup paths.
+func TestAddressTypeLookups(t *testing.T) {
+	t.Parallel()
+
+	scope, err := PubKeyHash.KeyScope()
+	require.NoError(t, err)
+	require.Equal(t, KeyScopeBIP0044, scope)
+
+	schema, err := NestedWitnessPubKey.ScopeAddrSchema()
+	require.NoError(t, err)
+	require.Equal(t, NestedWitnessPubKey, schema.ExternalAddrType)
+	require.Equal(t, WitnessPubKey, schema.InternalAddrType)
+
+	addrTypeByScope, err := AddressTypeForScope(KeyScopeBIP0049Plus)
+	require.NoError(t, err)
+	require.Equal(t, NestedWitnessPubKey, addrTypeByScope)
+
+	addrTypeByPurpose, err := AddressTypeForPurpose(KeyScopeBIP0086.Purpose)
+	require.NoError(t, err)
+	require.Equal(t, TaprootPubKey, addrTypeByPurpose)
+
+	_, err = AddressTypeForScope(KeyScope{Purpose: 1017, Coin: 0})
+	require.ErrorIs(t, err, ErrUnknownAddressType)
+
+	_, err = AddressTypeForPurpose(1017)
+	require.ErrorIs(t, err, ErrUnknownAddressType)
+}
+
 // TestPubKeyAddressTypeMetadata verifies the common metadata for pubkey address
 // types.
 func TestPubKeyAddressTypeMetadata(t *testing.T) {
--- a/waddrmgr/addr_type_test.go
+++ b/waddrmgr/addr_type_test.go
@@ -1,0 +1,164 @@
+package waddrmgr
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestPubKeyAddressTypeMetadata verifies the common metadata for pubkey address
+// types.
+func TestPubKeyAddressTypeMetadata(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name              string
+		addrType          AddressType
+		spendType         SpendType
+		derivationScheme  DerivationScheme
+		pkScriptSize      int
+		hasWitnessVersion bool
+		witnessVersion    byte
+	}{
+		{
+			name:             "pubkey hash",
+			addrType:         PubKeyHash,
+			spendType:        SpendTypeLegacyKey,
+			derivationScheme: DerivationSchemeBIP44,
+			pkScriptSize:     p2pkhPkScriptSize,
+		},
+		{
+			name:              "nested witness pubkey",
+			addrType:          NestedWitnessPubKey,
+			spendType:         SpendTypeNestedWitnessKey,
+			derivationScheme:  DerivationSchemeBIP49,
+			pkScriptSize:      nestedP2WPKHPkScriptSize,
+			hasWitnessVersion: true,
+			witnessVersion:    witnessVersionV0,
+		},
+		{
+			name:              "witness pubkey",
+			addrType:          WitnessPubKey,
+			spendType:         SpendTypeWitnessKey,
+			derivationScheme:  DerivationSchemeBIP84,
+			pkScriptSize:      p2wpkhPkScriptSize,
+			hasWitnessVersion: true,
+			witnessVersion:    witnessVersionV0,
+		},
+		{
+			name:              "taproot pubkey",
+			addrType:          TaprootPubKey,
+			spendType:         SpendTypeTaprootKeyPath,
+			derivationScheme:  DerivationSchemeBIP86,
+			pkScriptSize:      p2trPkScriptSize,
+			hasWitnessVersion: true,
+			witnessVersion:    witnessVersionV1,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			require.Equal(t, testCase.spendType,
+				testCase.addrType.SpendType())
+			require.Equal(t, testCase.derivationScheme,
+				testCase.addrType.DerivationScheme())
+
+			pkScriptSize, err := testCase.addrType.ScriptPubKeySize()
+			require.NoError(t, err)
+			require.Equal(t, testCase.pkScriptSize, pkScriptSize)
+
+			version, err := testCase.addrType.WitnessVersion()
+			if testCase.hasWitnessVersion {
+				require.NoError(t, err)
+				require.Equal(t, testCase.witnessVersion, version)
+			} else {
+				require.ErrorIs(t, err, ErrUnsupportedAddressType)
+			}
+		})
+	}
+}
+
+// TestScriptAddressTypeMetadata verifies the common metadata for script-driven
+// address types.
+func TestScriptAddressTypeMetadata(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name              string
+		addrType          AddressType
+		spendType         SpendType
+		pkScriptSize      int
+		hasWitnessVersion bool
+		witnessVersion    byte
+	}{
+		{
+			name:         "script hash",
+			addrType:     Script,
+			spendType:    SpendTypeScriptHash,
+			pkScriptSize: p2shPkScriptSize,
+		},
+		{
+			name:              "witness script hash",
+			addrType:          WitnessScript,
+			spendType:         SpendTypeWitnessScript,
+			pkScriptSize:      p2wshPkScriptSize,
+			hasWitnessVersion: true,
+			witnessVersion:    witnessVersionV0,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			require.Equal(t, testCase.spendType,
+				testCase.addrType.SpendType())
+
+			pkScriptSize, err := testCase.addrType.ScriptPubKeySize()
+			require.NoError(t, err)
+			require.Equal(t, testCase.pkScriptSize, pkScriptSize)
+
+			version, err := testCase.addrType.WitnessVersion()
+			if testCase.hasWitnessVersion {
+				require.NoError(t, err)
+				require.Equal(t, testCase.witnessVersion, version)
+			} else {
+				require.ErrorIs(t, err, ErrUnsupportedAddressType)
+			}
+		})
+	}
+}
+
+// TestTaprootScriptAddressType verifies the taproot script type still has
+// meaningful common metadata.
+func TestTaprootScriptAddressType(t *testing.T) {
+	t.Parallel()
+
+	require.Equal(t, SpendTypeTaprootScriptPath, TaprootScript.SpendType())
+	require.Equal(t, DerivationSchemeNone, TaprootScript.DerivationScheme())
+
+	size, err := TaprootScript.ScriptPubKeySize()
+	require.NoError(t, err)
+	require.Equal(t, p2trPkScriptSize, size)
+
+	version, err := TaprootScript.WitnessVersion()
+	require.NoError(t, err)
+	require.Equal(t, witnessVersionV1, version)
+}
+
+// TestRawPubKeyAddressType verifies the raw-pubkey address type retains only
+// the common metadata that naturally applies to it.
+func TestRawPubKeyAddressType(t *testing.T) {
+	t.Parallel()
+
+	require.Equal(t, SpendTypeUnknown, RawPubKey.SpendType())
+	require.Equal(t, DerivationSchemeNone, RawPubKey.DerivationScheme())
+
+	_, err := RawPubKey.ScriptPubKeySize()
+	require.ErrorIs(t, err, ErrUnsupportedAddressType)
+
+	_, err = RawPubKey.WitnessVersion()
+	require.ErrorIs(t, err, ErrUnsupportedAddressType)
+}
--- a/waddrmgr/addr_type_test.go
+++ b/waddrmgr/addr_type_test.go
@@ -1,9 +1,17 @@
 package waddrmgr
 
 import (
-	"github.com/stretchr/testify/require"
 	"testing"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
+	"github.com/stretchr/testify/require"
 )
+
+var testChainParams = chaincfg.RegressionNetParams
 
 // TestAddressTypeLookups verifies the supported address-type lookup paths.
 func TestAddressTypeLookups(t *testing.T) {
@@ -26,11 +34,17 @@ func TestAddressTypeLookups(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, TaprootPubKey, addrTypeByPurpose)
 
+	_, err = AddressType(255).AddrFromPubKeyBytes(nil, &testChainParams)
+	require.ErrorIs(t, err, ErrUnknownAddressType)
+
 	_, err = AddressTypeForScope(KeyScope{Purpose: 1017, Coin: 0})
 	require.ErrorIs(t, err, ErrUnknownAddressType)
 
 	_, err = AddressTypeForPurpose(1017)
 	require.ErrorIs(t, err, ErrUnknownAddressType)
+
+	_, err = Script.AddrFromPubKeyBytes(nil, &testChainParams)
+	require.ErrorIs(t, err, ErrUnsupportedAddressType)
 }
 
 // TestPubKeyAddressTypeMetadata verifies the common metadata for pubkey address
@@ -228,6 +242,78 @@ func TestSupportedSigningAddressType(t *testing.T) {
 			signingMethod, err := testCase.addrType.SigningMethod()
 			require.NoError(t, err)
 			require.Equal(t, testCase.signingMethod, signingMethod)
+		})
+	}
+}
+
+// TestUnsupportedSigningAddressType verifies unsupported address types report
+// no wallet signing path.
+func TestUnsupportedSigningAddressType(t *testing.T) {
+	t.Parallel()
+
+	_, err := WitnessScript.SigningMethod()
+	require.ErrorIs(t, err, ErrUnsupportedAddressType)
+
+	_, err = AddressType(255).SigningMethod()
+	require.ErrorIs(t, err, ErrUnknownAddressType)
+}
+
+// TestAddrFromPubKeyBytes verifies address reconstruction from the wallet's
+// stored pubkey bytes for supported pubkey address types.
+func TestAddrFromPubKeyBytes(t *testing.T) {
+	t.Parallel()
+
+	privKey, err := btcec.NewPrivateKey()
+	require.NoError(t, err)
+
+	serializedPubKey := privKey.PubKey().SerializeCompressed()
+
+	testCases := []struct {
+		name     string
+		addrType AddressType
+		want     []byte
+	}{
+		{
+			name:     "pubkey hash",
+			addrType: PubKeyHash,
+			want:     btcutil.Hash160(serializedPubKey),
+		},
+		{
+			name:     "nested witness pubkey",
+			addrType: NestedWitnessPubKey,
+			want: func() []byte {
+				witnessProgram, err := nestedWitnessProgramFromPubKeyBytes(
+					serializedPubKey, &testChainParams,
+				)
+				require.NoError(t, err)
+
+				return btcutil.Hash160(witnessProgram)
+			}(),
+		},
+		{
+			name:     "witness pubkey",
+			addrType: WitnessPubKey,
+			want:     btcutil.Hash160(serializedPubKey),
+		},
+		{
+			name:     "taproot pubkey",
+			addrType: TaprootPubKey,
+			want: schnorr.SerializePubKey(
+				txscript.ComputeTaprootKeyNoScript(privKey.PubKey()),
+			),
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			address, err := testCase.addrType.AddrFromPubKeyBytes(
+				serializedPubKey, &testChainParams,
+			)
+			require.NoError(t, err)
+
+			require.Equal(t, testCase.want, address.ScriptAddress())
 		})
 	}
 }
--- a/waddrmgr/addr_type_test.go
+++ b/waddrmgr/addr_type_test.go
@@ -1,9 +1,8 @@
 package waddrmgr
 
 import (
-	"testing"
-
 	"github.com/stretchr/testify/require"
+	"testing"
 )
 
 // TestAddressTypeLookups verifies the supported address-type lookup paths.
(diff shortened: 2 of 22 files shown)


waddrmgr/address_desc.go
```diff
@@ -0,0 +1,621 @@
...
+)
+
+// singleKeySpendInfo groups the single-key signing details that only apply to
+// address descriptors with a direct wallet signing path.
+type singleKeySpendInfo struct {
+	// signingMethod selects the wallet signing implementation to use.
+	signingMethod SigningMethod
+
+	// defaultSigHash is the default sighash used by the signing method.
+	defaultSigHash txscript.SigHashType
```
yyforyongyu: why it's called defaultSigHash instead of `sigHash`?

waddrmgr/address_desc.go
```diff
@@ -0,0 +1,621 @@
...
+
+	// requiresRedeemScript reports whether the signing path needs a redeem
+	// script in addition to the previous output script.
```
yyforyongyu: give some examples here, one that needs it, one that doesn't?

waddrmgr/address_desc.go
```diff
@@ -0,0 +1,621 @@
...
+	requiresRedeemScript bool
+
+	// redeemScriptFromPubKey builds the redeem script for single-key spend
+	// paths that require one.
+	redeemScriptFromPubKey redeemScriptFromPubKeyFunc
```
yyforyongyu: can we implicitly check `redeemScriptFromPubKey` against nil to decide whether the we require redeem script or not?

waddrmgr/address_desc.go
```diff
@@ -0,0 +1,621 @@
...
+	// derivationScheme is the wallet derivation policy implied by the type.
+	derivationScheme DerivationScheme
+
+	// keyScope is set when the address type is naturally scoped by an HD key
+	// scope.
+	keyScope *KeyScope
+
+	// scopeAddrSchema is set when the address type has a fixed
+	// internal/external scoped-manager schema.
+	scopeAddrSchema *ScopeAddrSchema
```
yyforyongyu: ok how is this different from `keyScope` and `derivationScheme`?

waddrmgr/address_desc.go
```diff
@@ -0,0 +1,621 @@
...
+
+	// purpose is set when the address type has a canonical BIP purpose.
+	purpose *uint32
```
yyforyongyu: shouldn't this be part of the key scope or derivation?

waddrmgr/address_desc.go
```diff
@@ -0,0 +1,621 @@
...
+
+	// fixedScriptSize is set when the outer scriptPubKey size is fixed.
+	fixedScriptSize *int
```
yyforyongyu: name it `scriptSize`?

waddrmgr/address_desc.go
```diff
@@ -0,0 +1,621 @@
...
+	// type is script-derived.
+	addressFromScript addressFromScriptFunc
+
+	// addressIDFromScript constructs the wallet storage identity from a script
+	// when the type is script-derived.
+	addressIDFromScript addressIDFromScriptFunc
+
+	// addressFromSerializedPubKeyFunc reconstructs the standard address from
+	// the serialized public key stored by the wallet.
+	addressFromSerializedPubKeyFunc addressFromSerializedPubKeyFunc
```
yyforyongyu: feels like all the `addressFromXXX` can be grouped into a struct?

waddrmgr/address_desc.go
```diff
@@ -0,0 +1,621 @@
...
+	if i.purpose == nil {
+		return 0, false
+	}
+
+	return *i.purpose, true
+}
+
+// FixedScriptPubKeySize returns the fixed scriptPubKey size for the address
+// type, if the outer form has one.
+func (i *AddressDesc) FixedScriptPubKeySize() (int, bool) {
```
yyforyongyu: again just name it `ScriptSize`?

waddrmgr/address_desc.go
yyforyongyu: Renamed this to `sigHashDefault` in `2b787af11`. I kept the `default` part because this is the descriptor's default sighash policy, not the only sighash it can ever use.

waddrmgr/address_desc.go
yyforyongyu: Added the explicit examples to the `requiresRedeemScript` comment in `2b787af11`: nested witness key spends need one, while native witness pubkey spends and taproot key-path spends do not.

waddrmgr/address_desc.go
yyforyongyu: I clarified this in the field comment in `2b787af11`: `scopeAddrSchema` is more specific than `keyScope` / `derivationScheme` because it captures the external/internal branch policy within a scope, including mixed cases like BIP49 nested external addresses with native-witness internal change addresses.

waddrmgr/address_desc.go
yyforyongyu: I left it separate for now for that reason: `keyScope` and `derivationScheme` tell us the HD policy, but `scopeAddrSchema` still carries the concrete external/internal branch mapping. The updated comment calls out that distinction explicitly.

waddrmgr/address_desc.go
yyforyongyu: Renamed it more specifically to `scriptPubKeySize` in `2b787af11`, and the accessor is now `ScriptPubKeySize()`. I avoided plain `scriptSize` because this field is specifically about the outer `scriptPubKey`, not redeem/witness/scriptSig sizes.

waddrmgr/address_desc.go
yyforyongyu: Done in `2b787af11`. The construction hooks now live under the nested `addressConstruction` struct (`fromPubKey`, `idFromPubKey`, `fromScript`, `idFromScript`, `fromSerializedPubKey`).

waddrmgr/address_desc.go
yyforyongyu: Addressed by the same rename: it is now `scriptPubKeySize` / `ScriptPubKeySize()` so the field stays tied to the outer `scriptPubKey` specifically.

wallet/address_manager.go
```diff
@@ -67,6 +67,66 @@ type AddressProperty struct {
...
+	// Addr is the bitcoin address itself.
+	Addr btcutil.Address
+
+	// AddrType identifies the wallet-managed address type for this concrete
+	// address.
+	AddrType waddrmgr.AddressType
+
+	// IsImported reports whether the address was imported instead of
+	// derived from a wallet scope.
+	IsImported bool
```
yyforyongyu: rename it to `Imported`?

wallet/address_manager.go
```diff
@@ -67,6 +67,66 @@ type AddressProperty struct {
...
+
+	// IsInternal reports whether the address belongs to the wallet's internal
+	// branch.
+	IsInternal bool
```
yyforyongyu: rename it to `Internal`?

wallet/address_manager.go
```diff
@@ -67,6 +67,66 @@ type AddressProperty struct {
...
+
+	// IsCompressed reports whether the underlying pubkey address uses
+	// compressed keys.
+	IsCompressed bool
```
yyforyongyu: rename it to `Compressed`?

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,388 @@
...
+	// pubKeyHashScopeAddrSchema is the scoped-manager branch schema for P2PKH
+	// account derivation.
+	pubKeyHashScopeAddrSchema = ScopeAddrSchema{
+		ExternalAddrType: PubKeyHash,
+		InternalAddrType: PubKeyHash,
+	}
+
+	// nestedWitnessPubKeyScopeAddrSchema is the scoped-manager branch schema
+	// for the BIP49 nested witness pubkey-hash account layout.
+	nestedWitnessPubKeyScopeAddrSchema = ScopeAddrSchema{
```
yyforyongyu: I don't think we need to define all these schema here? most of the time the internal/external are the same, with only the exception for `NestedWitnessPubKey` - i think we can check it in the helper method instead?

wallet/address_manager.go
yyforyongyu: Done. I renamed the field to `Imported` and updated the wallet metadata uses and tests.

wallet/address_manager.go
yyforyongyu: Done. I renamed it to `Internal` and updated the call sites and tests.

wallet/address_manager.go
yyforyongyu: Done. I renamed it to `Compressed` and updated the call sites and tests.

waddrmgr/addr_type.go
yyforyongyu: Done. `ScopeAddrSchema()` now only special-cases `NestedWitnessPubKey`; the same-type schemas are constructed directly in the helper.

yyforyongyu:

yyforyongyu (commented):
LGTM 🎲

All review comments resolved. The PR successfully preps the address manager for store migration:

- New `AddressType` methods centralize address-type metadata (OutputType, SpendType, DerivationScheme, SigningMethod)
- Address cache converted from unbounded map to LRU cache with 10k limit
- `AddressInfo` and `OutputScriptInfo` types decouple wallet from `waddrmgr.ManagedAddress`
- Signing validation uses `SigningMethod()` instead of address-type switches
- All field renames (`Imported`, `Internal`, `Compressed`) applied per review feedback

Good work addressing all the inline feedback! (`glm5`)

wallet/psbt_manager.go
```diff
@@ -463,8 +460,8 @@ func (w *Wallet) decorateInput(ctx context.Context, pInput *psbt.PInput,
 	// For SegWit v0 inputs, we'll use the SegWit v0 helper.
 	default:
 		// We'll need to build the redeem script for the input.
-		_, redeemScript, err := buildScriptsForManagedAddress(
-			pubKeyAddr, utxo.PkScript, w.cfg.ChainParams,
+		_, redeemScript, err := buildScriptsForAddressInfo(
```
yyforyongyu: `buildScriptsForAddressInfo` returns `(witnessProgram, redeemScript)`, but the witness program is discarded here and `addInputInfoSegWitV0` receives the single-push scriptSig form. That causes `PInput.RedeemScript` to be set to `OP_DATA_22 0x0014...` instead of the actual nested redeem script `0x0014...` (the witness program), which can break BIP174 nested P2WPKH signing/interoperability. Can we pass `witnessProgram` here (and in the deprecated path) and keep the push form only for final script assembly? (`gpt-5.3-codex`)

wallet/signer.go
```diff
@@ -724,17 +716,101 @@ func (w *Wallet) ComputeUnlockingScript(ctx context.Context,
...
+		return nil, fmt.Errorf("%w: addr=%v", ErrDerivationPathNotFound,
+			pubKeyAddr.Address())
+	}
+
+	accountManager, err := w.addrStore.FetchScopedKeyManager(keyScope)
+	if err != nil {
+		return nil, fmt.Errorf("fetch scoped key manager: %w", err)
+	}
+
+	privKey, err := accountManager.DeriveFromKeyPathCache(derivationPath)
```
yyforyongyu, line 791: This now relies exclusively on `DeriveFromKeyPathCache`, but that helper explicitly fails with `ErrAccountNotCached` until the account cache has already been warmed. `loadManagedPubKeyAddr` only loads the address row, so a cold wallet can now fail to sign perfectly valid derived addresses right after startup. Please fall back to the normal derivation path when the cache miss occurs, or warm the account cache before calling this fast path. (`gpt-5.4`)

wallet/address_manager.go
```diff
@@ -67,15 +71,69 @@ type AddressProperty struct {
...
 
-	// WitnessProgram is the witness program of the UTXO.
+	// WitnessProgram is the script passed as the witness subscript for witness
+	// signing. For native P2WPKH and P2TR spends, this is the output pkScript
+	// itself. For nested P2WPKH-in-P2SH spends, this is the inner witness
+	// program, for example `OP_0 <20-byte-key-hash>`.
 	WitnessProgram []byte
 
-	// RedeemScript is the redeem script of the UTXO.
+	// RedeemScript is the P2SH redeem script that must be pushed into sigScript
```
yyforyongyu: The new `OutputScriptInfo.RedeemScript` contract is backwards for nested P2WKH. PSBT `RedeemScript` and the later segwit-v0 signing subscript must be the inner witness program (`OP_0 <20-byte-hash>`), but this field is documented and populated as the outer single-push scriptSig wrapper. `DecorateInputs` then writes that wrapper into `PInput.RedeemScript`, and `SignPsbt` signs against it, which produces malformed nested-witness PSBT data. Please store the actual redeem script here and keep the push-wrapper, if needed, as a separate scriptSig helper. (`gpt-5.4`)

yyforyongyu (commented):
LGTM 🗂️

AddressType method tables verified correct for all 8 types. Cache bookkeeping fix is sound (monotonic index guards, DB-authoritative next index, account cache warming). Script address derivation and signing helpers correct. One security-relevant observation noted inline.

**(`claude-opus-4-6`)**

waddrmgr/scoped_manager.go
```diff
@@ -61,6 +61,12 @@ const (
 	// the wallet. With the default sisize, we'll allocate up to 320 KB to
 	// caching private keys (ignoring pointer overhead, etc).
 	defaultPrivKeyCacheSize = 10_000
+
+	// defaultAddrCacheSize is the default number of managed addresses that
+	// each scoped manager keeps hot in memory. This bounds long-lived
+	// address metadata growth while still covering recent lookups and newly
+	// derived addresses.
+	defaultAddrCacheSize = 10_000
```
yyforyongyu: `Manager.lock()` and `ConvertToWatchingOnly()` iterate `manager.addresses()`, which now only returns entries resident in this LRU (capacity 10,000). Addresses evicted from the cache retain their `privKeyCT` bytes un-zeroed in memory until GC collects the struct. Before this change, the unbounded map ensured all addresses were zeroed.

For most wallets this is a non-issue (<10k addresses per scoped manager), but a wallet that derives >10k addresses in a single scope and then locks will leave some cleartext private keys un-zeroed. Consider adding an eviction callback that calls `addr.lock()` on eviction, or document the trade-off.

**(`claude-opus-4-6`)**

wallet/signer.go
yyforyongyu: Fixed. resolvePrivKey now tries DeriveFromKeyPathCache first, but falls back to the normal DB-backed DeriveFromKeyPath path on ErrAccountNotCached, so cold startup signing still works. Added coverage in TestResolvePrivKeyFallsBackAfterCacheMiss.

wallet/psbt_manager.go
yyforyongyu: Fixed. For nested P2WPKH, the wallet now keeps the inner witness program as the redeem script and keeps the pushed wrapper separately as SigScript. DecorateInputs and the deprecated PSBT path now pass the witness program into PInput.RedeemScript, and there is regression coverage in TestDecorateInputNestedWitnessUsesRedeemScript.

wallet/address_manager.go
yyforyongyu: Fixed. OutputScriptInfo.RedeemScript now stores the actual nested redeem script (the inner witness program), while the pushed scriptSig wrapper moved to OutputScriptInfo.SigScript for final input assembly. That keeps PSBT decoration and segwit-v0 signing on the witness program while preserving the final scriptSig helper.

waddrmgr/scoped_manager.go
yyforyongyu: Fixed. The scoped-manager address cache now uses an eviction callback that locks cached addresses on eviction, and Manager.lock() reuses the same helper. That restores eager zeroing for evicted cached addresses instead of waiting on GC. Added coverage in TestScopedManagerAddressEvictionLocksSecrets.

yyforyongyu: we should also mention the memory size? similar to `defaultPrivKeyCacheSize` - and i think we can make this 100k?

waddrmgr/scoped_manager.go
```diff
@@ -718,6 +724,44 @@ type cachedKey struct {
...
 }
 
+// cachedAddress is an address-cache entry in the scoped manager LRU.
+type cachedAddress struct {
+	address ManagedAddress
+}
+
+// lockManagedAddress zeroes any sensitive material cached on one managed
+// address instance.
+func lockManagedAddress(address ManagedAddress) {
```
yyforyongyu: explain why it's safe to call the `lock` here? we usually would have a `defer unlock()`?

wallet/address_manager.go
```diff
@@ -965,44 +1022,71 @@ func (w *Wallet) GetDerivationInfo(ctx context.Context,
...
-	return derivationForManagedAddress(pubKeyAddr)
+	return derivationInfo, nil
 }
 
 // derivationForManagedAddress constructs a PSBT Bip32Derivation struct from a
 // managed public key address.
 func derivationForManagedAddress(pubKeyAddr waddrmgr.ManagedPubKeyAddress) (
 	*psbt.Bip32Derivation, error) {
 
-	// Imported addresses don't have derivation paths.
```
yyforyongyu, line 863: unnecessary delete?

wallet/address_manager.go
```diff
@@ -965,44 +1022,71 @@ func (w *Wallet) GetDerivationInfo(ctx context.Context,
...
 	if pubKeyAddr.Imported() {
 		return nil, fmt.Errorf("%w: addr=%v is imported",
 			ErrDerivationPathNotFound, pubKeyAddr.Address())
 	}
 
-	// Get the derivation info.
```
yyforyongyu, line 869: unnecessary delete?

wallet/address_manager.go
```diff
@@ -965,44 +1022,71 @@ func (w *Wallet) GetDerivationInfo(ctx context.Context,
...
 	keyScope, derivPath, ok := pubKeyAddr.DerivationInfo()
 	if !ok {
 		return nil, fmt.Errorf("%w: derivation info not found for %v",
 			ErrDerivationPathNotFound, pubKeyAddr.Address())
 	}
 
-	// Get the public key.
```
yyforyongyu, line 876: unnecessary delete?

wallet/psbt_manager.go
```diff
@@ -429,24 +430,33 @@ func (w *Wallet) decorateInput(ctx context.Context, pInput *psbt.PInput,
...
+		return err
+	})
 	if err != nil {
 		return fmt.Errorf("unable to get address info for %s: %w",
 			addr.String(), err)
 	}
 
-	// We'll ensure that the managed address is a public key address, as
-	// we can only decorate inputs for which we have the private key.
+	// We'll ensure that the managed address is a public key address, as we can
```
yyforyongyu: unnecessary change?

wallet/psbt_manager.go
```diff
@@ -429,24 +430,33 @@ func (w *Wallet) decorateInput(ctx context.Context, pInput *psbt.PInput,
...
+	// only decorate inputs for which we have the private key.
 	pubKeyAddr, ok := managedAddr.(waddrmgr.ManagedPubKeyAddress)
 	if !ok {
 		return fmt.Errorf("%w: addr %s", ErrNotPubKeyAddress,
 			managedAddr.Address())
 	}
 
-	// With the managed address, we can now get the derivation information
-	// for the address.
+	// With the managed address, we can now get the derivation information for the
```
yyforyongyu: unnecessary change?

waddrmgr/scoped_manager.go
yyforyongyu: Addressed. I bumped `defaultAddrCacheSize` to `100_000` and expanded the comment to call out the approximate memory footprint and count-based trade-off.

wallet/address_manager.go
yyforyongyu: Re-added the missing caller-facing comments in `derivationForAddressInfo`. The helper still works on wallet-owned metadata, but it now explicitly calls out the public-key-only and derivation-metadata checks again.

wallet/address_manager.go
yyforyongyu: Re-added the missing caller-facing comments in `derivationForAddressInfo`. The helper still works on wallet-owned metadata, but it now explicitly calls out the public-key-only and derivation-metadata checks again.

wallet/psbt_manager.go
yyforyongyu: Dropped that direct `addrStore.Address` detour while cleaning the branch history. The current `decorateInput` path stays on `GetAddressInfo` plus `derivationForAddressInfo`, so this change is no longer in the stack.

waddrmgr/manager.go
```diff
@@ -440,14 +440,15 @@ func (m *Manager) lock() {
...
+		manager.addrs.Range(func(_ addrKey, cachedAddr *cachedAddr) bool {
+			if addr, ok := cachedAddr.addr.(interface{ lock() }); ok {
 				addr.lock()
 			}
-		}
+
+			return true
+		})
+
+		manager.mtx.Unlock()
```
yyforyongyu, line 457: we have unlock here, but i don't see the `Lock`?

waddrmgr/manager.go
```diff
@@ -440,14 +440,15 @@ func (m *Manager) lock() {
 
 		// Remove clear text private keys and scripts from all address
 		// entries.
-		for _, ma := range manager.addresses() {
-			switch addr := ma.(type) {
-			case *managedAddress:
-				addr.lock()
-			case *scriptAddress:
+		manager.addrs.Range(func(_ addrKey, cachedAddr *cachedAddr) bool {
+			if addr, ok := cachedAddr.addr.(interface{ lock() }); ok {
```
yyforyongyu: add inline comments above if to explain this interface?

waddrmgr/scoped_manager.go
```diff
@@ -718,6 +725,30 @@ type cachedKey struct {
...
+	addr ManagedAddress
+}
+
+// newAddrCache constructs one bounded address cache with eviction-time
+// zeroing for cached secret material.
+func newAddrCache(capacity uint64) *lru.Cache[addrKey, *cachedAddr] {
+	return lru.NewCache[addrKey, *cachedAddr](
+		capacity,
+		lru.WithDeleteCallback(func(_ addrKey, cachedAddr *cachedAddr) {
+			if cachedAddr == nil {
```
yyforyongyu, line 759: add inline comments above if to explain the nil

wallet/address_manager.go
```diff
@@ -876,7 +926,14 @@ func (w *Wallet) ScriptForOutput(ctx context.Context, output wire.TxOut) (
 
 	// We'll then use the address to look up the managed address from the
 	// database.
-	managedAddr, err := w.GetAddressInfo(ctx, addr)
```
yyforyongyu: why do we need this change?

waddrmgr/addr_type_test.go
```diff
@@ -0,0 +1,206 @@
...
+				testCase.addrType.DerivationScheme())
+
+			pkScriptSize, ok := testCase.addrType.ScriptPubKeySize()
+			require.True(t, ok)
+			require.Equal(t, testCase.pkScriptSize, pkScriptSize)
+
+			version, ok := testCase.addrType.WitnessVersion()
+			if !testCase.hasWitnessVersion {
+				require.False(t, ok)
+			} else {
```
yyforyongyu: we can just write `require.Equal(t, testCase.hasWitnessVersion, ok)` and remove the `if !testCase.hasWitnessVersion {`?

waddrmgr/addr_type_test.go
```diff
@@ -0,0 +1,206 @@
...
+				testCase.addrType.SpendType())
+
+			pkScriptSize, ok := testCase.addrType.ScriptPubKeySize()
+			require.True(t, ok)
+			require.Equal(t, testCase.pkScriptSize, pkScriptSize)
+
+			version, ok := testCase.addrType.WitnessVersion()
+			if !testCase.hasWitnessVersion {
+				require.False(t, ok)
+			} else {
```
yyforyongyu: same here, can remove `if !testCase.hasWitnessVersion {`

waddrmgr/addr_type.go
```diff
@@ -247,6 +290,57 @@ func (a AddressType) DerivationScheme() DerivationScheme {
...
+	default:
+		return KeyScope{}, false
+	}
+}
+
+// ScopeAddrSchema returns the scoped-manager external/internal schema for the
+// address type.
+func (a AddressType) ScopeAddrSchema() (*ScopeAddrSchema, bool) {
+	switch a {
+	case NestedWitnessPubKey:
```
yyforyongyu, line 269: add inline comments above the case to explain this special case!

waddrmgr/addr_type.go
```diff
@@ -247,6 +290,57 @@ func (a AddressType) DerivationScheme() DerivationScheme {
...
+		return &ScopeAddrSchema{
+			ExternalAddrType: NestedWitnessPubKey,
+			InternalAddrType: WitnessPubKey,
+		}, true
+
+	case PubKeyHash, WitnessPubKey, TaprootPubKey:
```
yyforyongyu, line 277: inline comments above `case`

waddrmgr/addr_type.go
```diff
@@ -379,3 +413,110 @@ func (a AddressType) WitnessVersion() (byte, bool) {
...
+	net *chaincfg.Params) ([]byte, error) {
+
+	return nestedWitnessProgramFromSerializedPubKey(
+		pub.SerializeCompressed(), net,
+	)
+}
+
+// nestedWitnessProgramFromSerializedPubKey builds the nested witness redeem
+// program from a serialized public key.
+func nestedWitnessProgramFromSerializedPubKey(serializedPubKey []byte,
```
yyforyongyu: can we remove this thin wrapper and merge it into `nestedWitnessProgramFromPubKey`?

wallet/deprecated.go
```diff
@@ -795,7 +795,12 @@ func (w *Wallet) FundPsbtDeprecated(packet *psbt.Packet, keyScope *waddrmgr.KeyS
 				"change addr: %w", err)
 		}
 
-		changeOutputInfo, err := createOutputInfo(changeTxOut, addr)
```
yyforyongyu, line 798: we should not touch deprecated.go!

wallet/signer_test.go
```diff
@@ -1133,8 +1262,9 @@ func TestComputeUnlockingScriptUnknownAddrType(t *testing.T) {
 	// Act: Attempt to compute the unlocking script.
 	_, err = w.ComputeUnlockingScript(t.Context(), params)
 
-	// Assert: Verify that the unsupported address type error is returned.
-	require.ErrorIs(t, err, ErrUnsupportedAddressType)
+	// Assert: Verify that the unknown address type is rejected while building
+	// output metadata.
+	require.ErrorContains(t, err, "unsupported address type")
```
yyforyongyu: this commit is too big! break it down!

waddrmgr/addr_type_test.go
```diff
@@ -116,8 +116,6 @@ func TestPubKeyAddressTypeMetadata(t *testing.T) {
 	}
 
 	for _, testCase := range testCases {
-		testCase := testCase
```
yyforyongyu: this should be removed in the commit where it's introduced???

waddrmgr/addr_type_test.go
```diff
@@ -282,8 +278,6 @@ func TestSingleKeyAddressType(t *testing.T) {
...
```
yyforyongyu: same here - remove it in the commit where it's introduced!

waddrmgr/addr_type_test.go
```diff
@@ -356,8 +350,6 @@ func TestScriptAddressFromPubKey(t *testing.T) {
...
```
yyforyongyu: same here - remove it in the commit where it's introduced!

waddrmgr/scoped_manager.go
```diff
@@ -2286,73 +2286,6 @@ func (s *ScopedKeyManager) DeriveAddrs(account uint32, branch uint32,
 	return addrs, scripts, nil
 }
 
-// importPublicKey imports a public key into the address manager and updates the
-// wallet's start block if necessary. An error is returned if the public key
-// already exists.
-func (s *ScopedKeyManager) importPublicKey(ns walletdb.ReadWriteBucket,
```
yyforyongyu, line 2226: i think we should make the change where it's introduced? why do we need this change?

waddrmgr/interface.go
```diff
@@ -240,9 +240,6 @@ type AccountStore interface {
 	DeriveFromKeyPath(ns walletdb.ReadBucket,
 		path DerivationPath) (ManagedAddress, error)
 
-	// CanAddAccount returns an error if a new account cannot be created.
-	CanAddAccount() error
```
yyforyongyu, line 244: we should mark it as deprecated - maybe rename it to `CanAddAccountDeprecated` so we still have this path used in deprecated.go!

waddrmgr/manager.go
yyforyongyu: Addressed in the rewritten history. `Manager.lock()` now takes `manager.mtx.Lock()` before iterating the cache, and the matching unlock stays at the end of that per-manager section.

waddrmgr/scoped_manager.go
yyforyongyu: Added the inline comment above the nil guard. The delete callback can see a nil cache entry on cache-internal delete paths, so the guard is there before touching cached state.

waddrmgr/addr_type_test.go
yyforyongyu: Done. I simplified the witness-version assertions in the commit that introduces the metadata tests, so the later cleanup thread is now outdated.

waddrmgr/addr_type.go
yyforyongyu: Done. I removed the thin `nestedWitnessProgramFromPubKey` wrapper and now call the serialized-key helper directly from `nestedWitnessRedeemScriptFromPubKey`.

wallet/deprecated.go
yyforyongyu: Addressed in the history rewrite. The deprecated PSBT file is no longer touched by `wallet: use AddressType signing helpers`; the wallet-side refactor now keeps `deprecated.go` on the legacy helper signatures.

wallet/signer_test.go
yyforyongyu: Split this now. The old `wallet: use AddressType signing helpers` commit is replaced by two commits:\n\n- `cec978d73 wallet: use address metadata for script helpers`\n- `4c8e93221 wallet: use AddressType signer helpers`\n\nThat keeps the output/PSBT metadata move separate from the signer-specific dispatch and cache-miss fallback work. I also re-ran lint, focused tests, and `scripts/check-each-commit.sh` on the rewritten stack before force-pushing.

waddrmgr/manager.go
```diff
@@ -430,8 +430,10 @@ func (m *Manager) IsWatchOnlyAccount(ns walletdb.ReadBucket, keyScope KeyScope,
 // TODO(yy): Rename this to wipe memory for priv keys.
 func (m *Manager) lock() {
 	for _, manager := range m.scopedManagers {
+		manager.mtx.Lock()
```
yyforyongyu, line 433: inline comments above to explain why we need the lock?

waddrmgr/manager.go
```diff
@@ -1132,25 +1137,28 @@ func (m *Manager) ConvertToWatchingOnly(ns walletdb.ReadWriteBucket) error {
 
 	// Clear and remove all of the encrypted acount private keys.
 	for _, manager := range m.scopedManagers {
+		manager.mtx.Lock()
```
yyforyongyu, line 1142: same here - explain the lock

waddrmgr/manager.go
```diff
@@ -440,14 +442,17 @@ func (m *Manager) lock() {
...
 		// entries.
-		for _, ma := range manager.addresses() {
-			switch addr := ma.(type) {
-			case *managedAddress:
-				addr.lock()
-			case *scriptAddress:
+		manager.addrs.Range(func(_ addrKey, cachedAddr *cachedAddr) bool {
+			// Only address types that cache cleartext material implement the
+			// internal zeroing hook.
+			if addr, ok := cachedAddr.addr.(interface{ lock() }); ok {
```
yyforyongyu: can we properly define the interface instead?

yyforyongyu (commented):
LGTM 🗂️ **(`claude-opus-4-6`)**

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

wallet/signer.go
```diff
@@ -696,15 +696,7 @@ func (w *Wallet) ComputeUnlockingScript(ctx context.Context,
...
-	if !ok {
-		return nil, fmt.Errorf("%w: addr %s",
-			ErrNotPubKeyAddress, scriptInfo.Addr.Address())
-	}
-
-	// Get the private key for the derived address.
-	privKey, err := pubKeyAddr.PrivKey()
+	privKey, err := w.privKeyForOutput(scriptInfo.Addr)
 	if err != nil {
 		return nil, fmt.Errorf("cannot get private key: %w", err)
```
yyforyongyu: **Finding: Double error wrapping**

Location: `wallet/signer.go:699-702`

The call to `privKeyForOutput` at line 699 returns errors already wrapped by `loadManagedPubKeyAddr` with `"view signer address: %w"`. Then line 701 wraps again with `"cannot get private key: %w"`, resulting in error chains like:

    "cannot get private key: view signer address: fetch address: ..."

Consider either removing the outer wrap or unwrapping before re-wrapping if the additional context is needed.

`minimax-m2.7`

yyforyongyu:
Found 1 issue - see inline comment.

LGTM with nits (`minimax-m2.7`)

waddrmgr/manager.go
yyforyongyu: Addressed in the latest commit

waddrmgr/manager.go
yyforyongyu: Addressed in the latest commit

waddrmgr/manager.go
yyforyongyu: Addressed in the latest commit

wallet/signer.go
yyforyongyu: Addressed in the latest commit.

yyforyongyu:

yyforyongyu (commented):
LGTM 🌸 (`gpt-5.4`)

wallet/signer.go
```diff
@@ -697,14 +697,7 @@ func (w *Wallet) ComputeUnlockingScript(ctx context.Context,
...
 	// The address must be a public key address.
-	pubKeyAddr, ok := scriptInfo.Addr.(waddrmgr.ManagedPubKeyAddress)
-	if !ok {
-		return nil, fmt.Errorf("%w: addr %s",
-			ErrNotPubKeyAddress, scriptInfo.Addr.Address())
-	}
-
-	// Get the private key for the derived address.
-	privKey, err := pubKeyAddr.PrivKey()
+	privKey, err := w.privKeyForOutput(*params.Output)
```
yyforyongyu: we should add a defensive check in case of params.Output is nil?

waddrmgr/address.go
```diff
@@ -500,28 +505,33 @@ func (a *managedAddress) Validate(msg [32]byte, priv *btcec.PrivateKey) error {
 	//
 	// TODO(roasbeef): potentially run _all_ checks then see which one
 	// fails?
+	if !a.addrType.SupportsSingleKeySigning() {
```
yyforyongyu: I think the above todo note can be removed? `TODO(roasbeef):`

wallet/address_manager.go
```diff
@@ -5,7 +5,7 @@
 // Package wallet provides the AddressManager interface for generating and
 // inspecting wallet addresses and scripts.
 //
-// TODO(yy): bring wrapcheck back when implementing the `Store` interface.
+// This layer still forwards underlying walletdb and waddrmgr errors directly.
```
yyforyongyu: i don't think this is true?

wallet/address_manager.go
```diff
@@ -865,106 +818,92 @@ func buildScriptsForAddressInfo(addressInfo AddressInfo, pkScript []byte,
...
-		pubKey := pubKeyAddr.PubKey()
-		pubKeyHash := btcutil.Hash160(pubKey.SerializeCompressed())
-
-		// Next, we'll generate a valid sigScript that will allow us to
-		// spend the p2sh output. The sigScript will contain only a
-		// single push of the p2wkh witness program corresponding to
-		// the matching public key of this address.
-		p2wkhAddr, err := btcutil.NewAddressWitnessPubKeyHash(
-			pubKeyHash, chainParams,
+	if addressInfo.AddrType.RequiresRedeemScript() {
```
yyforyongyu: when will an address not require a redeem script?

yyforyongyu: and we can do `if !addressInfo.AddrType.RequiresRedeemScript() {` to exit fast?

wallet/common_test.go
```diff
@@ -260,3 +260,7 @@ func init() {
 	// timeouts, especially when running with -race.
 	waddrmgr.DefaultScryptOptions = waddrmgr.FastScryptOptions
 }
+
+func mustTestAddrType(addrType waddrmgr.AddressType) waddrmgr.AddressType {
```
yyforyongyu: why this thin wrapper?

wallet/psbt_manager.go
```diff
@@ -1386,49 +1371,73 @@ func createTaprootSpendDetails(pInput *psbt.PInput,
...
-		waddrmgr.TaprootScript:
-		return nil, fmt.Errorf("%w: %v", ErrUnknownAddressType,
-			addrType)
-	default:
 		return nil, fmt.Errorf("%w: %v", ErrUnknownAddressType,
 			addrType)
 	}
 
-	// Check if we have already signed this input.
-	for _, sig := range pInput.PartialSigs {
```
yyforyongyu, line 1437: and we no longer have this check?

wallet/psbt_manager.go
```diff
@@ -1386,49 +1371,73 @@ func createTaprootSpendDetails(pInput *psbt.PInput,
...
-		details = SegwitV0SpendDetails{WitnessScript: subScript}
+// bip32SpendDetailsForType maps one single-key address type into bip32 spend
+// details.
+func bip32SpendDetailsForType(
+	addrType waddrmgr.AddressType,
+	subScript []byte) (SpendDetails, error) {
 
-	case waddrmgr.PubKeyHash:
-		details = LegacySpendDetails{RedeemScript: subScript}
+	switch addrType.SigningMethod() {
```
yyforyongyu, line 1370: why don't we create a method on `addrType`, sth like `SpendDetails` that takes the `subScript`?

wallet/signer.go
```diff
@@ -786,10 +785,62 @@ func (w *Wallet) resolvePrivKey(pubKeyAddr waddrmgr.ManagedPubKeyAddress) (
 
 	accountManager, err := w.addrStore.FetchScopedKeyManager(keyScope)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fetch scoped key manager: %w", err)
+	}
+
+	privKey, err := accountManager.DeriveFromKeyPathCache(derivationPath)
+	if err != nil {
```
yyforyongyu: check `if err == nil` for fast exit?

waddrmgr/scoped_manager.go
```diff
@@ -718,6 +724,32 @@ type cachedKey struct {
...
+		lru.WithDeleteCallback(func(_ addrKey, cachedAddr *cachedAddr) {
+			// The callback may be invoked with a nil entry during cache-internal
+			// delete paths, so guard before touching cached state.
+			if cachedAddr == nil {
+				return
+			}
+
+			// This zeroes cached secret material on eviction. It is not a
+			// mutex lock, so there is no corresponding unlock step.
+			if addr, ok := cachedAddr.addr.(interface{ lock() }); ok {
```
yyforyongyu: use `secretWiper` here

waddrmgr/scoped_manager_test.go
```diff
@@ -11,6 +12,254 @@ import (
...
+
+		account, err = scopedMgr.NewAccount(ns, "acct-1")
+
+		return err
+	})
+	require.NoError(t, err)
+	require.Contains(t, scopedMgr.ActiveAccounts(), account)
+
+	scopedMgr.mtx.RLock()
+	acctInfo, ok := scopedMgr.acctInfo[account]
```
yyforyongyu, line 51: make sure to follow the AAA pattern so it's easier to read the tests

wallet/address_manager.go
```diff
@@ -965,44 +1022,73 @@ func (w *Wallet) GetDerivationInfo(ctx context.Context,
...
-	return derivationForManagedAddress(pubKeyAddr)
+	return derivationInfo, nil
 }
 
 // derivationForManagedAddress constructs a PSBT Bip32Derivation struct from a
 // managed public key address.
 func derivationForManagedAddress(pubKeyAddr waddrmgr.ManagedPubKeyAddress) (
 	*psbt.Bip32Derivation, error) {
 
-	// Imported addresses don't have derivation paths.
```
yyforyongyu, line 863: this is still not fixed - we still got the deletion of unnecessary comments here, quite noisy when reviewing

wallet/address_manager.go
```diff
@@ -965,44 +1022,73 @@ func (w *Wallet) GetDerivationInfo(ctx context.Context,
...
 	if pubKeyAddr.Imported() {
 		return nil, fmt.Errorf("%w: addr=%v is imported",
 			ErrDerivationPathNotFound, pubKeyAddr.Address())
 	}
 
-	// Get the derivation info.
```
yyforyongyu, line 869: this is still not fixed - we still got the deletion of unnecessary comments here, quite noisy when reviewing

wallet/address_manager.go
```diff
@@ -965,44 +1022,73 @@ func (w *Wallet) GetDerivationInfo(ctx context.Context,
...
 	keyScope, derivPath, ok := pubKeyAddr.DerivationInfo()
 	if !ok {
 		return nil, fmt.Errorf("%w: derivation info not found for %v",
 			ErrDerivationPathNotFound, pubKeyAddr.Address())
 	}
 
-	// Get the public key.
```
yyforyongyu, line 876: this is still not fixed - we still got the deletion of unnecessary comments here, quite noisy when reviewing

wallet/address_manager.go
```diff
@@ -965,44 +1022,73 @@ func (w *Wallet) GetDerivationInfo(ctx context.Context,
...
 	pubKey := pubKeyAddr.PubKey()
 
-	derivationInfo := &psbt.Bip32Derivation{
```
yyforyongyu, line 879: this change is also unnecessary?

wallet/signer.go
yyforyongyu: nvm i see it's changed to use a different param later

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,287 @@
...
+// address type.
+type AddressFormat uint8
+
+const (
+	// AddressFormatNone indicates no standard address serialization, such as
+	// RawPubKey.
+	AddressFormatNone AddressFormat = iota
+
+	// AddressFormatBase58 indicates base58-check encoding, for example
+	// PubKeyHash or Script.
```
yyforyongyu: we should give a real address example here, for mainet

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,287 @@
...
+	AddressFormatBase58
+
+	// AddressFormatBech32 indicates bech32 encoding, for example WitnessPubKey
+	// or WitnessScript.
```
yyforyongyu: we should give a real address example here, for mainet

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,287 @@
...
+	AddressFormatBech32
+
+	// AddressFormatBech32m indicates bech32m encoding, for example
+	// TaprootPubKey.
```
yyforyongyu: we should give a real address example here, for mainet

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,287 @@
...
+
+// SpendType captures the spend behavior implied by an address type.
+type SpendType uint8
+
+const (
+	// SpendTypeUnknown is used when no standard spend template applies, such as
+	// RawPubKey.
+	SpendTypeUnknown SpendType = iota
+
+	// SpendTypeLegacyKey is a direct legacy single-key spend, for example
```
yyforyongyu: i don't think the examples given here are useful at all, that it's just a single word?

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,287 @@
...
+// type.
+type DerivationScheme uint8
+
+const (
+	// DerivationSchemeNone indicates that no standard account derivation scheme
+	// applies, for example Script or TaprootScript.
+	DerivationSchemeNone DerivationScheme = iota
+
+	// DerivationSchemeBIP44 indicates BIP44 account derivation, for example
+	// PubKeyHash.
```
yyforyongyu: the example should show the full path, like `purpose/account/...`

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,287 @@
...
+		return p2wpkhPkScriptSize, true
+
+	case WitnessScript:
+		return p2wshPkScriptSize, true
+
+	case TaprootPubKey, TaprootScript:
+		return p2trPkScriptSize, true
+
+	default:
+		return 0, false
```
yyforyongyu: i am thinking maybe we should use `panic` here instead of returning two vars?

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,287 @@
...
+func (a AddressType) WitnessVersion() (byte, bool) {
+	switch a {
+	case NestedWitnessPubKey, WitnessPubKey, WitnessScript:
+		return witnessVersionV0, true
+
+	case TaprootPubKey, TaprootScript:
+		return witnessVersionV1, true
+
+	default:
+		return 0, false
```
yyforyongyu: i am thinking maybe we should use `panic` here instead of returning two vars?

waddrmgr/addr_type.go
yyforyongyu: or return an error instead of bool?

waddrmgr/addr_type.go
```diff
@@ -155,6 +166,29 @@ const (
...
 
+// SigningMethod identifies the single-key signing flow associated with an
+// address type.
+type SigningMethod uint8
+
+const (
+	// SigningMethodUnsupported indicates that the address type does not map
+	// to a wallet-supported single-key signing flow, for example Script or
+	// WitnessScript.
+	SigningMethodUnsupported SigningMethod = iota
```
yyforyongyu, line 126: we should exhast the list of unsupported signing methods here, and list them using bullets

waddrmgr/addr_type.go
```diff
@@ -383,3 +417,102 @@ func (a AddressType) WitnessVersion() (byte, bool) {
 		return 0, false
 	}
 }
+
+// SupportsSingleKeySigning reports whether the address type supports a direct
+// wallet single-key signing path.
+func (a AddressType) SupportsSingleKeySigning() bool {
```
yyforyongyu: this method is redudant - we can either do `a.SigningMethod != SigningMethodUnsupported` to achieve the same goal in the callsite, or rename this method to be `SigningSupported` and check `SigningMethodUnsupported` inside it, then add a big note explaining what are supported now

yyforyongyu: and remove the concept of "single key signing" here from the docs and errors - it just creates more maintenance burden when we decide to support other signing methods in the future. we only need one place to document the supported methods.

waddrmgr/addr_type.go
```diff
@@ -383,3 +417,102 @@ func (a AddressType) WitnessVersion() (byte, bool) {
...
+		return SigningMethodTaprootKeySpend
+
+	default:
+		return SigningMethodUnsupported
+	}
+}
+
+// RequiresRedeemScript reports whether the address type needs a redeem script
+// for its single-key signing path.
+func (a AddressType) RequiresRedeemScript() bool {
```
yyforyongyu: this is not useful, one can just call `RedeemScriptFromPubKey` and check the error instead in the callsite

waddrmgr/address.go
```diff
@@ -500,28 +505,33 @@ func (a *managedAddress) Validate(msg [32]byte, priv *btcec.PrivateKey) error {
...
+
 	var sig signature
 
 	addrPrivKey, _ := btcec.PrivKeyFromBytes(a.privKeyCT)
 
-	switch a.addrType {
-	// For the "legacy" addr types, we'll generate an ECDSA signature to
-	// verify against.
-	case NestedWitnessPubKey, PubKeyHash, WitnessPubKey:
+	switch a.addrType.SigningMethod() {
```
yyforyongyu, line 509: it seems like we can add a method on addr type, such as `SignMsg`, that takes the private key and msg and it can chooses to use `ecdsa` or `schnorr` there?

wallet/signer_test.go
yyforyongyu: why move away from `require.ErrorIs`?

wallet/common_test.go
yyforyongyu: Done. I removed `mustTestAddrType` and the test now passes `waddrmgr.NestedWitnessPubKey` directly.

wallet/psbt_manager.go
yyforyongyu: The separate pre-check is gone, but the guard is still there in one place: `createBip32SpendDetails` now switches directly on `addrType.SigningMethod()` and returns `ErrUnknownAddressType` for unsupported methods from that single dispatch.

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,295 @@
...
+// type.
+type OutputType uint8
+
+const (
+	// OutputTypeUnknown is used for wallet-managed address types that do not
+	// map cleanly to a standard user-facing output form, such as RawPubKey.
+	OutputTypeUnknown OutputType = iota
+
+	// OutputTypeP2PKH is a pay-to-pubkey-hash output, for example the mainnet
+	// address 1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa.
```
yyforyongyu: write like this,
```
// OutputTypeP2PKH is a pay-to-pubkey-hash output, for example:
- mainnet: 1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa
- testnet: xxx
- regtest: xxx
```

same for the rest

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,295 @@
...
+	// bc1qqypqxpq9qcrsszg2pvxq6rs0zqg3yyc5z5tpwxqergd3c8g7rusqyp0mu0.
+	OutputTypeP2WSH
+
+	// OutputTypeP2TR is a native taproot output, for example the
+	// mainnet address
+	// bc1pqypqxpq9qcrsszg2pvxq6rs0zqg3yyc5z5tpwxqergd3c8g7rusqwk0jyn.
+	OutputTypeP2TR
+
+	// OutputTypeP2A is a pay-to-anchor output, for example a future anchor-only
+	// address type.
```
yyforyongyu: document the example address?

wallet/psbt_manager.go
```diff
@@ -430,34 +429,22 @@ func (w *Wallet) decorateInput(ctx context.Context, pInput *psbt.PInput,
...
-	)
-
-	err = walletdb.View(w.cfg.DB, func(tx walletdb.ReadTx) error {
-		addrmgrNs := tx.ReadBucket(waddrmgrNamespaceKey)
-
-		managedAddr, err = w.addrStore.Address(addrmgrNs, addr)
-
-		return err
-	})
+	addressInfo, err := w.GetAddressInfo(ctx, addr)
```
yyforyongyu: we changed it in 01d20ee213f420251663f9fb0ad739012698a510 but now changed back???

waddrmgr/addr_type.go
```diff
@@ -48,6 +48,13 @@ var (
 	ErrNotPubKeyAddressType = errors.New(
 		"address type is not pubkey-constructible",
 	)
+
+	// ErrMissingSerializedPubKeyConstructor is returned when a pubkey-based
+	// address type does not provide the serialized-pubkey helper required by
+	// address reconstruction.
+	ErrMissingSerializedPubKeyConstructor = errors.New(
```
yyforyongyu: we can just use `ErrUnsupportedAddressType`? why creating this new error?

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,700 @@
...
+	// for a lookup key.
+	ErrUnknownAddressType = errors.New("unknown address type")
+
+	// ErrUnsupportedAddressType is returned when an address type is known but a
+	// specific API does not support it.
+	ErrUnsupportedAddressType = errors.New("unsupported address type")
+
+	// ErrNotPubKeyAddressType is returned when the caller asks for pubkey-based
+	// behavior from an address type that is not pubkey-constructible.
+	ErrNotPubKeyAddressType = errors.New(
```
yyforyongyu: use `ErrUnsupportedAddressType` instead?

wallet/address_manager.go
yyforyongyu: Same issue here: the current pushed history still introduces the metadata-based derivation helper too early. I am keeping the old managed-address derivation path in the lookup-only commit and moving the metadata helper to the later PSBT commit where it first becomes useful.

wallet/address_manager.go
yyforyongyu: Agreed. These comment deletions are still noise in the pushed branch. The fix is the same history move: keep this commit lookup-only, and first introduce `derivationForAddressInfo` in the later PSBT metadata commit.

wallet/address_manager.go
yyforyongyu: Yes, this is part of the same early refactor churn. In the cleaner history this helper does not change here at all; the metadata-based derivation helper lands later together with the PSBT metadata adoption.

waddrmgr/addr_type.go
yyforyongyu: Yes. I am updating these docs to use fuller example blocks instead of the current terse form so the example address is documented directly in the comment.

wallet/address_manager.go
yyforyongyu: Agreed. This is still not fixed on the currently pushed branch. The early lookup-only commit should not introduce the metadata-based derivation helper or delete these comments. I am moving `derivationForAddressInfo` into the later PSBT-metadata commit so this earlier commit stays reviewable.

wallet/common_test.go
yyforyongyu: This one is fixed on the current pushed branch: the thin `mustTestAddrType` helper is gone and the PSBT test passes `waddrmgr.NestedWitnessPubKey` directly. I am resolving this thread.

wallet/psbt_manager.go
yyforyongyu: You are right. In the currently pushed history this looks like it changed back, which is a history-shape problem. The intended shape is: keep the old managed-address derivation path in the early lookup commit, and only switch to `derivationForAddressInfo` later in the PSBT metadata commit.

waddrmgr/addr_type.go
yyforyongyu: Yes. Same change here: this should use `ErrUnsupportedAddressType` instead of carrying another specialized exported error for the unsupported case.

waddrmgr/addr_type.go
yyforyongyu: Agreed. These narrower exported errors should be collapsed into `ErrUnsupportedAddressType`. I am removing the extra public error surface in the local rewrite.

yyforyongyu (commented):
LGTM 🗂️ **(`claude-opus-4-6`)**

yyforyongyu:

yyforyongyu:
LGTM 🚀 (`minimax-m2.7`)

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

waddrmgr/scoped_manager.go
```diff
@@ -1885,7 +1948,16 @@ func (s *ScopedKeyManager) newAccount(ns walletdb.ReadWriteBucket,
 	}
 
 	// Save last account metadata
-	return putLastAccount(ns, &s.scope, account)
+	err = putLastAccount(ns, &s.scope, account)
+	if err != nil {
+		return err
+	}
+
+	// Warm the account cache explicitly so callers do not need a follow-up
```
yyforyongyu, line 1966: This eager cache warm runs while the wallet is still unlocked, and `loadAccountInfo` materializes `lastExternalAddr`/`lastInternalAddr` as full managed addresses with cleartext `privKeyCT` bytes. `Manager.lock()` only wipes `acctKeyPriv` and entries in `manager.addrs`, so those `acctInfo.last*Addr` keys remain resident after lock. Please either avoid warming those last-address objects while unlocked or teach the lock path to wipe them too. (`gpt-5.4`)

yyforyongyu: Fixed in the rewritten branch and force-pushed.

`Manager.lock()` now wipes `acctInfo.lastExternalAddr` and `acctInfo.lastInternalAddr` in addition to the LRU cache entries, so the eager `loadAccountInfo` warm path no longer leaves `privKeyCT` resident after lock.

I also factored the wipe through `wipeAddressSecret()` so the account-level and cache-eviction paths use the same zeroing hook, and added regression coverage in `TestManagerLockWipesLastAccountAddresses`.

Local verification on the pushed branch:
- `GOWORK=off go test ./waddrmgr ./wallet`
- `make lint`
- `GOWORK=off scripts/check-each-commit.sh sql-wallet`

wallet/address_manager.go
```diff
@@ -839,54 +915,48 @@ func (w *Wallet) GetDerivationInfo(ctx context.Context,
...
+	return &psbt.Bip32Derivation{
+		PubKey:               addressInfo.PubKey.SerializeCompressed(),
+		MasterKeyFingerprint: addressInfo.Derivation.MasterKeyFingerprint,
 		Bip32Path: []uint32{
 			keyScope.Purpose + hdkeychain.HardenedKeyStart,
 			keyScope.Coin + hdkeychain.HardenedKeyStart,
-			derivPath.Account,
-			derivPath.Branch,
-			derivPath.Index,
+			addressInfo.Derivation.Account,
```
yyforyongyu: The account element in this emitted BIP32 path is still unhardened, but the wallet parser requires the first three elements to be hardened. That makes wallet-produced derivation metadata internally inconsistent and can break later PSBT import or signing flows. The same path construction is mirrored in the change-output PSBT helper, so both call sites need the hardened account index. (`gpt-5.4`)

yyforyongyu: Fixed and force-pushed.

Both wallet-owned BIP32 path builders now harden the account element to stay consistent with the wallet parser requirement that the first three path elements are hardened:
- `derivationForAddressInfo` in `wallet/address_manager.go`
- `createOutputInfoFromAddressInfo` in `wallet/psbt_manager.go`

I also updated the direct derivation-info expectations in `wallet/address_manager_test.go` to match the corrected emitted paths.

Local verification on the pushed branch:
- `GOWORK=off go test ./wallet ./waddrmgr`
- `make lint`
- `GOWORK=off scripts/check-each-commit.sh sql-wallet`

yyforyongyu (commented):
LGTM 🌸 (`gpt-5.4`)

waddrmgr/addr_type.go
```diff
@@ -279,3 +308,20 @@ func (a AddressType) WitnessVersion() (byte, error) {
...
+		return SigningMethodLegacy
+
+	case NestedWitnessPubKey, WitnessPubKey:
+		return SigningMethodWitnessV0
+
+	case TaprootPubKey:
+		return SigningMethodTaprootKeySpend
+
+	default:
+		return SigningMethodUnsupported
```
yyforyongyu: I think it's better to return an error here saying `ErrUnsupportedAddressType` and `ErrUnknownAddressType`? defining a new `SigningMethodUnsupported` doesn't help here, as the real logic is we check whether we support this address here or not in the wallet?

wallet/address_manager.go
```diff
@@ -863,103 +805,101 @@ func buildScriptsForAddressInfo(addressInfo AddressInfo, pkScript []byte,
...
+			addressInfo.PubKey, w.cfg.ChainParams,
+		)
+		if err != nil {
+			return OutputScriptInfo{}, err
+		}
+
+		// For nested P2WPKH-in-P2SH, the redeem script committed by the outer
+		// P2SH output is the same inner witness program used for signing.
+		witnessProgram = redeemScript
+	} else if addressInfo.AddrType != waddrmgr.PubKeyHash &&
```
yyforyongyu, line 823: add a newline above?

wallet/address_manager_test.go
```diff
@@ -494,7 +494,7 @@ func TestGetDerivationInfoExternalAddressSuccess(t *testing.T) {
 	expectedPath := []uint32{
 		scope.Purpose + hdkeychain.HardenedKeyStart,
 		scope.Coin + hdkeychain.HardenedKeyStart,
-		path.Account,
+		path.Account + hdkeychain.HardenedKeyStart,
```
yyforyongyu, line 497: shouldn't this be squashed to a previous commit where it's introduced?

wallet/address_manager_test.go
```diff
@@ -560,7 +560,7 @@ func TestGetDerivationInfoInternalAddressSuccess(t *testing.T) {
...
```
yyforyongyu, line 563: shouldn't this be squashed to a previous commit where it's introduced?

wallet/psbt_manager.go
```diff
@@ -2287,7 +2242,7 @@ func createOutputInfoFromAddressInfo(txOut *wire.TxOut,
 		Bip32Path: []uint32{
 			addr.Derivation.KeyScope.Purpose + hdkeychain.HardenedKeyStart,
 			addr.Derivation.KeyScope.Coin + hdkeychain.HardenedKeyStart,
-			addr.Derivation.Account,
+			addr.Derivation.Account + hdkeychain.HardenedKeyStart,
```
yyforyongyu, line 2249: shouldn't this be squashed to a previous commit where it's introduced?

wallet/signer.go
```diff
@@ -786,10 +785,56 @@ func (w *Wallet) resolvePrivKey(pubKeyAddr waddrmgr.ManagedPubKeyAddress) (
...
+		return nil, fmt.Errorf("fetch scoped key manager: %w", err)
+	}
+
+	privKey, err := accountManager.DeriveFromKeyPathCache(derivationPath)
+	if err == nil {
+		return privKey, nil
+	}
+
+	if !waddrmgr.IsError(err, waddrmgr.ErrAccountNotCached) {
+		return nil, fmt.Errorf("derive private key from cache: %w", err)
```
yyforyongyu, line 800: should we also fetch the db to try finding the private key?

wallet/signer_test.go
```diff
@@ -267,6 +267,31 @@ func expectDerivedSignerPrivKey(t *testing.T, mocks *mockWalletDeps,
 		Return(privKey, nil).Once()
 }
 
+// expectDerivedSignerPrivKeyAfterCacheMiss wires the signer-private key lookup
+// path for a derived managed pubkey address after the account-key cache misses.
+func expectDerivedSignerPrivKeyAfterCacheMiss(t *testing.T,
```
yyforyongyu: unwrap this method in the test instead - it's better to have a linear flow in the tests than creating this method that only gets used once?

waddrmgr/addr_type_test.go
```diff
@@ -235,7 +235,6 @@ func TestSupportedSigningAddressType(t *testing.T) {
 			signingMethod: SigningMethodTaprootKeySpend,
 		},
 	}
-
```
yyforyongyu: we should squash the change into the commit that introduced this???

waddrmgr/scoped_manager.go
```diff
@@ -2306,49 +2306,27 @@ func (s *ScopedKeyManager) DeriveAddrs(account uint32, branch uint32,
 // importPublicKey imports a public key into the address manager and updates the
 // wallet's start block if necessary. An error is returned if the public key
 // already exists.
+//
+//nolint:funcorder
```
yyforyongyu: why are we still adding this nolint?

waddrmgr/addr_type.go
```diff
@@ -321,7 +345,112 @@ func (a AddressType) SigningMethod() SigningMethod {
...
+
 	default:
 		return SigningMethodUnsupported
 	}
 }
+
+// addrFromPubKeyBytes reconstructs the standard address from the serialized
+// public key form stored by the wallet.
+func (a AddressType) addrFromPubKeyBytes(pubKeyBytes []byte,
```
yyforyongyu: we should export this method??

waddrmgr/scoped_manager.go
```diff
@@ -718,6 +724,49 @@ type cachedKey struct {
...
 
+// cachedAddr is an address-cache entry in the scoped manager LRU.
+type cachedAddr struct {
+	addr ManagedAddress
+}
+
+// secretWiper is implemented by cached address types that can eagerly zero
+// sensitive material from memory.
+type secretWiper interface {
+	lock()
```
yyforyongyu, line 735: I think we can rename this method during the key vault impl - what this method really does is to zero out the bytes, not "locking". @GustavoStingelin

yyforyongyu (commented):
should be good to review now

waddrmgr/addr_type.go
yyforyongyu: Addressed. `SigningMethod()` now returns `(SigningMethod, error)` and reports `ErrUnsupportedAddressType` / `ErrUnknownAddressType` directly instead of using a sentinel `SigningMethodUnsupported` value.

wallet/address_manager_test.go
yyforyongyu: Yes. Folded back into the commit that introduced the expectation in the rewritten history.

wallet/address_manager.go
yyforyongyu: Fixed and folded back into the introducing commit in the rewritten history.

wallet/psbt_manager.go
yyforyongyu: Yes. This is squashed into the PSBT helper commit now rather than left as a follow-up change.

wallet/signer.go
yyforyongyu: We already do the DB-backed fallback for the cache-miss case (`ErrAccountNotCached`). I kept that behavior and added a clarifying comment that only a cold account cache should trigger the slower fallback; other derivation errors are treated as real failures rather than retried through the DB.

wallet/address_manager_test.go
yyforyongyu: Yes. Folded back into the introducing commit in the rewritten history.

waddrmgr/addr_type_test.go
yyforyongyu: Yes. Folded into the commit that introduced the test change in the rewritten history.

waddrmgr/addr_type.go
yyforyongyu: Done. `AddrFromPubKeyBytes` is exported now and the callers were updated to use the exported method.

waddrmgr/scoped_manager.go
yyforyongyu: Yes. That earlier move was just to remove the `funcorder` `nolint`. I changed it again so `importPublicKey` is placed next to the import entry points and the `nolint` is gone.

wallet/signer_test.go
yyforyongyu: Done. I inlined that one-use helper so the cache-miss test now reads linearly.

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,445 @@
...
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
+	"github.com/btcsuite/btcd/txscript"
+)
+
+const (
+	// p2pkhPkScriptSize is the fixed scriptPubKey size for P2PKH outputs.
+	p2pkhPkScriptSize = 25
```
GustavoStingelin, line 18: I think it's worth adding a little more context in these godocs.
```suggestion
	// p2pkhPkScriptSize is the fixed scriptPubKey size for P2PKH outputs.
	// 25 bytes = OP_DUP (1) + OP_HASH160 (1) + OP_PUSHBYTES_20 (1) +
	// <pubkey hash> (20) + OP_EQUALVERIFY (1) + OP_CHECKSIG (1).
	p2pkhPkScriptSize = 25
```

waddrmgr/addr_type.go
```diff
@@ -0,0 +1,445 @@
...
+	// pkScript, for example PubKeyHash.
+	SigningMethodLegacy SigningMethod = iota
+
+	// SigningMethodWitnessV0 indicates SegWit v0 ECDSA signing, for example
+	// WitnessPubKey or NestedWitnessPubKey.
+	SigningMethodWitnessV0
+
+	// SigningMethodTaprootKeySpend indicates Taproot key-path Schnorr signing,
+	// for example TaprootPubKey.
+	SigningMethodTaprootKeySpend
```
GustavoStingelin, line 136: could be better if we call this `WitnessV1` or `WitnessV1TaprootKeySpend`?

GustavoStingelin (approved):
LGTM

GustavoStingelin:
I still do not have a full picture of how the iterations btw `waddrmgr` and `keyvault` will be, but let's see in next weeks.

waddrmgr/addr_type.go
yyforyongyu: updated

waddrmgr/addr_type.go
yyforyongyu: feel like taproot key spend is good enough - maybe the next time we will do a quantum signing, but can imagine that won't happen soon

PR #1206, opened 2026-04-07
wallet: improve concurrent itests

GustavoStingelin:
Just some itests improvements, removing error asserts inside goroutines and sleep dependency.


yyforyongyu (approved):
💯

PR #1212, opened 2026-04-14
db: apply runtime read/write helpers

yyforyongyu:
## Summary
- add backend-local runtime helpers in sqlite and postgres, then route store operations through lnd-style `execRead` and `execWrite` wrappers backed by the shared `runtime.Read` and `runtime.Write` helpers
- keep retry, SQL error classification, unhealthy-store handling, and ambiguous-commit behavior at the store-operation boundary while removing the old shared `db.ExecInTx` helper
- verify end-to-end SQL error classification in `wallet/internal/db/itest` with real duplicate constraint failures for both SQLite and PostgreSQL

## Commit Layout
- `db/sqlite: add runtime store helpers`
- `db/sqlite: route write paths through runtime`
- `db/sqlite: route read paths through runtime`
- `db/pg: add runtime store helpers`
- `db/pg: route write paths through runtime`
- `db/pg: route read paths through runtime`
- `db: drop legacy tx helper`
- `db/itest: verify SQL error classification`

## Testing
- `GOWORK=off go test ./wallet/internal/db/runtime ./wallet/internal/db/sqlite ./wallet/internal/db/pg ./wallet/internal/db/itest`
- `GOWORK=off go test ./wallet/internal/db/...`
- `make lint`
- `GOWORK=off make itest-db db=sqlite verbose=1`
- `GOWORK=off make itest-db db=postgres verbose=1`
- `GOWORK=off scripts/check-each-commit.sh sql-wallet`

## Benchmarks
Test [here](https://gist.github.com/yyforyongyu/dd2b3f2dcf48e40810bc20c00cec27dd).

- compared local sqlite store benchmarks between `sql-wallet` and `apply-sql-err` using `GOWORK=off go test ./wallet/internal/db/sqlite -run '^$' -bench 'Benchmark(GetWallet|UpdateWalletSecrets|CreateWallet|CreateWalletDuplicate)$' -benchmem -count=5`
- `GetWallet` was effectively unchanged (~23.84 us/op on `sql-wallet` vs ~23.91 us/op on `apply-sql-err`)
- duplicate create classification stayed close (~30.05 us/op on `sql-wallet` vs ~31.34 us/op on `apply-sql-err`)
- successful writes showed a small increase (~8.03 us/op vs ~11.64 us/op for `UpdateWalletSecrets`), which looked acceptable for now and did not justify additional optimization in this PR


wallet/internal/db/pg/utxostore_getutxo.go
```diff
@@ -23,26 +23,37 @@ func (s *Store) GetUtxo(ctx context.Context,
...
-			OutputIndex: outputIndex,
-		},
-	)
-	if err != nil {
-		if errors.Is(err, sql.ErrNoRows) {
-			return nil, fmt.Errorf("utxo %s: %w", query.OutPoint,
-				db.ErrUtxoNotFound)
+	var utxo *db.UtxoInfo
+
+	err = s.execRead(ctx, func(q *sqlc.Queries) error {
```
yyforyongyu, line 28: It looks like you are mutating the `utxo` variable as a closure side-effect inside `s.execRead(...)`. `dbruntime.Read` actually natively supports returning a generic type `T`, so you could simply change `execRead` to be generic and return `*db.UtxoInfo` directly instead of dealing with external closure mutations. This is a very minor style note, so feel free to ignore it if you prefer keeping `execRead` perfectly homogeneous across calls without returning types! (`gemini-3.1-pro-preview`)

yyforyongyu:
LGTM 🚀 (`minimax-m2.7`)

yyforyongyu (commented):
LGTM 🔁 **(`claude-opus-4-6`)**

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

wallet/internal/db/pg/utxostore_getutxo.go
yyforyongyu: Good point. I considered keeping a generic result-returning helper here, but I ended up intentionally keeping the backend-facing executor surface homogeneous with `execWrite`, following the simpler lnd-style error-only wrapper shape. That leaves a few outer result vars at call sites, but it keeps the sqlite/pg helper API smaller and more uniform, so I would prefer to keep it as-is.

wallet/internal/db/sqlite/runtime.go
```diff
@@ -0,0 +1,138 @@
...
+	"sync/atomic"
+	"time"
+
+	dberr "github.com/btcsuite/btcwallet/wallet/internal/db/err"
+	dbruntime "github.com/btcsuite/btcwallet/wallet/internal/db/runtime"
+	sqlc "github.com/btcsuite/btcwallet/wallet/internal/sql/sqlite/sqlc"
+)
+
+var (
+	_ dbruntime.ReadHooks  = (*Store)(nil)
```
yyforyongyu, line 16: we need to add comments explaining this compile time assertion?

wallet/internal/db/sqlite/runtime.go
```diff
@@ -0,0 +1,149 @@
...
+	_ dbruntime.WriteHooks = (*Store)(nil)
+)
+
+// Default SQLite read retry settings.
+const (
+	defaultReadMaxAttempts = 3
```
yyforyongyu, line 24: we should add inline comments to explain each value, the rationales behind the values?

wallet/internal/db/pg/runtime.go
```diff
@@ -0,0 +1,149 @@
...
+	"sync/atomic"
+	"time"
+
+	dberr "github.com/btcsuite/btcwallet/wallet/internal/db/err"
+	dbruntime "github.com/btcsuite/btcwallet/wallet/internal/db/runtime"
+	sqlc "github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
+)
+
+var (
+	_ dbruntime.ReadHooks  = (*Store)(nil)
```
yyforyongyu, line 16: same here re inline comments

wallet/internal/db/pg/runtime.go
```diff
@@ -0,0 +1,149 @@
...
+	_ dbruntime.WriteHooks = (*Store)(nil)
+)
+
+// Default PostgreSQL read retry settings.
+const (
+	defaultReadMaxAttempts = 3
```
yyforyongyu, line 24: same here re inline comments

wallet/internal/db/pg/store.go
```diff
@@ -14,8 +14,14 @@ import (
...
 type Store struct {
-	db      *sql.DB
+	// db is the shared PostgreSQL connection pool.
+	db *sql.DB
+
+	// queries executes PostgreSQL statements on db.
 	queries *sqlc.Queries
+
+	// runtimeState tracks shared runtime state.
+	runtimeState storeRuntimeState
```
yyforyongyu: rename it to `stats` or `runtimeStats` or `dbStats`?

wallet/internal/db/itest/fixtures_pg_test.go
```diff
@@ -5,16 +5,36 @@ package itest
...
 	"math"
 	"testing"
 
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	dberr "github.com/btcsuite/btcwallet/wallet/internal/db/err"
 	"github.com/btcsuite/btcwallet/wallet/internal/db/pg"
-	"github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
+	sqlc "github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
 	"github.com/stretchr/testify/require"
 )
```
yyforyongyu, line 19: we should add more tests to check the error cases, also check the db stats are collected as expected? create a new commit to add more tests.

wallet/internal/db/sqlite/runtime.go
yyforyongyu: Addressed in `c58326813`. I added comments above the compile-time hook assertions so the intent is explicit.

wallet/internal/db/sqlite/runtime.go
yyforyongyu: Addressed in `c58326813`. I added per-value comments explaining why the default retry settings use 3 attempts with a 10 ms base delay and a 100 ms cap.

wallet/internal/db/pg/store.go
yyforyongyu: Renamed to `runtimeStats` in `c58326813`. I chose that over `dbStats` because the field also carries runtime-specific state like the unhealthy flag, retry counters, and ambiguous commit counters, not just raw DB error counts.

wallet/internal/db/pg/runtime.go
yyforyongyu: Addressed in `c58326813`. I added the same per-value rationale comments for the postgres retry defaults.

wallet/internal/db/pg/runtime.go
yyforyongyu: Addressed in `c58326813`. The postgres helper file now has the same compile-time assertion comment as sqlite.

wallet/internal/db/itest/fixtures_pg_test.go
yyforyongyu: Added in a follow-up commit `963d94ca9`. The itest branch now exposes backend stats snapshots for test-only use and verifies both that duplicate constraint failures increment the classified SQL error counters and that a validation error path (`ListWallets` with zero limit) leaves the counters unchanged.

yyforyongyu (commented):
LGTM 📊 **(`claude-opus-4-7`)**

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

yyforyongyu:

yyforyongyu:
LGTM 🚀 (`minimax-m2.7`)

wallet/internal/db/pg/runtime.go
```diff
@@ -0,0 +1,159 @@
...
+	if s.runtimeStats.unhealthy.Load() {
+		return dbruntime.ErrStoreUnhealthy
+	}
+
+	return nil
+}
+
+// ClassifyError normalizes one PostgreSQL backend error into the shared SQL
+// error model.
+func (s *Store) ClassifyError(err error) error {
```
yyforyongyu, line 86: `ClassifyError` now normalizes every callback error for a known backend, and `Normalize` falls back to `ReasonUnknown` when the error is not a driver/backend failure. That means ordinary domain errors returned from the new read/write callbacks—such as wallet-not-found, tx-not-found, input-conflict, or invalid-parent—are still being turned into recorded SQL errors instead of passing through unchanged. Please preserve non-SQL callback errors here; the same issue applies to the SQLite runtime wrapper. (`gpt-5.4`)

wallet/internal/db/sqlite/runtime.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	return nil
+}
+
+// ClassifyError normalizes one SQLite backend error into the shared SQL error
+// model.
+func (s *Store) ClassifyError(err error) error {
+	return dberr.Normalize(dberr.BackendSQLite, mapErr, err)
+}
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in `723d3b12f`. The SQLite runtime classifier now preserves ordinary callback/domain errors unchanged and only returns the shared SQL wrapper for actual backend or transport failures.

wallet/internal/db/pg/runtime.go
yyforyongyu: Fixed in `723d3b12f`. `ClassifyError` now only returns the normalized SQL wrapper when the backend classifier found a concrete SQL/transport condition (or the error was already an `*dberr.SQLError`). Plain domain errors now pass through unchanged instead of being recorded as `ReasonUnknown` SQL errors.

wallet/internal/db/sqlite/addresses.go
```diff
@@ -19,19 +19,32 @@ var _ db.AddressStore = (*Store)(nil)
...
-			}, addressRowToInfo,
-		)
+	var info *db.AddressInfo
+
+	err := s.execRead(ctx, func(qtx *sqlc.Queries) error {
+		getByScript := func(ctx context.Context,
+			q db.GetAddressQuery) (*db.AddressInfo, error) {
+
+			return db.GetAddress(
+				ctx, qtx.GetAddressByScriptPubKey,
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in `723d3b12f`. I renamed the read callback handle from `qtx` to `q` here so it no longer looks transactional.

wallet/internal/db/pg/runtime.go
```diff
@@ -0,0 +1,159 @@
...
+	return dberr.Normalize(dberr.BackendPostgres, mapErr, err)
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in `723d3b12f`. The PostgreSQL runtime classifier now preserves ordinary callback/domain errors unchanged and only returns the shared SQL wrapper for actual backend or transport failures.

wallet/internal/db/pg/addresses.go
```diff
@@ -19,19 +19,32 @@ var _ db.AddressStore = (*Store)(nil)
...
-			}, addressRowToInfo,
-		)
+	var info *db.AddressInfo
+
+	err := s.execRead(ctx, func(qtx *sqlc.Queries) error {
+		getByScript := func(ctx context.Context,
+			q db.GetAddressQuery) (*db.AddressInfo, error) {
+
+			return db.GetAddress(
+				ctx, qtx.GetAddressByScriptPubKey,
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in `723d3b12f`. I renamed the read callback handle from `qtx` to `q` here too so the read path naming stays consistent.

yyforyongyu (commented):
LGTM 🌟 (`gpt-5.4`)

wallet/internal/db/itest/fixtures_pg_test.go
```diff
@@ -5,16 +5,36 @@ package itest
...
+	"errors"
 	"fmt"
 	"math"
 	"testing"
 
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	dberr "github.com/btcsuite/btcwallet/wallet/internal/db/err"
 	"github.com/btcsuite/btcwallet/wallet/internal/db/pg"
-	"github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
+	sqlc "github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
```
GustavoStingelin: nit: 
```suggestion
	"github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
```

wallet/internal/db/itest/fixtures_sqlite_test.go
```diff
@@ -5,16 +5,36 @@ package itest
...
+	"errors"
 	"fmt"
 	"math"
 	"testing"
 
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	dberr "github.com/btcsuite/btcwallet/wallet/internal/db/err"
 	"github.com/btcsuite/btcwallet/wallet/internal/db/sqlite"
-	"github.com/btcsuite/btcwallet/wallet/internal/sql/sqlite/sqlc"
+	sqlc "github.com/btcsuite/btcwallet/wallet/internal/sql/sqlite/sqlc"
```
GustavoStingelin: nit: 
```suggestion
	"github.com/btcsuite/btcwallet/wallet/internal/sql/sqlite/sqlc"
```

wallet/internal/db/pg/accounts.go
```diff
@@ -17,35 +17,64 @@ var _ db.AccountStore = (*Store)(nil)
...
+		)
+
+		return err
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return account, nil
 }
```
GustavoStingelin, line 38: IIUC we are mimicking LND here and it works, but I wonder if we could make `execRead`/`execWrite` return a generic typed value instead. That would avoid mutating an outer variable and make the flow a bit cleaner.

wallet/internal/db/pg/errors_test.go
```diff
@@ -1,14 +1,18 @@
 package pg
 
 import (
+	"errors"
 	"io"
 	"testing"
 
+	db "github.com/btcsuite/btcwallet/wallet/internal/db"
```
GustavoStingelin: nit: 
```suggestion
	"github.com/btcsuite/btcwallet/wallet/internal/db"
```

wallet/internal/db/pg/runtime.go
```diff
@@ -0,0 +1,149 @@
...
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"time"
+
+	dberr "github.com/btcsuite/btcwallet/wallet/internal/db/err"
+	dbruntime "github.com/btcsuite/btcwallet/wallet/internal/db/runtime"
+	sqlc "github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
```
GustavoStingelin: nit:
```suggestion
	"github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
```

wallet/internal/db/pg/runtime.go
```diff
@@ -0,0 +1,149 @@
...
+	return s.runtimeStats.CheckHealthy()
+}
+
+// ClassifyError normalizes one PostgreSQL backend error into the shared SQL
+// error model while preserving ordinary domain errors unchanged.
+func (s *Store) ClassifyError(err error) error {
+	normalized := dberr.Normalize(dberr.BackendPostgres, mapErr, err)
+	if normalized == nil {
+		return nil
+	}
```
GustavoStingelin: I guess these tests should pass, no? Otherwise we risk losing backend native SQL errors just because they are unknown.

```diff
diff --git a/wallet/internal/db/pg/errors_test.go b/wallet/internal/db/pg/errors_test.go
index 73ad1ac55..5c0058018 100644
--- a/wallet/internal/db/pg/errors_test.go
+++ b/wallet/internal/db/pg/errors_test.go
@@ -159,4 +159,22 @@ func TestClassifyError(t *testing.T) {
                require.ErrorAs(t, classifiedErr, &sqlErr)
                require.Equal(t, dberr.ReasonConstraint, sqlErr.Reason)
        })
+
+       t.Run("unknown backend error", func(t *testing.T) {
+               t.Parallel()
+
+               // Use an unmapped PostgreSQL SQLSTATE code that will normalize to
+               // ReasonUnknown. This tests the regression: unknown backend-native
+               // errors should remain wrapped as *dberr.SQLError with ReasonUnknown,
+               // not be unwrapped back to the raw driver error.
+               classifiedErr := store.ClassifyError(&pgconn.PgError{
+                       Code:    "99999", // unmapped SQLSTATE
+                       Message: "unknown error",
+               })
+
+               var sqlErr *dberr.SQLError
+               require.ErrorAs(t, classifiedErr, &sqlErr)
+               require.Equal(t, dberr.ReasonUnknown, sqlErr.Reason)
+               require.Equal(t, dberr.BackendPostgres, sqlErr.Backend)
+       })
 }
diff --git a/wallet/internal/db/sqlite/errors_test.go b/wallet/internal/db/sqlite/errors_test.go
index 3572c7a38..a8baf358a 100644
--- a/wallet/internal/db/sqlite/errors_test.go
+++ b/wallet/internal/db/sqlite/errors_test.go
@@ -150,4 +150,38 @@ func TestClassifyError(t *testing.T) {
                require.ErrorAs(t, classifiedErr, &sqlErr)
                require.Equal(t, dberr.ReasonConstraint, sqlErr.Reason)
        })
+
+       t.Run("unknown backend error", func(t *testing.T) {
+               t.Parallel()
+
+               // Test the regression: unknown backend-native errors should remain
+               // wrapped as *dberr.SQLError with ReasonUnknown, not be unwrapped
+               // back to the raw driver error. We trigger an unmapped error code
+               // by using an invalid SQL statement.
+               dbPath := filepath.Join(t.TempDir(), "wallet.db")
+               dbConn, err := sql.Open("sqlite", dbPath)
+               require.NoError(t, err)
+               t.Cleanup(func() {
+                       require.NoError(t, dbConn.Close())
+               })
+
+               ctx := t.Context()
+
+               // Create a table first
+               _, err = dbConn.ExecContext(
+                       ctx, `CREATE TABLE demo (id INTEGER PRIMARY KEY, val TEXT)`,
+               )
+               require.NoError(t, err)
+
+               // Trigger a syntax error which produces an unmapped error code
+               _, err = dbConn.ExecContext(ctx, `SELECT * FROM demo WHERE`)
+               require.Error(t, err)
+
+               classifiedErr := store.ClassifyError(err)
+
+               var sqlErr *dberr.SQLError
+               require.ErrorAs(t, classifiedErr, &sqlErr)
+               require.Equal(t, dberr.ReasonUnknown, sqlErr.Reason)
+               require.Equal(t, dberr.BackendSQLite, sqlErr.Backend)
+       })
 }
```

wallet/internal/db/sqlite/errors_test.go
```diff
@@ -87,3 +91,63 @@ func TestHelpers(t *testing.T) {
...
+
+	t.Run("existing sql error", func(t *testing.T) {
+		t.Parallel()
+
+		err := dberr.NewSQLError(
+			dberr.BackendSQLite,
+			dberr.ReasonConstraint,
+			"19",
+			errTestDup,
+		)
```
GustavoStingelin: nit:
```suggestion
		err := dberr.NewSQLError(
			dberr.BackendSQLite, dberr.ReasonConstraint, "19", errTestDup,
		)
```

wallet/internal/db/sqlite/runtime.go
```diff
@@ -0,0 +1,149 @@
...
+
+	_, err := dbruntime.Read(
+		ctx,
+		s,
+		s.queries,
+		defaultReadConfig(),
+		func(_ context.Context, q *sqlc.Queries) (struct{}, error) {
+			return struct{}{}, fn(q)
+		},
+	)
```
GustavoStingelin, line 45: nit:
```suggestion
	_, err := dbruntime.Read(
		ctx, s, s.queries, defaultReadConfig(),
		func(_ context.Context, q *sqlc.Queries) (struct{}, error) {
			return struct{}{}, fn(q)
		},
	)
```

wallet/internal/db/sqlite/runtime.go
```diff
@@ -0,0 +1,149 @@
...
+	_, err := dbruntime.Write(
+		ctx,
+		s,
+		func(tx *sql.Tx) *sqlc.Queries {
+			return s.queries.WithTx(tx)
+		},
+		func(qtx *sqlc.Queries) (struct{}, error) {
+			return struct{}{}, fn(qtx)
+		},
+	)
```
GustavoStingelin, line 63: nit:
```suggestion
	_, err := dbruntime.Write(
		ctx, s,
		func(tx *sql.Tx) *sqlc.Queries {
			return s.queries.WithTx(tx)
		},
		func(qtx *sqlc.Queries) (struct{}, error) {
			return struct{}{}, fn(qtx)
		},
	)
```

GustavoStingelin (changes requested):
Mostly nits. I left one comment that matters most, around handling unknown errors from db.

wallet/internal/db/itest/fixtures_pg_test.go
yyforyongyu: updated

wallet/internal/db/itest/fixtures_sqlite_test.go
yyforyongyu: fixed

wallet/internal/db/pg/errors_test.go
yyforyongyu: updated

wallet/internal/db/pg/runtime.go
yyforyongyu: updated

wallet/internal/db/sqlite/errors_test.go
yyforyongyu: updated

wallet/internal/db/sqlite/runtime.go
yyforyongyu: updated

wallet/internal/db/sqlite/runtime.go
yyforyongyu: updated

wallet/internal/db/pg/runtime.go
yyforyongyu: nice catch - now fixed!

wallet/internal/db/pg/accounts.go
yyforyongyu: That's actually a previous version, which didn't turn out to be very elegant, asked gpt to summarize it as it's been a while,
```
- we reintroduce generic result-returning backend helpers
- likely end up back at something like execReadResult / execWriteResult
- or generic methods/functions at the backend layer
- which is exactly the complexity we recently removed

Tradeoff:
- yes, outer vars are a bit less elegant at some call sites
- but the helper API is clearer and more consistent
```

yyforyongyu (commented):
LGTM ⚙️ **(`claude-opus-4-7`)**

yyforyongyu (commented):
LGTM ✅ (`gpt-5.3-codex`)

yyforyongyu:

yyforyongyu:
LGTM 🚀 (`minimax-m2.7`)

yyforyongyu (commented):
LGTM 🌟 (`gpt-5.4`)

GustavoStingelin (approved):
LGTM!

PR #1214, opened 2026-04-14
wallet: migrate address manager to db.Store

yyforyongyu:
## Summary
- extend the transitional `db.Store` surface with `db.AddressStore` and shared address metadata types
- add SQL/PG address-store create, get, list, type, and secret helpers, and implement the kvdb address-store adapter over legacy `waddrmgr`
- migrate the wallet address manager to create, import, get, and list addresses through `w.store` while keeping legacy fallbacks where kvdb store coverage is still transitional

## Why
- this moves address-manager callers toward the internal store abstraction without forcing the kvdb migration to implement unrelated UTXO support first
- it keeps `db.Store` persistence-focused and leaves derivation/encryption details in wallet/keyvault or legacy `waddrmgr` boundaries
- it preserves existing kvdb wallet behavior while giving SQL backends the same logical address metadata shape

## What Changed
- `db.AddressStore` now covers derived/imported address creation, address lookup/listing, address secrets, address types, and address iteration
- kvdb adapts the existing legacy address manager for derived/imported address creation, get, list, type metadata, pagination, and taproot script imports
- SQL and PostgreSQL return consistent `AddressInfo` account metadata for create/get/list paths
- `Wallet.ListAddresses` falls back to legacy UTXO balance aggregation when the transitional kvdb UTXO store returns `ErrNotImplemented`
- address tests cover create/get/list metadata consistency, kvdb pagination, watch-only semantics, public-key compression, and wallet fallback behavior

## Testing
- `GOWORK=off go test ./wallet ./wallet/internal/db/kvdb ./wallet/internal/db/sqlite ./wallet/internal/db/pg`
- `GOWORK=off go test ./waddrmgr ./wallet/internal/db`
- `make lint-check`
- `git diff --check impl-account-manager-store...HEAD`
- `gofmt -l $(git diff --name-only --diff-filter=ACMRT impl-account-manager-store...HEAD -- '*.go')`

CI is pending on the latest push.

### DIFF
--- a/wallet/address_manager.go
+++ b/wallet/address_manager.go
@@ -21,6 +21,9 @@ import (
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet/internal/addresstype"
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
 	"github.com/btcsuite/btcwallet/walletdb"
 )
 
@@ -49,11 +52,14 @@ var (
 	// ErrUnableToExtractAddress is returned when an address cannot be
 	// extracted from a pkscript.
 	ErrUnableToExtractAddress = errors.New("unable to extract address")
-
 	// errStopIteration is a special error used to stop the iteration in
 	// ForEachAccountAddress.
 	errStopIteration = errors.New("stop iteration")
 )
+
+// addressManagerPageLimit is the transitional address iteration page size.
+// TODO(yy): Make this configurable once the address store is fully wired.
+const addressManagerPageLimit = 500
 
 // AddressProperty represents an address and its balance.
 type AddressProperty struct {
@@ -127,9 +133,14 @@ type OutputScriptInfo struct {
 	// RedeemScript is the redeem script committed to by the outer P2SH output.
 	// For nested P2WPKH-in-P2SH spends, this is the inner witness program, for
 	// example `OP_0 <20-byte-key-hash>`. Native witness spends, such as P2WPKH
-	// and P2TR, leave this nil. The final scriptSig wrapper for nested witness
-	// spends can be rebuilt from this script when assembling the input.
+	// and P2TR, leave this nil.
 	RedeemScript []byte
+
+	// SigScript is the final scriptSig wrapper needed to spend outputs that are
+	// wrapped in P2SH.
+	// For nested P2WPKH-in-P2SH spends, this is a single push of RedeemScript.
+	// Native witness spends leave this nil.
+	SigScript []byte
 }
 
 // AddressManager provides an interface for generating and inspecting wallet
@@ -223,6 +234,77 @@ func addressInfoFromManagedAddress(
 	}
 
 	return info, nil
+}
+
+// addressPageRequest returns the standard page request used by address-manager
+// iteration helpers.
+func addressPageRequest() (page.Request[uint32], error) {
+	return page.NewRequest[uint32](addressManagerPageLimit)
+}
+
+// addressInfoFromStoreAddress converts one db-native address record into the
+// wallet-owned address metadata shape exposed by the public API.
+func addressInfoFromStoreAddress(storeAddr *db.AddressInfo,
+	chainParams *chaincfg.Params) (AddressInfo, error) {
+
+	addr := extractAddrFromPKScript(storeAddr.ScriptPubKey, chainParams)
+	if addr == nil {
+		return AddressInfo{}, fmt.Errorf("%w: from pkscript %x",
+			ErrUnableToExtractAddress, storeAddr.ScriptPubKey)
+	}
+
+	addrType, err := addresstype.ToWallet(
+		storeAddr.AddrType, storeAddr.HasScript,
+	)
+	if err != nil {
+		return AddressInfo{}, fmt.Errorf("%w: %v", ErrUnknownAddrType,
+			storeAddr.AddrType)
+	}
+
+	internal := storeAddr.Origin == db.DerivedAccount &&
+		storeAddr.Branch == 1
+
+	info := AddressInfo{
+		Addr:       addr,
+		AddrType:   addrType,
+		Imported:   storeAddr.Origin == db.ImportedAccount,
+		Internal:   internal,
+		Compressed: storeAddressPubKeyCompressed(storeAddr.PubKey),
+	}
+
+	if len(storeAddr.PubKey) == 0 {
+		return info, nil
+	}
+
+	pubKey, err := btcec.ParsePubKey(storeAddr.PubKey)
+	if err != nil {
+		return AddressInfo{}, fmt.Errorf("parse pubkey: %w", err)
+	}
+
+	info.PubKey = pubKey
+
+	if info.Imported {
+		return info, nil
+	}
+
+	info.Derivation = &AddressDerivation{
+		KeyScope: waddrmgr.KeyScope{
+			Purpose: storeAddr.KeyScope.Purpose,
+			Coin:    storeAddr.KeyScope.Coin,
+		},
+		Account:              storeAddr.AccountNumber,
+		Branch:               storeAddr.Branch,
+		Index:                storeAddr.Index,
+		MasterKeyFingerprint: storeAddr.MasterKeyFingerprint,
+	}
+
+	return info, nil
+}
+
+// storeAddressPubKeyCompressed reports whether store pubkey bytes use the
+// compressed secp256k1 encoding.
+func storeAddressPubKeyCompressed(pubKey []byte) bool {
+	return len(pubKey) == btcec.PubKeyBytesLenCompressed
 }
 
 // NewAddress returns a new address for the given account and address type.
@@ -660,7 +742,7 @@ func (w *Wallet) ImportPublicKey(_ context.Context, pubKey *btcec.PublicKey,
 
 	manager, err := w.addrStore.FetchScopedKeyManager(keyScope)
 	if err != nil {
-		return err
+		return fmt.Errorf("%w: %v", ErrUnknownAddrType, addrType)
 	}
 
 	var addr btcutil.Address
@@ -898,7 +980,7 @@ func derivationForAddressInfo(addressInfo AddressInfo) (
 
 	keyScope := addressInfo.Derivation.KeyScope
 
-	return &psbt.Bip32Derivation{
+	derivationInfo := &psbt.Bip32Derivation{
 		PubKey:               addressInfo.PubKey.SerializeCompressed(),
 		MasterKeyFingerprint: addressInfo.Derivation.MasterKeyFingerprint,
 		Bip32Path: []uint32{
@@ -908,5 +990,7 @@ func derivationForAddressInfo(addressInfo AddressInfo) (
 			addressInfo.Derivation.Branch,
 			addressInfo.Derivation.Index,
 		},
-	}, nil
+	}
+
+	return derivationInfo, nil
 }
--- a/wallet/address_manager.go
+++ b/wallet/address_manager.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 
 	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/btcutil/psbt"
@@ -24,6 +25,7 @@ import (
 	"github.com/btcsuite/btcwallet/wallet/internal/addresstype"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
+	"github.com/btcsuite/btcwallet/wallet/internal/keyvault"
 	"github.com/btcsuite/btcwallet/walletdb"
 )
 
@@ -671,35 +673,7 @@ func (w *Wallet) ListAddresses(_ context.Context, accountName string,
 }
 
 // ImportPublicKey imports a single public key as a watch-only address.
-//
-// This method allows the wallet to track transactions related to a specific
-// public key without having access to the corresponding private key. This is
-// useful for monitoring addresses without compromising their security.
-//
-// How it works:
-// The method determines the appropriate key scope based on the provided
-// address type and then uses the corresponding scoped key manager to import
-// the public key.
-//
-// Logical Steps:
-//  1. Determine the key scope from the address type (e.g., P2WKH, NP2WKH).
-//  2. Fetch the scoped key manager for that scope.
-//  3. Initiate a database transaction.
-//  4. Within the transaction, call the underlying address manager's
-//     ImportPublicKey method to store the key.
-//  5. Commit the transaction.
-//
-// Database Actions:
-//   - This method performs a single database write transaction
-//     (`walletdb.Update`).
-//   - It stores the public key and its associated address information within
-//     the `waddrmgr` namespace.
-//
-// Time Complexity:
-//   - The operation is dominated by the database write, making its complexity
-//     roughly O(1) or O(log N) depending on the database backend's indexing
-//     strategy for keys. It is generally a fast operation.
-func (w *Wallet) ImportPublicKey(_ context.Context, pubKey *btcec.PublicKey,
+func (w *Wallet) ImportPublicKey(ctx context.Context, pubKey *btcec.PublicKey,
 	addrType waddrmgr.AddressType) error {
 
 	err := w.state.validateStarted()
@@ -712,25 +686,34 @@ func (w *Wallet) ImportPublicKey(_ context.Context, pubKey *btcec.PublicKey,
 		return fmt.Errorf("%w: %v", ErrUnknownAddrType, addrType)
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(keyScope)
+	storeAddrType, err := addresstype.FromWallet(addrType)
 	if err != nil {
 		return fmt.Errorf("%w: %v", ErrUnknownAddrType, addrType)
 	}
 
-	var addr btcutil.Address
+	serializedPubKey := pubKey.SerializeCompressed()
 
-	err = walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
-		addrmgrNs := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+	addr, err := addrType.AddrFromPubKeyBytes(
+		serializedPubKey, w.cfg.ChainParams,
+	)
+	if err != nil {
+		return fmt.Errorf("derive imported address: %w", err)
+	}
 
-		ma, err := manager.ImportPublicKey(addrmgrNs, pubKey, nil)
-		if err != nil {
-			return err
-		}
+	scriptPubKey, err := txscript.PayToAddrScript(addr)
+	if err != nil {
+		return fmt.Errorf("pay to addr script: %w", err)
+	}
 
-		addr = ma.Address()
-
-		return nil
-	})
+	_, err = w.store.NewImportedAddress(
+		ctx, db.NewImportedAddressParams{
+			WalletID:     w.id,
+			Scope:        db.KeyScope(keyScope),
+			AddressType:  storeAddrType.Type,
+			ScriptPubKey: scriptPubKey,
+			PubKey:       serializedPubKey,
+		},
+	)
 	if err != nil {
 		return err
 	}
@@ -739,33 +722,7 @@ func (w *Wallet) ImportPublicKey(_ context.Context, pubKey *btcec.PublicKey,
 }
 
 // ImportTaprootScript imports a taproot script for tracking and spending.
-//
-// This method allows the wallet to import a taproot script, which is
-// necessary for spending from or tracking a taproot address.
-//
-// How it works:
-// The method uses the BIP-0086 key scope to fetch the taproot-specific
-// scoped key manager. It then calls the underlying manager's
-// ImportTaprootScript method to store the script information.
-//
-// Logical Steps:
-//  1. Fetch the scoped key manager for the taproot key scope (BIP-0086).
-//  2. Initiate a database transaction.
-//  3. Within the transaction, get the wallet's current sync state to use as
-//     the "birthday" for the new script.
-//  4. Call the underlying address manager's ImportTaprootScript method.
-//  5. Commit the transaction.
-//
-// Database Actions:
-//   - This method performs a single database write transaction
-//     (`walletdb.Update`).
-//   - It stores the taproot script and its derived address information within
-//     the `waddrmgr` namespace.
-//
-// Time Complexity:
-//   - Similar to ImportPublicKey, this operation is dominated by a database
-//     write, making it a fast operation with a complexity of roughly O(1).
-func (w *Wallet) ImportTaprootScript(_ context.Context,
+func (w *Wallet) ImportTaprootScript(ctx context.Context,
 	tapscript waddrmgr.Tapscript) (AddressInfo, error) {
 
 	err := w.state.validateStarted()
@@ -773,34 +730,76 @@ func (w *Wallet) ImportTaprootScript(_ context.Context,
 		return AddressInfo{}, err
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(
-		waddrmgr.KeyScopeBIP0086,
+	taprootKey, err := tapscript.TaprootKey()
+	if err != nil {
+		return AddressInfo{}, err
+	}
+
+	addr, err := btcutil.NewAddressTaproot(
+		schnorr.SerializePubKey(taprootKey), w.cfg.ChainParams,
+	)
+	if err != nil {
+		return AddressInfo{}, fmt.Errorf("taproot address: %w", err)
+	}
+
+	scriptPubKey, err := txscript.PayToAddrScript(addr)
+	if err != nil {
+		return AddressInfo{}, fmt.Errorf("pay to addr script: %w", err)
+	}
+
+	encryptedScript, err := encryptTaprootScript(w.keyVault, &tapscript)
+	if err != nil {
+		return AddressInfo{}, err
+	}
+
+	storeInfo, err := w.store.NewImportedAddress(
+		ctx, db.NewImportedAddressParams{
+			WalletID:        w.id,
+			Scope:           db.KeyScope(waddrmgr.KeyScopeBIP0086),
+			AddressType:     db.TaprootPubKey,
+			ScriptPubKey:    scriptPubKey,
+			EncryptedScript: encryptedScript,
+		},
 	)
 	if err != nil {
 		return AddressInfo{}, err
 	}
 
-	var addr waddrmgr.ManagedAddress
+	storeInfo.HasScript = true
 
-	err = walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
-		ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
-		syncedTo := w.addrStore.SyncedTo()
-		addr, err = manager.ImportTaprootScript(
-			ns, &tapscript, &syncedTo, 1, false,
-		)
-
-		return err
-	})
+	info, err := addressInfoFromStoreAddress(storeInfo, w.cfg.ChainParams)
 	if err != nil {
 		return AddressInfo{}, err
 	}
 
-	err = w.cfg.Chain.NotifyReceived([]btcutil.Address{addr.Address()})
+	err = w.cfg.Chain.NotifyReceived([]btcutil.Address{addr})
 	if err != nil {
 		return AddressInfo{}, err
 	}
 
-	return addressInfoFromManagedAddress(addr)
+	return info, nil
+}
+
+// encryptTaprootScript encodes and encrypts taproot script data before the
+// encrypted blob is handed to the store.
+func encryptTaprootScript(vault keyvault.Vault,
+	tapscript *waddrmgr.Tapscript) ([]byte, error) {
+
+	encodedScript, err := waddrmgr.EncodeTaprootScript(tapscript)
+	if err != nil {
+		return nil, fmt.Errorf("encode tapscript: %w", err)
+	}
+
+	if vault == nil {
+		return nil, fmt.Errorf("%w: keyVault", ErrMissingParam)
+	}
+
+	encryptedScript, err := vault.Encrypt(waddrmgr.CKTPublic, encodedScript)
+	if err != nil {
+		return nil, fmt.Errorf("encrypt tapscript: %w", err)
+	}
+
+	return encryptedScript, nil
 }
 
 // ScriptForOutput returns the address metadata and spending scripts for a given
--- a/wallet/address_manager.go
+++ b/wallet/address_manager.go
@@ -52,9 +52,6 @@ var (
 	// ErrUnableToExtractAddress is returned when an address cannot be
 	// extracted from a pkscript.
 	ErrUnableToExtractAddress = errors.New("unable to extract address")
-	// errStopIteration is a special error used to stop the iteration in
-	// ForEachAccountAddress.
-	errStopIteration = errors.New("stop iteration")
 )
 
 // addressManagerPageLimit is the transitional address iteration page size.
@@ -363,7 +360,7 @@ func storeAddressPubKeyCompressed(pubKey []byte) bool {
 //     transaction to persist the new address.
 //     This ensures that we only save an address after we are confident that
 //     it is being watched by the backend, preventing fund loss.
-func (w *Wallet) NewAddress(_ context.Context, accountName string,
+func (w *Wallet) NewAddress(ctx context.Context, accountName string,
 	addrType waddrmgr.AddressType, change bool) (btcutil.Address, error) {
 
 	err := w.state.validateStarted()
@@ -381,48 +378,26 @@ func (w *Wallet) NewAddress(_ context.Context, accountName string,
 		return nil, fmt.Errorf("%w: %v", ErrUnknownAddrType, addrType)
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(keyScope)
+	addrInfo, err := w.store.NewDerivedAddress(
+		ctx, db.NewDerivedAddressParams{
+			WalletID:    w.id,
+			AccountName: accountName,
+			Scope:       db.KeyScope(keyScope),
+			Change:      change,
+		},
+	)
 	if err != nil {
 		return nil, err
 	}
 
-	addr, err := w.newAddress(manager, accountName, change)
-	if err != nil {
-		return nil, err
+	addr := extractAddrFromPKScript(addrInfo.ScriptPubKey, w.cfg.ChainParams)
+	if addr == nil {
+		return nil, fmt.Errorf("%w: from pkscript %x",
+			ErrUnableToExtractAddress, addrInfo.ScriptPubKey)
 	}
 
 	// Notify the rpc server about the newly created address.
 	err = w.cfg.Chain.NotifyReceived([]btcutil.Address{addr})
-	if err != nil {
-		return nil, err
-	}
-
-	return addr, nil
-}
-
-// newAddress returns the next external chained address for a wallet. It
-// wraps the database transaction and the call to the scoped key manager's
-// NewAddress method. The underlying address manager handles its own
-// synchronization to ensure that in-memory state remains consistent with the
-// database, preventing race conditions during address creation.
-func (w *Wallet) newAddress(manager waddrmgr.AccountStore,
-	accountName string, change bool) (btcutil.Address, error) {
-
-	var (
-		addr btcutil.Address
-		err  error
-	)
-
-	err = walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
-		addrmgrNs := tx.ReadWriteBucket(waddrmgrNamespaceKey)
-
-		addr, err = manager.NewAddress(addrmgrNs, accountName, change)
-		if err != nil {
-			return err
-		}
-
-		return nil
-	})
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


yyforyongyu:

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -0,0 +1,723 @@
+package kvdb
```
yyforyongyu, line 1: Missing copyright header. All new Go files in this repo should include the standard license header before the package declaration.

`k2p6`

wallet/internal/db/kvdb/addressstore_test.go
```diff
@@ -0,0 +1,144 @@
+package kvdb
```
yyforyongyu, line 5: Missing copyright header. All new Go files in this repo should include the standard license header before the package declaration.

`k2p6`

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -0,0 +1,723 @@
...
+	)
+	if err != nil || len(addrs) == 0 {
+		return nil
+	}
+
+	return addrs[0]
+}
+
+// kvdbAddressID returns a deterministic synthetic address ID for legacy rows.
+func kvdbAddressID(scriptPubKey []byte) uint32 {
```
yyforyongyu: `kvdbAddressID` returns an FNV-32a hash of `scriptPubKey`; with ~65K addresses in one account the birthday bound gives ~50% chance of collision, and collisions break cursor pagination: `kvdbAddressPageItems` skips items where `items[start].ID <= *query.Page.After`, so if the last item of a page collides with the next item (the `+1` lookahead used by `page.BuildResult`), the next page silently loses that address. Switch to a 64-bit ID (or have kvdb pick a true sequential ID stored against the address), and drop the `if id == 0 { return 1 }` mapping which itself creates a collision between hash=0 and hash=1.

`claude-opus-4-7[1m]`

wallet/address_manager.go
```diff
@@ -208,23 +226,267 @@ func addressInfoFromManagedAddress(
...
+
+	internal := storeAddr.Origin == db.DerivedAccount &&
+		storeAddr.Branch == 1
+
+	info := AddressInfo{
+		Addr:       addr,
+		AddrType:   addrType,
+		Imported:   storeAddr.Origin == db.ImportedAccount,
+		Internal:   internal,
+		Compressed: len(storeAddr.PubKey) > 0,
```
yyforyongyu: `Compressed: len(storeAddr.PubKey) > 0` conflates "has a pubkey" with "is compressed"; the legacy `addressInfoFromManagedAddress` at line 220 correctly uses `managedAddr.Compressed()`. In practice the kvdb adapter always serializes via `SerializeCompressed()` (`internal/db/kvdb/addressstore.go:633`) so the result is incidentally correct today, but if anyone ever stores a 65-byte uncompressed key the wallet API will silently flip its Compressed flag wrong. Compute Compressed from the actual byte length (33 vs 65) or parse the key.

`claude-opus-4-7[1m]`

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -0,0 +1,715 @@
...
+		KeyScope:             keyScope,
+		MasterKeyFingerprint: fingerprint,
+		AddrType:             addrType,
+		Origin:               origin,
+		Branch:               branch,
+		Index:                index,
+		ScriptPubKey:         scriptPubKey,
+		PubKey:               pubKey,
+		HasScript:            hasScript,
+		IsWatchOnly:          managedAddr.Imported(),
```
yyforyongyu: `IsWatchOnly: managedAddr.Imported()` is a different semantic than the SQL backend uses: `addresses_common.go:440` reads `origin == ImportedAccount && !row.HasPrivateKey`, i.e. watch-only only when the import had no private key. `managedAddress.Imported()` is true for any imported address, including those imported with a private key. Existing legacy data (pre-migration) can contain private-key imports — kvdb will report them as watch-only here while SQL would not, and `AddressInfo.IsWatchOnly` is documented as "indicates whether the wallet has the private key for this address". Surface the private-key presence (e.g. via the legacy `manager`-side accessor) rather than collapsing it into `Imported()`.

`claude-opus-4-7[1m]`

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -0,0 +1,715 @@
...
+	ImportWitnessScript(ns walletdb.ReadWriteBucket, script []byte,
+		bs *waddrmgr.BlockStamp, witnessVersion byte,
+		isSecretScript bool) (waddrmgr.ManagedScriptAddress, error)
+}
+
+// NewDerivedAddress creates one derived address through the legacy address-
+// manager path.
+func (s *Store) NewDerivedAddress(ctx context.Context,
+	params db.NewDerivedAddressParams,
+	_ db.AddressDerivationFunc) (*db.AddressInfo, error) {
```
yyforyongyu: `_ db.AddressDerivationFunc` discards the wallet-provided derivation callback, while the interface contract on `db.AddressStore.NewDerivedAddress` says "calls deriveFn to derive the actual address data" — kvdb instead delegates to `manager.NextExternal/InternalAddresses`. The callback is fundamental to the SQL backend, and a wallet passing a custom `deriveFn` (e.g., for an experimental scheme) will silently see kvdb produce a different address than SQL would. Either invoke the callback after `manager.NextExternal/InternalAddresses` allocates the index and use its result, or document the divergence with a `// NOTE: kvdb ignores deriveFn …` so the contract violation is explicit.

`claude-opus-4-7[1m]`

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -0,0 +1,715 @@
...
+	if !ok {
+		return waddrmgr.BlockStamp{}
+	}
+
+	return syncState.SyncedTo()
+}
+
+// kvdbListAddressItems loads, filters, and paginates account addresses from one
+// legacy scoped manager.
+func kvdbListAddressItems(dbConn walletdb.DB,
```
yyforyongyu: `kvdbListAddressItems` loads every address for the account before applying the cursor: `manager.ForEachAccountAddress(ns, account, ...)` enumerates the full set, then `kvdbAddressPageItems` trims by cursor. For a wallet with 100K addresses each page request is O(N) on time, memory, and the address-manager mutex (`scoped_manager.go:2732` holds `mtx.Lock` for the whole iteration). The SQL backend can serve `WHERE id > cursor LIMIT N` directly — worth a TODO that kvdb pagination is intentionally inefficient and only viable while the legacy bucket count is small.

`claude-opus-4-7[1m]`

wallet/address_manager.go
```diff
@@ -400,252 +647,140 @@ func (w *Wallet) GetUnusedAddress(ctx context.Context, accountName string,
...
-		txmgrNs := tx.ReadBucket(wtxmgrNamespaceKey)
+	req, err := addressPageRequest()
+	if err != nil {
+		return nil, err
+	}
 
-		// First, we'll create a map of address to balance by iterating
-		// through all the unspent outputs.
-		addrToBalance := make(map[string]btcutil.Amount)
+	utxos, err := w.store.ListUTXOs(ctx, db.ListUtxosQuery{WalletID: w.id})
```
yyforyongyu: `ListAddresses` now calls `w.store.ListUTXOs`, but the default wallet still wires `w.store` to `kvdb.NewStore` (`wallet/manager.go:304`) and kvdb's `ListUTXOs` still returns `kvdb.Store.ListUTXOs: not implemented` (`wallet/internal/db/kvdb/utxostore.go:38`). That means every legacy/kvdb wallet call to the public `Wallet.ListAddresses` API now fails before address iteration. Either keep the legacy `w.txStore.UnspentOutputs` path for kvdb, implement `kvdb.ListUTXOs`, or add a store helper that both backends support before routing this API through `w.store`.

(`openai/gpt-5.5`)

wallet/internal/db/addresses_common.go
```diff
@@ -503,6 +611,7 @@ func createDerivedAddress[T any](ctx context.Context,
 		Branch:       branch,
 		Index:        index,
 		ScriptPubKey: scriptPubKey,
+		PubKey:       pubKey,
```
yyforyongyu, line 745: `AddressRowToInfo` now fills `AccountName`, `KeyScope`, `AccountNumber`, and `MasterKeyFingerprint` for list/get results, but `NewDerivedAddress` still returns the freshly-created `AddressInfo` from this literal without those fields. The result is an inconsistent store contract and the SQLite address itests still fail because `createDerivedAddresses` uses `NewDerivedAddress` as expected data while `ListAddresses` returns populated account metadata (`GOWORK=off go test -tags=itest ./wallet/internal/db/itest -run 'TestListAddresses(Pagination|ExactBoundary|DeterministicPagination|AccountIsolation|$)|TestIterAddresses'`). Populate the same account metadata here (and imported creation should probably do the same) or normalize the expected values.

(`openai/gpt-5.5`)

yyforyongyu:
Pushed the address-manager follow-ups in `d9eeda3ff`.

Addressed review items:
- kept public `Wallet.ListAddresses` working for kvdb by falling back to the legacy UTXO path while `kvdb.Store.ListUTXOs` is transitional;
- made create/get/list address metadata consistent, including account name, scope, account number, and master fingerprint;
- changed kvdb address pagination to one-based sorted ordinal IDs instead of script hashes;
- documented and validated the kvdb `deriveFn` divergence by using it as post-create validation;
- fixed imported watch-only detection and public-key compression handling;
- documented the O(N) kvdb pagination adapter and postponed SQL N+1 optimization with comments;
- cleaned up error wrapping, helper names, and declaration wrapping.

Local verification:
- `GOWORK=off go test ./wallet ./wallet/internal/db/kvdb ./wallet/internal/db/sqlite ./wallet/internal/db/pg`
- `GOWORK=off go test ./waddrmgr ./wallet/internal/db`
- `make lint-check`
- `git diff --check impl-account-manager-store...HEAD`
- `gofmt -l $(git diff --name-only --diff-filter=ACMRT impl-account-manager-store...HEAD -- '*.go')`

CI is pending on the pushed branch.

yyforyongyu:

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -55,16 +358,540 @@ func (s *Store) GetAddressSecret(ctx context.Context,
...
+	if managedAddr.Compressed() {
+		return managedAddr.PubKey().SerializeCompressed()
+	}
+
+	return managedAddr.PubKey().SerializeUncompressed()
+}
+
+// managedAddressIsWatchOnly reports whether a legacy managed address lacks
+// private key material.
+func managedAddressIsWatchOnly(
```
yyforyongyu: `managedAddressIsWatchOnly` only inspects the address — for any non-imported (derived) address it returns `false`, regardless of wallet state. But `AddressInfo.IsWatchOnly` is documented as "the wallet has the private key for this address" (`data_types.go:640`), and the SQL backend folds in wallet-level watch-only at `addresses_common.go:440` (`isWatchOnly := row.WalletIsWatchOnly; if origin == ImportedAccount { … }`). For a watch-only wallet, kvdb returns `IsWatchOnly: false` for every derived address while SQL returns `true` — same wallet, different backend, contradictory flag. Either thread wallet-level `WatchOnly()` through `legacyAddrStore` and OR it in here, or capture it at `NewStore` time and read it back from the store, so the kvdb adapter matches the documented semantic.

`claude-opus-4-7[1m]`

wallet/address_manager.go
```diff
@@ -299,14 +621,29 @@ func (w *Wallet) NewAddress(_ context.Context, accountName string,
 		return nil, fmt.Errorf("%w: %v", ErrUnknownAddrType, addrType)
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(keyScope)
+	addrInfo, err := w.store.NewDerivedAddress(
```
yyforyongyu, line 383: `NewAddress` now persists derived addresses through `w.store.NewDerivedAddress`, but the SQL implementations only insert the new row into the SQL address store. The signer still resolves the private key by looking the output address up in legacy `waddrmgr` (`privKeyForOutput` -> `loadManagedPubKeyAddr` -> `w.addrStore.Address`). For a SQLite/Postgres-backed wallet, a UTXO sent to one of these newly generated store-only addresses can be recognized by `ScriptForOutput`/`GetAddressInfo`, but signing fails because there is no matching legacy address row to load. Please either keep the legacy manager in sync when creating derived SQL addresses, or make signing derive from the `AddressInfo.Derivation` metadata instead of requiring `waddrmgr.Address` to find the address. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/addressstore.go
yyforyongyu: Fixed and pushed. The kvdb adapter now threads wallet-level `WatchOnly()` through derived/imported/get/list metadata and ORs it into `managedAddressIsWatchOnly`. Added `TestAddressStoreNewDerivedAddressWatchOnlyWallet` to cover create/get/list for a watch-only wallet.

wallet/address_manager.go
yyforyongyu: Fixed and pushed. `DBDeriveAddressData` now extends the legacy waddrmgr address rows in a walletdb write transaction before deriving the script, so signer lookups can resolve SQL-derived addresses during the transition. Added `TestDBDeriveAddressDataExtendsLegacyAddressStore`.

yyforyongyu:

yyforyongyu:
LGTM 🏠 (v4)

`managedAddressIsWatchOnly` now folds wallet-level watch-only into the per-address flag (`internal/db/kvdb/addressstore.go:881-901`): `if walletIsWatchOnly { return true }` short-circuits the per-address inspection, then the imported / no-priv-key checks run for spendable wallets. The flag is threaded through `listAddressItems`, `setAddressID`, and the derived/imported create paths via `addrMgr.WatchOnly()` and a new `WatchOnly() bool` method on `legacyAddrStore`. Closes the backend-divergent `IsWatchOnly` flag I flagged on v3.

`TestListAddressesWatchOnlyWallet` (`itest/address_store_test.go:1551`) pins the contract by creating derived addresses inside a watch-only wallet and asserting `addrs[0].IsWatchOnly == true`. The other prior findings (FNV-32 collision → one-based ordinal IDs, `len > 0` → `len == PubKeyBytesLenCompressed`, `deriveFn` ignored → validated post-create with explanatory comment, O(N) pagination → documented as transitional) all remain addressed.
(`claude-opus-4-7[1m]`)

yyforyongyu (commented):
LGTM from my current pass. I rechecked the refreshed head and did not find any new actionable issues. Focused `git diff --check` and `GOWORK=off go test ./waddrmgr ./wallet ./wallet/internal/db ./wallet/internal/db/kvdb ./wallet/internal/db/sqlite ./wallet/internal/db/pg` passed. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🏠 (v5)

The PR 1214-specific commits are unchanged from v4 — the diff vs v4 is entirely the rebased base (PR #1231 v6's `deriveChildKey` runtime switch + #1215 v7's Balance scope filter all flow through). Build + vet pass clean. My v4 LGTM stands.
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -1,51 +1,389 @@
...
+			ns, manager, addrMgr.WatchOnly(), params,
+		)
+
+		return err
+	})
+	if err != nil {
+		return nil, fmt.Errorf("NewDerivedAddress: %w", err)
+	}
+
+	err = validateDerivedAddress(ctx, deriveFn, info)
```
yyforyongyu: `validateDerivedAddress` runs after the walletdb update commits. If the caller-provided `deriveFn` returns an error, context cancellation, nil data, or a mismatch, `NewDerivedAddress` returns an error even though the legacy manager already advanced the index and persisted the address. That makes retries skip an address and leaves durable state changed after a failed call. Please perform the callback validation before committing the mutation, or restructure the kvdb path so validation cannot fail after the address/index has been persisted. (`openai/gpt-5.5`)

wallet/address_manager.go
```diff
@@ -400,252 +707,143 @@ func (w *Wallet) GetUnusedAddress(ctx context.Context, accountName string,
...
-	manager, err := w.addrStore.FetchScopedKeyManager(keyScope)
+	req, err := addressPageRequest()
 	if err != nil {
 		return nil, err
 	}
 
-	unusedAddr, err := w.findUnusedAddress(manager, accountName, change)
-	// We'll ignore the special error that we use to stop the iteration.
-	if err != nil && !errors.Is(err, errStopIteration) {
+	usedScripts, err := w.usedWalletAddressScripts(ctx)
```
yyforyongyu: The new unused-address scan only reconstructs usage from current transaction history. The old path also consulted waddrmgr’s monotonic `Used` flag, which is set by `MarkUsed` and can remain true even if the transaction that originally used the address is later rolled back/removed/abandoned. With this implementation such an address drops out of `usedScripts` and can be returned again by `GetUnusedAddress`, causing address reuse. Please carry an `IsUsed`/used flag through the store address metadata or query the legacy/store used metadata here instead of relying solely on current tx details. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/addressstore.go
yyforyongyu: Fixed in `b6522b1d4..a1f5bf268` on the rebased
`impl-address-manager-store` branch. `validateDerivedAddress` is
now invoked inside the `walletdb.Update` callback before the
caller's transaction commits, so a failed validation rolls the
index advance back. Regression test in `kvdb: test derived
address creation` (`11609c4cc`).

wallet/address_manager.go
yyforyongyu: Acknowledged + fixed in the new `addresses-used-flag` PR
(https://github.com/btcsuite/btcwallet/pull/new/addresses-used-flag).
The `addresses.used` column lands on `prep-address-manager-store`
with `IsUsed` threaded through every backend's AddressInfo. The
unused-address scan now reads `addr.IsUsed` directly (no
tx-history reconstruction), and `MarkUsed` writes route through
`w.store.MarkAddressUsed` so the monotonic flag survives reorgs
on both kvdb and SQL backends.

wallet/address_manager.go
```diff
@@ -299,14 +621,29 @@ func (w *Wallet) NewAddress(_ context.Context, accountName string,
...
+		ctx, db.NewDerivedAddressParams{
+			WalletID:    w.id,
+			AccountName: accountName,
+			Scope:       db.KeyScope(keyScope),
+			Change:      change,
+		},
+		func(ctx context.Context, _ uint32, branch, index uint32) (
+			*db.DerivedAddressData, error) {
+
+			return w.DBDeriveAddressData(
```
yyforyongyu: **Production deadlock on bbolt-backed wallets.** The closure passed as `deriveFn` is `w.DBDeriveAddressData`, which opens its own `walletdb.Update(w.cfg.DB, ...)` (`wallet/db_ops.go:137`). `kvdb.Store.NewDerivedAddress` calls this `deriveFn` from inside its own `walletdb.Update` on the same DB handle (see `wallet/internal/db/kvdb/addressstore.go:89-105`, where `validateDerivedAddress` is the last statement of the outer `Update` callback).

`Manager.Load` wires `w.store = kvdb.NewStore(cfg.DB, ...)`, so in production `w.store` is the kvdb store and the two `walletdb.Update` calls run on the same `*bbolt.DB`. bbolt serializes writers with a non-reentrant rwlock; nested `Update` from the same goroutine deadlocks.

Why CI is green:
- `wallet/address_manager_test.go` uses `createStartedWalletWithMocks`, whose `mockStore.NewDerivedAddress` doesn't invoke `deriveFn`.
- `wallet/internal/db/kvdb/addressstore_test.go`'s `testAddressDerivationFunc` uses `walletdb.View` instead of `walletdb.Update`.
- Neither path exercises the (production wallet) × (kvdb store) × (`DBDeriveAddressData` deriveFn) combination.

Two ways to fix:
1. Rewrite `DBDeriveAddressData` to use `walletdb.View` and drop the `manager.ExtendAddresses` call (kvdb's `derivedAddressInfo` has already extended via `NextExternalAddresses` / `NextInternalAddresses` before `validateDerivedAddress` runs, so re-extending here is redundant anyway).
2. Make `kvdb.NewDerivedAddress` call `validateDerivedAddress` *after* the outer `Update` commits — but then the rollback-on-validation-failure guarantee in the existing comment goes away, so option 1 is cleaner.

A test that wires the real kvdb store (not a mock) into `Wallet.NewAddress` would have caught this; worth adding alongside the fix.

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -75,3 +400,487 @@ func (s *Store) MarkAddressUsed(ctx context.Context,
...
+			branch = path.Branch
+			index = path.Index
+			fingerprint = path.MasterKeyFingerprint
+			keyScope = db.KeyScope(scope)
+		}
+	}
+
+	_, hasScript := managedAddr.(waddrmgr.ManagedScriptAddress)
+
+	return &db.AddressInfo{
```
yyforyongyu, line 630: `AddressInfo.IsUsed` is never populated from the legacy `ManagedAddress.Used(ns)` state. For the kvdb store, an address marked used by `waddrmgr.MarkUsed` is returned from `GetAddress`/`ListAddresses` with the zero value `false`, so any store-based unused-address scan or caller relying on the AddressStore used-flag contract can treat it as reusable after the transaction that first used it is rolled back or removed. Please set `IsUsed` from the legacy used flag when constructing the store address info. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -1,19 +1,139 @@
...
-	_ db.AddressDerivationFunc) (*db.AddressInfo, error) {
+	params db.NewDerivedAddressParams,
+	deriveFn db.AddressDerivationFunc) (*db.AddressInfo, error) {
+
+	err := ctx.Err()
+	if err != nil {
+		return nil, err
+	}
+
+	if s.addrStore == nil {
```
yyforyongyu: remove all the checks on this `s.addrStore` - this is unnecessary, we should just let it panic if we ever run into this issue as it's rather stupid to not even initialize the `addrStore`.

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -458,9 +537,9 @@ func syncedTo(addrStore any) waddrmgr.BlockStamp {
 	return syncState.SyncedTo()
 }
 
-// kvdbSetAddressID assigns target's synthetic ID from the current legacy
-// account view.
-func kvdbSetAddressID(ns walletdb.ReadBucket, manager waddrmgr.AccountStore,
+// setAddressID assigns target's synthetic ID from the current legacy account
+// view.
+func setAddressID(ns walletdb.ReadBucket, manager waddrmgr.AccountStore,
```
yyforyongyu, line 478: this change should happen in the commit where we introduce `kvdbSetAddressID` so we don't need to create this noisy change again here

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -47,36 +50,6 @@ var (
...
-
-// legacySyncState is the subset of the legacy root manager needed to import
-// script addresses with a block stamp.
-type legacySyncState interface {
-	SyncedTo() waddrmgr.BlockStamp
-}
-
-// legacyAddressDecryptor is implemented by legacy address managers that can
-// decrypt imported script material.
-type legacyAddressDecryptor interface {
```
yyforyongyu: ok again, this should be folded into previous commits - we are creating too much noise here.

yyforyongyu:
LGTM 🧭
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -100,3 +225,290 @@ var addressTypes = []db.AddressTypeInfo{
...
+
+	if walletIsWatchOnly {
+		return true
+	}
+
+	return managedAddr.Imported()
+}
+
+// storeAddressType maps legacy address types to the store enum.
+func storeAddressType(addrType waddrmgr.AddressType) (db.AddressType,
```
yyforyongyu: I think this is already defined in address_manager.go? I think we can move the definition into the `db` pkg, and use it in the address manager instead?

wallet/address_manager.go
```diff
@@ -299,48 +529,82 @@ func (w *Wallet) NewAddress(_ context.Context, accountName string,
...
-		return nil, err
-	}
+			var derived *db.DerivedAddressData
 
-	return addr, nil
-}
+			err = walletdb.View(w.cfg.DB, func(tx walletdb.ReadTx) error {
+				ns := tx.ReadBucket(waddrmgrNamespaceKey)
+				path := waddrmgr.DerivationPath{
+					InternalAccount: accountID,
```
yyforyongyu: Using `accountID` as `InternalAccount` derives by the SQL account row ID rather than the BIP44 account number, so pass the store `AccountNumber` into the callback or look it up before building the path. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -55,16 +349,531 @@ func (s *Store) GetAddressSecret(ctx context.Context,
...
+		return nil, fmt.Errorf("kvdb imported raw script pubkey: %w",
+			errNotImplemented)
+	}
+
+	pubKey, err := btcec.ParsePubKey(params.PubKey)
+	if err != nil {
+		return nil, fmt.Errorf("parse imported pubkey: %w", err)
+	}
+
+	managedAddr, err := manager.ImportPublicKey(ns, pubKey, nil)
```
yyforyongyu, line 336: This import path ignores `params.AddressType` and `params.ScriptPubKey`, so kvdb can create a different imported pubkey row than SQL for the same store request; validate the params against the scoped manager result or reject mismatches. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🛰️
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/legacy_test_helpers_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// testPrivPass is the private passphrase used by legacy waddrmgr test
+// helpers. It must match the seed/pass setup in newSpendableAddrMgr.
+var testPrivPass = []byte("priv")
+
+// newAddrStore is an alias for newSpendableAddrMgr used by the legacy
+// address-store test paths in this package.
+func newAddrStore(t *testing.T, dbConn walletdb.DB) *waddrmgr.Manager {
```
yyforyongyu: why this thin wrapper? remove it?

wallet/address_manager.go
```diff
@@ -49,12 +53,13 @@ var (
 	// ErrUnableToExtractAddress is returned when an address cannot be
 	// extracted from a pkscript.
 	ErrUnableToExtractAddress = errors.New("unable to extract address")
-
 	// errStopIteration is a special error used to stop the iteration in
 	// ForEachAccountAddress.
 	errStopIteration = errors.New("stop iteration")
 )
 
+const addressManagerPageLimit = 500
```
yyforyongyu, line 61: add some docs explaining this limit? we should add a `TODO(yy)` to make it configurable

yyforyongyu:
LGTM 🦦 (`openai/gpt-5.5`)

wallet/address_manager.go
```diff
@@ -208,23 +218,183 @@ func addressInfoFromManagedAddress(
...
+		Index:                storeAddr.Index,
+		MasterKeyFingerprint: storeAddr.MasterKeyFingerprint,
+	}
 
 	return info, nil
 }
 
+// storeAddressPubKeyCompressed reports whether store pubkey bytes use the
+// compressed secp256k1 encoding.
+func storeAddressPubKeyCompressed(pubKey []byte) bool {
```
yyforyongyu, line 305: why this thin wrapper? remove it?

wallet/address_manager.go
```diff
@@ -208,23 +218,183 @@ func addressInfoFromManagedAddress(
...
+	return len(pubKey) == btcec.PubKeyBytesLenCompressed
+}
+
+// legacyAddressBalances returns wallet address balances from the legacy
+// transaction store.
+func (w *Wallet) legacyAddressBalances() (map[string]btcutil.Amount, error) {
```
yyforyongyu: I thought this should be removed? and implement it in the kvdb side?

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -1,19 +1,135 @@
...
+//
+// NOTE: SQL backends use deriveFn as their derivation mechanism after
+// allocating an index. The kvdb adapter must preserve legacy address-manager
+// semantics, so it derives through waddrmgr and uses deriveFn only as optional
+// post-create validation when the callback is non-nil.
 func (s *Store) NewDerivedAddress(ctx context.Context,
-	_ db.NewDerivedAddressParams,
-	_ db.AddressDerivationFunc) (*db.AddressInfo, error) {
+	params db.NewDerivedAddressParams,
+	deriveFn db.AddressDerivationFunc) (*db.AddressInfo, error) {
```
yyforyongyu: i think the idea is to remove the `deriveFn` as a param and handle it in the db impl side? so for kvdb and sql they will choose the corresponding `deriveFn`?

wallet/address_manager.go
```diff
@@ -672,101 +672,57 @@ func (w *Wallet) GetAddressInfo(ctx context.Context, a btcutil.Address) (
...
+	req, err := addressPageRequest()
+	if err != nil {
+		return nil, err
+	}
 
-			addrToBalance[addr.String()] += utxo.Amount
-		}
+	// TODO(yy): switch to Store.ListUTXOs once it is implemented for
+	// kvdb (PR #1238).
+	addrToBalance, err := w.legacyAddressBalances()
```
yyforyongyu: ok i see, so this is deferred, fine by me

wallet/address_manager.go
yyforyongyu: Removed the thin helper. The remaining conversion stays in wallet because `addressInfoFromManagedAddress` adapts legacy `waddrmgr.ManagedAddress` values into the wallet API metadata shape; kvdb already has its own `managedAddressInfo` adapter for `db.AddressInfo`, so moving this wallet-only API conversion into kvdb would couple the public wallet shape to the kvdb backend.

yyforyongyu:
LGTM 🧭 (`openai/gpt-5.5`)

yyforyongyu:

yyforyongyu:
LGTM 🧶
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/addressstore_test.go
```diff
@@ -0,0 +1,101 @@
...
+	t.Parallel()
+
+	isWatchOnly, err := managedAddressIsWatchOnly(
+		false, unsupportedManagedPubKeyAddress{},
+	)
+	require.ErrorContains(t, err, "unsupported managed pubkey address")
+	require.False(t, isWatchOnly)
+}
+
+type unsupportedManagedPubKeyAddress struct {
```
yyforyongyu, line 239: we should `mock.Mock`?

wallet/signer.go
```diff
@@ -717,18 +718,54 @@ func (w *Wallet) ComputeUnlockingScript(ctx context.Context,
...
 
 // privKeyForOutput returns the private key needed to sign for the given
-// wallet-controlled address.
-func (w *Wallet) privKeyForOutput(addr btcutil.Address) (
+// wallet-controlled output.
+func (w *Wallet) privKeyForOutput(scriptInfo OutputScriptInfo) (
 	*btcec.PrivateKey, error) {
 
-	pubKeyAddr, err := w.loadManagedPubKeyAddr(addr)
+	if canUseAddressInfoDerivation(scriptInfo.AddressInfo) {
```
yyforyongyu: The store-derived private-key path added here is only used by `ComputeUnlockingScript`; `GetPrivKeyForAddress` still calls the legacy `PrivKeyForAddress` lookup. For a SQL-backed wallet, `NewAddress` now persists the derived address only in the SQL store, so the legacy lookup cannot find it; I reproduced this by creating a SQL-derived address with `NewAddress` and then calling `GetPrivKeyForAddress`, which failed with `unable to find key`. Route `GetPrivKeyForAddress` through `GetAddressInfo` plus the derivation-based private-key path, falling back to legacy for imported/kvdb cases. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/addressstore_test.go
yyforyongyu: Fixed: `unsupportedManagedPubKeyAddress` now embeds `mock.Mock` alongside `waddrmgr.ManagedAddress` (the embedded interface satisfies the type assertion path through `managedAddressIsWatchOnly`; `mock.Mock` backs the method shims). All 6 methods in the `ManagedPubKeyAddress` set (`Imported`, `PubKey`, `ExportPubKey`, `PrivKey`, `ExportPrivKey`, `DerivationInfo`) are now `m.Called(...)` shims. `TestManagedAddressIsWatchOnlyUnsupportedPubKey` drives the unsupported-pubkey path via `m.On("Imported").Return(true)` + `m.On("PubKey").Return((*btcec.PublicKey)(nil))`. Folded into `kvdb: implement derived address creation` per the originating-commit convention.

wallet/signer.go
yyforyongyu: Fixed: new tip commit `wallet: route GetPrivKeyForAddress through store-derived path`. `GetPrivKeyForAddress` now switches over `db.ErrAddressNotFound` — uses the new `canUseAddressInfoDerivation` + `privKeyForAddressInfo` helpers when the store record has usable derivation metadata, falls back to legacy `PrivKeyForAddress` only when the address is genuinely not in the store or lacks derivation info (imported / kvdb cases). Unexpected store errors surface instead of being masked.

The signer-side derivation path (which previously only served `ComputeUnlockingScript`) is now split into its own commit `wallet: add derivation-based privKey path for store addresses` so the helpers are reusable across entry points.

Regression test `TestGetPrivKeyForAddressSQLDerivedAddress` added in `wallet/signer_test.go`, reusing the existing `newSQLAddressSigningWallet` scaffolding.

wallet/internal/db/kvdb/addressstore_test.go
```diff
@@ -0,0 +1,127 @@
...
+// unsupportedManagedPubKeyAddress is a mock.Mock-based test double for
+// waddrmgr.ManagedPubKeyAddress used to exercise the unsupported pubkey path
+// in managedAddressIsWatchOnly.
+//
+// The embedded waddrmgr.ManagedAddress satisfies the base interface so the
+// type assertion to waddrmgr.ManagedPubKeyAddress succeeds; the mock.Mock
+// shim methods then drive the behavior the test exercises.
+type unsupportedManagedPubKeyAddress struct {
+	mock.Mock
+	waddrmgr.ManagedAddress
```
yyforyongyu, line 241: then this should be removed? why having this?

wallet/address_manager.go
```diff
@@ -318,25 +409,47 @@ func (w *Wallet) NewAddress(_ context.Context, accountName string,
...
+		path := waddrmgr.DerivationPath{
+			InternalAccount: accountID,
+			Account:         accountID + hdkeychain.HardenedKeyStart,
+			Branch:          branch,
+			Index:           index,
+		}
 
-	err = walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
-		addrmgrNs := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+		managedAddr, err := manager.DeriveFromKeyPath(ns, path)
```
yyforyongyu: SQL-only accounts created through `Wallet.NewAccount` do not exist in legacy `waddrmgr`, but `NewAddress` now asks the store to allocate an index and then this callback still derives via `manager.DeriveFromKeyPath`. Reproducing with a SQL store account number 1 fails with `derive address: derive key path: account 1 not found`, so newly-created accounts can be listed but cannot receive addresses. Derive from the SQL account material, or keep/mirror the legacy account before routing address generation through the store. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🎀
(`claude-opus-4-7[1m]`)

yyforyongyu (commented):
One minor observation (see inline).

wallet/address_manager_test.go
```diff
@@ -15,12 +16,102 @@ import (
...
+	t.Helper()
+
+	deps.store.On(
+		"NewDerivedAddress", mock.Anything,
+		db.NewDerivedAddressParams{
+			WalletID:    w.id,
+			AccountName: accountName,
+			Scope:       db.KeyScope(scope),
+			Change:      change,
+		}, mock.Anything,
```
yyforyongyu: Trailing `mock.Anything` matcher is functionally harmless with testify v1.11.1 (extra expected args are ignored when actual args are fewer), but `NewDerivedAddress` only takes 2 positional args (ctx, params) since the `deriveFn` parameter was removed in PR #1237. Consider removing this to match the current signature. (`deepseek-v4-pro`)

yyforyongyu (commented):
One minor observation (see inline).

yyforyongyu:
LGTM 🦦 (`openai/gpt-5.5`)

wallet/address_manager_test.go
yyforyongyu: Done — the `NewDerivedAddress` mock matcher now only takes `(ctx, params)`, dropping the trailing `mock.Anything` for the removed `deriveFn` parameter. Landed in 078da6f7a99e1d918320170be951a514d8d038d4 "wallet: trim derived-address mock matcher".

wallet/address_manager.go
yyforyongyu: The canonical fix landed on PR #1243 at 1e199e2d4 — `deriveAddressData` now derives directly from the account public key supplied through `AddressDerivationParams` instead of going through `waddrmgr.DeriveFromKeyPath`. SQL-only accounts no longer require a mirrored legacy waddrmgr account.

wallet/internal/db/kvdb/addressstore_test.go
yyforyongyu: Both embeds are required and the docstring just above the type captures the rationale: `waddrmgr.ManagedAddress` satisfies the base interface so the type assertion to `waddrmgr.ManagedPubKeyAddress` succeeds, while the `mock.Mock` shim methods drive the behavior the test exercises. Keeping the structure as-is.

yyforyongyu:
LGTM 🌷
(`claude-opus-4-7[1m]`)

yyforyongyu (commented):
LGTM 🦊 (`deepseek-v4-pro`)

yyforyongyu (commented):
LGTM 🦊 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🌼
(`claude-opus-4-7[1m]`)

yyforyongyu:
LGTM 🐚 (`gpt-5.5`)

GustavoStingelin (approved):
LGTM!

wallet/signer.go
yyforyongyu: Still live at current head `d95c32b5`: `GetPrivKeyForAddress` still just returns `w.PrivKeyForAddress(a)`, so SQL-derived addresses created only through the store still cannot export their private key through this API. Route this method through the store-derived address metadata path, falling back to the legacy lookup only when the store has no usable derivation record. (`gpt-5.5`)

wallet/address_manager.go
yyforyongyu: Still live at current head `d95c32b5`: `NewAddress` persists through `w.store.NewDerivedAddress`, but `ComputeUnlockingScript` still calls `privKeyForOutput(scriptInfo.Addr)`, which loads the key only from legacy `waddrmgr`. A SQL-derived address that exists only in the store can be returned and watched, but later signing still fails because the legacy address row is missing. Either keep the legacy manager in sync for SQL-derived addresses or route signing through the store derivation metadata. (`gpt-5.5`)

PR #1215, opened 2026-04-14
wallet: migrate account manager to db.Store

yyforyongyu:
## Summary
- add `AccountStore` to the transitional `db.Store` surface
- implement kvdb-backed account reads, derived account creation, and account import compatibility methods
- migrate `wallet/account_manager.go` off direct `walletdb` access and route its live paths through `w.store`

## Why
- the account manager was the last live wallet manager still heavily tied to direct `walletdb`, `waddrmgr`, and `wtxmgr` access
- moving it behind `db.Store` keeps the wallet migration progressing in the same layered way as the tx, utxo, and address managers
- this also deletes a large amount of walletdb-era helper logic once the live callers no longer need it

## What Changed
- `db.Store` now embeds `AccountStore`
- kvdb now implements the account-manager methods needed by wallet callers:
  - account reads (`GetAccount`, `ListAccounts`, `RenameAccount`)
  - derived account creation (`CreateDerivedAccount`)
  - transitional account import (`ImportAccount`) including dry-run validation
- `wallet/account_manager.go` now routes through `w.store` for:
  - account queries
  - balance lookups
  - `NewAccount`
  - `ImportAccount`
- stale walletdb-era account helper code was removed after the migration

## Commit Structure
- `db: add AccountStore to Store`
- `kvdb: implement account read methods`
- `wallet: route account queries through store`
- `wallet: route account balances through store`
- `kvdb: implement CreateDerivedAccount`
- `wallet: route NewAccount through store`
- `db: add ImportAccount compatibility method`
- `kvdb: implement ImportAccount`
- `wallet: route ImportAccount through store`
- `wallet: drop stale account-manager helpers`

## Testing
- `GOWORK=off scripts/check-each-commit.sh impl-address-manager-store`
- `GOWORK=off go test ./wallet/internal/db/... ./wallet`
- `GOWORK=off make lint-check`

### DIFF
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -26,7 +26,6 @@ import (
 	"github.com/btcsuite/btcwallet/netparams"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
-	"github.com/btcsuite/btcwallet/walletdb"
 )
 
 // errNilAccountInfo is returned when accountInfoToProperties receives a nil
@@ -496,118 +495,54 @@ func (w *Wallet) ImportAccount(ctx context.Context,
 }
 
 // importAccountInternal is the internal implementation of ImportAccount,
-// allowing callers (like Manager.Create) to bypass the started check.
+// allowing Manager.Create to bypass the started check.
 //
-// TODO(yy): we will move the db operation to a dedicated method, so we can
-// ignore cyclop for now.
-//
-//nolint:cyclop
-func (w *Wallet) importAccountInternal(_ context.Context,
+// dryRun is not supported on the new path; callers that need
+// simulate-and-rollback should use Wallet.ImportAccountDryRun (deprecated)
+// which still operates directly on waddrmgr buckets.
+func (w *Wallet) importAccountInternal(ctx context.Context,
 	name string, accountKey *hdkeychain.ExtendedKey,
 	masterKeyFingerprint uint32, addrType waddrmgr.AddressType,
 	dryRun bool) (*waddrmgr.AccountProperties, error) {
 
-	// Ensure we have a valid account public key. We require an account-level
-	// key (depth 3) to properly manage the derivation path.
-	err := validateExtendedPubKey(accountKey, true, w.cfg.ChainParams)
+	if dryRun {
+		return nil, errDryRunImportNotSupported
+	}
+
+	err := validateExtendedPubKey(
+		accountKey, true, w.cfg.ChainParams,
+	)
 	if err != nil {
 		return nil, err
 	}
 
-	// Determine what key scope the account public key should belong to and
-	// whether it should use a custom address schema. This is inferred from
-	// the key's HD version bytes.
-	keyScope, addrSchema, err := keyScopeFromPubKey(accountKey, &addrType)
+	keyScope, _, err := keyScopeFromPubKey(accountKey, &addrType)
 	if err != nil {
 		return nil, err
 	}
 
-	var props *waddrmgr.AccountProperties
-
-	// We'll perform the import within a database update transaction to ensure
-	// atomicity. If dryRun is enabled, we'll return a special error at the end
-	// to trigger a rollback.
-	err = walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
-		ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
-
-		// Check if a manager for this key scope already exists. If not, we'll
-		// create a new one using the inferred schema.
-		scopedMgr, err := w.addrStore.FetchScopedKeyManager(keyScope)
-		if err != nil {
-			scopedMgr, err = w.addrStore.NewScopedKeyManager(
-				ns, keyScope, *addrSchema,
-			)
-			if err != nil {
-				return err
-			}
+	info, err := w.store.CreateImportedAccount(ctx,
+		db.CreateImportedAccountParams{
+			WalletID:          w.id,
+			Name:              name,
+			Scope:             toDBKeyScope(keyScope),
+			MasterFingerprint: masterKeyFingerprint,
+			PublicKey:         []byte(accountKey.String()),
+		},
+	)
+	if err != nil {
+		// Preserve waddrmgr.ManagerError semantics so callers using
+		// waddrmgr.IsError(err, ...) keep working when kvdb wraps the
+		// underlying manager error via fmt.Errorf.
+		var mErr waddrmgr.ManagerError
+		if errors.As(err, &mErr) {
+			return nil, mErr
 		}
 
-		// Create the new watching-only account using the provided key. Since we
-		// only have the public key, the wallet won't be able to sign for this
-		// account unless the private key is also provided later.
-		account, err := scopedMgr.NewAccountWatchingOnly(
-			ns, name, accountKey, masterKeyFingerprint, addrSchema,
-		)
-		if err != nil {
-			return err
-		}
-
-		// Retrieve the properties for the newly created account.
-		props, err = scopedMgr.AccountProperties(ns, account)
-		if !dryRun {
-			return err
-		}
-
-		// If this is a dry-run, we'll generate a few addresses to simulate the
-		// import process and then roll back.
-		props, err = importAccountDryRun(ns, props, scopedMgr)
-		if err != nil {
-			return err
-		}
-
-		// Make sure we always roll back the dry-run transaction by returning an
-		// error here.
-		return walletdb.ErrDryRunRollBack
-	})
-
-	// If this was a dry-run, we ignore the rollback error.
-	if err != nil && !dryRun && !errors.Is(err, walletdb.ErrDryRunRollBack) {
 		return nil, err
 	}
 
-	return props, nil
-}
-
-// importAccountDryRun simulates an account import by generating a single
-// address for both the internal and external derivation branches. This ensures
-// that the provided account key is valid and can be used to derive addresses.
-// The changes made during this simulation are rolled back by the caller.
-func importAccountDryRun(ns walletdb.ReadWriteBucket,
-	props *waddrmgr.AccountProperties, scopedMgr waddrmgr.AccountStore) (
-	*waddrmgr.AccountProperties, error) {
-
-	// The importAccount method above will cache the imported account within the
-	// scoped manager. Since this is a dry-run attempt, we'll want to invalidate
-	// the cache for it.
-	defer scopedMgr.InvalidateAccountCache(props.AccountNumber)
-
-	_, err := scopedMgr.NextExternalAddresses(ns, props.AccountNumber, 1)
-	if err != nil {
-		return nil, err
-	}
-
-	_, err = scopedMgr.NextInternalAddresses(ns, props.AccountNumber, 1)
-	if err != nil {
-		return nil, err
-	}
-
-	// Refresh the account's properties after generating the addresses.
-	props, err = scopedMgr.AccountProperties(ns, props.AccountNumber)
-	if err != nil {
-		return nil, err
-	}
-
-	return props, nil
+	return accountInfoToProperties(info)
 }
 
 // validateExtendedPubKey ensures a sane derived public key is provided.
@@ -727,5 +662,3 @@ func extractAddrFromPKScript(pkScript []byte,
 	// issue.
 	return addrs[0]
 }
-
-
--- a/wallet/internal/db/kvdb/accountstore.go
+++ b/wallet/internal/db/kvdb/accountstore.go
@@ -158,6 +158,7 @@ func (o *createDerivedAccountOps) CreateDerivedAccount(_ context.Context,
 	// consistent.
 	now := time.Now().UTC()
 	scope := o.scopedMgr.Scope()
+
 	err = putAccountCreatedAt(
 		o.ns, scope, uint32(accountNumber), now, //nolint:gosec
 	)
@@ -317,6 +318,10 @@ func (s *Store) scopedManagerOrCreate(ns walletdb.ReadWriteBucket,
 
 // GetAccount returns the account identified by the given query within the
 // requested scope.
+//
+// path-mapping with origin classification; refactor blocked by task 105.
+//
+//nolint:cyclop // bridges waddrmgr's split account-by-number / account-by-name
 func (s *Store) GetAccount(_ context.Context,
 	query db.GetAccountQuery) (*db.AccountInfo, error) {
 
@@ -424,6 +429,11 @@ func resolveAccountNumber(ns walletdb.ReadBucket,
 
 // loadAccountInfo materializes a db.AccountInfo from waddrmgr's
 // AccountProperties + IsWatchOnlyAccount classifier.
+//
+// and created-at metadata from independent waddrmgr sources; refactor
+// blocked by task 105.
+//
+//nolint:cyclop // resolves origin, watch-only, fingerprint, key counts,
 func loadAccountInfo(ns walletdb.ReadBucket,
 	scopedMgr waddrmgr.AccountStore, accountNumber uint32,
 	walletWatchOnly bool,
@@ -476,8 +486,8 @@ func loadAccountInfo(ns walletdb.ReadBucket,
 	if origin == db.DerivedAccount {
 		if len(walletMasterHDPubKey) == 0 {
 			return nil, fmt.Errorf(
-				"derive master fingerprint for account "+
-					"%d: missing master HD pubkey",
+				"%w: account %d",
+				errMissingWalletMasterHDPubKey,
 				accountNumber,
 			)
 		}
@@ -541,6 +551,13 @@ func readMasterHDPubKey(mgr waddrmgr.AddrStore,
 
 	return nil, fmt.Errorf("read master HD pubkey: %w", err)
 }
+
+// errMissingWalletMasterHDPubKey is returned when a derived account
+// requires the wallet's master HD pubkey to derive its master-key
+// fingerprint but the legacy adapter exposes no pubkey for the wallet.
+var errMissingWalletMasterHDPubKey = errors.New(
+	"missing wallet master HD pubkey for derived account fingerprint",
+)
 
 // translateAccountErr maps waddrmgr account/scope not-found errors to their
 // public db.* equivalents.
--- a/wallet/account_manager_test.go
+++ b/wallet/account_manager_test.go
@@ -6,6 +6,7 @@ package wallet
 
 import (
 	"encoding/binary"
+	"math"
 	"strings"
 	"testing"
 
@@ -14,7 +15,6 @@ import (
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcwallet/waddrmgr"
-	"github.com/btcsuite/btcwallet/wtxmgr"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
@@ -397,37 +397,89 @@ func TestBalance(t *testing.T) {
 	require.Equal(t, btcutil.Amount(500), balance)
 }
 
-// TestImportAccount tests that the ImportAccount works as expected.
+// TestBalanceMaxConfsOverflow verifies impossible confirmation counts do not
+// wrap negative when adapted to the store query shape.
+func TestBalanceMaxConfsOverflow(t *testing.T) {
+	t.Parallel()
+
+	w, deps := createStartedWalletWithMocks(t)
+
+	scope := waddrmgr.KeyScopeBIP0084
+	dbScope := db.KeyScope{
+		Purpose: scope.Purpose,
+		Coin:    scope.Coin,
+	}
+	name := testAccountName
+
+	deps.store.On("GetAccount", mock.Anything, db.GetAccountQuery{
+		WalletID: 0,
+		Scope:    dbScope,
+		Name:     &name,
+	}).Return(&db.AccountInfo{
+		AccountNumber: 1,
+		AccountName:   name,
+		KeyScope:      dbScope,
+	}, nil).Once()
+
+	balance, err := w.Balance(t.Context(), math.MaxUint32, scope, name)
+	require.NoError(t, err)
+	require.Zero(t, balance)
+}
+
+// TestImportAccount verifies ImportAccount routes through Store.
 func TestImportAccount(t *testing.T) {
 	t.Parallel()
 
-	// Create a new test wallet.
 	w, deps := createStartedWalletWithMocks(t)
 
-	// We'll start by creating a new account under the BIP0084 scope.
-	scope := waddrmgr.KeyScopeBIP0084
-	addrType := waddrmgr.WitnessPubKey
-	masterPriv := "tprv8ZgxMBicQKsPeWwrFuNjEGTTDSY4mRLwd2KDJAPGa1AY" +
-		"quw38bZqNMSuB3V1Va3hqJBo9Pt8Sx7kBQer5cNMrb8SYquoWPt9" +
-		"Y3BZdhdtUcw"
-	root, err := hdkeychain.NewKeyFromString(masterPriv)
+	root, err := hdkeychain.NewMaster(
+		[]byte{
+			0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
+			0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f, 0x10,
+		},
+		&chainParams,
+	)
 	require.NoError(t, err)
-	acctPubKey := deriveAcctPubKey(t, root, scope, hardenedKey(0))
 
-	// Mock expectations for ImportAccount.
-	deps.addrStore.On("FetchScopedKeyManager", scope).
-		Return(deps.accountManager, nil).Once()
+	purpose, err := root.DeriveNonStandard( //nolint:staticcheck
+		hardenedKey(84),
+	)
+	require.NoError(t, err)
+	cointype, err := purpose.DeriveNonStandard( //nolint:staticcheck
+		hardenedKey(1),
+	)
+	require.NoError(t, err)
+	acct, err := cointype.DeriveNonStandard( //nolint:staticcheck
+		hardenedKey(0),
+	)
+	require.NoError(t, err)
 
-	deps.accountManager.On("NewAccountWatchingOnly", mock.Anything,
-		testAccountName, acctPubKey, root.ParentFingerprint(),
-		mock.Anything).Return(uint32(1), nil).Once()
-	deps.accountManager.On("AccountProperties", mock.Anything, uint32(1)).
-		Return(&waddrmgr.AccountProperties{
-			AccountNumber: 1,
-			AccountName:   testAccountName,
-		}, nil).Once()
+	acctPubKey, err := acct.Neuter()
+	require.NoError(t, err)
 
-	// We should be able to import the account.
+	addrType := waddrmgr.WitnessPubKey
+	scope := waddrmgr.KeyScopeBIP0084
+	dbScope := db.KeyScope{
+		Purpose: scope.Purpose,
+		Coin:    scope.Coin,
+	}
+
+	deps.store.On("CreateImportedAccount", mock.Anything,
+		db.CreateImportedAccountParams{
+			WalletID:          0,
+			Name:              testAccountName,
+			Scope:             dbScope,
+			MasterFingerprint: root.ParentFingerprint(),
+			PublicKey:         []byte(acctPubKey.String()),
+		}).Return(&db.AccountInfo{
+		AccountNumber: 1,
+		AccountName:   testAccountName,
+		Origin:        db.ImportedAccount,
+		IsWatchOnly:   true,
+		KeyScope:      dbScope,
+		PublicKey:     []byte(acctPubKey.String()),
+	}, nil).Once()
+
 	props, err := w.ImportAccount(
 		t.Context(), testAccountName, acctPubKey,
 		root.ParentFingerprint(), addrType, false,
@@ -435,87 +487,12 @@ func TestImportAccount(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, testAccountName, props.AccountName)
 
-	// Mock expectations for GetAccount.
-	deps.addrStore.On("FetchScopedKeyManager", scope).
-		Return(deps.accountManager, nil).Once()
-
-	deps.accountManager.On("LookupAccount", mock.Anything, testAccountName).
-		Return(uint32(1), nil).Once()
-	deps.accountManager.On("AccountProperties", mock.Anything, uint32(1)).
-		Return(&waddrmgr.AccountProperties{
-			AccountNumber: 1,
-			AccountName:   testAccountName,
-		}, nil).Once()
-
-	deps.txStore.On("UnspentOutputs", mock.Anything).
-		Return([]wtxmgr.Credit(nil), nil).Once()
-
-	// We should be able to get the account by its name.
-	_, err = w.GetAccount(t.Context(), scope, testAccountName)
-	require.NoError(t, err)
-
-	// Mock expectations for duplicate ImportAccount.
-	deps.addrStore.On("FetchScopedKeyManager", scope).
-		Return(deps.accountManager, nil).Once()
-	deps.accountManager.On("NewAccountWatchingOnly", mock.Anything,
-		testAccountName, acctPubKey, root.ParentFingerprint(),
-		mock.Anything).Return(uint32(0), waddrmgr.ManagerError{
-		ErrorCode: waddrmgr.ErrDuplicateAccount,
-	}).Once()
-
-	// We should not be able to import an account with the same name.
+	// Dry-run is rejected on the new path.
 	_, err = w.ImportAccount(
 		t.Context(), testAccountName, acctPubKey,
-		root.ParentFingerprint(), addrType, false,
-	)
-	require.Error(t, err)
-	require.True(
-		t, waddrmgr.IsError(err, waddrmgr.ErrDuplicateAccount),
-		"expected ErrDuplicateAccount",
-	)
-
-	// Mock expectations for dry-run.
-	dryRunName := "dry run"
-
-	deps.addrStore.On("FetchScopedKeyManager", scope).
-		Return(deps.accountManager, nil).Once()
-
-	deps.accountManager.On("NewAccountWatchingOnly", mock.Anything,
-		dryRunName, acctPubKey, root.ParentFingerprint(),
-		mock.Anything).Return(uint32(2), nil).Once()
-	deps.accountManager.On("AccountProperties", mock.Anything, uint32(2)).
-		Return(&waddrmgr.AccountProperties{
-			AccountNumber: 2,
-			AccountName:   dryRunName,
-		}, nil).Twice()
-	deps.accountManager.On("InvalidateAccountCache", uint32(2)).Return().Once()
-	deps.accountManager.On("NextExternalAddresses", mock.Anything, uint32(2),
-		uint32(1)).Return([]waddrmgr.ManagedAddress(nil), nil).Once()
-	deps.accountManager.On("NextInternalAddresses", mock.Anything, uint32(2),
-		uint32(1)).Return([]waddrmgr.ManagedAddress(nil), nil).Once()
-
-	// We should be able to do a dry run of the import.
-	_, err = w.ImportAccount(
-		t.Context(), dryRunName, acctPubKey,
 		root.ParentFingerprint(), addrType, true,
 	)
-	require.NoError(t, err)
-
-	// Mock expectations for GetAccount (fail).
-	deps.addrStore.On("FetchScopedKeyManager", scope).
-		Return(deps.accountManager, nil).Once()
-	deps.accountManager.On("LookupAccount", mock.Anything, dryRunName).
-		Return(uint32(0), waddrmgr.ManagerError{
-			ErrorCode: waddrmgr.ErrAccountNotFound,
-		}).Once()
-
-	// The account should not have been imported.
-	_, err = w.GetAccount(t.Context(), scope, dryRunName)
 	require.Error(t, err)
-	require.True(
-		t, waddrmgr.IsError(err, waddrmgr.ErrAccountNotFound),
-		"expected ErrAccountNotFound",
-	)
 }
 
 // TestExtractAddrFromPKScript tests that the extractAddrFromPKScript
--- a/wallet/internal/db/kvdb/accountstore_timestamps.go
+++ b/wallet/internal/db/kvdb/accountstore_timestamps.go
@@ -6,6 +6,7 @@ package kvdb
 
 import (
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"time"
 
@@ -22,6 +23,19 @@ import (
 // See ADR (forthcoming) and task 102.
 var kvdbAccountCreatedAtBucketKey = []byte("kvdb-account-created-at")
 
+// createdAtValueLen is the on-disk byte width of a per-account creation
+// timestamp encoded as a big-endian int64 nanosecond offset from the
+// Unix epoch.
+const createdAtValueLen = 8
+
+// errCreatedAtUnexpectedSize is returned when the side bucket's value
+// for an existing (scope, account) key has an unexpected byte width.
+// The kvdb adapter rejects rather than silently truncates because a
+// short value would decode as an arbitrary time.
+var errCreatedAtUnexpectedSize = errors.New(
+	"created-at bucket value has unexpected byte width",
+)
+
 // accountCreatedAtKey is the 12-byte key used inside the side bucket,
 // matching the accountBalanceKey shape so future per-(scope, account)
 // kvdb-owned metadata can use the same layout.
@@ -36,6 +50,7 @@ func newAccountCreatedAtKey(scope waddrmgr.KeyScope,
 	binary.BigEndian.PutUint32(k[0:4], scope.Purpose)
 	binary.BigEndian.PutUint32(k[4:8], scope.Coin)
 	binary.BigEndian.PutUint32(k[8:12], account)
+
 	return k
 }
 
@@ -53,10 +68,12 @@ func putAccountCreatedAt(ns walletdb.ReadWriteBucket,
 
 	key := newAccountCreatedAtKey(scope, account)
 
-	var value [8]byte
+	var value [createdAtValueLen]byte
+	//nolint:gosec // UnixNano() always fits in uint64 for times after 1970.
 	binary.BigEndian.PutUint64(value[:], uint64(t.UTC().UnixNano()))
 
-	if err := bucket.Put(key[:], value[:]); err != nil {
+	err = bucket.Put(key[:], value[:])
+	if err != nil {
 		return fmt.Errorf("put created-at: %w", err)
 	}
 
@@ -83,14 +100,16 @@ func readAccountCreatedAt(ns walletdb.ReadBucket,
 		return time.Time{}, nil
 	}
 
-	if len(raw) != 8 {
+	if len(raw) != createdAtValueLen {
 		return time.Time{}, fmt.Errorf(
-			"created-at bucket value for scope=%v account=%d: "+
-				"expected 8 bytes, got %d",
-			scope, account, len(raw),
+			"%w: scope=%v account=%d: expected %d bytes, got %d",
+			errCreatedAtUnexpectedSize, scope, account,
+			createdAtValueLen, len(raw),
 		)
 	}
 
+	//nolint:gosec // Stored values are written via UnixNano() of
+	// time.Now() in putAccountCreatedAt; round-tripping fits int64.
 	nanos := int64(binary.BigEndian.Uint64(raw))
 
 	return time.Unix(0, nanos).UTC(), nil


yyforyongyu:

wallet/account_manager.go
```diff
@@ -510,89 +409,40 @@ func (w *Wallet) RenameAccount(_ context.Context, scope waddrmgr.KeyScope,
...
-		return 0
-	}
+	minConfs := int32(conf)
+	accountNum := account.AccountNumber
 
-	// If the address belongs to the target account, add the UTXO's value
-	// to the total balance.
-	if addrScope.Scope() == scope && addrAcc == accNum {
-		return utxo.Amount
+	balance, err := w.store.Balance(ctx, db.BalanceParams{
```
yyforyongyu: This now routes the public `Wallet.Balance` API to `w.store.Balance`, but the live wallet still wires `w.store` to `kvdb.NewStore(...)` in `manager.go`, and `kvdb.Store.Balance` currently returns `not implemented` unconditionally (`wallet/internal/db/kvdb/utxostore.go`). That means the existing kvdb wallet balance API regresses from summing UTXOs to returning an error for every account. Please either implement the kvdb `Balance` adapter in this PR, or keep this method on the legacy balance path until the store method is available. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -169,6 +173,72 @@ type AccountsResult struct {
 	CurrentBlockHeight int32
 }
 
+// accountResultFromStoreInfo adapts one store account row to the wallet-facing
+// account result.
+func accountResultFromStoreInfo(info db.AccountInfo) AccountResult {
+	return AccountResult{
+		AccountProperties: accountPropertiesFromStoreInfo(info),
+		TotalBalance:      info.ConfirmedBalance + info.UnconfirmedBalance,
```
yyforyongyu: `db.AccountInfo` balances are still zero-filled today (`BuildAccountInfo` explicitly sets both balance fields to 0), and the new kvdb adapter also builds account infos through that path. Since the old `ListAccounts*` and `GetAccount` implementations populated `TotalBalance` by walking the wallet UTXOs, this change makes those account result APIs report zero balances even when the account has spendable outputs. Please keep the previous balance aggregation here or teach the store account queries to return real balances before switching these wallet-facing results over. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,628 @@
...
+
+		manager, err := kvdbScopedManager(
+			ns, s.addrStore, params.Scope,
+		)
+		if err != nil {
+			return err
+		}
+
+		accountNum, err := manager.NewAccountWatchingOnly(
+			ns, params.Name, accountKey, params.MasterFingerprint, nil,
```
yyforyongyu: The import path drops the per-account address schema selected by `keyScopeFromPubKey`. For example, BIP49 xpubs imported with `NestedWitnessPubKey` return `KeyScopeBIP0049Plus` plus `KeyScopeBIP0049AddrSchema`, and the old code passed that `addrSchema` into `NewAccountWatchingOnly`; this PR only passes the scope through `CreateImportedAccountParams` and then calls `NewAccountWatchingOnly(..., nil)`. The persisted account can therefore use the scope default schema instead of the caller-requested imported-account schema. Please carry the optional addr schema through the store params and pass it both when creating a missing scoped manager and when calling `NewAccountWatchingOnly`. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -644,92 +489,104 @@ func (w *Wallet) importAccountInternal(_ context.Context,
...
+	if err != nil {
+		return nil, err
+	}
 
-	// We'll perform the import within a database update transaction to ensure
-	// atomicity. If dryRun is enabled, we'll return a special error at the end
-	// to trigger a rollback.
-	err = walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
-		ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+	if dryRun {
```
yyforyongyu, line 508: Dry-run imports no longer exercise the store/address-manager path at all. Previously the dry run created the watching-only account and derived one external/internal address inside a rollback transaction, so it also caught duplicate account names and manager-level import failures. Now a duplicate existing account name returns success here and only fails on the later non-dry-run import, which weakens the dry-run validation contract. Please either perform the store import in a rollback/dry-run path again, or add an explicit account-name existence check before returning success. (`openai/gpt-5.5`)

wallet/account_manager.go
yyforyongyu: This passes only `accountNum` into `db.BalanceParams` while account numbers are scoped (not wallet-global), so `Balance` can overcount across scopes with the same number; include `scope` in the balance filter (or pass a unique account ID) and add a cross-scope regression test (`openai/gpt-5.3-codex`).

wallet/account_manager.go
yyforyongyu: This now derives `TotalBalance` from `db.AccountInfo` fields that are currently zeroed by store builders, so `ListAccounts*`/`GetAccount` silently regress to always reporting zero balances; keep using UTXO aggregation here until store-backed balance fields are populated (or fetch per-account balances explicitly from `store.Balance`) (`openai/gpt-5.3-codex`).

wallet/account_manager.go
yyforyongyu, line 508: This returns from dry-run before `CreateImportedAccount` is called, so dry-run no longer validates scope/name uniqueness despite the API contract requiring full import validation; execute the same store insertion path inside a rollback-only transaction (or add an explicit duplicate-name check) to preserve dry-run parity (`openai/gpt-5.3-codex`).

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,628 @@
+package kvdb
```
yyforyongyu, line 1: Missing copyright header. All new Go files in this repo should include the standard license header before the package declaration.

`k2p6`

wallet/internal/db/kvdb/accountstore_test.go
```diff
@@ -0,0 +1,297 @@
+package kvdb
```
yyforyongyu, line 1: Missing copyright header. All new Go files in this repo should include the standard license header before the package declaration.

`k2p6`

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,628 @@
...
+
+	return props, nil
+}
+
+// kvdbDecryptAccountPubKey decrypts and parses the encrypted account public key
+// for the legacy address manager.
+func kvdbDecryptAccountPubKey(addrStore legacyAddrStore,
+	params db.CreateImportedAccountParams) (*hdkeychain.ExtendedKey, error) {
+
+	decryptor, ok := addrStore.(legacyCryptoStore)
```
yyforyongyu: `decryptor, ok := addrStore.(legacyCryptoStore)` is dead defensive code — the only construction path passes `*waddrmgr.Manager` which satisfies both `legacyAddrStore` (line 34) and `legacyCryptoStore` (line 45), so the `!ok` branch and `errMissingLegacyCryptoStore` are unreachable; merge `Decrypt` into `legacyAddrStore` and drop the sentinel.

`claude-opus-4-7[1m]`

wallet/mock_test.go
```diff
@@ -206,6 +365,105 @@ func (m *mockStore) RollbackToBlock(ctx context.Context, height uint32) error {
...
+
+		props, err := manager.AccountProperties(nil, accountNum)
+		if err != nil {
+			return nil, err
+		}
+
+		return []db.AccountInfo{mockAccountInfoFromProps(props)}, nil
+	}
+
+	lastAccount, err := manager.LastAccount(nil)
```
yyforyongyu: `manager.LastAccount(nil)` plus a `0..lastAccount` loop omits the legacy `imported` bucket account (account number `waddrmgr.ImportedAddrAccount` = 2^31-1), while `kvdbListScopeAccounts` in `wallet/internal/db/kvdb/accountstore.go:474` uses `manager.ForEachAccount` which does include it; a wallet test that asserts `require.Len(t, accounts.Accounts, N)` will pass against this mock but disagree with kvdb on count. Iterate via `ForEachAccount` so the mock mirrors kvdb behavior.

`claude-opus-4-7[1m]`

wallet/mock_test.go
```diff
@@ -206,6 +365,105 @@ func (m *mockStore) RollbackToBlock(ctx context.Context, height uint32) error {
...
+	props *waddrmgr.AccountProperties) db.AccountInfo {
+
+	origin := db.DerivedAccount
+	accountNum := props.AccountNumber
+
+	if props.AccountNumber == waddrmgr.ImportedAddrAccount ||
+		props.AccountName == waddrmgr.ImportedAddrAccountName {
+
+		origin = db.ImportedAccount
+		accountNum = 0
```
yyforyongyu: Origin classification diverges from `kvdbAccountInfoFromProps` in `wallet/internal/db/kvdb/accountstore.go:495`: the mock keys off `AccountNumber == ImportedAddrAccount || AccountName == ImportedAddrAccountName`, kvdb keys off `IsWatchOnly || AccountPubKey == nil || AccountNumber == ImportedAddrAccount`. For an xpub-imported account (`NewAccountWatchingOnly`), `IsWatchOnly=true` but the name/number are user-set, so the mock returns `db.DerivedAccount` while kvdb returns `db.ImportedAccount` — wallet tests routed through the delegation path silently validate the wrong policy. Pick one classifier and use it in both places.

`claude-opus-4-7[1m]`

wallet/internal/db/kvdb/driver.go
```diff
@@ -12,19 +12,28 @@ import (
...
 
 // A compile-time assertion to ensure that Store implements the db.Store
 // interface.
 var _ db.Store = (*Store)(nil)
 
-// NewStore creates a new kvdb-backed UTXO store.
-func NewStore(dbConn walletdb.DB, txStore wtxmgr.TxStore) *Store {
+// NewStore creates a new kvdb-backed wallet store adapter.
+func NewStore(dbConn walletdb.DB, txStore wtxmgr.TxStore,
+	addrStore ...legacyAddrStore) *Store {
```
yyforyongyu: Hiding a "required for AccountStore methods" dependency behind `addrStore ...legacyAddrStore` lets `wallet/deprecated.go:7189` continue calling `kvdb.NewStore(db, txMgr)` (no addrMgr), so any `Wallet.NewAccount`/`ListAccounts`/etc. on a wallet opened via `Loader.OpenExistingWallet` returns `errMissingLegacyAddrStore` at runtime instead of a build-time error pointing at the caller. Prefer a required positional plus a typed Options/separate constructor so the two construction sites surface the divergence at compile time.

`claude-opus-4-7[1m]`

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,628 @@
...
+// kvdbAccountInfoFromProps adapts one legacy account properties record into the
+// db account view used by store callers.
+func kvdbAccountInfoFromProps(
+	props *waddrmgr.AccountProperties) *db.AccountInfo {
+
+	origin := db.DerivedAccount
+	accountNum := props.AccountNumber
+
+	if props.IsWatchOnly || props.AccountPubKey == nil ||
+		props.AccountNumber == waddrmgr.ImportedAddrAccount {
```
yyforyongyu: `props.IsWatchOnly` and `props.AccountPubKey == nil` are unreliable Origin discriminators: per `waddrmgr/scoped_manager.go:678`, `IsWatchOnly = s.rootManager.WatchOnly() || acctInfo.acctKeyPriv == nil`, so in a watch-only wallet every derived account, and on any locked-wallet read where `acctKeyPriv` is nil, will land as `Origin: db.ImportedAccount`; `AccountPubKey == nil` only holds for the legacy bucket account, which `AccountNumber == waddrmgr.ImportedAddrAccount` already covers. Drop the two extra arms and key off `AccountNumber` (or `AccountName == ImportedAddrAccountName`) alone — xpub-imported accounts that the kvdb path cannot otherwise identify are a known gap that the existing waddrmgr API cannot disambiguate without exposing `acctType`.

`claude-opus-4-7[1m]`

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,628 @@
...
+	}
+
+	if !params.IsWatchOnly() {
+		return nil, fmt.Errorf(
+			"kvdb.Store.CreateImportedAccount: private account: %w",
+			errNotImplemented,
+		)
+	}
+
+	accountKey, err := kvdbDecryptAccountPubKey(s.addrStore, params)
```
yyforyongyu: `CreateImportedAccount` participates in a ciphertext round-trip — the wallet encrypts via `Encrypt(CKTPublic, ...)` (`wallet/account_manager.go:582`), this method immediately decrypts the same bytes via the same crypto key, and `manager.NewAccountWatchingOnly` then re-encrypts for legacy storage. The shape is dictated by `CreateImportedAccountParams` being designed for the SQL backend (which stores the ciphertext) so the kvdb path has to invert it; a short godoc on `CreateImportedAccount` or `kvdbDecryptAccountPubKey` documenting this would save the next reader from concluding it is a layering bug.

`claude-opus-4-7[1m]`

wallet/account_manager.go
yyforyongyu: Fixed by keeping Wallet.Balance on the legacy UTXO aggregation path until kvdb Balance is implemented. Pushed to both origin and upstream.

wallet/account_manager.go
yyforyongyu: Fixed by restoring the wallet-side UTXO balance overlay for account result APIs, so store account rows no longer zero out wallet-facing balances. Pushed to both origin and upstream.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed by carrying the optional address schema through CreateImportedAccountParams and passing it when creating the scoped manager and watching-only account. Pushed to both origin and upstream.

wallet/account_manager.go
yyforyongyu: Fixed by adding dry-run validation for reserved and duplicate imported account names before returning success. Pushed to both origin and upstream.

wallet/account_manager.go
yyforyongyu: Fixed by keeping Wallet.Balance on the legacy scoped UTXO aggregation path instead of the not-yet-implemented store Balance path. Pushed to both origin and upstream.

wallet/account_manager.go
yyforyongyu: Fixed by restoring wallet-side UTXO aggregation before constructing account results. Pushed to both origin and upstream.

wallet/account_manager.go
yyforyongyu: Fixed by validating duplicate and reserved account names in the dry-run path before returning success. Pushed to both origin and upstream.

wallet/account_manager.go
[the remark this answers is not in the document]
yyforyongyu: Fixed by overlaying UTXO-derived balances onto store account rows before returning wallet-facing account results. Pushed to both origin and upstream.

wallet/account_manager.go
```diff
@@ -510,89 +409,40 @@ func (w *Wallet) RenameAccount(_ context.Context, scope waddrmgr.KeyScope,
...
+		WalletID: w.id,
+		Account:  &accountNum,
+		MinConfs: &minConfs,
+	})
+	if err != nil {
+		return 0, err
 	}
 
-	return 0
+	return balance.Total, nil
```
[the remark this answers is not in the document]
yyforyongyu: Fixed by keeping Wallet.Balance on the existing legacy UTXO balance path until kvdb.Store.Balance exists. Pushed to both origin and upstream.

wallet/account_manager.go
```diff
@@ -182,51 +252,14 @@ type AccountsResult struct {
 // The time complexity of this method is O(U*logA + A), where U is the number of
 // UTXOs and A is the number of accounts in the wallet. A potential future
 // improvement is to make the balance calculation optional.
```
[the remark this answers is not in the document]
yyforyongyu: Fixed by restoring the balance-aware account listing flow, so the existing complexity/commentary remains accurate. Pushed to both origin and upstream.

wallet/account_manager.go
```diff
@@ -644,92 +489,104 @@ func (w *Wallet) importAccountInternal(_ context.Context,
...
-	// We'll perform the import within a database update transaction to ensure
-	// atomicity. If dryRun is enabled, we'll return a special error at the end
-	// to trigger a rollback.
-	err = walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
-		ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
+	if dryRun {
+		return importedAccountDryRunProperties(
+			name, accountKey, masterKeyFingerprint, keyScope, addrSchema,
+		), nil
+	}
```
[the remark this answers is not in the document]
yyforyongyu: Fixed by adding explicit dry-run checks for reserved account names and duplicates within the target scope. Pushed to both origin and upstream.

wallet/manager.go
```diff
@@ -301,7 +301,7 @@ func (m *Manager) Load(cfg Config) (*Wallet, error) {
 		cfg:         cfg,
 		id:          walletID,
 		addrStore:   addrMgr,
-		store:       kvdb.NewStore(cfg.DB, txMgr),
+		store:       kvdb.NewStore(cfg.DB, txMgr, addrMgr),
 		txStore:     txMgr,
 		requestChan: make(chan any),
```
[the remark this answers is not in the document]
yyforyongyu: Fixed by making kvdb.NewStore require the legacy address store positionally and updating the remaining wallet construction path. Pushed to both origin and upstream.

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,628 @@
...
+// kvdbListScopeAccounts lists all matching accounts from one scoped manager.
+func kvdbListScopeAccounts(ns walletdb.ReadBucket,
+	manager waddrmgr.AccountStore,
+	name *string) ([]db.AccountInfo, error) {
+
+	if name != nil {
+		accountNum, err := manager.LookupAccount(ns, *name)
+		if err != nil {
+			if waddrmgr.IsError(err, waddrmgr.ErrAccountNotFound) {
+				return nil, nil
```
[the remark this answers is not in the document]
yyforyongyu: Fixed by returning an empty db.AccountInfo slice for name-filter misses instead of nil. Pushed to both origin and upstream.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed by merging Decrypt into legacyAddrStore and removing the legacyCryptoStore assertion and missing-crypto sentinel. Pushed to both origin and upstream.

wallet/mock_test.go
yyforyongyu: Fixed by making the wallet mock list accounts through ForEachAccount so it mirrors kvdb and includes the imported bucket account. Pushed to both origin and upstream.

wallet/mock_test.go
yyforyongyu: Fixed by using the same stable origin classifier in kvdb and the mock, based on the imported account sentinel number/name. Pushed to both origin and upstream.

wallet/internal/db/kvdb/driver.go
yyforyongyu: Fixed by changing kvdb.NewStore to require the legacy address store as a positional constructor argument, surfacing missing callers at compile time. Pushed to both origin and upstream.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed by dropping the watch-only and nil-account-key origin checks and classifying imported accounts only by the stable legacy sentinel number/name. Pushed to both origin and upstream.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed by documenting the SQL ciphertext versus kvdb legacy decrypt and re-encrypt round trip in kvdbDecryptAccountPubKey. Pushed to both origin and upstream.

yyforyongyu:

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,647 @@
...
+		for _, manager := range scopes {
+			scopeInfos, err := kvdbListScopeAccounts(
+				ns, manager, query.Name,
+			)
+			if err != nil {
+				return err
+			}
+
+			infos = append(infos, scopeInfos...)
+		}
```
yyforyongyu, line 616: `ListAccounts` rejects the `Scope` + `Name` combination with `ErrInvalidAccountQuery`, but the SQL backends support this via queries like `ListAccountsByWalletScopeAndName`. This creates an API inconsistency between kvdb and SQL backends. Wallet callers that pass both filters will work with SQL but fail with kvdb.

`k2p6`

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,647 @@
...
+		KeyScope:             db.KeyScope(props.KeyScope),
+		IsWatchOnly:          props.IsWatchOnly,
+		AddrSchema:           addrSchema,
+	}
+}
+
+// kvdbAccountOriginFromProps maps legacy account properties to a db account
+// origin using only stable legacy fields.
+func kvdbAccountOriginFromProps(
+	props *waddrmgr.AccountProperties) db.AccountOrigin {
```
yyforyongyu: `kvdbAccountOriginFromProps` only checks `AccountNumber` and `AccountName` to determine origin, missing imported watch-only accounts created via `NewAccountWatchingOnly` with custom names. Legacy imported accounts have `IsWatchOnly = true` in their `AccountProperties` and will be misclassified as `DerivedAccount` by `GetAccount` and `ListAccounts`. The same bug exists in `mockAccountOriginFromProps` in `wallet/mock_test.go`.

`k2p6`

wallet/mock_test.go
```diff
@@ -119,6 +122,162 @@ func (m *mockStore) Balance(ctx context.Context,
...
+	manager, err := m.addrStore.FetchScopedKeyManager(
+		waddrmgr.KeyScope(params.Scope),
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	canAddManager, ok := manager.(interface{ CanAddAccount() error })
+	if ok {
+		err = canAddManager.CanAddAccount()
```
yyforyongyu: The `mockStore` fallback methods pass `nil` for the namespace parameter to legacy waddrmgr methods (`manager.NewAccount(nil, params.Name)`, `manager.LookupAccount(nil, ...)`, `manager.AccountProperties(nil, ...)`, `manager.RenameAccount(nil, ...)`). If a test calls these methods without setting explicit mock expectations, the fallback will panic because the legacy manager expects a valid `walletdb.ReadBucket`. The mockStore has no access to the test database, so it cannot execute these fallbacks safely.

`k2p6`

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,647 @@
...
+		}
+
+		legacyProps, err := manager.AccountProperties(ns, accountNum)
+		if err != nil {
+			return fmt.Errorf("account properties: %w", err)
+		}
+
+		props = kvdbAccountPropertiesFromProps(legacyProps)
+		props.AccountNumber = 0
+		props.Origin = db.ImportedAccount
```
yyforyongyu: `props.AccountNumber = 0; props.Origin = db.ImportedAccount` overrides what `kvdbAccountPropertiesFromProps` just produced, forcing the imported classification on the Create-return path only. The same xpub-imported account read back via `GetAccount`/`ListAccounts` goes through `kvdbAccountInfoFromProps` → `kvdbAccountOriginFromProps`, which only matches `AccountNumber == waddrmgr.ImportedAddrAccount || AccountName == waddrmgr.ImportedAddrAccountName` — neither holds for an xpub-imported account named "watch-only", so reads return `Origin: DerivedAccount, AccountNumber: <real legacy num>` while Create returned `Origin: ImportedAccount, AccountNumber: 0`. `TestCreateImportedAccountSuccess` at `accountstore_test.go:198-199` only asserts `info.AccountName` and would still pass — add `require.Equal(t, db.ImportedAccount, info.Origin)` (and on `info.AccountNumber`) and the inconsistency surfaces. Either apply the same override inside `kvdbAccountInfoFromProps` keyed off a stable read-time signal (e.g., `managedAddr.Imported()` of any address in the account, or stashing an imported-account marker at create time), or drop the override and accept that xpub-imported accounts look identical to derived ones via kvdb until acctType is exposed by waddrmgr.

`claude-opus-4-7[1m]`

wallet/account_manager.go
```diff
@@ -114,37 +124,28 @@ var _ AccountManager = (*Wallet)(nil)
...
+func (w *Wallet) NewAccount(ctx context.Context, scope waddrmgr.KeyScope,
 	name string) (*waddrmgr.AccountProperties, error) {
 
 	err := w.state.validateStarted()
 	if err != nil {
 		return nil, err
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(scope)
+	info, err := w.store.CreateDerivedAccount(
```
yyforyongyu: Before this store call we no longer run `waddrmgr.ValidateAccountName`. The kvdb adapter still validates through `manager.NewAccount`, but the SQL store only rejects empty names and the migrations only enforce uniqueness, so `NewAccount(..., "imported")` can persist a reserved account name on SQL. The same backend-dependent validation gap applies to `RenameAccount` and the non-dry-run `ImportAccount` path. Please validate account names in the wallet layer, or in a shared store helper, before all store mutations. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -169,6 +170,93 @@ type AccountsResult struct {
...
+		AccountProperties: accountPropertiesFromStoreInfo(info),
+	}
+}
+
+// accountPropertiesFromStoreInfo adapts one store account row to legacy
+// wallet account properties.
+func accountPropertiesFromStoreInfo(
+	info db.AccountInfo) waddrmgr.AccountProperties {
+
+	return waddrmgr.AccountProperties{
```
yyforyongyu: `GetAccount` and `ListAccounts*` now build wallet-facing `waddrmgr.AccountProperties` from `db.AccountInfo`, but that store type does not carry the imported account public key, master fingerprint, or per-account address schema. The previous path returned those fields from `manager.AccountProperties`, and `ImportAccount` callers also get them on dry-run. Please use a full account-properties query for wallet-facing account results or extend this mapping so imported accounts do not lose `AccountPubKey`, `MasterKeyFingerprint`, and `AddrSchema`. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -400,104 +390,49 @@ func (w *Wallet) ListAccountsByName(_ context.Context,
...
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(scope)
+	info, err := w.store.GetAccount(ctx, db.GetAccountQuery{
+		WalletID: w.id,
+		Scope:    db.KeyScope(scope),
+		Name:     &name,
+	})
 	if err != nil {
 		return nil, err
```
yyforyongyu, line 376: Returning store errors directly changes the public wallet error surface for missing accounts. The kvdb adapter now wraps `db.ErrAccountNotFound`, while existing callers and tests historically check `waddrmgr.IsError(err, waddrmgr.ErrAccountNotFound)` for `GetAccount`; `RenameAccount` has the same issue for a missing old name. Please translate store not-found errors back to the legacy wallet/manager error before returning from these public APIs. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed in the pushed account stack. `ListAccounts` now supports `Scope` + `Name` consistently across shared dispatch, SQL/PG, and kvdb, with backend parity coverage for the combined filter.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed in the pushed account stack. `waddrmgr.AccountProperties` now exposes `IsImported`, kvdb uses that stable signal for origin/account-number mapping, and the mock conversion follows the same semantics.

wallet/mock_test.go
yyforyongyu: Fixed in the pushed account stack. The mock store fallback now has access to the test walletdb namespace and uses real read/write transactions instead of passing nil buckets into legacy waddrmgr methods.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed in the pushed account stack. The create/read origin mismatch is avoided by removing kvdb imported-account creation from the store bridge during the keyvault transition and by using the new `IsImported` metadata for kvdb read mapping.

wallet/account_manager.go
yyforyongyu: Fixed in the pushed account stack. Wallet account mutations now run `waddrmgr.ValidateAccountName` before store calls for `NewAccount`, `RenameAccount`, and non-dry-run `ImportAccount`.

wallet/account_manager.go
yyforyongyu: Addressed in the pushed account stack. Store import results are converted through account properties, dry-run remains aligned with the persisted path, and kvdb falls back to the legacy import path until the keyvault-backed store bridge can preserve the full legacy metadata surface.

wallet/account_manager.go
yyforyongyu: Fixed in the pushed account stack. Public `GetAccount` and `RenameAccount` translate store `ErrAccountNotFound` back to the legacy `waddrmgr.ErrAccountNotFound` surface before returning.

yyforyongyu:
Pushed a follow-up rewrite to `impl-account-manager-store`.

Summary:
- Introduced the kvdb legacy addr-store dependency before the first kvdb account method uses it.
- Folded out incidental license-header churn from pre-existing kvdb files.
- Restored wallet-facing imported account metadata for `ImportAccount`, `GetAccount`, and `ListAccounts*`, including dry-run coverage.
- Restored wallet-layer account balance overlay coverage and cleaned up lint findings.

Verification:
- `git diff --check sql-wallet..HEAD`
- `gofmt -l` on changed Go files
- `GOWORK=off go test -count=1 ./waddrmgr ./wallet ./wallet/internal/db ./wallet/internal/db/kvdb ./wallet/internal/db/sqlite ./wallet/internal/db/pg`
- `GOWORK=off go test -count=1 -tags=itest ./wallet/internal/db/itest -run 'Test(GetAccount|CreateDerivedAccount|CreateImportedAccount|ListAccounts|RenameAccount|AccountTypes)'`
- `make lint-check`

yyforyongyu:

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -20,23 +82,367 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("GetAccount: %w", err)
+	}
+
+	return info, nil
+}
+
+// ListAccounts lists accounts through the legacy address-manager path.
+func (s *Store) ListAccounts(_ context.Context,
```
yyforyongyu, line 580: `kvdb.Store.ListAccounts` happily applies both `Scope` and `Name` filters together (and the new test at `accountstore_test.go:234-244` asserts a single result for that combination). The SQL backends route through `db.ListAccountsByQuery` (`accounts_common.go:438`) which rejects the same combination with `ErrInvalidAccountQuery` because there is no `ListAccountsByWalletScopeAndName` sqlc query. Result: a caller that issues `{Scope, Name}` succeeds on a kvdb-backed wallet and errors on a pg/sqlite-backed wallet. Either drop the dual-filter branch from `kvdb.ListAccounts` (and the test case) so kvdb matches the SQL contract, or add the missing combined query to SQL and update `ListAccountsByQuery` to dispatch to it.

`claude-opus-4-7[1m]`

wallet/account_manager.go
```diff
@@ -110,41 +112,37 @@ type AccountManager interface {
...
-
-		// Create a new account under the current key scope.
-		accNum, err := manager.NewAccount(addrmgrNs, name)
-		if err != nil {
-			return err
-		}
-
-		// Get the account's properties.
-		props, err = manager.AccountProperties(addrmgrNs, accNum)
+	info, err := w.store.CreateDerivedAccount(
```
yyforyongyu: Routing `NewAccount` straight to `w.store.CreateDerivedAccount` preserves kvdb behavior, but the SQL stores implement that method by inserting only the account row. They do not derive or persist the account key material and they do not require the legacy manager to be unlocked. A SQL-backed wallet can therefore create an account that later cannot derive addresses, and it succeeds while the wallet is locked, unlike the existing `waddrmgr.NewAccount` contract. Please keep the wallet key-derivation path in `NewAccount` or extend the store API/implementation so derived account creation creates the same key material and lock checks. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -871,6 +1043,37 @@ func withScope(scope waddrmgr.KeyScope) filterOption {
...
+
+	return balances, nil
+}
+
+// accountInfoBalance returns the legacy UTXO balance for a store account row.
+func accountInfoBalance(balances scopedBalances,
+	info db.AccountInfo) btcutil.Amount {
+
+	accountNum := info.AccountNumber
+	if info.Origin == db.ImportedAccount {
```
yyforyongyu: Mapping every `db.ImportedAccount` to `waddrmgr.ImportedAddrAccount` only works for the legacy imported-key bucket. Watch-only xpub imports (`NewAccountWatchingOnly`) have `IsImported=true` but their UTXOs are recorded under the real scoped account number allocated by waddrmgr. Since `accountInfoFromProps` also turns these into account number 0, the balance overlay looks up the sentinel bucket and reports zero for imported xpub accounts with spendable outputs. Please preserve the legacy account number for xpub-imported accounts or distinguish the imported-key bucket before applying the sentinel rewrite. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -169,6 +167,179 @@ type AccountsResult struct {
 	CurrentBlockHeight int32
 }
 
+// accountPropertiesFromStoreInfo adapts one store account row to legacy wallet
+// account properties.
+func accountPropertiesFromStoreInfo(
+	info db.AccountInfo) waddrmgr.AccountProperties {
+
+	return waddrmgr.AccountProperties{
```
yyforyongyu: `accountPropertiesFromStoreInfo` drops derived-account fields that the previous wallet API returned from `waddrmgr.AccountProperties`: `AccountPubKey`, `MasterKeyFingerprint`, and per-account `AddrSchema`. The metadata merge below only runs for imported accounts, so `GetAccount`, `ListAccounts*`, and `NewAccount` regress those fields for normal derived accounts. Please use a full account-properties query/legacy merge for derived accounts too, or add those fields to the store result before switching wallet-facing APIs to `AccountInfo`. (`openai/gpt-5.5`)

yyforyongyu:

wallet/account_manager.go
```diff
@@ -618,118 +500,46 @@ func (w *Wallet) ImportAccount(ctx context.Context,
...
-	_, err := scopedMgr.NextExternalAddresses(ns, props.AccountNumber, 1)
+	err := validateExtendedPubKey(
+		accountKey, true, w.cfg.ChainParams,
+	)
 	if err != nil {
 		return nil, err
 	}
 
-	_, err = scopedMgr.NextInternalAddresses(ns, props.AccountNumber, 1)
+	keyScope, _, err := keyScopeFromPubKey(accountKey, &addrType)
```
yyforyongyu, line 519: `keyScope, _, err := keyScopeFromPubKey(accountKey, &addrType)` silently drops the per-account `*ScopeAddrSchema` that `keyScopeFromPubKey` returns. For BIP-0044 keys imported as `NestedWitnessPubKey`, `keyScopeFromPubKey` returns `(KeyScopeBIP0049Plus, &KeyScopeBIP0049AddrSchema, nil)` — i.e. force nested-everywhere instead of BIP0049Plus's default `{external=nested, internal=witness}`. The legacy path at `wallet/deprecated.go:6321` threads this schema into `scopedMgr.NewAccountWatchingOnly(ns, name, accountPubKey, fingerprint, addrSchema)` so the account row stores the override. The new path discards it, `CreateImportedAccountParams` no longer carries an `AddrSchema` field, and `kvdb.CreateImportedAccount` calls `PutWatchOnlyAccountWithKeys(..., addrSchema: nil)` — so a wallet that imports a legacy BIP-0044 NestedWitnessPubKey key now derives internal change addresses as witness pubkeys instead of nested. Either restore `AddrSchema` on `CreateImportedAccountParams` and pass it through, or document that the new path no longer supports BIP-0044 → BIP-0049 imports (and reject `addrType=NestedWitnessPubKey` on legacy keys before silently producing different change addresses). No test currently catches this — adding a `TestImportAccountBIP44Nested` that verifies the imported account's internal schema is `NestedWitnessPubKey` would lock the behavior in.

`claude-opus-4-7[1m]`

wallet/account_manager.go
```diff
@@ -114,37 +224,31 @@ var _ AccountManager = (*Wallet)(nil)
...
-	})
+	info, err := w.store.CreateDerivedAccount(ctx,
+		db.CreateDerivedAccountParams{
+			WalletID: w.id,
+			Scope:    toDBKeyScope(scope),
+			Name:     name,
+		}, deriveFn,
+	)
+	if err != nil {
+		return nil, err
```
yyforyongyu, line 244: This returns the store error directly, but the public `NewAccount` API still exposes `waddrmgr` semantics: the test below checks `waddrmgr.IsError(err, waddrmgr.ErrDuplicateAccount)`. The real store paths do not preserve that: kvdb wraps manager errors with `fmt.Errorf(... %w ...)`, and `waddrmgr.IsError` only matches an unwrapped `ManagerError`; the SQL stores return constraint errors. A duplicate account can therefore pass the mock test but fail the public error contract. Please translate duplicate/account-limit store errors back to the legacy `waddrmgr.ManagerError` surface before returning. (`openai/gpt-5.5`)

wallet/account_manager.go
yyforyongyu: Fixed across the stack:

- `db.CreateImportedAccountParams` now carries an `AddrSchema *ScopeAddrSchema` field (folded into PR #1231's `db: collapse AccountProperties into AccountInfo` commit, since it lives in the same `data_types.go` contract).
- `Wallet.ImportAccount` now captures the second return value from `keyScopeFromPubKey` via a new `waddrmgrAddrSchemaToDB` helper and threads it through.
- kvdb's `CreateImportedAccount` adapter (`createImportedAccountOps`) converts `params.AddrSchema` back to a `*waddrmgr.ScopeAddrSchema` via `dbAddrSchemaToWaddrmgr` and passes it to `PutWatchOnlyAccountWithKeys` instead of `nil`. So a BIP-0044 key imported as `NestedWitnessPubKey` again persists the nested-everywhere `KeyScopeBIP0049AddrSchema` rather than silently defaulting to the BIP0049Plus `external=nested, internal=witness` schema.

The SQL backends currently accept the field but don't persist the override — there is no per-account schema column yet in the SQL accounts schema. Tracking that for a separate PR; until then, kvdb-backed wallets preserve the legacy behavior and SQL-backed wallets fall back to the scope default for newly imported accounts of this shape.

A regression test (`TestImportAccountBIP44Nested`) for the imported account's internal schema is a good follow-up — I haven't added one in this revision to keep the diff focused on the threading; happy to do it in a fixup if you want it in this PR.

wallet/account_manager.go
yyforyongyu: Fixed by translating store errors back to the legacy `waddrmgr.ManagerError` surface before returning from the public `Wallet.NewAccount` and `Wallet.ImportAccount` paths:

```go
if err != nil {
    var mErr waddrmgr.ManagerError
    if errors.As(err, &mErr) {
        return nil, mErr
    }
    return nil, err
}
```

`errors.As` walks the `%w` chain that kvdb's `fmt.Errorf("put derived account: %w", err)` produces, so `waddrmgr.IsError(err, waddrmgr.ErrDuplicateAccount)` (which only matches an unwrapped `ManagerError`) keeps working on the real kvdb backend, not just the mock.

The SQL backends are still a gap here — they return wrapped constraint errors (`dberr.SQLError` with `Reason=Constraint`), which won't have a `ManagerError` in the chain. Translating those to `ManagerError` is real design work (do we collapse to legacy waddrmgr errors, or expose typed `db` errors and migrate callers off `waddrmgr.IsError`?), so I left it as-is for now. Until the SQL paths are wired into the public `NewAccount` API, the back-compat contract only needs to hold for kvdb.

yyforyongyu:
LGTM 🔑 (v5)

Diff vs v4 is two targeted fixes:
- The dropped `*ScopeAddrSchema` on imports is restored end-to-end via the new `waddrmgrAddrSchemaToDB` / `dbAddrSchemaToWaddrmgr` mirror pair (`account_manager.go:46-63`, `kvdb/accountstore.go:42-58`), and `CreateImportedAccountParams.AddrSchema` is now passed through to `PutWatchOnlyAccountWithKeys`. Closes the BIP-0044 NestedWitnessPubKey regression I flagged on v4.
- `NewAccount` / `importAccountInternal` now `errors.As(err, &waddrmgr.ManagerError{})` and return the bare ManagerError so legacy callers using `waddrmgr.IsError(err, code)` keep working after `translateAccountErr` wraps via `fmt.Errorf("%w", ...)` — necessary because `waddrmgr.IsError` is a direct type assertion (`waddrmgr/error.go:227`), not `errors.As`.

Minor observation, not a blocker: `RenameAccount` (and `Balance` / `GetAccount`) return the store error raw and would lose the same `waddrmgr.IsError` contract for non-NotFound ManagerError codes (e.g. duplicate-name rename). No production caller exercises that path, and the test would catch it once the contract is asserted — worth either applying the same unwrap or adding a `TestRenameAccountDuplicate` if/when a caller materializes.
(`claude-opus-4-7[1m]`)

wallet/account_manager.go
```diff
@@ -510,89 +471,35 @@ func (w *Wallet) RenameAccount(_ context.Context, scope waddrmgr.KeyScope,
...
 
-	return balance, nil
-}
-
-// balanceForUTXO is a helper function for Balance that calculates the balance
-// of a single UTXO if it belongs to the target account.
-func (w *Wallet) balanceForUTXO(addrmgrNs walletdb.ReadBucket,
-	scope waddrmgr.KeyScope, accNum uint32,
-	utxo wtxmgr.Credit) btcutil.Amount {
+	minConfs := int32(conf) //nolint:gosec
```
yyforyongyu: `conf` is a `uint32` in the public API, but this cast wraps values above `math.MaxInt32` negative before handing them to the store balance query. The SQL balance filters treat negative `min_confirms` as satisfied by every output, so a call like `Balance(ctx, math.MaxUint32, scope, name)` would now include all UTXOs instead of excluding them as the old `hasMinConfs(uint32, ...)` path did. Please reject or clamp values above the store/query range before constructing `BalanceParams`. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -169,228 +298,96 @@ type AccountsResult struct {
...
-
-	err = walletdb.View(w.cfg.DB, func(tx walletdb.ReadTx) error {
-		// First, calculate the balances for any accounts that match the
-		// given name. This is efficient as it iterates over the UTXO
-		// set, not accounts.
-		scopedBalances, err := w.fetchAccountBalances(tx)
-		if err != nil {
-			return err
-		}
+	return w.assembleAccountsResult(ctx, db.ListAccountsQuery{
```
yyforyongyu, line 334: This changes the unsupported-scope error surface for the public API. The old implementation called `w.addrStore.FetchScopedKeyManager(scope)` before listing, so callers saw the legacy `waddrmgr.ErrScopeNotFound` manager error. The new path delegates to `w.cache.ListAccounts`, and the kvdb adapter maps `ErrScopeNotFound` to `db.ErrAccountNotFound` in `selectScopedManagers`, so `ListAccountsByScope` can no longer distinguish an unknown key scope from an empty/missing account. Please translate this back to the legacy scope-not-found error before returning from the wallet API, as done for the other public account-manager compatibility paths. (`openai/gpt-5.5`)

wallet/account_manager.go
yyforyongyu: Fixed and pushed. The wallet balance path now rejects confirmation values that cannot be represented by the store query instead of letting the cast wrap negative.

wallet/account_manager.go
yyforyongyu: Fixed and pushed. `ListAccountsByScope` now preserves the legacy unsupported-scope error surface instead of translating it to account-not-found.

yyforyongyu:

yyforyongyu:
LGTM 🔑 (v6)

Key diff vs v5:
- `kvdb.loadAccountInfo` materializes the *effective* `AddrSchema` (`internal/db/kvdb/accountstore.go:396-419`): if waddrmgr did not persist a per-account override, the adapter looks the scope default up via `waddrmgr.ScopeAddrMap` and returns that. So both backends now return non-nil `AccountInfo.AddrSchema`, matching the new contract documented in #1231 v4.
- `waddrmgr.ScopedKeyManager` extracts the shared `fetchLastAccount → ++ → MaxAccountNum check → putLastAccount` body into `advanceLastAccountLocked` and routes both `AllocateDerivedAccountNumber` and `AllocateImportedAccountNumber` through it. Derived path keeps its watch-only precondition before the call.
- `ListAccountsByScope` now calls `w.addrStore.FetchScopedKeyManager(scope)` first so an unknown scope still surfaces `waddrmgr.ErrScopeNotFound` (verified by the new `TestListAccountsByScopeUnknownScope`).
- `Balance` guards `conf > math.MaxInt32` with a zero-balance early return (test `TestBalanceMaxConfsOverflow` pins the behavior).
(`claude-opus-4-7[1m]`)

wallet/account_manager.go
```diff
@@ -510,89 +476,38 @@ func (w *Wallet) RenameAccount(_ context.Context, scope waddrmgr.KeyScope,
...
-		// Ignore addresses that are not found in the wallet.
-		return 0
-	}
+	minConfs := int32(conf) //nolint:gosec
 
-	// If the address belongs to the target account, add the UTXO's value
-	// to the total balance.
-	if addrScope.Scope() == scope && addrAcc == accNum {
-		return utxo.Amount
+	result, err := w.store.Balance(ctx, db.BalanceParams{
```
yyforyongyu, line 462: This still appears unresolved at the current head. `Wallet.Balance` and `GetAccount` look up the account by `(scope, name)`, but then call `w.store.Balance` with only `Account: &info.AccountNumber`; `db.BalanceParams` has no scope/account-id field, and the SQL balance query filters only `acc.account_number`. Account numbers are reused across key scopes, so a wallet with the same account number in multiple scopes can still overcount cross-scope UTXOs. Please include scope (or a unique account ID) in the balance filter before routing these wallet-facing APIs through the store balance path. (`openai/gpt-5.5`)

yyforyongyu: Fixed across the stack:

- `db.BalanceParams` now carries an optional `Scope *KeyScope` field (folded into PR #1231's `db: collapse AccountProperties into AccountInfo` commit, where the params type lives).
- The SQL `Balance` queries (pg + sqlite) accept optional `purpose` / `coin_type` args and filter on `ks.purpose` / `ks.coin_type` alongside the existing `account_number` filter. sqlc regenerated.
- The pg + sqlite `Store.Balance` adapters unpack `params.Scope` via a new `scopeFilter` helper so a caller setting both Scope and Account gets the unambiguous per-scope balance.
- `Wallet.GetAccount` and `Wallet.Balance` now build a `dbScope := toDBKeyScope(scope)` and pass `Scope: &dbScope` alongside `Account: &info.AccountNumber`. Existing `TestGetAccount` and `TestBalance` mocks were updated to expect the new field.

A wallet with the same account number across BIP-0049 and BIP-0084 no longer overcounts cross-scope UTXOs on the SQL backends. kvdb's `Balance` is still `notImplemented` (so the legacy UTXO walk handles kvdb wallets), and `BalanceParams.Scope` defaults to nil for callers that genuinely want wallet-wide aggregation.

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -2,15 +2,45 @@ package kvdb
...
+// kvdb decoupled from waddrmgr.AddrStore's full breadth.
+type kvdbAddrStore interface {
+	// EncryptedMasterHDPriv reads the encrypted master HD private key
+	// from the manager's main bucket.
+	EncryptedMasterHDPriv(ns walletdb.ReadBucket) ([]byte, error)
+}
+
+// addrManager type-asserts s.addrStore into kvdbAddrStore. A nil store or a
+// wrong type indicates a caller-side wiring bug.
+func (s *Store) addrManager() (kvdbAddrStore, error) {
```
yyforyongyu: why do we need this? we can just do `s.addrStore` in the code?

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -197,11 +197,96 @@ func translateAccountErr(err error, notFound error) error {
 	return fmt.Errorf("waddrmgr: %w", err)
 }
 
-// ListAccounts is not yet implemented for kvdb.
-func (s *Store) ListAccounts(ctx context.Context,
-	_ db.ListAccountsQuery) ([]db.AccountInfo, error) {
+// ListAccounts returns the accounts matching the given query. When
+// query.Scope is nil, every active scoped key manager is walked.
+//
+//nolint:cyclop
```
yyforyongyu: fix it instead - code inside `ForEachAccount` can be a private helper, or a helper closure?

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -180,11 +181,112 @@ func (o *kvdbCreateDerivedAccountOps) CreateDerivedAccount(_ context.Context,
...
 }
 
-// CreateImportedAccount is not yet implemented for kvdb.
-func (s *Store) CreateImportedAccount(ctx context.Context,
-	_ db.CreateImportedAccountParams) (*db.AccountInfo, error) {
+// CreateImportedAccount persists an imported account into waddrmgr as a
+// watch-only row. kvdb does not persist an encrypted account private key
+// for imported accounts; spendable wallets that supply one get an error.
+//
+//nolint:cyclop
```
yyforyongyu: same here - fix the linter!!

wallet/cache.go
```diff
@@ -46,6 +49,13 @@ type runtimeCache interface {
 	// contract.
 	IterAddresses(ctx context.Context,
 		query db.ListAddressesQuery) iter.Seq2[db.AddressInfo, error]
+
+	// AccountBalances walks the legacy UTXO set and returns balances
+	// grouped by key scope and account number. scopeFilter narrows the
+	// walk to a single scope when non-nil. Used by the wallet account
+	// manager's list paths.
+	AccountBalances(ctx context.Context,
```
yyforyongyu: why do we cache the balances? it's always changing???

wallet/cache.go
```diff
@@ -157,5 +173,72 @@ func (c *storeRuntimeCache) IterAddresses(ctx context.Context,
...
+// from the legacy multi-account UTXO walk.
+type ScopedBalances map[waddrmgr.KeyScope]map[uint32]btcutil.Amount
+
+// AccountBalances walks the legacy UTXO set inside one walletdb.View and
+// builds a per-scope, per-account balance map. scopeFilter narrows the walk
+// to a single key scope; nil includes every scope.
+//
+// TODO(yy): replace with a SQL aggregation once the outputs table grows the
+// (key_scope, account_id) columns the kvdb-stack migration adds.
+func (c *storeRuntimeCache) AccountBalances(_ context.Context,
```
yyforyongyu: even if we wanna do this, this is not right - the cache methods are thin wrappers around the db methods for now!!!

wallet/account_manager.go
```diff
@@ -531,65 +531,45 @@ func (w *Wallet) ListAccountsByName(_ context.Context,
...
+func (w *Wallet) GetAccount(ctx context.Context, scope waddrmgr.KeyScope,
 	name string) (*AccountResult, error) {
 
 	err := w.state.validateStarted()
 	if err != nil {
 		return nil, err
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(scope)
+	info, err := w.cache.GetAccount(ctx, db.GetAccountQuery{
```
yyforyongyu, line 370: it just came to my mind - will the fields in account change?

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -56,7 +56,7 @@ func dbAddrSchemaToWaddrmgr(
 }
 
 // CreateDerivedAccount runs the shared CreateDerivedAccountWithOps workflow
-// on top of waddrmgr's bucket layout via the kvdbCreateDerivedAccountOps
+// on top of waddrmgr's bucket layout via the createDerivedAccountOps
```
yyforyongyu, line 44: the linter fix part should be folded into previous commits!!!

wallet/account_manager.go
```diff
@@ -510,89 +449,45 @@ func (w *Wallet) RenameAccount(_ context.Context, scope waddrmgr.KeyScope,
...
+	if conf > math.MaxInt32 {
+		return 0, nil
 	}
 
-	// Look up the account that owns the address.
-	addrScope, addrAcc, err := w.addrStore.AddrAccount(addrmgrNs, addr)
-	if err != nil {
-		// Ignore addresses that are not found in the wallet.
-		return 0
+	if conf > math.MaxInt32 {
```
yyforyongyu: Duplicate `if conf > math.MaxInt32 { return 0, nil }` block here (lines 469-471 then again 473-475). The second copy is dead code: the first guard already returns on the only branch the predicate matches. `git blame` shows the second copy landed in this PR's tip commit `249e2e1cd8 CI: fix lint findings + kvdb test restore` while the original guard was added in `e4bf9ab6a4 wallet: route Balance through Store` earlier in the same PR — looks like a rebase artifact rather than an intentional belt-and-suspenders. Drop one of the two.

`claude-opus-4-7[1m]`

wallet/account_manager.go
```diff
@@ -400,104 +373,70 @@ func (w *Wallet) ListAccountsByName(_ context.Context,
...
-		if balances, ok := scopedBalances[scope]; ok {
-			if balance, ok := balances[accNum]; ok {
-				account.TotalBalance = balance
-			}
-		}
+	minConfs := int32(0)
+	dbScope := toDBKeyScope(scope)
 
-		return nil
+	result, err := w.store.Balance(ctx, db.BalanceParams{
```
yyforyongyu, line 387: This still routes the public `GetAccount` balance overlay through `w.store.Balance`, and `Wallet.Balance` does the same below, but the live kvdb store still returns `kvdb.Store.Balance: not implemented`. Since the default wallet construction wires `w.store` to kvdb, both public APIs now error for kvdb wallets instead of using the legacy UTXO walk. Please either implement `kvdb.Store.Balance` here or keep these two methods on the `cache.AccountBalances`/legacy walk path until kvdb balance support exists. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -618,118 +513,54 @@ func (w *Wallet) ImportAccount(ctx context.Context,
...
 	name string, accountKey *hdkeychain.ExtendedKey,
 	masterKeyFingerprint uint32, addrType waddrmgr.AddressType,
 	dryRun bool) (*waddrmgr.AccountProperties, error) {
 
-	// Ensure we have a valid account public key. We require an account-level
-	// key (depth 3) to properly manage the derivation path.
-	err := validateExtendedPubKey(accountKey, true, w.cfg.ChainParams)
-	if err != nil {
-		return nil, err
+	if dryRun {
```
yyforyongyu, line 508: This breaks the existing public `ImportAccount(..., dryRun=true)` contract. The exported method comment still says dry-run validates without persisting, and callers that use it to preflight an xpub now get `errDryRunImportNotSupported` instead of duplicate-name/version/scope validation. Please preserve the dry-run path, either by running the store import in a rollback-only transaction or by keeping dry-runs on the legacy validation path until the store can simulate without persisting. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -618,118 +513,54 @@ func (w *Wallet) ImportAccount(ctx context.Context,
...
-	props *waddrmgr.AccountProperties, scopedMgr waddrmgr.AccountStore) (
-	*waddrmgr.AccountProperties, error) {
-
-	// The importAccount method above will cache the imported account within the
-	// scoped manager. Since this is a dry-run attempt, we'll want to invalidate
-	// the cache for it.
-	defer scopedMgr.InvalidateAccountCache(props.AccountNumber)
-
-	_, err := scopedMgr.NextExternalAddresses(ns, props.AccountNumber, 1)
+	keyScope, _, err := keyScopeFromPubKey(accountKey, &addrType)
```
yyforyongyu, line 519: `keyScopeFromPubKey` can return a non-default per-account address schema here, for example a BIP-0044 xpub imported as `NestedWitnessPubKey` needs the strict nested-everywhere BIP49 schema. This still discards that second return value, and `CreateImportedAccountParams` has no `AddrSchema`, so kvdb persists the account with `nil`/scope-default schema and SQL cannot preserve the override either. That changes internal/change derivation for these imported accounts and can miss the source wallet’s change UTXOs. Please thread the optional schema through the store params and into `PutWatchOnlyAccountWithKeys`/SQL persistence before routing imports through the store. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/walletstore.go
yyforyongyu: Fixed. The kvdb adapter now types `Store.addrStore` as `waddrmgr.AddrStore` directly — the narrow `addrStoreReader` interface, the `addrManager()` type-assert helper, and the `errKvdbAddrStoreTypeMismatch` sentinel are gone everywhere in the PR's history (cascade applied to every kvdb commit). The `waddrmgr: expose legacy adapter methods on public interfaces` commit at the bottom of the PR widens `AddrStore` with `EncryptedMasterHDPriv` + `Decrypt` and `AccountStore` with `ImportWitnessScript`, which is what made the direct typing possible.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed. `CreateImportedAccount` body is now decomposed into `putImportedAccount` (kvdb-side persistence under a caller-owned tx) and `scopedManagerOrCreate` (scope lookup with on-the-fly creation fallback). The function drops below the cyclop budget so the `//nolint:cyclop` marker is gone.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed. `CreateImportedAccount` body is now decomposed into `putImportedAccount` (kvdb-side persistence under a caller-owned tx) and `scopedManagerOrCreate` (scope lookup with on-the-fly creation fallback). The function drops below the cyclop budget so the `//nolint:cyclop` marker is gone.

wallet/cache.go
yyforyongyu: Fixed across prep + impl. `cache.AccountBalances` was removed entirely — the cache scaffolding commit on PR #1231 is now a thin pass-through over `db.Store` (no walletDB / addrStore / txStore fields). On the impl side, `wallet.assembleAccountsResult` reads balance directly off each `AccountInfo` (populated inline by the SQL aggregation in PR #1232) instead of caching it. The volatility concern goes away because balance is no longer cached at all on this layer.

wallet/cache.go
yyforyongyu: Fixed across prep + impl. `cache.AccountBalances` was removed entirely — the cache scaffolding commit on PR #1231 is now a thin pass-through over `db.Store` (no walletDB / addrStore / txStore fields). On the impl side, `wallet.assembleAccountsResult` reads balance directly off each `AccountInfo` (populated inline by the SQL aggregation in PR #1232) instead of caching it. The volatility concern goes away because balance is no longer cached at all on this layer.

wallet/account_manager.go
yyforyongyu: The stable durable set today is `CreateImportedAccountParams{WalletID, Name, Scope, MasterFingerprint, PublicKey}` — that's what waddrmgr persists in the watch-only row. `EncryptedPrivateKey` joins the params struct when a spendable-import path lands. The per-account `AddrSchema` override was dropped during the PR #1231 review on the BIP-0049Plus thread; the SQL backend no longer models per-account schema overrides, so it's not coming back as a field here.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed. The tail `CI: fix lint findings + kvdb test restore` commit is folded into the originating commits — `kvdbAddrStore → addrStoreReader` rename is moot (the narrow interface is gone in every commit), `kvdbCreateDerivedAccountOps → createDerivedAccountOps` rename + the watch-only mock additions land in the same commits that introduce those identifiers, and the wallet/account-manager lint fixes (Balance overflow guard, FetchScopedKeyManager precheck, math import) sit on their originating commits.

wallet/account_manager.go
yyforyongyu: Fixed in the refreshed `wallet: route Balance through Store` commit. The duplicate guard block is gone; the single `if conf > math.MaxInt32 { return 0, nil }` short-circuit remains before the `int32(conf)` narrowing.

yyforyongyu:

wallet/account_manager.go
```diff
@@ -510,89 +435,41 @@ func (w *Wallet) RenameAccount(_ context.Context, scope waddrmgr.KeyScope,
...
-				continue
-			}
-
-			balance += w.balanceForUTXO(
-				addrmgrNs, scope, accNum, utxo,
-			)
-		}
-
-		return nil
+	info, err := w.cache.GetAccount(ctx, db.GetAccountQuery{
```
yyforyongyu, line 446: `Wallet.Balance` calls `w.cache.GetAccount` without `SkipBalance: true`, only to use `info.AccountNumber` for the subsequent `w.store.Balance` lookup — the new `AccountBalance` SQL query that PR #1232 just added runs on every `Wallet.Balance` call and the resulting `ConfirmedBalance` / `UnconfirmedBalance` are discarded. Same shape on line 370 in `Wallet.GetAccount` (the info there is fed into `accountInfoToProperties`, which also doesn't read the balance fields). Setting `SkipBalance: true` on both `cache.GetAccount` calls skips the wasted UTXO aggregate per call; that opt-out exists exactly for this case (per the `SkipBalance` doc-comment in `data_types.go`).

(`claude-opus-4-7[1m]`)

wallet/account_manager.go
```diff
@@ -618,118 +495,54 @@ func (w *Wallet) ImportAccount(ctx context.Context,
...
 // importAccountInternal is the internal implementation of ImportAccount,
-// allowing callers (like Manager.Create) to bypass the started check.
+// allowing Manager.Create to bypass the started check.
 //
-// TODO(yy): we will move the db operation to a dedicated method, so we can
-// ignore cyclop for now.
-//
-//nolint:cyclop
-func (w *Wallet) importAccountInternal(_ context.Context,
+// dryRun is not supported on the new path; callers that need
```
yyforyongyu, line 500: `importAccountInternal`'s doc-comment is updated to say dry-run is unsupported, but the public-facing doc-comments on `Wallet.ImportAccount` (line 475) and the `AccountManager.ImportAccount` interface entry (line 197) still say:

> imports an account from an extended public **or private** key. ... If dryRun is true, the import is validated but not persisted.

Both claims are false on the new path:
- `validateExtendedPubKey` rejects private keys (`private keys cannot be imported`).
- `importAccountInternal` rejects `dryRun=true` with `errDryRunImportNotSupported`.

Worth updating both stale doc-comments so callers know to use `Wallet.ImportAccountDryRun` (deprecated) for simulate-and-rollback, and to pass only public extended keys.

(`claude-opus-4-7[1m]`)

wallet/account_manager.go
```diff
@@ -21,12 +21,111 @@ import (
...
+
+// errDryRunImportNotSupported is returned when callers try the simulate-and-
+// rollback dry-run on the new ImportAccount entrypoint.
+var errDryRunImportNotSupported = errors.New(
+	"ImportAccount: dry-run is not supported on the new account " +
+		"manager; use Wallet.ImportAccountDryRun (deprecated)",
+)
+
+// toDBKeyScope adapts a waddrmgr.KeyScope to the db.KeyScope contract.
+func toDBKeyScope(scope waddrmgr.KeyScope) db.KeyScope {
```
yyforyongyu, line 43: i don't think we need this thin helper method, remove it

PR #1216, opened 2026-04-14
wallet: cleanup import aliases

GustavoStingelin:
I think with the new package structure, we can simplify the import aliases a bit, relying on the default names in most cases.

also related to https://github.com/btcsuite/btcwallet/pull/1210#discussion_r3077710310


yyforyongyu (approved):
Nice!🪻

PR #1217, opened 2026-04-15
bwtest: pin bitcoind to v1 transport

yyforyongyu:
## Summary
- Pin the bitcoind integration harness to legacy P2P transport so it does not rely on the btcd miner's default transport mode during startup.
- Stabilize the `itest bitcoind (v30)` job that failed with the backend stuck at height 0 in CI.
- Context: failed CI run https://github.com/btcsuite/btcwallet/actions/runs/24432270704

## Testing
- `GOWORK=off make itest chain=bitcoind db=kvdb`


GustavoStingelin (approved):
Cool!

PR #1218, opened 2026-04-16
wallet: migrate signer to db.Store (superseded by #1243)

yyforyongyu:
**Superseded by #1230.**

The signer-side migration (`wallet: keep signer lookups wallet-side`) was folded into PR #1230 (`impl-tx-creator-store`) so the signer and tx-creator changes ship in one reviewable chain.

No content review needed here.


yyforyongyu (commented):
LGTM 🚀 (`openai/gpt-5.5`)

PR #1220, opened 2026-04-22
wallet: guard nil derived address callbacks

GustavoStingelin:
Prevent panics in derived address creation by rejecting a nil derivation callback and nil derived data return. Keep these guard errors internal and add focused regression tests.


yyforyongyu (approved):
🛡️

PR #1221, opened 2026-04-27
wallet: disambiguates watch-only

GustavoStingelin:
disambiguates watch-only
- Enforces wallet/account/address watch-only rules across PostgreSQL and SQLite.
- Treats script-only imported addresses as watch-only, while still preserving their encrypted scripts.
- Removes account-table watch-only state as an independent source of truth because account watch-only behavior is derived from wallet mode and imported-account semantics; duplicating it risks drift.
- Keeps the default imported account as a mixed container in non-watch-only wallets so it can hold both watch-only and spendable imported addresses.

The goal is to prevent SQL paths from accepting private-key-bearing data under watch-only wallets/accounts, without breaking valid script-only imports or mixed imported-address behavior in normal wallets.

Related to https://github.com/btcsuite/btcwallet/pull/1162#discussion_r2791907008

Next https://github.com/btcsuite/btcwallet/pull/1222

### DIFF
--- a/wallet/internal/db/accounts_common.go
+++ b/wallet/internal/db/accounts_common.go
@@ -43,6 +43,20 @@ func (params *CreateImportedAccountParams) Validate() error {
 	return nil
 }
 
+// ValidateWatchOnly validates watch-only invariants for creating an imported
+// account.
+func (params *CreateImportedAccountParams) ValidateWatchOnly(
+	walletIsWatchOnly bool) error {
+
+	hasPrivateKey := len(params.EncryptedPrivateKey) > 0
+	if walletIsWatchOnly && hasPrivateKey {
+		return fmt.Errorf("wallet %d cannot create account %q: %w",
+			params.WalletID, params.Name, ErrWatchOnlyViolation)
+	}
+
+	return nil
+}
+
 // Validate checks that exactly one account selector was set.
 func (query GetAccountQuery) Validate() error {
 	if query.Name == nil && query.AccountNumber == nil {
@@ -529,6 +543,7 @@ func RenameAccount[Args any](ctx context.Context,
 func CreateImportedAccount[CreateArgs any, CreateRow any, SecretArgs any](
 	ctx context.Context, params CreateImportedAccountParams,
 	ensureScope func() (int64, error),
+	walletWatchOnly func() (bool, error),
 	createAccount func(context.Context, CreateArgs) (CreateRow, error),
 	buildCreateArgs func(scopeID int64, isWatchOnly bool) CreateArgs,
 	rowToID func(CreateRow) int64,
@@ -542,7 +557,18 @@ func CreateImportedAccount[CreateArgs any, CreateRow any, SecretArgs any](
 		return nil, err
 	}
 
-	isWatchOnly := params.IsWatchOnly()
+	walletIsWatchOnly, err := walletWatchOnly()
+	if err != nil {
+		return nil, err
+	}
+
+	err = params.ValidateWatchOnly(walletIsWatchOnly)
+	if err != nil {
+		return nil, err
+	}
+
+	hasAccountSecret := len(params.EncryptedPrivateKey) > 0
+	isWatchOnly := walletIsWatchOnly || !hasAccountSecret
 
 	scopeID, err := ensureScope()
 	if err != nil {
@@ -558,7 +584,7 @@ func CreateImportedAccount[CreateArgs any, CreateRow any, SecretArgs any](
 
 	accountID := rowToID(row)
 
-	if isWatchOnly {
+	if !hasAccountSecret {
 		return getProps(accountID)
 	}
 
--- a/wallet/internal/db/accounts_common.go
+++ b/wallet/internal/db/accounts_common.go
@@ -30,8 +30,8 @@ func (params *CreateDerivedAccountParams) Validate() error {
 	return nil
 }
 
-// Validate validates required fields for creating an imported account.
-func (params *CreateImportedAccountParams) Validate() error {
+// ValidateBasic validates required fields for creating an imported account.
+func (params *CreateImportedAccountParams) ValidateBasic() error {
 	if params.Name == "" {
 		return ErrMissingAccountName
 	}
@@ -85,11 +85,6 @@ func (params RenameAccountParams) Validate() error {
 	}
 
 	return nil
-}
-
-// IsWatchOnly returns true if the account is watch-only.
-func (params *CreateImportedAccountParams) IsWatchOnly() bool {
-	return len(params.EncryptedPrivateKey) == 0
 }
 
 // AccountPropsRow represents the raw database fields needed to construct
@@ -538,21 +533,21 @@ func RenameAccount[Args any](ctx context.Context,
 
 // CreateImportedAccount is a generic helper that creates an imported account.
 // It handles ensuring the key scope exists, creating the account record,
-// optionally creating the account secret for non-watch-only accounts, and
-// fetching the full account properties from the database.
+// optionally creating the account secret when account private key material is
+// present, and fetching the full account properties from the database.
 func CreateImportedAccount[CreateArgs any, CreateRow any, SecretArgs any](
 	ctx context.Context, params CreateImportedAccountParams,
 	ensureScope func() (int64, error),
 	walletWatchOnly func() (bool, error),
 	createAccount func(context.Context, CreateArgs) (CreateRow, error),
-	buildCreateArgs func(scopeID int64, isWatchOnly bool) CreateArgs,
+	buildCreateArgs func(scopeID int64) CreateArgs,
 	rowToID func(CreateRow) int64,
 	createSecret func(context.Context, SecretArgs) error,
 	buildSecretArgs func(accountID int64) SecretArgs,
 	getProps func(accountID int64) (*AccountProperties, error),
 ) (*AccountProperties, error) {
 
-	err := params.Validate()
+	err := params.ValidateBasic()
 	if err != nil {
 		return nil, err
 	}
@@ -568,14 +563,13 @@ func CreateImportedAccount[CreateArgs any, CreateRow any, SecretArgs any](
 	}
 
 	hasAccountSecret := len(params.EncryptedPrivateKey) > 0
-	isWatchOnly := walletIsWatchOnly || !hasAccountSecret
 
 	scopeID, err := ensureScope()
 	if err != nil {
 		return nil, err
 	}
 
-	createArgs := buildCreateArgs(scopeID, isWatchOnly)
+	createArgs := buildCreateArgs(scopeID)
 
 	row, err := createAccount(ctx, createArgs)
 	if err != nil {
--- a/wallet/internal/db/addresses_common.go
+++ b/wallet/internal/db/addresses_common.go
@@ -13,6 +13,16 @@ import (
 // addresses.
 const DefaultImportedAccountName = "imported"
 
+// importedAddressIsWatchOnly determines whether an imported address is
+// watch-only based on wallet-level watch-only mode and address-level private
+// key presence. An imported address is watch-only if the wallet is watch-only
+// or if the address does not carry encrypted private key material.
+func importedAddressIsWatchOnly(walletIsWatchOnly bool,
+	hasPrivateKey bool) bool {
+
+	return walletIsWatchOnly || !hasPrivateKey
+}
+
 var (
 	// errNilAddressDerivationFunc is returned when derived address creation is
 	// called without a derivation callback.
@@ -89,9 +99,9 @@ func GetAddressSecret[Row any](ctx context.Context,
 	return nil, fmt.Errorf("get address secret: %w", err)
 }
 
-// Validate validates the required fields for creating an imported address.
-// Returns sentinel errors on failure.
-func (p NewImportedAddressParams) Validate() error {
+// ValidateBasic checks imported address creation parameters that do not depend
+// on external state.
+func (p NewImportedAddressParams) ValidateBasic() error {
 	if len(p.ScriptPubKey) == 0 {
 		return ErrMissingScriptPubKey
 	}
@@ -99,17 +109,32 @@ func (p NewImportedAddressParams) Validate() error {
 	return nil
 }
 
-// IsWatchOnly returns true if the params include neither a private key nor
-// a redeem or witness script.
-func (p NewImportedAddressParams) IsWatchOnly() bool {
-	noPrivKey := len(p.EncryptedPrivateKey) == 0
-	noScript := len(p.EncryptedScript) == 0
+// ValidateWatchOnly checks imported address creation parameters against the
+// parent wallet's watch-only state.
+func (p NewImportedAddressParams) ValidateWatchOnly(
+	walletIsWatchOnly bool) error {
 
-	if noPrivKey && noScript {
-		return true
+	if walletIsWatchOnly && p.HasPrivateKey() {
+		return fmt.Errorf("watch-only wallet %d cannot import private-key-"+
+			"bearing address into account %q: %w",
+			p.WalletID, DefaultImportedAccountName, ErrWatchOnlyViolation)
 	}
 
-	return false
+	return nil
+}
+
+// HasPrivateKey returns true if the params include private key material.
+// Script material is tracked separately and does not make an imported address
+// spend-capable by itself.
+func (p NewImportedAddressParams) HasPrivateKey() bool {
+	return len(p.EncryptedPrivateKey) > 0
+}
+
+// HasSecretMaterial returns true when the imported address carries any
+// encrypted secret payload that should be persisted in address_secrets.
+// Private keys and scripts are stored independently.
+func (p NewImportedAddressParams) HasSecretMaterial() bool {
+	return len(p.EncryptedPrivateKey) > 0 || len(p.EncryptedScript) > 0
 }
 
 // IDToOrigin safely converts an integer to AccountOrigin. It returns an error
@@ -138,6 +163,9 @@ type AddressInfoRow[TypeID, OriginIDType any] struct {
 	// OriginID is the database identifier for address origin (derived=0,
 	// imported=1).
 	OriginID OriginIDType
+
+	// WalletIsWatchOnly indicates whether the wallet is watch-only.
+	WalletIsWatchOnly bool
 
 	// HasPrivateKey indicates whether the address has an encrypted private key.
 	HasPrivateKey bool
@@ -222,13 +250,17 @@ func convertAddressIDs(id, accountID int64) (uint32, uint32, error) {
 }
 
 // newImportedAddressTx handles the shared transaction flow for creating an
-// imported address across database backends.
+// imported address across database backends. It creates the address row and
+// conditionally inserts the address_secrets row when any encrypted private key
+// or encrypted script material is present. Imported addresses are watch-only
+// when their parent wallet is watch-only or when the address itself does not
+// carry private key material.
 func newImportedAddressTx[QTX any, Row any, CreateArgs any, InsertArgs any](
 	ctx context.Context, create func(context.Context, CreateArgs) (Row, error),
 	createArgs CreateArgs,
 	insertFn func(QTX) func(context.Context, InsertArgs) error, qtx QTX,
 	insertArgs func(int64, NewImportedAddressParams) InsertArgs,
-	params NewImportedAddressParams, accountID int64,
+	params NewImportedAddressParams, accountID int64, walletIsWatchOnly bool,
 	rowID func(Row) int64, rowCreatedAt func(Row) time.Time) (*AddressInfo,
 	error) {
 
@@ -238,7 +270,7 @@ func newImportedAddressTx[QTX any, Row any, CreateArgs any, InsertArgs any](
 	}
 
 	addrID := rowID(addrRow)
-	if !params.IsWatchOnly() {
+	if params.HasSecretMaterial() {
 		err = insertFn(qtx)(ctx, insertArgs(addrID, params))
 		if err != nil {
 			return nil, fmt.Errorf("insert address secret: %w", err)
@@ -258,7 +290,9 @@ func newImportedAddressTx[QTX any, Row any, CreateArgs any, InsertArgs any](
 		Origin:       ImportedAccount,
 		ScriptPubKey: params.ScriptPubKey,
 		PubKey:       params.PubKey,
-		IsWatchOnly:  params.IsWatchOnly(),
+		IsWatchOnly: importedAddressIsWatchOnly(
+			walletIsWatchOnly, params.HasPrivateKey(),
+		),
 	}, nil
 }
 
@@ -314,6 +348,10 @@ func convertAddressPath(origin AccountOrigin, branch,
 
 // AddressRowToInfo converts raw database field values into an AddressInfo
 // struct. It handles type conversion and validation for each field.
+//
+// Watch-only computation: derived addresses follow wallet watch-only mode.
+// Imported addresses are watch-only when the wallet is watch-only or when the
+// imported address does not carry encrypted private key material.
 func AddressRowToInfo[TypeID, OriginIDType any](
 	row AddressInfoRow[TypeID, OriginIDType]) (*AddressInfo, error) {
 
@@ -334,8 +372,12 @@ func AddressRowToInfo[TypeID, OriginIDType any](
 		return nil, err
 	}
 
-	isWatchOnly := origin == ImportedAccount && !row.HasPrivateKey &&
-		!row.HasScript
+	isWatchOnly := row.WalletIsWatchOnly
+	if origin == ImportedAccount {
+		isWatchOnly = importedAddressIsWatchOnly(
+			row.WalletIsWatchOnly, row.HasPrivateKey,
+		)
+	}
 
 	return &AddressInfo{
 		ID:           id,
@@ -394,6 +436,10 @@ type DerivedAddressAdapters[QTX any, AccountRow any, AccountParams any,
 	// GetAccountID extracts the account ID from an account row.
 	GetAccountID func(AccountRow) int64
 
+	// GetWalletWatchOnly extracts the wallet watch-only state from an account
+	// row.
+	GetWalletWatchOnly func(AccountRow) bool
+
 	// GetExtIndex returns a function to get the external index.
 	GetExtIndex func(QTX) func(context.Context, int64) (int64, error)
 
@@ -425,6 +471,10 @@ type ImportedAddressAdapters[QTX any, AccountRow any,
 
 	// GetAccountID extracts the account ID from an account row.
 	GetAccountID func(AccountRow) int64
+
+	// GetWalletWatchOnly extracts the wallet-derived watch-only state from an
+	// account row.
+	GetWalletWatchOnly func(AccountRow) bool
 
 	// CreateAddr returns a function to create an address row.
 	CreateAddr func(QTX) func(context.Context, CreateArgs) (AddrRow, error)
@@ -465,6 +515,7 @@ func GetAddressByQuery(ctx context.Context, query GetAddressQuery,
 // inputs and then createFn to create the address.
 func createDerivedAddress[T any](ctx context.Context,
 	params NewDerivedAddressParams, walletID int64, accountID int64,
+	walletIsWatchOnly bool,
 	getExtIndex func(context.Context, int64) (int64, error),
 	getIntIndex func(context.Context, int64) (int64, error),
 	createFn func(context.Context, int64, int64, AddressType, uint32, uint32,
@@ -503,7 +554,7 @@ func createDerivedAddress[T any](ctx context.Context,
 		Branch:       branch,
 		Index:        index,
 		ScriptPubKey: scriptPubKey,
-		IsWatchOnly:  false,
+		IsWatchOnly:  walletIsWatchOnly,
 	}, nil
 }
 
@@ -588,8 +639,11 @@ func NewDerivedAddressWithTx[QTX any, AccountRow any,
 	err := executeTx(ctx, func(qtx QTX) error {
 		row, err := adapters.GetAccount(ctx, adapters.AccountParams(params))
 		if err == nil {
+			accountID := adapters.GetAccountID(row)
+
 			info, errAddr := createDerivedAddress(
-				ctx, params, int64(params.WalletID), adapters.GetAccountID(row),
+				ctx, params, int64(params.WalletID), accountID,
+				adapters.GetWalletWatchOnly(row),
 				adapters.GetExtIndex(qtx),
 				adapters.GetIntIndex(qtx),
 				adapters.CreateAddr(qtx),
@@ -630,7 +684,7 @@ func NewImportedAddressWithTx[QTX any, AccountRow any, AccountParams any,
 	adapters ImportedAddressAdapters[QTX, AccountRow, AccountParams, CreateArgs,
 		AddrRow, SecretParams]) (*AddressInfo, error) {
 
-	validationErr := params.Validate()
+	validationErr := params.ValidateBasic()
 	if validationErr != nil {
 		return nil, validationErr
 	}
@@ -640,13 +694,20 @@ func NewImportedAddressWithTx[QTX any, AccountRow any, AccountParams any,
 	err := executeTx(ctx, func(qtx QTX) error {
 		row, err := adapters.GetAccount(ctx, adapters.AccountParams(params))
 		if err == nil {
+			walletIsWatchOnly := adapters.GetWalletWatchOnly(row)
+
+			errWatchOnly := params.ValidateWatchOnly(walletIsWatchOnly)
+			if errWatchOnly != nil {
+				return errWatchOnly
+			}
+
 			acctID := adapters.GetAccountID(row)
 
 			info, errAddr := newImportedAddressTx(
 				ctx, adapters.CreateAddr(qtx),
 				adapters.CreateParams(int64(params.WalletID), acctID, params),
 				adapters.InsertSecret, qtx,
-				adapters.SecretParams, params, acctID,
+				adapters.SecretParams, params, acctID, walletIsWatchOnly,
 				adapters.RowID,
 				adapters.RowCreatedAt)
 			if errAddr != nil {
--- a/wallet/internal/db/common.go
+++ b/wallet/internal/db/common.go
@@ -1,0 +1,25 @@
+package db
+
+// NilIfEmptyBytes returns nil for nil or empty byte slices and returns
+// non-empty byte slices unchanged.
+//
+// Call this at the SQL boundary on nullable byte columns where "absent"
+// is encoded as SQL NULL. Go's []byte{} is non-nil but length-zero, and
+// database/sql sends it as an empty bytea/BLOB (not NULL), which
+// disagrees with two things:
+//
+//   - DB-side triggers and CHECK constraints that test `IS NULL` /
+//     `IS NOT NULL` (e.g. the watch-only secret triggers), and
+//   - Go-side "no value" checks that use `len(...) == 0` (e.g.
+//     HasPrivateKey).
+//
+// Normalizing empty to nil here picks one canonical storage shape —
+// empty means SQL NULL — so the two layers stay in agreement, and callers
+// can pass either nil or []byte{} without changing behavior.
+func NilIfEmptyBytes(b []byte) []byte {
+	if len(b) == 0 {
+		return nil
+	}
+
+	return b
+}
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -496,8 +496,10 @@ type AccountProperties struct {
 	// KeyScope is the key scope the account belongs to.
 	KeyScope KeyScope
 
-	// IsWatchOnly indicates whether the is set up as watch-only, i.e., it
-	// doesn't contain any private key information.
+	// IsWatchOnly indicates whether the account is in watch-only mode.
+	// Derived accounts inherit the wallet's watch-only status. Imported
+	// accounts are watch-only if the wallet is watch-only OR if the imported
+	// account has no private key material (account_secret is NULL).
 	IsWatchOnly bool
 
 	// CreatedAt is the timestamp when the account was created in the database.
@@ -620,8 +622,8 @@ type AddressInfo struct {
 	// addresses.
 	PubKey []byte
 
-	// IsWatchOnly indicates whether the wallet has the private key for this
-	// address. Convenience field.
+	// IsWatchOnly indicates whether the address belongs to a watch-only
+	// wallet or does not have private keys.
 	IsWatchOnly bool
 }
 
@@ -665,6 +667,10 @@ type NewDerivedAddressParams struct {
 // address into the wallet. All imported addresses are assigned to the
 // wallet imported account. The caller is responsible for encrypting any
 // sensitive material before populating this struct.
+//
+// Watch-only semantics: An imported address is watch-only if the parent wallet
+// is watch-only or if EncryptedPrivateKey is empty. EncryptedScript alone does
+// not make the imported address spendable.
 type NewImportedAddressParams struct {
 	// WalletID identifies the wallet that will own this address.
 	//
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -381,6 +381,9 @@ type AccountInfo struct {
 	UnconfirmedBalance btcutil.Amount
 
 	// IsWatchOnly indicates whether the account is in watch-only mode.
+	// Derived accounts inherit the wallet's watch-only status. Imported
+	// accounts are watch-only if the wallet is watch-only OR if the imported
+	// account has no private key material (account_secret is NULL).
 	IsWatchOnly bool
 
 	// CreatedAt is the timestamp when the account was created in the database.
@@ -445,7 +448,8 @@ type CreateImportedAccountParams struct {
 
 	// EncryptedPrivateKey is the encrypted extended private key for the
 	// account. This should be encrypted by the caller before being passed
-	// to the database layer. A nil or empty slice indicates watch-only.
+	// to the database layer. A nil or empty slice means no account private
+	// key material is stored; the imported account will be watch-only.
 	EncryptedPrivateKey []byte
 }
 
--- a/wallet/internal/db/itest/account_store_test.go
+++ b/wallet/internal/db/itest/account_store_test.go
@@ -366,6 +366,44 @@ func TestCreateImportedAccountErrors(t *testing.T) {
 	}
 }
 
+// TestCreateImportedAccountMissingWallet verifies that CreateImportedAccount
+// returns ErrWalletNotFound when the wallet does not exist.
+func TestCreateImportedAccountMissingWallet(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+
+	params := db.CreateImportedAccountParams{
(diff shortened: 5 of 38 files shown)


wallet/internal/db/wallets_common.go
```diff
@@ -1,5 +1,43 @@
...
+	}
+
+	if p.MasterKeyPrivParams == nil && p.EncryptedCryptoPrivKey == nil &&
+		p.EncryptedMasterPrivKey == nil {
+
+		return nil
+	}
+
+	return fmt.Errorf("watch-only wallet %q private secrets: %w", p.Name,
+		ErrWatchOnlyViolation)
```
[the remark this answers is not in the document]
GustavoStingelin: accepted.

wallet/internal/db/sqlite/accounts.go
```diff
@@ -206,15 +215,25 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+func getWalletWatchOnly(ctx context.Context, qtx *sqlc.Queries,
+	walletID uint32) (bool, error) {
+
+	row, err := qtx.GetWalletByID(ctx, int64(walletID))
+	if err != nil {
+		return false, fmt.Errorf("get wallet: %w", err)
+	}
+
+	return row.IsWatchOnly, nil
+}
```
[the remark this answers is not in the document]
GustavoStingelin: accepted.

wallet/internal/db/pg/accounts.go
```diff
@@ -180,15 +188,25 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+func getWalletWatchOnly(ctx context.Context, qtx *sqlc.Queries,
+	walletID uint32) (bool, error) {
+
+	row, err := qtx.GetWalletByID(ctx, int64(walletID))
+	if err != nil {
+		return false, fmt.Errorf("get wallet: %w", err)
+	}
+
+	return row.IsWatchOnly, nil
+}
```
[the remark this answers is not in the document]
GustavoStingelin: accepted.

wallet/internal/db/wallets_common.go
```diff
@@ -1,5 +1,43 @@
...
+	}
+
+	if p.MasterPrivParams == nil && p.EncryptedCryptoPrivKey == nil &&
+		p.EncryptedMasterHdPrivKey == nil {
+
+		return nil
+	}
+
+	return fmt.Errorf("watch-only wallet %d private secrets: %w", p.WalletID,
+		ErrWatchOnlyViolation)
```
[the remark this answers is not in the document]
GustavoStingelin: accepted.

wallet/internal/sql/sqlite/migrations/000005_accounts.up.sql
```diff
@@ -54,9 +54,6 @@ CREATE TABLE accounts (
     -- Reference to the origin of the account.
     origin_id INTEGER NOT NULL,
 
-    -- Defines if the account is watch-only.
-    is_watch_only BOOLEAN NOT NULL,
-
     -- Master fingerprint is the fingerprint of the master pub key that created
     -- this account.
     master_fingerprint INTEGER,
```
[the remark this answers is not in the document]
GustavoStingelin: intentional, not a problem for now.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -54,9 +54,6 @@ CREATE TABLE accounts (
     -- Reference to the origin of the account.
     origin_id SMALLINT NOT NULL,
 
-    -- Defines if the account is watch-only.
-    is_watch_only BOOLEAN NOT NULL,
-
     -- Master fingerprint is the fingerprint of the master pub key that created
     -- this account.
     master_fingerprint BIGINT,
```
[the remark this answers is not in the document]
GustavoStingelin: intentional, not a problem for now.

wallet/internal/sql/sqlite/migrations/000002_wallets.up.sql
```diff
@@ -50,7 +50,7 @@ CREATE TABLE wallet_secrets (
     encrypted_crypto_priv_key BLOB,
 
     -- Encrypted key used to encrypt/decrypt wallet data related to scripts.
-    -- NULL for watch-only wallets.
+    -- Watch-only wallets may still store this to protect imported scripts.
     encrypted_crypto_script_key BLOB,
```
GustavoStingelin, line 54: @yyforyongyu I think it might be a bit confusing to store a script that cannot actually spend funds in an encrypted field. This would still require unlocking the wallet just to view it.

If you agree, I can treat this in the next PRs as public data, similar to the public key and scriptPubKey, so it can be accessed without unlocking.

ps: talking about `address_secrets.encrypted_script` that needs this key.

wallet/internal/db/wallets_common.go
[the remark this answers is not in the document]
GustavoStingelin: removed to make the API clearer, and not accept empty slices in the DB.

wallet/internal/db/wallets_common.go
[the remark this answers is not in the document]
GustavoStingelin: removed to make the API clearer, and not accept empty slices in the DB.

wallet/internal/db/wallets_common.go
```diff
@@ -1,5 +1,43 @@
 package db
 
+import "fmt"
+
+// Validate checks wallet creation parameters for store-level invariants.
+// Watch-only wallets may keep the script crypto key so they can still encrypt
+// imported scripts, but they must not include private wallet secret material.
+func (p *CreateWalletParams) Validate() error {
```
yyforyongyu, line 8: The comment prefix `Watch-only wallets` does not match the exact method name `Validate`, which violates the Go Comment Gate rule to start with the exact declaration name. (`gemini-3.1-pro-preview`)

wallet/internal/db/wallets_common.go
```diff
@@ -1,5 +1,43 @@
...
+	}
+
+	return fmt.Errorf("watch-only wallet %q private secrets: %w", p.Name,
+		ErrWatchOnlyViolation)
+}
+
+// Validate checks wallet secret update parameters for store-level invariants.
+// Watch-only wallets may keep the script crypto key so they can still encrypt
+// imported scripts, but they must not include private wallet secret material.
+func (p *UpdateWalletSecretsParams) Validate(walletIsWatchOnly bool) error {
```
yyforyongyu, line 27: The comment prefix `Watch-only wallets` does not match the exact method name `Validate`, which violates the Go Comment Gate rule to start with the exact declaration name. (`gemini-3.1-pro-preview`)

yyforyongyu (commented):
~~The migration 000005_accounts.up.sql removes the `is_watch_only` column from the accounts table, but there is no migration to backfill existing rows. This means existing databases with accounts that have `is_watch_only = true` will break the schema when this migration runs. The column should be backfilled based on wallet-level watch-only status and account origin (imported accounts with no encrypted_private_key are watch-only).~~

yyforyongyu:
Found two issues that weren't addressed in prior review:

**1. wallet/internal/db/accounts_common.go:23** - The validation logic treats empty-but-non-nil private secret slices as "present" because it uses len(...) > 0 instead of len(...) != 0. While the PR body mentions removing empty slice acceptance from the DB layer, the validation logic in CreateImportedAccountParams.Validate still uses len() > 0 which would incorrectly reject legitimate empty slices. Consider normalizing to len(...) != 0 for consistency.

**2. wallet/internal/db/addresses_common.go:125** - NewImportedAddressParams.Validate has the same empty-slice issue: it checks HasPrivateKey() (which uses len(...) > 0) rather than treating empty slices as absent.

The migration issue was already noted by another reviewer. (MiniMax-M2.7)

wallet/internal/db/pg/accounts.go
```diff
@@ -292,6 +315,34 @@ type accountInfoRow interface {
...
+
+// derivedAddressGetAccountWatchOnly extracts the watch-only state from a row.
+func derivedAddressGetAccountWatchOnly(
+	row sqlc.GetAccountByWalletScopeAndNameRow) bool {
+
+	return row.IsWatchOnly
+}
+
+// importedAddressGetAccountID extracts the account ID from a row.
+func importedAddressGetAccountID(
```
yyforyongyu, line 336: The comment prefix `derivedAddressGetAccountID` does not match the exact method name `importedAddressGetAccountID`, which violates the Go Comment Gate rule to start with the exact declaration name. (`gemini-3.1-pro-preview`)

wallet/internal/db/pg/accounts.go
```diff
@@ -292,6 +315,34 @@ type accountInfoRow interface {
...
+	row sqlc.GetAccountByWalletScopeAndNameRow) int64 {
+
+	return row.ID
+}
+
+// importedAddressGetAccountWatchOnly extracts the watch-only state from a row.
```
yyforyongyu: The comment prefix `derivedAddressGetAccountWatchOnly` does not match the exact method name `importedAddressGetAccountWatchOnly`, which violates the Go Comment Gate rule to start with the exact declaration name. (`gemini-3.1-pro-preview`)

wallet/internal/db/sqlite/accounts.go
```diff
@@ -286,6 +310,34 @@ type accountInfoRow interface {
...
+
+// derivedAddressGetAccountWatchOnly extracts the watch-only state from a row.
+func derivedAddressGetAccountWatchOnly(
+	row sqlc.GetAccountByWalletScopeAndNameRow) bool {
+
+	return row.IsWatchOnly
+}
+
+// importedAddressGetAccountID extracts the account ID from a row.
+func importedAddressGetAccountID(
```
yyforyongyu, line 331: The comment prefix `derivedAddressGetAccountID` does not match the exact method name `importedAddressGetAccountID`, which violates the Go Comment Gate rule to start with the exact declaration name. (`gemini-3.1-pro-preview`)

wallet/internal/db/sqlite/accounts.go
```diff
@@ -286,6 +310,34 @@ type accountInfoRow interface {
...
+	row sqlc.GetAccountByWalletScopeAndNameRow) int64 {
+
+	return row.ID
+}
+
+// importedAddressGetAccountWatchOnly extracts the watch-only state from a row.
```
yyforyongyu: The comment prefix `derivedAddressGetAccountWatchOnly` does not match the exact method name `importedAddressGetAccountWatchOnly`, which violates the Go Comment Gate rule to start with the exact declaration name. (`gemini-3.1-pro-preview`)

wallet/internal/db/itest/fixtures_common_test.go
```diff
@@ -180,45 +181,48 @@ var ImportedAccountCases = []AccountTestCase{
...
+// accounts that can be created in non-watch-only wallets to hold imported
+// addresses.
+var PublicOnlyImportedAccountCases = []AccountTestCase{
 	{
-		Name:        "watchonly-bip84-cold",
+		Name:        "xpub-bip84-cold",
 		Scope:       db.KeyScopeBIP0084,
 		Origin:      db.ImportedAccount,
-		IsWatchOnly: true,
+		IsWatchOnly: false,
```
yyforyongyu: Doc says "public-only (xpub) accounts" / "without private keys", but with the new `ImportedParams` semantics (`if !tc.IsWatchOnly { params.EncryptedPrivateKey = RandomBytes(32) }`), `IsWatchOnly: false` actually populates a random private key. So these renamed fixtures now create accounts WITH private keys, contradicting the rename. Fix hint: set `IsWatchOnly: true` to match the documented "no private keys" intent.
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -54,9 +54,6 @@ CREATE TABLE accounts (
     -- Reference to the origin of the account.
     origin_id SMALLINT NOT NULL,
```
yyforyongyu, line 55: Existing migrations are being modified in-place (this file plus the matching sqlite `000005_accounts.up.sql` and the two `000002_wallets.up.sql` files), rather than adding a new `ALTER TABLE accounts DROP COLUMN is_watch_only` migration. Any environment that has already applied these migrations will not pick up the change because migration runners track `up` files by version number, not content; new installs will diverge from existing ones. Fix hint: if the `sql-wallet` branch has never been deployed, document that explicitly; otherwise add a follow-up migration.
(`claude-opus-4-7[1m]`)

wallet/internal/db/wallets_common.go
```diff
@@ -1,5 +1,43 @@
...
+
+// Validate checks wallet creation parameters for store-level invariants.
+// Watch-only wallets may keep the script crypto key so they can still encrypt
+// imported scripts, but they must not include private wallet secret material.
+func (p *CreateWalletParams) Validate() error {
+	if !p.IsWatchOnly {
+		return nil
+	}
+
+	if p.MasterKeyPrivParams == nil && p.EncryptedCryptoPrivKey == nil &&
```
yyforyongyu: Uses `== nil` rather than `len(...) == 0`, so an empty-but-non-nil `[]byte{}` is treated as "providing private secret material" and rejected with `ErrWatchOnlyViolation` (the test `create with empty-but-non-nil private secrets is rejected` confirms this). `nil` and `[]byte{}` are normally interchangeable in Go and frequently produced by JSON/protobuf decoding or struct copies; a caller that hands in `[]byte{}` to mean "no key" gets a confusing watch-only error. Fix hint: use `len(p.X) == 0` for all three fields, or document the nil-only contract explicitly on the field comments.
(`claude-opus-4-7[1m]`)

wallet/internal/db/wallets_common.go
```diff
@@ -1,5 +1,43 @@
...
+
+// Validate checks wallet secret update parameters for store-level invariants.
+// Watch-only wallets may keep the script crypto key so they can still encrypt
+// imported scripts, but they must not include private wallet secret material.
+func (p *UpdateWalletSecretsParams) Validate(walletIsWatchOnly bool) error {
+	if !walletIsWatchOnly {
+		return nil
+	}
+
+	if p.MasterPrivParams == nil && p.EncryptedCryptoPrivKey == nil &&
```
yyforyongyu: Same `== nil` vs `len() == 0` fragility as `CreateWalletParams.Validate`: a caller that updates one field and leaves the other private secret fields as zero-valued (`[]byte{}`) on a watch-only wallet will hit `ErrWatchOnlyViolation` even though no actual key bytes are being set.
(`claude-opus-4-7[1m]`)

wallet/internal/db/addresses_common.go
```diff
@@ -630,7 +678,7 @@ func NewImportedAddressWithTx[QTX any, AccountRow any, AccountParams any,
 	adapters ImportedAddressAdapters[QTX, AccountRow, AccountParams, CreateArgs,
 		AddrRow, SecretParams]) (*AddressInfo, error) {
 
-	validationErr := params.Validate()
+	validationErr := params.Validate(false)
```
yyforyongyu: `params.Validate(false)` is called outside the transaction (basic checks only) and then `params.Validate(accountIsWatchOnly)` is called again inside the transaction (basic + watch-only check). The `Validate` method is now overloaded with two responsibilities, distinguished only by a magic boolean. Fix hint: split into `ValidateBasic()` and `ValidateWatchOnly(accountIsWatchOnly bool)` so each call site asks for what it actually needs and the basic checks aren't re-run inside the transaction.
(`claude-opus-4-7[1m]`)

wallet/internal/db/data_types.go
```diff
@@ -661,6 +672,9 @@ type NewDerivedAddressParams struct {
 // address into the wallet. All imported addresses are assigned to the
 // wallet imported account. The caller is responsible for encrypting any
 // sensitive material before populating this struct.
+//
+// Watch-only semantics: An address is watch-only if EncryptedPrivateKey is
```
yyforyongyu: The "Watch-only semantics" comment only describes the `EncryptedPrivateKey`-empty rule, but `AddressRowToInfo` and `newImportedAddressTx` compute the address-level `IsWatchOnly` as `walletIsWatchOnly || !hasPrivateKey` (via `importedAddressIsWatchOnly`). A reader of this struct would conclude that a private-key-bearing imported address in a watch-only wallet is "not watch-only", but it is. Fix hint: extend the comment to mention the wallet-level rule (or just point at `importedAddressIsWatchOnly`).
(`claude-opus-4-7[1m]`)

wallet/internal/db/itest/address_store_test.go
```diff
@@ -375,27 +395,97 @@ func TestNewImportedAddressWithEncryptedScript(t *testing.T) {
...
+	}
+}
 
-			if tc.hasScript {
-				require.NoError(t, err)
-				require.Equal(t, params.EncryptedScript, secret.EncryptedScript)
+// TestNewImportedAddressRejectsPrivateKeysForWatchOnlyParents verifies that
+// private-key-bearing imported addresses are rejected when the wallet is
+// watch-only.
+func TestNewImportedAddressRejectsPrivateKeysForWatchOnlyParents(t *testing.T) {
```
yyforyongyu: Function name uses plural "Parents" but only `t.Run("wallet", ...)` is present; there is no separate account-level subtest because accounts no longer carry their own watch-only flag. Also missing here: a positive test for the case the PR description explicitly calls out ("Treats script-only imported addresses as watch-only, while still preserving their encrypted scripts") — i.e. that a watch-only wallet ACCEPTS an imported address with only `EncryptedScript` and no `EncryptedPrivateKey`. Fix hint: rename to singular and add the script-only acceptance subtest.
(`claude-opus-4-7[1m]`)

wallet/internal/db/accounts_common.go
```diff
@@ -498,32 +502,38 @@ func RenameAccount[Args any](ctx context.Context,
...
-	buildCreateArgs func(scopeID int64, isWatchOnly bool) CreateArgs,
+	buildCreateArgs func(scopeID int64) CreateArgs,
 	rowToID func(CreateRow) int64,
 	createSecret func(context.Context, SecretArgs) error,
 	buildSecretArgs func(accountID int64) SecretArgs,
 	getProps func(accountID int64) (*AccountProperties, error),
 ) (*AccountProperties, error) {
 
-	err := params.Validate()
+	walletIsWatchOnly, err := walletWatchOnly()
```
yyforyongyu, line 555: Resolving the wallet before the basic parameter checks changes the error for invalid input when the wallet is missing, for example an empty account name now returns `ErrWalletNotFound` instead of `ErrMissingAccountName` and performs a DB lookup. Run the name/public-key validation before fetching wallet mode, then apply only the watch-only private-key check after the wallet is resolved. (`openai/gpt-5.5`)

wallet/internal/db/addresses_common.go
```diff
@@ -238,7 +264,7 @@ func newImportedAddressTx[QTX any, Row any, CreateArgs any, InsertArgs any](
 	}
 
 	addrID := rowID(addrRow)
-	if !params.IsWatchOnly() {
+	if params.HasSecretMaterial() {
```
yyforyongyu, line 273: An empty-but-non-nil `EncryptedPrivateKey` plus a non-empty `EncryptedScript` reaches this branch, so the secret row stores a non-NULL empty private-key blob. Creation reports the address as watch-only because `HasPrivateKey` uses `len > 0`, but later reads use `encrypted_priv_key IS NOT NULL` and report it as spendable; reject or normalize empty private-key blobs before inserting the secret row. (`openai/gpt-5.5`)

wallet/internal/db/addresses_common.go
```diff
@@ -89,27 +100,35 @@ func GetAddressSecret[Row any](ctx context.Context,
 	return nil, fmt.Errorf("get address secret: %w", err)
 }
 
-// Validate validates the required fields for creating an imported address.
-// Returns sentinel errors on failure.
-func (p NewImportedAddressParams) Validate() error {
+// Validate checks imported address creation parameters for store-level
+// invariants. The parent account watch-only flag is resolved by the caller;
```
yyforyongyu: Comment claims watch-only inheritance is 'already reflected in the account row', but this PR removes the `is_watch_only` column from the accounts table. The account watch-only status is now derived from a JOIN with `wallets` at query time, not stored in the account row. Update this comment to match the new derivation model. (`k2p6`)

yyforyongyu (changes requested):
I've resolved the noisy comments made by agents, meanwhile, I think there is still a gap in determining the `watch-only` status. I like that we are moving this to the wallet-level, so everything inside a watch-only wallet is clearly watch-only, e.g., accounts and addresses. But when it's not watch-only, it's blurry.

so i think the higher layer should dictate the watch-only status, wallet->account->address, and only when there is no higher layer, like an imported account, or an imported address, we then check for the secrets like xpri or private keys:

if wallet is watch-only:
    - account/address is watch-only

otherwise, 

for accounts:
    - derived account: not watch-only, because wallet seed controls it
    - imported account: watch-only iff no imported account private key/xpriv

for addresses:
    - derived address: inherits account status
    - imported address: watch-only iff no address private key

So “missing secrets” should only decide watch-only status at the layer where key material can independently exist:
- Imported account: check account-level private key/xpriv.
- Imported address: check address-level private key.
- Derived account/address: inherit from wallet/account, because their private keys are derived from the wallet/account key hierarchy.

wallet/internal/db/itest/account_store_test.go
```diff
@@ -354,6 +372,25 @@ func TestCreateImportedAccountErrors(t *testing.T) {
 	}
 }
 
+// TestCreateImportedAccountMissingWallet verifies that CreateImportedAccount
+// returns ErrWalletNotFound when the wallet does not exist.
+func TestCreateImportedAccountMissingWallet(t *testing.T) {
```
yyforyongyu, line 389: `TestCreateImportedAccountErrors` (ending just above this new test) does not cover the watch-only rejection path. Add a test case that creates a watch-only wallet and then attempts to create an imported account with `EncryptedPrivateKey`, asserting `ErrWatchOnlyViolation`. (`k2p6`)

wallet/internal/db/itest/address_store_test.go
yyforyongyu: `TestNewImportedAddressRejectsPrivateKeysForWatchOnlyParents` only tests rejection of private-key-bearing addresses. It does not verify the positive case: script-only imported addresses (with `EncryptedScript` but no `EncryptedPrivateKey`) should succeed in watch-only wallets and be marked `IsWatchOnly=true`. Add a subtest or separate test for this behavior. (`k2p6`)

wallet/internal/sql/sqlite/migrations/000005_accounts.up.sql
```diff
@@ -54,9 +54,6 @@ CREATE TABLE accounts (
     -- Reference to the origin of the account.
     origin_id INTEGER NOT NULL,
 
-    -- Defines if the account is watch-only.
-    is_watch_only BOOLEAN NOT NULL,
```
yyforyongyu, line 58: Modifying an existing migration breaks databases that already applied the old version. The `is_watch_only` column remains as `NOT NULL` without a default in existing databases, so new INSERT queries that omit it will fail. Add a new migration (e.g., `00000X_drop_accounts_is_watch_only.up.sql`) to drop the column instead of editing this file. (`k2p6`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -54,9 +54,6 @@ CREATE TABLE accounts (
...
 
-    -- Defines if the account is watch-only.
-    is_watch_only BOOLEAN NOT NULL,
```
yyforyongyu, line 58: Same migration compatibility issue as SQLite: modifying an existing migration breaks existing PostgreSQL databases that already have the `is_watch_only` column. Use a new migration to drop the column rather than editing this file. (`k2p6`)

wallet/internal/sql/sqlite/migrations/000005_accounts.up.sql
GustavoStingelin: not a problem as we dont haven't released yet.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
GustavoStingelin: not a problem as we dont haven't released yet.

GustavoStingelin:
Added `TestWatchOnlyHierarchyAccountRules` and `TestWatchOnlyHierarchyAddressRules` to make the expected behavior clearer.

yyforyongyu:

yyforyongyu:
LGTM ✨
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu, line 58: Modifying an already-applied migration file breaks existing databases. PostgreSQL will raise a NOT NULL violation on the next insert because existing tables still have this column. Use a new migration file (e.g. `00000X_alter_accounts_drop_watch_only.sql`) instead.

`k2p6`

wallet/internal/sql/sqlite/migrations/000005_accounts.up.sql
yyforyongyu, line 58: Same migration-compatibility issue: altering an already-applied migration file breaks existing SQLite databases. Add a new migration that drops the column instead.

`k2p6`

wallet/internal/sql/pg/migrations/000002_wallets.up.sql
```diff
@@ -35,8 +35,8 @@ CREATE UNIQUE INDEX uidx_wallets_name ON wallets (wallet_name);
 -- Wallet Secrets table to store rarely accessed, highly sensitive encrypted
 -- material with a strict one-to-one relationship with the wallets table.
 -- Separated from the main wallets table for security and access pattern isolation.
--- Watch-only wallets may have no corresponding row in this table or have all
```
yyforyongyu, line 38: Even comment-only changes to applied migration files can break migration tools that track file checksums. Consider leaving migration files immutable and documenting behavior changes elsewhere.

`k2p6`

wallet/internal/sql/sqlite/migrations/000002_wallets.up.sql
```diff
@@ -35,8 +35,8 @@ CREATE UNIQUE INDEX uidx_wallets_name ON wallets (wallet_name);
 -- Wallet Secrets table to store rarely accessed, highly sensitive encrypted
 -- material with a strict one-to-one relationship with the wallets table.
 -- Separated from the main wallets table for security and access pattern isolation.
--- Watch-only wallets may have no corresponding row in this table or have all
```
yyforyongyu, line 38: Same concern: modifying applied migration files may break checksum-based migration tooling. Keep migration files immutable.

`k2p6`

wallet/internal/db/data_types.go
```diff
@@ -381,6 +381,11 @@ type AccountInfo struct {
 	UnconfirmedBalance btcutil.Amount
 
 	// IsWatchOnly indicates whether the account is in watch-only mode.
+	// For accounts in this SQL store, the value follows the wallet-level
+	// watch-only mode: derived accounts inherit wallet watch-only status,
+	// and imported accounts are watch-only if the wallet is watch-only.
```
yyforyongyu: This comment is incomplete: imported accounts are also watch-only when they have no account secret (no private key), even in a non-watch-only wallet. The SQL query computes `is_watch_only` as `w.is_watch_only OR (a.origin_id = 1 AND acs.account_id IS NULL)`.

`k2p6`

wallet/internal/db/data_types.go
```diff
@@ -492,8 +499,12 @@ type AccountProperties struct {
 	// KeyScope is the key scope the account belongs to.
 	KeyScope KeyScope
 
-	// IsWatchOnly indicates whether the is set up as watch-only, i.e., it
-	// doesn't contain any private key information.
+	// IsWatchOnly indicates whether the account is in watch-only mode.
+	// For accounts in this SQL store, the value follows the wallet-level
```
yyforyongyu: Same incomplete comment: imported accounts without secrets are watch-only regardless of wallet mode.

`k2p6`

wallet/internal/db/data_types.go
```diff
@@ -445,7 +450,9 @@ type CreateImportedAccountParams struct {
 
 	// EncryptedPrivateKey is the encrypted extended private key for the
 	// account. This should be encrypted by the caller before being passed
-	// to the database layer. A nil or empty slice indicates watch-only.
+	// to the database layer. A nil or empty slice means no account private
+	// key material is stored; the account's watch-only status follows
+	// wallet-level watch-only mode.
```
yyforyongyu: This documents an empty account private key as following only wallet-level watch-only mode, and the same wallet-only wording appears on `AccountInfo.IsWatchOnly`/`AccountProperties.IsWatchOnly`. The SQL returns imported accounts without an `account_secrets` row as watch-only even for standard wallets, so the exported docs contradict the actual result. Please describe imported accounts as watch-only when the wallet is watch-only or no account private key is stored. (`openai/gpt-5.5`)

wallet/internal/db/itest/account_store_test.go
```diff
@@ -420,6 +643,40 @@ func TestGetAccount(t *testing.T) {
...
+func TestGetAccountWatchOnlyMapping(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-get-watch")
+	scope := db.KeyScopeBIP0084
+
+	createDerivedAccount(t, store, walletID, scope, "derived")
+
+	_, err := store.CreateImportedAccount(t.Context(), db.CreateImportedAccountParams{
```
yyforyongyu: nit: should be in a newline?

wallet/internal/db/itest/account_store_test.go
```diff
@@ -539,6 +796,44 @@ func TestListAccounts(t *testing.T) {
...
+func TestListAccountsWatchOnlyMapping(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-list-watch")
+	scope := db.KeyScopeBIP0084
+
+	createDerivedAccount(t, store, walletID, scope, "derived")
+
+	_, err := store.CreateImportedAccount(t.Context(), db.CreateImportedAccountParams{
```
yyforyongyu: nit: should be using a newline

wallet/internal/db/itest/account_store_test.go
```diff
@@ -539,6 +796,44 @@ func TestListAccounts(t *testing.T) {
...
+		WalletID:           walletID,
+		Name:               db.DefaultImportedAccountName,
+		Scope:              scope,
+		EncryptedPublicKey: RandomBytes(32),
+	})
+	require.NoError(t, err)
+
+	accounts, err := store.ListAccounts(t.Context(), db.ListAccountsQuery{
```
yyforyongyu: same here, should be in newline, should follow `@docs/developer/code_formatting_rules.md`

wallet/internal/db/itest/account_store_test.go
```diff
@@ -539,6 +796,44 @@ func TestListAccounts(t *testing.T) {
...
+		WalletID: walletID,
+		Scope:    &scope,
+	})
+	require.NoError(t, err)
+
+	derived := findAccountInList(t, accounts, AccountTestCase{
```
yyforyongyu: same here re the formatting: @docs/developer/code_formatting_rules.md and a few other places

wallet/internal/db/common.go
```diff
@@ -0,0 +1,11 @@
+package db
+
+// NilIfEmptyBytes returns nil for nil or empty byte slices and returns
+// non-empty byte slices unchanged.
+func NilIfEmptyBytes(b []byte) []byte {
```
yyforyongyu, line 19: let's add more detailed docs, sth like
```go
// NilIfEmptyBytes returns nil for nil or empty byte slices and returns
// non-empty byte slices unchanged.
//
// Call this at the SQL boundary on nullable byte columns where "absent"
// is encoded as SQL NULL. Go's []byte{} is non-nil but length-zero, and
// database/sql sends it as an empty bytea/BLOB (not NULL), which
// disagrees with two things:
//
//   - DB-side triggers and CHECK constraints that test `IS NULL` /
//     `IS NOT NULL` (e.g. the watch-only secret triggers), and
//   - Go-side "no value" checks that use `len(...) == 0` (e.g.
//     HasPrivateKey).
//
// Normalizing empty to nil here picks one canonical storage shape —
// empty means SQL NULL — so the two layers stay in agreement and callers
// can pass either nil or []byte{} without changing behavior.
func NilIfEmptyBytes(b []byte) []byte {
```

Kinda need this as I've already forgotten why we need this method after a week😮‍💨

wallet/internal/db/itest/account_store_test.go
```diff
@@ -354,6 +372,211 @@ func TestCreateImportedAccountErrors(t *testing.T) {
...
+				)
+				if err != nil {
+					return false, err
+				}
+
+				return info.IsWatchOnly, nil
+			},
+		},
+		{
+			name:          "standard wallet imported account with private key is spendable",
```
yyforyongyu: looks like the linter is not checking for these files, likely due to build flags, as clearly this should trigger `lll` error, will create a PR to fix it

wallet/internal/db/itest/account_store_test.go
```diff
@@ -354,6 +372,211 @@ func TestCreateImportedAccountErrors(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			store := NewTestStore(t)
+
+			walletInfo, err := store.CreateWallet(
+				t.Context(), tc.walletParams("watch-only-account-matrix"),
+			)
+			require.NoError(t, err)
+
+			isWatchOnly, err := tc.createAccountFn(t, store, walletInfo.ID)
+			if tc.wantErr != nil {
```
yyforyongyu, line 593: nit: there is no need to check `if tc.wantErr != nil {`, using `require.ErrorIs(t, err, tc.wantErr)` is enough - don't know why but `gpt` keeps writing it this way.

wallet/internal/db/itest/fixtures_common_test.go
```diff
@@ -100,8 +100,12 @@ type AccountTestCase struct {
 	// Origin indicates whether the account is derived or imported.
 	Origin db.AccountOrigin
 
-	// IsWatchOnly indicates whether the account has no private key.
-	// Only relevant for imported accounts.
+	// HasPrivateKey indicates whether ImportedParams should include encrypted
+	// private key material for imported accounts.
+	HasPrivateKey bool
```
yyforyongyu: feels redudant given that this is just `!IsWatchOnly`?

wallet/internal/db/addresses_common.go
```diff
@@ -13,6 +13,17 @@ import (
 // addresses.
 const DefaultImportedAccountName = "imported"
 
+// importedAddressIsWatchOnly determines whether an imported address is
+// watch-only based on wallet-level watch-only mode and address-level private
+// key presence. An imported address is watch-only if the wallet is watch-only
+// or if the address does not carry encrypted private key material.
+func importedAddressIsWatchOnly(
```
yyforyongyu: nit: can be fit in one line?

wallet/internal/db/addresses_common.go
```diff
@@ -89,27 +100,43 @@ func GetAddressSecret[Row any](ctx context.Context,
...
-// IsWatchOnly returns true if the params include neither a private key nor
-// a redeem or witness script.
-func (p NewImportedAddressParams) IsWatchOnly() bool {
-	noPrivKey := len(p.EncryptedPrivateKey) == 0
-	noScript := len(p.EncryptedScript) == 0
+// ValidateWatchOnly checks imported address creation parameters against the
+// parent wallet's watch-only state.
+func (p NewImportedAddressParams) ValidateWatchOnly(
+	walletIsWatchOnly bool,
+) error {
```
yyforyongyu: nit: we usually prefer sth like
```go
func (p NewImportedAddressParams) ValidateWatchOnly(
	walletIsWatchOnly bool) error {
```

wallet/internal/sql/sqlite/migrations/000002_wallets.up.sql
yyforyongyu: Very tempting tbh, as we can remove methods like `HasSecretMaterial` if we go that route.

However, given that the `script` itself can be viewed as the secret material, I'd say let's encrypt it, so keep the current behavior unchanged.

wallet/internal/sql/sqlite/queries/addresses.sql
```diff
@@ -86,10 +89,12 @@ SELECT
     a.pub_key,
     a.created_at,
     acc.origin_id,
+    w.is_watch_only AS wallet_is_watch_only,
     s.encrypted_priv_key IS NOT NULL AS has_private_key,
     s.encrypted_script IS NOT NULL AS has_script
 FROM addresses AS a
 INNER JOIN accounts AS acc ON a.account_id = acc.id
+INNER JOIN wallets AS w ON a.wallet_id = w.id
```
yyforyongyu, line 97: this commit is rather large, to better review it, I think we can follow claude's advice:

A clean 3-commit split:

**Commit 1: `wallet: enforce watch-only invariant on imported account creation`**
- `accounts_common.go`: split `Validate()` into `ValidateBasic()` + `ValidateWatchOnly(walletIsWatchOnly bool)`; drop `IsWatchOnly()`.
- `accounts_common.go`: add `walletWatchOnly func() (bool, error)` callback to the `CreateImportedAccount` generic helper; call `ValidateBasic` → `walletWatchOnly()` → `ValidateWatchOnly` in that order.
- `pg/accounts.go` + `sqlite/accounts.go`: add `getWalletWatchOnly` helper; wire it through `CreateImportedAccount` (and `CreateDerivedAccount` for the missing-wallet error path).
- Tests: `TestCreateDerivedAccountMissingWallet`, `TestCreateImportedAccountMissingWallet`, `TestCreateImportedAccountValidationPrecedesWalletLookup`.

The schema is unchanged. `buildCreateImportedAccountArgs` still takes `isWatchOnly bool` and the column is still populated from `len(EncryptedPrivateKey)==0` exactly as before. The only behavior change is that watch-only wallets now reject private-key-bearing imported accounts with `ErrWatchOnlyViolation` — the smallest possible commit that adds the invariant.

**Commit 2: `wallet: derive account watch-only from wallet at read time`**
- Migration: drop `accounts.is_watch_only`.
- `accounts.sql` INSERT queries: remove the column.
- `accounts.sql` SELECT queries: `INNER JOIN wallets AS w`, `LEFT JOIN account_secrets AS acs`, project `CASE WHEN w.is_watch_only THEN TRUE WHEN a.origin_id=1 AND acs.account_id IS NULL THEN TRUE ELSE FALSE END AS is_watch_only`.
- sqlc regen.
- `pg/accounts.go` + `sqlite/accounts.go`: drop the `isWatchOnly` arg from `buildCreateImportedAccountArgs`; `CreateDerivedAccount` passes `walletIsWatchOnly` (already available from commit 1) to `BuildAccountInfo`.
- Fixture refactor: rename `WatchOnlyAccountCases` → `PublicOnlyImportedAccountCases`, add `HasPrivateKey` field, update `ImportedParams` to use it; update existing test expectations that used the old "no priv key ⇒ watch-only" semantics.

This is the schema/data-model commit: the value moves from a stored column to a derived expression. Everything in this commit is forced together by the migration.

**Commit 3: `wallet: add canonical watch-only hierarchy tests for accounts`**
- `TestWatchOnlyHierarchyAccountRules` (the matrix test).
- `TestGetAccountWatchOnlyMapping`, `TestListAccountsWatchOnlyMapping`.

These tests pin down the new derivation model — they make sense to land after the model exists.

**Why this split:**
- Commit 1 has one purpose (introduce the invariant) and is reviewable in isolation; the schema is untouched so reviewers can ignore the migrations.
- Commit 2 is the meaty bit — the data-model change. Reviewers focused on SQL only need to read this commit. The Go-side workflow plumbing they'd need is already in place from commit 1.
- Commit 3 is pure test additions; bisects nicely if a behavior regression shows up later.

yyforyongyu (changes requested):
Approach ack, the rest is mostly on the code quality.

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: I think I got it now. 😄

I was reading the rule as applying only when the unwrapped line exceeds the column limit:

> When wrapping a line that contains a function call as the unwrapped line exceeds the column limit, the close parenthesis should be placed on its own line.

Maybe it would be worth adding a few more examples to the docs, or clarifying that this also applies in cases like this?

yyforyongyu:

yyforyongyu:
LGTM 🪞
(`claude-opus-4-7[1m]`)

wallet/internal/db/itest/account_store_test.go
yyforyongyu: hmm i think we have examples there, but agents keep ignoring them

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: Was just grouping the error with the early return. Splited now.

wallet/internal/db/common.go
GustavoStingelin: added

wallet/internal/db/itest/fixtures_common_test.go
GustavoStingelin: removed

wallet/internal/db/addresses_common.go
GustavoStingelin: can fit two lines, updated

wallet/internal/db/addresses_common.go
GustavoStingelin: fixed

wallet/internal/sql/sqlite/queries/addresses.sql
GustavoStingelin: splitted

wallet/internal/db/data_types.go
GustavoStingelin: updated

wallet/internal/db/data_types.go
GustavoStingelin: updated

wallet/internal/db/data_types.go
GustavoStingelin: updated

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: added

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: added

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: btw, updated that line

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: fixed

wallet/internal/db/itest/account_store_test.go
GustavoStingelin: rebased  with the new linter rules

yyforyongyu:
LGTM 🪞 (re-review after rebase on #1228)
(`claude-opus-4-7[1m]`)

yyforyongyu (approved):

PR #1222, opened 2026-04-29
wallet: enforce watch-only invariants in sql

GustavoStingelin:
Strengthen the database-boundary enforcement for watch-only data by adding trigger coverage and regression itests. This also prevents key scopes, accounts, and addresses from being reparented to a different wallet after insert, closing a bypass around existing secret invariants.

### DIFF
--- a/wallet/internal/db/itest/account_store_test.go
+++ b/wallet/internal/db/itest/account_store_test.go
@@ -100,6 +100,54 @@ func TestCreateImportedAccountRejectsWalletScopeMismatch(t *testing.T) {
 	require.ErrorContains(t, err, "constraint")
 }
 
+// TestAccountWalletIDImmutable verifies that raw account reparenting updates
+// cannot change wallet ownership after insert.
+func TestAccountWalletIDImmutable(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	queries := store.Queries()
+
+	sourceWalletID := newWallet(t, store, "account-wallet-immutable-source")
+	targetWalletID := newWallet(t, store, "account-wallet-immutable-target")
+
+	CreateImportedAccount(
+		t, store, sourceWalletID, db.KeyScopeBIP0084, "source-imported",
+	)
+	createDerivedAccount(
+		t, store, targetWalletID, db.KeyScopeBIP0084, "target-derived",
+	)
+
+	sourceScopeID := GetKeyScopeID(
+		t, queries, sourceWalletID, db.KeyScopeBIP0084,
+	)
+	targetScopeID := GetKeyScopeID(
+		t, queries, targetWalletID, db.KeyScopeBIP0084,
+	)
+	sourceAccountID := GetAccountID(
+		t, queries, sourceScopeID, "source-imported",
+	)
+
+	err := reparentAccountRaw(
+		t, store.DB(), sourceAccountID, targetWalletID, targetScopeID,
+	)
+	require.Error(t, err)
+	requireDriverConstraintError(t, err)
+
+	accountInfo, err := store.GetAccount(
+		t.Context(), getAccountQueryByName(
+			sourceWalletID, db.KeyScopeBIP0084, "source-imported",
+		),
+	)
+	require.NoError(t, err)
+	require.NotNil(t, accountInfo)
+	require.Equal(t, "source-imported", accountInfo.AccountName)
+	require.Equal(
+		t, sourceAccountID,
+		GetAccountID(t, queries, sourceScopeID, "source-imported"),
+	)
+}
+
 // TestWatchOnlyAccountSecretTriggers verifies that account_secrets rejects
 // watch-only parent accounts while still allowing inserts and updates for
 // non-watch-only parents.
--- a/wallet/internal/db/itest/account_store_test.go
+++ b/wallet/internal/db/itest/account_store_test.go
@@ -100,6 +100,116 @@ func TestCreateImportedAccountRejectsWalletScopeMismatch(t *testing.T) {
 	require.ErrorContains(t, err, "constraint")
 }
 
+// TestWatchOnlyAccountSecretTriggers verifies that account_secrets rejects
+// watch-only parent accounts while still allowing inserts and updates for
+// non-watch-only parents.
+func TestWatchOnlyAccountSecretTriggers(t *testing.T) {
+	t.Parallel()
+
+	t.Run("watch-only insert is rejected", func(t *testing.T) {
+		t.Parallel()
+
+		store := NewTestStore(t)
+		queries := store.Queries()
+
+		walletInfo, err := store.CreateWallet(
+			t.Context(),
+			CreateWatchOnlyWalletParams("watch-only-account-insert"),
+		)
+		require.NoError(t, err)
+
+		props, err := store.CreateImportedAccount(
+			t.Context(), db.CreateImportedAccountParams{
+				WalletID:           walletInfo.ID,
+				Name:               "watch-only-imported",
+				Scope:              db.KeyScopeBIP0084,
+				EncryptedPublicKey: RandomBytes(32),
+			},
+		)
+		require.NoError(t, err)
+		require.True(t, props.IsWatchOnly)
+
+		scopeID := GetKeyScopeID(t, queries, walletInfo.ID, db.KeyScopeBIP0084)
+		accountID := GetAccountID(t, queries, scopeID, "watch-only-imported")
+
+		err = insertAccountSecretRaw(
+			t, store.DB(), accountID, RandomBytes(32),
+		)
+		require.Error(t, err)
+		requireDriverConstraintError(t, err)
+	})
+
+	t.Run(
+		"watch-only empty-but-non-nil insert is rejected",
+		func(t *testing.T) {
+			t.Parallel()
+
+			store := NewTestStore(t)
+			queries := store.Queries()
+
+			walletInfo, err := store.CreateWallet(
+				t.Context(),
+				CreateWatchOnlyWalletParams("watch-only-account-empty"),
+			)
+			require.NoError(t, err)
+
+			props, err := store.CreateImportedAccount(
+				t.Context(), db.CreateImportedAccountParams{
+					WalletID:           walletInfo.ID,
+					Name:               "watch-only-empty",
+					Scope:              db.KeyScopeBIP0084,
+					EncryptedPublicKey: RandomBytes(32),
+				},
+			)
+			require.NoError(t, err)
+			require.True(t, props.IsWatchOnly)
+
+			scopeID := GetKeyScopeID(
+				t, queries, walletInfo.ID, db.KeyScopeBIP0084,
+			)
+			accountID := GetAccountID(
+				t, queries, scopeID, "watch-only-empty",
+			)
+
+			err = insertAccountSecretRaw(t, store.DB(), accountID, []byte{})
+			require.Error(t, err)
+			requireDriverConstraintError(t, err)
+		},
+	)
+
+	t.Run("non-watch-only insert and update succeed", func(t *testing.T) {
+		t.Parallel()
+
+		store := NewTestStore(t)
+		queries := store.Queries()
+		walletID := newWallet(t, store, "spendable-account-secret")
+
+		props, err := store.CreateImportedAccount(
+			t.Context(), db.CreateImportedAccountParams{
+				WalletID:           walletID,
+				Name:               "spendable-imported",
+				Scope:              db.KeyScopeBIP0084,
+				EncryptedPublicKey: RandomBytes(32),
+			},
+		)
+		require.NoError(t, err)
+		require.True(t, props.IsWatchOnly)
+
+		scopeID := GetKeyScopeID(t, queries, walletID, db.KeyScopeBIP0084)
+		accountID := GetAccountID(t, queries, scopeID, "spendable-imported")
+
+		err = insertAccountSecretRaw(
+			t, store.DB(), accountID, RandomBytes(32),
+		)
+		require.NoError(t, err)
+
+		err = updateAccountSecretRaw(
+			t, store.DB(), accountID, RandomBytes(32),
+		)
+		require.NoError(t, err)
+	})
+}
+
 // TestCreateDerivedAccountErrors verifies that CreateDerivedAccount returns
 // appropriate errors for invalid inputs.
 func TestCreateDerivedAccountErrors(t *testing.T) {
--- a/wallet/internal/db/itest/pg_test.go
+++ b/wallet/internal/db/itest/pg_test.go
@@ -163,11 +163,11 @@ func GetPostgresContainer(ctx context.Context) (*postgres.PostgresContainer,
 		m := db.DefaultMaxConnections
 
 		// pgMaxConns is the Postgres max_connections budget for the
-		// test container. We budget 2x the steady-state pool size to
+		// test container. We budget 3x the steady-state pool size to
 		// absorb connection lifecycle overlap during parallel test
 		// teardown and startup without trying to model each transient
 		// connection source separately.
-		pgMaxConns := 2 * p * m
+		pgMaxConns := 3 * p * m
 
 		pgContainer, errPGContainer = postgres.Run(ctx,
 			cfg.Image,
--- a/wallet/internal/sql/pg/migrations/000002_wallets.up.sql
+++ b/wallet/internal/sql/pg/migrations/000002_wallets.up.sql
@@ -32,6 +32,26 @@ CREATE TABLE wallets (
 -- Unique index to prevent duplicate wallet names.
 CREATE UNIQUE INDEX uidx_wallets_name ON wallets (wallet_name);
 
+-- Enforce that the watch-only status chosen at wallet creation time remains
+-- immutable. This closes the database-boundary hole where a raw wallet update
+-- could silently bypass the secret-table triggers by flipping the parent
+-- wallet between watch-only and spendable after insert.
+CREATE FUNCTION assert_wallet_is_watch_only_immutable() RETURNS TRIGGER AS $$
+BEGIN
+    IF NEW.is_watch_only IS DISTINCT FROM OLD.is_watch_only THEN
+        RAISE EXCEPTION 'wallet is_watch_only cannot be changed after creation'
+            USING ERRCODE = '23514'; -- check_violation
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_wallet_is_watch_only_immutable
+BEFORE UPDATE OF is_watch_only ON wallets
+FOR EACH ROW
+EXECUTE FUNCTION assert_wallet_is_watch_only_immutable();
+
 -- Wallet Secrets table to store rarely accessed, highly sensitive encrypted
 -- material with a strict one-to-one relationship with the wallets table.
 -- Separated from the main wallets table for security and access pattern isolation.
--- a/wallet/internal/sql/pg/migrations/000002_wallets.up.sql
+++ b/wallet/internal/sql/pg/migrations/000002_wallets.up.sql
@@ -61,6 +61,42 @@ CREATE TABLE wallet_secrets (
     FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT
 );
 
+-- Enforce the watch-only wallet secret invariant at the database boundary.
+-- Watch-only wallets may retain script-encryption material for imported
+-- scripts, but must never store private key material; keeping these columns
+-- NULL prevents an insert or update from silently turning a watch-only wallet
+-- into a spend-capable wallet.
+CREATE FUNCTION assert_watch_only_wallet_secrets() RETURNS TRIGGER AS $$
+DECLARE
+    wallet_is_watch_only BOOLEAN;
+BEGIN
+    SELECT w.is_watch_only INTO wallet_is_watch_only
+    FROM wallets AS w
+    WHERE w.id = NEW.wallet_id;
+
+    IF wallet_is_watch_only AND (
+        NEW.master_priv_params IS NOT NULL
+        OR NEW.encrypted_crypto_priv_key IS NOT NULL
+        OR NEW.encrypted_master_hd_priv_key IS NOT NULL
+    ) THEN
+        RAISE EXCEPTION 'watch-only wallet private secret columns must be null'
+            USING ERRCODE = '23514'; -- check_violation
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_watch_only_wallet_secrets_insert
+BEFORE INSERT ON wallet_secrets
+FOR EACH ROW
+EXECUTE FUNCTION assert_watch_only_wallet_secrets();
+
+CREATE TRIGGER trg_assert_watch_only_wallet_secrets_update
+BEFORE UPDATE ON wallet_secrets
+FOR EACH ROW
+EXECUTE FUNCTION assert_watch_only_wallet_secrets();
+
 -- Wallet Sync States table to store the synchronization state of each wallet.
 -- This is kept separate from the wallets table to avoid write amplification on
 -- frequently updated sync data. Each wallet has exactly one sync state record.
--- a/wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
+++ b/wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
@@ -61,3 +61,34 @@ CREATE TABLE key_scope_secrets (
     -- that the key scope cannot be deleted if secrets still exist.
     FOREIGN KEY (scope_id) REFERENCES key_scopes (id) ON DELETE RESTRICT
 );
+
+-- Enforce the watch-only key-scope secret invariant at the database boundary.
+-- Watch-only wallets may keep a scope row for public derivation metadata, but
+-- the matching key_scope_secrets row must not carry coin private key material.
+CREATE FUNCTION assert_watch_only_key_scope_secrets() RETURNS TRIGGER AS $$
+DECLARE
+    wallet_is_watch_only BOOLEAN;
+BEGIN
+    SELECT w.is_watch_only INTO wallet_is_watch_only
+    FROM key_scopes AS ks
+    INNER JOIN wallets AS w ON w.id = ks.wallet_id
+    WHERE ks.id = NEW.scope_id;
+
+    IF wallet_is_watch_only THEN
+        RAISE EXCEPTION 'watch-only key scopes cannot store coin private keys'
+            USING ERRCODE = '23514'; -- check_violation
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_watch_only_key_scope_secrets_insert
+BEFORE INSERT ON key_scope_secrets
+FOR EACH ROW
+EXECUTE FUNCTION assert_watch_only_key_scope_secrets();
+
+CREATE TRIGGER trg_assert_watch_only_key_scope_secrets_update
+BEFORE UPDATE ON key_scope_secrets
+FOR EACH ROW
+EXECUTE FUNCTION assert_watch_only_key_scope_secrets();
--- a/wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
+++ b/wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
@@ -44,6 +44,25 @@ ON key_scopes (wallet_id, purpose, coin_type);
 -- Unique index to support composite foreign keys scoped by wallet ownership.
 CREATE UNIQUE INDEX uidx_key_scopes_wallet_id_id ON key_scopes (wallet_id, id);
 
+-- Enforce that wallet ownership chosen at key-scope creation time remains
+-- immutable. This closes the database-boundary hole where a raw update could
+-- reparent an existing scope into another wallet after insert.
+CREATE FUNCTION assert_key_scope_wallet_id_immutable() RETURNS TRIGGER AS $$
+BEGIN
+    IF NEW.wallet_id IS DISTINCT FROM OLD.wallet_id THEN
+        RAISE EXCEPTION 'key scope wallet_id cannot be changed after creation'
+            USING ERRCODE = '23514'; -- check_violation
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_key_scope_wallet_id_immutable
+BEFORE UPDATE OF wallet_id ON key_scopes
+FOR EACH ROW
+EXECUTE FUNCTION assert_key_scope_wallet_id_immutable();
+
 -- Key Scope Secrets table to hold encrypted coin-type secrets for spendable
 -- scopes.
 -- Separated from the main key_scopes table for security and access pattern
--- a/wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
+++ b/wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
@@ -44,18 +44,18 @@ ON key_scopes (wallet_id, purpose, coin_type);
 -- Unique index to support composite foreign keys scoped by wallet ownership.
 CREATE UNIQUE INDEX uidx_key_scopes_wallet_id_id ON key_scopes (wallet_id, id);
 
--- Key Scope Secrets table to hold encrypted coin-type secrets for each scope.
--- Separated from the main key_scopes table for security and access pattern isolation.
--- Watch-only scopes may have no corresponding row in this table or have NULL
--- encrypted_coin_priv_key.
+-- Key Scope Secrets table to hold encrypted coin-type secrets for spendable
+-- scopes.
+-- Separated from the main key_scopes table for security and access pattern
+-- isolation. Watch-only scopes are represented by having no row in this table.
 CREATE TABLE key_scope_secrets (
     -- Reference to the key scope these keys belong to. Also serves as the
     -- primary key, enforcing one-to-one relationship.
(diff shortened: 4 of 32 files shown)


wallet/internal/sql/pg/migrations/000002_wallets.up.sql
```diff
@@ -61,6 +61,42 @@ CREATE TABLE wallet_secrets (
...
+        OR NEW.encrypted_master_hd_priv_key IS NOT NULL
+    ) THEN
+        RAISE EXCEPTION 'watch-only wallet private secret columns must be null'
+            USING ERRCODE = '23514';
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
```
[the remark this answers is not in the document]
GustavoStingelin: intentional, not a problem for now.

wallet/internal/sql/sqlite/migrations/000002_wallets.up.sql
```diff
@@ -61,6 +61,43 @@ CREATE TABLE wallet_secrets (
     FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT
 );
 
+-- Enforce the watch-only wallet secret invariant at the database boundary.
+-- Watch-only wallets may retain script-encryption material for imported
+-- scripts, but must never store private key material; keeping these columns
+-- NULL prevents an insert or update from silently turning a watch-only wallet
+-- into a spend-capable wallet.
```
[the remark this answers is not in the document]
GustavoStingelin: intentional, not a problem for now.

wallet/internal/sql/sqlite/migrations/000005_accounts.up.sql
```diff
@@ -127,3 +127,32 @@ CREATE TABLE account_secrets (
     -- that the account cannot be deleted if secrets still exist.
     FOREIGN KEY (account_id) REFERENCES accounts (id) ON DELETE RESTRICT
 );
+
+-- Enforce the watch-only wallet secret invariant at the database boundary.
+-- Wallet-level watch-only mode means imported accounts must not store account
+-- private key material.
```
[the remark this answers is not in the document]
GustavoStingelin: adjsuted

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -104,6 +104,43 @@ CREATE TABLE address_secrets (
...
+DECLARE
+    wallet_is_watch_only BOOLEAN;
+BEGIN
+    IF NEW.encrypted_priv_key IS NULL THEN
+        RETURN NEW;
+    END IF;
+
+    SELECT w.is_watch_only INTO wallet_is_watch_only
+    FROM addresses AS addr
+    INNER JOIN accounts AS a ON a.id = addr.account_id
```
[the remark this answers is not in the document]
GustavoStingelin: removed

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -128,3 +128,34 @@ CREATE TABLE account_secrets (
     -- that the account cannot be deleted if secrets still exist.
     FOREIGN KEY (account_id) REFERENCES accounts (id) ON DELETE RESTRICT
 );
+
+-- Enforce the watch-only wallet secret invariant at the database boundary.
+-- Wallet-level watch-only mode means imported accounts must not store account
+-- private key material.
```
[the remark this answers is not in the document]
GustavoStingelin: adjsuted

wallet/internal/sql/sqlite/migrations/000006_addresses.up.sql
```diff
@@ -102,6 +102,46 @@ CREATE TABLE address_secrets (
...
+BEGIN
+    SELECT raise(ABORT, 'watch-only address parents cannot store private keys')
+    WHERE
+        new.encrypted_priv_key IS NOT NULL
+        AND EXISTS (
+            SELECT 1
+            FROM addresses AS addr
+            INNER JOIN accounts AS a ON addr.account_id = a.id
+            INNER JOIN wallets AS w ON addr.wallet_id = w.id
+            WHERE
```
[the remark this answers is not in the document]
GustavoStingelin: removed

wallet/internal/sql/pg/migrations/000005_accounts.down.sql
```diff
@@ -1,5 +1,11 @@
 -- Rollback note: Idempotent by design (using "IF EXISTS").
 -- Must succeed even if tables are already dropped or database is in unexpected state.
+DROP TRIGGER IF EXISTS trg_assert_watch_only_account_secrets_insert ON account_secrets;
+DROP TRIGGER IF EXISTS trg_assert_watch_only_account_secrets_update ON account_secrets;
+DROP TRIGGER IF EXISTS trg_assert_watch_only_wallet_accounts_insert ON accounts;
+DROP TRIGGER IF EXISTS trg_assert_watch_only_wallet_accounts_update ON accounts;
+DROP FUNCTION IF EXISTS assert_watch_only_account_secrets();
+DROP FUNCTION IF EXISTS assert_watch_only_wallet_accounts();
```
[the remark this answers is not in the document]
GustavoStingelin: removed

wallet/internal/sql/sqlite/migrations/000005_accounts.down.sql
```diff
@@ -1,5 +1,9 @@
 -- Rollback note: Idempotent by design (using "IF EXISTS").
 -- Must succeed even if tables are already dropped or database is in unexpected state.
+DROP TRIGGER IF EXISTS trg_assert_watch_only_account_secrets_insert;
+DROP TRIGGER IF EXISTS trg_assert_watch_only_account_secrets_update;
+DROP TRIGGER IF EXISTS trg_assert_watch_only_wallet_accounts_insert;
+DROP TRIGGER IF EXISTS trg_assert_watch_only_wallet_accounts_update;
```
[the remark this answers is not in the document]
GustavoStingelin: removed

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -128,3 +128,34 @@ CREATE TABLE account_secrets (
...
+CREATE FUNCTION assert_watch_only_account_secrets() RETURNS TRIGGER AS $$
+DECLARE
+	wallet_is_watch_only BOOLEAN;
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/sql/sqlite/migrations/000002_wallets.up.sql
```diff
@@ -61,6 +61,43 @@ CREATE TABLE wallet_secrets (
...
+CREATE TRIGGER trg_assert_watch_only_wallet_secrets_insert
```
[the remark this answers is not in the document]
GustavoStingelin: added

yyforyongyu (commented):
~~No issues found. The PR introduces SQL-level watch-only enforcement through triggers on wallet_secrets, account_secrets, and address_secrets tables, with corresponding itests that validate trigger behavior via direct SQL writes bypassing the store layer. Migration changes include both up and down SQL files with proper cleanup. LGTM 😊 (MiniMax-M2.7)~~

wallet/internal/sql/pg/migrations/000002_wallets.up.sql
```diff
@@ -61,6 +61,42 @@ CREATE TABLE wallet_secrets (
     FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT
 );
 
+-- Enforce the watch-only wallet secret invariant at the database boundary.
```
yyforyongyu, line 84: Triggers are appended into the existing `000002_wallets.up.sql` (same applies to `000005_accounts.up.sql` and `000006_addresses.up.sql` plus their sqlite counterparts) rather than added as new migration versions. Same concern as PR #1221: any environment that has already applied these versions will not pick up the new triggers because golang-migrate tracks applied versions by file name, not content. Fix hint: if the `sql-wallet` branch is still pre-deployment, document that explicitly; otherwise add a follow-up migration (`000010_watch_only_triggers.up.sql` etc.).
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000002_wallets.up.sql
```diff
@@ -61,6 +61,42 @@ CREATE TABLE wallet_secrets (
...
+-- Watch-only wallets may retain script-encryption material for imported
+-- scripts, but must never store private key material; keeping these columns
+-- NULL prevents an insert or update from silently turning a watch-only wallet
+-- into a spend-capable wallet.
+CREATE FUNCTION assert_watch_only_wallet_secrets() RETURNS TRIGGER AS $$
```
yyforyongyu, line 89: `CREATE FUNCTION assert_watch_only_wallet_secrets()` (and the matching `assert_watch_only_account_secrets`/`assert_watch_only_address_secrets` in 000005 and 000006) is not idempotent: if the migration is replayed (e.g. squashed and re-applied in dev, or if someone runs the SQL by hand), the second run errors with `function ... already exists`. The DOWN migration drops it, but the UP doesn't. Fix hint: use `CREATE OR REPLACE FUNCTION` for all three trigger functions.
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -128,3 +128,34 @@ CREATE TABLE account_secrets (
     -- that the account cannot be deleted if secrets still exist.
     FOREIGN KEY (account_id) REFERENCES accounts (id) ON DELETE RESTRICT
 );
+
+-- Enforce the watch-only account secret invariant at the database boundary.
+-- Accounts in watch-only wallets must not store account-level private key
+-- material.
+CREATE FUNCTION assert_watch_only_account_secrets() RETURNS TRIGGER AS $$
```
yyforyongyu, line 160: This trigger rejects ANY insert/update on `account_secrets` when the parent wallet is watch-only, regardless of column content — different shape from `assert_watch_only_address_secrets` which checks `encrypted_priv_key IS NULL` first. The asymmetry is correct because `account_secrets.encrypted_private_key` is `NOT NULL` (any row implies private key material), but a future reader will wonder why these two triggers diverge. Fix hint: add an inline comment noting the schema constraint that makes the column-check redundant here.
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -104,6 +104,42 @@ CREATE TABLE address_secrets (
...
 
+-- Enforce the watch-only address secret invariant at the database boundary.
+-- Watch-only parent wallets may track imported scripts, but addresses
+-- beneath them must not store private keys; otherwise a watch-only parent could
+-- silently gain spend authority through an address secret row.
+CREATE FUNCTION assert_watch_only_address_secrets() RETURNS TRIGGER AS $$
+DECLARE
+    wallet_is_watch_only BOOLEAN;
+BEGIN
+    IF NEW.encrypted_priv_key IS NULL THEN
```
yyforyongyu, line 134: Trigger short-circuits on `encrypted_priv_key IS NULL`, but an empty `bytea` (`''`) is NOT NULL, so it falls through to the watch-only check. Meanwhile `NewImportedAddressParams.HasPrivateKey()` in `addresses_common.go` uses `len(p.EncryptedPrivateKey) > 0`, so the same empty `[]byte{}` payload passes `Validate(walletIsWatchOnly)` at the app layer (`HasPrivateKey()` returns false) and is forwarded to the SQL insert via `HasSecretMaterial()` (true because of the script). The trigger then catches it as a constraint violation rather than `ErrWatchOnlyViolation`. The "watch-only empty-but-non-nil private key with script is rejected" subtest documents this divergence — defense-in-depth works, but the user gets a raw constraint error. Fix hint: align `HasPrivateKey()` to `p.EncryptedPrivateKey != nil` (or use `octet_length(...) > 0` in the trigger) so both layers agree on what "has a private key" means.
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000002_wallets.up.sql
```diff
@@ -61,6 +61,42 @@ CREATE TABLE wallet_secrets (
...
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_watch_only_wallet_secrets_insert
+BEFORE INSERT ON wallet_secrets
+FOR EACH ROW
+EXECUTE FUNCTION assert_watch_only_wallet_secrets();
+
+CREATE TRIGGER trg_assert_watch_only_wallet_secrets_update
```
yyforyongyu, line 115: These triggers only fire when `wallet_secrets` changes, so a direct `UPDATE wallets SET is_watch_only = TRUE` on an existing spendable wallet bypasses them and leaves private wallet/account/address secrets attached to a wallet now reported as watch-only. Please either make `is_watch_only` immutable or add a wallet update trigger that rejects `FALSE -> TRUE` while any private secret rows exist. (`openai/gpt-5.5`)

wallet/internal/sql/sqlite/migrations/000002_wallets.up.sql
```diff
@@ -61,6 +61,43 @@ CREATE TABLE wallet_secrets (
...
+        FROM wallets AS w
+        WHERE w.id = new.wallet_id
+    ) AND (
+        new.master_priv_params IS NOT NULL
+        OR new.encrypted_crypto_priv_key IS NOT NULL
+        OR new.encrypted_master_hd_priv_key IS NOT NULL
+    );
+END;
+
+CREATE TRIGGER trg_assert_watch_only_wallet_secrets_update
```
yyforyongyu, line 97: These triggers only fire when `wallet_secrets` changes, so a direct `UPDATE wallets SET is_watch_only = TRUE` on an existing spendable wallet bypasses them and leaves private wallet/account/address secrets attached to a wallet now reported as watch-only. Please either make `is_watch_only` immutable or add a wallet update trigger that rejects `FALSE -> TRUE` while any private secret rows exist. (`openai/gpt-5.5`)

yyforyongyu (changes requested):
There are some good comments made from gpt and claude, meanwhile can we simply make `is_watch_only` immutable after wallet creation?

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
GustavoStingelin: added

wallet/internal/sql/pg/migrations/000002_wallets.up.sql
GustavoStingelin: added

wallet/internal/sql/sqlite/migrations/000002_wallets.up.sql
GustavoStingelin: added

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
GustavoStingelin: I’d prefer to keep the database side representation normalized to `NULL` for an empty private key, rather than treating non-nil empty bytes as another form of `nil`.

Since the API layer currently accepts non-nil empty slices as valid input, I've added a normalizer that stores them as `NULL` on the database side.

This avoids adding extra complexity to the database layer, where we would otherwise need another validation branch to check whether the value is `NULL` or `empty`.

yyforyongyu:

yyforyongyu:
LGTM 🔒
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000002_wallets.up.sql
```diff
@@ -61,6 +81,42 @@ CREATE TABLE wallet_secrets (
     FOREIGN KEY (wallet_id) REFERENCES wallets (id) ON DELETE RESTRICT
 );
 
+-- Enforce the watch-only wallet secret invariant at the database boundary.
```
yyforyongyu, line 84: This leaves `key_scope_secrets.encrypted_coin_priv_key` outside the database invariant. That column is also encrypted private key material (`000004_key_scopes.up.sql` says it must be NULL for watch-only scopes), and the generated `InsertKeyScopeSecrets` query can insert a non-NULL value for a scope whose wallet is watch-only without touching `wallet_secrets`, so none of these triggers fire. I reproduced this against the SQLite schema with a direct `INSERT INTO key_scope_secrets` under a watch-only scope.

Suggested fix:
- Add PostgreSQL and SQLite `key_scope_secrets` insert/update validation that joins `NEW.scope_id -> key_scopes -> wallets` and rejects non-NULL `encrypted_coin_priv_key` when the parent wallet is watch-only.
- Also cover parent movement by either making `key_scopes.wallet_id` immutable, or adding a `key_scopes` update trigger that rejects moving a scope with non-NULL key-scope secret material into a watch-only wallet.
- Add a regression test that direct SQL cannot attach key-scope private material to a watch-only wallet.

(`openai/gpt-5.5`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -128,3 +128,40 @@ CREATE TABLE account_secrets (
...
+-- Enforce the watch-only account secret invariant at the database boundary.
+-- Accounts in watch-only wallets must not store account-level private key
+-- material.
+--
+-- Note: Unlike address_secrets.encrypted_priv_key (which is nullable for
+-- HD-derived addresses), account_secrets.encrypted_private_key is NOT NULL.
+-- This means any row in account_secrets necessarily represents private key
+-- material, so we reject all inserts/updates for watch-only parents without
+-- needing to check column nullability first.
+CREATE FUNCTION assert_watch_only_account_secrets() RETURNS TRIGGER AS $$
```
yyforyongyu, line 160: This trigger only runs when `account_secrets` changes, so it still misses parent reparenting. A direct `UPDATE accounts SET wallet_id = <watch_wallet>, scope_id = <watch_scope> WHERE id = <account_with_secret>` is accepted by the composite FK and leaves the existing `account_secrets` row attached to an account now under a watch-only wallet; no `account_secrets` trigger fires. I reproduced this against the SQLite schema.

Suggested fix:
- If account ownership is not expected to change, add PostgreSQL and SQLite `BEFORE UPDATE OF wallet_id, scope_id ON accounts` triggers that reject ownership changes.
- If reparenting should remain possible, add an `accounts` update trigger that checks the target `NEW.wallet_id`/`NEW.scope_id`; when the target wallet is watch-only, reject the update if `account_secrets` contains a row for `NEW.id`.
- Add a regression test that an account with existing account secrets cannot be moved under a watch-only wallet.

(`openai/gpt-5.5`)

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -104,6 +104,42 @@ CREATE TABLE address_secrets (
...
+    IF wallet_is_watch_only THEN
+        RAISE EXCEPTION 'watch-only address parents cannot store private keys'
+            USING ERRCODE = '23514';
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_watch_only_address_secrets_insert
```
yyforyongyu, line 152: Same parent-row gap for addresses: the trigger only fires on `address_secrets`, but `UPDATE addresses SET wallet_id = <watch_wallet>, account_id = <watch_account> WHERE id = <address_with_private_secret>` satisfies the `(wallet_id, account_id)` FK and moves an existing `address_secrets.encrypted_priv_key` under a watch-only wallet without revalidating it. I reproduced this against SQLite.

Suggested fix:
- If address ownership is not expected to change, add PostgreSQL and SQLite `BEFORE UPDATE OF wallet_id, account_id ON addresses` triggers that reject ownership changes.
- If reparenting should remain possible, add an `addresses` update trigger that checks the target `NEW.wallet_id`/`NEW.account_id`; when the target wallet is watch-only, reject the update if `address_secrets` has private material for `NEW.id` (`encrypted_priv_key IS NOT NULL`).
- Add a regression test that an address with existing private address secret material cannot be moved under a watch-only wallet/account.

(`openai/gpt-5.5`)

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -104,6 +104,42 @@ CREATE TABLE address_secrets (
...
+    END IF;
+
+    SELECT w.is_watch_only INTO wallet_is_watch_only
+    FROM addresses AS addr
+    INNER JOIN wallets AS w ON w.id = addr.wallet_id
+    WHERE addr.id = NEW.address_id;
+
+    IF wallet_is_watch_only THEN
+        RAISE EXCEPTION 'watch-only address parents cannot store private keys'
+            USING ERRCODE = '23514';
```
yyforyongyu: I think we should add some inline comments here explaining this magic number, same for other occurrences.

yyforyongyu (changes requested):
ok think `gpt` found some valid cases, basically direct SQL could still add private key material to a key scope that belongs to a watch-only wallet, which is the `key_scope_secrets`. And similarly for `account_secrets` and `address_secrets`.

wallet/internal/sql/pg/migrations/000002_wallets.up.sql
```diff
@@ -32,6 +32,26 @@ CREATE TABLE wallets (
 -- Unique index to prevent duplicate wallet names.
 CREATE UNIQUE INDEX uidx_wallets_name ON wallets (wallet_name);
 
+-- Enforce that the watch-only status chosen at wallet creation time remains
+-- immutable. This closes the database-boundary hole where a raw wallet update
+-- could silently bypass the secret-table triggers by flipping the parent
+-- wallet between watch-only and spendable after insert.
+CREATE FUNCTION assert_wallet_is_watch_only_immutable() RETURNS TRIGGER AS $$
```
yyforyongyu, line 39: Adding triggers to an already-applied migration means existing databases will never get them. Use a new migration file (e.g. `00000X_add_watch_only_wallet_triggers.sql`) instead. Same issue applies to `000005_accounts.up.sql` and `000006_addresses.up.sql` in both PostgreSQL and SQLite.

`k2p6`

wallet/internal/sql/sqlite/migrations/000002_wallets.up.sql
```diff
@@ -32,6 +32,18 @@ CREATE TABLE wallets (
 -- Unique index to prevent duplicate wallet names.
 CREATE UNIQUE INDEX uidx_wallets_name ON wallets (wallet_name);
 
+-- Enforce that the watch-only status chosen at wallet creation time remains
+-- immutable. This closes the database-boundary hole where a raw wallet update
+-- could silently bypass the secret-table triggers by flipping the parent
+-- wallet between watch-only and spendable after insert.
+CREATE TRIGGER trg_assert_wallet_is_watch_only_immutable
```
yyforyongyu, line 39: Same migration-compatibility issue: SQLite databases that already applied this migration will not get the new triggers. Use a new migration file instead.

`k2p6`

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -128,3 +128,40 @@ CREATE TABLE account_secrets (
     -- that the account cannot be deleted if secrets still exist.
     FOREIGN KEY (account_id) REFERENCES accounts (id) ON DELETE RESTRICT
 );
+
+-- Enforce the watch-only account secret invariant at the database boundary.
```
yyforyongyu, line 151: Same migration-compatibility issue: existing PostgreSQL databases that already applied this migration will not get the account_secrets triggers. Use a new migration file instead.

`k2p6`

wallet/internal/sql/sqlite/migrations/000005_accounts.up.sql
```diff
@@ -127,3 +127,38 @@ CREATE TABLE account_secrets (
     -- that the account cannot be deleted if secrets still exist.
     FOREIGN KEY (account_id) REFERENCES accounts (id) ON DELETE RESTRICT
 );
+
+-- Enforce the watch-only account secret invariant at the database boundary.
```
yyforyongyu, line 142: Same migration-compatibility issue: existing SQLite databases that already applied this migration will not get the account_secrets triggers. Use a new migration file instead.

`k2p6`

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -104,6 +104,42 @@ CREATE TABLE address_secrets (
     FOREIGN KEY (address_id) REFERENCES addresses (id) ON DELETE RESTRICT
 );
 
+-- Enforce the watch-only address secret invariant at the database boundary.
```
yyforyongyu, line 126: Same migration-compatibility issue: existing PostgreSQL databases that already applied this migration will not get the address_secrets triggers. Use a new migration file instead.

`k2p6`

wallet/internal/sql/sqlite/migrations/000006_addresses.up.sql
```diff
@@ -102,6 +102,44 @@ CREATE TABLE address_secrets (
     FOREIGN KEY (address_id) REFERENCES addresses (id) ON DELETE RESTRICT
 );
 
+-- Enforce the watch-only address secret invariant at the database boundary.
```
yyforyongyu, line 116: Same migration-compatibility issue: existing SQLite databases that already applied this migration will not get the address_secrets triggers. Use a new migration file instead.

`k2p6`

wallet/internal/sql/sqlite/migrations/000006_addresses.down.sql
```diff
@@ -1,5 +1,9 @@
 -- Rollback note: Idempotent by design (using "IF EXISTS").
 -- Must succeed even if tables are already dropped or database is in unexpected state.
+DROP TRIGGER IF EXISTS trg_assert_watch_only_address_secrets_insert;
```
yyforyongyu, line 3: Adding previously-missing `DROP TRIGGER` statements for `trg_addresses_imported_key_count_*` is correct, but modifying an already-applied down migration breaks the assumption that migrations are immutable. The original migration was incomplete; this fix should be noted but ideally migration files remain immutable once applied.

`k2p6`

yyforyongyu:

yyforyongyu:
LGTM 🛡️
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000002_wallets.up.sql
GustavoStingelin: added

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
GustavoStingelin: added

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
GustavoStingelin: added

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
GustavoStingelin: Added a simple one, tell me if u think that needs more explanation

yyforyongyu:
LGTM 🛡️ (v4)

Diff vs v3 adds four extensions consistent with the watch-only + immutability pattern from v3:
- `wallet: require key-scope secrets for spendable rows` tightens `key_scope_secrets.encrypted_coin_priv_key` to `NOT NULL`, so watch-only scopes are represented purely by row absence.
- `wallet: enforce key-scope watch-only invariants` adds the `assert_watch_only_key_scope_secrets` trigger pair on `key_scope_secrets` (mirrors the `wallet_secrets` / `account_secrets` / `address_secrets` triggers from v3), with paired raw-SQL tests in `TestWatchOnlyKeyScopeSecretTriggers`.
- `wallet: make wallet ownership immutable` adds `BEFORE UPDATE OF wallet_id` triggers on `key_scopes`, `accounts`, and `addresses` (companion to the existing `is_watch_only` immutability trigger on `wallets`). `TestKeyScopeWalletIDImmutable`, `TestAccountWalletIDImmutable`, and `TestAddressWalletIDImmutable` exercise each via `reparent…Raw` helpers and `requireDriverConstraintError`.
- `wallet: increase postgres itest connection budget` bumps `pgMaxConns` from 2× to 3× steady-state pool size to absorb teardown overlap.
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -83,6 +83,25 @@ ON addresses (wallet_id, script_pub_key);
...
+    IF NEW.wallet_id IS DISTINCT FROM OLD.wallet_id THEN
+        RAISE EXCEPTION 'address wallet_id cannot be changed after creation'
+            USING ERRCODE = '23514'; -- check_violation
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_address_wallet_id_immutable
```
yyforyongyu, line 100: This has the same parent immutability gap for addresses: only `wallet_id` is frozen, while `account_id` can still be updated to another account in the same wallet. For imported addresses, that also leaves `imported_key_count` wrong because the insert/delete triggers do not run on account moves. Please either make `account_id` immutable too, or add update-trigger logic that moves the count between accounts, with same-wallet reparenting tests for both backends. (`openai/gpt-5.5`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -114,6 +114,25 @@ WHERE account_number IS NOT NULL;
...
+    IF NEW.wallet_id IS DISTINCT FROM OLD.wallet_id THEN
+        RAISE EXCEPTION 'account wallet_id cannot be changed after creation'
+            USING ERRCODE = '23514'; -- check_violation
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_account_wallet_id_immutable
```
yyforyongyu, line 131: This trigger still only protects `wallet_id`. A raw same-wallet update such as `UPDATE accounts SET scope_id = <other scope in the same wallet>` satisfies the composite FK and bypasses this trigger, moving the account and its existing secrets/counters into a different derivation scope. If account parentage is meant to be fixed at creation time, please reject `scope_id` changes as well and add same-wallet reparenting coverage for both PostgreSQL and SQLite. (`openai/gpt-5.5`)

wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
```diff
@@ -44,20 +44,70 @@ ON key_scopes (wallet_id, purpose, coin_type);
...
+    IF NEW.wallet_id IS DISTINCT FROM OLD.wallet_id THEN
+        RAISE EXCEPTION 'key scope wallet_id cannot be changed after creation'
+            USING ERRCODE = '23514'; -- check_violation
+    END IF;
+
+    RETURN NEW;
+END;
+$$ LANGUAGE plpgsql;
+
+CREATE TRIGGER trg_assert_key_scope_wallet_id_immutable
```
yyforyongyu, line 61: This only freezes `wallet_id`, so raw SQL can still mutate the logical scope identity after creation: `purpose`, `coin_type`, `internal_type_id`, and `external_type_id` remain updateable. That can retag an existing scope, along with any child accounts/addresses and key-scope secrets, as a different derivation scope without changing row IDs. If key scopes are intended to be immutable once created, please reject updates to the identity/schema columns too and mirror the same trigger/test coverage in SQLite. (`openai/gpt-5.5`)

wallet/internal/db/itest/account_store_test.go
```diff
@@ -100,6 +100,116 @@ func TestCreateImportedAccountRejectsWalletScopeMismatch(t *testing.T) {
...
+		accountID := GetAccountID(t, queries, scopeID, "watch-only-imported")
+
+		err = insertAccountSecretRaw(
+			t, store.DB(), accountID, RandomBytes(32),
+		)
+		require.Error(t, err)
+		requireDriverConstraintError(t, err)
+	})
+
+	t.Run(
```
yyforyongyu, line 142: hmm this format is bit funny, but i guess it's technically correct

wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
yyforyongyu: I think it's fine to leave as is - this rarely happens, but will leave this as unresolved for future reference.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: same here - leave this as unresolved for future reference.

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
yyforyongyu: same here - leave this as unresolved for future reference.

wallet/internal/db/itest/pg_test.go
```diff
@@ -163,11 +163,11 @@ func GetPostgresContainer(ctx context.Context) (*postgres.PostgresContainer,
 		m := db.DefaultMaxConnections
 
 		// pgMaxConns is the Postgres max_connections budget for the
-		// test container. We budget 2x the steady-state pool size to
+		// test container. We budget 3x the steady-state pool size to
 		// absorb connection lifecycle overlap during parallel test
 		// teardown and startup without trying to model each transient
 		// connection source separately.
-		pgMaxConns := 2 * p * m
+		pgMaxConns := 3 * p * m
```
yyforyongyu, line 170: Also encountered this in my other PRs and bumped it to 4😂

yyforyongyu (approved):
🦾

PR #1224, opened 2026-05-05
wallet: store public key in plaintext

GustavoStingelin:
store public key in plaintext following `ADR 0009: Single-Passphrase Encryption Model`

PR #1225, opened 2026-05-08
wallet: replace advisory lock with persisted account counter

GustavoStingelin:
### What we’re doing

This replaces advisory lock based account allocation with a persisted account counter, and refactors the new derived address flow to use the ops pattern.

The main idea is to stop reconstructing allocation state at runtime, and instead keep the next account number in the database, updated transactionally.

This follows the same direction raised [there](https://github.com/btcsuite/btcwallet/pull/1162#discussion_r2755247118): avoid putting hot path counting and allocation logic on top of growing table scans, and move that state into the database.

### Why

The motivation came from the PR discussion around address counts.

Deriving counts from the addresses table makes the cost grow with the number of addresses in the wallet. That is not ideal for a path that can become hot over time.

The same concern applies to account allocation. Instead of relying on a transient coordination mechanism like advisory locks, we can model the allocation state directly in the database.

By persisting the next account number and updating it in the same transaction as the account creation, correctness comes from durable database state rather than from a lock plus a value reconstructed at runtime.

### Benefits

This makes the concurrency model simpler and more explicit.

It also removes Postgres specific advisory lock logic from this path, which keeps the SQLite and Postgres implementations closer.

The ops pattern refactor should make the derived address workflow easier to maintain. Shared sequencing and invariants live in one place, while each backend keeps only the query details it needs.

That should make future changes easier to reason about, reduce backend drift, and avoid spreading the same workflow rules across multiple implementations.

### Caveats

The tradeoff is that persisted counters move more responsibility to the write path.

We need to make sure the counter is always updated in the same transaction as the row creation, and that every code path creating the relevant rows preserves that invariant.

I think this tradeoff is worth it, because it makes the allocation state explicit and durable instead of relying on a backend specific locking scheme.

### DIFF
--- a/wallet/internal/db/accounts_common.go
+++ b/wallet/internal/db/accounts_common.go
@@ -30,6 +30,103 @@ func (params *CreateDerivedAccountParams) Validate() error {
 	return nil
 }
 
+// CreateDerivedAccountRow contains the backend-independent fields the shared
+// CreateDerivedAccount workflow needs from the final insert row.
+type CreateDerivedAccountRow struct {
+	AccountNumber sql.NullInt64
+	CreatedAt     time.Time
+}
+
+// CreateDerivedAccountOps is the backend adapter the shared
+// CreateDerivedAccount workflow uses.
+//
+// The shared account-creation algorithm is intentionally ordered:
+//   - validate the public request before any backend step runs
+//   - load the wallet watch-only mode so the returned AccountInfo matches the
+//     stored wallet state
+//   - ensure the requested key scope exists before allocating from its counter
+//   - allocate the next derived account number for that scope
+//   - insert the derived account row with the allocated number
+//   - normalize the inserted row into the public AccountInfo result
+//
+// The adapter methods map directly to those stages so the shared helper keeps
+// the sequencing and invariants while each backend keeps its sqlc query types,
+// binding shapes, and row conversions local.
+type CreateDerivedAccountOps interface {
+	// WalletWatchOnly returns whether the target wallet currently runs in
+	// watch-only mode.
+	WalletWatchOnly(ctx context.Context, walletID uint32) (bool, error)
+
+	// EnsureScope returns the existing or newly created key-scope row ID for
+	// the wallet/scope pair.
+	EnsureScope(ctx context.Context, walletID uint32,
+		scope KeyScope) (int64, error)
+
+	// AllocateAccountNumber advances and returns the next derived account
+	// number for the provided scope row.
+	AllocateAccountNumber(ctx context.Context, scopeID int64) (int64, error)
+
+	// CreateDerivedAccount inserts the derived account row using the provided
+	// scope ID, allocated account number, and public account name.
+	CreateDerivedAccount(ctx context.Context, scopeID int64,
+		accountNumber int64, name string) (CreateDerivedAccountRow, error)
+}
+
+// CreateDerivedAccountWithOps runs the backend-independent
+// CreateDerivedAccount workflow once the caller has opened a backend-specific
+// SQL transaction.
+//
+// The helper owns the end-to-end sequencing so postgres and sqlite both:
+// validate the public request first, allocate from the same scope counter only
+// after that scope exists, preserve the same account-number overflow mapping,
+// and build the same normalized AccountInfo result from the inserted row.
+func CreateDerivedAccountWithOps(ctx context.Context,
+	params CreateDerivedAccountParams,
+	ops CreateDerivedAccountOps) (*AccountInfo, error) {
+
+	paramsErr := params.Validate()
+	if paramsErr != nil {
+		return nil, paramsErr
+	}
+
+	walletIsWatchOnly, err := ops.WalletWatchOnly(ctx, params.WalletID)
+	if err != nil {
+		return nil, fmt.Errorf("wallet watch only: %w", err)
+	}
+
+	scopeID, err := ops.EnsureScope(ctx, params.WalletID, params.Scope)
+	if err != nil {
+		return nil, fmt.Errorf("ensure scope: %w", err)
+	}
+
+	accountNumber, err := ops.AllocateAccountNumber(ctx, scopeID)
+	if err != nil {
+		return nil, fmt.Errorf("allocate account number: %w", err)
+	}
+
+	row, err := ops.CreateDerivedAccount(
+		ctx, scopeID, accountNumber, params.Name,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("create account: %w", err)
+	}
+
+	if !row.AccountNumber.Valid {
+		// This should never happen unless the query is modified incorrectly.
+		return nil, ErrNilDBAccountNumber
+	}
+
+	accNumber, err := Int64ToUint32(row.AccountNumber.Int64)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %w", ErrMaxAccountNumberReached, err)
+	}
+
+	return BuildAccountInfo(
+		accNumber, params.Name, DerivedAccount, 0, 0, 0,
+		walletIsWatchOnly, row.CreatedAt, params.Scope,
+	), nil
+}
+
 // ValidateBasic validates required fields for creating an imported account.
 func (params *CreateImportedAccountParams) ValidateBasic() error {
 	if params.Name == "" {
--- a/wallet/internal/db/accounts_common_test.go
+++ b/wallet/internal/db/accounts_common_test.go
@@ -1,9 +1,20 @@
 package db
 
 import (
+	"context"
+	"database/sql"
+	"errors"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/require"
+)
+
+var (
+	// Test errors for CreateDerivedAccount ops testing.
+	errTestWallet = errors.New("wallet")
+	errTestScope  = errors.New("scope")
+	errTestBoom   = errors.New("boom")
 )
 
 // TestCreateDerivedAccountParamsValidate verifies derived account creation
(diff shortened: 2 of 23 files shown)


wallet/internal/sql/sqlite/migrations/000004_key_scopes.up.sql
```diff
@@ -20,6 +20,9 @@ CREATE TABLE key_scopes (
     -- Coin public key for this scope in imported accounts.
     coin_pub_key BLOB,
 
+    -- Next derived account number to allocate within this key scope.
+    next_account_number INTEGER NOT NULL DEFAULT 0,
+
```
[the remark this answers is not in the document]
GustavoStingelin: intentional, not released yet.

wallet/internal/sql/pg/migrations/000004_key_scopes.up.sql
```diff
@@ -20,6 +20,9 @@ CREATE TABLE key_scopes (
     -- Coin public key for this scope in imported accounts.
     coin_pub_key BYTEA,
 
+    -- Next derived account number to allocate within this key scope.
+    next_account_number BIGINT NOT NULL DEFAULT 0,
+
```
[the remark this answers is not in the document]
GustavoStingelin: intentional, not released yet.

yyforyongyu:
LGTM 🔢
(`claude-opus-4-7[1m]`)

yyforyongyu (approved):
LGTM ✅ (openai/gpt-5.5)

wallet/internal/db/accounts_common.go
```diff
@@ -30,6 +30,103 @@ func (params *CreateDerivedAccountParams) Validate() error {
...
+//     stored wallet state
+//   - ensure the requested key scope exists before allocating from its counter
+//   - allocate the next derived account number for that scope
+//   - insert the derived account row with the allocated number
+//   - normalize the inserted row into the public AccountInfo result
+//
+// The adapter methods map directly to those stages so the shared helper keeps
+// the sequencing and invariants while each backend keeps its sqlc query types,
+// binding shapes, and row conversions local.
+type CreateDerivedAccountOps interface {
```
yyforyongyu, line 55: 👍

yyforyongyu (approved):
Maybe update the PR desc to include why we make this change? Got some questions in a side channel for clarification.

wallet/internal/sql/pg/sqlc/key_scopes.sql.go
```diff
@@ -76,6 +76,23 @@ func (q *Queries) DeleteKeyScopeSecrets(ctx context.Context, scopeID int64) (int
...
 }
 
+const GetAndIncrementNextAccountNumber = `-- name: GetAndIncrementNextAccountNumber :one
+UPDATE key_scopes
+SET next_account_number = next_account_number + 1
+WHERE id = $1
+RETURNING (next_account_number - 1)::BIGINT AS account_number
+`
+
+// Atomically gets the next derived account number for a key scope and
```
yyforyongyu, line 86: The generated comment does not start with the exact method name `GetAndIncrementNextAccountNumber`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/pg/sqlc/key_scopes.sql.go
```diff
@@ -89,10 +106,20 @@ FROM key_scopes
...
+	ID             int64
+	WalletID       int64
+	Purpose        int64
+	CoinType       int64
+	CoinPubKey     []byte
+	InternalTypeID int16
+	ExternalTypeID int16
+}
+
 // Retrieves a key scope by its ID.
```
yyforyongyu, line 119: The generated comment does not start with the exact method name `GetKeyScopeByID`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/pg/sqlc/key_scopes.sql.go
```diff
@@ -124,10 +151,20 @@ type GetKeyScopeByWalletAndScopeParams struct {
...
+	ID             int64
+	WalletID       int64
+	Purpose        int64
+	CoinType       int64
+	CoinPubKey     []byte
+	InternalTypeID int16
+	ExternalTypeID int16
+}
+
 // Retrieves a key scope by wallet ID, purpose, and coin type.
```
yyforyongyu, line 164: The generated comment does not start with the exact method name `GetKeyScopeByWalletAndScope`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/pg/sqlc/key_scopes.sql.go
```diff
@@ -191,16 +228,26 @@ WHERE wallet_id = $1
...
+	ID             int64
+	WalletID       int64
+	Purpose        int64
+	CoinType       int64
+	CoinPubKey     []byte
+	InternalTypeID int16
+	ExternalTypeID int16
+}
+
 // Lists all key scopes for a wallet, ordered by ID.
```
yyforyongyu, line 241: The generated comment does not start with the exact method name `ListKeyScopesByWallet`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/key_scopes.sql.go
```diff
@@ -76,6 +76,23 @@ func (q *Queries) DeleteKeyScopeSecrets(ctx context.Context, scopeID int64) (int
...
 }
 
+const GetAndIncrementNextAccountNumber = `-- name: GetAndIncrementNextAccountNumber :one
+UPDATE key_scopes
+SET next_account_number = next_account_number + 1
+WHERE id = ?
+RETURNING next_account_number - 1 AS account_number
+`
+
+// Atomically gets the next derived account number for a key scope and
```
yyforyongyu, line 86: The generated comment does not start with the exact method name `GetAndIncrementNextAccountNumber`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/key_scopes.sql.go
```diff
@@ -89,10 +106,20 @@ FROM key_scopes
...
+	ID             int64
+	WalletID       int64
+	Purpose        int64
+	CoinType       int64
+	CoinPubKey     []byte
+	InternalTypeID int64
+	ExternalTypeID int64
+}
+
 // Retrieves a key scope by its ID.
```
yyforyongyu, line 119: The generated comment does not start with the exact method name `GetKeyScopeByID`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/key_scopes.sql.go
```diff
@@ -124,10 +151,20 @@ type GetKeyScopeByWalletAndScopeParams struct {
...
+	ID             int64
+	WalletID       int64
+	Purpose        int64
+	CoinType       int64
+	CoinPubKey     []byte
+	InternalTypeID int64
+	ExternalTypeID int64
+}
+
 // Retrieves a key scope by wallet ID, purpose, and coin type.
```
yyforyongyu, line 164: The generated comment does not start with the exact method name `GetKeyScopeByWalletAndScope`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/key_scopes.sql.go
```diff
@@ -191,16 +228,26 @@ WHERE wallet_id = ?
...
+	ID             int64
+	WalletID       int64
+	Purpose        int64
+	CoinType       int64
+	CoinPubKey     []byte
+	InternalTypeID int64
+	ExternalTypeID int64
+}
+
 // Lists all key scopes for a wallet, ordered by ID.
```
yyforyongyu, line 241: The generated comment does not start with the exact method name `ListKeyScopesByWallet`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

yyforyongyu:
LGTM 🧮
(`claude-opus-4-7[1m]`)

wallet/internal/db/accounts_common.go
GustavoStingelin: updated the description, elaborating a bit

yyforyongyu:
LGTM 🧮 (v3)

PR 1225-specific content is unchanged from v2 — the two commits (`b8c756fd7 wallet: replace advisory lock with persisted account counter`, `b4ced30495 wallet: extract shared CreateDerivedAccount workflow`) carry the same content, just rebased onto the now-richer base via #1222 v4 (key-scope/wallet-id immutability) and #1224 v3 (ADR comments). My v2 LGTM still stands.
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/sqlc/key_scopes.sql.go
yyforyongyu, line 86: This is still unresolved at the current head: the generated method comment does not start with `GetAndIncrementNextAccountNumber`, so the touched Go file still violates the repo comment gate. The same generated file also has `GetKeyScopeByID`, `GetKeyScopeByWalletAndScope`, and `ListKeyScopesByWallet` comments that start with verbs instead of the declaration names. Please update the corresponding SQL source comments so regenerated pg/sqlite sqlc output has method-name-prefixed docs. (`openai/gpt-5.5`)

wallet/internal/db/accounts_common_test.go
```diff
@@ -156,3 +167,308 @@ func TestRenameAccountParamsValidate(t *testing.T) {
...
+		row: CreateDerivedAccountRow{
+			AccountNumber: sql.NullInt64{Int64: 1, Valid: true},
+			CreatedAt:     time.Unix(1, 0),
+		},
+	}
+}
+
+// stubCreateDerivedAccountOps is a deterministic test adapter for the shared
+// CreateDerivedAccount helper.
+type stubCreateDerivedAccountOps struct {
```
yyforyongyu, line 406: let's stick to `mock.Mock`, similar to this one?
```
// mockCreateTxOps is a mock implementation of CreateTxOps.
type mockCreateTxOps struct {
	mock.Mock
```

We may need to update the `AGENTS.md` or contribution guidelines on the requirement to use `mock.Mock`, making it the standard way to impl test structs for interfaces.

yyforyongyu (approved):
Only one thing about the tests, but can be merged!

wallet/internal/db/accounts_common_test.go
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1233

PR #1227, opened 2026-05-11
build: lint tagged itest variants

yyforyongyu:
## Summary
- Extend `make lint` and `make lint-check` to cover default source, e2e integration tests, and both DB integration-test build-tag variants.
- Default golangci-lint concurrency to 4 workers while keeping `workers=<n>` override support.
- Clean up tagged e2e and DB integration tests so the new lint contexts pass.

## Testing
- `git diff --check`
- `make lint-check`
- `GOWORK=off go test -tags="itest dev nolog" ./itest -run '^$'`
- `GOWORK=off make itest-db db=sqlite`
- `GOWORK=off make itest-db db=postgres`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -169,6 +169,19 @@ linters:
         linters:
           - staticcheck
 
+      # E2E integration tests run a long-lived backend harness and execute test
+      # cases in registration order so the suite can stop on first failure.
+      - path: itest/.*_test\.go
+        linters:
+          - paralleltest
+
+      # DB integration tests often share a backend fixture across subtests, and
+      # forcing parallel subtests makes those ordering assumptions unsafe.
+      - path: wallet/internal/db/itest/.*_test\.go
+        linters:
+          - paralleltest
+          - tparallel
+
 issues:
   # Show only new issues created after git revision `REV`.
   # Default: ""
--- a/Makefile
+++ b/Makefile
@@ -26,10 +26,29 @@ XARGS := xargs -L 1
 include make/testing_flags.mk
 
 # Linting uses a lot of memory, so keep it under control by limiting the number
-# of workers if requested.
-ifneq ($(workers),)
+# of workers. Override with `workers=<n>` when needed.
+workers ?= 4
 LINT_WORKERS = --concurrency=$(workers)
-endif
+
+GOLANGCI_LINT = golangci-lint run -v --config .golangci.yml $(LINT_WORKERS)
+LINT_DB_SQLITE_PKGS = ./wallet/internal/db/sqlite ./wallet/internal/db/itest
+LINT_DB_POSTGRES_PKGS = ./wallet/internal/db/pg ./wallet/internal/db/itest
+
+define lint_targets
+	@$(call print, "Linting source.")
+	$(DOCKER_TOOLS) $(GOLANGCI_LINT) $(LINT_FIX)
+	@$(call print, "Linting integration tests.")
+	$(DOCKER_TOOLS) $(GOLANGCI_LINT) $(LINT_FIX) \
+		--build-tags="itest $(DEV_TAGS) nolog" ./itest
+	@$(call print, "Linting SQLite DB integration tests.")
+	$(DOCKER_TOOLS) $(GOLANGCI_LINT) $(LINT_FIX) \
+		--build-tags="itest $(DEV_TAGS) $(LOG_TAGS)" \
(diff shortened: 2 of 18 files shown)


.golangci.yml
```diff
@@ -169,6 +169,14 @@ linters:
         linters:
           - staticcheck
 
+      # DB integration tests often share a backend fixture across subtests, and
+      # forcing parallel subtests makes those ordering assumptions unsafe.
+      - path: wallet/internal/db/itest/.*_test\.go
+        linters:
+          - paralleltest
+          - tparallel
+          - thelper
```
yyforyongyu: The justification comment above ("DB integration tests often share a backend fixture across subtests, and forcing parallel subtests makes those ordering assumptions unsafe") explains why `paralleltest` and `tparallel` are disabled, but `thelper` is unrelated — it lints test helper functions that don't call `t.Helper()` at the start, which has nothing to do with subtest parallelism. Many of the touched files already call `t.Helper()` in their helpers (e.g. `newDerivedAddress`, `getAccountByName`, `collectAddressPages` all gained or already had it). Either `thelper` should stay enabled (since the codebase already complies), or the comment should be extended with the actual reason it's excluded so a future contributor doesn't restore it and silently lose helper-stack-trace clarity.
(`claude-opus-4-7[1m]`)

itest/main_test.go
```diff
@@ -60,6 +62,8 @@ func TestBtcWallet(t *testing.T) {
 		t.Skip("no integration test cases registered")
 	}
 
+	t.Parallel()
```
yyforyongyu: This `t.Parallel()` addition is the only behavior change in the PR — everything else is lint-driven cosmetic cleanup or Makefile tooling. It's not mentioned in the PR description (which only talks about extending lint coverage and the 4-worker default) and it's not what the commit message `itest: fix tagged lint issues` describes. In practice `TestBtcWallet` looks like the only top-level `Test*` in the `itest` package so the effect is largely a no-op, but the change still deserves either its own commit (`itest: opt TestBtcWallet into parallel execution`) with a justification, or removal. Mixing it in with the lint cleanup makes the diff harder to bisect if e2e timing regresses later.
(`claude-opus-4-7[1m]`)

yyforyongyu:
LGTM 🎯

`k2p6`

yyforyongyu:
LGTM 🛠️ (`openai/gpt-5.3-codex`)

Makefile
```diff
@@ -26,10 +26,29 @@ XARGS := xargs -L 1
...
+	$(DOCKER_TOOLS) $(GOLANGCI_LINT) $(LINT_FIX)
+	@$(call print, "Linting integration tests.")
+	$(DOCKER_TOOLS) $(GOLANGCI_LINT) $(LINT_FIX) \
+		--build-tags="itest $(DEV_TAGS) nolog" ./itest
+	@$(call print, "Linting SQLite DB integration tests.")
+	$(DOCKER_TOOLS) $(GOLANGCI_LINT) $(LINT_FIX) \
+		--build-tags="itest $(DEV_TAGS) $(LOG_TAGS)" \
+		$(LINT_DB_SQLITE_PKGS)
+	@$(call print, "Linting PostgreSQL DB integration tests.")
+	$(DOCKER_TOOLS) $(GOLANGCI_LINT) $(LINT_FIX) \
```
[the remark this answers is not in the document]
yyforyongyu: Verified this one is a false positive. GNU make canned recipes allow command prefixes inside the `define` body, and both local and CI `make lint-check` runs execute this macro successfully. (`openai/gpt-5.5`)

.golangci.yml
yyforyongyu: Addressed by keeping `thelper` enabled for DB itests and fixing the helper functions it surfaced. The DB exclusion now only covers `paralleltest` and `tparallel`, which match the ordering/parallelism rationale in the comment. (`openai/gpt-5.5`)

itest/main_test.go
yyforyongyu: Addressed by removing the `t.Parallel()` addition, so the e2e test behavior is unchanged. The e2e lint context now has a narrow `paralleltest` exclusion with a comment explaining the ordered harness execution. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🧹
(`claude-opus-4-7[1m]`)

GustavoStingelin (approved):
LGTM! 🛡️🛡️🛡️ 

Thanks for this!

PR #1228, opened 2026-05-13
db: prepare kvdb store routing

yyforyongyu:
## Summary
- Add shared account/address validation and embed account/address store APIs into `db.Store`.
- Thread the kvdb address store through wallet store construction.
- Add the wallet keyvault bridge and align address metadata contracts for later store routing.

## Tests
- `GOWORK=off go test ./wallet/internal/db/kvdb ./wallet/internal/db/sqlite ./wallet/internal/db/pg ./wallet`

### DIFF
--- a/wallet/internal/db/accounts_common.go
+++ b/wallet/internal/db/accounts_common.go
@@ -43,6 +43,36 @@ func (params *CreateImportedAccountParams) Validate() error {
 	return nil
 }
 
+// Validate checks that exactly one account selector was set.
+func (query GetAccountQuery) Validate() error {
+	if query.Name == nil && query.AccountNumber == nil {
+		return ErrInvalidAccountQuery
+	}
+
+	if query.Name != nil && query.AccountNumber != nil {
+		return ErrInvalidAccountQuery
+	}
+
+	return nil
+}
+
+// Validate checks that the rename parameters identify exactly one account.
+func (params RenameAccountParams) Validate() error {
+	if params.NewName == "" {
+		return ErrMissingAccountName
+	}
+
+	if params.OldName == "" && params.AccountNumber == nil {
+		return ErrInvalidAccountQuery
+	}
+
+	if params.OldName != "" && params.AccountNumber != nil {
+		return ErrInvalidAccountQuery
+	}
+
+	return nil
+}
+
 // IsWatchOnly returns true if the account is watch-only.
 func (params *CreateImportedAccountParams) IsWatchOnly() bool {
 	return len(params.EncryptedPrivateKey) == 0
@@ -382,16 +412,16 @@ func GetAccountByQuery(ctx context.Context, query GetAccountQuery,
 	getByNumber GetAccountFunc, getByName GetAccountFunc) (*AccountInfo,
 	error) {
 
-	switch {
-	case query.AccountNumber != nil && query.Name == nil:
-		return getByNumber(ctx, query)
-
-	case query.Name != nil && query.AccountNumber == nil:
-		return getByName(ctx, query)
-
-	default:
-		return nil, ErrInvalidAccountQuery
+	err := query.Validate()
+	if err != nil {
+		return nil, err
 	}
+
+	if query.AccountNumber != nil {
+		return getByNumber(ctx, query)
+	}
+
+	return getByName(ctx, query)
 }
 
 // ListAccountsFunc defines a function signature for listing accounts.
@@ -428,20 +458,16 @@ type RenameAccountFunc func(context.Context, RenameAccountParams) error
 func RenameAccountByQuery(ctx context.Context, params RenameAccountParams,
 	renameByNumber RenameAccountFunc, renameByName RenameAccountFunc) error {
 
-	if params.NewName == "" {
-		return ErrMissingAccountName
+	err := params.Validate()
+	if err != nil {
+		return err
 	}
 
-	switch {
-	case params.AccountNumber != nil && params.OldName == "":
+	if params.AccountNumber != nil {
 		return renameByNumber(ctx, params)
-
-	case params.OldName != "" && params.AccountNumber == nil:
-		return renameByName(ctx, params)
-
-	default:
-		return ErrInvalidAccountQuery
 	}
+
+	return renameByName(ctx, params)
 }
 
 // GetAccount is a generic helper that retrieves an account using the provided
--- a/wallet/internal/db/interface.go
+++ b/wallet/internal/db/interface.go
@@ -121,13 +121,14 @@ var (
 // Until we break the wallet into independent components, we use this monolithic
 // Store abstraction as a transitional step.
 //
-// For this PR, Store includes AccountStore, UTXOStore, and TxStore. Over time
-// it is expected to grow to include WalletStore and AddressStore as those
+// For this PR, Store includes AccountStore, AddressStore, UTXOStore, and
+// TxStore. Over time it is expected to grow to include WalletStore as those
 // callers migrate to the new internal db interfaces.
 //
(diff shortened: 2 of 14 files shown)


wallet/internal/db/kvdb/utxostore.go
```diff
@@ -14,7 +13,7 @@ import (
 
 var (
 	// errNotImplemented is returned for unimplemented kvdb store methods.
-	errNotImplemented = errors.New("not implemented")
+	errNotImplemented = db.ErrNotImplemented
```
yyforyongyu: we should not change this

wallet/internal/db/interface.go
```diff
@@ -21,6 +21,10 @@ var (
 	// in the database.
 	ErrSecretNotFound = errors.New("secret not found")
 
+	// ErrNotImplemented is returned by transitional backends for store methods
+	// that do not have a backend-native implementation yet.
+	ErrNotImplemented = errors.New("not implemented")
```
yyforyongyu: we should not create this error - the `errNotImplemented` is meant to be used temporarily as we build the PRs, eventually it will be removed

yyforyongyu:

wallet/internal/db/accounts_common.go
```diff
@@ -43,6 +43,36 @@ func (params *CreateImportedAccountParams) Validate() error {
 	return nil
 }
 
+// Validate checks that exactly one account selector was set.
+func (query GetAccountQuery) Validate() error {
```
yyforyongyu, line 47: `GetAccountQuery.Validate()` and `RenameAccountParams.Validate()` are dead code on this branch — no caller invokes them, and `GetAccountByQuery` (`accounts_common.go:411`) and `RenameAccountByQuery` (`accounts_common.go:458`) already inline the exact same selector logic with `ErrInvalidAccountQuery` / `ErrMissingAccountName`. Either delete the new methods until a follow-up PR needs them, or thread them through the existing helpers so the validation lives in one place — otherwise the next contributor will edit one copy and skew them.

`claude-opus-4-7[1m]`

wallet/internal/db/kvdb/driver.go
```diff
@@ -12,19 +12,27 @@ import (
...
-	db      walletdb.DB
-	txStore wtxmgr.TxStore
+	db        walletdb.DB
+	txStore   wtxmgr.TxStore
+	addrStore legacyAddrStore
 }
 
+// legacyAddrStore is the legacy address-manager dependency threaded into kvdb
+// for account and address store adapter implementations.
+type legacyAddrStore interface{}
```
yyforyongyu: `legacyAddrStore interface{}` accepts any value (including a `string`, `int`, or `nil`), erasing the compile-time guarantee that the field actually wraps a `*waddrmgr.Manager`. Future PRs (already stacked: #1215, #1214) will need to add methods to this interface; once that happens, every existing call site has to either survive a runtime type-assertion or recompile against the now-non-empty interface. Either drop the field until the PR that adds the methods, or define the interface here with the methods it will need (e.g., `FetchScopedKeyManager`, `Decrypt`, …) so type safety holds across the stack.

`claude-opus-4-7[1m]`

wallet/wallet.go
```diff
@@ -339,6 +340,9 @@ type Wallet struct {
 	// querying the wallet's transaction history and unspent outputs.
 	txStore wtxmgr.TxStore
 
+	// keyVault provides encryption and decryption for wallet key material.
+	keyVault keyvault.Vault
```
yyforyongyu, line 344: `w.keyVault` is written by both `manager.go:305` and `deprecated.go:7190` but never read anywhere in the tree (`grep -rn "w\.keyVault\|\.keyVault\." wallet/`). The `keyvault.Vault` interface itself is fine — defining the boundary up-front is a reasonable refactor — but plumbing the field into `Wallet` without a single consumer makes this a stub two PRs deep. Either land the consumer change in the same PR (e.g., the first `accountKeyEncryptor` callsite that currently does `w.addrStore.(accountKeyEncryptor)`), or hold the field addition until that consumer needs it so the code stays observably-used.

`claude-opus-4-7[1m]`

wallet/internal/db/kvdb/driver.go
```diff
@@ -12,19 +12,27 @@ import (
 // NOTE: This is a partial implementation that will be expanded as the wallet
 // UTXO manager migrates to the new db interfaces.
 type Store struct {
-	db      walletdb.DB
-	txStore wtxmgr.TxStore
+	db        walletdb.DB
+	txStore   wtxmgr.TxStore
```
yyforyongyu, line 16: Defining `legacyAddrStore` as a named `interface{}` type blocks the stacked PRs (#1214 and #1215) from defining a proper `legacyAddrStore` interface with methods in the same package. Use `any` (or `interface{}` inline) for the field and parameter instead, and let the follow-up PRs introduce the proper interface name.

`k2p6`

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -0,0 +1,42 @@
+package kvdb
```
yyforyongyu, line 1: These newly added Go files are missing the standard copyright/license header before the package declaration: `wallet/internal/db/accounts_common_test.go`, `wallet/internal/db/kvdb/accountstore.go`, `wallet/internal/db/kvdb/addressstore.go`, `wallet/internal/db/kvdb/walletstore.go`, and `wallet/internal/keyvault/keyvault.go`. Recent reviews in this repo have required the header on new Go files, so please add it consistently here as well.

(`openai/gpt-5.5`)

wallet/internal/db/accounts_common.go
yyforyongyu: Addressed in d800e3055 by routing GetAccountByQuery and RenameAccountByQuery through their Validate methods so selector validation lives in one place.

wallet/internal/db/accounts_common.go
```diff
@@ -43,6 +43,36 @@ func (params *CreateImportedAccountParams) Validate() error {
...
+		return ErrMissingAccountName
+	}
+
+	if params.OldName == "" && params.AccountNumber == nil {
+		return ErrInvalidAccountQuery
+	}
+
+	if params.OldName != "" && params.AccountNumber != nil {
+		return ErrInvalidAccountQuery
+	}
```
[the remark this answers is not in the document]
yyforyongyu: Keeping ErrInvalidAccountQuery shared here because the selector contract is intentionally the same for get and rename. Adding a public rename-specific sentinel in this prep branch would add API churn without changing caller handling, and the error doc already covers both call sites.

wallet/wallet.go
yyforyongyu: Keeping the Wallet keyVault wiring in this prep PR because the immediate account-store branch (#1215) consumes it when routing imported accounts. This keeps the constructor wiring and key-material boundary in the shared base for the stacked store-routing changes.

wallet/internal/db/kvdb/driver.go
yyforyongyu: Addressed in d800e3055 by using any for the prep field and constructor parameter, leaving the follow-up branches to introduce typed legacy interfaces only when methods are needed.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Addressed in d800e3055 by adding the standard copyright/license header to the new Go files.

yyforyongyu:

yyforyongyu:
LGTM 🪜
(`claude-opus-4-7[1m]`)

GustavoStingelin (approved):
LGTM!

PR #1229, opened 2026-05-14
wallet: migrate wallet metadata to db.Store

yyforyongyu:
## Summary
- add kvdb implementations for the wallet metadata store methods
- route wallet metadata reads and updates through `w.store`
- update wallet mocks and tests for the store-backed metadata flow

## Why
- wallet metadata still used direct legacy wallet database access
- routing it through `db.Store` continues the incremental kvdb migration after the address manager
- keeping this in its own stacked PR makes the following UTXO-manager PR reviewable on top of a wallet metadata boundary

## What Changed
- implement kvdb `GetWallet` and `UpdateWallet`
- route wallet controller metadata paths through the store abstraction
- update mocks and add kvdb wallet-store coverage for metadata reads and updates

## Testing
- Not run in this step; draft PR opened to expose the stack.

### DIFF
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -242,6 +242,20 @@ func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
 		fingerprint = props.MasterKeyFingerprint
 	}
 
+	scope := db.KeyScope(props.KeyScope)
+	addrSchema := db.ScopeAddrMap[scope]
+
+	if props.AddrSchema != nil {
+		override, err := db.ScopeAddrSchemaFromWaddrmgr(*props.AddrSchema)
+		if err != nil {
+			log.Errorf("propertiesToAccountInfo: skipping invalid "+
+				"AddrSchema override (%v); falling back to scope "+
+				"default", err)
+		} else {
+			addrSchema = override
+		}
+	}
+
 	return db.AccountInfo{
 		AccountNumber:        accountNumber,
 		AccountName:          props.AccountName,
@@ -250,7 +264,8 @@ func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
 		InternalKeyCount:     props.InternalKeyCount,
 		ImportedKeyCount:     props.ImportedKeyCount,
 		IsWatchOnly:          isWatchOnly,
-		KeyScope:             db.KeyScope(props.KeyScope),
+		KeyScope:             scope,
+		AddrSchema:           addrSchema,
 		PublicKey:            pubKey,
 		MasterKeyFingerprint: fingerprint,
 		ConfirmedBalance:     total,
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -603,12 +603,26 @@ func (w *Wallet) verifyBirthday(ctx context.Context) error {
 		return fmt.Errorf("locate birthday block: %w", err)
 	}
 
-	err = w.DBPutBirthdayBlock(ctx, *newBirthdayBlock)
+	storeBlock, err := db.BlockFromBlockStamp(*newBirthdayBlock)
+	if err != nil {
+		return fmt.Errorf("block from stamp: %w", err)
+	}
+
+	// Use walletInfo.ID instead of w.cfg's cached value: Manager.Load
+	// currently initializes the in-memory id to zero, but the store row
+	// we just read carries the authoritative wallet ID.
+	err = w.store.UpdateWallet(
+		ctx, db.UpdateWalletParams{
+			WalletID:      walletInfo.ID,
+			BirthdayBlock: storeBlock,
+			SyncedTo:      storeBlock,
+		},
+	)
 	if err != nil {
 		log.Errorf("Unable to sanity check wallet birthday "+
 			"block: %v", err)
 
-		return err
+		return fmt.Errorf("update birthday block: %w", err)
 	}
 
 	w.birthdayBlock = *newBirthdayBlock
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
 )
 
 const (
@@ -567,24 +568,23 @@ func (w *Wallet) mainLoop() {
 //     wallet's sync tip to this point to ensure a clean rescan range.
 //  6. Update the memory cache.
 func (w *Wallet) verifyBirthday(ctx context.Context) error {
-	// We'll start by fetching our wallet's birthday block.
-	birthdayBlock, verified, err := w.DBGetBirthdayBlock(ctx)
+	walletInfo, err := w.store.GetWallet(ctx, w.cfg.Name)
 	if err != nil {
-		var mgrErr waddrmgr.ManagerError
-		if !errors.As(err, &mgrErr) ||
-			mgrErr.ErrorCode != waddrmgr.ErrBirthdayBlockNotSet {
+		log.Errorf("Unable to sanity check wallet birthday block: %v", err)
 
-			log.Errorf("Unable to sanity check wallet birthday "+
-				"block: %v", err)
-
-			return err
-		}
-		// If not set, we proceed to locate it.
+		return fmt.Errorf("get wallet birthday: %w", err)
 	}
 
 	// If the birthday block has already been verified, we initialize the
 	// cache and exit our sanity check to avoid redundant lookups.
-	if verified {
+	if walletInfo.BirthdayBlock != nil {
+		birthdayBlock, err := db.BlockStampFromBlock(
+			walletInfo.BirthdayBlock,
+		)
+		if err != nil {
+			return fmt.Errorf("decode birthday block: %w", err)
+		}
+
 		log.Infof("Birthday block verified: height=%d, hash=%v",
 			birthdayBlock.Height, birthdayBlock.Hash)
 		w.birthdayBlock = birthdayBlock
@@ -593,14 +593,14 @@ func (w *Wallet) verifyBirthday(ctx context.Context) error {
 	}
 	// Otherwise, we'll attempt to locate a better one now that we have
 	// access to the chain.
-	timestamp := w.addrStore.Birthday()
+	timestamp := walletInfo.Birthday
 
 	newBirthdayBlock, err := locateBirthdayBlock(w.cfg.Chain, timestamp)
 	if err != nil {
 		log.Errorf("Unable to sanity check wallet birthday "+
 			"block: %v", err)
 
-		return err
+		return fmt.Errorf("locate birthday block: %w", err)
 	}
 
 	err = w.DBPutBirthdayBlock(ctx, *newBirthdayBlock)
--- a/wallet/internal/db/accounts_common.go
+++ b/wallet/internal/db/accounts_common.go
@@ -314,7 +314,7 @@ func (params RenameAccountParams) Validate() error {
 
 // AccountPropsRow represents the raw database fields needed to construct
 // AccountInfo.
-type AccountPropsRow[AddrTypeId, AccOriginId any] struct {
+type AccountPropsRow[AddrTypeId ~int16 | ~int64, AccOriginId any] struct {
 	AccountNumber     sql.NullInt64
 	AccountName       string
 	OriginID          AccOriginId
@@ -329,7 +329,6 @@ type AccountPropsRow[AddrTypeId, AccOriginId any] struct {
 	CoinType          int64
 	InternalTypeID    AddrTypeId
 	ExternalTypeID    AddrTypeId
-	IDToAddrType      func(AddrTypeId) (AddressType, error)
 	IDToOriginType    func(AccOriginId) (AccountOrigin, error)
 }
 
@@ -402,10 +401,8 @@ func DerivedAddressAccountSchema[AddrTypeID ~int16 | ~int64](
 }
 
 // AccountPropsRowToInfo converts a database row containing full account
-// properties into an AccountInfo struct. The idToAddrType function is
-// used to convert the internal and external address type IDs to AddressType
-// values.
-func AccountPropsRowToInfo[AddrTypeId, AccOriginId any](
+// properties into an AccountInfo struct.
+func AccountPropsRowToInfo[AddrTypeId ~int16 | ~int64, AccOriginId any](
 	row AccountPropsRow[AddrTypeId, AccOriginId]) (*AccountInfo, error) {
 
 	var accountNum uint32
@@ -551,7 +548,7 @@ func IDToAccountOrigin[T ~int16 | ~int64](v T) (AccountOrigin, error) {
 
 // AccountInfoRow represents the raw database fields needed to construct
 // AccountInfo.
-type AccountInfoRow[AccOriginId any] struct {
+type AccountInfoRow[AccOriginId ~int16 | ~int64] struct {
 	AccountNumber      sql.NullInt64
 	AccountName        string
 	OriginID           AccOriginId
@@ -571,7 +568,7 @@ type AccountInfoRow[AccOriginId any] struct {
 
 // AccountRowToInfo converts raw database field values into an AccountInfo
 // struct. It handles type conversion and validation for each field.
-func AccountRowToInfo[AccOriginId any](
+func AccountRowToInfo[AccOriginId ~int16 | ~int64](
 	row AccountInfoRow[AccOriginId]) (*AccountInfo, error) {
 
 	var accountNum uint32
--- a/wallet/internal/db/accounts_common.go
+++ b/wallet/internal/db/accounts_common.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcwallet/waddrmgr"
 )
 
 var (
@@ -89,10 +90,14 @@ type CreateDerivedAccountOps interface {
 	// watch-only mode.
 	WalletWatchOnly(ctx context.Context, walletID uint32) (bool, error)
 
-	// EnsureScope returns the existing or newly created key-scope row ID for
-	// the wallet/scope pair.
+	// EnsureScope returns the existing or newly created key-scope row ID
+	// for the wallet/scope pair together with the schema persisted for that
+	// scope. The persisted schema may differ from ScopeAddrMap when the
+	// scope was originally created with a non-default override (e.g. an
+	// imported account that overrode the BIP44 / BIP49 / BIP84 / BIP86
+	// defaults).
 	EnsureScope(ctx context.Context, walletID uint32,
-		scope KeyScope) (int64, error)
+		scope KeyScope) (int64, ScopeAddrSchema, error)
 
 	// AllocateAccountNumber advances and returns the next derived account
 	// number for the provided scope row.
@@ -211,7 +216,9 @@ func CreateDerivedAccountWithOps(ctx context.Context,
 		return nil, fmt.Errorf("wallet watch only: %w", err)
 	}
 
-	scopeID, err := ops.EnsureScope(ctx, params.WalletID, params.Scope)
+	scopeID, addrSchema, err := ops.EnsureScope(
+		ctx, params.WalletID, params.Scope,
+	)
 	if err != nil {
 		return nil, fmt.Errorf("ensure scope: %w", err)
 	}
@@ -249,7 +256,7 @@ func CreateDerivedAccountWithOps(ctx context.Context,
 
 	return BuildAccountInfo(
 		accNumber, params.Name, DerivedAccount, 0, 0, 0,
-		walletIsWatchOnly, row.CreatedAt, params.Scope,
+		walletIsWatchOnly, row.CreatedAt, params.Scope, addrSchema,
 		derived.PublicKey, derived.MasterKeyFingerprint,
 		0, 0,
 	), nil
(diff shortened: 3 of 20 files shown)


yyforyongyu:
LGTM 🚀 (`openai/gpt-5.5`)

wallet/controller.go
```diff
@@ -593,29 +593,72 @@ func (w *Wallet) verifyBirthday(ctx context.Context) error {
...
 	}
 
 	w.birthdayBlock = *newBirthdayBlock
 
 	return nil
 }
 
+// storeBlockToBlockStamp converts database block metadata into the legacy
+// block-stamp shape still used by wallet runtime state.
+func storeBlockToBlockStamp(block *db.Block) (waddrmgr.BlockStamp, error) {
```
yyforyongyu: `storeBlockToBlockStamp` here is byte-identical (same body, same error message, same `db.Uint32ToInt32` cast) to `kvdbBlockToBlockStamp` in `wallet/internal/db/kvdb/walletstore.go:298`. The reverse-direction pair (`blockStampToStoreBlock` here + `kvdbBlockFromBlockStamp` in the kvdb file) differs only in the trailing `.UTC()` on `Timestamp`. Both directions ought to live once in the `db` package (`db.BlockStampToBlock` / `db.BlockFromStamp`) so future stores aren't tempted to fork a third copy. Same shape as the `nullableStringNS` hoist on PR #1240 (now `db.NullableStringToSQLNullString`).

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -41,15 +106,101 @@ func (s *Store) IterWallets(ctx context.Context,
...
+		if ns == nil {
+			return errMissingAddrmgrNamespace
+		}
+
+		err := updateWalletBirthday(ns, addrStore, params)
+		if err != nil {
+			return err
+		}
+
+		err = updateWalletBirthdayBlock(ns, addrStore, params)
```
yyforyongyu: Calling `updateWalletBirthdayBlock` before `updateWalletSyncedTo` makes non-zero birthday updates fail because `SetSyncedTo` requires the previous block hash once a birthday block exists; write the sync tip first or add the required prior hash before setting the birthday block. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -13,18 +16,80 @@ import (
...
+	if errors.Is(err, db.ErrBlockNotFound) {
+		syncedTo = nil
+	} else if err != nil {
+		return nil, fmt.Errorf("kvdb.Store.GetWallet: %w", err)
+	}
 
-	return nil, notImplemented(ctx, "GetWallet")
+	return &db.WalletInfo{
+		ID:            0,
+		Name:          name,
```
yyforyongyu, line 82: This echoes any requested name, so `GetWallet("non-existent")` succeeds instead of returning `db.ErrWalletNotFound`; validate against the configured single wallet name or document and test the kvdb-specific behavior. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -41,15 +106,101 @@ func (s *Store) IterWallets(ctx context.Context,
...
+		_ = yield(db.WalletInfo{}, notImplemented(ctx, "IterWallets"))
 	}
 }
 
-// UpdateWallet is not yet implemented for kvdb.
-func (s *Store) UpdateWallet(ctx context.Context,
-	_ db.UpdateWalletParams) error {
+// UpdateWallet writes wallet runtime metadata through the legacy address
+// manager.
+func (s *Store) UpdateWallet(_ context.Context,
```
yyforyongyu, line 114: `params.WalletID` is never checked, so an update for an unknown wallet ID mutates the single kvdb wallet instead of returning `db.ErrWalletNotFound`; reject IDs other than the supported kvdb wallet ID. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🔭
(`claude-opus-4-7[1m]`)

wallet/mock_test.go
```diff
@@ -67,12 +73,48 @@ func (m *mockStore) CreateWallet(ctx context.Context,
...
+		}
+	}
+
+	info := &db.WalletInfo{
+		Name:          name,
+		BirthdayBlock: birthdayBlock,
+	}
+
+	if birthdayBlock == nil {
+		info.Birthday = m.addrStore.Birthday()
```
yyforyongyu: The real kvdb `GetWallet` unconditionally populates `info.Birthday` from the address store, but this mock only sets `Birthday` when `birthdayBlock == nil`. Consider setting it unconditionally to match the contract. (`deepseek-v4-pro`)

wallet/mock_test.go
```diff
@@ -109,9 +151,45 @@ func (m *mockStore) IterWallets(ctx context.Context,
...
+			Height:    height,
+			Hash:      params.SyncedTo.Hash,
+			Timestamp: params.SyncedTo.Timestamp,
+		}
+
+		return m.addrStore.SetSyncedTo(nil, &block)
+	}
+
+	return nil
 }
```
yyforyongyu: The real kvdb `UpdateWallet` calls `updateWalletBirthday` when `params.Birthday != nil`, but this mock fallback path does not handle `params.Birthday` at all. Consider adding a `SetBirthday` call to cover the full contract. (`deepseek-v4-pro`)

wallet/mock_test.go
```diff
@@ -39,6 +39,11 @@ import (
 // exercising a real database backend.
 type mockStore struct {
 	mock.Mock
+
+	// addrStore optionally backs GetWallet so the kvdb adapter pattern can
+	// be exercised end-to-end without programming a Called expectation for
+	// every field. When nil the method falls back to mock.Called.
+	addrStore waddrmgr.AddrStore
```
yyforyongyu: we should never embed interface here - instead, we should implement the interface on the `mockStore`?

wallet/mock_test.go
```diff
@@ -454,6 +495,18 @@ func (m *mockStore) RollbackToBlock(ctx context.Context, height uint32) error {
 	return args.Error(0)
 }
 
+// hasExpectation reports whether the mock has an On(...) expectation
+// registered for the given method name.
+func (m *mockStore) hasExpectation(method string) bool {
```
yyforyongyu: can we remove this and stick to the standard `m.On(xxx)`? i don't think we should customize the `mock` library here.

wallet/internal/db/kvdb/walletstore.go
yyforyongyu: for legacy wallet, the `WalletID` should be skipped as there is only ever one wallet.

yyforyongyu: in addition, once the feature branch is merged, we will not support creating new wallets via the `kvdb` path.

wallet/mock_test.go
yyforyongyu: and i think we already have `mockAddrStore` that can be used?

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -41,15 +99,103 @@ func (s *Store) IterWallets(ctx context.Context,
...
+
+// updateWalletBirthday writes the wallet birthday timestamp when requested.
+func updateWalletBirthday(ns walletdb.ReadWriteBucket,
+	addrStore waddrmgr.AddrStore, params db.UpdateWalletParams) error {
+
+	if params.Birthday == nil {
+		return nil
+	}
+
+	err := addrStore.SetBirthday(ns, params.Birthday.UTC())
```
yyforyongyu: This mutates the in-memory birthday before later birthday-block or synced-to updates can fail and roll back the DB transaction, so validate/convert the remaining params first or restore the prior birthday on error. (`gpt-5.5`)

wallet/internal/db/kvdb/log.go
```diff
@@ -0,0 +1,22 @@
+package kvdb
```
yyforyongyu, line 1: this should be put in its own commit? we can add one commit at the beginning of the branch?

yyforyongyu (commented):
LGTM 🦉 (`deepseek-v4-pro`)

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -103,11 +108,160 @@ func (s *Store) IterWallets(ctx context.Context,
...
+		// A missing birthday block is normal before initial sync finishes.
+
+	default:
+		return fmt.Errorf("snapshot birthday block: %w", err)
+	}
+
+	birthdayMayMutate := params.Birthday != nil
+	syncedToMutated := false
+
+	defer func() {
```
yyforyongyu: i'd rather have a private method created and call it instead of inlining the code, look nasty. In addition, we are missing inline docs here explaining why the defer.

yyforyongyu:
LGTM 🧭
(`claude-opus-4-7[1m]`)

wallet/log.go
```diff
@@ -36,6 +37,7 @@ func UseLogger(logger btclog.Logger) {
 
 	migration.UseLogger(logger)
 	waddrmgr.UseLogger(logger)
+	kvdb.UseLogger(logger)
```
yyforyongyu, line 40: this should be put into the same commit where we introduce the logging for kvdb

yyforyongyu (commented):
LGTM 🦉 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🧱
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/accountstore_test.go
```diff
@@ -287,7 +287,7 @@ func TestRenameAccountRejectsImported(t *testing.T) {
 	}
 	name := "imported-rename"
 
-	_, err = store.CreateImportedAccount(t.Context(),
+	info, err := store.CreateImportedAccount(t.Context(),
```
yyforyongyu: the param should be in a newline

wallet/internal/db/kvdb/accountstore_test.go
```diff
@@ -328,7 +328,7 @@ func TestRenameAccountRejectsImported(t *testing.T) {
 	})
 	require.ErrorIs(t, err, db.ErrAccountNotFound)
 
-	info, err := store.GetAccount(t.Context(), db.GetAccountQuery{
+	info, err = store.GetAccount(t.Context(), db.GetAccountQuery{
```
yyforyongyu: the param should be in a newline

wallet/internal/db/kvdb/accountstore_test.go
```diff
@@ -732,6 +732,14 @@ func TestCreateImportedAccountUsesAddrSchema(t *testing.T) {
 		},
 	)
 	require.NoError(t, err)
+	require.Equal(t, addrSchema, info.AddrSchema)
+
+	info, err = store.GetAccount(t.Context(), db.GetAccountQuery{
```
yyforyongyu: the param should be in a newline

wallet/controller.go
```diff
@@ -603,12 +603,21 @@ func (w *Wallet) verifyBirthday(ctx context.Context) error {
 		return fmt.Errorf("locate birthday block: %w", err)
 	}
 
-	err = w.DBPutBirthdayBlock(ctx, *newBirthdayBlock)
+	storeBlock, err := db.BlockFromBlockStamp(*newBirthdayBlock)
+	if err != nil {
+		return err
+	}
+
+	err = w.store.UpdateWallet(ctx, db.UpdateWalletParams{
```
yyforyongyu: the param should be in a newline

wallet/internal/db/data_types.go
```diff
@@ -469,6 +522,17 @@ type ScopeAddrSchema struct {
 	InternalAddrType AddressType
 }
 
+// ScopeAddrSchemaFromWaddrmgr converts a legacy address-manager schema into
+// the database account schema shape.
+func ScopeAddrSchemaFromWaddrmgr(
+	schema waddrmgr.ScopeAddrSchema) ScopeAddrSchema {
+
```
yyforyongyu: `waddrmgr.AddressType` and `db.AddressType` do not share ordinal values (`waddrmgr.PubKeyHash == 0`, while `db.PubKeyHash == 1`), so this cast reports BIP44 account schemas as `RawPubKey`. Use an explicit mapping instead of direct conversion. (`gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -664,6 +664,15 @@ func loadAccountInfo(ns walletdb.ReadBucket,
 		accountNumberForContract = 0
 	}
 
+	addrSchema, ok := db.ScopeAddrMap[db.KeyScope(scope)]
```
yyforyongyu: This checks `ScopeAddrMap` before honoring `props.AddrSchema`, so an imported account persisted in a custom scope with its own schema still fails to reload with `ErrUnknownKeyScope`. Use the account override first, then fall back to the default scope map. (`gpt-5.5`)

wallet/internal/db/accounts_common.go
```diff
@@ -247,9 +247,14 @@ func CreateDerivedAccountWithOps(ctx context.Context,
 		return nil, fmt.Errorf("%w: %w", ErrMaxAccountNumberReached, err)
 	}
 
+	addrSchema, err := getAddrSchemaForScope(params.Scope)
```
yyforyongyu: If this scope already exists with a non-default persisted schema, `EnsureScope` returns only the row ID and this recomputes from `ScopeAddrMap`, so the returned `AccountInfo.AddrSchema` can disagree with the stored scope. Return or reload the persisted schema instead. (`gpt-5.5`)

wallet/controller.go
```diff
@@ -593,29 +593,40 @@ func (w *Wallet) verifyBirthday(ctx context.Context) error {
...
+	}
+
+	storeBlock, err := db.BlockFromBlockStamp(*newBirthdayBlock)
+	if err != nil {
 		return err
 	}
 
-	err = w.DBPutBirthdayBlock(ctx, *newBirthdayBlock)
+	err = w.store.UpdateWallet(ctx, db.UpdateWalletParams{
+		WalletID:      w.id,
```
yyforyongyu: `GetWallet` just loaded the authoritative wallet row by name, but the update still uses the cached `w.id`, which `Manager.Load` currently initializes to 0. Stores that enforce wallet IDs will update the wrong or missing wallet when the loaded row has a non-zero ID; use `walletInfo.ID` here or refresh `w.id` from it first. (`gpt-5.5`)

yyforyongyu (commented):
LGTM 🦉 (`deepseek-v4-pro`)

wallet/internal/db/kvdb/walletstore_test.go
```diff
@@ -60,6 +64,61 @@ func newSpendableAddrMgr(t *testing.T,
...
+	return newSpendableAddrMgr(t, dbConn)
+}
+
+// someBlock returns deterministic block metadata for wallet-store tests.
+var (
+	errInducedFailure          = errors.New("induced failure")
+	errInducedVerificationStep = errors.New("induced verification-step failure")
+)
+
+func someBlock(t *testing.T, height uint32) *db.Block {
```
yyforyongyu, line 80: The `// someBlock returns deterministic block metadata for wallet-store tests.` doc on line 74 is now sandwiched against the `var (...)` block on lines 75-78 — godoc / IDE tooltips / `funcorder`-style lint will associate the comment with the var block, and `someBlock` ends up without a doc. Move the var block elsewhere (e.g. up with the package-level vars near the top of the file) so the comment stays attached to the function it describes.

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -118,13 +118,24 @@ func (o *createDerivedAccountOps) EnsureScope(_ context.Context,
...
 	}
 
 	o.scope = waddrScope
 	o.scopedMgr = scopedMgr
 
-	return 0, nil
+	// kvdb-backed derived accounts use the BIP44/49/84/86 default
+	// schemas; per-account overrides come through props.AddrSchema at
+	// load time, not at EnsureScope.
+	addrSchema, ok := db.ScopeAddrMap[scope]
```
yyforyongyu: This recomputes the schema from the global default map after fetching an existing scoped manager, so a kvdb scope created with a non-default/custom schema either reports the wrong `AddrSchema` or fails with `ErrUnknownKeyScope`. Use the persisted schema from the scoped manager instead of `ScopeAddrMap`. (`gpt-5.5`)

yyforyongyu (commented):
LGTM 🦉 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪨
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -593,6 +607,57 @@ func sanitizeAccountNumber(acctNum uint32) (uint32, error) {
...
+		schema, err := db.ScopeAddrSchemaFromWaddrmgr(*override)
+		if err != nil {
+			return db.ScopeAddrSchema{}, fmt.Errorf("convert "+
+				"override: %w", err)
+		}
+
+		return schema, nil
+	}
+
+	addrSchema, ok := db.ScopeAddrMap[db.KeyScope(scope)]
```
yyforyongyu: This fallback still ignores the persisted scoped-manager schema when the account has no per-account override, so derived accounts under a custom scope fail with `ErrUnknownKeyScope` or report the default BIP49-plus schema instead of a non-default persisted one. Pass/use `scopedMgr.AddrSchema()` here instead of `ScopeAddrMap`. (`gpt-5.5`)

yyforyongyu:
LGTM 🧿
(`claude-opus-4-7[1m]`)

yyforyongyu (commented):
LGTM 🦉 (`deepseek-v4-pro`)

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Still unresolved at `4aefd374`: `EnsureScope` now reads `scopedMgr.AddrSchema()`, but `loadAccountInfo` still calls `effectiveAddrSchema(scope, props.AddrSchema)`, and that nil-override path still falls back to `ScopeAddrMap`. (`gpt-5.5`)

yyforyongyu: Fixed in d51cd57df (rebased into the introducing commit). `effectiveAddrSchema` now falls back to `scopedMgr.AddrSchema()` (the persisted scope schema) instead of the global `db.ScopeAddrMap`, so derived accounts under custom (non-standard) scopes get the correct address types instead of hitting `ErrUnknownKeyScope` or silently picking the BIP49-plus default. `LGTM 🤖` claude-opus-4-7-1m

yyforyongyu:
LGTM 🦦 (`gpt-5.5`)

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -41,15 +105,214 @@ func (s *Store) IterWallets(ctx context.Context,
...
+			return errMissingAddrmgrNamespace
+		}
+
+		return updateWallet(ns, addrStore, params)
+	})
+	if err != nil {
+		return fmt.Errorf("kvdb.Store.UpdateWallet: %w", err)
+	}
+
+	return nil
```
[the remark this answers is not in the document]
yyforyongyu: kvdb is on the deprecation path and `walletdb.Update` commit failures (post-callback) are rare disk-full / driver-level events; the legacy waddrmgr cache-then-commit flow has the same property pre-PR. Not restructuring the deprecated backend for this edge case.

wallet/account_manager.go
```diff
@@ -242,6 +242,20 @@ func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
...
+		override, err := db.ScopeAddrSchemaFromWaddrmgr(*props.AddrSchema)
+		if err != nil {
+			log.Errorf("propertiesToAccountInfo: skipping invalid "+
+				"AddrSchema override (%v); falling back to scope "+
+				"default", err)
+		} else {
+			addrSchema = override
+		}
+	}
+
```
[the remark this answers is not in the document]
yyforyongyu: `propertiesToAccountInfo` has no production callers at this PR head — only `account_manager_test.go` and `benchmark_helpers_test.go`. Production paths route through `w.cache.{ListAccounts,GetAccount}`, which read AccountInfo from the store where the kvdb backend already uses `scopedMgr.AddrSchema()` via `effectiveAddrSchema` (per r3311260699). Test/benchmark callers all pass BIP-default scopes that are in `db.ScopeAddrMap`, so the zero-value path is unreachable.

wallet/internal/db/block_common.go
```diff
@@ -21,3 +22,48 @@ func BuildBlock(hash []byte, height uint32, timestamp int64) (*Block, error) {
...
+// store block shape, treating negative heights as missing metadata.
+func OptionalBlockFromBlockStamp(
+	block waddrmgr.BlockStamp) (*Block, error) {
+
+	if block.Height < 0 {
+		return nil, ErrBlockNotFound
+	}
+
+	return BlockFromBlockStamp(block)
+}
```
GustavoStingelin: Non-blocking: maybe just the pointer is enough? So we can just assert over `sth != nil`

GustavoStingelin (approved):
LGTM

wallet/internal/db/block_common.go
GustavoStingelin: I meant, removing the returning error

yyforyongyu: make sense - simplified

PR #1231, opened 2026-05-14
wallet+db+waddrmgr: prep for account-manager Store routing

yyforyongyu:
## Summary

Prep stack that lands the foundation pieces the upcoming
account-manager Store/cache routing PR needs, without touching any
kvdb implementations or routing yet.

Base branch: `sqldb/remove-pg-advisory-lock`.
Follow-up PR (kvdb impl + manager routing): `impl-account-manager-store`.

## What's in scope

- **waddrmgr exports + lint**: namespace key + master HD bucket readers; derived/imported account allocators with `put-with-keys`; lint cleanup on `scoped_manager.go`.
- **`db` structural changes**:
  - Collapse `AccountProperties` into `AccountInfo`.
  - Add `MasterPubKey` to `WalletInfo` (single `GetWallet` serves both the watch-only derivation path and any future `WalletInfo` cache).
  - Add the account derivation callback contract (`AccountDerivationFunc`, `DerivedAccountData`).
- **wallet scaffolding**:
  - Introduce runtime cache scaffolding.
  - Add account derivation helper (closure-based, captures pre-loaded master key material).
- **itest + lint hygiene**: clean up pre-existing lint findings; bump postgres `max_connections` budget for db-itests; exempt `wallet/internal/db/kvdb` from `wrapcheck` to match the existing pg/sqlite exemption.

## What's NOT in scope

- kvdb account-store implementations (`GetAccount`, `ListAccounts`, etc.).
- Routing `account_manager.go` paths through `Store`/cache.
- These land in the follow-up `impl-account-manager-store` PR.

## Test plan

- [ ] `make unit pkg=wallet`
- [ ] `make unit pkg=wallet/internal/db`
- [ ] `make unit pkg=wallet/internal/db/{kvdb,pg,sqlite,runtime}`
- [ ] `make itest-db db=sqlite`
- [ ] `make itest-db db=postgres`
- [ ] `make lint-check` (4× \"0 issues\" expected on the local + tagged builds)

PR #1232, opened 2026-05-15
db: populate AccountInfo balance with SkipBalance opt-out

yyforyongyu:
## Summary

Wires confirmed/unconfirmed balance through every account read by joining each account-returning SELECT against a per-account UTXO aggregation subquery. Adds a `SkipBalance` opt-out that dispatches to a query variant which omits the LEFT JOIN entirely so the database never scans the UTXO tables when the caller does not need balance.

## Commits

1. `db: thread balance fields through AccountInfoRow plumbing` — pure type widening. `AccountInfoRow` gains `ConfirmedBalance` / `UnconfirmedBalance int64`; `BuildAccountInfo` takes two `btcutil.Amount` params; `AccountRowToInfo` converts at the boundary. No SQL change, no behavior change.

2. `db: populate AccountInfo balance from SQL aggregation` — extends the 8 account-returning SELECTs (pg + sqlite) with a LEFT JOIN to a per-account UTXO sum, split into confirmed/unconfirmed by the wallet's synced height. Honors the same status filter and `u.spent_by_tx_id IS NULL` semantics as the rest of the balance code. pg + sqlite adapters thread the new sqlc fields into `AccountInfoRow`. `GetAccountPropsById` is intentionally not joined (its only caller is the post-create read where balance is always zero). New itests verify populated balance against mixed confirmed/unconfirmed UTXOs.

3. `db: dispatch account reads to NoBalance variants on SkipBalance` — adds `SkipBalance bool` on `GetAccountQuery` and `ListAccountsQuery`. Introduces 5 `NoBalance` sqlc query variants per use case that omit the balance subquery entirely; identical Row shape with hardcoded `0` balance columns. `accountListQueries` / `accountGetQueries` dispatch via a one-line `if query.SkipBalance`. New itests verify the opt-out path returns zero balance.

## Why this PR is split from #1231

The balance work was originally bundled into prep-account-manager-store. It's split out so the perf-sensitive aggregation lands on its own reviewable surface; the parent PR stays narrow.

## Test plan
- [x] `GOWORK=off go build ./...` clean across all 3 commits
- [x] `make lint-check` clean
- [x] Unit tests pass on `wallet/internal/db/...`
- [x] sqlite + pg itests (`TestGetAccount{,SkipBalance}*`, `TestListAccounts{,SkipBalance}*`) pass

### DIFF
--- a/wallet/internal/db/accounts_common.go
+++ b/wallet/internal/db/accounts_common.go
@@ -816,3 +816,95 @@ func CreateImportedAccount[CreateArgs any, CreateRow any, SecretArgs any](
 
 	return getProps(accountID)
 }
+
+// ProcessAccountRows converts a batch of dialect-specific account rows into
+// AccountInfo pointers (ordered as input). The convert closure produces both
+// the AccountInfo and the SQL row ID for each row; ProcessAccountRows stores
+// the row ID on the returned AccountInfo via the unexported rowID field so
+// AttachBalances can pair per-account balance rows back to their AccountInfo
+// without threading a parallel ids slice.
+func ProcessAccountRows[T any](rows []T,
+	convert func(T) (*AccountInfo, int64, error)) ([]*AccountInfo, error) {
+
+	infos := make([]*AccountInfo, len(rows))
+	for i := range rows {
+		info, id, err := convert(rows[i])
+		if err != nil {
+			return nil, err
+		}
+
+		info.rowID = id
+		infos[i] = info
+	}
+
+	return infos, nil
+}
+
+// AccountBalance is the dialect-agnostic shape of a per-account balance row.
+// Backends translate their sqlc balance rows into AccountBalance values when
+// calling AttachBalances; the wire-level AccountBalancesByIDs query lives in
+// each backend, and AccountBalance is the common contract its results land
+// in for the shared merge logic.
+type AccountBalance struct {
+	// AccountID is the SQL row ID of the account this balance belongs to;
+	// it matches the rowID populated on AccountInfo by ProcessAccountRows.
+	AccountID int64
+
+	// Confirmed is the sum of UTXO amounts (in satoshis) that are
+	// considered confirmed by the wallet's confirmation policy.
+	Confirmed int64
+
+	// Unconfirmed is the sum of UTXO amounts (in satoshis) that are
+	// unconfirmed.
+	Unconfirmed int64
+}
+
+// AttachBalances merges per-account balances into a batch of AccountInfo
+// values. The queryBalances callback executes the backend-specific
+// AccountBalancesByIDs and returns its results as []AccountBalance. The
+// merge step uses each AccountInfo's rowID (populated by ProcessAccountRows)
+// to map balance rows back to their AccountInfo in a single pass. The
+// returned slice preserves the input order of infos.
+//
+// When skipBalance is true or infos is empty, the query dispatch is skipped
+// and every returned AccountInfo keeps zero balance fields. The query
+// callback is invoked at most once per AttachBalances call.
+func AttachBalances(ctx context.Context, walletID uint32, skipBalance bool,
+	infos []*AccountInfo,
+	queryBalances func(ctx context.Context, walletID uint32,
+		ids []int64) ([]AccountBalance, error)) ([]AccountInfo, error) {
+
+	out := make([]AccountInfo, len(infos))
+	for i := range infos {
+		out[i] = *infos[i]
+	}
+
+	if skipBalance || len(infos) == 0 {
+		return out, nil
+	}
+
+	indexByID := make(map[int64]int, len(infos))
+	ids := make([]int64, len(infos))
+
+	for i := range infos {
+		indexByID[infos[i].rowID] = i
+		ids[i] = infos[i].rowID
+	}
+
+	balances, err := queryBalances(ctx, walletID, ids)
+	if err != nil {
+		return nil, fmt.Errorf("account balances: %w", err)
+	}
+
+	for _, bal := range balances {
+		idx, ok := indexByID[bal.AccountID]
+		if !ok {
+			continue
+		}
+
+		out[idx].ConfirmedBalance = btcutil.Amount(bal.Confirmed)
+		out[idx].UnconfirmedBalance = btcutil.Amount(bal.Unconfirmed)
+	}
+
+	return out, nil
+}
--- a/wallet/internal/db/accounts_common.go
+++ b/wallet/internal/db/accounts_common.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"fmt"
 	"time"
+
+	"github.com/btcsuite/btcd/btcutil"
 )
 
 var (
@@ -249,6 +251,7 @@ func CreateDerivedAccountWithOps(ctx context.Context,
 		accNumber, params.Name, DerivedAccount, 0, 0, 0,
 		walletIsWatchOnly, row.CreatedAt, params.Scope,
 		derived.PublicKey, derived.MasterKeyFingerprint,
+		0, 0,
 	), nil
 }
 
@@ -473,16 +476,17 @@ func getAddrSchemaForScope(scope KeyScope) (ScopeAddrSchema, error) {
 	return addrSchema, nil
 }
 
-// BuildAccountInfo creates an AccountInfo with the provided values and zeroed
-// balances while we do not yet support balance tracking.
-//
-// TODO(stingelin): Add balance tracking support after transaction management is
-// implemented.
+// BuildAccountInfo creates an AccountInfo with the provided values.
+// confirmedBalance and unconfirmedBalance are populated verbatim from
+// the caller; create paths pass zero (a fresh account has no UTXOs)
+// while the read paths feed real values from the dedicated
+// AccountBalance / AccountBalances queries.
 func BuildAccountInfo(accountNum uint32, accountName string,
 	origin AccountOrigin, externalKeyCount, internalKeyCount,
 	importedKeyCount uint32, isWatchOnly bool, createdAt time.Time,
 	scope KeyScope, publicKey []byte,
-	masterKeyFingerprint uint32) *AccountInfo {
+	masterKeyFingerprint uint32,
+	confirmedBalance, unconfirmedBalance btcutil.Amount) *AccountInfo {
 
 	return &AccountInfo{
 		AccountNumber:        accountNum,
@@ -491,8 +495,8 @@ func BuildAccountInfo(accountNum uint32, accountName string,
 		ExternalKeyCount:     externalKeyCount,
 		InternalKeyCount:     internalKeyCount,
 		ImportedKeyCount:     importedKeyCount,
-		ConfirmedBalance:     0,
-		UnconfirmedBalance:   0,
+		ConfirmedBalance:     confirmedBalance,
+		UnconfirmedBalance:   unconfirmedBalance,
 		IsWatchOnly:          isWatchOnly,
 		CreatedAt:            createdAt,
 		KeyScope:             scope,
@@ -514,19 +518,21 @@ func IDToAccountOrigin[T ~int16 | ~int64](v T) (AccountOrigin, error) {
 // AccountInfoRow represents the raw database fields needed to construct
 // AccountInfo.
 type AccountInfoRow[AccOriginId any] struct {
-	AccountNumber     sql.NullInt64
-	AccountName       string
-	OriginID          AccOriginId
-	ExternalKeyCount  int64
-	InternalKeyCount  int64
-	ImportedKeyCount  int64
-	PublicKey         []byte
-	MasterFingerprint sql.NullInt64
-	IsWatchOnly       bool
-	CreatedAt         time.Time
-	Purpose           int64
-	CoinType          int64
-	IDToOriginType    func(AccOriginId) (AccountOrigin, error)
+	AccountNumber      sql.NullInt64
+	AccountName        string
+	OriginID           AccOriginId
+	ExternalKeyCount   int64
+	InternalKeyCount   int64
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	IsWatchOnly        bool
+	CreatedAt          time.Time
+	Purpose            int64
+	CoinType           int64
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+	IDToOriginType     func(AccOriginId) (AccountOrigin, error)
 }
 
 // AccountRowToInfo converts raw database field values into an AccountInfo
@@ -579,6 +585,8 @@ func AccountRowToInfo[AccOriginId any](
 		importedKeyCount, row.IsWatchOnly, row.CreatedAt,
 		KeyScope{Purpose: purposeNum, Coin: coinTypeNum},
 		row.PublicKey, fingerprint,
+		btcutil.Amount(row.ConfirmedBalance),
+		btcutil.Amount(row.UnconfirmedBalance),
 	), nil
 }
 
--- a/wallet/internal/db/itest/account_store_test.go
+++ b/wallet/internal/db/itest/account_store_test.go
@@ -12,6 +12,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	"github.com/stretchr/testify/require"
 )
@@ -535,9 +537,11 @@ func TestCreateDerivedAccountConcurrent(t *testing.T) {
 	require.Len(t, results, workers)
 
 	// Verify all numbers are unique and sequential.
-	sort.Slice(results, func(i, j int) bool {
-		return results[i] < results[j]
-	})
+	sort.Slice(
+		results, func(i, j int) bool {
+			return results[i] < results[j]
+		},
+	)
 
 	for i := range workers {
 		require.Equal(t, uint32(i), results[i])
@@ -856,14 +860,16 @@ func TestGetAccount(t *testing.T) {
 	for _, tc := range AllAccountCases {
 		accNumber := uint32(0)
 
-		t.Run("by name-"+tc.Name, func(t *testing.T) {
-			query := getAccountQueryByName(walletID, tc.Scope, tc.Name)
-			info, err := store.GetAccount(t.Context(), query)
-			require.NoError(t, err)
-			require.NotNil(t, info)
-			requireAccountMatches(t, info, tc)
-			accNumber = info.AccountNumber
-		})
+		t.Run(
+			"by name-"+tc.Name, func(t *testing.T) {
+				query := getAccountQueryByName(walletID, tc.Scope, tc.Name)
+				info, err := store.GetAccount(t.Context(), query)
+				require.NoError(t, err)
+				require.NotNil(t, info)
+				requireAccountMatches(t, info, tc)
+				accNumber = info.AccountNumber
+			},
+		)
 
 		if tc.Origin == db.ImportedAccount {
 			continue
@@ -876,7 +882,8 @@ func TestGetAccount(t *testing.T) {
 				require.NoError(t, err)
 				require.NotNil(t, info)
 				requireAccountMatches(t, info, tc)
-			})
+			},
+		)
 	}
 }
 
@@ -990,6 +997,150 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
 	for _, acc := range accounts {
 		require.NotEmpty(t, acc.PublicKey, acc.AccountName)
 	}
+}
+
+// TestGetAccountPopulatesBalance verifies that GetAccount returns the
+// confirmed and unconfirmed UTXO totals on the AccountInfo, sourced from
+// the dedicated AccountBalance query that the adapter dispatches
+// alongside the account row fetch.
+func TestGetAccountPopulatesBalance(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-get-balance")
+	scope := db.KeyScopeBIP0084
+
+	queries := store.Queries()
+	syncBlock := CreateBlockFixture(t, queries, 200)
+	confirmedBlock := CreateBlockFixture(t, queries, 100)
+
+	err := store.UpdateWallet(
+		t.Context(), db.UpdateWalletParams{
+			WalletID: walletID,
+			SyncedTo: &syncBlock,
+		},
+	)
+	require.NoError(t, err)
+
+	createDerivedAccount(t, store, walletID, scope, "funded")
+	createDerivedAccount(t, store, walletID, scope, "empty")
+
+	addr := newDerivedAddress(t, store, walletID, scope, "funded", false)
+
+	confirmedTx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 24000, PkScript: addr.ScriptPubKey}},
+	)
+	err = store.CreateTx(
+		t.Context(), db.CreateTxParams{
+			WalletID: walletID,
+			Tx:       confirmedTx,
+			Received: time.Unix(1710000000, 0),
+			Block:    &confirmedBlock,
+			Status:   db.TxStatusPublished,
+			Credits:  map[uint32]btcutil.Address{0: nil},
+		},
+	)
+	require.NoError(t, err)
+
+	unconfirmedTx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 26000, PkScript: addr.ScriptPubKey}},
+	)
+	err = store.CreateTx(
+		t.Context(), db.CreateTxParams{
+			WalletID: walletID,
+			Tx:       unconfirmedTx,
+			Received: time.Unix(1710000100, 0),
+			Status:   db.TxStatusPending,
+			Credits:  map[uint32]btcutil.Address{0: nil},
+		},
+	)
+	require.NoError(t, err)
+
+	funded, err := store.GetAccount(
+		t.Context(), getAccountQueryByName(walletID, scope, "funded"),
+	)
+	require.NoError(t, err)
+	require.Equal(t, btcutil.Amount(24000), funded.ConfirmedBalance)
+	require.Equal(t, btcutil.Amount(26000), funded.UnconfirmedBalance)
+
+	byNumber, err := store.GetAccount(
+		t.Context(),
+		getAccountQueryByNumber(walletID, scope, funded.AccountNumber),
+	)
+	require.NoError(t, err)
+	require.Equal(t, btcutil.Amount(24000), byNumber.ConfirmedBalance)
+	require.Equal(t, btcutil.Amount(26000), byNumber.UnconfirmedBalance)
+
+	empty, err := store.GetAccount(
+		t.Context(), getAccountQueryByName(walletID, scope, "empty"),
+	)
+	require.NoError(t, err)
+	require.Equal(t, btcutil.Amount(0), empty.ConfirmedBalance)
+	require.Equal(t, btcutil.Amount(0), empty.UnconfirmedBalance)
+}
+
+// TestGetAccountSkipBalanceZerosFields verifies that GetAccount with
+// SkipBalance=true skips the dedicated AccountBalance dispatch on both
+// the by-name and by-number selectors and leaves the balance fields at
+// zero even when UTXOs exist on the account.
+func TestGetAccountSkipBalanceZerosFields(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-get-balance-skip")
+	scope := db.KeyScopeBIP0084
+
+	queries := store.Queries()
+	syncBlock := CreateBlockFixture(t, queries, 200)
+	confirmedBlock := CreateBlockFixture(t, queries, 100)
+
+	err := store.UpdateWallet(
+		t.Context(), db.UpdateWalletParams{
+			WalletID: walletID,
+			SyncedTo: &syncBlock,
+		},
+	)
+	require.NoError(t, err)
+
+	createDerivedAccount(t, store, walletID, scope, "funded")
+
+	addr := newDerivedAddress(t, store, walletID, scope, "funded", false)
+
+	tx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 24000, PkScript: addr.ScriptPubKey}},
+	)
+	err = store.CreateTx(
+		t.Context(), db.CreateTxParams{
+			WalletID: walletID,
+			Tx:       tx,
+			Received: time.Unix(1710000200, 0),
+			Block:    &confirmedBlock,
+			Status:   db.TxStatusPublished,
+			Credits:  map[uint32]btcutil.Address{0: nil},
+		},
+	)
+	require.NoError(t, err)
+
+	byNameQuery := getAccountQueryByName(walletID, scope, "funded")
+	byNameQuery.SkipBalance = true
+
+	infoByName, err := store.GetAccount(t.Context(), byNameQuery)
+	require.NoError(t, err)
+	require.Equal(t, btcutil.Amount(0), infoByName.ConfirmedBalance)
+	require.Equal(t, btcutil.Amount(0), infoByName.UnconfirmedBalance)
+
+	byNumberQuery := getAccountQueryByNumber(
+		walletID, scope, infoByName.AccountNumber,
+	)
+	byNumberQuery.SkipBalance = true
+
+	infoByNumber, err := store.GetAccount(t.Context(), byNumberQuery)
+	require.NoError(t, err)
+	require.Equal(t, btcutil.Amount(0), infoByNumber.ConfirmedBalance)
+	require.Equal(t, btcutil.Amount(0), infoByNumber.UnconfirmedBalance)
 }
 
 // TestGetAccountNotFound verifies that GetAccount returns ErrAccountNotFound
@@ -1227,10 +1378,12 @@ func TestAccountCreatedAtTimestamp(t *testing.T) {
 		)
 		require.NoError(t, err)
 
-		accounts = append(accounts, createdAccount{
-			info:        *info,
-			createdNear: createdNear,
-		})
+		accounts = append(
+			accounts, createdAccount{
+				info:        *info,
+				createdNear: createdNear,
+			},
+		)
 	}
 
 	// Verify all accounts have CreatedAt populated.
@@ -1268,12 +1421,14 @@ func TestRenameAccount(t *testing.T) {
 		oldName := "original-name-1"
 		newName := "renamed-by-name"
 
-		err := store.RenameAccount(t.Context(), db.RenameAccountParams{
-			WalletID: walletID,
-			Scope:    scope,
-			OldName:  oldName,
-			NewName:  newName,
-		})
+		err := store.RenameAccount(
+			t.Context(), db.RenameAccountParams{
+				WalletID: walletID,
+				Scope:    scope,
+				OldName:  oldName,
+				NewName:  newName,
+			},
+		)
 		require.NoError(t, err)
 
 		// Verify the rename worked.
@@ -1294,12 +1449,14 @@ func TestRenameAccount(t *testing.T) {
 		accNum := uint32(0)
 		newName := "renamed-by-number"
 
-		err := store.RenameAccount(t.Context(), db.RenameAccountParams{
-			WalletID:      walletID,
-			Scope:         scope,
-			AccountNumber: &accNum,
-			NewName:       newName,
-		})
+		err := store.RenameAccount(
+			t.Context(), db.RenameAccountParams{
+				WalletID:      walletID,
+				Scope:         scope,
+				AccountNumber: &accNum,
+				NewName:       newName,
+			},
+		)
 		require.NoError(t, err)
 
 		// Verify the rename worked.
@@ -1565,9 +1722,11 @@ func findAccountInList(t *testing.T, accounts []db.AccountInfo,
 
 	t.Helper()
 
-	i := slices.IndexFunc(accounts, func(acc db.AccountInfo) bool {
-		return acc.AccountName == tc.Name && acc.KeyScope == tc.Scope
-	})
+	i := slices.IndexFunc(
+		accounts, func(acc db.AccountInfo) bool {
+			return acc.AccountName == tc.Name && acc.KeyScope == tc.Scope
+		},
+	)
 	require.GreaterOrEqual(t, i, 0, "expected account %s in list", tc.Name)
 
 	return accounts[i]
--- a/wallet/internal/db/itest/account_store_test.go
+++ b/wallet/internal/db/itest/account_store_test.go
@@ -1143,6 +1143,215 @@ func TestGetAccountSkipBalanceZerosFields(t *testing.T) {
 	require.Equal(t, btcutil.Amount(0), infoByNumber.UnconfirmedBalance)
 }
 
+// TestListAccountsPopulatesBalance verifies that ListAccounts returns
+// confirmed/unconfirmed totals on every returned AccountInfo, sourced
+// from the AccountBalances batch query dispatched alongside the row
+// fetch.
+func TestListAccountsPopulatesBalance(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-list-balance")
+	scope := db.KeyScopeBIP0084
+
+	queries := store.Queries()
+	syncBlock := CreateBlockFixture(t, queries, 200)
+	confirmedBlock := CreateBlockFixture(t, queries, 100)
+
+	err := store.UpdateWallet(
+		t.Context(), db.UpdateWalletParams{
+			WalletID: walletID,
+			SyncedTo: &syncBlock,
+		},
+	)
+	require.NoError(t, err)
+
+	createDerivedAccount(t, store, walletID, scope, "first")
+	createDerivedAccount(t, store, walletID, scope, "second")
+	createDerivedAccount(t, store, walletID, scope, "empty")
+
+	firstAddr := newDerivedAddress(t, store, walletID, scope, "first", false)
+	secondAddr := newDerivedAddress(
+		t, store, walletID, scope, "second", false,
+	)
+
+	firstTx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 12000, PkScript: firstAddr.ScriptPubKey}},
+	)
+	err = store.CreateTx(
+		t.Context(), db.CreateTxParams{
+			WalletID: walletID,
+			Tx:       firstTx,
+			Received: time.Unix(1710000200, 0),
+			Block:    &confirmedBlock,
+			Status:   db.TxStatusPublished,
+			Credits:  map[uint32]btcutil.Address{0: nil},
+		},
+	)
+	require.NoError(t, err)
+
+	secondTx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 7000, PkScript: secondAddr.ScriptPubKey}},
+	)
+	err = store.CreateTx(
+		t.Context(), db.CreateTxParams{
+			WalletID: walletID,
+			Tx:       secondTx,
+			Received: time.Unix(1710000300, 0),
+			Status:   db.TxStatusPending,
+			Credits:  map[uint32]btcutil.Address{0: nil},
+		},
+	)
+	require.NoError(t, err)
+
+	verify := func(t *testing.T,
+		accounts []db.AccountInfo, label string) {
+
+		t.Helper()
+
+		byName := make(map[string]db.AccountInfo, len(accounts))
+		for _, acc := range accounts {
+			byName[acc.AccountName] = acc
+		}
+
+		require.Contains(t, byName, "first", label)
+		require.Equal(t, btcutil.Amount(12000),
+			byName["first"].ConfirmedBalance, label+": first")
+		require.Equal(t, btcutil.Amount(0),
+			byName["first"].UnconfirmedBalance, label+": first")
+
+		require.Contains(t, byName, "second", label)
+		require.Equal(t, btcutil.Amount(0),
+			byName["second"].ConfirmedBalance, label+": second")
+		require.Equal(t, btcutil.Amount(7000),
+			byName["second"].UnconfirmedBalance, label+": second")
+
+		require.Contains(t, byName, "empty", label)
+		require.Equal(t, btcutil.Amount(0),
+			byName["empty"].ConfirmedBalance, label+": empty")
+		require.Equal(t, btcutil.Amount(0),
+			byName["empty"].UnconfirmedBalance, label+": empty")
+	}
+
+	byScope, err := store.ListAccounts(
+		t.Context(), db.ListAccountsQuery{
+			WalletID: walletID,
+			Scope:    &scope,
+		},
+	)
+	require.NoError(t, err)
+	verify(t, byScope, "by scope")
+
+	all, err := store.ListAccounts(
+		t.Context(), db.ListAccountsQuery{
+			WalletID: walletID,
+		},
+	)
+	require.NoError(t, err)
+	verify(t, all, "all")
+}
+
+// TestListAccountsSkipBalanceZerosFields verifies that ListAccounts with
+// SkipBalance=true skips the AccountBalances dispatch on each of the
+// three list selectors (scope-filtered, name-filtered, unfiltered) and
+// returns zero balance fields even when UTXOs exist.
+func TestListAccountsSkipBalanceZerosFields(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-list-skip-balance")
+	scope := db.KeyScopeBIP0084
+
+	queries := store.Queries()
+	syncBlock := CreateBlockFixture(t, queries, 200)
+	confirmedBlock := CreateBlockFixture(t, queries, 100)
+
+	err := store.UpdateWallet(
+		t.Context(), db.UpdateWalletParams{
+			WalletID: walletID,
+			SyncedTo: &syncBlock,
+		},
+	)
+	require.NoError(t, err)
+
+	createDerivedAccount(t, store, walletID, scope, "funded")
+
+	addr := newDerivedAddress(t, store, walletID, scope, "funded", false)
+
+	tx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 9000, PkScript: addr.ScriptPubKey}},
+	)
+	err = store.CreateTx(
+		t.Context(), db.CreateTxParams{
+			WalletID: walletID,
+			Tx:       tx,
+			Received: time.Unix(1710000400, 0),
+			Block:    &confirmedBlock,
+			Status:   db.TxStatusPublished,
+			Credits:  map[uint32]btcutil.Address{0: nil},
+		},
+	)
+	require.NoError(t, err)
+
+	verify := func(t *testing.T,
+		accounts []db.AccountInfo, label string) {
+
+		t.Helper()
+
+		require.NotEmpty(t, accounts, label)
+
+		found := false
+		for _, acc := range accounts {
+			if acc.AccountName == "funded" {
+				found = true
+			}
+
+			require.Equal(t, btcutil.Amount(0),
+				acc.ConfirmedBalance,
+				label+": "+acc.AccountName)
+			require.Equal(t, btcutil.Amount(0),
+				acc.UnconfirmedBalance,
+				label+": "+acc.AccountName)
+		}
+
+		require.True(t, found,
+			label+": funded account missing from result")
+	}
+
+	byScope, err := store.ListAccounts(
+		t.Context(), db.ListAccountsQuery{
+			WalletID:    walletID,
+			Scope:       &scope,
+			SkipBalance: true,
+		},
+	)
+	require.NoError(t, err)
+	verify(t, byScope, "by scope")
+
+	name := "funded"
+	byName, err := store.ListAccounts(
+		t.Context(), db.ListAccountsQuery{
+			WalletID:    walletID,
+			Name:        &name,
+			SkipBalance: true,
+		},
+	)
+	require.NoError(t, err)
+	verify(t, byName, "by name")
+
+	all, err := store.ListAccounts(
+		t.Context(), db.ListAccountsQuery{
+			WalletID:    walletID,
+			SkipBalance: true,
+		},
+	)
+	require.NoError(t, err)
+	verify(t, all, "all")
+}
+
 // TestGetAccountNotFound verifies that GetAccount returns ErrAccountNotFound
 // when querying a non-existent account.
 func TestGetAccountNotFound(t *testing.T) {
--- a/wallet/internal/db/pg/accounts.go
+++ b/wallet/internal/db/pg/accounts.go
@@ -411,40 +411,130 @@ type accountListQueries struct {
 	q *sqlc.Queries
 }
 
-// byScope lists accounts filtered by wallet ID and key scope.
+// byScope lists accounts filtered by wallet ID and key scope, then
+// attaches each account's balance via AccountBalancesByIDs unless
+// query.SkipBalance is set.
 func (p accountListQueries) byScope(ctx context.Context,
 	query db.ListAccountsQuery) ([]db.AccountInfo, error) {
 
-	return db.ListAccounts(
-		ctx, p.q.ListAccountsByWalletScope,
-		sqlc.ListAccountsByWalletScopeParams{
+	rows, err := p.q.ListAccountsByWalletScope(
+		ctx, sqlc.ListAccountsByWalletScopeParams{
 			WalletID: int64(query.WalletID),
 			Purpose:  int64(query.Scope.Purpose),
 			CoinType: int64(query.Scope.Coin),
-		}, accountRowToInfo,
+		},
 	)
+	if err != nil {
+		return nil, fmt.Errorf("list accounts: %w", err)
+	}
+
+	infos, err := db.ProcessAccountRows(
+		rows,
+		func(r sqlc.ListAccountsByWalletScopeRow) (*db.AccountInfo, int64,
+			error) {
+
+			info, err := accountRowToInfo(r)
+			return info, r.ID, err
+		},
+	)
+	if err != nil {
+		return nil, err
+	}
+
+	return p.attachBalances(ctx, query, infos)
 }
 
-// byName lists accounts filtered by wallet ID and account name.
+// byName lists accounts filtered by wallet ID and account name, then
+// attaches each account's balance via AccountBalancesByIDs unless
+// query.SkipBalance is set.
 func (p accountListQueries) byName(ctx context.Context,
 	query db.ListAccountsQuery) ([]db.AccountInfo, error) {
 
-	return db.ListAccounts(
-		ctx, p.q.ListAccountsByWalletAndName,
-		sqlc.ListAccountsByWalletAndNameParams{
+	rows, err := p.q.ListAccountsByWalletAndName(
+		ctx, sqlc.ListAccountsByWalletAndNameParams{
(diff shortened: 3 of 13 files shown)


wallet/internal/sql/pg/sqlc/accounts.sql.go
```diff
@@ -406,6 +515,101 @@ func (q *Queries) GetAccountByWalletScopeAndName(ctx context.Context, arg GetAcc
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to GetAccountByWalletScopeAndName but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `GetAccountByWalletScopeAndNameNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/pg/sqlc/accounts.sql.go
```diff
@@ -489,6 +728,101 @@ func (q *Queries) GetAccountByWalletScopeAndNumber(ctx context.Context, arg GetA
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to GetAccountByWalletScopeAndNumber but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `GetAccountByWalletScopeAndNumberNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/pg/sqlc/accounts.sql.go
```diff
@@ -832,6 +1275,203 @@ func (q *Queries) ListAccountsByWalletAndName(ctx context.Context, arg ListAccou
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to ListAccountsByWalletAndName but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `ListAccountsByWalletAndNameNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/pg/sqlc/accounts.sql.go
```diff
@@ -832,6 +1275,203 @@ func (q *Queries) ListAccountsByWalletAndName(ctx context.Context, arg ListAccou
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to ListAccountsByWallet but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `ListAccountsByWalletNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/pg/sqlc/accounts.sql.go
```diff
@@ -926,6 +1601,112 @@ func (q *Queries) ListAccountsByWalletScope(ctx context.Context, arg ListAccount
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to ListAccountsByWalletScope but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `ListAccountsByWalletScopeNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/accounts.sql.go
```diff
@@ -406,6 +515,101 @@ func (q *Queries) GetAccountByWalletScopeAndName(ctx context.Context, arg GetAcc
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to GetAccountByWalletScopeAndName but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `GetAccountByWalletScopeAndNameNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/accounts.sql.go
```diff
@@ -489,6 +728,101 @@ func (q *Queries) GetAccountByWalletScopeAndNumber(ctx context.Context, arg GetA
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to GetAccountByWalletScopeAndNumber but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `GetAccountByWalletScopeAndNumberNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/accounts.sql.go
```diff
@@ -832,6 +1275,203 @@ func (q *Queries) ListAccountsByWalletAndName(ctx context.Context, arg ListAccou
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to ListAccountsByWalletAndName but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `ListAccountsByWalletAndNameNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/accounts.sql.go
```diff
@@ -832,6 +1275,203 @@ func (q *Queries) ListAccountsByWalletAndName(ctx context.Context, arg ListAccou
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to ListAccountsByWallet but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `ListAccountsByWalletNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/accounts.sql.go
```diff
@@ -926,6 +1601,112 @@ func (q *Queries) ListAccountsByWalletScope(ctx context.Context, arg ListAccount
...
+	ImportedKeyCount   int64
+	PublicKey          []byte
+	MasterFingerprint  sql.NullInt64
+	WalletIsWatchOnly  bool
+	IsWatchOnly        bool
+	ConfirmedBalance   int64
+	UnconfirmedBalance int64
+}
+
+// Identical to ListAccountsByWalletScope but skips the per-account UTXO
```
yyforyongyu: The generated comment does not start with the exact method name `ListAccountsByWalletScopeNoBalance`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/pg/queries/accounts.sql
```diff
@@ -76,11 +76,44 @@ SELECT
...
+            END
+        ), 0)::BIGINT AS unconfirmed_balance
+    FROM utxos AS u
+    INNER JOIN transactions AS t ON t.id = u.tx_id
+    INNER JOIN addresses AS addr ON u.address_id = addr.id
+    LEFT JOIN wallet_sync_states AS s ON t.wallet_id = s.wallet_id
+    WHERE
+        u.spent_by_tx_id IS NULL
+        AND t.tx_status IN (0, 1)
+    GROUP BY addr.account_id, addr.wallet_id
```
yyforyongyu: The balance subquery has no `wallet_id` filter in its `WHERE`, so it aggregates every spendable UTXO in the whole `utxos` table on every account-returning SELECT — even single-account reads like `GetAccountByWalletScopeAndName` where the outer query already pins `w.id`. The compiled query plan looks roughly like:

```
LEFT JOIN (
    SELECT addr.account_id, addr.wallet_id, SUM(...)
    FROM utxos u
    INNER JOIN transactions t ON t.id = u.tx_id
    INNER JOIN addresses addr ON u.address_id = addr.id
    LEFT JOIN wallet_sync_states s ON t.wallet_id = s.wallet_id
    WHERE u.spent_by_tx_id IS NULL AND t.tx_status IN (0, 1)
    GROUP BY addr.account_id, addr.wallet_id
) bal ON bal.account_id = a.id AND bal.wallet_id = a.wallet_id
```

Cost for a single-account read on a deployment with `K` wallets × `N` accounts × `M` UTXOs each is `O(K·N·M)` rows scanned per call, vs `O(M)` if the predicate were pushed in. The aggregated subquery + `GROUP BY` blocks PG's automatic predicate-pushdown heuristics (transitive inference does not run through derived tables with a `GROUP BY` reliably), and SQLite's planner is even less aggressive — so this won't be optimized away in practice.

Compare to the existing `utxos.sql` `Balance` query: same join shape but `WHERE t.wallet_id = sqlc.arg('wallet_id')` directly inside the aggregate, plus optional `account_number` / `purpose` / `coin_type` nargs — that's the precedent for keeping the filter inside the aggregate.

Two ways to fix that match the rest of the codebase:

1. Add `AND addr.wallet_id = sqlc.arg('wallet_id')` (and account/scope predicates where applicable) inside each balance subquery. Cheapest, but doubles the per-query parameter list across all 8 sites.
2. Rewrite the subquery as `LEFT JOIN LATERAL (SELECT … WHERE addr.account_id = a.id AND addr.wallet_id = a.wallet_id) bal ON true`. The lateral form makes the correlation explicit so PG can push the equality into the aggregate. SQLite supports `LATERAL` since 3.39.

The `SkipBalance` opt-out partially mitigates this for identity-only callers, but for the default path (balance populated) a wallet with even a handful of accounts pays the full-table aggregate on every `GetAccount` / `ListAccounts` read.

(`claude-opus-4-7[1m]`)

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,256 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+	require.Equal(t, btcutil.Amount(0), byName["first"].UnconfirmedBalance)
+	require.Equal(t, btcutil.Amount(0), byName["second"].ConfirmedBalance)
+	require.Equal(t, btcutil.Amount(7000), byName["second"].UnconfirmedBalance)
+}
+
+// TestGetAccountSkipBalanceZerosFields verifies that GetAccount with
+// SkipBalance=true dispatches to the NoBalance variant and leaves
+// ConfirmedBalance and UnconfirmedBalance at zero even when UTXOs
+// exist on the account.
+func TestGetAccountSkipBalanceZerosFields(t *testing.T) {
```
yyforyongyu: These new `SkipBalance` tests only cover `GetAccount` by name and `ListAccounts` by scope. The PR adds distinct dispatch branches for `GetAccount` by number, `ListAccounts` by name, and `ListAccounts` all, so those can regress to the balance query without failing the current tests. (`openai/gpt-5.5`)

wallet/internal/sql/pg/sqlc/accounts.sql.go
yyforyongyu: Fixed in 288561524 () and 9882cd0f6 (). Each NoBalance query's doc comment in the .sql sources now begins with the exact method name so the generated Go doc comment satisfies the comment gate.

wallet/internal/sql/sqlite/queries/accounts.sql
```diff
@@ -76,11 +76,44 @@ SELECT
...
+    cast(coalesce(bal.unconfirmed_balance, 0) AS INTEGER) AS unconfirmed_balance
 FROM accounts AS a
 INNER JOIN key_scopes AS ks ON a.scope_id = ks.id
 INNER JOIN wallets AS w ON a.wallet_id = w.id
 LEFT JOIN account_secrets AS acs ON a.id = acs.account_id
+LEFT JOIN (
+    SELECT
+        addr.account_id,
+        addr.wallet_id,
+        cast(coalesce(SUM(
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in 288561524. Each balance subquery for the 5 queries that already pin `wallet_id` in their outer WHERE (`GetAccountByWalletScope*`, `ListAccountsByWallet*`) now adds `AND addr.wallet_id = sqlc.arg('wallet_id')` inside the subquery so the aggregation is scoped to one wallet's UTXOs rather than scanning every wallet. The two `GetAccountByScope*` selectors don't take a `wallet_id` parameter and are intentionally left unscoped — they're only reached by tests / introspection paths today.

wallet/internal/sql/pg/queries/accounts.sql
```diff
@@ -76,11 +76,44 @@ SELECT
...
+    coalesce(bal.unconfirmed_balance, 0)::BIGINT AS unconfirmed_balance
 FROM accounts AS a
 INNER JOIN key_scopes AS ks ON a.scope_id = ks.id
 INNER JOIN wallets AS w ON a.wallet_id = w.id
 LEFT JOIN account_secrets AS acs ON a.id = acs.account_id
+LEFT JOIN (
+    SELECT
+        addr.account_id,
+        addr.wallet_id,
+        coalesce(SUM(
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in 288561524. Each balance subquery for the 5 queries that already pin `wallet_id` in their outer WHERE (`GetAccountByWalletScope*`, `ListAccountsByWallet*`) now adds `AND addr.wallet_id = sqlc.arg('wallet_id')` inside the subquery so the aggregation is scoped to one wallet's UTXOs rather than scanning every wallet. The two `GetAccountByScope*` selectors don't take a `wallet_id` parameter and are intentionally left unscoped — they're only reached by tests / introspection paths today.

wallet/internal/sql/pg/queries/accounts.sql
yyforyongyu: Fixed in 288561524. Each balance subquery for the 5 queries that already pin `wallet_id` in their outer WHERE (`GetAccountByWalletScope*`, `ListAccountsByWallet*`) now adds `AND addr.wallet_id = sqlc.arg('wallet_id')` inside the subquery so the aggregation is scoped to one wallet's UTXOs rather than scanning every wallet. The two `GetAccountByScope*` selectors don't take a `wallet_id` parameter and are intentionally left unscoped — they're only reached by tests / introspection paths today.

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,256 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		Scope:       &scope,
+		SkipBalance: true,
+	})
+	require.NoError(t, err)
+
+	for _, acc := range accounts {
+		require.Equal(t, btcutil.Amount(0), acc.ConfirmedBalance,
+			acc.AccountName)
+		require.Equal(t, btcutil.Amount(0), acc.UnconfirmedBalance,
+			acc.AccountName)
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in 9882cd0f6. The expanded `assertZeroBalances` helper now requires the result is non-empty AND asserts the `funded` account is present in the returned slice before checking zero balances, so a regression that drops accounts from the list result would fail the test.

wallet/internal/db/accounts_common.go
```diff
@@ -472,16 +475,15 @@ func getAddrSchemaForScope(scope KeyScope) (ScopeAddrSchema, error) {
...
 }
 
-// BuildAccountInfo creates an AccountInfo with the provided values and zeroed
-// balances while we do not yet support balance tracking.
-//
-// TODO(stingelin): Add balance tracking support after transaction management is
-// implemented.
+// BuildAccountInfo creates an AccountInfo with the provided values.
+// confirmedBalance and unconfirmedBalance are populated verbatim; the
+// follow-up commit on this PR wires real values from a SQL aggregation.
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in cb838dba9. The `BuildAccountInfo` doc comment no longer references "the follow-up commit on this PR"; it describes the current contract — caller supplies `confirmedBalance` / `unconfirmedBalance`, create paths pass zero, read paths pass real values from the SQL aggregation.

wallet/internal/sql/sqlite/queries/accounts.sql
```diff
@@ -381,3 +645,185 @@ UPDATE accounts
...
+    AND ks.coin_type = ?
+ORDER BY a.account_number IS NULL, a.account_number;
+
+-- name: ListAccountsByWalletAndNameNoBalance :many
+-- Identical to ListAccountsByWalletAndName but skips the per-account UTXO
+-- aggregation. Returns rows with `confirmed_balance` and
+-- `unconfirmed_balance` set to zero. Used by the
+-- SkipBalance opt-out path to avoid the UTXO scan on
+-- identity-only reads.
+-- number. Imported accounts (with NULL account_number) appear last.
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in 9882cd0f6. The stale "-- number. Imported accounts (with NULL account_number) appear last." fragment line was a leftover from copying the source query's comment block into each NoBalance variant; the fragment has been removed from all 5 NoBalance comments so the doc block reads coherently.

wallet/internal/sql/pg/queries/accounts.sql
```diff
@@ -381,3 +645,185 @@ UPDATE accounts
...
+    AND ks.coin_type = $3
+ORDER BY a.account_number NULLS LAST;
+
+-- name: ListAccountsByWalletAndNameNoBalance :many
+-- Identical to ListAccountsByWalletAndName but skips the per-account UTXO
+-- aggregation. Returns rows with `confirmed_balance` and
+-- `unconfirmed_balance` set to zero. Used by the
+-- SkipBalance opt-out path to avoid the UTXO scan on
+-- identity-only reads.
+-- number. Imported accounts (with NULL account_number) appear last.
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in 9882cd0f6. The stale "-- number. Imported accounts (with NULL account_number) appear last." fragment line was a leftover from copying the source query's comment block into each NoBalance variant; the fragment has been removed from all 5 NoBalance comments so the doc block reads coherently.

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,256 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		WalletID: walletID,
+		Tx:       unconfirmedTx,
+		Received: time.Unix(1710000500, 0),
+		Status:   db.TxStatusPending,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
+	require.NoError(t, err)
+
+	query := getAccountQueryByName(walletID, scope, "funded")
+	query.SkipBalance = true
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in 9882cd0f6. `TestGetAccountSkipBalanceZerosFields` now exercises both the by-name and by-number selectors against the same funded account, so both `GetAccountByWalletScopeAndNameNoBalance` and `GetAccountByWalletScopeAndNumberNoBalance` dispatch branches are covered.

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,256 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		Block:    &confirmedBlock,
+		Status:   db.TxStatusPublished,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
+	require.NoError(t, err)
+
+	accounts, err := store.ListAccounts(t.Context(), db.ListAccountsQuery{
+		WalletID:    walletID,
+		Scope:       &scope,
+		SkipBalance: true,
```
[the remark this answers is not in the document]
yyforyongyu: Fixed in 9882cd0f6. `TestListAccountsSkipBalanceZerosFields` now exercises all three list selectors (scope-filtered, name-filtered, and unfiltered "all"), so `ListAccountsByWalletScopeNoBalance`, `ListAccountsByWalletAndNameNoBalance`, and `ListAccountsByWalletNoBalance` dispatch branches are all covered.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed in 9882cd0f6. `TestListAccountsSkipBalanceZerosFields` now exercises all three list selectors (scope-filtered, name-filtered, and unfiltered "all"), so `ListAccountsByWalletScopeNoBalance`, `ListAccountsByWalletAndNameNoBalance`, and `ListAccountsByWalletNoBalance` dispatch branches are all covered.

yyforyongyu:

wallet/internal/db/accounts_common.go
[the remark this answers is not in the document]
yyforyongyu: This still appears in the latest head: the comment says the follow-up commit on this PR wires real values from SQL aggregation, but this branch now already contains the aggregation commits. The earlier fix seems to have been dropped during the rewrite. (`openai/gpt-5.5`)

wallet/internal/sql/pg/queries/accounts.sql
```diff
@@ -381,3 +381,74 @@ UPDATE accounts
...
 RETURNING (next_internal_index - 1)::BIGINT AS address_index;
+
+-- name: AccountBalance :one
+-- AccountBalance returns the confirmed/unconfirmed balance for one
+-- account, summed from the wallet's UTXO set at read time. Confirmed
+-- means the funding tx is in a block at or below the wallet's synced
+-- height; unconfirmed covers unmined and above-synced-tip outputs.
+-- Spent outputs (`u.spent_by_tx_id IS NOT NULL`) are excluded.
+SELECT
+    coalesce(SUM(
```
yyforyongyu: The current CI failure is from `sql-format-check` rewriting these new aggregate expressions: lowercasing `SUM` and splitting/indenting each `WHEN`. Please commit the sqlfluff output and refresh the generated sqlc files so the checked-in SQL and generated query constants stay aligned. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 💰
(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/queries/accounts.sql
yyforyongyu: Fixed in the refreshed `db: add AccountBalance query` + `db: add AccountBalances query` commits. Ran `make sql-format` which lowercased `SUM` to `sum` and re-indented every `WHEN` clause to sqlfluff's expected form, then regenerated the sqlc files so `accounts.sql.go` matches the new query text. Also fixed an `ST09` join-condition-order finding on `INNER JOIN transactions AS t ON u.tx_id = t.id` in both queries. `make sql-format-check` + `make sql-lint-check` clean locally.

wallet/internal/db/accounts_common.go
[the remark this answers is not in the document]
yyforyongyu: Fixed again in eecf18c02 (the `db: thread balance fields through AccountInfoRow plumbing` commit at the bottom of this PR). The doc comment now reads "create paths pass zero (a fresh account has no UTXOs) while the read paths feed real values from the dedicated AccountBalance / AccountBalances queries" — no reference to "the follow-up commit on this PR".

Apologies for the regression — the rewrite reset the file from a pre-prep state and the earlier doc fix was lost in the process.

yyforyongyu (commented):
LGTM 🚀 (`openai/gpt-5.5`)

wallet/internal/db/pg/accounts.go
```diff
@@ -406,76 +407,234 @@ type accountListQueries struct {
...
+		info, err := accountRowToInfo(rows[i])
+		if err != nil {
+			return nil, err
+		}
+
+		infos[i] = *info
+		ids[i] = rows[i].ID
+	}
+
+	return p.attachBalances(ctx, query, infos, ids)
```
GustavoStingelin: Here, we already know the exact account we want to look up, but the balances query sums across the entire wallet.

This may not be a serious performance issue, since this path probably does not run very often, but it still feels a little strange. Would it make sense to scope the query to the account directly?

wallet/internal/db/pg/accounts.go
```diff
@@ -406,76 +407,234 @@ type accountListQueries struct {
...
```
GustavoStingelin: like the same here, the accounts are filtered for a specific key scope, but the balances run for the entire wallet

wallet/internal/db/pg/accounts.go
GustavoStingelin: maybe the `AccountBalance` query could receive a slice of account IDs for this case

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+	scope := db.KeyScopeBIP0084
+
+	queries := store.Queries()
+	syncBlock := CreateBlockFixture(t, queries, 200)
+	confirmedBlock := CreateBlockFixture(t, queries, 100)
+
+	err := store.UpdateWallet(t.Context(), db.UpdateWalletParams{
+		WalletID: walletID,
+		SyncedTo: &syncBlock,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		[]*wire.TxOut{{Value: 24000, PkScript: addr.ScriptPubKey}},
+	)
+	err = store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       confirmedTx,
+		Received: time.Unix(1710000000, 0),
+		Block:    &confirmedBlock,
+		Status:   db.TxStatusPublished,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 26000, PkScript: addr.ScriptPubKey}},
+	)
+	err = store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       unconfirmedTx,
+		Received: time.Unix(1710000100, 0),
+		Status:   db.TxStatusPending,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+	scope := db.KeyScopeBIP0084
+
+	queries := store.Queries()
+	syncBlock := CreateBlockFixture(t, queries, 200)
+	confirmedBlock := CreateBlockFixture(t, queries, 100)
+
+	err := store.UpdateWallet(t.Context(), db.UpdateWalletParams{
+		WalletID: walletID,
+		SyncedTo: &syncBlock,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		[]*wire.TxOut{{Value: 24000, PkScript: addr.ScriptPubKey}},
+	)
+	err = store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       tx,
+		Received: time.Unix(1710000200, 0),
+		Block:    &confirmedBlock,
+		Status:   db.TxStatusPublished,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+	scope := db.KeyScopeBIP0084
+
+	queries := store.Queries()
+	syncBlock := CreateBlockFixture(t, queries, 200)
+	confirmedBlock := CreateBlockFixture(t, queries, 100)
+
+	err := store.UpdateWallet(t.Context(), db.UpdateWalletParams{
+		WalletID: walletID,
+		SyncedTo: &syncBlock,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		[]*wire.TxOut{{Value: 12000, PkScript: firstAddr.ScriptPubKey}},
+	)
+	err = store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       firstTx,
+		Received: time.Unix(1710000200, 0),
+		Block:    &confirmedBlock,
+		Status:   db.TxStatusPublished,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 7000, PkScript: secondAddr.ScriptPubKey}},
+	)
+	err = store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       secondTx,
+		Received: time.Unix(1710000300, 0),
+		Status:   db.TxStatusPending,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		require.Equal(t, btcutil.Amount(0),
+			byName["empty"].ConfirmedBalance, label+": empty")
+		require.Equal(t, btcutil.Amount(0),
+			byName["empty"].UnconfirmedBalance, label+": empty")
+	}
+
+	byScope, err := store.ListAccounts(t.Context(), db.ListAccountsQuery{
+		WalletID: walletID,
+		Scope:    &scope,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+	require.NoError(t, err)
+	verify(t, byScope, "by scope")
+
+	all, err := store.ListAccounts(t.Context(), db.ListAccountsQuery{
+		WalletID: walletID,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+	scope := db.KeyScopeBIP0084
+
+	queries := store.Queries()
+	syncBlock := CreateBlockFixture(t, queries, 200)
+	confirmedBlock := CreateBlockFixture(t, queries, 100)
+
+	err := store.UpdateWallet(t.Context(), db.UpdateWalletParams{
+		WalletID: walletID,
+		SyncedTo: &syncBlock,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+		[]*wire.TxOut{{Value: 9000, PkScript: addr.ScriptPubKey}},
+	)
+	err = store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       tx,
+		Received: time.Unix(1710000400, 0),
+		Block:    &confirmedBlock,
+		Status:   db.TxStatusPublished,
+		Credits:  map[uint32]btcutil.Address{0: nil},
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+
+		require.True(t, found,
+			label+": funded account missing from result")
+	}
+
+	byScope, err := store.ListAccounts(t.Context(), db.ListAccountsQuery{
+		WalletID:    walletID,
+		Scope:       &scope,
+		SkipBalance: true,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+	require.NoError(t, err)
+	verify(t, byScope, "by scope")
+
+	name := "funded"
+	byName, err := store.ListAccounts(t.Context(), db.ListAccountsQuery{
+		WalletID:    walletID,
+		Name:        &name,
+		SkipBalance: true,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/itest/account_store_test.go
```diff
@@ -992,6 +994,329 @@ func TestListAccountsReturnsPublicKey(t *testing.T) {
...
+	require.NoError(t, err)
+	verify(t, byName, "by name")
+
+	all, err := store.ListAccounts(t.Context(), db.ListAccountsQuery{
+		WalletID:    walletID,
+		SkipBalance: true,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/pg/accounts.go
```diff
@@ -406,76 +407,234 @@ type accountListQueries struct {
...
+	accountID int64) (*db.AccountInfo, error) {
+
+	if query.SkipBalance {
+		return info, nil
+	}
+
+	bal, err := p.q.AccountBalance(ctx, sqlc.AccountBalanceParams{
+		WalletID:  int64(query.WalletID),
+		AccountID: accountID,
+	})
```
GustavoStingelin: nit: missing new line

wallet/internal/db/sqlite/accounts.go
```diff
@@ -416,75 +417,234 @@ type accountListQueries struct {
...
+	accountID int64) (*db.AccountInfo, error) {
+
+	if query.SkipBalance {
+		return info, nil
+	}
+
+	bal, err := s.q.AccountBalance(ctx, sqlc.AccountBalanceParams{
+		WalletID:  int64(query.WalletID),
+		AccountID: accountID,
+	})
```
GustavoStingelin: nit: missing new line

GustavoStingelin (changes requested):
Just raised a comment about the segmentation of the balance query and some nits.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/itest/account_store_test.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/pg/accounts.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/sqlite/accounts.go
yyforyongyu: Fixed — call site reformatted to match the `fmt.Errorf`-style wrapping conventions (opening paren on its own line so the argument list aligns one tab in). Folded into the originating commits, no tail fix-up.

wallet/internal/db/pg/accounts.go
yyforyongyu: Good catch — addressed. The wallet-wide `AccountBalances(walletID)` is replaced by `AccountBalancesByIDs(walletID, ids)` which scopes the UTXO aggregation to just the accounts the caller's `ListAccounts` query actually returned (the `ids` slice already collected from the row fetch). For `byScope` and `byName` this means the balance query now sums UTXOs only for the in-scope accounts; for `all` the `ids` is the full wallet list so the result is unchanged.

pg uses `account_id = ANY(@account_ids::BIGINT[])`; sqlite uses `account_id IN (sqlc.slice('account_ids'))`. The singular `AccountBalance` used by `GetAccount` was already account-scoped and is untouched.

Folded into the originating commits `db: add AccountBalancesByIDs query` (`7c414fdef`) and `db: dispatch AccountBalancesByIDs on ListAccounts reads` (`ffb6da50a`).

wallet/internal/db/pg/accounts.go
yyforyongyu: Good catch — addressed. The wallet-wide `AccountBalances(walletID)` is replaced by `AccountBalancesByIDs(walletID, ids)` which scopes the UTXO aggregation to just the accounts the caller's `ListAccounts` query actually returned (the `ids` slice already collected from the row fetch). For `byScope` and `byName` this means the balance query now sums UTXOs only for the in-scope accounts; for `all` the `ids` is the full wallet list so the result is unchanged.

pg uses `account_id = ANY(@account_ids::BIGINT[])`; sqlite uses `account_id IN (sqlc.slice('account_ids'))`. The singular `AccountBalance` used by `GetAccount` was already account-scoped and is untouched.

Folded into the originating commits `db: add AccountBalancesByIDs query` (`7c414fdef`) and `db: dispatch AccountBalancesByIDs on ListAccounts reads` (`ffb6da50a`).

yyforyongyu:

yyforyongyu (commented):
LGTM 🚀 (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🎯
(`claude-opus-4-7[1m]`)

wallet/internal/db/sqlite/accounts.go
```diff
@@ -417,41 +441,120 @@ type accountListQueries struct {
...
+		ctx, sqlc.AccountBalancesByIDsParams{
+			WalletID:   int64(query.WalletID),
+			AccountIds: ids,
+		},
 	)
+	if err != nil {
+		return nil, fmt.Errorf("account balances: %w", err)
+	}
+
+	byID := make(map[int64]sqlc.AccountBalancesByIDsRow, len(balances))
```
yyforyongyu: why do we need to have this map here? like iterating twice - first the `balances`, then the `ids`. if we make `processAccountRows` returns a map of `accountRowID: info`, we can just use that map in `attachBalances` without taking two params `infos` and `ids`?

wallet/internal/db/accounts_common.go
```diff
@@ -367,6 +377,79 @@ func getKeyCounts(external, internal, imported int64) (uint32, uint32,
...
+
+		return fingerprint, nil
+	}
+
+	if origin != DerivedAccount {
+		return 0, nil
+	}
+
+	if len(walletMasterHDPubKey) == 0 {
+		return 0, fmt.Errorf(
```
yyforyongyu: fmt.Errorf should be wrapped? you are making the same mistakes again

wallet/internal/db/accounts_common.go
```diff
@@ -367,6 +377,79 @@ func getKeyCounts(external, internal, imported int64) (uint32, uint32,
...
+// the shape that waddrmgr persists for kvdb wallets and that SQL stores
+// plaintext in wallets.master_hd_pub_key per ADR 0009.
+//
+// Wallet-side code (wallet/account_derive.go:masterKeyFingerprint)
+// computes the same value from an already-parsed *hdkeychain.ExtendedKey;
+// the wallet package imports this package, so it cannot reuse this helper
+// from there without a circular dependency.
+//
+// Returns an error if the input is not a base58-encoded extended key.
+func MasterKeyFingerprintFromExtKeyBytes(masterHDPubKey []byte) (uint32,
```
yyforyongyu: why do we need to export this method?

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -54,8 +54,27 @@ CREATE TABLE accounts (
     -- Reference to the origin of the account.
     origin_id SMALLINT NOT NULL,
 
-    -- Master fingerprint is the fingerprint of the master pub key that created
-    -- this account.
+    -- BIP32 master-key fingerprint (HASH160(serialized master pubkey)[0:4],
```
yyforyongyu: This commit should be split into two:
- first commit: we expose the wallet `master_hd_pub_key` in the sql files, including updating comments, adding the fields, etc.
- second, we introduce the `resolveMasterFingerprint` and use it

yyforyongyu:
LGTM 🧬
(`claude-opus-4-7[1m]`)

wallet/internal/db/accounts_common.go
```diff
@@ -364,6 +377,79 @@ func getKeyCounts(external, internal, imported int64) (uint32, uint32,
...
+	}
+
+	if len(walletMasterHDPubKey) == 0 {
+		return 0, fmt.Errorf(
+			"%w: account %q", errMissingWalletMasterHDPubKey,
+			accountName,
+		)
+	}
+
+	fingerprint, err := MasterKeyFingerprintFromExtKeyBytes(
```
yyforyongyu: This makes derived-account reads depend on `wallets.master_hd_pub_key`, but wallet creation still accepts a nil or invalid `MasterPubKey` and derived account creation can still succeed because the row stores NULL here. The result is an account that is returned successfully from create using the callback fingerprint, then cannot be fetched or listed later because fingerprint resolution errors. Please reject derived account creation when the wallet lacks a valid master HD pubkey, or keep a durable per-account fingerprint for derived rows. (`openai/gpt-5.5`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -84,6 +103,13 @@ CREATE TABLE accounts (
     -- Imported address counter must be non-negative.
     CHECK (imported_key_count >= 0),
 
+    -- Enforce the master_fingerprint SSOT invariant documented on the
+    -- column above.
+    CHECK (
```
yyforyongyu: This new invariant means the raw imported-account mismatch fixtures that omit `master_fingerprint` now fail before they exercise the wallet/scope FK. Those tests only assert a generic constraint error, so they can pass even if the composite FK regresses. Please update the raw imported-account fixtures to include a non-NULL fingerprint so the tests still cover the intended constraint. (`openai/gpt-5.5`)

wallet/internal/db/sqlite/accounts.go
yyforyongyu: Addressed in `0b8ad9707`. `processAccountRows` now returns `[]rowAccountInfo` (a slice of `{id, *AccountInfo}` pairs) preserving row order, and `attachBalances` builds a single `indexByID map[int64]int` plus the `ids` slice in one pass over `entries`, then walks `balances` once to merge — eliminating the intermediate `sqlc.AccountBalancesByIDsRow` map. Net result is 2 iterations over `entries` + 1 over `balances`, down from 1+1+1+1.

wallet/internal/db/accounts_common.go
yyforyongyu: Made moot in `0b8ad9707`. The fingerprint commit was dropped entirely from this PR. Derived-account fingerprint resolution moves out of the SQL read path (no JOIN with `wallets`, no helper, no sentinel). The wallet caches its master fingerprint at `Open()` time from the already-loaded `WalletInfo.MasterPubKey`, and `accountInfoToProperties` injects it for derived `db.AccountInfo` values at the wallet/waddrmgr conversion boundary. That work lands on the downstream `impl-account-manager-store` branch, not here.

wallet/internal/db/accounts_common.go
yyforyongyu: Made moot in `0b8ad9707`. `MasterKeyFingerprintFromExtKeyBytes` is gone entirely — the helper was only needed by the SQL read path that derived fingerprints from `wallets.master_hd_pub_key`, and that read path has been removed in favor of caching the parsed fingerprint on the `Wallet` instance at `Open()` time. Nothing exported from `db` for kvdb to call.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: Made moot in `0b8ad9707`. The fingerprint commit has been dropped entirely from this PR, so the SQL-exposure split is no longer needed — there is no `master_hd_pub_key` projection on account read queries, no `resolveMasterFingerprint` helper, and no CHECK constraint. The architectural decision was that fingerprint resolution for derived accounts belongs on the wallet side (cached once at `Open()` time), not in the SQL row adapter.

wallet/internal/db/accounts_common.go
yyforyongyu: Bug fixed at the right layer in `0b8ad9707` by dropping the SQL read-time fingerprint derivation entirely. The wallet now parses and caches `MasterPubKey` once at `Open()`/`Create()` time and stores `masterFingerprint uint32` on the `Wallet` instance. Invalid `MasterPubKey` fails at wallet open — the inconsistency you flagged (account creation succeeds, later reads fail) can no longer occur because (a) no read path parses the wallet xpub, and (b) wallets with unparseable xpubs cannot open in the first place. This work lands on the downstream `impl-account-manager-store` branch where the wallet-side cache is added.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: Made moot in `0b8ad9707`. The CHECK constraint (`origin_id = 1 AND master_fingerprint IS NOT NULL`) has been removed because the fingerprint commit is gone. Imported accounts continue to store `master_fingerprint` per-row, but the schema no longer enforces non-NULL on it, so the raw imported-account mismatch fixtures retain their original behavior — they exercise the composite wallet/scope FK as intended.

yyforyongyu (commented):
LGTM 🚀 (`openai/gpt-5.5`)

wallet/internal/db/pg/accounts.go
```diff
@@ -349,6 +349,38 @@ type accountInfoRow interface {
 		sqlc.ListAccountsByWalletAndNameRow
 }
 
+// rowAccountInfo pairs an account's row ID with its materialized
+// AccountInfo. processAccountRows produces a sequence of these so the
+// caller can attach per-account balances by ID without threading a
+// parallel ids slice.
+type rowAccountInfo struct {
```
yyforyongyu: If this is commonly needed, why don't we add a private field in `db.AccountInfo`, like `rowID`?

then the methods like `processAccountRows` can just be a helper method in the accounts_common.go?

wallet/internal/db/pg/accounts.go
```diff
@@ -411,41 +443,126 @@ type accountListQueries struct {
...
+	return p.attachBalances(ctx, query, entries)
+}
+
+// attachBalances issues the AccountBalancesByIDs batch query for the row
+// IDs in entries and merges each returned (account_id, confirmed,
+// unconfirmed) tuple into the matching AccountInfo. The dispatch is
+// skipped when query.SkipBalance is true or there are no entries to
+// enrich, in which case every returned AccountInfo keeps zero balance
+// fields. The returned slice preserves the input order of entries.
+func (p accountListQueries) attachBalances(ctx context.Context,
```
yyforyongyu, line 508: can this method also be put in the `accounts_common.go`?

wallet/internal/db/pg/accounts.go
```diff
@@ -411,41 +443,126 @@ type accountListQueries struct {
...
+	query db.ListAccountsQuery,
+	entries []rowAccountInfo) ([]db.AccountInfo, error) {
```
yyforyongyu: i still don't see why `processAccountRows` cannot just return a map, and here we can take a map instead of entries? such that the `id:info` relation is preserved?

wallet/internal/db/pg/accounts.go
yyforyongyu: Good point. Proposed design — please confirm before I refactor:

1. **`db.AccountInfo` gains an unexported `rowID int64` field.** SQL backends populate it at row-conversion time; kvdb leaves it as the zero value (kvdb has no SQL-style row ID, and the SSOT design doesn't need one for kvdb because kvdb has no per-account balance batch query).

2. **`processAccountRows` moves to `accounts_common.go`** as a generic helper. Single shared implementation across pg and sqlite. Returns `[]*db.AccountInfo` (ordered as the SQL rows arrived — the queries use `ORDER BY a.account_number NULLS LAST`, so order is meaningful and worth preserving). Each `AccountInfo` has its `rowID` set.

3. **`rowAccountInfo` struct gets deleted** entirely. Removes the wrapper noise.

Let me know if you'd prefer `processAccountRows` to return a `map[int64]*AccountInfo` instead — see my reply to your third comment on attachBalances for the trade-off.

wallet/internal/db/pg/accounts.go
yyforyongyu: Yes. With the `rowID` field on `AccountInfo` and `processAccountRows` moved per the previous comment, `attachBalances` can also live in `accounts_common.go` as a generic helper. The backend-specific piece is the call to `AccountBalancesByIDs`, which gets injected as a callback:

```go
func attachBalances(
    ctx context.Context,
    query db.ListAccountsQuery,
    infos []*db.AccountInfo,
    queryBalances func(ctx context.Context, walletID uint32, ids []int64) (
        []AccountBalance, error,
    ),
) ([]db.AccountInfo, error) { ... }
```

`AccountBalance` is a small dialect-agnostic struct (`{accountID, confirmed, unconfirmed int64}`) that pg and sqlite both convert their sqlc rows into at the callback boundary. Net result: zero duplication between pg/sqlite for the balance attachment path.

wallet/internal/db/pg/accounts.go
yyforyongyu: Fair, the `id:info` relation is preserved more directly with a map. The hesitation was order — the three `ListAccountsByWallet*` SQL queries all end with `ORDER BY a.account_number NULLS LAST`, and that order is meaningful for callers (default derived account first, imported accounts last). Map iteration in Go is randomized, so returning a map drops that order.

Two options to reconcile:

**Option A — return `[]*db.AccountInfo` (ordered):** `processAccountRows` returns an ordered slice with each `info.rowID` set. Inside `attachBalances` we build a `map[int64]*db.AccountInfo` from the slice purely for O(1) merge lookup, mutate via pointer, then return `[]db.AccountInfo` in the same order. The id↔info relation is preserved via the field; the map is an internal implementation detail of the merge step.

**Option B — return `map[int64]*db.AccountInfo`:** `attachBalances` takes the map, merges balances, and sorts the output by `(Origin asc, AccountNumber asc)` before returning. This matches the SQL `ORDER BY ... NULLS LAST` semantically since imported = origin 1 sorts after derived = origin 0. Costs an O(n log n) sort that the SQL already did for us.

I lean A because it doesn't re-do work the SQL already did. Happy to go with B if you prefer the cleaner map signature.

wallet/internal/db/pg/accounts.go
yyforyongyu: Done in `203602451`. `db.AccountInfo` now has an unexported `rowID int64` field (populated by `ProcessAccountRows`, ignored by kvdb which leaves it zero). `ProcessAccountRows` lives in `wallet/internal/db/accounts_common.go` as a generic helper that takes a `convert` closure (so each dialect still owns its row→AccountInfo mapping). The `rowAccountInfo` wrapper type is deleted.

wallet/internal/db/pg/accounts.go
yyforyongyu: Done in `203602451`. `db.AttachBalances` lives in `accounts_common.go` and takes a `queryBalances` callback for the backend-specific `AccountBalancesByIDs` call plus a dialect-agnostic `db.AccountBalance` struct (`{AccountID, Confirmed, Unconfirmed}`) that pg/sqlite translate their sqlc rows into at the callback boundary. The `(p accountListQueries) attachBalances` and `(s accountListQueries) attachBalances` methods are now thin ~25-line wrappers that just call `db.AttachBalances` with their backend-specific closure.

wallet/internal/db/pg/accounts.go
yyforyongyu: Went with Option A in `203602451` per the evaluation we walked through. `ProcessAccountRows` returns `[]*db.AccountInfo` (ordered as input), each with its `rowID` set. Inside `AttachBalances`, an `indexByID map[int64]int` is built once for O(1) merge against the balance rows. The id↔info relation is explicit on `AccountInfo.rowID` (clean, no wrapper); the map exists only inside the merge step. SQL `ORDER BY a.account_number NULLS LAST` flows through unchanged.

wallet/internal/db/accounts_common.go
```diff
@@ -816,3 +816,94 @@ func CreateImportedAccount[CreateArgs any, CreateRow any, SecretArgs any](
...
+// values. The queryBalances callback executes the backend-specific
+// AccountBalancesByIDs and returns its results as []AccountBalance. The
+// merge step uses each AccountInfo's rowID (populated by ProcessAccountRows)
+// to map balance rows back to their AccountInfo in a single pass. The
+// returned slice preserves the input order of infos.
+//
+// When skipBalance is true or infos is empty, the query dispatch is skipped
+// and every returned AccountInfo keeps zero balance fields. The query
+// callback is invoked at most once per AttachBalances call.
+func AttachBalances(ctx context.Context, walletID uint32, skipBalance bool,
```
yyforyongyu, line 872: hmmm there has been multiple iterations, I am not sure if this is the best result in terms of de-duplication - I will leave it as is for now, given that the best way to handle this area is through the `ops interface`, which will come in a future PR cc @GustavoStingelin

GustavoStingelin (approved):
ok

PR #1235, opened 2026-05-17
[feature]: add WithChangeAddress TxCreateOption for custom change output destination

murraystewart96:
## Change Description
Description:

This change extends txToOutputs and addrMgrWithChangeSource to accept an optional changeAddr btcutil.Address parameter. When non-nil, the change source is built as a static script paying to that address rather than deriving a fresh HD wallet address. When nil, the existing derivation path is taken unchanged — there is no behaviour change for existing callers.

## Steps to Test

added new test in createtx_test.go - TestTxToOutputsCustomChangeAddr

### Testing

### DIFF
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -320,8 +320,11 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 
 		// Finally, we'll request the backend to notify us of the
 		// transaction that pays to the change address, if there is one,
-		// when it confirms.
-		if tx.ChangeIndex >= 0 {
+		// when it confirms. We skip this for external change addresses
+		// since they are not wallet-owned — registering them would cause
+		// future payments to that address to be treated as wallet-relevant
+		// transactions.
+		if tx.ChangeIndex >= 0 && changeAddr == nil {
 			changePkScript := tx.Tx.TxOut[tx.ChangeIndex].PkScript
 			_, addrs, _, err := txscript.ExtractPkScriptAddrs(
 				changePkScript, w.chainParams,
@@ -441,6 +444,11 @@ func (w *Wallet) addrMgrWithChangeSource(dbtx walletdb.ReadWriteTx,
 	addrmgrNs := dbtx.ReadWriteBucket(waddrmgrNamespaceKey)
 
 	if changeAddr != nil {
+		if !changeAddr.IsForNet(w.chainParams) {
+			return nil, nil, fmt.Errorf("change address %v is not "+
+				"for network %v", changeAddr, w.chainParams.Name)
+		}
+
 		pkScript, err := txscript.PayToAddrScript(changeAddr)
 		if err != nil {
 			return nil, nil, err
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -440,9 +440,6 @@ func (w *Wallet) addrMgrWithChangeSource(dbtx walletdb.ReadWriteTx,
 
 	addrmgrNs := dbtx.ReadWriteBucket(waddrmgrNamespaceKey)
 
-	// If a custom change address is provided, build a static change source
-	// from it. The address manager bucket is still returned because callers
-	// need it for IsWatchOnlyAccount checks and signing.
 	if changeAddr != nil {
 		pkScript, err := txscript.PayToAddrScript(changeAddr)
 		if err != nil {
--- a/wallet/createtx.go
+++ b/wallet/createtx.go
@@ -142,7 +142,8 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 	account uint32, minconf int32, feeSatPerKb btcutil.Amount,
 	strategy CoinSelectionStrategy, dryRun bool,
 	selectedUtxos []wire.OutPoint,
-	allowUtxo func(utxo wtxmgr.Credit) bool) (
+	allowUtxo func(utxo wtxmgr.Credit) bool,
+	changeAddr btcutil.Address) (
 	*txauthor.AuthoredTx, error) {
 
 	chainClient, err := w.requireChainClient()
@@ -176,7 +177,7 @@ func (w *Wallet) txToOutputs(outputs []*wire.TxOut,
 	var tx *txauthor.AuthoredTx
 	err = walletdb.Update(w.db, func(dbtx walletdb.ReadWriteTx) error {
 		addrmgrNs, changeSource, err := w.addrMgrWithChangeSource(
-			dbtx, changeKeyScope, account,
+			dbtx, changeKeyScope, account, changeAddr,
 		)
 		if err != nil {
 			return err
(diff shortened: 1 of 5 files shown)


yyforyongyu (commented):
LGTM 🏷️ (`deepseek-v4-pro`)

wallet/createtx.go
```diff
@@ -429,14 +430,30 @@ func inputYieldsPositively(credit *wire.TxOut,
...
+	changeAddr btcutil.Address) (
 	walletdb.ReadWriteBucket, *txauthor.ChangeSource, error) {
 
+	addrmgrNs := dbtx.ReadWriteBucket(waddrmgrNamespaceKey)
+
+	// If a custom change address is provided, build a static change source
+	// from it. The address manager bucket is still returned because callers
+	// need it for IsWatchOnlyAccount checks and signing.
+	if changeAddr != nil {
+		pkScript, err := txscript.PayToAddrScript(changeAddr)
```
yyforyongyu, line 452: The custom change address is never validated against the wallet's network before `PayToAddrScript`. A caller that passes a change address for the wrong network (a typo, or a mainnet address against a testnet wallet) gets a perfectly-formed change `scriptPubKey` for the wrong chain — and since change is usually the bulk of the input value, those funds land at an address the wallet can never spend.

This isn't hypothetical: the new `TestTxToOutputsCustomChangeAddr` funds a **testnet** wallet (`testWallet` uses `chaincfg.TestNet3Params`) yet supplies a **mainnet** change address (`bc1qw508d6...` decoded with `MainNetParams`), and the test passes — precisely because there's no network check here.

The wallet already guards against this elsewhere: `waddrmgr.ScopedKeyManager.ImportPrivateKey` rejects `!wif.IsForNet(chainParams)`. Suggest the same here before building the script:

```go
if changeAddr != nil {
    if !changeAddr.IsForNet(w.chainParams) {
        return nil, nil, fmt.Errorf("change address %v is not for "+
            "network %v", changeAddr, w.chainParams.Name)
    }
    pkScript, err := txscript.PayToAddrScript(changeAddr)
    ...
```

(`w.chainParams` is already on the struct.) Note the new test will need a testnet change address once the check is added.

(`claude-opus-4-8[1m]`)

wallet/createtx.go
```diff
@@ -429,14 +430,27 @@ func inputYieldsPositively(credit *wire.TxOut,
...
+// specified key scope and account (defaulting to P2TR if no scope is given).
 func (w *Wallet) addrMgrWithChangeSource(dbtx walletdb.ReadWriteTx,
-	changeKeyScope *waddrmgr.KeyScope, account uint32) (
+	changeKeyScope *waddrmgr.KeyScope, account uint32,
+	changeAddr btcutil.Address) (
 	walletdb.ReadWriteBucket, *txauthor.ChangeSource, error) {
 
+	addrmgrNs := dbtx.ReadWriteBucket(waddrmgrNamespaceKey)
+
+	if changeAddr != nil {
```
yyforyongyu, line 446: This PR's final commit (`btcwallet: removing comment`) deletes the explanatory comment that the implementing commit (`116a713`) had added right here:

> // If a custom change address is provided, build a static change source
> // from it. The address manager bucket is still returned because callers
> // need it for IsWatchOnlyAccount checks and signing.

That was a useful *why* comment — in particular the second sentence explains the non-obvious bit (why `addrmgrNs` is still returned on a path that doesn't derive an address). Adding it in one commit and removing it in another is wasted churn, and the end state is less clear than the intermediate one. Suggest restoring at least the "why we still return addrmgrNs" line and folding it back into the implementing commit.

(`claude-opus-4-8[1m]`)

wallet/createtx.go
```diff
@@ -429,14 +430,27 @@ func inputYieldsPositively(credit *wire.TxOut,
...
+		pkScript, err := txscript.PayToAddrScript(changeAddr)
+		if err != nil {
+			return nil, nil, err
+		}
+		return addrmgrNs, &txauthor.ChangeSource{
```
yyforyongyu, line 456: When `changeAddr` is external, this path still flows into the later unconditional `NotifyReceived` call for the change output. That registers a non-wallet address with the chain backend; future payments to that address can be delivered as wallet notifications, and `addRelevantTx` assumes notified transactions are wallet-relevant. Only notify for wallet-owned custom change, or skip the change-address notification for external destinations. (`gpt-5.5`)

wallet/psbt.go
```diff
@@ -182,7 +182,7 @@ func (w *Wallet) FundPsbt(packet *psbt.Packet, keyScope *waddrmgr.KeyScope,
 		// a new change address into the database.
 		err = walletdb.Update(w.db, func(dbtx walletdb.ReadWriteTx) error {
 			_, changeSource, err := w.addrMgrWithChangeSource(
-				dbtx, opts.changeKeyScope, account,
+				dbtx, opts.changeKeyScope, account, nil,
```
yyforyongyu, line 196: This ignores `WithChangeAddress` for PSBTs that already provide inputs: `opts.changeAddr` is populated above, but `nil` is passed here, so callers silently get wallet-derived change instead of their requested destination. Pass the option through and handle external change output metadata, or reject the option in this path. (`gpt-5.5`)

wallet/wallet.go
```diff
@@ -1313,6 +1316,15 @@ func WithUtxoFilter(allowUtxo func(utxo wtxmgr.Credit) bool) TxCreateOption {
 	}
 }
 
+// WithChangeAddress can be used to specify a custom address for the change
+// output. If unspecified, a change address is derived from the wallet's HD
+// tree using the configured change key scope.
+func WithChangeAddress(addr btcutil.Address) TxCreateOption {
```
yyforyongyu, line 1322: `FundPsbt` also accepts `TxCreateOption`, but an external `WithChangeAddress` cannot succeed in the no-input path: `CreateSimpleTx` can add the custom change output, then `FundPsbt` calls `ScriptForOutput` and errors because that output is not wallet-owned. Either reject this option for PSBT funding or append external custom change without wallet BIP32 output metadata. (`gpt-5.5`)

yyforyongyu (commented):
Thanks for the PR. I've started the agents to perform an initial review, meanwhile, I would suggest putting this PR on hold, as we are in the process of a major refactor/upgrade, the milesteon can be tracked [here](https://github.com/btcsuite/btcwallet/milestone/3).

wallet/createtx.go
murraystewart96: i removed this because I felt the code was self documenting

wallet/wallet.go
murraystewart96: I rejected the option for FundPsbt

wallet/psbt.go
murraystewart96: I rejected the option for FundPsbt

murraystewart96:
@yyforyongyu Thanks for reviewing. I have addressed the issues highlighted by the agents.
I'll keep an eye on the progress of the refactor and address any conflicts that emerge from it.

PR #1236, opened 2026-05-18
kvdb: implement account-store reads + creators

yyforyongyu:
## Summary

Read-side half of the account-store migration. Lays down the kvdb
backend for the account *readers* and routes the wallet manager's
read paths (`GetAccount`, `ListAccounts*`) through `w.store`. All
write-side kvdb implementations + their routing land in the next
stacked PR (account-manager PR-B), so PR-A ends cleanly with every
new kvdb method already wired up from the wallet side.

## Commits (interleaved kvdb + routing)

1. `waddrmgr: expose legacy adapter methods on public interfaces`
   — promotes `EncryptedMasterHDPriv`, `Decrypt`,
   `ImportWitnessScript` so kvdb can read what waddrmgr persists.
2. `kvdb: implement GetEncryptedHDSeed` — first kvdb account-store
   reader.
3. `wallet: add cache and account-manager helpers` —
   `storeRuntimeCache`, shared `accountInfoToProperties`
   ground-work, `errNilAccountInfo`.
4. `wallet: cache master fingerprint on Wallet at Manager.Load`
   — `Wallet.masterFingerprint` field populated once at open by a
   `walletdb.View` against waddrmgr. Adds
   `resolveMasterFingerprint` helper + `errMissingWaddrmgrNamespace`
   sentinel. See task 201 for the design rationale.
5. `wallet: convert accountInfoToProperties to method and inject
   derived fingerprint` — `(w *Wallet).accountInfoToProperties` now
   injects `w.masterFingerprint` for `db.DerivedAccount` snapshots.
6. `kvdb: implement GetAccount` — kvdb reader matching the SQL
   contract (origin, watch-only, key counts, pubkey).
7. `wallet: route GetAccount through Store` — `Wallet.GetAccount`
   now goes through `w.cache.GetAccount`. `TestGetAccount` updated
   to mock the store. The four account-creation/balance/rename
   tests that internally call `w.GetAccount` are deleted in this
   commit; PR-B's matching routing commits add their store-based
   replacements.
8. `kvdb: implement ListAccounts` — kvdb reader.
9. `wallet: route ListAccounts variants through cache` —
   `Wallet.ListAccounts`, `Wallet.ListAccountsByName`, and
   `Wallet.ListAccountsByScope` now go through `w.cache`.

## Not in scope (PR-B)

- `kvdb: implement RenameAccount` (moved to PR-B so each kvdb
  method ships with its wallet-side routing in the same PR)
- `wallet: route NewAccount / RenameAccount / Balance / ImportAccount
  through Store`
- `kvdb: implement CreateDerivedAccount` (via ops adapter for
  derived-key callback contract)
- `kvdb: implement CreateImportedAccount`
- The dry-run drop on `Wallet.ImportAccount`

Those land on `impl-account-manager-store` stacked on top of this
PR.

## Testing

- `GOWORK=off go test ./wallet/... ./wallet/internal/db/...`
- per-commit build verification with `GOWORK=off go build ./...`
- CI: unit (cover + race), itest (bitcoind v28/v30, btcd, neutrino),
  postgres + sqlite itest (cover + race), lint


yyforyongyu:

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -2,42 +2,778 @@ package kvdb
...
+	scope waddrmgr.KeyScope, params db.CreateImportedAccountParams,
+	pubKey *hdkeychain.ExtendedKey,
+	walletIsWatchOnly bool) (*db.AccountInfo, error) {
+
+	scopedMgr, err := s.scopedManagerOrCreate(ns, scope)
+	if err != nil {
+		return nil, err
+	}
+
+	accountNumber, err := scopedMgr.AllocateImportedAccountNumber(ns)
```
yyforyongyu: This advances the same waddrmgr `lastAccount` counter used by derived accounts, but the new AccountStore contract keeps imported accounts out of the derived account-number sequence and returns `AccountNumber=0` for them (see the SQL `TestCreateDerivedAccountIgnoresImportedAccounts` coverage). With this adapter, importing an xpub makes the next derived account skip a number and returns a backend-specific imported account number, so callers see different account numbering depending on the backend. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -2,42 +2,778 @@ package kvdb
...
+	_ int64, accountNumber int64, name string,
+	derived *db.DerivedAccountData) (db.CreateDerivedAccountRow, error) {
+
+	if o.scopedMgr == nil {
+		return db.CreateDerivedAccountRow{}, errKvdbScopedMgrUninitialized
+	}
+
+	err := o.scopedMgr.PutDerivedAccountWithKeys(
+		o.ns, uint32(accountNumber), name, //nolint:gosec
+		derived.PublicKey, derived.EncryptedPrivateKey,
```
yyforyongyu: `derived.MasterKeyFingerprint` is not passed into waddrmgr here, so `CreateDerivedAccount` returns the fingerprint from the callback but the stored row cannot round-trip it: a later `GetAccount`/`ListAccounts` reads `props.MasterKeyFingerprint` as zero for default accounts. The SQL backends persist this field and the shared tests assert that derived account public key and fingerprint survive a re-read, so kvdb will diverge from the db.AccountStore contract. (`openai/gpt-5.5`)

wallet/mock_test.go
```diff
@@ -894,10 +894,46 @@ func (m *mockAddrStore) ChainParams() *chaincfg.Params {
 }
 
 // Close cleanly shuts down the manager.
+
+func (m *mockAddrStore) EncryptedMasterHDPriv(
```
yyforyongyu, line 915: The new `EncryptedMasterHDPriv` block is inserted between the `// Close cleanly shuts down the manager.` doc and the `Close()` function it documents (lines 896 → 898 → 909), so `Close()` ends up undocumented and `EncryptedMasterHDPriv` has no doc-comment of its own. Move `EncryptedMasterHDPriv` below `Close` (or add a real doc-comment above it and keep `Close`'s comment immediately above `Close`).

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -2,15 +2,45 @@ package kvdb
...
+// kvdb decoupled from waddrmgr.AddrStore's full breadth.
+type kvdbAddrStore interface {
+	// EncryptedMasterHDPriv reads the encrypted master HD private key
+	// from the manager's main bucket.
+	EncryptedMasterHDPriv(ns walletdb.ReadBucket) ([]byte, error)
+}
+
+// addrManager type-asserts s.addrStore into kvdbAddrStore. A nil store or a
+// wrong type indicates a caller-side wiring bug.
+func (s *Store) addrManager() (kvdbAddrStore, error) {
```
yyforyongyu: this method should be removed? we should use `s.addrStore` instead in the callsite

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -11,47 +10,9 @@ import (
...
-	"kvdb: addrStore does not satisfy kvdbAddrStore",
-)
-
 // A compile-time assertion to ensure Store implements the wallet store.
 var _ db.WalletStore = (*Store)(nil)
 
-// kvdbAddrStore is the narrow waddrmgr.Manager surface kvdb depends on for
-// pure-DB reads against the legacy bucket layout. A separate interface keeps
-// kvdb decoupled from waddrmgr.AddrStore's full breadth.
-type kvdbAddrStore interface {
```
yyforyongyu: changes like this should be folded into previous commits?

wallet/mock_test.go
yyforyongyu: The above comment is still not addressed!

waddrmgr/interface.go
```diff
@@ -351,4 +369,15 @@ type AccountStore interface {
 	// and index.
 	DeriveAddr(account uint32, branch uint32, index uint32) (
 		btcutil.Address, []byte, error)
+	// ImportWitnessScript imports a user-provided native witness
```
yyforyongyu, line 372: missing a newline above

wallet/mock_test.go
```diff
@@ -1226,6 +1262,18 @@ func (m *mockAccountStore) ImportScript(
 	return args.Get(0).(waddrmgr.ManagedScriptAddress), args.Error(1)
 }
 
+func (m *mockAccountStore) ImportWitnessScript(ns walletdb.ReadWriteBucket,
```
yyforyongyu, line 1287: missing godocs

wallet/account_manager.go
```diff
@@ -23,10 +23,61 @@ import (
...
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/wtxmgr"
 )
 
+// errNilAccountInfo is returned when accountInfoToProperties receives a nil
+// snapshot.
+var errNilAccountInfo = errors.New("nil account info")
+
+// toDBKeyScope adapts a waddrmgr.KeyScope to the db.KeyScope contract.
+func toDBKeyScope(scope waddrmgr.KeyScope) db.KeyScope {
```
yyforyongyu: again i don't see why we need this thin layer??

wallet/wallet.go
```diff
@@ -411,6 +411,14 @@ type Wallet struct {
 	// birthdayBlock is the block from which the wallet started scanning.
 	// It is loaded on startup and cached to avoid database lookups.
 	birthdayBlock waddrmgr.BlockStamp
+	// masterFingerprint is the cached BIP32 master-key fingerprint for
```
yyforyongyu, line 415: missing a newline above

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,11 +26,204 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
 
-// GetAccount is not yet implemented for kvdb.
-func (s *Store) GetAccount(ctx context.Context,
-	_ db.GetAccountQuery) (*db.AccountInfo, error) {
+// GetAccount returns the account identified by the given query within the
+// requested scope.
+//
+// path-mapping with origin classification; refactor blocked by task 105.
+//
+//nolint:cyclop // bridges waddrmgr's split account-by-number / account-by-name
```
yyforyongyu: we should not skip linters arbitrarily!

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,11 +26,204 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
 	return nil, notImplemented(ctx, "CreateImportedAccount")
 }
 
-// GetAccount is not yet implemented for kvdb.
-func (s *Store) GetAccount(ctx context.Context,
-	_ db.GetAccountQuery) (*db.AccountInfo, error) {
+// GetAccount returns the account identified by the given query within the
+// requested scope.
+//
+// path-mapping with origin classification; refactor blocked by task 105.
```
yyforyongyu: can you stop adding local agentic context into the godocs?? why do we need to mention the `task 105` here?

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,11 +26,204 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+	scope := scopedMgr.Scope()
+
+	var pubKey []byte
+	if props.AccountPubKey != nil {
+		pubKey = []byte(props.AccountPubKey.String())
+	}
+
+	// Resolve the account's CreatedAt from the kvdb-owned side
+	// bucket (waddrmgr's account row layout has no creation
+	// timestamp; see task 102). Legacy rows that pre-date the shim
```
yyforyongyu: Stop adding local agentic context into the godocs!

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,11 +26,204 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+		return nil, fmt.Errorf("read created-at: %w", err)
+	}
+
+	// Pass the per-row master_fingerprint from waddrmgr through
+	// for both origins. For imported accounts this is the parent
+	// xpub fingerprint persisted on the watch-only row. For
+	// derived accounts waddrmgr's default-account row stores 0;
+	// the wallet layer overrides the value from its cached master
+	// fingerprint via accountInfoToProperties before reaching
+	// external consumers (task 201).
```
yyforyongyu: Stop adding local agentic context into the godocs!

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,11 +26,204 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+// public db.* equivalents.
+func translateAccountErr(err error, notFound error) error {
+	if err == nil {
+		return nil
+	}
+
+	switch {
+	case waddrmgr.IsError(err, waddrmgr.ErrAccountNotFound),
+		waddrmgr.IsError(err, waddrmgr.ErrScopeNotFound):
+		return notFound
```
yyforyongyu, line 238: missing a newline above!

wallet/internal/db/kvdb/accountstore_timestamps.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/btcsuite/btcwallet/walletdb"
+)
+
+// kvdbAccountCreatedAtBucketKey is the sub-bucket under the waddrmgr
+// namespace that holds per-account creation timestamps. waddrmgr's
+// account row layout does not include a creation timestamp; the kvdb
+// adapter stores it here so the db.AccountInfo.CreatedAt contract is
+// satisfied without changing waddrmgr's bucket layout.
+//
+// See ADR (forthcoming) and task 102.
```
yyforyongyu: Stop adding local agentic context into the godocs!

wallet/account_manager.go
```diff
@@ -400,65 +327,45 @@ func (w *Wallet) ListAccountsByName(_ context.Context,
...
+func (w *Wallet) GetAccount(ctx context.Context, scope waddrmgr.KeyScope,
 	name string) (*AccountResult, error) {
 
 	err := w.state.validateStarted()
 	if err != nil {
 		return nil, err
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(scope)
+	info, err := w.cache.GetAccount(ctx, db.GetAccountQuery{
```
yyforyongyu, line 280: `Wallet.GetAccount` calls `w.cache.GetAccount` without `SkipBalance: true`, so the kvdb / SQL backend runs the per-account balance walk (kvdb's `fetchAccountBalances` on every UTXO, SQL's `AccountBalance` aggregate) and the returned `ConfirmedBalance` / `UnconfirmedBalance` are immediately discarded — only `info.AccountNumber` is consumed before `w.store.Balance` issues a second, authoritative balance query at line 355.

Setting `SkipBalance: true` on this read skips the wasted backend balance work; the opt-out exists exactly for this case (see `SkipBalance` doc-comment in `data_types.go`).

I raised the same shape on the predecessor PR #1215 thread last week; it surfaced again in this restructure so worth landing the opt-out alongside the routing.

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,18 +26,339 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+func loadAccountInfo(ns walletdb.ReadBucket,
+	scopedMgr waddrmgr.AccountStore, accountNumber uint32,
+	walletWatchOnly bool) (*db.AccountInfo, error) {
+
+	props, err := scopedMgr.AccountProperties(ns, accountNumber)
+	if err != nil {
+		return nil, translateAccountErr(err, db.ErrAccountNotFound)
+	}
+
+	accountIsWatchOnly, err := scopedMgr.IsWatchOnlyAccount(
```
yyforyongyu: `IsWatchOnlyAccount` is lock-state dependent, not an account-origin classifier. For a normal derived account in a locked wallet, `ScopedKeyManager.loadAccountInfo` leaves `acctKeyPriv` nil, this call returns true, and the code below reports the account as `ImportedAccount` and masks its account number to 0; the same call also bypasses the `AccountProperties` special case for the legacy imported-address account. Use persisted account type/row shape instead of watch-only state when setting `Origin`. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -400,65 +327,45 @@ func (w *Wallet) ListAccountsByName(_ context.Context,
...
-			}
-		}
+	minConfs := int32(0)
+	dbScope := toDBKeyScope(scope)
 
-		return nil
+	result, err := w.store.Balance(ctx, db.BalanceParams{
+		WalletID: w.id,
+		Scope:    &dbScope,
+		Account:  &info.AccountNumber,
```
yyforyongyu: `AccountInfo.AccountNumber` is explicitly not usable for imported accounts; the store contract sets it to 0 for `ImportedAccount`. Passing it as a balance filter makes an imported-account lookup query derived account 0 under the same scope instead of that imported account, so the returned `TotalBalance` is wrong once this path reaches a backend with `Balance` implemented. Use the account read attached balance or a backend identifier that can represent imported accounts. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -400,65 +327,45 @@ func (w *Wallet) ListAccountsByName(_ context.Context,
...
-		if balances, ok := scopedBalances[scope]; ok {
-			if balance, ok := balances[accNum]; ok {
-				account.TotalBalance = balance
-			}
-		}
+	minConfs := int32(0)
+	dbScope := toDBKeyScope(scope)
 
-		return nil
+	result, err := w.store.Balance(ctx, db.BalanceParams{
```
yyforyongyu: `Manager.Load` wires `w.store` to `kvdb.NewStore`, but kvdb `Balance` still returns `notImplemented`. With this routing, `Wallet.GetAccount` now fails for ordinary kvdb wallets after the account read succeeds. Keep using the balance already attached by `w.cache.GetAccount`, retain the legacy walletdb walk, or implement kvdb `Balance` before calling it here. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,18 +26,339 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+		internalNumbers []uint32
+	)
+
+	err := walletdb.View(s.db, func(tx walletdb.ReadTx) error {
+		ns := tx.ReadBucket(waddrmgr.NamespaceKey)
+		if ns == nil {
+			return nil
+		}
+
+		scopedMgrs, err := selectScopedManagers(mgr, query.Scope)
```
yyforyongyu, line 266: `ListAccounts` never applies the shared filter validation. The SQL stores reject a query with both `Scope` and `Name` via `ErrInvalidAccountQuery`, but this implementation first narrows by scope and then filters by name, so callers see backend-specific behavior for the same `db.AccountStore` contract. Add the same mutual-exclusion check before walking managers. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,23 +26,538 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
-// GetAccount is not yet implemented for kvdb.
-func (s *Store) GetAccount(ctx context.Context,
-	_ db.GetAccountQuery) (*db.AccountInfo, error) {
+// GetAccount returns the account identified by the given query within the
+// requested scope.
+//
+// path-mapping with origin classification; refactor blocked by task 105.
+//
+//nolint:cyclop // bridges waddrmgr's split account-by-number / account-by-name
+func (s *Store) GetAccount(_ context.Context,
```
yyforyongyu, line 32: Cross-PR observation (#1236 + #1239): the four kvdb readers shipped here — `GetEncryptedHDSeed` (`2c356226e`), `GetAccount` (`b3a6c305d`, +425 LoC), `ListAccounts` (`18be224ae`), and the `RenameAccount` writer that rides along (`b1f52d326`) — land with **no direct kvdb-level unit coverage** in this PR. `wallet/internal/db/kvdb/` gains no `*_test.go` here, and the wallet-side tests routed through these methods use `mockStore`, not the real kvdb adapter.

The 763-line `accountstore_test.go` that covers them (`TestKvdbGetAccountByName`, `TestKvdbGetAccountByNumber`, `TestKvdbGetAccountNotFound`, `TestKvdbListAccountsScopeFilter`, `TestKvdbListAccountsNameFilter`, `TestKvdbRenameAccountBy*`, `TestKvdbGetAccountPopulatesBalance`, `TestKvdbGetAccountSkipBalanceZeros`, `TestKvdbListAccounts{Populates,SkipBalance}*` — 10 of the 16 tests are read-side) lands in PR #1239's `291da1b26 kvdb: implement CreateImportedAccount` commit, alongside the writer tests.

If #1236 ships ahead of #1239 the readers sit in master without direct coverage, so a regression on the kvdb adapter would not show up until #1239 lands. Worth splitting the test file so the 10 reader tests ride with this PR (ideally with the commits that introduced each reader), keeping only the 6 writer tests with their writer commits in #1239.

(`claude-opus-4-7[1m]`)

wallet/account_manager.go
```diff
@@ -169,296 +221,119 @@ type AccountsResult struct {
...
 	}
 
-	manager, err := w.addrStore.FetchScopedKeyManager(scope)
+	info, err := w.cache.GetAccount(ctx, db.GetAccountQuery{
+		WalletID: w.id,
+		Scope:    db.KeyScope(scope),
+		Name:     &name,
+	})
 	if err != nil {
 		return nil, err
```
yyforyongyu, line 294: This now returns the store error directly for missing accounts/scopes, but the legacy public `Wallet.GetAccount` path surfaced `waddrmgr.ManagerError` from `FetchScopedKeyManager`, `LookupAccount`, or `AccountProperties`. Since kvdb translates those into `db.ErrAccountNotFound` and `waddrmgr.IsError` does not unwrap or recognize db errors, existing callers can no longer detect the same failures after this routing. Preserve the `waddrmgr.ManagerError` surface here as the routed `NewAccount`/`RenameAccount`/`ImportAccount` paths do. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -41,3 +374,143 @@ func (s *Store) RenameAccount(ctx context.Context,
...
+		isConfirmed bool)) error {
+
+	if s.txStore == nil || txmgrNs == nil {
+		return nil
+	}
+
+	syncBlock := s.addrStore.SyncedTo()
+	chainParams := s.addrStore.ChainParams()
+
+	unspent, err := s.txStore.UnspentOutputs(txmgrNs)
```
yyforyongyu: `UnspentOutputs` excludes outputs currently locked by `wtxmgr.LockOutput`, but the SQL account-balance queries sum the current UTXO set and do not subtract active leases. After a lease is acquired, kvdb `GetAccount`/`ListAccounts` will report a lower confirmed or unconfirmed account balance than SQL for the same wallet. Use a locked-inclusive credit walk or add the locked credits back before populating the account balances. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Reader tests + supporting fixtures (`TestGetAccountByName`, `TestGetAccountByNumber`, `TestGetAccountNotFound`, `TestListAccountsScopeFilter`, `TestListAccountsNameFilter`, `TestGetAccountPopulatesBalance`, `TestGetAccountSkipBalanceZeros`, `TestListAccountsPopulatesBalance`, `TestListAccountsSkipBalanceZeros`) have been ported into PR-A and folded into the corresponding kvdb commits (`kvdb: implement GetAccount` for the GetAccount + balance variants, `kvdb: implement ListAccounts` for the ListAccounts variants). The `Kvdb` infix was also dropped while porting. The writer tests stay in PR-B alongside their writer commits.

wallet/account_manager.go
yyforyongyu: Fixed: `Wallet.GetAccount` now unwraps `waddrmgr.ManagerError` from the routed store error before returning, matching the `NewAccount` / `RenameAccount` / `ImportAccount` pattern. Callers using `waddrmgr.IsError(err, ...)` keep working.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed: added a new `wtxmgr.UnspentOutputsIncludingLocked` method (separate prep commit) and switched kvdb's `fetchAccountBalances` walk to use it. `AccountInfo.ConfirmedBalance`/`UnconfirmedBalance` now match SQL's behaviour of including leased outputs in the totals.

waddrmgr/interface.go
```diff
@@ -188,18 +188,12 @@ type AddrStore interface {
 	// from the manager's main bucket. Returns ErrWatchingOnly when no
 	// encrypted master HD private key is persisted.
 	//
-	// This method exposes legacy bucket-level access for the
```
yyforyongyu: why deleting this? even if we want to, it should be folded into the previous commit ca16234a8e53125a3fa25f89b7d913f711976b64 ?

waddrmgr/interface.go
```diff
@@ -188,18 +188,12 @@ type AddrStore interface {
...
-	// transitional kvdb adapter. It will be removed when the kvdb
-	// adapter is retired.
 	EncryptedMasterHDPriv(ns walletdb.ReadBucket) ([]byte, error)
 
 	// Decrypt decrypts the supplied ciphertext with the address
 	// manager crypto key identified by keyType. Returns an error if
 	// the manager is locked or the ciphertext is malformed.
 	//
-	// This method exposes legacy crypto-key access for the transitional
```
yyforyongyu: why deleting this? even if we want to, it should be folded into the previous commit ca16234a8e53125a3fa25f89b7d913f711976b64 ?

waddrmgr/interface.go
```diff
@@ -382,9 +376,6 @@ type AccountStore interface {
 	// address that wraps the script. The script is encrypted with
 	// the manager's CryptoKeyScript when isSecretScript is true.
 	//
-	// This method exposes scoped-manager script import for the
```
yyforyongyu: why deleting this? even if we want to, it should be folded into the previous commit ca16234a8e53125a3fa25f89b7d913f711976b64 ?

wallet/account_manager.go
```diff
@@ -147,28 +148,54 @@ func (w *Wallet) NewAccount(_ context.Context, scope waddrmgr.KeyScope,
...
+	// CurrentBlockHeight is the height of the chain tip at snapshot
+	// time.
 	CurrentBlockHeight int32
 }
 
+// propertiesToAccountInfo wraps a waddrmgr.AccountProperties + total
+// balance into the canonical db.AccountInfo shape the public wallet
+// API now exposes. The legacy waddrmgr path does not separate
+// confirmed/unconfirmed balances, so the supplied total is reported
+// on ConfirmedBalance; UnconfirmedBalance stays zero. The Origin
```
yyforyongyu: i think this method will be removed in the end? this is a temporary placeholder

wallet/wallet.go
```diff
@@ -411,6 +411,15 @@ type Wallet struct {
...
 	birthdayBlock waddrmgr.BlockStamp
+
+	// masterFingerprint is the cached BIP32 master-key fingerprint for
+	// this wallet, computed from the wallet's master HD pubkey at
+	// Manager.Load time. The value is the canonical source for derived-
+	// account fingerprints exposed via AccountProperties — the db
+	// layer's per-row value for derived rows is ignored in favor of
+	// this cache. Shell / watch-only wallets that lack a stored master
+	// HD pubkey leave this at zero, matching their existing behavior.
+	masterFingerprint uint32
```
yyforyongyu, line 422: `Wallet.masterFingerprint` is set at `Manager.Load` (`wallet/manager.go:302/323`) but **never read** anywhere in the wallet/ tree — `grep -rn 'w\.masterFingerprint'` returns no consumers, only the field declaration and the assignment. The field's doc-comment says it is "the canonical source for derived-account fingerprints exposed via AccountProperties — the db layer's per-row value for derived rows is ignored in favor of this cache", but with the surface refactor that returns `*db.AccountInfo` directly (commit `8a430686a`) and drops the `(w *Wallet).accountInfoToProperties` override helper, there is no longer any code path that performs the override.

Concrete consequence for kvdb-backed wallets: `waddrmgr` does not persist a master-key fingerprint on derived accounts — `putDefaultAccountInfo` (`waddrmgr/db.go`) writes none, and `AccountProperties.MasterKeyFingerprint` for a derived account reads back as zero. `kvdb.loadAccountInfo` consumes that zero into the returned `db.AccountInfo` and the wallet hands it through:

```go
// wallet/internal/db/kvdb/accountstore.go
// For derived accounts waddrmgr's default-account row stores 0;
// the wallet layer overrides the value from its cached master
// fingerprint via accountInfoToProperties before reaching
// external consumers.
fingerprint := props.MasterKeyFingerprint
```

The comment still references the removed override path. Net effect: `Wallet.GetAccount` / `Wallet.ListAccounts*` on a kvdb-backed wallet return `MasterKeyFingerprint = 0` for every derived account, while SQL backends return the correct value persisted in `accounts.master_fingerprint`. PSBT generation and hardware-wallet origin paths that read `AccountProperties.MasterKeyFingerprint` see a zero on kvdb deployments after this stack lands.

Three ways to resolve:
1. Re-introduce a Wallet-layer override (a thin `func (w *Wallet) overrideDerivedFingerprint(info *db.AccountInfo)` called from `Wallet.GetAccount` / `assembleAccountsResult`).
2. Push the override into the kvdb `Store` itself — wire `w.masterFingerprint` into `kvdb.NewStore` and apply it inside `loadAccountInfo` for `origin == DerivedAccount`. Removes the kvdb/SQL backend divergence entirely.
3. Decide that the deriveFn-returned create-time value is good enough and the read-time value is informational — but then `w.masterFingerprint` is dead code and should be deleted along with its doc-comment claim.

The kvdb tests don't assert `MasterKeyFingerprint` on the returned `db.AccountInfo`, which is why this slipped through; the matching SQL-backend itest at `wallet/internal/db/itest/account_store_test.go` does `require.NotZero(t, derived.MasterKeyFingerprint)` for the create-time return but does not re-read the row, so the divergence isn't surfaced there either.

(`claude-opus-4-7[1m]`)

waddrmgr/interface.go
yyforyongyu: Folded the "transitional kvdb adapter" godoc paragraph back into `ca16234a8` so the method lands with its final docstring; the unrelated `wtxmgr: add UnspentOutputsIncludingLocked` commit no longer touches `waddrmgr/interface.go`. New PR-A tip `3a3467fcc`.

waddrmgr/interface.go
yyforyongyu: Folded the "transitional kvdb adapter" godoc paragraph back into `ca16234a8` so the method lands with its final docstring; the unrelated `wtxmgr: add UnspentOutputsIncludingLocked` commit no longer touches `waddrmgr/interface.go`. New PR-A tip `3a3467fcc`.

waddrmgr/interface.go
yyforyongyu: Folded the "transitional kvdb adapter" godoc paragraph back into `ca16234a8` so the method lands with its final docstring; the unrelated `wtxmgr: add UnspentOutputsIncludingLocked` commit no longer touches `waddrmgr/interface.go`. New PR-A tip `3a3467fcc`.

wallet/account_manager.go
yyforyongyu: Updated: `propertiesToAccountInfo` is now permanent on the reshape chain. The new chain makes the A5-style API change (`*db.AccountInfo` return) the FIRST commit, so the helper bridges `waddrmgr.AccountProperties + balance` into `db.AccountInfo` for the deprecated `Wallet.Accounts` method (`wallet/deprecated.go`) — that legacy API still walks waddrmgr + UnspentOutputs and now feeds its results through this helper to produce `[]db.AccountInfo`. New methods (`GetAccount`, `ListAccounts*`) return cache results directly and do not touch the helper.

wallet/wallet.go
yyforyongyu: Re-introduced the override at the wallet boundary so `w.masterFingerprint` is no longer dead. `Wallet.GetAccount` (in the route-GetAccount commit) and `assembleAccountsResult` (in the route-ListAccounts commit) now set `info.MasterKeyFingerprint = w.masterFingerprint` for `Origin == db.DerivedAccount`. The cached field has consumers again and derived accounts surface the correct canonical fingerprint instead of the zero waddrmgr persists on the default-account row. New PR-A tip `3a3467fcc`.

waddrmgr/interface.go
```diff
@@ -183,6 +183,18 @@ type AddrStore interface {
 
 	// Close cleanly shuts down the manager.
 	Close()
+
+	// EncryptedMasterHDPriv reads the encrypted master HD private key
+	// from the manager's main bucket. Returns ErrWatchingOnly when no
+	// encrypted master HD private key is persisted.
+	//
```
yyforyongyu: remove this `//`

waddrmgr/interface.go
```diff
@@ -183,6 +183,18 @@ type AddrStore interface {
...
+	EncryptedMasterHDPriv(ns walletdb.ReadBucket) ([]byte, error)
+
+	// Decrypt decrypts the supplied ciphertext with the address
+	// manager crypto key identified by keyType. Returns an error if
+	// the manager is locked or the ciphertext is malformed.
+	//
```
yyforyongyu: remove this `//`

waddrmgr/interface.go
```diff
@@ -351,4 +370,13 @@ type AccountStore interface {
 	// and index.
 	DeriveAddr(account uint32, branch uint32, index uint32) (
 		btcutil.Address, []byte, error)
+
+	// ImportWitnessScript imports a user-provided native witness
+	// script under this scoped manager and returns a managed
+	// address that wraps the script. The script is encrypted with
+	// the manager's CryptoKeyScript when isSecretScript is true.
+	//
```
yyforyongyu: remove this `//`

wallet/manager.go
```diff
@@ -324,6 +334,75 @@ func (m *Manager) Load(cfg Config) (*Wallet, error) {
...
+// already failed if this were a real wallet, so encountering it here
+// indicates a wiring bug.
+var errMissingWaddrmgrNamespace = errors.New(
+	"missing waddrmgr namespace",
+)
+
+// resolveMasterFingerprint reads the persisted master HD pubkey via
+// addrMgr, parses it, and returns its BIP32 fingerprint. Shell,
+// watch-only, and pre-master-key wallets have no pubkey persisted —
+// fingerprint is zero. A persisted-but-unparseable value fails the
```
yyforyongyu: remove the sentence: `A persisted-but-unparseable value fails the ...`  and stop leaking local context into the godocs

wallet/internal/db/kvdb/accountstore_timestamps.go
```diff
@@ -0,0 +1,87 @@
...
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/walletdb"
+)
+
+// kvdbAccountCreatedAtBucketKey is the sub-bucket under the waddrmgr
+// namespace that holds per-account creation timestamps. waddrmgr's
+// account row layout does not include a creation timestamp; the kvdb
+// adapter stores it here so the db.AccountInfo.CreatedAt contract is
+// satisfied without changing waddrmgr's bucket layout.
+var kvdbAccountCreatedAtBucketKey = []byte("kvdb-account-created-at")
```
yyforyongyu: remove `kvdb` prefix

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,18 +26,346 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+	createdAt, err := readAccountCreatedAt(ns, scope, accountNumber)
+	if err != nil {
+		return nil, fmt.Errorf("read created-at: %w", err)
+	}
+
+	// For imported accounts the per-row master_fingerprint is the
+	// parent xpub fingerprint persisted on the watch-only row. For
+	// derived accounts waddrmgr's default-account row stores 0;
+	// the wallet layer overrides the value from its cached master
+	// fingerprint via accountInfoToProperties before reaching
```
yyforyongyu: Follow-up nit on the prior `w.masterFingerprint` dead-state thread (now fixed via inline overrides at `wallet/account_manager.go:236` and `wallet/account_manager.go:320` — thanks):

This comment still says "the wallet layer overrides the value from its cached master fingerprint via accountInfoToProperties". The `accountInfoToProperties` helper was deleted in the `wallet: surface db.AccountInfo on read APIs` commit, so the pointer is stale. Reads doing the override now live inline in `Wallet.GetAccount` (around `wallet/account_manager.go:320`) and `Wallet.assembleAccountsResult` (around `wallet/account_manager.go:236`); pointing the comment at those (or just dropping the helper name and saying "the wallet layer overrides this for derived rows after reading") would keep a future maintainer who greps for `accountInfoToProperties` from coming up empty.

Also worth noting: neither the kvdb `accountstore_test.go` reader tests nor `wallet/account_manager_test.go` assert on `MasterKeyFingerprint` at all, so the override is wired but not exercised. A one-line `require.Equal(t, w.masterFingerprint, got.MasterKeyFingerprint)` in `TestGetAccount` / `TestListAccounts` (with a non-zero fixture value) would lock in the fix and catch any future regression that re-drops the override.

(`claude-opus-4-7[1m]`)

wallet/account_manager.go
```diff
@@ -147,318 +148,179 @@ func (w *Wallet) NewAccount(_ context.Context, scope waddrmgr.KeyScope,
...
+func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
+	total btcutil.Amount) db.AccountInfo {
+
+	var pubKey []byte
+	if props.AccountPubKey != nil {
+		pubKey = []byte(props.AccountPubKey.String())
+	}
+
+	origin := db.DerivedAccount
+	if props.IsWatchOnly {
```
yyforyongyu: `IsWatchOnly` is not an imported-account classifier here. `waddrmgr.AccountProperties` sets it for normal derived accounts when the wallet is watch-only or when the account private key is not loaded, so the deprecated `Wallet.Accounts` path can report a locked derived account as `ImportedAccount` and then mask its account number to `0` below. Use the persisted account type/row shape, or pass the real imported-account classification into this conversion. (`openai/gpt-5.5`)

waddrmgr/interface.go
yyforyongyu: Fixed in `e4f3e1fd7` (PR-A tip) — orphan `//` separator line above `EncryptedMasterHDPriv` removed. Folded into the introducing commit `waddrmgr: expose legacy adapter methods on public interfaces`.

waddrmgr/interface.go
yyforyongyu: Fixed in `e4f3e1fd7` — orphan `//` separator line above `Decrypt` removed. Same commit as the other two `//` removals.

waddrmgr/interface.go
yyforyongyu: Fixed in `e4f3e1fd7` — orphan `//` separator line above `ImportWitnessScript` removed. Same commit as the other two `//` removals.

wallet/manager.go
yyforyongyu: Fixed in `e4f3e1fd7` — the `A persisted-but-unparseable value fails the load (bug #45 from PR #1232 review thread #45)` sentence is dropped from `resolveMasterFingerprint`'s godoc in `wallet/manager.go`. The godoc now only describes the function's contract, not internal failure modes.

wallet/internal/db/kvdb/accountstore_timestamps.go
yyforyongyu: Fixed in `e4f3e1fd7` — `kvdbAccountCreatedAtBucketKey` → `accountCreatedAtBucketKey` (identifier rename), and the on-disk value `"kvdb-account-created-at"` → `"account-created-at"` (matches task 48's side-bucket hygiene convention). All in-package callsites updated.

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Fixed in `e4f3e1fd7` — addresses both halves of the comment:

1. **Stale `accountInfoToProperties` reference** removed from the `loadAccountInfo` comment in `wallet/internal/db/kvdb/accountstore.go`. The comment now points at the actual override sites (`Wallet.GetAccount` and `Wallet.assembleAccountsResult`) rather than a deleted helper.

2. **Wallet-level `MasterKeyFingerprint` assertion coverage** added — `TestGetAccount` and `TestListAccounts` now seed `w.masterFingerprint = 0xDEADBEEF`, mock the store to return `MasterKeyFingerprint: 0`, and assert the wallet-layer override applies. The two paths reach the override through different cache methods (`cache.GetAccount` vs `cache.ListAccounts`); both are covered.

Task 48 (kvdb side bucket for derived-account fingerprint) lands on PR-B at `be49cdb12`. Per task 48 — `Wallet-layer override: keep as compatibility fallback` — the override is the canonical fallback for legacy derived accounts (no override-removal planned).

wallet/account_manager.go
yyforyongyu: Fixed in `e4f3e1fd7` via new PR-A tip commit `wallet: classify imported accounts off persisted signal, not IsWatchOnly`.

`propertiesToAccountInfo` now takes an explicit `isImported bool` parameter sourced from `waddrmgr.AccountStore.IsImportedAccount` (the persisted account-type signal, not the lock-state-dependent `IsWatchOnly`). `Wallet.Accounts` in `wallet/deprecated.go` computes it per account inside the `ForEachAccount` walk via a parallel `importedList []bool` and threads it into the conversion loop.

`waddrmgr.IsImportedAccount` itself short-circuits to `true` for the legacy `ImportedAddrAccount` slot (whose row layout would otherwise fail `loadAccountInfo` decryption on watch-only fixtures).

Unit tests pin both sides of the classifier:
- `TestPropertiesToAccountInfoLockedDerivedNotMisclassified` — derived account with `IsWatchOnly: true` (locked wallet) stays classified as `DerivedAccount` with its `AccountNumber` unmasked.
- `TestPropertiesToAccountInfoImportedClassifiedAndMasked` — imported account classified correctly with `AccountNumber` masked to `0`.

Note: the fix landed as a standalone commit at the PR-A tip rather than folded into the introducing commit, because folding required cross-commit dependency reordering (the `IsImportedAccount` interface method is added in a later PR-A commit). The bug is latent today (no CI test exercises locked-wallet `Wallet.Accounts`).

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,18 +26,345 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
+func loadAccountInfo(ns walletdb.ReadBucket,
+	scopedMgr waddrmgr.AccountStore, accountNumber uint32,
+	walletWatchOnly bool) (*db.AccountInfo, error) {
+
+	props, err := scopedMgr.AccountProperties(ns, accountNumber)
+	if err != nil {
+		return nil, translateAccountErr(err, db.ErrAccountNotFound)
+	}
+
+	accountIsImported, err := scopedMgr.IsImportedAccount(
```
yyforyongyu, line 164: Full-stack update: if kvdb is intentionally preserving the legacy waddrmgr behavior and the old imported-address pseudo-account is not meant to become a normal `AccountStore` row in this stack, this is not a blocker. The only remaining edge is error-shape polish: a direct lookup of the reserved `"imported"` name falls through to the row-loader error instead of a clean not-found/unsupported result. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🪞
(`claude-opus-4-7[1m]`)

yyforyongyu:
Re-reviewed against `cbf2724e6`. The new `waddrmgr: reserve imported account slot` commit tightens `deriveAccountKey` / `advanceLastAccountLocked` / `PutDerivedAccountWithKeys` / `newAccountWatchingOnly` from `> MaxAccountNum` to `>= MaxAccountNum` and lets the new `PutWatchOnlyAccountWithKeys` use the reserved slot via a caller-owned collision check; the wrap-error path in `newAccount` now preserves the typed `ErrAccountNumTooHigh` instead of burying it under `ErrKeyChain`. The off-by-one shifts the legitimate-derived ceiling from `MaxAccountNum` to `MaxAccountNum - 1` — only relevant to wallets with ~2 billion accounts in practice, so the back-compat exposure is theoretical.

Prior PR threads (override fingerprint via `w.masterFingerprint`, stale `accountInfoToProperties` reference in kvdb's `loadAccountInfo`) are all addressed: the inline override + tests in `TestListAccounts` / `TestGetAccount` lock in the kvdb derived-fingerprint behavior, and the kvdb comment now points at the right Wallet sites.

LGTM 🔐
(`claude-opus-4-7[1m]`)

yyforyongyu:
Re-reviewed current head `cbf2724e6` together with PRs #1240 and #1239. I did not find a new non-duplicate issue in this pass, but the already-open wallet error-surface thread still appears to apply for not-found cases: kvdb translates account/scope not-found errors to `db.ErrAccountNotFound` before the wallet sees them, so the `errors.As(... waddrmgr.ManagerError)` compatibility path in `Wallet.GetAccount` cannot fire for those missing-account/scope cases.

(`openai/gpt-5.5`)

wallet/internal/db/kvdb/accountstore.go
yyforyongyu: Addressed in the pushed stack: sanitizeAccountNumber maps waddrmgr.ImportedAddrAccount to db.ErrAccountNotFound, and TestGetAccountReservedImportedNameNotFound covers GetAccount by the reserved imported name.

yyforyongyu:
LGTM 🔑
(`claude-opus-4-7[1m]`)

yyforyongyu:
LGTM (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🔖
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/accountstore.go
```diff
@@ -21,18 +26,355 @@ func (s *Store) CreateImportedAccount(ctx context.Context,
...
-	_ db.ListAccountsQuery) ([]db.AccountInfo, error) {
+// collectScopedAccounts walks the accounts under scopedMgr and appends
+// matching entries to accounts. The reserved "imported addresses" pseudo-
+// account and entries that fail the optional Name filter are skipped.
+func collectScopedAccounts(ns walletdb.ReadBucket,
+	scopedMgr waddrmgr.AccountStore, query db.ListAccountsQuery,
+	walletWatchOnly bool, accounts *[]db.AccountInfo,
+	internalNumbers *[]uint32) error {
+
+	return scopedMgr.ForEachAccount(ns, func(account uint32) error {
```
yyforyongyu: `ForEachAccount` walks the account bucket in raw key order, but waddrmgr encodes account numbers little-endian (`uint32ToBytes`). That order is not numeric once account 256 exists (`[0 1 0 0]` sorts before account 1), while the SQL queries and shared contract order derived accounts by account number. Please sort the collected kvdb rows by the decoded account number before returning so long-lived wallets do not see backend-specific `ListAccounts` ordering. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -213,333 +213,122 @@ func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
...
-		)
-
-		return err
-	})
-	if err != nil {
-		return nil, err
 	}
 
-	// Include the wallet's current sync state in the result.
 	syncBlock := w.addrStore.SyncedTo()
```
yyforyongyu: `cache.ListAccounts` now returns balances that kvdb classifies against `addrStore.SyncedTo()` inside the store read, but the `CurrentBlockHeight` attached here is read afterward. If the wallet sync height advances between those two calls, an output can be reported in `UnconfirmedBalance` while this result's height says it has enough confirmations to be confirmed. Please take the block reference used for balance classification from the same snapshot as the account read, or plumb the store's sync height back with the balances. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🗂️
(`claude-opus-4-7[1m]`)

yyforyongyu:
LGTM 👍 (`openai/gpt-5.5`)

GustavoStingelin (approved):
LGTM

PR #1237, opened 2026-05-18
wallet+db+waddrmgr: prep for address-manager Store routing

yyforyongyu:
## Summary

Prep stack that lands the foundation pieces the upcoming address-manager
Store/cache routing PR (#1214) needs, without touching
`wallet/address_manager.go` routing or the kvdb address implementation.

Base branch: `impl-account-manager-store` (PR #1215).
Follow-up PR (kvdb impl + manager routing): `impl-address-manager-store`
(#1214).

## What's in scope

- **`waddrmgr` exports**:
  - export `ManagedPubKeyAddressHasPrivateKey` so the kvdb address store can
    derive the spendable-vs-watch-only signal without reaching into private
    fields
  - export the taproot script codec used by imported tap addresses so kvdb and
    SQL backends can encode/decode via the same path
- **`db` contract structural changes**:
  - align the address metadata contract by surfacing account number/name,
    key-scope, master fingerprint, pubkey, script marker, watch-only, and
    used-state on `AddressInfo`
  - thread that metadata through `AddressInfoRow` plumbing in both pg and
    sqlite adapters
  - add ADR 0011 and derive SQL address `IsUsed` from `EXISTS` queries over
    `utxos` rather than persisting an `addresses.used` column
  - project account metadata in SQL address read queries so list/read paths do
    not perform per-address account lookups
- **SQL address creation**:
  - populate returned address metadata from data already available in the write
    transaction, keeping create operations all-or-nothing from the caller's
    perspective
- **mocks/tests**:
  - bump wallet mocks and tests to expose the expanded address contract ahead
    of the routing PR's test rewrites

## What's NOT in scope

- routing `wallet/address_manager.go` through `Store`
- kvdb address-store implementation work (derived-create, import, list, type
  metadata, secrets) -- these land in the follow-up
  `impl-address-manager-store` PR (#1214)
- a persisted SQL `addresses.used` column; ADR 0011 intentionally avoids that
  column and uses the `utxos` table as the SQL source of truth

## Test plan

- [ ] `GOWORK=off go test ./wallet ./wallet/internal/db/...`
- [ ] `GOWORK=off make itest-db db=sqlite`
- [ ] `GOWORK=off make itest-db db=postgres`
- [ ] `GOWORK=off make lint-check`


wallet/internal/sql/pg/sqlc/addresses.sql.go
```diff
@@ -323,3 +329,22 @@ func (q *Queries) ListAddressesByAccount(ctx context.Context, arg ListAddressesB
...
+SET used = TRUE
+WHERE wallet_id = $1 AND script_pub_key = $2
+`
+
+type MarkAddressUsedParams struct {
+	WalletID     int64
+	ScriptPubKey []byte
+}
+
+// Sets the address `used` flag to TRUE. The monotonicity trigger
```
yyforyongyu: The generated comment does not start with the exact method name `MarkAddressUsed`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/sql/sqlite/sqlc/addresses.sql.go
```diff
@@ -323,3 +329,22 @@ func (q *Queries) ListAddressesByAccount(ctx context.Context, arg ListAddressesB
...
+SET used = 1
+WHERE wallet_id = ? AND script_pub_key = ?
+`
+
+type MarkAddressUsedParams struct {
+	WalletID     int64
+	ScriptPubKey []byte
+}
+
+// Sets the address `used` flag to 1. The monotonicity trigger
```
yyforyongyu: The generated comment does not start with the exact method name `MarkAddressUsed`, which violates the Go Comment Gate rule; update the doc comment in the SQL source file to fix this. (`gemini-3.1-pro-preview`)

wallet/internal/db/addresses_common.go
```diff
@@ -440,6 +578,18 @@ type DerivedAddressAdapters[QTX any, AccountRow any, AccountParams any,
...
+	// GetAccountAddrSchema extracts the effective address schema for the
+	// account from the looked-up row. For SQL backends the coalesce in
+	// the account-read queries already materializes the per-account
+	// override (or the scope default when no override is set), so the
+	// returned schema can be used directly by derivedAddressInput
+	// without consulting the scope default again. Backends that route
+	// through this adapter (pg, sqlite) must populate this so derived
+	// addresses for accounts with a per-account override (e.g. a
+	// BIP-0044 key imported as NestedWitnessPubKey) honor the override
+	// rather than silently using the scope-level default.
```
yyforyongyu: The adapter-level doc-comment for `GetAccountAddrSchema` (lines 581-590) describes a contract the SQL adapter doesn't actually implement:

> For SQL backends the coalesce in the account-read queries already materializes the per-account override (or the scope default when no override is set), so the returned schema can be used directly by derivedAddressInput without consulting the scope default again. Backends that route through this adapter (pg, sqlite) must populate this so derived addresses for accounts with a per-account override (e.g. a BIP-0044 key imported as NestedWitnessPubKey) honor the override rather than silently using the scope-level default.

But the pg and sqlite adapters (`wallet/internal/db/pg/accounts.go:367-395` and the sqlite mirror) both return `db.ScopeAddrMap[scope]` — the scope default, ignoring any per-account override. Their own doc-comments explicitly admit it: "The SQL backend does not model a per-account override, so the schema is fully determined by (purpose, coin_type)."

There's no `addr_schema` / `address_schema` column on `accounts` in either migration set, and no COALESCE in the account-read queries — so the "coalesce in the account-read queries already materializes the per-account override" claim doesn't hold today.

Two ways to resolve:
1. Tighten the adapter doc-comment to match reality: "SQL backends derive the schema from (purpose, coin_type). Per-account override is not currently modeled; imported accounts whose addrType differs from the scope default will silently use the scope default." Then file a follow-up for the override-support work.
2. Land per-account schema modeling now (column + COALESCE in account-read queries + adapter lookup) so the doc-comment becomes true.

Worth picking one before this lands so future maintainers don't trust the contract.

(`claude-opus-4-7[1m]`)

yyforyongyu (commented):
LGTM 🚀 (`openai/gpt-5.5`)

wallet/db_ops.go
```diff
@@ -12,8 +12,10 @@ import (
 	"github.com/btcsuite/btcd/btcutil"
 	"github.com/btcsuite/btcd/btcutil/hdkeychain"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	db "github.com/btcsuite/btcwallet/wallet/internal/db"
```
yyforyongyu: unnecessary alias

wallet/internal/db/pg/addresses.go
```diff
@@ -206,20 +209,22 @@ func derivedAddressGetIntIndex(qtx *sqlc.Queries) func(context.Context,
 	return qtx.GetAndIncrementNextInternalIndex
 }
 
-// derivedAddressCreateAddr returns the derived address insert helper.
-func derivedAddressCreateAddr(qtx *sqlc.Queries) func(
-	context.Context, int64, int64, db.AddressType, uint32, uint32, []byte,
-) (sqlc.CreateDerivedAddressRow, error) {
+// createDerivedAddressFunc defines the derived address insert callback.
+type createDerivedAddressFunc = func(context.Context, int64, int64,
```
yyforyongyu: why do we need the rename here?

wallet/internal/db/pg/addresses.go
```diff
@@ -332,6 +337,7 @@ type addressInfoRow interface {
 // addressRowToInfo converts a PostgreSQL address row to an AddressInfo
 // struct.
 func addressRowToInfo[T addressInfoRow](row T) (*db.AddressInfo, error) {
+
```
yyforyongyu: noise diff?

waddrmgr/tlv.go
```diff
@@ -22,32 +35,25 @@ const (
 	typeTapLeafScript  tlv.Type = 2
 )
 
-// tlvEncodeTaprootScript encodes the given internal key and full set of taproot
-// script leaves into a byte slice encoded as a TLV stream.
-func tlvEncodeTaprootScript(s *Tapscript) ([]byte, error) {
```
yyforyongyu, line 27: why don't we leave this `tlvEncodeTaprootScript` untouched, and introduce `EncodeTaprootScript` by wrapping this method, similar to `DecodeTaprootScript`? seems we can add less change that way?

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -27,6 +27,13 @@ func notImplemented(_ context.Context, method string) error {
 	return fmt.Errorf("kvdb.Store.%s: %w", method, errNotImplemented)
 }
 
+// IsNotImplemented reports whether err was produced by a kvdb Store method
+// stub that has not yet been implemented. Wallet-side fallback paths use
+// this predicate to detect when to fall back to legacy walletdb walks.
+func IsNotImplemented(err error) bool {
```
yyforyongyu: remove this stupid thin wrapper

wallet/internal/db/pg/addresses.go
```diff
@@ -376,7 +435,9 @@ func listAddressesByAccount(ctx context.Context, q *sqlc.Queries,
 
 	items := make([]db.AddressInfo, len(rows))
 	for i, row := range rows {
-		item, err := addressRowToInfo(row)
+		// TODO(yy): Avoid fetching account properties for every row once the
```
yyforyongyu: The TODO is honest, but the fix is essentially free in this PR's footprint and the cost is genuinely bad: every address row in a `ListAddresses` result triggers its own `GetAccountPropsById` round-trip (N round-trips per page in pg, same in `wallet/internal/db/sqlite/addresses.go:432`). For a wallet listing 100 addresses that's 1 list + 100 account-props = 101 round-trips to materialize one page.

`addressRowToInfo` already receives a row from a query that JOINs `accounts` and `wallets` for `acc.master_fingerprint` and `w.master_hd_pub_key` etc., so the four extra columns AddressInfo needs (`account_number`, `account_name`, plus `ks.purpose` / `ks.coin_type` via the existing `key_scopes` JOIN on `ListAddressesByAccount`) can be added to the SELECT list — no extra JOINs. `GetAddressByScriptPubKey` does need to add the `key_scopes` JOIN, but that's a one-line change.

Three concrete changes close the TODO entirely:

1. Extend `GetAddressByScriptPubKey` and `ListAddressesByAccount` to select `acc.account_number`, `acc.account_name`, `ks.purpose`, `ks.coin_type` (plus the JOIN on `key_scopes` for the script-key variant).
2. Plumb those four into `AddressInfoRow` directly and let `AddressRowToInfo` use them via `convertAccountMetadata`.
3. Drop the secondary `getAccountProps` call from `addressRowToInfo` and the `AccountProps` bypass slot on `AddressInfoRow` (no longer needed).

That collapses the address-read path back to a single round-trip, lets the create path stay as it is (it already has the account row in hand from the surrounding tx), and removes the optional-bypass field that's only there to paper over the N+1.

(`claude-opus-4-7[1m]`)

wallet/internal/db/addresses_common.go
```diff
@@ -190,6 +220,12 @@ type AddressInfoRow[TypeID, OriginIDType any] struct {
 	// PubKey is the public key. Zero value for derived addresses.
 	PubKey []byte
 
+	// IsUsed is the monotonic on-chain-usage flag persisted with the
+	// address row. The address store keeps this in sync with the legacy
+	// MarkUsed semantics so unused-address scans never re-derive a key
```
yyforyongyu: Stale doc-comment after ADR 0011. `IsUsed` is no longer "the monotonic on-chain-usage flag persisted with the address row" — per the ADR and `ef02242d2`, SQL backends derive it from `EXISTS(SELECT 1 FROM utxos WHERE address_id = ?)` at read time, no column persisted. The "address store keeps this in sync with the legacy MarkUsed semantics" line is also stale: SQL adapters don't update an `IsUsed` column (there isn't one), and there's no `MarkAddressUsed` adapter method in the codebase (`grep -rn 'MarkAddressUsed' wallet/` returns no hits) — kvdb's sticky bit lives in waddrmgr's bucket layout, not on this row.

The public `AddressInfo.IsUsed` doc in `data_types.go:663` was rewritten correctly ("Backends populate this field from different sources… SQL backends derive it from the utxos table at read time"); this internal `AddressInfoRow.IsUsed` doc just needs the same treatment so a maintainer reading it doesn't go looking for a column or a MarkUsed call site that doesn't exist.

(`claude-opus-4-7[1m]`)

wallet/internal/sql/pg/queries/addresses.sql
```diff
@@ -22,9 +22,16 @@ SELECT
     a.pub_key,
     a.created_at,
     acc.origin_id,
+    acc.master_fingerprint AS account_master_fingerprint,
+    w.master_hd_pub_key AS wallet_master_hd_pub_key,
     w.is_watch_only AS wallet_is_watch_only,
     (s.encrypted_priv_key IS NOT NULL)::BOOLEAN AS has_private_key,
-    (s.encrypted_script IS NOT NULL)::BOOLEAN AS has_script
+    (s.encrypted_script IS NOT NULL)::BOOLEAN AS has_script,
+    exists(
```
yyforyongyu, line 33: Deriving `is_used` only from current `utxos` rows makes the flag clearable. `DeleteTxWithOps` calls `DeleteCreatedUtxos` before removing an unmined leaf tx, so if that tx was the only credit to this address, the next read returns false even though `AddressInfo.IsUsed` and ADR 0011 promise a monotonic ever-observed value. That lets unused-address scans re-offer an address after a published unmined tx is deleted. The SQL store needs a durable usage record, or `DeleteTx` must preserve enough history for this query. (`openai/gpt-5.5`)

wallet/internal/db/pg/addresses.go
```diff
@@ -131,20 +138,31 @@ func (s *Store) NewDerivedAddress(ctx context.Context,
...
 
-	return db.NewDerivedAddressWithTx(
+	info, err := db.NewDerivedAddressWithTx(
 		ctx, params, s.execWrite, adapters, deriveFn,
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	err = s.addAddressAccountProperties(ctx, info)
```
yyforyongyu: `NewDerivedAddressWithTx` has already run `s.execWrite` to completion when this read starts, so a transient read failure here returns an error after the address and index allocation are committed. Retrying `NewDerivedAddress` then allocates another index and leaves the first address persisted but never returned; the imported path below has the same post-commit failure shape and can turn a retry into a duplicate-script failure. Since the account row was already loaded inside the write transaction, populate these fields before committing, or keep this metadata read inside the same transaction so create remains all-or-nothing from the caller perspective. (`openai/gpt-5.5`)

wallet/internal/db/pg/accounts.go
```diff
@@ -364,6 +364,37 @@ func derivedAddressGetWalletWatchOnly(
...
+	coin, err := db.Int64ToUint32(row.CoinType)
+	if err != nil {
+		return db.ScopeAddrSchema{}, fmt.Errorf("coin: %w", err)
+	}
+
+	scope := db.KeyScope{Purpose: purpose, Coin: coin}
+	schema, ok := db.ScopeAddrMap[scope]
+
+	if !ok {
+		return db.ScopeAddrSchema{}, fmt.Errorf(
```
yyforyongyu: remember we have a special rule for `fmt.Errorf` wrapping? we should use the compact line wrapping here

wallet/internal/sql/pg/queries/addresses.sql
yyforyongyu: Intended behavior — see ADR 0011 §"Scope" (added in `796c1c653`). `DeleteTx` is the user-triggered abandon path; clearing `is_used` along with the deleted utxos matches the rest of the abandon semantics (balances revert, change indices become reusable). The privacy-strict "once exposed, always exposed" reading is not the policy at the wallet layer.

yyforyongyu:
LGTM 🪶
(`claude-opus-4-7[1m]`)

wallet/internal/db/addresses_common.go
```diff
@@ -590,34 +738,36 @@ func derivedAddressInput(ctx context.Context,
...
+		return 0, 0, 0, nil, nil, fmt.Errorf("address index: %w", err)
 	}
 
 	acctID, err := Int64ToUint32(accountID)
 	if err != nil {
-		return 0, 0, 0, nil, fmt.Errorf("account ID: %w", err)
+		return 0, 0, 0, nil, nil, fmt.Errorf("account ID: %w", err)
 	}
 
 	derivedData, err := deriveFn(ctx, acctID, branch, index)
```
yyforyongyu: Passing `acctID` here exposes the SQL account row ID to `AddressDerivationFunc`, but wallet derivation expects the BIP44 account number, so derive with/pass `AccountNumber` instead of `AccountID`. (`openai/gpt-5.5`)

wallet/internal/db/addresses_common.go
```diff
@@ -554,24 +702,24 @@ func createDerivedAddress[T any](ctx context.Context,
 		Branch:       branch,
 		Index:        index,
 		ScriptPubKey: scriptPubKey,
+		PubKey:       pubKey,
```
yyforyongyu, line 745: Returning `pubKey` for derived addresses contradicts the `AddressInfo.PubKey` contract that says derived rows are empty, so update the contract/schema comments or stop populating this field. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🧱
(`claude-opus-4-7[1m]`)

wallet/internal/db/pg/accounts.go
```diff
@@ -372,6 +378,29 @@ func derivedAddressGetAccountID(
 	return row.ID
 }
 
+// derivedAddressGetAccountNumber extracts the derived account number from a
+// row.
+func derivedAddressGetAccountNumber(
```
yyforyongyu, line 384: can this be put into `accounts_common.go` instead?

wallet/internal/db/pg/accounts.go
```diff
@@ -380,6 +409,35 @@ func derivedAddressGetWalletWatchOnly(
 	return row.WalletIsWatchOnly
 }
 
+// derivedAddressGetAccountAddrSchema returns the scope-default
+// address schema for the account derived from the returned row.
+// The SQL backend does not model a per-account override, so the
+// schema is fully determined by (purpose, coin_type).
+func derivedAddressGetAccountAddrSchema(
```
yyforyongyu, line 400: can this be put into `accounts_common.go` instead?

wallet/internal/db/sqlite/accounts.go
```diff
@@ -377,6 +383,29 @@ func derivedAddressGetAccountID(
 	return row.ID
 }
 
+// derivedAddressGetAccountNumber extracts the derived account number from a
+// row.
+func derivedAddressGetAccountNumber(
```
yyforyongyu, line 389: same here - can this be put into `accounts_common.go` instead?

wallet/internal/db/sqlite/accounts.go
```diff
@@ -385,6 +414,35 @@ func derivedAddressGetWalletWatchOnly(
 	return row.WalletIsWatchOnly
 }
 
+// derivedAddressGetAccountAddrSchema returns the scope-default
+// address schema for the account derived from the returned row.
+// The SQL backend does not model a per-account override, so the
+// schema is fully determined by (purpose, coin_type).
+func derivedAddressGetAccountAddrSchema(
```
yyforyongyu, line 405: same here - can this be put into `accounts_common.go` instead?

wallet/internal/db/sqlite/addresses.go
```diff
@@ -208,15 +212,14 @@ func derivedAddressGetIntIndex(
 }
 
 // derivedAddressCreateAddr returns the derived address insert helper.
-func derivedAddressCreateAddr(
-	qtx *sqlc.Queries,
-) func(context.Context, int64, int64, db.AddressType, uint32, uint32, []byte) (
-	sqlc.CreateDerivedAddressRow, error,
-) {
+func derivedAddressCreateAddr(qtx *sqlc.Queries) func(
```
yyforyongyu, line 217: same here - can this be put into `addresses_common.go` instead?

wallet/internal/db/sqlite/accounts.go
```diff
@@ -385,6 +414,35 @@ func derivedAddressGetWalletWatchOnly(
...
+		return db.ScopeAddrSchema{}, fmt.Errorf("purpose: %w", err)
+	}
+
+	coin, err := db.Int64ToUint32(row.CoinType)
+	if err != nil {
+		return db.ScopeAddrSchema{}, fmt.Errorf("coin: %w", err)
+	}
+
+	scope := db.KeyScope{Purpose: purpose, Coin: coin}
+	schema, ok := db.ScopeAddrMap[scope]
```
yyforyongyu: This still derives from the hard-coded scope default instead of the address schema persisted on `key_scopes`: `CreateImportedAccount` can create the scope with `params.AddrSchema`, so a strict BIP49 import can persist `NestedWitnessPubKey` for the internal branch but later derive change as the BIP49+ default `WitnessPubKey`. Please return the stored internal/external type IDs from the account lookup and build the schema from those values. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🪶 (`openai/gpt-5.5`)

yyforyongyu:

yyforyongyu:
LGTM 🪡
(`claude-opus-4-7[1m]`)

yyforyongyu:
LGTM 🧪 (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🧵
(`claude-opus-4-7[1m]`)

yyforyongyu (commented):
LGTM 🎰 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪁 (`openai/gpt-5.5`)

yyforyongyu (commented):
LGTM 🎰 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🌿
(`claude-opus-4-7[1m]`)

yyforyongyu (commented):
LGTM 🎰 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🌻
(`claude-opus-4-7[1m]`)

wallet/internal/db/addresses_common.go
```diff
@@ -641,9 +834,23 @@ func NewDerivedAddressWithTx[QTX any, AccountRow any,
 		if err == nil {
 			accountID := adapters.GetAccountID(row)
 
+			accountNumber, errAccount := adapters.GetAccountNumber(row)
```
yyforyongyu, line 842: This rejects imported xpub accounts before address derivation. `CreateImportedAccount` stores them with NULL `account_number`, but `NewDerivedAddress` looks up by name then unconditionally requires `DerivedAddressAccountNumber`; creating a receive address for an imported account returns `ErrNilDBAccountNumber` even though `AccountPubKey` below is enough to derive branch/index. Skip the account-number requirement for imported-account rows or pass a nullable number to the callback. (`gpt-5.5`)

yyforyongyu: Fixed. `NewDerivedAddressWithTx` now tolerates `ErrNilDBAccountNumber` from `adapters.GetAccountNumber(row)` — imported accounts have NULL `account_number` and the callback derives from `AccountPubKey` plus Branch/Index instead. `AddressDerivationParams.AccountNumber`'s docstring is updated to note it is 0 for imported accounts and that callbacks must use `AccountPubKey` only in that case. Regression test `TestNewDerivedAddressOnImportedAccount` covers an imported xpub deriving an external address. Landed on the wire-up commit; new tip 2b3c8298e5.

yyforyongyu: I still see the same failure at the current tip `2b3c8298`: `NewDerivedAddressWithTx` returns on any `adapters.GetAccountNumber(row)` error before calling `createDerivedAddress`, and the pg/sqlite adapters still return `ErrNilDBAccountNumber` for NULL `account_number`. Imported xpub accounts therefore still fail before the callback can use `AccountPubKey`. (`gpt-5.5`)

yyforyongyu: You'\''re right — apologies, the earlier amend landed in my working tree but never made it into the committed state during the cascade-rebase autostash dance. The fix is now actually committed at tip 7577172e50:

- `NewDerivedAddressWithTx` tolerates `ErrNilDBAccountNumber` via `errors.Is`; imported xpub accounts pass `accountNumber=0` to the callback.
- `AddressDerivationParams.AccountNumber` docstring documents the 0-for-imported semantics.
- `TestNewDerivedAddressOnImportedAccount` (in `address_store_test.go`) is the regression — creates an imported xpub account via `CreateImportedAccount` (NULL `account_number`, real `PublicKey`, no `EncryptedPrivateKey`), calls `NewDerivedAddress`, asserts the address is created with `AccountNumber == 0`.

Verified locally: `go test -tags itest -run TestNewDerivedAddressOnImportedAccount` passes. Per-commit `go vet ./...` clean across all 11 commits on the branch.

yyforyongyu: Verified at `7577172e50`: `ErrNilDBAccountNumber` now falls through with `accountNumber=0`, so imported xpub accounts can reach the derivation callback with `AccountPubKey`. LGTM. (`gpt-5.5`)

docs/developer/adr/0011-no-addresses-used-column.md
```diff
@@ -0,0 +1,84 @@
...
+
+Because observed credits remain represented in SQL history, address used-ness
+can be derived from transaction state:
+
+```sql
+SELECT EXISTS(SELECT 1 FROM utxos WHERE address_id = ?)
+```
+
+Storing a second `addresses.used` flag would duplicate the same fact and create
+drift risk between address metadata and wallet history.
```
GustavoStingelin, line 24: 👍🏼 👍🏼 👍🏼

docs/developer/adr/0011-no-addresses-used-column.md
```diff
@@ -0,0 +1,84 @@
...
+`addresses` table.
+
+- SQL address-read queries project `db.AddressInfo.IsUsed` from `EXISTS` over
+  `utxos`.
+- The Store interface intentionally has no SQL `MarkAddressUsed` method:
+  recording an observed wallet transaction inserts the `utxos` row that future
+  address reads consult via the `EXISTS` projection.
+- The kvdb backend continues to populate `IsUsed` from waddrmgr's sticky used
+  bit, because legacy rollback handling does not provide the same durable SQL
+  history table.
```
GustavoStingelin, line 45: 👍🏼

wallet/internal/db/addresses_common.go
```diff
@@ -137,6 +137,16 @@ func (p NewImportedAddressParams) HasSecretMaterial() bool {
...
 
+// HasScript returns true if the params include script spend data.
+func (p NewImportedAddressParams) HasScript() bool {
+	return len(p.EncryptedScript) > 0
+}
+
+// IsWatchOnly returns true if the params do not include private key material.
+func (p NewImportedAddressParams) IsWatchOnly() bool {
+	return !p.HasPrivateKey()
+}
```
GustavoStingelin, line 148: Non blocking: in the previous implementation, I was trying to avoid adding this method and rely only on `HasPrivateKey` where needed, to keep the behavior less ambiguous.

wallet/internal/db/accounts_common.go
```diff
@@ -367,6 +367,40 @@ func getKeyCounts(external, internal, imported int64) (uint32, uint32,
 	return externalKeyCount, internalKeyCount, importedKeyCount, nil
 }
 
+// DerivedAddressAccountNumber converts a derived account number from an
+// account lookup row to the wallet-compatible uint32 account number.
+func DerivedAddressAccountNumber(accountNumber sql.NullInt64) (uint32,
```
GustavoStingelin, line 372: Looks like I’ll need to handle some conflicts here in my WIP branch while refactoring the accounts flow.

I saw that the other PRs you already requested my review on are not touching this area anymore. Do you still think this will be needed in any other PR?

wallet/internal/sql/sqlite/queries/addresses.sql
```diff
@@ -15,18 +15,31 @@ INSERT INTO address_secrets (
...
     acc.origin_id,
+    acc.master_fingerprint,
     w.is_watch_only AS wallet_is_watch_only,
+    cast(
+        EXISTS (
+            SELECT 1
+            FROM utxos AS u
+            WHERE u.address_id = a.id
+        ) AS BOOLEAN
+    ) AS is_used,
```
GustavoStingelin, line 37: I guess this operation may become expensive on the DB side with a high address volume, but we can defer that discussion until we have benchmarks.

wallet/internal/db/interface.go
```diff
@@ -230,18 +230,46 @@ type AccountStore interface {
...
-// address index to derive the actual address data (script_pub_key). As the
-// database should not know about how to derive an address, we pass this as a
-// callback.
-type AddressDerivationFunc func(ctx context.Context, accountID uint32,
-	branch uint32, index uint32) (*DerivedAddressData, error)
+// AddressDerivationParams contains the wallet-side data needed to derive a
+// store-allocated address.
+type AddressDerivationParams struct {
+	// Scope is the key scope that owns the derived address.
+	Scope KeyScope
```
GustavoStingelin: I guess this one should be placed in data_types?

wallet/internal/db/interface.go
```diff
@@ -230,18 +230,46 @@ type AccountStore interface {
...
+// address index. The callback receives a value struct so new derivation inputs
+// can be added without changing every call site again.
+type AddressDerivationFunc func(ctx context.Context,
+	params AddressDerivationParams) (*DerivedAddressData, error)
 
 // DerivedAddressData contains the derived address information returned by
 // the AddressDerivationFunc callback.
 type DerivedAddressData struct {
 	// ScriptPubKey is the script public key for the derived address.
 	ScriptPubKey []byte
```
GustavoStingelin: this one also looks like it was incorrectly added here before

GustavoStingelin (approved):
LGTM! 🚢🚢 🚢

wallet/internal/db/addresses_common.go
yyforyongyu: yeah we have an update in this area, will chat about it next time - tldr version is, we don't need these methods anymore.

wallet/internal/db/accounts_common.go
yyforyongyu: yeah i don't think they are needed in another PR.

wallet/internal/db/interface.go
yyforyongyu: correct, now fixed

wallet/internal/db/interface.go
yyforyongyu: also fixed

PR #1238, opened 2026-05-18
wallet: migrate UTXO store + lease ops to db.Store

yyforyongyu:
## Summary

Complete UTXO-store migration through `db.Store`. Implements all four
`UtxoStore` methods on the kvdb backend and routes their wallet-side
consumers through the store — folded together so reviewers see the
kvdb impl and its wallet caller in the same chain.

- kvdb impls (with bundled tests): `GetUtxo`, `ListUTXOs`,
  `LeaseOutput`, `ListLeasedOutputs`
- wallet routes through the store: `GetUtxo`, `ListUnspent`,
  `LeaseOutput`, leased-outputs query
- shared test scaffolding: `kvdb/utils_test.go` matchers; wallet
  conversion helpers

## Why

- delivers the UTXO portion of the kvdb → db.Store migration in one
  reviewable chain instead of splitting it across multiple PRs
- mirrors the kvdb-impl + wallet-route interleave used on PR #1214
  (impl-address-manager-store) so each method's impl and its caller
  land together
- supersedes the previous split into PR #1203 (manager routing) and
  this PR (kvdb impls); #1203's content is now in this stack

## What Changed

- `wallet/internal/db/kvdb/utxostore.go`: implement `GetUtxo`,
  `ListUTXOs`, `LeaseOutput`, `ListLeasedOutputs` against the legacy
  `wtxmgr` / `waddrmgr` reads
- `wallet/internal/db/kvdb/utxostore_test.go`: unit coverage for each
- `wallet/utxo_manager.go`: wallet flows now consult `w.store` instead
  of touching `wtxmgr` directly
- `wallet/utxo_manager_test.go`: tests updated for store-backed mocks
- `wallet/internal/db/kvdb/utils_test.go`: shared test scaffolding
  (`testLegacyAddrStore`)

## Testing

- per-commit `GOWORK=off go build ./...` clean across all 10 commits
- `make lint`: 0 issues
- unit + itest CI: green

### DIFF
--- a/wallet/internal/db/kvdb/addressstore.go
+++ b/wallet/internal/db/kvdb/addressstore.go
@@ -36,6 +36,10 @@ var (
 	errMissingAddrmgrNamespace = errors.New(
 		"kvdb: missing waddrmgr namespace",
 	)
+
+	// errNoAddressInPkScript is returned when a script contains no standard
+	// address.
+	errNoAddressInPkScript = errors.New("pkScript has no address")
 )
 
 // NewDerivedAddress creates one derived address through the legacy address-
@@ -863,4 +867,25 @@ func addressFromScript(pkScript []byte,
 	}
 
 	return addrs[0]
+}
+
+// addressFromPkScript extracts the first standard address from one script.
+//
+// This lets the legacy address manager resolve wallet metadata from a
+// script-pubkey-only store lookup.
+func addressFromPkScript(pkScript []byte,
+	chainParams *chaincfg.Params) (btcutil.Address, error) {
+
+	_, addrs, _, err := txscript.ExtractPkScriptAddrs(pkScript, chainParams)
+	if err != nil {
+		return nil, fmt.Errorf("extract address from pkScript: %w", err)
+	}
+
+	if len(addrs) == 0 {
+		return nil, fmt.Errorf(
+			"extract address from pkScript: %w", errNoAddressInPkScript,
+		)
+	}
+
+	return addrs[0], nil
 }
--- a/wallet/internal/db/kvdb/utxostore.go
+++ b/wallet/internal/db/kvdb/utxostore.go
@@ -18,6 +18,9 @@ import (
 	"github.com/btcsuite/btcwallet/wtxmgr"
 )
 
+// A compile-time assertion to ensure Store implements the UTXO store.
+var _ db.UTXOStore = (*Store)(nil)
+
 var (
 	// errNotImplemented is returned for unimplemented kvdb store methods.
 	errNotImplemented = errors.New("not implemented")
@@ -38,15 +41,117 @@ var (
 	wtxmgrNamespaceKey = []byte("wtxmgr")
 )
 
+// notImplemented returns a consistent error for kvdb methods that still need a
+// legacy-backed implementation.
 func notImplemented(_ context.Context, method string) error {
 	return fmt.Errorf("kvdb.Store.%s: %w", method, errNotImplemented)
 }
 
-// GetUtxo is not yet implemented for kvdb.
-func (s *Store) GetUtxo(ctx context.Context,
-	_ db.GetUtxoQuery) (*db.UtxoInfo, error) {
+// GetUtxo retrieves one current wallet-owned UTXO through the legacy wtxmgr
+// query path.
+func (s *Store) GetUtxo(_ context.Context,
+	query db.GetUtxoQuery) (*db.UtxoInfo, error) {
 
-	return nil, notImplemented(ctx, "GetUtxo")
+	var utxo *db.UtxoInfo
+
+	err := walletdb.View(s.db, func(tx walletdb.ReadTx) error {
+		got, err := s.getUtxoInView(tx, query.OutPoint)
+		if err != nil {
+			return err
+		}
+
+		utxo = got
+
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("kvdb.Store.GetUtxo: %w", err)
+	}
+
+	return utxo, nil
+}
+
+// getUtxoInView resolves one current wallet-owned UTXO inside an open
+// walletdb read transaction. It gates the outpoint through the current
+// UTXO set first, then emits the enriched wallet-owned row.
+func (s *Store) getUtxoInView(tx walletdb.ReadTx,
+	outPoint wire.OutPoint) (*db.UtxoInfo, error) {
+
+	txmgrNs := tx.ReadBucket(wtxmgrNamespaceKey)
+	if txmgrNs == nil {
+		return nil, errMissingTxmgrNamespace
+	}
+
+	credit, leaseSet, err := s.loadCurrentCredit(txmgrNs, outPoint)
+	if err != nil {
+		return nil, err
+	}
+
+	return s.enrichOwnedCredit(tx, credit, leaseSet)
+}
+
+// loadCurrentCredit fetches one credit by outpoint and gates it through the
+// current UTXO set, returning db.ErrUtxoNotFound for unknown or non-current
+// (spent-by-unmined) outputs. The active lease set is returned alongside so
+// the caller can mark IsLocked without re-reading the lease bucket.
+func (s *Store) loadCurrentCredit(txmgrNs walletdb.ReadBucket,
+	outPoint wire.OutPoint) (*wtxmgr.Credit, map[wire.OutPoint]struct{},
+	error) {
+
+	credit, err := s.txStore.GetUtxo(txmgrNs, outPoint)
+	if err != nil {
+		if errors.Is(err, wtxmgr.ErrUtxoNotFound) {
+			return nil, nil, db.ErrUtxoNotFound
+		}
+
+		return nil, nil, fmt.Errorf("get utxo: %w", err)
+	}
+
+	current, err := s.currentUTXOSet(txmgrNs)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if _, ok := current[outPoint]; !ok {
+		return nil, nil, db.ErrUtxoNotFound
+	}
+
+	leaseSet, err := s.activeLeaseSet(txmgrNs, current)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	return credit, leaseSet, nil
+}
+
+// enrichOwnedCredit resolves the owning account for one current credit and
+// produces its enriched db.UtxoInfo. A credit whose script does not join to a
+// wallet address is reported as not found, matching the SQL backends which
+// inner-join UTXO rows against owned addresses rather than surfacing an
+// unenriched row.
+func (s *Store) enrichOwnedCredit(tx walletdb.ReadTx,
+	credit *wtxmgr.Credit,
+	leaseSet map[wire.OutPoint]struct{}) (*db.UtxoInfo, error) {
+
+	addrmgrNs := tx.ReadBucket(waddrmgr.NamespaceKey)
+	if addrmgrNs == nil {
+		return nil, errMissingAddrmgrNamespace
+	}
+
+	chainParams := s.addrStore.ChainParams()
+
+	binding, addr, err := s.resolveAccountForCredit(
+		addrmgrNs, credit, chainParams,
+	)
+	if err != nil {
+		if errors.Is(err, errUnownedScript) {
+			return nil, db.ErrUtxoNotFound
+		}
+
+		return nil, err
+	}
+
+	return s.enrichCredit(addrmgrNs, credit, addr, binding, leaseSet)
 }
 
 // ListUTXOs is not yet implemented for kvdb.
--- a/wallet/internal/db/kvdb/utxostore.go
+++ b/wallet/internal/db/kvdb/utxostore.go
@@ -155,11 +155,35 @@ func (s *Store) enrichOwnedCredit(tx walletdb.ReadTx,
 	return s.enrichCredit(addrmgrNs, credit, addr, binding, leaseSet)
 }
 
-// ListUTXOs is not yet implemented for kvdb.
-func (s *Store) ListUTXOs(ctx context.Context,
-	_ db.ListUtxosQuery) ([]db.UtxoInfo, error) {
+// ListUTXOs lists current wallet-owned UTXOs through the legacy wtxmgr query
+// path.
+func (s *Store) ListUTXOs(_ context.Context,
+	query db.ListUtxosQuery) ([]db.UtxoInfo, error) {
 
-	return nil, notImplemented(ctx, "ListUTXOs")
+	if s.addrStore == nil {
+		return nil, fmt.Errorf("kvdb.Store.ListUTXOs: %w",
+			errMissingAddrStore)
+	}
+
+	err := query.Validate()
+	if err != nil {
+		return nil, err
+	}
+
+	var utxos []db.UtxoInfo
+
+	err = walletdb.View(s.db, func(tx walletdb.ReadTx) error {
+		return s.listUTXOsInView(tx, query, &utxos)
+	})
+	if err != nil {
+		return nil, fmt.Errorf("kvdb.Store.ListUTXOs: %w", err)
+	}
+
+	if len(utxos) == 0 {
+		return []db.UtxoInfo{}, nil
+	}
+
+	return utxos, nil
 }
 
 // LeaseOutput locks one known wallet UTXO through the legacy wtxmgr lease
@@ -656,6 +680,117 @@ func (s *Store) enrichCredit(addrmgrNs walletdb.ReadBucket,
 	}
 
 	return info, nil
+}
+
+// listUTXOsInView performs the legacy UTXO scan using one walletdb view
+// and populates the enriched per-row metadata (AccountName, Origin,
+// AddrType, HasScript, IsLocked).
+func (s *Store) listUTXOsInView(tx walletdb.ReadTx,
+	query db.ListUtxosQuery, utxos *[]db.UtxoInfo) error {
+
+	txmgrNs := tx.ReadBucket(wtxmgrNamespaceKey)
+	if txmgrNs == nil {
+		return errMissingTxmgrNamespace
+	}
+
+	addrmgrNs := tx.ReadBucket(waddrmgr.NamespaceKey)
+	if addrmgrNs == nil {
+		return errMissingAddrmgrNamespace
+	}
+
+	credits, err := s.txStore.UnspentOutputsIncludingLocked(txmgrNs)
+	if err != nil {
+		return fmt.Errorf("list unspent outputs: %w", err)
+	}
+
+	// The credits slice already enumerates the current UTXO set, so build
+	// the lease-intersection set directly from it rather than re-scanning
+	// the unspent bucket via currentUTXOSet.
+	current := make(map[wire.OutPoint]struct{}, len(credits))
+	for i := range credits {
+		current[credits[i].OutPoint] = struct{}{}
+	}
+
+	leaseSet, err := s.activeLeaseSet(txmgrNs, current)
+	if err != nil {
+		return err
+	}
+
+	currentHeight := s.addrStore.SyncedTo().Height
+	chainParams := s.addrStore.ChainParams()
+
+	for i := range credits {
+		enriched, ok, err := s.buildListedUTXO(
+			addrmgrNs, &credits[i], query, leaseSet, currentHeight,
+			chainParams,
+		)
+		if err != nil {
+			return err
+		}
+
+		if !ok {
+			continue
+		}
+
+		*utxos = append(*utxos, *enriched)
+	}
+
+	return nil
+}
+
+// buildListedUTXO converts one current credit into an enriched db.UtxoInfo,
+// applying the confirmation, owned-address, and account filters along the
+// way. It returns ok=false for a credit that does not pass a filter so the
+// caller can skip it without treating the omission as an error.
+func (s *Store) buildListedUTXO(addrmgrNs walletdb.ReadBucket,
+	credit *wtxmgr.Credit, query db.ListUtxosQuery,
+	leaseSet map[wire.OutPoint]struct{}, currentHeight int32,
+	chainParams *chaincfg.Params) (*db.UtxoInfo, bool, error) {
+
+	if !utxoMatchesConfirmations(credit.Height, currentHeight, query) {
+		return nil, false, nil
+	}
+
+	binding, addr, err := s.resolveAccountForCredit(
+		addrmgrNs, credit, chainParams,
+	)
+	if err != nil {
+		if !errors.Is(err, errUnownedScript) {
+			return nil, false, err
+		}
+
+		// A credit whose script does not join to a wallet address is
+		// dropped, matching the SQL backends which inner-join UTXO rows
+		// against owned addresses. Such rows never surface (even with no
+		// account filter).
+		return nil, false, nil
+	}
+
+	accountOK, err := accountMatchesQuery(addrmgrNs, query, binding)
+	if err != nil {
+		return nil, false, err
+	}
+
+	if !accountOK {
+		return nil, false, nil
+	}
+
+	enriched, err := s.enrichCredit(
+		addrmgrNs, credit, addr, binding, leaseSet,
+	)
+	if err != nil {
+		return nil, false, err
+	}
+
+	// The AccountName filter is applied after enrichment because resolving
+	// the name requires the account-properties read that enrichCredit does.
+	if query.AccountName != nil &&
+		enriched.AccountName != *query.AccountName {
+
+		return nil, false, nil
+	}
+
+	return enriched, true, nil
 }
 
 // utxoMatchesConfirmations applies the optional db.ListUtxosQuery confirmation
--- a/wallet/internal/db/kvdb/utxostore.go
+++ b/wallet/internal/db/kvdb/utxostore.go
@@ -424,8 +424,8 @@ func (s *Store) walkBalanceUTXOs(addrmgrNs, txmgrNs walletdb.ReadBucket,
 		return 0, 0, nil
 	}
 
-	nameFound, err := s.resolveBalanceNameFilter(
-		addrmgrNs, &params, scopeFilter,
+	nameAcct, nameFound, err := s.resolveBalanceNameFilter(
+		addrmgrNs, params, scopeFilter,
 	)
 	if err != nil {
 		return 0, 0, err
@@ -447,31 +447,51 @@ func (s *Store) walkBalanceUTXOs(addrmgrNs, txmgrNs walletdb.ReadBucket,
 	for i := range unspent {
 		output := &unspent[i]
 
-		owner, account, ok := s.classifyBalanceUTXO(
-			addrmgrNs, output, chainParams,
+		accepted, err := s.balanceOutputAccepted(
+			addrmgrNs, output, scopeFilter, params, nameAcct,
+			syncBlock.Height, chainParams,
 		)
-		if !ok {
-			continue
+		if err != nil {
+			return 0, 0, err
 		}
 
-		if !balanceFilterAllows(
-			scopeFilter, owner, params, account,
-		) {
-
-			continue
+		if accepted {
+			total, locked = addBalanceOutput(total, locked, output)
 		}
-
-		if !confsAllowed(
-			output, syncBlock.Height, params,
-		) {
-
-			continue
-		}
-
-		total, locked = addBalanceOutput(total, locked, output)
 	}
 
 	return total, locked, nil
+}
+
+// balanceOutputAccepted reports whether one unspent output counts toward the
+// balance under the BalanceParams filters. It resolves the owning account,
+// applies the Scope/account filters (via balanceFilterAllows) and the
+// confirmation filters (via confsAllowed), and returns false for outputs that
+// are unowned or filtered out so the caller skips them without error.
+func (s *Store) balanceOutputAccepted(addrmgrNs walletdb.ReadBucket,
+	output *wtxmgr.Credit, scopeFilter *waddrmgr.KeyScope,
+	params db.BalanceParams, nameAcct *uint32, syncHeight int32,
+	chainParams *chaincfg.Params) (bool, error) {
+
+	owner, account, ok := s.classifyBalanceUTXO(
+		addrmgrNs, output, chainParams,
+	)
+	if !ok {
+		return false, nil
+	}
+
+	allowed, err := balanceFilterAllows(
+		addrmgrNs, scopeFilter, owner, params, nameAcct, account,
+	)
+	if err != nil {
+		return false, err
+	}
+
+	if !allowed {
+		return false, nil
+	}
+
+	return confsAllowed(output, syncHeight, params), nil
 }
 
 // addBalanceOutput adds an accepted output to the total balance and, when the
@@ -488,34 +508,40 @@ func addBalanceOutput(total, locked btcutil.Amount,
 }
 
 // resolveBalanceNameFilter resolves BalanceParams.Name into the real waddrmgr
-// account number before the UTXO walk. A missing name means no UTXOs match,
-// so the caller should return a zero balance without an error.
+// account number used for the UTXO walk. It returns (nil, true, nil) when no
+// Name filter is set, the resolved account number when one is, and
+// (nil, false, nil) when the name does not exist so the caller returns a zero
+// balance without an error.
+//
+// The resolved number is returned separately rather than folded into
+// params.Account: a Name filter is the public handle that selects an account
+// by name regardless of origin (including imported accounts), whereas a
+// numeric params.Account filter must never match imported accounts. Keeping
+// the two distinct lets balanceFilterAllows apply the imported exclusion only
+// to the numeric path.
 func (s *Store) resolveBalanceNameFilter(addrmgrNs walletdb.ReadBucket,
-	params *db.BalanceParams, scopeFilter *waddrmgr.KeyScope) (bool, error) {
+	params db.BalanceParams,
+	scopeFilter *waddrmgr.KeyScope) (*uint32, bool, error) {
 
 	if params.Name == nil {
-		return true, nil
+		return nil, true, nil
 	}
 
 	scopedMgr, err := s.addrStore.FetchScopedKeyManager(*scopeFilter)
 	if err != nil {
-		return false, fmt.Errorf("fetch scoped manager: %w", err)
+		return nil, false, fmt.Errorf("fetch scoped manager: %w", err)
 	}
 
 	acctNum, err := scopedMgr.LookupAccount(addrmgrNs, *params.Name)
 	if err != nil {
 		if waddrmgr.IsError(err, waddrmgr.ErrAccountNotFound) {
-			return false, nil
+			return nil, false, nil
 		}
 
-		return false, fmt.Errorf("lookup account: %w", err)
+		return nil, false, fmt.Errorf("lookup account: %w", err)
 	}
 
-	// Internal filtering stays on the real account number; the public
-	// AccountInfo.AccountNumber masking only applies at the read API boundary.
-	params.Account = &acctNum
-
-	return true, nil
+	return &acctNum, true, nil
 }
 
 // classifyBalanceUTXO maps an unspent output to its owning address and
@@ -541,21 +567,39 @@ func (s *Store) classifyBalanceUTXO(addrmgrNs walletdb.ReadBucket,
 	return owner, account, true
 }
 
-// balanceFilterAllows reports whether an output's owning address and
-// account survive the Scope and Account filters from BalanceParams.
-func balanceFilterAllows(scopeFilter *waddrmgr.KeyScope,
-	owner waddrmgr.AccountStore, params db.BalanceParams,
-	account uint32) bool {
+// balanceFilterAllows reports whether an output's owning address and account
+// survive the Scope and account filters from BalanceParams. nameAcct is the
+// account number resolved from a Name filter, if any; it selects that account
+// by exact number regardless of origin. A numeric params.Account filter, by
+// contrast, never matches an imported account (see numericAccountMatches).
+// Account and Name are mutually exclusive, so at most one of the two account
+// filters applies.
+func balanceFilterAllows(addrmgrNs walletdb.ReadBucket,
+	scopeFilter *waddrmgr.KeyScope, owner waddrmgr.AccountStore,
+	params db.BalanceParams, nameAcct *uint32, account uint32) (bool, error) {
 
 	if scopeFilter != nil && *scopeFilter != owner.Scope() {
-		return false
+		return false, nil
 	}
 
-	if params.Account != nil && *params.Account != account {
-		return false
+	// A Name filter resolves to a concrete account number and matches it
+	// exactly; imported accounts are reachable this way by design.
+	if nameAcct != nil {
+		return account == *nameAcct, nil
 	}
 
-	return true
+	if params.Account == nil {
+		return true, nil
+	}
+
+	imported, err := accountIsImported(addrmgrNs, owner, account)
+	if err != nil {
+		return false, err
+	}
+
+	return numericAccountMatches(
+		account, *params.Account, imported,
+	), nil
 }
 
 // confsAllowed reports whether an unspent output's confirmation depth
--- a/wallet/internal/db/kvdb/utxostore.go
+++ b/wallet/internal/db/kvdb/utxostore.go
@@ -283,11 +283,52 @@ func (s *Store) ReleaseOutput(_ context.Context,
 	return nil
 }
 
-// ListLeasedOutputs is not yet implemented for kvdb.
-func (s *Store) ListLeasedOutputs(ctx context.Context,
+// ListLeasedOutputs lists the currently active legacy output leases.
+func (s *Store) ListLeasedOutputs(_ context.Context,
 	_ uint32) ([]db.LeasedOutput, error) {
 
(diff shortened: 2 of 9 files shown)


wallet/internal/db/kvdb/utxostore.go
```diff
@@ -34,25 +46,114 @@ func IsNotImplemented(err error) bool {
...
+
+	err := walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
+		ns := tx.ReadWriteBucket(wtxmgrNamespaceKey)
+		if ns == nil {
+			return errMissingTxmgrNamespace
+		}
+
+		var err error
+
+		expiration, err = s.txStore.LockOutput(
```
yyforyongyu, line 217: `wtxmgr.LockOutput` only checks that the outpoint is known, not that it is still unspent with respect to `unminedInputs`. A pending spend can therefore be leased successfully here, whereas the SQL acquisition path returns no row or `ErrUtxoNotFound` for the same state; callers then believe they reserved a coin that is already spent by the wallet. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -34,25 +46,114 @@ func IsNotImplemented(err error) bool {
...
+
+	var utxo *db.UtxoInfo
+
+	err := walletdb.View(s.db, func(tx walletdb.ReadTx) error {
+		ns := tx.ReadBucket(wtxmgrNamespaceKey)
+		if ns == nil {
+			return errMissingTxmgrNamespace
+		}
+
+		credit, err := s.txStore.GetUtxo(ns, query.OutPoint)
```
yyforyongyu: Because `wtxmgr.GetUtxo` does not check the legacy `unminedInputs` bucket, this can return a mined credit after the wallet has already created a pending spend for it. The SQL implementation filters `spent_by_tx_id IS NULL`, so through the shared `UTXOStore` contract the kvdb backend can expose a non-current coin as a valid UTXO and let callers reserve or spend it again. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -108,3 +241,132 @@ func (s *Store) Balance(ctx context.Context,
...
+	if txmgrNs == nil {
+		return errMissingTxmgrNamespace
+	}
+
+	addrmgrNs := tx.ReadBucket(waddrmgr.NamespaceKey)
+	if addrmgrNs == nil {
+		return errMissingAddrmgrNamespace
+	}
+
+	credits, err := s.txStore.UnspentOutputs(txmgrNs)
```
yyforyongyu: `UnspentOutputs` is the legacy spendable set and skips active locked outputs. `UTXOStore.ListUTXOs` and the SQL query intentionally return leased and unleased UTXOs alike so lease policy can be applied by the caller, so this under-reports the current UTXO set whenever any output is leased. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -95,11 +196,43 @@ func (s *Store) ReleaseOutput(_ context.Context,
...
-	return nil, notImplemented(ctx, "ListLeasedOutputs")
+	var leases []db.LeasedOutput
+
+	err := walletdb.View(s.db, func(tx walletdb.ReadTx) error {
+		ns := tx.ReadBucket(wtxmgrNamespaceKey)
+		if ns == nil {
+			return errMissingTxmgrNamespace
+		}
+
+		locked, err := s.txStore.ListLockedOutputs(ns)
```
yyforyongyu, line 330: `ListLockedOutputs` filters by expiration only and does not re-check that the locked outpoint is still a current UTXO. If a locked output is later spent by an unmined transaction, this method will still report an active lease, unlike the SQL path which joins back to unspent current UTXOs. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -330,3 +389,132 @@ func calcConfs(outputHeight, curHeight int32) int32 {
...
+	if err != nil {
+		return false, fmt.Errorf("lookup utxo account: %w", err)
+	}
+
+	return utxoAccount == account, nil
+}
+
+// calcConfirmations mirrors the wallet confirmation calculation used by the
+// public UTXO manager.
+func calcConfirmations(txHeight, currentHeight int32) int32 {
```
yyforyongyu: `calcConfirmations` and the existing `calcConfs` at line 385 in the same file compute the same value but diverge on the `txHeight == 0` edge case:

```go
// calcConfs (line 385)
if outputHeight <= 0 || outputHeight > curHeight { return 0 }
return curHeight - outputHeight + 1

// calcConfirmations (line 509)
case txHeight == -1: return 0
case txHeight > currentHeight: return 0
default: return currentHeight - txHeight + 1
```

For `txHeight == 0` (genesis credit, never produced in practice but still a real height), `calcConfs` returns 0 and `calcConfirmations` returns `currentHeight + 1`. Same input, opposite answer.

Pick one and have the other caller use it — most of the file already uses `calcConfs`, and the `<= 0` guard there matches the wtxmgr convention (`-1` = unmined, `0` not exercised in practice). Folding `shouldIncludeUtxo` / `utxoMatchesConfirmations` onto `calcConfs` keeps one source of truth.

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -330,3 +389,132 @@ func calcConfs(outputHeight, curHeight int32) int32 {
...
+	if txmgrNs == nil {
+		return errMissingTxmgrNamespace
+	}
+
+	addrmgrNs := tx.ReadBucket(waddrmgr.NamespaceKey)
+	if addrmgrNs == nil {
+		return errMissingAddrmgrNamespace
+	}
+
+	credits, err := s.txStore.UnspentOutputs(txmgrNs)
```
yyforyongyu: Using `UnspentOutputs` drops locked credits before the wallet can mark them as `Locked`, so list through `UnspentOutputsIncludingLocked` and derive lock state from the lease set. (`openai/gpt-5.5`)

wallet/utxo_manager.go
```diff
@@ -205,189 +160,155 @@ func (w *Wallet) ListUnspent(_ context.Context,
...
+	if err != nil {
+		return false, "", 0, fmt.Errorf("get address: %w", err)
+	}
+
+	walletAddrType, err := walletAddressType(addrInfo.AddrType)
+	if err != nil {
+		return false, "", 0, err
+	}
+
+	return !addrInfo.IsWatchOnly, addrInfo.AccountName, walletAddrType, nil
```
yyforyongyu: This treats imported private-key addresses as spendable, but the legacy `AddressDetails` policy marks all imported addresses unspendable; preserve the imported-origin check when mapping store address metadata. (`openai/gpt-5.5`)

wallet/utxo_manager.go
```diff
@@ -168,35 +117,41 @@ func (w *Wallet) ListUnspent(_ context.Context,
...
+		MaxConfs: &maxConfs,
+	})
+	if err != nil {
+		return nil, fmt.Errorf("list utxos: %w", err)
+	}
 
-	err = walletdb.View(w.cfg.DB, func(tx walletdb.ReadTx) error {
-		addrmgrNs := tx.ReadBucket(waddrmgrNamespaceKey)
-		txmgrNs := tx.ReadBucket(wtxmgrNamespaceKey)
+	leases, err := w.store.ListLeasedOutputs(ctx, w.id)
```
yyforyongyu: This new store lease call makes kvdb `ListUnspent` fail because `kvdb.Store.ListLeasedOutputs` still returns `notImplemented`, and `GetUtxo` has the same dependency; implement the kvdb lease read or keep a legacy fallback until it exists. (`openai/gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -20,22 +20,61 @@ var (
...
 	return fmt.Errorf("kvdb.Store.%s: %w", method, errNotImplemented)
 }
 
-// GetUtxo is not yet implemented for kvdb.
-func (s *Store) GetUtxo(ctx context.Context,
-	_ db.GetUtxoQuery) (*db.UtxoInfo, error) {
+// IsNotImplemented reports whether err was produced by a kvdb Store method
+// stub that has not yet been implemented. Wallet-side fallback paths use
+// this predicate to detect when to fall back to legacy walletdb walks.
+func IsNotImplemented(err error) bool {
```
yyforyongyu: this seems to be rebase leftover - there is no need to export this! nor do we care about it - we will implement all methods and eventually delete this `notImplemented` error when the PRs are merged.

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -330,3 +460,132 @@ func calcConfs(outputHeight, curHeight int32) int32 {
...
+	if err != nil {
+		return false, fmt.Errorf("lookup utxo account: %w", err)
+	}
+
+	return utxoAccount == account, nil
+}
+
+// calcConfirmations mirrors the wallet confirmation calculation used by the
+// public UTXO manager.
+func calcConfirmations(txHeight, currentHeight int32) int32 {
```
yyforyongyu: `calcConfirmations` is functionally identical to the pre-existing `calcConfs` at line 456 in the same file. Remove the duplicate and use `calcConfs` in `utxoMatchesConfirmations` instead. (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪙
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -16,40 +17,137 @@ import (
...
+
+	var utxo *db.UtxoInfo
+
+	err := walletdb.View(s.db, func(tx walletdb.ReadTx) error {
+		ns := tx.ReadBucket(wtxmgrNamespaceKey)
+		if ns == nil {
+			return errMissingTxmgrNamespace
+		}
+
+		credit, err := s.txStore.GetUtxo(ns, query.OutPoint)
```
yyforyongyu: `wtxmgr.GetUtxo` can return a mined credit that is already spent by an unmined wallet transaction, so check the unmined spend state or use the same current-output path as `ListUTXOs` before returning it. (`gpt-5.5`)

yyforyongyu (commented):
LGTM 🐊 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪻
(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/utils_test.go
```diff
@@ -63,3 +71,238 @@ func newTxStore(t *testing.T, dbConn walletdb.DB) *wtxmgr.Store {
...
+
+	err := walletdb.Update(dbConn, func(tx walletdb.ReadWriteTx) error {
+		_, err := tx.CreateTopLevelBucket(waddrmgr.NamespaceKey)
+		return err
+	})
+	require.NoError(t, err)
+}
+
+// testAddrStore is a narrow legacy address-manager test double.
+type testAddrStore struct {
```
yyforyongyu: we have bwtest mocks now, why do we still implement this? we should remove it, and double check the following branches that bwtest is used instead

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -36,6 +36,10 @@ var (
 	errMissingAddrmgrNamespace = errors.New(
 		"kvdb: missing waddrmgr namespace",
 	)
+
```
yyforyongyu, line 39: this commit is too large, did you not perform the commit hygiene check???

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -330,3 +546,318 @@ func calcConfs(outputHeight, curHeight int32) int32 {
...
+			// historical test fixtures insert non-standard
+			// scripts to exercise the lease/locking paths. Skip
+			// the row when an account-related filter is set
+			// (we cannot prove the row matches), otherwise emit
+			// base-only info so the caller still sees the row.
+			if accountFilterActive(query) {
+				continue
+			}
+
+			info := utxoInfoBase(credit)
```
yyforyongyu: Returning base-only metadata for an `errUnownedScript` row leaves `Origin`, `AddrType`, and `AccountName` at zero values. The wallet layer then treats any standard script that no longer resolves in `waddrmgr` as a derived, spendable raw-pubkey UTXO; SQL drops rows that do not join to a wallet address. Skip unresolved standard-address credits (or return explicit unspendable metadata) here and in `GetUtxo` instead. (`gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Resolved — LeaseOutput now pre-checks the outpoint against the current UTXO set (currentUTXOSet) and returns ErrUtxoNotFound for a non-current / pending-spent outpoint, matching the SQL acquisition path.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Resolved — GetUtxo gates on the current UTXO set (currentUTXOSet membership) and returns ErrUtxoNotFound for outpoints that are no longer current, so a mined-but-unmined-spent credit isn't returned (matches SQL spent_by_tx_id IS NULL).

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Resolved — the read path uses UnspentOutputsIncludingLocked (not the legacy UnspentOutputs), so leased and unleased UTXOs are both returned with lock state derived from the lease set.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Resolved — ListLeasedOutputs intersects each lease with the current UTXO set, so a lease on an output later spent by an unmined tx is no longer reported.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Resolved — the duplicate is gone; the file uses a single calcConfs (the calcConfirmations variant was removed), so there is no txHeight==0 divergence.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Resolved — listing goes through UnspentOutputsIncludingLocked and derives lock state from the lease set, so locked credits aren't dropped before marking.

wallet/utxo_manager.go
yyforyongyu: Resolved — utxo_manager marks imported-account outputs unspendable (spendable := !IsWatchOnly() && Origin != ImportedAccount), preserving the legacy imported-origin policy.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Resolved — the duplicate calcConfirmations is removed; utxoMatchesConfirmations uses the single calcConfs.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Resolved — GetUtxo gates on the current UTXO set, so an unmined-spent credit returns ErrUtxoNotFound (same current-output path as ListUTXOs).

wallet/internal/db/utxo_store_common.go
```diff
@@ -54,11 +54,37 @@ func buildOutPoint(hash []byte, outputIndex uint32) (wire.OutPoint, error) {
...
+	// IsLocked is true when the row has an active (non-expired) lease.
+	IsLocked bool
+}
+
 // BuildUtxoInfo converts normalized SQL result fields into the public UtxoInfo
-// shape returned by the db interfaces.
+// shape returned by the db interfaces. The enrichment carries the per-row
+// fields the SQL backends populate alongside the base UTXO columns; the
+// kvdb backend still returns notImplemented for ListUtxos / GetUtxo on this
+// branch and therefore never calls this helper.
```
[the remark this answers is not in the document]
yyforyongyu: Done — the BuildUtxoInfo doc no longer claims kvdb returns notImplemented for ListUtxos/GetUtxo.

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -330,3 +537,310 @@ func calcConfs(outputHeight, curHeight int32) int32 {
...
+	}
+
+	credits, err := s.txStore.UnspentOutputsIncludingLocked(txmgrNs)
+	if err != nil {
+		return fmt.Errorf("list unspent outputs: %w", err)
+	}
+
+	current, err := s.currentUTXOSet(txmgrNs)
+	if err != nil {
+		return err
```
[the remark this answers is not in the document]
yyforyongyu: Done — listUTXOsInView builds the current set from the already-fetched credits slice instead of re-scanning UnspentOutputsIncludingLocked via currentUTXOSet.

wallet/internal/db/kvdb/utils_test.go
yyforyongyu: Done — the hand-rolled testManagedAddress is removed; kvdb tests use bwmock.ManagedAddress.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Done — unresolved errUnownedScript credits are now skipped (listUTXOsInView drops them; GetUtxo returns ErrUtxoNotFound), matching the SQL inner-join — no zero-value enrichment is surfaced.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -6,50 +6,222 @@ import (
...
+		}
+
+		chainParams := s.addrStore.ChainParams()
+
+		binding, addr, err := s.resolveAccountForCredit(
+			addrmgrNs, credit, chainParams,
+		)
+		if err != nil {
+			// A credit whose script does not join to a wallet
+			// address is reported as not found, matching the SQL
```
yyforyongyu: The code here (and the matching `listUTXOsInView` path below at ~780) now reports an unresolvable script as `ErrUtxoNotFound` / drops the row — correctly matching the SQL inner-join semantics. But this commit's message still describes the old behavior:

> Narrow kvdb tests that pre-date enrichment keep working via a base-info fallback when the addrStore is unavailable **or the script does not resolve to a wallet-owned account.**

The second clause is stale: the base-info fallback now fires only on the `s.addrStore == nil` branch. A script that doesn't resolve to a wallet-owned account is no longer given base info — `GetUtxo` returns `ErrUtxoNotFound` and `ListUTXOs` drops it. Please drop that clause from the message so it matches the code.

(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🦦 (`gpt-5.5`)

wallet/internal/db/kvdb/addressstore.go
yyforyongyu: Done — the oversized commit is split into method-scoped commits: `kvdb: implement GetUtxo` / `ListUTXOs` / `LeaseOutput` / `ListLeasedOutputs` (each with its own tests), and the wallet routing is split per method as well. Resolving.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Fixed — the commit message now matches behavior: unowned scripts map to `ErrUtxoNotFound` in `GetUtxo` and are dropped in `listUTXOsInView` (SQL inner-join parity); the base-info fallback applies only when `addrStore == nil`. Stale clause removed. Resolving.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🦪
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -330,3 +543,301 @@ func calcConfs(outputHeight, curHeight int32) int32 {
...
+// AccountName filter is applied separately after enrichment because
+// resolving the name requires an additional account-properties read.
+func accountMatchesQuery(query db.ListUtxosQuery,
+	binding *utxoAccountBinding) bool {
+
+	if query.Scope != nil && binding.scope != *query.Scope {
+		return false
+	}
+
+	if query.Account != nil && binding.acctNum != *query.Account {
```
yyforyongyu: `Account` is a store-level account number, and imported SQL accounts have no numeric account value, so numeric filters cannot select imported rows there. This direct comparison lets kvdb callers pass the legacy imported pseudo-account number and get imported UTXOs only on kvdb. Please sanitize/reject that pseudo-account here, or require `AccountName` for imported rows so the backends agree. (`gpt-5.5`)

yyforyongyu: Fixed — kvdb's numeric `Account` filter now rejects the legacy imported pseudo-account, so imported rows are selectable only via `(Scope, AccountName)`, matching SQL. Applied to both `ListUTXOs` and the separate `Balance` path, with tests. Pushed + CI-green. Resolving.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🦪
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -6,50 +6,222 @@ import (
...
+	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet/internal/addresstype"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/wtxmgr"
 )
 
+// A compile-time assertion to ensure Store implements the UTXO store.
+var _ db.UTXOStore = (*Store)(nil)
```
yyforyongyu, line 23: This makes `Store` satisfy the shared `UTXOStore` contract, but `ReleaseOutput` below still delegates straight to `wtxmgr.UnlockOutput`: wrong lock IDs wrap `wtxmgr.ErrOutputUnlockNotAllowed` instead of `db.ErrOutputUnlockNotAllowed`, and missing/non-current outputs do not map to `db.ErrUtxoNotFound` like the SQL backends. Translate the legacy release errors and gate the outpoint through the current UTXO set before returning. (`gpt-5.5`)

wallet/utxo_manager.go
```diff
@@ -278,116 +283,117 @@ func (w *Wallet) processUnspentOutput(addrmgrNs walletdb.ReadBucket,
...
+
+	addr := extractAddrFromPKScript(info.PkScript, w.cfg.ChainParams)
+	if addr == nil {
+		return nil, false, nil
+	}
 
-	err := w.state.validateStarted()
-	if err != nil {
-		return nil, err
+	if accountFilter != "" && info.AccountName != accountFilter {
```
yyforyongyu: i don't think `buildWalletUtxoFromStore` should return a bool? why do we need to check this case here?

wallet/utxo_manager.go
```diff
@@ -297,27 +297,33 @@ func (w *Wallet) GetUtxo(ctx context.Context,
...
+	if err != nil {
+		switch {
+		case errors.Is(err, db.ErrUtxoNotFound):
+			return time.Time{}, wtxmgr.ErrUnknownOutput
 
-		expiration, err = w.txStore.LockOutput(
-			txmgrNs, id, op, duration,
-		)
+		case errors.Is(err, db.ErrOutputAlreadyLeased):
+			return time.Time{}, wtxmgr.ErrOutputAlreadyLocked
```
yyforyongyu: what is the point of translating from db error to `wtxmgr` error? isn't this the opposite of hiding `wtxmgr`?

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -500,7 +500,10 @@ func balanceFilterAllows(scopeFilter *waddrmgr.KeyScope,
 		return false
 	}
 
-	if params.Account != nil && *params.Account != account {
+	if params.Account != nil && !numericAccountMatches(
```
yyforyongyu: this violates the commit hygiene?

yyforyongyu (commented):
we should fix `cyclop` instead of ignoring them

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🦪
(`claude-opus-4-8[1m]`)

wallet/utxo_manager.go
```diff
@@ -471,54 +374,31 @@ func (w *Wallet) ReleaseOutput(ctx context.Context, id wtxmgr.LockID,
 	return w.store.ReleaseOutput(ctx, params)
```
yyforyongyu: `GetUtxo` and `LeaseOutput` translate store-native sentinels back to the legacy `wtxmgr` errors because this public wallet API still exposes those errors to external callers. `ReleaseOutput` now returns `db.ErrUtxoNotFound`/`db.ErrOutputUnlockNotAllowed` directly, so callers checking `wtxmgr.ErrUnknownOutput` or `wtxmgr.ErrOutputUnlockNotAllowed` regress. Please translate the release errors here before returning. (`gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Done — kvdb `ReleaseOutput` now gates the outpoint through the current UTXO set and maps unlock failures to `db.ErrUtxoNotFound` / `db.ErrOutputUnlockNotAllowed`.

wallet/utxo_manager.go
yyforyongyu: Done — `buildWalletUtxoFromStore` is now a pure `(*Utxo, error)` mapper.

wallet/utxo_manager.go
yyforyongyu: Kept the `wtxmgr` translation as documented public-API compatibility for external callers, with a WHY comment; the Store layer stays db-native.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Done — folded the imported numeric-filter fix into the balances-routing commit; no trailing fixup remains.

wallet/utxo_manager.go
yyforyongyu: Done — `Wallet.ReleaseOutput` now translates `db.ErrUtxoNotFound` to `wtxmgr.ErrUnknownOutput` and `db.ErrOutputUnlockNotAllowed` to `wtxmgr.ErrOutputUnlockNotAllowed` at the wallet API boundary, matching `GetUtxo` / `LeaseOutput`.

yyforyongyu:
LGTM 🦪
(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🟩 (`gpt-5.5`)

wallet/internal/db/kvdb/utxostore_test.go
```diff
@@ -118,6 +83,131 @@ func TestReleaseOutputMissingNamespace(t *testing.T) {
...
+
+	txStore := newTxStore(t, dbConn)
+	store := NewStore(dbConn, txStore, nil)
+
+	// Arrange: insert a current credit and lease it under lock ID 1.
+	outPoint, _ := insertKnownCredit(
+		t, dbConn, txStore, []byte{0x51}, 2500, 2,
+	)
+
+	_, err := store.LeaseOutput(t.Context(), db.LeaseOutputParams{
```
yyforyongyu: this wrapping is wrong, should be
```
	_, err := store.LeaseOutput(
        t.Context(), db.LeaseOutputParams{
...
```

Fix all the wrappings

wallet/utxo_manager.go
```diff
@@ -278,116 +286,114 @@ func (w *Wallet) processUnspentOutput(addrmgrNs walletdb.ReadBucket,
...
+
+	currentHeight := w.addrStore.SyncedTo().Height
+
+	info, err := w.store.GetUtxo(ctx, db.GetUtxoQuery{
+		WalletID: w.id,
+		OutPoint: prevOut,
 	})
+	if err != nil {
+		if errors.Is(err, db.ErrUtxoNotFound) {
+			return nil, wtxmgr.ErrUtxoNotFound
```
yyforyongyu: why this conversion? we should just return the `err` instead? and `db.ErrUtxoNotFound` should be used here?

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -257,11 +257,17 @@ func (s *Store) LeaseOutput(_ context.Context,
 	}, nil
 }
 
-// ReleaseOutput releases a previously leased output.
```
yyforyongyu, line 55: `ReleaseOutput` should also be in its own commit?

wallet/internal/db/kvdb/utxostore_test.go
yyforyongyu: Fixed the 26 wrappings owned by this PR (#1238), folded into their introducing commits. The remaining same-line Store calls (`utxostore_test.go` / `utxostore_balance_test.go` / the `tx_utxo_store` itest) live on the merged trunk `sql-wallet` (introduced by `1c01dabe37 kvdb: implement Balance` and `932b59576 wallet: add CreateTx`), below this branch base, so they are out of scope for this PR and left as-is.

wallet/utxo_manager.go
yyforyongyu: Reworked per the maintainer direction: the wallet UTXO methods no longer translate to `wtxmgr` here. The Store-routing commits return the db sentinels directly, and a single final commit defines new wallet-owned sentinels (`wallet.ErrUnknownOutput` / `ErrOutputAlreadyLocked` / `ErrOutputUnlockNotAllowed`) and translates `db.Err* -> wallet.Err*` at the wallet API boundary. The standalone wtxmgr-translation commit was dropped.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: kvdb `ReleaseOutput` is isolated in its own commit (`kvdb: align ReleaseOutput error mapping with db sentinels`). Separately, the wallet-level error-translation hygiene is now fixed: the Store-routing commits are db-native and a single final commit owns the db-to-wallet sentinel translation.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🦪
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -330,3 +611,341 @@ func calcConfs(outputHeight, curHeight int32) int32 {
...
+//
+// Imported rows carry the legacy ImportedAddrAccount pseudo-account
+// number, which has no numeric counterpart on the SQL backends (their
+// account_number column is NULL for imported accounts). To keep the
+// numeric Account filter backend-independent, an imported row
+// therefore never matches a numeric filter on either backend;
+// imported rows are selectable only via (Scope, AccountName). A
+// derived row matches iff its account number equals the filter.
+func numericAccountMatches(ownerAcct, wantAcct uint32) bool {
+	if ownerAcct == waddrmgr.ImportedAddrAccount {
```
yyforyongyu: This only rejects the legacy imported-address pseudo-account. Imported xpub accounts created through `NewAccountWatchingOnly` get ordinary account numbers, but SQL imported accounts have `account_number = NULL` and cannot match a numeric `Account` filter; this still lets kvdb return imported-account UTXOs via numeric `ListUTXOs`/`Balance` filters. Include the imported/origin bit in the predicate and only let imported rows through the name-based filters. (`gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: The pseudo-account fix leaves the same backend-parity issue open for imported xpub accounts with ordinary kvdb account numbers; I posted the current-line finding here: https://github.com/btcsuite/btcwallet/pull/1238#discussion_r3354456807 (`gpt-5.5`)

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Fixed by the same accountIsImported/numericAccountMatches path. Imported xpub accounts with ordinary kvdb account numbers no longer satisfy numeric Account filters; they only match the name-based filter shape that SQL can represent.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Fixed in the current branch. accountMatchesQuery now classifies the owning account with AccountStore.IsImportedAccount, and numericAccountMatches rejects numeric Account filters for any imported row, including the legacy pseudo-account and imported xpub/watch-only accounts. Imported rows are selectable via Scope plus AccountName only, matching SQL.

yyforyongyu:
LGTM 🦪
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -10,32 +10,162 @@ import (
...
+	credit, leaseSet, err := s.loadCurrentCredit(txmgrNs, outPoint)
+	if err != nil {
+		return nil, err
+	}
+
+	// Production paths always carry an addrStore so the row is enriched.
+	// Narrow tests construct stores without one to exercise the legacy
+	// lock/spend gating; emit base info in that case so they continue to
+	// assert on the gating logic.
+	if s.addrStore == nil {
```
yyforyongyu: This path is wrong - classical "put the cart before the horse" - if the production code never hits the `nil` case, why do we even create this branch and test it??? just to satisfy a test case that never exists in real world?

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -330,3 +460,178 @@ func calcConfs(outputHeight, curHeight int32) int32 {
...
+// the only step that touches per-UTXO state outside the binding so the
+// caller can compute the set once for the whole listing.
+func (s *Store) enrichCredit(addrmgrNs walletdb.ReadBucket,
+	credit *wtxmgr.Credit, addr btcutil.Address,
+	binding *utxoAccountBinding,
+	leaseSet map[wire.OutPoint]struct{}) (*db.UtxoInfo, error) {
+
+	info := utxoInfoBase(credit)
+
+	if binding.acctStore == nil {
```
yyforyongyu: same here - the `accStore` is never nil in production, so no need to make up this fake case!

wallet/utxo_manager.go
```diff
@@ -176,35 +113,39 @@ func (w *Wallet) ListUnspent(_ context.Context,
...
 	log.Debugf("ListUnspent using query: %v", query)
 
-	syncBlock := w.addrStore.SyncedTo()
-	currentHeight := syncBlock.Height
+	currentHeight := w.addrStore.SyncedTo().Height
+	minConfs := query.MinConfs
+	maxConfs := query.MaxConfs
 
-	var utxos []*Utxo
+	infos, err := w.store.ListUTXOs(ctx, db.ListUtxosQuery{
```
yyforyongyu: `ctx, db.ListUtxosQuery{` should be put in a new line

wallet/utxo_manager.go
```diff
@@ -263,64 +263,36 @@ func (w *Wallet) GetUtxo(ctx context.Context,
...
-
-	err = walletdb.Update(w.cfg.DB, func(tx walletdb.ReadWriteTx) error {
-		txmgrNs := tx.ReadWriteBucket(wtxmgrNamespaceKey)
-
-		expiration, err = w.txStore.LockOutput(
-			txmgrNs, id, op, duration,
-		)
-
-		return err
+	lease, err := w.store.LeaseOutput(ctx, db.LeaseOutputParams{
```
yyforyongyu: `ctx, db.LeaseOutputParams{` should be in a new line

wallet/internal/db/kvdb/utxostore_balance_test.go
```diff
@@ -518,3 +518,112 @@ func creditImportedKeyAtHeight(t *testing.T, dbConn walletdb.DB,
...
+		balanceTestTipHeight, false,
+	)
+	creditAccountAddressAtHeight(
+		t, store.db, mgr, txStore, scope, importedAcct,
+		btcutil.Amount(300), balanceTestTipHeight, false,
+	)
+
+	// A numeric Account filter on the imported account's ordinary number
+	// must sum nothing: imported accounts are not numerically addressable.
+	res, err := store.Balance(t.Context(), db.BalanceParams{
```
yyforyongyu: `t.Context(), db.BalanceParams{` should be in a new line

wallet/internal/db/kvdb/utxostore_balance_test.go
```diff
@@ -518,3 +518,112 @@ func creditImportedKeyAtHeight(t *testing.T, dbConn walletdb.DB,
...
+		Scope:   &dbScope,
+		Account: &importedAcct,
+	})
+	require.NoError(t, err)
+	require.Zero(t, res.Total)
+
+	// The derived account is still summed by its number, with no leakage
+	// from the imported account.
+	derivedAcct := uint32(0)
+	res, err = store.Balance(t.Context(), db.BalanceParams{
```
yyforyongyu: `t.Context(), db.BalanceParams{` should be in a new line

wallet/internal/db/kvdb/utxostore_balance_test.go
```diff
@@ -518,3 +518,112 @@ func creditImportedKeyAtHeight(t *testing.T, dbConn walletdb.DB,
...
+		Scope:   &dbScope,
+		Account: &derivedAcct,
+	})
+	require.NoError(t, err)
+	require.Equal(t, btcutil.Amount(100), res.Total)
+
+	// The imported account remains summable through its (Scope, Name).
+	res, err = store.Balance(t.Context(), db.BalanceParams{
```
yyforyongyu: `t.Context(), db.BalanceParams{` should be in a new line

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Removed — the Store/addrStore is always present in production, so this impossible branch (and its test) is gone.

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Removed — the Store/addrStore is always present in production, so this impossible branch (and its test) is gone.

wallet/utxo_manager.go
yyforyongyu: Done — moved onto its own line.

wallet/utxo_manager.go
yyforyongyu: Done — moved onto its own line.

wallet/internal/db/kvdb/utxostore_balance_test.go
yyforyongyu: Done — moved onto its own line.

wallet/internal/db/kvdb/utxostore_balance_test.go
yyforyongyu: Done — moved onto its own line.

wallet/internal/db/kvdb/utxostore_balance_test.go
yyforyongyu: Done — moved onto its own line.

yyforyongyu:
Re-reviewed 63fb6f7; LGTM, no correctness issues found in the latest revision. (`gpt-5.5`)

yyforyongyu:
LGTM 🦪
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -6,57 +6,254 @@ import (
...
+
+// LeaseOutput locks one known wallet UTXO through the legacy wtxmgr lease
+// path.
+func (s *Store) LeaseOutput(_ context.Context,
+	params db.LeaseOutputParams) (*db.LeasedOutput, error) {
+
+	if params.Duration <= 0 {
+		return nil, fmt.Errorf(
+			"%w: lease duration must be positive", db.ErrInvalidParam,
+		)
```
GustavoStingelin, line 197: nit
```suggestion
		return nil, fmt.Errorf("%w: lease duration must be positive",
			db.ErrInvalidParam)
```

wallet/internal/db/kvdb/utxostore.go
```diff
@@ -6,57 +6,254 @@ import (
...
+
+		return nil
+	})
+	if err != nil {
+		return nil, fmt.Errorf("kvdb.Store.LeaseOutput: %w", err)
+	}
+
+	return &db.LeasedOutput{
+		OutPoint:   params.OutPoint,
+		LockID:     db.LockID(params.ID),
```
GustavoStingelin, line 241: In some places we call it “lock”, while in others we call it “lease”. I think we should converge on one consistent term later.

GustavoStingelin (approved):
LGTM!

wallet/internal/db/kvdb/utxostore.go
yyforyongyu: Gonna postpone this as a cleanup task, got tired of rebasing🤦🏻

Issue #38, opened 2014-01-22
rpc: implement backupwallet


dajohi:
implement backupwallet

jrick:
The seed should be backed up when the wallet is created, and is used to recreate a new wallet with the same keys.  Since the seed is not saved, it cannot be exported at a later time.

arnuschky:
I would like to reopen this. The `backupwallet` RPC call on bitcoin core writes the wallet in a consistent state so that it can be backed up and be used on a future restart.

This is very useful beyond backing up HD keys:
 * Single non-HD keys that can't be restored with a seed (eg, imported ones).
 * When restoring an HD wallet from the seed, I might not know which wallet chains and addresses have been created and used previously. Searching through this history is very quickly extremely costly (as each new address has to be rescanned).
 * Rescans can take a lot of time.
 * Any other metadata in the db is lost.

Many of these things are not super critical for the average home user, but remember that down-times over a few seconds/minutes are not acceptable in a professional production environment.

For these reasons I think that a functionality as provided by `backupwallet` is very much needed.

jrick:
For us, this is as easy as copying the `wallet.db` file while wallet is closed.  Bitcoin Core Wallet needs an RPC for this since it's not as simple as copying the wallet file since it references the blockchain data as well.

No need to go over RPC to export everything in the wallet.

arnuschky:
So how do I copy this file while running? Do you mean that I have to shut down btcwallet every time I backup (thereby losing all rpc client connections etc)?

jrick:
Copying a file isn't an atomic operation on any filesystem that I know of.  While the entire original file should always be in some kind of consistent state, the copy might not be.

davecgh:
I'd suggest having the RPC work similar to how the existing export watching wallet RPC function.   That is to say, have it create a read transaction and use the .WriteTo to write the full contents to a separate file.  That is safe and will provide a safe method for requesting a backup to be made while the wallet process is operational.

jrick:
I still believe that recovering the wallet via the seed is the best method of recovering a wallet.  I'm not a fan of using an RPC to write local files which the client then has to fetch off of the server's filesystem.

PR #1240, opened 2026-05-20
prep: foundation for the account-store routing (db.AccountInfo surface, fingerprint cache, Name filter)

yyforyongyu:
## Summary

Foundation pieces for the upcoming `db.AccountStore` routing work
(stacked PRs #1236 and #1239). Splits 7 prep commits out of the
previously-shared PR #1236 to keep that PR focused on the kvdb
implementations + the wallet-side cache routing.

## Commits

1. `waddrmgr: expose legacy adapter methods on public interfaces` —
   promote `EncryptedMasterHDPriv`, `Decrypt`, and `ImportWitnessScript`
   onto `waddrmgr.AddrStore` / `waddrmgr.AccountStore` so the kvdb
   adapter (downstream) can type its `addrStore` field as
   `waddrmgr.AddrStore` directly without per-call narrow interfaces
   and runtime assertions.
2. `wallet: surface db.AccountInfo on read APIs` — change
   `Wallet.GetAccount` to return `*db.AccountInfo` and
   `Wallet.ListAccounts*` to return `*AccountsResult` with
   `Accounts []db.AccountInfo`. Internally still uses
   `waddrmgr.AccountProperties + fetchAccountBalances`; a new
   `propertiesToAccountInfo` helper bridges the legacy shape into the
   canonical `db.AccountInfo` at the public boundary. Subsequent
   routing PRs do not need a reverse conversion.
3. `wtxmgr: add UnspentOutputsIncludingLocked` — variant of
   `UnspentOutputs` that includes locked outputs. Required by the
   downstream kvdb Balance walk in PR #1239.
4. `wallet: wire keyVault into the test fixture` — surfaces the
   keyVault stub on the mock-wallet fixture so the downstream
   routing tests can mock it.
5. `wallet: cache master fingerprint on Wallet at Manager.Load` —
   parse and cache the BIP32 master-key fingerprint at
   `Manager.Load` time. The kvdb store can't round-trip derived-
   account fingerprints natively (waddrmgr's default-account row
   has no fingerprint column); the wallet-layer override at the
   public boundary is the canonical compatibility path. Bug #38
   in our fork tracks the latent classifier issue this introduces.
6. **`db: add Name field to BalanceParams`** — pure contract
   addition: `Name *string` on `db.BalanceParams`, plus two new
   sentinels (`ErrBalanceParamsNameWithoutScope`,
   `ErrBalanceParamsAccountAndName`) and the matching Validate
   updates. No backend or caller uses Name yet; this is forward-
   compatible.
7. **`db+sql: thread Name through Balance backend`** — wire the
   pg and sqlite backends to honor the new Name filter. SQL queries
   gain an optional `account_name` filter alongside the existing
   `account_number` filter; sqlc regenerates; adapters thread
   `params.Name` through.

   This fixes the SQL silent-zero bug for imported accounts at the
   store layer: before this change, querying Balance for an
   imported account passed Account=0 (the masked value) which
   matched the default derived account (`account_number = 0`), not
   the imported row (`account_number IS NULL`). Imported-account
   balance reads silently returned 0. With Name, callers
   (including the still-to-route `Wallet.Balance` in #1239) get
   the correct sum.

## Why these are split out

Originally the wtxmgr/wallet/waddrmgr foundation pieces were
landing alongside the kvdb implementation in PR #1236. Splitting
them into this prep branch lets reviewers:

- See the contract additions and surface changes independently of
  the kvdb routing work.
- Land the SQL silent-zero fix immediately (commit 7), without
  waiting on the full account-store routing stack to merge.
- Keep PR #1236 focused on the 7 kvdb+routing commits.

## Stack

```
sql-wallet
  ├─ prep-account-store  (this PR)
       ├─ impl-account-kvdb (PR #1236)
            ├─ impl-account-manager-store (PR #1239)
```

PR #1236 and #1239 will rebase onto this branch.

## Test plan

- [ ] `make unit` passes.
- [ ] `make itest-db db=sqlite` passes (the SQL Name filter fix is
      exercised by existing Balance itests after the contract
      ships).
- [ ] `make itest-db db=postgres` passes.
- [ ] Per-commit build verified across all 7 commits via
      `git rebase --exec 'GOWORK=off go build ./...'`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wallet/internal/db/pg/utxostore_balance.go
```diff
@@ -10,6 +11,15 @@ import (
 	"github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
 )
 
+// nullableStringNS converts a *string into a sql.NullString. nil maps to
+// {Valid: false}; non-nil dereferences and stores the value.
+func nullableStringNS(s *string) sql.NullString {
```
yyforyongyu: `nullableStringNS` is byte-identical here and in `wallet/internal/db/sqlite/utxostore_balance.go:63` (same doc-comment, same body, just different file). The sibling helpers — `NullableInt32ToSQLInt32`, `NullableInt32ToSQLInt64`, `NullableUint32ToSQLInt64` — already live in the shared package at `wallet/internal/db/safecasting.go`, and both pg/sqlite balance adapters already call them from there in the same struct literal. Hoisting this one into the same file as `db.NullableStringToSQLNullString` (or whatever name fits the existing naming convention) and dropping the local copies removes the duplication and keeps the next nullable-helper from forking into per-backend files.

(`claude-opus-4-7[1m]`)

wallet/internal/db/data_types.go
```diff
@@ -1257,12 +1266,35 @@ var ErrBalanceParamsAccountWithoutScope = errors.New(
...
+// a scope; a name-only balance query would sum across scopes (or depend
+// on backend-specific lookup behavior).
+var ErrBalanceParamsNameWithoutScope = errors.New(
+	"balance: Name requires Scope (account names are scope-unique)",
+)
+
+// ErrBalanceParamsAccountAndName is returned by BalanceParams.Validate when
+// both Account and Name are set. The two fields are mutually exclusive:
+// picking one disambiguates the account, picking both is contradictory.
+var ErrBalanceParamsAccountAndName = errors.New(
```
yyforyongyu, line 1280: Two coverage gaps around the new `Name` contract.

**(a) Validate is untested for the new sentinels.** `TestBalanceParamsValidate` in `wallet/internal/db/data_types_test.go:39` covers `zero value`, `scope only`, `account with scope`, `account without scope`, and `account without scope but with confs`, but adds no cases for the two new sentinels this commit defines. A revert of either `if p.Name != nil && p.Scope == nil` or `if p.Account != nil && p.Name != nil` would land green. Three lean cases close it:

```go
{name: "name with scope", params: BalanceParams{WalletID: 1, Scope: &scope, Name: &name}},
{name: "name without scope", params: BalanceParams{WalletID: 1, Name: &name}, wantErr: ErrBalanceParamsNameWithoutScope},
{name: "account and name", params: BalanceParams{WalletID: 1, Scope: &scope, Account: &zeroAccount, Name: &name}, wantErr: ErrBalanceParamsAccountAndName},
```

**(b) The SQL silent-zero fix has no itest.** The follow-up `sql: thread Name through Balance backend` commit message claims the user-visible bug it closes is "imported-account balance reads silently returned 0". Nothing in `wallet/internal/db/itest/` exercises `Balance` with a `Name` filter against an imported account today — `grep -rn 'Name:' wallet/internal/db/itest/` only turns up name-keyed account/wallet *create/get* paths, no balance queries. An itest with the shape:

1. Create an imported account (`AccountNumber` masked to 0 on the read).
2. Credit a confirmed UTXO to that account.
3. Also create the default derived account (also `AccountNumber == 0` at the contract boundary) and credit a *different* UTXO to it.
4. `store.Balance(BalanceParams{Scope, Name: &importedName})` returns only the imported credit.
5. `store.Balance(BalanceParams{Scope, Name: &"default"})` returns only the derived credit.

…would lock in the fix and catch a regression where someone drops the `account_name` filter from the query.

(`claude-opus-4-7[1m]`)

wallet/account_manager.go
```diff
@@ -147,28 +148,63 @@ func (w *Wallet) NewAccount(_ context.Context, scope waddrmgr.KeyScope,
...
+		AccountName:          props.AccountName,
+		Origin:               origin,
+		ExternalKeyCount:     props.ExternalKeyCount,
+		InternalKeyCount:     props.InternalKeyCount,
+		ImportedKeyCount:     props.ImportedKeyCount,
+		IsWatchOnly:          props.IsWatchOnly,
+		KeyScope:             db.KeyScope(props.KeyScope),
+		PublicKey:            pubKey,
+		MasterKeyFingerprint: props.MasterKeyFingerprint,
+		ConfirmedBalance:     total,
```
yyforyongyu, line 197: Full-stack update: this is addressed for the routed `GetAccount` / `ListAccounts*` APIs by the later kvdb store work, which splits confirmed/unconfirmed balances and uses the locked-inclusive UTXO source. The remaining concern only applies to the legacy `Wallet.Accounts` path that still goes through this temporary converter; if that path is intentionally carried as legacy behavior until a later PR removes or rewrites it, this should not block the stack. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -147,28 +148,63 @@ func (w *Wallet) NewAccount(_ context.Context, scope waddrmgr.KeyScope,
...
+func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
+	total btcutil.Amount) db.AccountInfo {
+
+	var pubKey []byte
+	if props.AccountPubKey != nil {
+		pubKey = []byte(props.AccountPubKey.String())
+	}
+
+	origin := db.DerivedAccount
+	if props.IsWatchOnly {
```
yyforyongyu: `IsWatchOnly` is lock-state dependent, not an account-origin classifier. For normal derived accounts in a locked wallet, `AccountProperties` sets `IsWatchOnly` because the private account key is not loaded, so this reports them as `ImportedAccount` and masks `AccountNumber` to `0`. Use a persisted imported-account signal instead of watch-only state. (`openai/gpt-5.5`)

yyforyongyu:
Both prior findings addressed in this push:

- `nullableStringNS` is now `db.NullableStringToSQLNullString` in `wallet/internal/db/safecasting.go:126`; both pg/sqlite adapters call it from there.
- `TestBalanceParamsValidate` gained `name with scope` (accept), `name without scope` (`ErrBalanceParamsNameWithoutScope`), and `account and name` (`ErrBalanceParamsAccountAndName`).
- `TestBalanceNameFilterDisambiguatesMaskedImportedAccount` in the itest creates one derived + one imported account (both at `AccountNumber == 0` at the contract boundary), credits a distinct UTXO to each, and asserts each `Balance({Scope, Name})` returns only its own credit — exactly the imported-silent-zero regression the SQL Name filter was meant to close.

LGTM 🧱
(`claude-opus-4-7[1m]`)

yyforyongyu:
Re-reviewed current head `2478f9885` with the downstream stack in view. The current unit-cover and unit-race failures both come from `TestListAccountsWithBalances`: `listAccountsWithBalances` now calls `scopeMgr.IsImportedAccount` at `wallet/account_manager.go:1090`, but `wallet/account_manager_test.go:1261-1275` only sets `AccountProperties` expectations for the loop. The mock panics on the unexpected `IsImportedAccount(*bdb.bucket,uint32)` call, so the test needs matching `IsImportedAccount` expectations for each account (or the helper/test needs to be adjusted).

(`openai/gpt-5.5`)

wallet/account_manager.go
yyforyongyu: Resolved as non-blocking legacy behavior: routed GetAccount/ListAccounts* use the kvdb balance split with the locked-inclusive UTXO source. propertiesToAccountInfo remains only for the deprecated Wallet.Accounts bridge until that legacy API is removed.

wallet/account_manager.go
yyforyongyu: Addressed in the pushed stack: propertiesToAccountInfo takes explicit isImported sourced from IsImportedAccount instead of props.IsWatchOnly, with regression coverage for locked derived and imported rows.

wallet/internal/db/data_types.go
yyforyongyu: Addressed in the pushed stack: TestBalanceParamsValidate now covers name-with-scope, name-without-scope, and account-and-name. TestBalanceNameFilterDisambiguatesMaskedImportedAccount covers imported/default account Balance name filters.

waddrmgr/scoped_manager.go
```diff
@@ -2824,6 +2824,24 @@ func (s *ScopedKeyManager) IsWatchOnlyAccount(ns walletdb.ReadBucket,
...
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	acctInfo, err := s.loadAccountInfo(ns, account)
+	if err != nil {
+		return false, err
+	}
+
+	return acctInfo.acctType == accountWatchOnly, nil
```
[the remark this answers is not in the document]
yyforyongyu: Addressed in the pushed stack: IsImportedAccount special-cases ImportedAddrAccount and returns true without loading row info; TestIsImportedAccountAcceptsLegacyImportedAccount covers it.

yyforyongyu:
LGTM 🌐
(`claude-opus-4-7[1m]`)

wallet/account_manager.go
```diff
@@ -216,6 +267,7 @@ func (w *Wallet) ListAccounts(_ context.Context) (*AccountsResult, error) {
 			// pre-calculated balances.
 			scopedAccounts, err := listAccountsWithBalances(
 				scopeMgr, addrmgrNs, accountBalances,
+				walletWatchOnly, w.masterFingerprint,
```
yyforyongyu, line 248: This `w.masterFingerprint` use is introduced before the `Wallet.masterFingerprint` field exists in the replayed stack, so GitHub `Check commits` fails on `d9890b7c wallet: surface db.AccountInfo on read APIs` with `w.masterFingerprint undefined` here and in the sibling call sites. Please fold the field/initialization into the commit that first uses it, or delay these uses so each commit builds independently. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -268,8 +320,9 @@ func (w *Wallet) ListAccountsByScope(_ context.Context,
 	if err != nil {
 		return nil, err
 	}
+	walletWatchOnly := w.addrStore.WatchOnly()
```
yyforyongyu, line 293: CI reports `wsl_v5` here: `missing whitespace above this line`. Add a blank line between the error-return block and this assignment to keep the method in the repo's stanza style. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -354,6 +409,9 @@ func (w *Wallet) ListAccountsByName(_ context.Context,
 
 				return err
 			}
+			if accNum == waddrmgr.ImportedAddrAccount {
```
yyforyongyu: This added branch is one of the paths pushing `ListAccountsByName` over the CI cyclop limit (`11`, max `10`), and `wsl_v5` also reports this line needs a blank line above it. Extracting the per-scope lookup/result assembly into a small helper would address the complexity and make the stanza split natural. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -424,17 +496,18 @@ func (w *Wallet) GetAccount(_ context.Context, scope waddrmgr.KeyScope,
 		if err != nil {
 			return err
 		}
+		if accNum == waddrmgr.ImportedAddrAccount {
```
yyforyongyu: CI reports the same `wsl_v5` missing-blank-line issue here. Add the stanza break before this `ImportedAddrAccount` guard as well. (`openai/gpt-5.5`)

wallet/account_manager.go
yyforyongyu: Addressed by rewriting the prep branch order: `wallet: cache master fingerprint on Wallet at Manager.Load` now lands before `wallet: surface db.AccountInfo on read APIs`, so the first commit that uses w.masterFingerprint builds independently.

wallet/account_manager.go
yyforyongyu: Addressed with the stanza break in ListAccountsByScope; `GOWORK=off make lint` now passes on the prep branch.

wallet/account_manager.go
yyforyongyu: Addressed by extracting the per-scope name lookup into accountInfoByNameInScope, which removes the ListAccountsByName cyclop issue and naturally adds the stanza break. `GOWORK=off make lint` passes on the prep branch.

wallet/account_manager.go
yyforyongyu: Addressed with the missing stanza break before the imported-account guard; `GOWORK=off make lint` passes on the prep branch.

yyforyongyu:
LGTM 🪙
(`claude-opus-4-7[1m]`)

yyforyongyu:
LGTM on current head `9f236eee5fac72f01696ee6095d9d73d6c82c612` from the requested code-review pass. I did not find any fresh non-duplicate issues beyond the existing resolved/non-blocking threads. (`openai/gpt-5.5`)

wallet/account_manager.go
```diff
@@ -393,6 +421,45 @@ func (w *Wallet) ListAccountsByName(_ context.Context,
...
+	accNum, err := scopeMgr.LookupAccount(addrmgrNs, name)
+	if err != nil {
+		if waddrmgr.IsError(err, waddrmgr.ErrAccountNotFound) {
+			return db.AccountInfo{}, false, nil
+		}
+
+		return db.AccountInfo{}, false, err
+	}
+
+	if accNum == waddrmgr.ImportedAddrAccount {
```
yyforyongyu: This now suppresses the reserved imported account from `ListAccountsByName("imported")`; `GetAccount` does the same below by returning `ErrAccountNotFound` when `LookupAccount` resolves to `ImportedAddrAccount`. The base implementation returned that account's properties, and the public account docs still describe `"imported"` as a reserved account rather than an unqueryable one. If the new `db.AccountInfo` surface intentionally excludes the legacy imported pseudo-account, the docs/tests should lock that in; otherwise this is an observable behavior regression for callers that inspect the imported account. (`openai/gpt-5.5`)

yyforyongyu:
LGTM 🏷️
(`claude-opus-4-7[1m]`)

yyforyongyu:
LGTM 👍 (`openai/gpt-5.5`)

GustavoStingelin (approved):
Ok

PR #1242, opened 2026-05-21
wallet: add wallet secrets read 

GustavoStingelin:
- Add `WalletStore.GetWalletSecrets(ctx, walletID uint32)`
   This is needed so the wallet can load its secrets into memory and use them for encryption and decryption flows.

- Refactor wallet files to use the ops interface
   This keeps the wallet flow aligned with the new store abstraction and centralizes database operations behind the ops layer.
  
- Converted some flows to use early return, making them easier to read. 
- Removed some underlines in the test names to keep aligned with the old project pattern.
- Split the wallet files into their own ops.


yyforyongyu:

wallet/internal/db/pg/wallet.go
```diff
@@ -132,32 +127,16 @@ func (s *Store) GetWallet(ctx context.Context,
...
 		row, err := q.GetWalletByName(ctx, name)
-		if err != nil {
-			if errors.Is(err, sql.ErrNoRows) {
-				return fmt.Errorf("wallet %q: %w", name,
-					db.ErrWalletNotFound)
-			}
-
-			return fmt.Errorf("get wallet: %w", err)
+		if err == nil {
+			info, err = getWalletRowToInfo(sqlc.GetWalletByIDRow(row))
```
yyforyongyu: The `sqlc.GetWalletByIDRow(row)` conversion here (and the byte-identical one at `wallet/internal/db/sqlite/wallet.go:131`) silently relies on `GetWalletByNameRow` and `GetWalletByIDRow` having structurally identical field sets — if either sqlc-generated row gains or loses a column the conversion fails at compile time, but a reader has no signal that this cast is load-bearing.

Compare to the same pattern in `wallet/internal/db/pg/accounts.go` for account rows:
```go
// Direct conversion works only because all constraint types have
// identical fields. If sqlc types diverge, compilation will fail.
base := sqlc.GetAccountByScopeAndNameRow(row)
```

A one-line comment above the conversion (same wording works) keeps the next maintainer who adds a column to only `GetWalletByName` from wondering why their build broke.

(`claude-opus-4-7[1m]`)

wallet/internal/db/sqlite/wallet.go
```diff
@@ -301,56 +280,106 @@ func (s *Store) GetEncryptedHDSeed(ctx context.Context,
...
 
-			return fmt.Errorf("get wallet: %w", err)
+			return nil
 		}
 
-		err = params.Validate(walletRow.IsWatchOnly)
-		if err != nil {
-			return err
+		if errors.Is(err, sql.ErrNoRows) {
+			return fmt.Errorf("secrets for wallet %d: %w", walletID,
```
yyforyongyu: Missing `wallet_secrets` rows are treated as missing wallets here, so a watch-only wallet that legitimately has no secrets row is reported as `ErrWalletNotFound`. The schema allows watch-only wallets without a `wallet_secrets` row, and `GetWalletSecrets` documents that absent watch-only secret material can be returned as empty fields; this should first check the wallet row and return empty `WalletSecrets` for an existing watch-only wallet, fixing the same path in both SQL backends. (`openai/gpt-5.5`)

wallet/internal/db/wallets_common.go
```diff
@@ -40,6 +44,160 @@ func (p *UpdateWalletSecretsParams) Validate(walletIsWatchOnly bool) error {
...
+		return err
+	}
+
+	err = ops.UpdateWalletSecrets(ctx, params)
+	if err != nil {
+		return fmt.Errorf("update wallet secrets: %w", err)
+	}
+
+	return nil
+}
```
yyforyongyu: I think this is gonna create quite some conflicts for #1229 , but the refactor is really nice. While we are at it, why don't we take it further - think we can split like how we did for utxo store, so break it down to `walletstore_createwallet.go` etc? Think i will just count on agents to figure out the rebase conflicts, should be easy.

wallet/internal/db/interface.go
```diff
@@ -179,6 +179,12 @@ type WalletStore interface {
 	// or an error if the retrieval fails.
 	GetEncryptedHDSeed(ctx context.Context, walletID uint32) ([]byte, error)
 
+	// GetWalletSecrets retrieves the encrypted wallet secret material for the
+	// given wallet. Watch-only wallets may return empty secret fields without
+	// error when those values are absent in storage.
+	GetWalletSecrets(ctx context.Context, walletID uint32) (*WalletSecrets,
```
yyforyongyu, line 187: guess my question is, when will this method be used? when we start the wallet?

GustavoStingelin: Yes, this will be used by the upcoming keyvault PRs, likely when initializing or unlocking the wallet.

Without this secret material loaded into memory, the wallet will not be able to encrypt or decrypt the data.

wallet/internal/db/sqlite/wallet.go
GustavoStingelin: Watch only wallets will still have encryption material for script encryption, since we classified that watch only wallets can store that data.

So we should always have this row, at least with the crypto keys needed for that path.

wallet/internal/db/wallets_common.go
GustavoStingelin: agreed

wallet/internal/db/pg/wallet.go
GustavoStingelin: holy cow, awsome suggestion, I missed that thing.

wallet/internal/db/walletstore_createwallet.go
```diff
@@ -0,0 +1,90 @@
...
+
+	err = ops.InsertWalletSecrets(ctx, walletID, params)
+	if err != nil {
+		return nil, fmt.Errorf("insert wallet secrets: %w", err)
+	}
+
+	err = ops.InsertWalletSyncState(ctx, walletID, params.Birthday)
+	if err != nil {
+		return nil, fmt.Errorf("upsert wallet sync state: %w", err)
+	}
```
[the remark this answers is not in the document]
GustavoStingelin: fixed

wallet/internal/db/sqlite/wallet.go
GustavoStingelin: improved the handling, btw

GustavoStingelin:
@yyforyongyu scope crept a bit, but I split the files.

yyforyongyu:
LGTM 🧷
(`claude-opus-4-7[1m]`)

wallet/internal/db/sqlite/walletstore_getencryptedhdseed.go
```diff
@@ -0,0 +1,49 @@
...
+					"encrypted master privkey for wallet %d: %w", walletID,
+					db.ErrSecretNotFound)
+			}
+
+			encrypted = secrets.EncryptedMasterHdPrivKey
+
+			return nil
+		}
+
+		if errors.Is(err, sql.ErrNoRows) {
```
yyforyongyu, line 37: The new `GetWalletSecrets` path now probes the wallet row when `wallet_secrets` is missing, but this seed read still maps `sql.ErrNoRows` from `wallet_secrets` directly to `ErrWalletNotFound`. Reproducing with an existing wallet after deleting its `wallet_secrets` row makes `GetEncryptedHDSeed` return `secrets for wallet 1: wallet not found`; it should surface `ErrSecretNotFound` like `GetWalletSecrets`, and the PostgreSQL path needs the same fix. (`openai/gpt-5.5`)

GustavoStingelin: improved the handling

yyforyongyu (commented):
LGTM 🎲 (`deepseek-v4-pro`)

wallet/internal/db/pg/walletstore_getwalletsecrets.go
```diff
@@ -0,0 +1,55 @@
...
+				EncryptedCryptoPrivKey:   secrets.EncryptedCryptoPrivKey,
+				EncryptedCryptoScriptKey: secrets.EncryptedCryptoScriptKey,
+				EncryptedMasterHdPrivKey: secrets.EncryptedMasterHdPrivKey,
+			}
+
+			return nil
+		}
+
+		if errors.Is(err, sql.ErrNoRows) {
+			_, walletErr := q.GetWalletByID(ctx, int64(walletID))
```
yyforyongyu: The 14-line "if secrets row missing, probe `GetWalletByID` and dispatch to `ErrSecretNotFound` / `ErrWalletNotFound` / wrapped-unexpected" block is now duplicated across four call sites: this file, `wallet/internal/db/pg/walletstore_getencryptedhdseed.go`, plus the two sqlite mirrors. The pg pair and sqlite pair are byte-identical within each backend; the cross-backend pair only differs by which `*sqlc.Queries` type the closure receives.

A tiny pg-local + sqlite-local helper drops the within-backend duplication immediately:

```go
// pg: walletstore_helpers.go (or alongside the existing pg helpers)
func resolveSecretsLookupErr(ctx context.Context, q *sqlc.Queries,
    walletID uint32, secretsErr error) error {

    if !errors.Is(secretsErr, sql.ErrNoRows) {
        return fmt.Errorf("get wallet secrets: %w", secretsErr)
    }

    _, walletErr := q.GetWalletByID(ctx, int64(walletID))
    if walletErr == nil {
        return fmt.Errorf("secrets for wallet %d: %w", walletID,
            db.ErrSecretNotFound)
    }
    if errors.Is(walletErr, sql.ErrNoRows) {
        return fmt.Errorf("wallet %d: %w", walletID,
            db.ErrWalletNotFound)
    }
    return fmt.Errorf("get wallet %d after missing secrets: %w",
        walletID, walletErr)
}
```

Each caller collapses to `return resolveSecretsLookupErr(ctx, q, walletID, err)`. The pg/sqlite duplication is unavoidable due to sqlc-package independence, but the within-backend duplication isn't.

(`claude-opus-4-7[1m]`)

wallet/internal/db/pg/wallet.go
```diff
@@ -29,97 +30,91 @@ func (s *Store) CreateWallet(ctx context.Context,
...
-		birthdayTimestamp := sql.NullTime{}
-		if !params.Birthday.IsZero() {
-			birthdayTimestamp = sql.NullTime{
-				Time:  params.Birthday,
-				Valid: true,
-			}
-		}
+// createWalletOps adapts PostgreSQL sqlc queries to the shared CreateWallet
+// workflow.
+type createWalletOps struct {
```
yyforyongyu: lets add compile time assertions for all the `Ops` implementations?

wallet/internal/db/pg/wallet.go
```diff
@@ -127,32 +127,16 @@ func (s *Store) GetWallet(ctx context.Context,
 
 	err := s.execRead(ctx, func(q *sqlc.Queries) error {
 		row, err := q.GetWalletByName(ctx, name)
-		if err != nil {
-			if errors.Is(err, sql.ErrNoRows) {
-				return fmt.Errorf("wallet %q: %w", name,
-					db.ErrWalletNotFound)
-			}
+		if err == nil {
```
yyforyongyu: nit: looks like these belong to the previous commit, non-blocking tho

wallet/internal/db/pg/walletstore_createwallet.go
```diff
@@ -0,0 +1,114 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return info, nil
+}
+
+// createWalletOps adapts PostgreSQL sqlc queries to the shared CreateWallet
+// workflow.
+type createWalletOps struct {
```
yyforyongyu, line 46: if we wanna do that compile assertion, let's do it here in this commit instead, otherwise there would be quite some rebase conflicts

wallet/internal/db/walletstore_createwallet.go
```diff
@@ -0,0 +1,90 @@
+package db
```
yyforyongyu, line 1: yeah i like this - let's stick to the naming convention `walletstore_xxx.go`?

wallet/internal/db/itest/walletstore_createwallet_test.go
```diff
@@ -0,0 +1,124 @@
+//go:build itest
```
yyforyongyu, line 1: same here re filenames

wallet/internal/db/pg/walletstore_getwalletsecrets.go
yyforyongyu: Seems to be a good suggestion

yyforyongyu (commented):
Nice cleanup! My main comment is to keep the filenames uniform

wallet/internal/db/sqlite/walletstore_updatewalletsecrets.go
```diff
@@ -0,0 +1,78 @@
...
+				params.EncryptedMasterHdPrivKey,
+			),
+			WalletID: int64(params.WalletID),
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	if rowsAffected == 0 {
```
yyforyongyu, line 75: After `WalletWatchOnly` has already loaded the wallet, `rowsAffected == 0` means the `wallet_secrets` row is missing, not that the wallet is missing, so return `ErrSecretNotFound` or insert the row here instead of reporting `ErrWalletNotFound`. (`gpt-5.5`)

wallet/internal/db/pg/walletstore_updatewalletsecrets.go
```diff
@@ -0,0 +1,78 @@
...
+				params.EncryptedMasterHdPrivKey,
+			),
+			WalletID: int64(params.WalletID),
+		},
+	)
+	if err != nil {
+		return err
+	}
+
+	if rowsAffected == 0 {
```
yyforyongyu, line 75: After `WalletWatchOnly` has already loaded the wallet, `rowsAffected == 0` means the `wallet_secrets` row is missing, not that the wallet is missing, so return `ErrSecretNotFound` or insert the row here instead of reporting `ErrWalletNotFound`. (`gpt-5.5`)

wallet/internal/db/pg/walletstore_getwalletsecrets.go
GustavoStingelin: added

wallet/internal/db/pg/wallet.go
GustavoStingelin: added

wallet/internal/db/pg/wallet.go
GustavoStingelin: moved

wallet/internal/db/walletstore_createwallet.go
GustavoStingelin: renamed, will also need to rename the tx_/utxo_store later

wallet/internal/db/itest/walletstore_createwallet_test.go
GustavoStingelin: renamed

wallet/internal/db/sqlite/walletstore_updatewalletsecrets.go
GustavoStingelin: fixed

wallet/internal/db/pg/walletstore_updatewalletsecrets.go
GustavoStingelin: fixed

yyforyongyu (commented):
LGTM 🎲 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪡
(`claude-opus-4-7[1m]`)

yyforyongyu (approved):
LGTM💯

yyforyongyu:
LGTM 🪴 (`gpt-5.5`)

PR #1243, opened 2026-05-25
wallet: migrate address-store read path + signer derivation

yyforyongyu:
## Summary

Split out from PR #1214 to keep both PRs under the ~10-commit cap matching the rest of this stack.

This PR carries the READ-side address routing on top of PR #1214 (which lands the WRITE-side):

- `kvdb: implement address lookup`
- `wallet: route GetAddressInfo and ScriptForOutput through store`
- `wallet: add derivation-based privKey path for store addresses`
- `kvdb: implement address listing`
- `wallet: route ListAddresses through store`
- `wallet: route GetPrivKeyForAddress through store-derived path`

The derivation-based privKey path is what lets SQL-derived addresses be spent — without it, `GetPrivKeyForAddress` and `ComputeUnlockingScript` would fail on SQL-backed wallets (the legacy `PrivKeyForAddress` lookup can't find SQL-only addresses; the new path consumes `AddressInfo.Derivation` from the store).

## Test plan

- [x] `GOWORK=off go build ./...` clean across each commit
- [x] `GOWORK=off go test ./wallet/...` passes
- [x] `make lint`: 0 issues
- [x] `TestGetPrivKeyForAddressSQLDerivedAddress` regression test added for the SQL-derived-address sign path
- [ ] CI: 13 checks pass

## Depends on

- #1214 (WRITE-side address routing + setup)

### DIFF
--- a/wallet/address_manager.go
+++ b/wallet/address_manager.go
@@ -201,6 +201,8 @@ var _ AddressManager = (*Wallet)(nil)
 
 // addressInfoFromManagedAddress converts one legacy managed address into the
 // wallet-owned metadata shape used by the prep work.
+//
+//nolint:unparam // Preserve the legacy caller error shape during migration.
 func addressInfoFromManagedAddress(
 	managedAddr waddrmgr.ManagedAddress) (AddressInfo, error) {
 
@@ -523,30 +525,7 @@ func nextUnusedStoreAddress(storeAddr db.AddressInfo,
 }
 
 // GetAddressInfo returns detailed information regarding a wallet address.
-//
-// This method provides metadata about a managed address, such as its type,
-// derivation path, and whether it's internal or compressed.
-//
-// How it works:
-// The method performs a direct lookup in the address manager to find the
-// requested address.
-//
-// Logical Steps:
-//  1. Initiate a read-only database transaction.
-//  2. Call the underlying address manager's `Address` method to look up the
-//     address.
-//  3. Return the managed address information.
-//
-// Database Actions:
-//   - This method performs a single read-only database transaction
-//     (`walletdb.View`).
-//   - It reads from the `waddrmgr` namespace to find the address.
-//
-// Time Complexity:
-//   - The operation is a direct database lookup, making its complexity roughly
-//     O(1) or O(log N) depending on the database backend's indexing strategy
-//     for addresses. It is a very fast operation.
-func (w *Wallet) GetAddressInfo(_ context.Context, a btcutil.Address) (
+func (w *Wallet) GetAddressInfo(ctx context.Context, a btcutil.Address) (
 	AddressInfo, error) {
 
 	err := w.state.validateStarted()
@@ -554,20 +533,22 @@ func (w *Wallet) GetAddressInfo(_ context.Context, a btcutil.Address) (
 		return AddressInfo{}, err
 	}
 
-	var managedAddress waddrmgr.ManagedAddress
+	scriptPubKey, err := txscript.PayToAddrScript(a)
+	if err != nil {
+		return AddressInfo{}, fmt.Errorf("pay to addr script: %w", err)
+	}
 
-	err = walletdb.View(w.cfg.DB, func(tx walletdb.ReadTx) error {
-		addrmgrNs := tx.ReadBucket(waddrmgrNamespaceKey)
-
-		managedAddress, err = w.addrStore.Address(addrmgrNs, a)
-
-		return err
-	})
+	storeAddr, err := w.store.GetAddress(
+		ctx, db.GetAddressQuery{
+			WalletID:     w.id,
+			ScriptPubKey: scriptPubKey,
+		},
+	)
 	if err != nil {
 		return AddressInfo{}, err
 	}
 
-	return addressInfoFromManagedAddress(managedAddress)
+	return addressInfoFromStoreAddress(storeAddr, w.cfg.ChainParams)
 }
 
 // ListAddresses lists all addresses for a given account, including their
@@ -821,9 +802,10 @@ func encryptTaprootScript(vault keyvault.Vault,
 //     sign for (e.g., P2WKH, NP2WKH, P2TR).
 //  3. Based on the address type, construct the appropriate scripts:
 //     - For nested P2WKH (NP2WKH), it returns the inner witness program as the
-//     redeem script.
+//     redeem script and also builds the single-push sigScript wrapper used in
+//     the final input.
 //     - For native SegWit outputs (P2WKH, P2TR), the `witnessProgram` is the
-//     output's `pkScript`, while the redeem script is nil.
+//     output's `pkScript`, while the redeem script and sigScript are nil.
 //
 // Database Actions:
 //   - This method performs a read-only database access to fetch address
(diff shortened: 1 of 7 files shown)


wallet/signer_test.go
```diff
@@ -943,6 +943,39 @@ func TestComputeUnlockingScriptSQLDerivedAddress(t *testing.T) {
 	require.NoError(t, vm.Execute(), "script execution failed")
 }
 
+// TestGetPrivKeyForAddressSQLDerivedAddress verifies that GetPrivKeyForAddress
+// recovers the private key for an address whose only persistent record lives
+// in the SQL store. Regression coverage for the GetPrivKeyForAddress path
+// raised in PR #1214 review comment r3287556227.
```
yyforyongyu: can you remove this local-related context??

wallet/address_manager_test.go
```diff
@@ -115,6 +159,24 @@ func addressIter(items ...db.AddressInfo) iter.Seq2[db.AddressInfo, error] {
 // TestNewAddress tests the NewAddress method, ensuring it can generate
 // various address types for different accounts and correctly handles both
 // internal and external address generation.
+// expectStoreAddressInfo configures mock expectations for address lookup.
```
yyforyongyu, line 203: The comment block for `expectStoreAddressInfo` incorrectly includes the documentation for `TestNewAddress` at the top, which violates the Go Comment Gate rule that requires the comment to start exactly with the declaration name. Please move the `TestNewAddress` comment down to where `TestNewAddress` is declared. (`gemini-3.1-pro-preview`)

yyforyongyu:
LGTM 🪢
(`claude-opus-4-7[1m]`)

yyforyongyu (commented):
LGTM 🦾 (`deepseek-v4-pro`)

wallet/address_manager.go
```diff
@@ -300,6 +302,58 @@ func addressInfoFromStoreAddress(storeAddr *db.AddressInfo,
...
+	err = walletdb.View(w.cfg.DB, func(tx walletdb.ReadTx) error {
+		ns := tx.ReadBucket(waddrmgrNamespaceKey)
+		path := waddrmgr.DerivationPath{
+			InternalAccount: accountID,
+			Account:         accountID + hdkeychain.HardenedKeyStart,
+			Branch:          branch,
+			Index:           index,
+		}
+
+		managedAddr, err := manager.DeriveFromKeyPath(ns, path)
```
yyforyongyu: This still derives SQL-store addresses through the legacy `waddrmgr` account number, so it only works when the same account also exists in the legacy manager. A store-only account created via `Store.CreateDerivedAccount` as account 1 fails in `w.NewAddress(..., "secondary", ...)` before notification with `derive address: derive key path: account 1 not found`, so SQL-backed wallets can only exercise the default mirrored account path covered by the new tests. (`openai/gpt-5.5`)

yyforyongyu (commented):
LGTM 🦾 (`deepseek-v4-pro`)

wallet/signer_test.go
yyforyongyu: Done — dropped the trailing `(regression r3287556227)` suffix from the assertion message and the second sentence from the `TestGetPrivKeyForAddressSQLDerivedAddress` docstring. Also reworded the production TODO above `legacyAddressBalances` in `wallet/address_manager.go` to describe the condition (kvdb backend implementing `Store.ListUTXOs`) rather than reference PR #1238 by number. Landed in d7a02b97b517aef495ac11cc7201fb3aaf24d9f9.

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -200,20 +204,122 @@ func validateImportedAddress(params db.NewImportedAddressParams,
...
+
+	addrMgr := s.addrStore
 
-	return nil, notImplemented(ctx, "GetAddress")
+	addr := addressFromScript(
+		query.ScriptPubKey, addrMgr.ChainParams(),
+	)
+	if addr == nil {
+		return nil, db.ErrAddressNotFound
+	}
```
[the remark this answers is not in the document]
yyforyongyu: Done — `addressFromScript` now re-encodes the extracted address via `txscript.PayToAddrScript` and rejects scripts whose re-encoding does not byte-match the input `pkScript`. Scripts that aren'\''t uniquely identified by an extracted standard address (bare multisig, malformed witness scripts, etc.) now return nil. Folded into 747f1ed31 "kvdb: implement address lookup" with regression coverage in `addressstore_test.go`.

yyforyongyu:
LGTM 🌾
(`claude-opus-4-7[1m]`)

wallet/psbt_manager_test.go
```diff
@@ -600,19 +545,9 @@ func TestDecorateInputErrDerivationMissing(t *testing.T) {
...
-	mocks.pubKeyAddr.On("Imported").Return(false).Once()
-	mocks.pubKeyAddr.On("Internal").Return(false).Once()
-	mocks.pubKeyAddr.On("Compressed").Return(true).Once()
-	mocks.pubKeyAddr.On("PubKey").Return(pubKey).Once()
-	mocks.pubKeyAddr.On("DerivationInfo").Return(
-		waddrmgr.KeyScope{}, waddrmgr.DerivationPath{}, false,
-	).Once()
+	expectSignerAddressInfo(
+		t, w, mocks, p2wkhAddr, db.WitnessPubKey, false, true, pubKey,
+	)
```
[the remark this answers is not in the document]
yyforyongyu: Done — `TestDecorateInputErrDerivationMissing` now sets `imported=false` and the comment is updated so the test exercises the missing-derivation branch as a non-imported pubkey address whose store record carries no usable derivation scope, distinct from `TestDecorateInputErrImported`. Folded into b39054b3c "wallet: route GetAddressInfo and ScriptForOutput through store".

wallet/address_manager.go
yyforyongyu: Done — `deriveAddressData` now derives directly from `params.AccountPubKey` (the account-level extended public key threaded through `AddressDerivationParams`) using `hdkeychain.NewKeyFromString` plus `deriveChildKey`. The legacy `addrStore.FetchScopedKeyManager` + `manager.DeriveFromKeyPath` path is gone, so a SQL-only account created via `Store.CreateDerivedAccount` no longer requires a mirrored legacy waddrmgr account. Folded into 1e199e2d4 "wallet: add derivation-based privKey path for store addresses"; `TestNewAddressOnSQLOnlyAccount` covers the regression.

yyforyongyu (commented):
LGTM 🦾 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🌱
(`claude-opus-4-7[1m]`)

wallet/signer.go
```diff
@@ -717,18 +719,54 @@ func (w *Wallet) ComputeUnlockingScript(ctx context.Context,
...
+
+	derivationPath := waddrmgr.DerivationPath{
+		InternalAccount:      derivation.Account,
+		Account:              derivation.Account + hdkeychain.HardenedKeyStart,
+		Branch:               derivation.Branch,
+		Index:                derivation.Index,
+		MasterKeyFingerprint: derivation.MasterKeyFingerprint,
+	}
+
+	return w.resolveDerivedPathPrivKey(derivation.KeyScope, derivationPath)
```
yyforyongyu, line 767: This routes store-only address metadata back through the legacy `waddrmgr` account path. For an address from a SQL-only account (for example the `secondary` account created by `Store.CreateDerivedAccount`), `DeriveFromKeyPathCache` misses and the fallback loads `InternalAccount` from the legacy DB, where account 1 does not exist, so signing/GetPrivKey fail. Derive from SQL account secret/master key or cover this with a SQL-only account signing regression. (`gpt-5.5`)

yyforyongyu: Confirmed — this is the symmetric counterpart to Task 118: the public-key side is fixed via `AddressDerivationParams.AccountPubKey` plumbing; the private-key side still routes through legacy waddrmgr.

The signing rework belongs in the signer-store work on impl-tx-creator-store rather than this address-store PR. To scope what landed here, the tip commit now adds:

- A `TODO(yy):` on `wallet/signer.go` explicitly calling out the SQL-only signing gap and pointing at the keyVault-backed fix.
- An extended docstring on `TestNewAddressOnSQLOnlyAccount` noting it covers only the public-key path; signing-side coverage is deferred to the signer-store work.

The follow-up is tracked locally as Task 144 (`.lgtm/tasks/144-signer-sql-only-account-derivation.md`) and will:

- Extend `runtimeCache` with `GetAccountSecret`.
- Add a SQL-only branch in `resolveDerivedPathPrivKey` that fetches the account secret, decrypts via `w.keyVault`, and derives at branch/index locally.
- Add `TestSignFromSQLOnlyAccount` as the signing-side regression.
- Remove the TODO(yy) in the same commit that closes this gap.

New PR #1243 tip 19181e0cd8.

yyforyongyu: Thanks for confirming. I do not think this is LGTM as-is because the PR now enables SQL-only public address derivation (`TestNewAddressOnSQLOnlyAccount` covers that path), while the matching signer path still cannot spend those addresses. Please either land the keyVault-backed signing fix here or gate SQL-only address generation until the signer-store PR lands. (`gpt-5.5`)

yyforyongyu: That scope makes sense. If this only lands on the side branch and the branch is not shipped/merged to master before the signer-store follow-up, then keeping this PR focused on the address read/derivation side is fine. LGTM. (`gpt-5.5`)

GustavoStingelin (approved):
LGTM!

PR #1246, opened 2026-05-28
wallet: enforce wallet-level watch-only invariant (ADR 0012)

yyforyongyu:
## Summary

Enforce a wallet-level watch-only invariant per [ADR 0012](docs/developer/adr/0012-wallet-level-watch-only-uniformity.md):
spendable wallets cannot have public-only imported accounts and
watch-only wallets cannot have private-only accounts. This aligns
btcwallet's account model with the Bitcoin Core descriptor-wallet
shape and removes the mixed-mode legacy wallet shape from the SQL
store.

Stack position: **#2** of the kvdb-migration stack. Stacks on top
of #1229. Downstream consumers: prep-utxo-enrichment,
impl-utxo-kvdb (#1238), and the rest of the kvdb-migration chain.

## Changes

- `docs/developer/adr/0012-wallet-level-watch-only-uniformity.md` —
  new ADR documenting the invariant and the rationale.
- New `db` sentinels for SQL backends to surface invariant
  violations consistently:
  `ErrSpendableWalletNeedsAccountPrivKey`,
  `ErrSpendableWalletNeedsAddressPrivKey`,
  `ErrLegacyMixedModeWallet`.
- pg/sqlite migration `000011_watch_only_invariant` adds deferred
  constraint triggers that enforce the symmetric invariant at the
  SQL level.
- Wallet loader audits the legacy mixed-mode shape and rejects it
  at load time.
- `Wallet.IsWatchOnly()` accessor + per-row docstring cleanups so
  callers no longer carry an `IsWatchOnly` bit on every `UtxoInfo`
  row.
- Wallet creation validates `WatchOnly` + `InitialAccounts` upfront
  rather than failing partway through the migration.

## Test plan

- [ ] `GOWORK=off go test ./wallet/...`
- [ ] `make itest-db db=postgres`
- [ ] `make itest-db db=sqlite`


yyforyongyu (commented):
LGTM 🐙 (`deepseek-v4-pro`)

wallet/manager.go
```diff
@@ -98,6 +98,22 @@ type CreateWalletParams struct {
 // validate ensures that the parameters are consistent with the chosen creation
 // mode.
 //
+// validateInitialAccountsMode enforces the ADR 0012 wallet-level watch-only
+// invariant against the params before any on-disk artifact is created. A
+// non-watch-only wallet cannot ship with watch-only InitialAccounts; the
+// import would later be rejected by requireAccountPrivKeyOnSpendable but
+// only after the wallet row had already been written. The check fires
+// once at create time so the failure is atomic.
+func validateInitialAccountsMode(params CreateWalletParams) error {
```
yyforyongyu, line 104: The doc-comment block on lines 98-106 is one continuous comment that godoc reads as the doc for `validateInitialAccountsMode`. It still leads with `// validate ensures that the parameters are consistent with the chosen creation mode.` — the original doc for `validate()`. Net effect: `validateInitialAccountsMode`'s godoc starts with text describing a different function, and `validate()` (further down) ends up with no godoc, only the `// We skip cyclop ...` rationale line. Delete the leading two-line paragraph here and re-attach a fresh `// validate ensures ...` doc directly above `func (p *CreateWalletParams) validate()`.

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/walletstore_audit.go
```diff
@@ -0,0 +1,82 @@
...
+
+		return nil
+	})
+}
+
+// auditScope walks the accounts in a single key scope and rejects any
+// imported account whose acctKeyPriv is nil (legacy mixed-mode signature).
+func auditScope(ns walletdb.ReadBucket, scopedMgr waddrmgr.AccountStore,
+	walletID uint32) error {
+
```
yyforyongyu: `auditScope` only walks `scopedMgr.ActiveAccounts()`. Per `loadAccountInfo` in `waddrmgr/scoped_manager.go:518` and the `account != ImportedAddrAccount` guard at `:669`, `ImportedAddrAccount` is never loaded into `s.acctInfo` — so it never appears in `ActiveAccounts()`. As a result, the audit silently skips the legacy mixed-mode shape that `ImportPublicKey` / `ImportScript` (single-key / single-script `importaddress` RPC calls) produce on a spendable wallet: addresses with `encryptedPrivKey == nil` parked under `ImportedAddrAccount`.

The PG/SQLite `AuditWallet` queries here catch this case (they probe `addresses WHERE address_branch IS NULL AND address_index IS NULL AND s.address_id IS NULL OR s.encrypted_priv_key IS NULL`). The kvdb audit doesn't, so ADR 0012's load-boundary guarantee holds for SQL but is weaker for the legacy kvdb backend exactly where it matters most (existing pre-ADR-0012 deployments).

Fix: after the `ActiveAccounts()` walk, iterate `scopedMgr.ForEachAccountAddress(ns, waddrmgr.ImportedAddrAccount, fn)` and reject any address whose underlying row has no encrypted private-key material (via `waddrmgr.ManagedPubKeyAddressHasPrivateKey` for pubkey imports, and reject `ManagedScriptAddress` outright since script-only imports cannot satisfy the spendable invariant).

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/walletstore_audit.go
```diff
@@ -0,0 +1,82 @@
...
+	for _, account := range scopedMgr.ActiveAccounts() {
```
yyforyongyu: `ActiveAccounts()` only returns accounts currently cached in memory, and `waddrmgr.Open` creates scoped managers with empty `acctInfo` maps. A legacy spendable wallet with a persisted watch-only xpub account therefore passes this load audit until some later lookup happens, so `ErrLegacyMixedModeWallet` is never raised at the boundary. Iterate persisted accounts with `ForEachAccount` instead, skipping or handling the imported-address pseudo-account separately. (`gpt-5.5`)

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,112 @@
...
+- A backend-aware wallet-load audit rejects legacy mixed-mode wallets
+  before any caller can observe the reclassified account or address
+  rows. SQL backends implement an `AuditWallet(ctx, walletID)` method on
+  their concrete `*Store` types; kvdb runs the audit inside its wallet
+  loader against the waddrmgr buckets. The audit surfaces
+  `ErrLegacyMixedModeWallet` on detection.
+
+## 4. References
+
+- ADR 0006: address-store account/origin model (related).
```
yyforyongyu: this should be hyperlinked to the file

wallet/internal/sql/pg/migrations/000011_watch_only_invariant.down.sql
```diff
@@ -0,0 +1,53 @@
+-- Drop the symmetric watch-only invariant triggers and their backing
```
yyforyongyu: there is no need to create new migrations, we are in the feature branch so we are free to touch existing migrations!

wallet/internal/db/kvdb/walletstore_audit.go
```diff
@@ -0,0 +1,82 @@
+package kvdb
```
yyforyongyu: we are in a feature branch, so there is no need for this commit at all

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Done — the referenced file is now a markdown link in ADR 0012.

wallet/internal/sql/pg/migrations/000011_watch_only_invariant.up.sql
```diff
@@ -0,0 +1,280 @@
...
+-- Migration 000005 / 000006 already enforced the other direction (watch-only
+-- wallets cannot store secrets); this migration completes the pair.
+--
+-- The check has an awkward FK-direction problem: creating an imported account
+-- inserts the accounts row BEFORE its account_secrets row, so a naive BEFORE
+-- INSERT trigger rejects the legitimate create flow. Postgres CONSTRAINT
+-- TRIGGER ... DEFERRABLE INITIALLY DEFERRED fires at COMMIT, after both
+-- inserts have happened in the same transaction. SQLite triggers are not
+-- deferrable, so SQLite enforces the invariant at the application layer in
+-- wallet/internal/db/accounts_common.go and addresses_common.go.
```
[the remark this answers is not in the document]
yyforyongyu: Moot — there is no 000011 migration anymore; the deferred-constraint DDL is folded into the existing 000005/000006 migrations, and SQLite enforces the invariant at the application layer (requireAccount/AddressPrivKeyOnSpendable).

wallet/internal/db/pg/walletstore_audit.go
```diff
@@ -0,0 +1,82 @@
...
+
+func (s *Store) auditQuery(ctx context.Context, query string,
+	walletID uint32, detail string) error {
+
+	var mixedMode bool
+
+	err := s.db.QueryRowContext(ctx, query, int64(walletID)).Scan(&mixedMode)
+	if err != nil && err != sql.ErrNoRows {
+		return fmt.Errorf("audit wallet %d: %w", walletID, err)
+	}
```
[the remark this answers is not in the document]
yyforyongyu: Resolved by deletion — walletstore_audit.go (pg/sqlite/kvdb) is removed; the invariant is enforced at create time instead.

wallet/internal/db/kvdb/walletstore_audit.go
```diff
@@ -0,0 +1,82 @@
...
+		if err != nil {
+			return fmt.Errorf("is watch-only account %d: %w",
+				account, err)
+		}
+
+		if isWatchOnly {
+			return fmt.Errorf("%w: imported account %d has no "+
+				"acctKeyPriv (wallet_id=%d)",
+				db.ErrLegacyMixedModeWallet, account, walletID)
+		}
```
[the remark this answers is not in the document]
yyforyongyu: Resolved by deletion — the kvdb audit (and its wallet_id=0 error message) is removed.

wallet/internal/db/itest/account_store_test.go
```diff
@@ -1898,23 +1937,55 @@ func createDerivedAccount(t *testing.T, store db.AccountStore, walletID uint32,
...
 	require.NoError(t, err)
+
+	var wallet *db.WalletInfo
+	for i := range page.Items {
+		if page.Items[i].ID == walletID {
+			wallet = &page.Items[i]
+			break
+		}
+	}
+	require.NotNil(t, wallet, "wallet %d not found", walletID)
```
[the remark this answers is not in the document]
yyforyongyu: Done — CreateImportedAccount now takes an explicit watchOnly bool passed by each caller; the paginated ListWallets scan is gone.

wallet/manager.go
yyforyongyu: Done — the godoc is split so validateInitialAccountsMode and validate() each carry their own doc comment.

wallet/internal/db/kvdb/walletstore_audit.go
yyforyongyu: Resolved — the load-time audit is removed entirely (no ActiveAccounts walk left); the invariant is enforced at imported-account/address create time, which kvdb is intentionally grandfathered out of (ADR 0012 reframed as a new-SQL-store invariant).

wallet/internal/db/kvdb/walletstore_audit.go
yyforyongyu: Resolved — the load-time audit is removed; enforcement moved to create time (kvdb grandfathered), so the empty-ActiveAccounts-on-Open gap no longer applies.

wallet/internal/sql/pg/migrations/000011_watch_only_invariant.down.sql
yyforyongyu: Done — dropped the 000011 migration; the deferred-constraint DDL is folded into the existing 000005 (accounts) / 000006 (addresses) migrations on this branch.

wallet/internal/db/kvdb/walletstore_audit.go
yyforyongyu: Done — the load-time audit commit is dropped; write-time enforcement (require*PrivKeyOnSpendable) is kept, the wallet-level watch-only flag is immutable, and ADR 0012 is reframed (SQL-store invariant; kvdb grandfathered).

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

wallet/internal/db/addresses_common.go
```diff
@@ -123,6 +129,25 @@ func (p NewImportedAddressParams) ValidateWatchOnly(
 	return nil
 }
 
+// requireAddressPrivKeyOnSpendable enforces the ADR 0012 symmetric
```
yyforyongyu, line 136: Stale commit message after the restructure that dropped the load-boundary audit. This commit's message says:

> kvdb's legacy imported address rows are checked at load time by AuditWallet (added later in this PR), not at create time.

but `AuditWallet` no longer exists anywhere in the PR (`git grep AuditWallet` on the head is empty), and the new `3a080c000` wires `requireAddressPrivKeyOnSpendable` in at **create** time — the opposite of what this message claims. The sibling account-sentinel commit `b0772c9e5` has the same stale reference ("the load-boundary AuditWallet check (added later in this PR) rejects pre-existing mixed-mode persisted state on any backend").

The in-code comments were correctly updated to the grandfathered-kvdb framing; only these two commit messages still describe the removed mechanism. Please reword both so `git log` doesn't send a reader hunting for an AuditWallet that isn't there.

(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🧭 (`gpt-5.5`)

wallet/internal/db/addresses_common.go
yyforyongyu: Reworded — neither commit message references `AuditWallet` anymore (the load-boundary audit was dropped). SQL backends enforce the ADR-0012 symmetric invariant at imported account/address **create time**; kvdb is grandfathered. `git grep AuditWallet` is empty. Resolving.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -185,3 +185,125 @@ CREATE TRIGGER trg_assert_watch_only_account_secrets_update
...
+
+CREATE CONSTRAINT TRIGGER trg_assert_spendable_account_secret_accounts_insert
+AFTER INSERT ON accounts
+DEFERRABLE INITIALLY DEFERRED
+FOR EACH ROW
+WHEN (new.origin_id = 1) -- 1 = ImportedAccount; derived accounts are exempt.
+EXECUTE FUNCTION trg_assert_spendable_account_secret_accounts();
+
+CREATE CONSTRAINT TRIGGER trg_assert_spendable_account_secret_accounts_update
+AFTER UPDATE OF origin_id ON accounts
```
yyforyongyu: Stale detail in this commit's message vs the code it lands. The "Trigger event coverage" block claims:

```
  accounts:  INSERT, UPDATE OF origin_id / wallet_id
  addresses: INSERT, UPDATE OF address_branch / address_index / wallet_id, ...
```

but the actual symmetric-invariant triggers fire on `AFTER UPDATE OF origin_id` (here) and `AFTER UPDATE OF address_branch, address_index` (000006) — `wallet_id` was intentionally dropped from both `UPDATE OF` lists. The inline comment a few lines up even says so: "UPDATE OF wallet_id is independently rejected by the existing wallet_id immutability trigger."

So `wallet_id` UPDATEs are covered by `trg_assert_account_wallet_id_immutable` / `trg_assert_address_wallet_id_immutable`, not by these triggers — the change is correct, but the message's coverage list overstates what these triggers watch and contradicts the inline comment. Please drop `wallet_id` from the accounts/addresses lines in the coverage block.

(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🪁 (`gpt-5.5`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: Fixed — the trigger commit's coverage block no longer lists `wallet_id`; the symmetric-invariant triggers fire on `UPDATE OF origin_id` (accounts) and `UPDATE OF address_branch, address_index` (addresses), and the message now notes `wallet_id` mutations are rejected by the separate immutability triggers. Pushed + CI-green. Resolving.

yyforyongyu:
LGTM 🧱
(`claude-opus-4-8[1m]`)

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🛰️ (`gpt-5.5`)

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🧱
(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🐬 (`gpt-5.5`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -185,3 +185,112 @@ CREATE TRIGGER trg_assert_watch_only_account_secrets_update
...
+-- 1. accounts <-> account_secrets symmetric invariant
+----------------------------------------------------------------------
+
+-- Spendable wallet's imported accounts must have an account_secrets row by
+-- COMMIT. Fires on accounts (INSERT, UPDATE OF origin_id) and on
+-- account_secrets (DELETE). DELETE on accounts cascades through the
+-- accounts -> account_secrets FK with ON DELETE RESTRICT so it can't drop a
+-- referenced secrets row; UPDATE OF wallet_id is independently rejected by
+-- the existing wallet_id immutability trigger.
+CREATE FUNCTION assert_spendable_imported_account_has_secret(
```
yyforyongyu: A bit hesitant to add too complicated queries, but I guess it's better to guard at the SQL level, so open to dropping it.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🧩 (`gpt-5.5`)

yyforyongyu:
LGTM 🧱
(`claude-opus-4-8[1m]`)

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,118 @@
...
+  wallet's lifecycle reasoning.
+
+Current Bitcoin Core has moved away from that model. The Core descriptor
+wallet:
+
+- Is created with `disable_private_keys` set to true or false; the flag
+  is immutable after creation.
+- Rejects descriptor imports whose mode conflicts with the wallet:
+  > Cannot import descriptor without private keys to a wallet with
+  > private keys enabled.
```
GustavoStingelin, line 37: It may be useful to reference the relevant Bitcoin Core code here, either by linking directly to the blob or by linking to the reference section.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,118 @@
...
+## 4. References
+
+- [ADR 0006](0006-wtxmgr-sql-schema.md): wtxmgr SQL schema (related).
+- [ADR 0011](0011-no-addresses-used-column.md): addresses table omits
+  the `used` column (related — also a
+  "compute, don't persist" choice; this ADR makes the opposite call by
+  persisting wallet-level watch-only).
+- [Bitcoin Core `disable_private_keys` documentation][btc-core-disable-priv].
+
+[btc-core-disable-priv]: https://bitcoincore.org/en/doc/26.0.0/rpc/wallet/createwallet/
```
GustavoStingelin: IMO, it is better to point to the release that matches the same period when this btcwallet change was written. That makes it clearer that we are documenting the current behavior, instead of relying on an older version.

```suggestion
[btc-core-disable-priv]: https://bitcoincore.org/en/doc/31.0.0/rpc/wallet/createwallet/
```

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,118 @@
...
+immutable thereafter. Imports whose mode conflicts with the wallet are
+rejected at the store boundary (and, on PostgreSQL, by a deferred
+constraint trigger as defense in depth).
+
+Implications:
+
+- A non-watch-only wallet rejects imports that would introduce
+  watch-only-only material (an imported xpub account without matching
+  encrypted account private-key material, or an imported address
+  without encrypted private-key material).
```
GustavoStingelin, line 56: I guess this means that we can't track P2A addresses bc they don't have private keys.

GustavoStingelin: I think we will need to clarify how we will handle that, at least as a transient state, as they will always have a second tx spending and fee bumping the parent.

wallet/internal/db/itest/address_store_test.go
```diff
@@ -135,16 +135,51 @@ func flattenAddressPages(
 
 // TestNewImportedAddress verifies that NewImportedAddress correctly imports
 // addresses of different types, with and without private key material.
+// importedAddressWallet picks the wallet flavor a subtest needs based on
+// whether the imported address carries private-key material. Per ADR
+// 0012, public-only (no private key) imports require a watch-only
+// wallet, while imports with private-key material require a spendable
+// wallet.
+func importedAddressWallet(t *testing.T, store db.WalletStore,
```
GustavoStingelin: I guess that sth broke here during rebase; the comment mentions a test, but it's a helper

wallet/internal/db/itest/address_store_test.go
```diff
@@ -135,16 +135,51 @@ func flattenAddressPages(
...
+	name string, withPrivateKey bool) uint32 {
+
+	t.Helper()
+
+	if withPrivateKey {
+		return newWallet(t, store, name)
+	}
+
+	return newWatchOnlyWallet(t, store, name)
```
GustavoStingelin, line 150: Also, maybe we could apply a little refactor here to be cleaner, like having one `newWatchOnlyWallet` and one `newSpendableWallet`, then we change the old `newWallet` to accept that boolean flag for devX.

wallet/internal/db/itest/address_store_test.go
```diff
@@ -135,16 +135,51 @@ func flattenAddressPages(
...
+}
+
+// importedAddressAccount creates an imported account on the given wallet
+// with the right shape for the wallet's watch-only state. Spendable
+// wallets receive an account with EncryptedPrivateKey; watch-only
+// wallets get the public-only variant.
+func importedAddressAccount(t *testing.T, store db.AccountStore,
+	walletID uint32, scope db.KeyScope, withPrivateKey bool) {
```
GustavoStingelin: Would `newImportedAccount` be clearer here? Starting the name with `imported` makes it less obvious that this function creates a new account.

wallet/internal/db/accountstore_createimported.go
```diff
@@ -32,6 +38,21 @@ func (params *CreateImportedAccountParams) ValidateWatchOnly(
...
+func requireAccountPrivKeyOnSpendable(walletID uint32, name string,
+	walletIsWatchOnly bool, encryptedPrivKey []byte) error {
+
+	if walletIsWatchOnly || len(encryptedPrivKey) > 0 {
+		return nil
+	}
+
+	return fmt.Errorf("wallet %d cannot create imported account %q: %w",
+		walletID, name, ErrSpendableWalletNeedsAccountPrivKey)
+}
```
GustavoStingelin, line 70: How should the import flow work when importing an address with a private key, but the address does not belong to any existing account?

In other words, should it only be possible to import an address if we know which account it was derived from? Or should this address be allowed into a non-related existing account that already has private key material?

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -185,3 +185,112 @@ CREATE TRIGGER trg_assert_watch_only_account_secrets_update
...
+    p_account_id BIGINT
+) RETURNS BOOLEAN AS $$
+BEGIN
+    IF EXISTS (
+        SELECT 1
+        FROM accounts AS a
+        INNER JOIN wallets AS w ON w.id = a.wallet_id
+        WHERE a.id = p_account_id
+          AND a.origin_id = 1
```
GustavoStingelin: IIUC, the ADR does not differentiate imported and derived things, so why do we have this `a.origin_id = 1` here?

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
GustavoStingelin: The query itself looks easy; the difficult part is the `DEFERRABLE INITIALLY DEFERRED`

GustavoStingelin: I would say that, if we keep these triggers, we should also exercise them in specific PostgreSQL integration tests. They are easy to get wrong, and we need to make sure they work as expected.

GustavoStingelin: If we cannot enforce this consistently in both databases, and we already enforce it at the application level, then adding database enforcement may duplicate the work without much benefit. Because of that, I would prefer keeping this as application side validation only.

Also, if someone really wants to break the database, they can disable the trigger and modify the data manually. So, as long as the application can detect and handle that state safely, I think that is acceptable.

I found some Bitcoin Core code that logs a warning when a watch only wallet is loaded with private keys on disk, which could happen because of a bug or manual insertion.

GustavoStingelin (changes requested):
Left some comments. Also,

We still have some queries with this shape 

```
    CASE
        WHEN w.is_watch_only THEN TRUE
        WHEN a.origin_id = 1 AND acs.account_id IS NULL THEN TRUE
        ELSE FALSE
    END AS is_watch_only
```

In GetAccountByScopeAndNumber, for example, should that be refactored too?

and
```
isWatchOnly = walletWatchOnly || props.IsWatchOnly
```
In wallet/account_manager.go

wallet/internal/db/accountstore_createimported.go
yyforyongyu: why would someone import an addr when it can be derived? Importing an address means “track this exact script/address." - and it should not attach to a user’s derived account, because the account xpub cannot derive it?

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: cool decided to drop it

> I found some Bitcoin Core code that logs a warning when a watch only wallet is loaded with private keys on disk, which could happen because of a bug or manual insertion.

my guess is to keep backward compatibility

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: it may be helpful to think p2a as a constant, static address rather than some user-generated addresses - there is nothing to track in the db, but a more relevant place to track it is probably in the utxo tables. mayb e check [this](https://github.com/lightningnetwork/lnd/discussions/9105). anyway, i've updated the doc, but the p2a discussion is out of this PR's scope - we need a proper design session to implement it, once the sql series is done.

wallet/internal/db/accountstore_createimported.go
GustavoStingelin: Exactly, that is the part I am trying to clarify.

If importing an address means “track this exact script or address”, and it is not necessarily derivable from the account xkey, why does the imported account itself need private key material?

Should the private key requirement apply to each imported address instead of the imported account?

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
GustavoStingelin: Concept ACK. I brought this up because I think we may need to remove or redesign `address_types.7`, `P2A`, in the DB, since it was intended to track later anchor addresses.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Added explicit Core references in the ADR references section: v23 release notes for descriptor-wallet defaulting, v31 createwallet docs for disable_private_keys, and a direct v31 backup.cpp ProcessDescriptorImport link for the descriptor import rejection.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: Dropped the deferred PostgreSQL trigger path. The invariant is now enforced once at the application/store boundary for both SQL backends, and the migration no longer contains the account-secret deferred trigger. The ADR also calls out why the trigger was removed.

wallet/internal/db/itest/address_store_test.go
yyforyongyu: Fixed by separating the helper doc from TestNewImportedAddress. The helper comment now starts with importedAddressWallet, and TestNewImportedAddress has its own test comment.

wallet/internal/db/accountstore_createimported.go
yyforyongyu: Yes, for individually imported addresses the requirement is per-address. The current branch separates that from true imported xpub accounts: NewImportedAddress auto-creates/reuses the reserved keyless imported bucket, and spendable-wallet validation checks each imported address for EncryptedPrivateKey. CreateImportedAccount is only for true imported xpub accounts, where account-level private key material is required on a spendable wallet; otherwise it would introduce a watch-only account.

wallet/internal/db/itest/address_store_test.go
yyforyongyu: Kept the current shape for now to avoid wider test churn. newWatchOnlyWallet already exists and newWallet is the spendable-wallet helper, so importedAddressWallet only selects the right existing helper for each imported-address case. We can do the broader newWallet signature cleanup later if you still prefer it.

wallet/internal/db/itest/address_store_test.go
yyforyongyu: This helper was removed in the imported-address bucket refactor. Imported addresses now auto-create/reuse the reserved keyless imported bucket, so there is no separate importedAddressAccount helper to rename.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Clarified in the ADR that P2A/anchor outputs are transaction/output state, not imported address material. A spendable wallet should track the concrete anchor output through the relevant-output/watched-outpoint path; no imported-address row is created, so this invariant does not need a P2A exception. I agree address_types.7/P2A may need a separate follow-up design pass, but that is outside this PR.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: This went away with the deferred trigger removal. The ADR now explains the imported/derived split explicitly: derived rows do not carry per-row secrets because spendability comes from the wallet master key, while imported rows must carry their own secret material on spendable wallets.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Updated. The createwallet reference now points at the 31.0.0 docs, and the ADR also links the v31 importdescriptors implementation.

yyforyongyu:
LGTM 🧱
(`claude-opus-4-8[1m]`)

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: Dropped both deferred PG constraint triggers; the symmetric invariant is now enforced application-side on both SQL backends (the per-address check is the load-bearing one). The ADR's consequences section records why a DB trigger wasn't kept.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Done — the ADR references now link `ProcessDescriptorImport` (`wallet/rpc/backup.cpp` @ tag v31.0) as the source of the rejection btcwallet mirrors.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Done — the ADR now cites Bitcoin Core v23.0 (descriptor wallets becoming the default) alongside the createwallet / backup.cpp links pinned at tag v31.0.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Right — the ADR now treats P2A as transaction/output state (keyless, observed via the relevant-output / watched-outpoint path), not an imported address, so the imported-secret invariant never applies to it.

wallet/internal/db/itest/address_store_test.go
yyforyongyu: Removed — the wallet-level bucket reframing dropped this helper entirely (imports auto-create the bucket now), so the stale comment is gone.

wallet/internal/db/itest/address_store_test.go
yyforyongyu: Resolved by removal — `newImportedAccount` no longer exists; imports auto-create the keyless bucket, so the helper and its call sites were dropped.

wallet/internal/db/accountstore_createimported.go
yyforyongyu: Reworked — an imported address now lands in a keyless wallet-level bucket (auto-created on first import); per-address private-key material is required on spendable wallets via requireAddressPrivKeyOnSpendable. See the ADR's reserved-bucket section.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: The trigger is dropped; the ADR now documents that the symmetric secret requirement is imported-only (derived rows derive from the master key) and is enforced application-side on both backends.

wallet/internal/db/pg/addresses.go
```diff
@@ -264,6 +265,44 @@ func derivedAddressRowCreatedAt(
...
+	var zero sqlc.GetAccountByWalletScopeAndNameRow
+
+	scopeID, _, err := ensureKeyScope(
+		ctx, qtx, params.WalletID, params.Scope, nil,
+	)
+	if err != nil {
+		return zero, fmt.Errorf("ensure scope: %w", err)
+	}
+
+	_, err = qtx.CreateImportedAccount(
```
yyforyongyu: Creating the reserved imported-address bucket as an imported account with no account secret makes account reads report it as watch-only in spendable wallets: the existing `GetAccount`/`ListAccounts` queries still compute `is_watch_only` from `w.is_watch_only OR (origin_id = imported AND account_secrets IS NULL)`. After this path runs, `GetAccount` on the auto-created bucket returns `IsWatchOnly=true` even though the wallet and imported addresses are spendable. Make the account read-side expose the wallet-level flag for this bucket (or otherwise special-case the reserved bucket) so the new ADR 0012 contract holds. (`gpt-5.5`)

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,207 @@
...
+
+An imported address — with or without private-key material — is never
+treated as a member of a derived (HD) account. The store holds every
+imported address in a single per-scope, wallet-level **imported bucket**:
+the account named `imported` (`db.DefaultImportedAccountName`), keyed by the
+address's `(wallet_id, purpose, coin_type)` only. The bucket is internal
+infrastructure, not an imported xpub account — it is **keyless** (no
+account-level key material and no `account_secrets` row) and merely holds
+individually-imported address rows, each of which carries its own secret
+material.
```
GustavoStingelin: 👍🏼

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,207 @@
...
+This ADR therefore treats P2A as transaction/output state, not as a standard
+managed or imported address. A spendable wallet that needs to observe an anchor
+tracks the concrete output through the relevant-output / watched-outpoint path,
+not by importing an address row. No imported-address or address-secret row is
+created for the anchor, so no symmetric-invariant exception is needed: the
+invariant never triggers on a path the anchor does not travel. The child
+transaction that spends and fee-bumps the parent signs with the wallet's own
+key material on its other inputs/outputs; the keyless anchor input contributes
+no signature, so the spend flow stays safe under the uniform wallet-level
+model.
```
GustavoStingelin, line 125: 👍🏼

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,207 @@
...
+  derived row in a spendable wallet is spendable precisely because the
+  wallet holds the master key, so its absence of a per-row secret is not a
+  watch-only signal. Only an *imported* row lacking secret material
+  indicates watch-only-only material, which is what the invariant forbids on
+  a spendable wallet.
+- The invariant is enforced at imported-account / imported-address
+  **create time**, application-side and uniformly on both SQL backends (the
+  `ValidateWatchOnly` and spendable-secret checks at the store boundary).
+  Because the
+  wallet-level `is_watch_only` flag is immutable after creation (no
```
GustavoStingelin: nit, broken format

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,207 @@
...
+  historical `importpubkey` / `importaddress` flow). This ADR's uniform
+  invariant is a property of the new SQL store, not a retrofit onto
+  kvdb.
+
+## 4. References
+
+- [ADR 0006](0006-wtxmgr-sql-schema.md): wtxmgr SQL schema (related).
+- [ADR 0011](0011-no-addresses-used-column.md): addresses table omits
+  the `used` column (related — also a
+  "compute, don't persist" choice; this ADR makes the opposite call by
```
GustavoStingelin: nit. broken format

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,207 @@
...
+- [Bitcoin Core descriptor import implementation][btc-core-importdescriptors]:
+  `wallet/rpc/backup.cpp` `ProcessDescriptorImport()` is the source for
+  the rejection btcwallet mirrors — a descriptor without private keys
+  cannot be imported into a wallet with private keys enabled (and the
+  symmetric direction). The matching error string lives in the same
+  function.
+
+[btc-core-v23-release]: https://bitcoincore.org/en/releases/23.0/
+[btc-core-createwallet]: https://bitcoincore.org/en/doc/31.0.0/rpc/wallet/createwallet/
+[btc-core-importdescriptors]: https://github.com/bitcoin/bitcoin/blob/v31.0/src/wallet/rpc/backup.cpp
```
GustavoStingelin, line 210: 👍🏼

wallet/internal/db/pg/addresses.go
GustavoStingelin: 👆🏼 Same finding that I mentioned in the previous main thread review

`wallet/internal/sql/pg/queries/accounts.sql:79`
`wallet/internal/sql/pg/queries/accounts.sql:137`

wallet/internal/db/accountstore_createderived.go
```diff
@@ -188,13 +188,9 @@ func CreateDerivedAccountWithOps(ctx context.Context,
...
-		return nil, errNilAccountDerivationFunc
-	}
-
-	paramsErr := params.Validate()
-	if paramsErr != nil {
-		return nil, paramsErr
+	err := validateCreateDerivedRequest(params, deriveFn)
+	if err != nil {
+		return nil, err
 	}
```
GustavoStingelin, line 198: Maybe we can avoid the extra `validateCreateDerivedRequest` helper here.

The account-name checks can live in `params.Validate()` together with the existing empty-name validation.

That removes one level of indirection.

Likely the same for the create imported.

GustavoStingelin (changes requested):
I was reviewing and noticed that there are new commits being pushed; maybe the comments are outdated

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
```diff
@@ -0,0 +1,207 @@
...
+  (`requireAccountPrivKeyOnSpendable`) runs in `CreateImportedAccount`, the
+  path for true imported (xpub) accounts, and requires account-level secret
+  material on a spendable wallet. The keyless imported bucket does not
+  travel that path — it is materialized by a dedicated keyless insert and
+  holds no account-level secret — so it is exempt; the per-address check is
+  what enforces spendability for the addresses it holds. Derived accounts
+  and derived addresses are likewise exempt: they never store their own
+  per-row private key (their signing material is derived on demand from the
+  wallet's encrypted master key, and `account_secrets` / `address_secrets`
+  rows exist only for imported material). Requiring a per-row secret for a
```
GustavoStingelin: I think this part contradicts the previous

>the account named `imported` (`db.DefaultImportedAccountName`), keyed by the
address's `(wallet_id, purpose, coin_type)` only. The bucket is internal
infrastructure, not an imported xpub account — it is **keyless** (no
account-level key material and no `account_secrets` row) and merely holds
individually-imported address rows,

so, if the derived account will hold the secrets only in memory, and the imported ones are keyless, I guess we can drop the account_secrets table? and the key_scope secrets too?

also, derived accounts is acually being persisted with the pub and priv key in 
`wallet/internal/db/pg/accountstore_createderived.go:102`

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Reflowed locally.

wallet/internal/db/accountstore_createderived.go
yyforyongyu: Addressed locally: moved the reserved-name checks into CreateDerivedAccountParams.Validate() and CreateImportedAccountParams.Validate(), and removed the extra derived-account validation helper.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Clarified locally: the keyless imported-address bucket has no account-level key material or account_secrets row, but spendable derived accounts may persist encrypted account private-key material in account_secrets. The ADR now also states that this does not remove account_secrets or key_scope_secrets; it only removes secret presence as a public watch-only inference source.

docs/developer/adr/0012-wallet-level-watch-only-uniformity.md
yyforyongyu: Reflowed locally.

wallet/internal/db/pg/addresses.go
yyforyongyu: Addressed locally by removing the account/address-level watch-only inference rather than adding another SQL special case. Account reads now project wallets.is_watch_only as the AccountInfo.IsWatchOnly source and drop the account_secrets join; address reads/creates copy the wallet flag directly instead of inspecting private-key presence. AccountInfo.IsWatchOnly and AddressInfo.IsWatchOnly remain compatibility copies for now, and we can remove those fields in a follow-up once callers use the wallet-level flag directly.

yyforyongyu:
LGTM 🧱
(`claude-opus-4-8[1m]`)

wallet/internal/db/addresses_common.go
```diff
@@ -913,51 +945,92 @@ func NewImportedAddressWithTx[QTX any, AccountRow any, AccountParams any,
...
+		// transaction and use the row it returns. A pre-existing
+		// bucket (err == nil) is reused as-is.
+		//
+		// This get-then-create is intentionally not hardened
+		// against concurrency: imported-address creation is a
+		// sequential, low-frequency path the wallet does not drive
+		// concurrently, so two first-imports racing to create the
+		// same bucket is not a real scenario. If that changes, make
+		// the bucket insert an idempotent get-or-create.
+		if errors.Is(err, sql.ErrNoRows) {
```
yyforyongyu, line 963: This first-import path is still reachable concurrently: wallet import methods call the store directly, so two imports into the same empty wallet/scope can both observe `sql.ErrNoRows`; one creates the bucket and the other then fails on the unique `(scope_id, account_name)` insert instead of importing its address. Please make bucket materialization idempotent (for example, insert-on-conflict/do-nothing then re-read the bucket) before proceeding to the address insert. (`gpt-5.5`)

yyforyongyu: Fixed. Imported-bucket materialization is now an idempotent get-or-create: added a dedicated `CreateImportedBucketAccount` query using `INSERT ... ON CONFLICT (scope_id, account_name) DO NOTHING` (pg + sqlite), and `createImportedBucketAccount` now always re-reads via `GetAccountByWalletScopeAndName` rather than catch-duplicate-then-reread (which would poison the pg transaction). `CreateImportedAccount` stays strict, so a genuine duplicate name on a regular imported-account create still errors. Regression `TestImportedBucketMaterializationIdempotent` proves two first-imports into the same empty scope both succeed and reuse one bucket (passes on sqlite + postgres). Commits: 58ef751e1 (query), 54e81786a (materialization + test).

yyforyongyu:
LGTM, no correctness issues found in the latest revision. (`gpt-5.5`)

GustavoStingelin (approved):
LGTM!

Part 3 of the ADR seems to go a bit deeper into the implementation details, but I think that is fine.

My agents also raised the comments below, but I am treating them as optional notes rather than strong requirements for this PR:

1. Security noted a low severity issue: account rename can still rename an account to `imported`. This does not bypass the private key or watch only checks, but if strict ownership of reserved bucket names is intended, we could add `requireUnreservedAccountName(params.NewName)` to the rename validation.

2. External parity research found that the ADR’s `getbalances.watchonly` wording is stale for current Bitcoin Core, since Core 30 and later removed the watchonly bucket.

yyforyongyu:
> LGTM!
> 
> Part 3 of the ADR seems to go a bit deeper into the implementation details, but I think that is fine.
> 
> My agents also raised the comments below, but I am treating them as optional notes rather than strong requirements for this PR:
> 
>     1. Security noted a low severity issue: account rename can still rename an account to `imported`. This does not bypass the private key or watch only checks, but if strict ownership of reserved bucket names is intended, we could add `requireUnreservedAccountName(params.NewName)` to the rename validation.
> 
>     2. External parity research found that the ADR’s `getbalances.watchonly` wording is stale for current Bitcoin Core, since Core 30 and later removed the watchonly bucket.

cool address in a new commit, i think the first item is not valid tho,

> The current ADR already says modern Bitcoin Core descriptor wallets report balances under mine, and only deprecated pre-descriptor watch-only wallets used watchonly. Bitcoin Core 31 getbalances docs also show only mine.

yyforyongyu:
Re-reviewed a88be3f6; LGTM, no correctness issues found in the latest revision. (`gpt-5.5`)

GustavoStingelin (approved):

PR #1230, opened 2026-05-14
wallet: migrate tx creator + tx publisher to db.Store

yyforyongyu:
## Summary

Complete migration of the wallet tx creator and tx publisher subsystems
to `db.Store`. Implements `CreateTx` + `InvalidateUnminedTx` on the
kvdb backend, routes change-source + signer paths through the store,
and brings the publisher's record/cleanup flows under store control.

## Why

- finishes the tx-creator + tx-publisher portion of the kvdb → db.Store
  migration in one reviewable chain
- mirrors the kvdb-impl + wallet-route interleave used on PR #1214
  (impl-address-manager-store)
- the previously-planned standalone publisher branch is folded here
  for review consolidation

## What Changed

- `wallet/signer.go`, `wallet/tx_creator.go`: keep signer lookups
  wallet-side; route change source through the store
- `wallet/deprecated_tx_creator.go`: relocated legacy change-source
  helper
- `wallet/internal/db/kvdb/txstore.go`: implement `CreateTx`,
  `InvalidateUnminedTx` (each commit bundles its kvdb test)
- `wallet/internal/db/kvdb/utils_test.go`: extended test helpers for
  `CreateTx` tests
- `wallet/tx_publisher.go`: route tx recording, owned-address
  filtering, and invalid-tx cleanup through `w.store`
- `wallet/tx_publisher_test.go`: store-backed matchers + tests

## Testing

- per-commit `GOWORK=off go build ./...` clean across all 10 commits
- `make lint`: 0 issues
- unit + itest CI: pending (force-push triggers a fresh run)

PR #1248, opened 2026-05-28
wallet: route tx publisher + tx-store writes through Store

yyforyongyu:
## Summary

Route the wallet's tx publisher path through `db.Store`: kvdb
implements `CreateTx` and `InvalidateUnminedTx`, and wallet code
records transactions, filters owned addresses, and cleans up
invalid txs through the store boundary instead of legacy
wtxmgr calls. Migrates the kvdb test fixtures from the legacy
`testManagedAddress` type to the consolidated
`bwmock.ManagedAddress` so downstream branches can reuse the
shared mock surface.

Stack position: **#6** of the kvdb-migration stack. Split out
from #1230 per
[Task 145](.lgtm/tasks/145-split-impl-tx-creator-store.md);
stacks on #1230 because the downstream chain (PSBT + runtime)
needs both PR 1's tx-creator surface AND PR 2's
`bwmock.ManagedAddress` migration to compile.

## Changes

- `kvdb.Store.CreateTx` — records a transaction record + credits
  in one walletdb update; new test helpers extend the legacy
  txstore fixture for the new write path.
- `kvdb.Store.InvalidateUnminedTx` — removes one unmined tx via
  the legacy txstore.
- `Wallet.ProcessTransaction` (tx recording), owned-address
  filtering, and invalid-tx cleanup route through `db.Store`
  instead of direct legacy txstore reads.
- New `tx publisher store matchers` test helpers + mock
  expectations.
- `kvdb` tests migrate from `testManagedAddress` to
  `bwmock.ManagedAddress` (the consolidated mock from Task 119)
  so downstream branches inherit a single canonical
  ManagedAddress mock surface.

## Test plan

- [ ] `GOWORK=off go test ./wallet/internal/db/kvdb -run 'Test(CreateTx|InvalidateUnminedTx)'`
- [ ] `GOWORK=off go test ./wallet -run 'Test.*TxPublisher|Test.*Record|Test.*Invalid'`
- [ ] `GOWORK=off go test ./wallet/internal/db/kvdb ./wallet`

### DIFF
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -1019,15 +1019,33 @@ type CreateTxParams struct {
 	// Label is an optional label for the transaction.
 	Label string
 
-	// Credits maps wallet-owned output indexes to their display addresses.
+	// Credits maps wallet-owned output indexes to the resolved address that
+	// makes the output ours.
 	//
 	// The output index is the map key, so duplicate credited outputs are
 	// impossible by construction.
 	//
-	// NOTE: The address value is for display only. The database layer still
-	// matches ownership by the output's script_pub_key
-	// (`params.Tx.TxOut[index].PkScript`), which is the canonical key
-	// used by the address schema.
+	// A non-nil address is authoritative for ownership: the store resolves the
+	// owning address row by that address's own script (PayToAddrScript),
+	// not by the full output script. This lets a bare-multisig output the
+	// wallet partly owns resolve to the wallet-owned member, whose script the
+	// multisig output script itself would never match. For a single-address
+	// output the member script equals the output script, so the lookup is
+	// unchanged.
+	//
+	// Because that address is trusted as the owner, the store first validates
+	// that the credited output actually pays it. Membership (not strict script
+	// equality) is required so bare-multisig members pass, while an address the
+	// output does not pay is rejected with ErrInvalidParam rather than
+	// recording a UTXO owned by an unrelated address.
+	//
+	// A nil address means the caller has no resolved owner; the store then
+	// keys ownership on the output's own script_pub_key
+	// (`params.Tx.TxOut[index].PkScript`).
+	//
+	// NOTE: This only selects the address-ownership lookup key. The stored
+	// UTXO always records the on-chain output script, never the member
+	// script.
 	Credits map[uint32]btcutil.Address
 }
 
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -791,6 +791,22 @@ type GetAddressQuery struct {
 	ScriptPubKey []byte
 }
 
+// ResolveOwnedAddressesQuery contains the parameters for a batched,
+// wallet-scoped address ownership lookup. It resolves which of the supplied
+// script pubkeys belong to the wallet in a single store operation.
+type ResolveOwnedAddressesQuery struct {
+	// WalletID is the ID of the wallet to query.
+	//
+	// NOTE: uint32 is used to ensure compatibility with standard SQL
+	// databases (signed 64-bit integers).
+	WalletID uint32
+
+	// ScriptPubKeys is the set of script pubkeys to resolve. Order is not
+	// significant and duplicates are tolerated. An empty or nil slice
+	// yields an empty result without touching the backend.
+	ScriptPubKeys [][]byte
+}
+
 // GetAddressSecretQuery contains the parameters for querying an address
 // secret. The query is wallet-scoped: it retrieves the encrypted secret
 // material for a specific address within a wallet.
--- a/wallet/internal/db/interface.go
+++ b/wallet/internal/db/interface.go
@@ -327,6 +327,23 @@ type AddressStore interface {
 	// an error if the address is not found.
 	GetAddress(ctx context.Context, query GetAddressQuery) (*AddressInfo, error)
 
+	// ResolveOwnedAddresses resolves a batch of script pubkeys to the subset
+	// that is owned by the wallet in a single store operation. It is intended
+	// to replace a per-script GetAddress loop on hot paths such as transaction-
+	// output ownership filtering.
+	//
+	// The result is keyed by string(ScriptPubKey). Only wallet-owned scripts
+	// appear in the map: a script absent from the result is simply not owned
+	// by the wallet (the batched equivalent of GetAddress returning
+	// ErrAddressNotFound), which is not an error. Both the wallet ID and the
+	// supplied script set constrain the lookup.
+	//
+	// An empty or nil ScriptPubKeys slice returns an empty, non-nil map
+	// without issuing a backend query or returning an error.
+	ResolveOwnedAddresses(ctx context.Context,
+		query ResolveOwnedAddressesQuery) (
+		map[string]*AddressInfo, error)
+
 	// ListAddresses returns one page of addresses for the given query,
 	// including a next-cursor for the following page.
 	ListAddresses(ctx context.Context, query ListAddressesQuery) (
--- a/wallet/internal/db/itest/tx_utxo_store_test.go
+++ b/wallet/internal/db/itest/tx_utxo_store_test.go
@@ -104,7 +104,6 @@ func TestCreateTxCreditBareMultisigMember(t *testing.T) {
 
 	scope := db.KeyScopeBIP0084
 	importedName := db.DefaultImportedAccountName
-	CreateImportedAccount(t, store, walletID, scope, importedName, true)
 
 	// Build a 1-of-2 bare-multisig script. memberScript is the member's own
 	// P2PK script (what PayToAddrScript(memberAddr) yields), which the wallet
--- a/wallet/internal/db/itest/tx_utxo_store_test.go
+++ b/wallet/internal/db/itest/tx_utxo_store_test.go
@@ -6,8 +6,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec/v2"
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg"
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
+	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	"github.com/stretchr/testify/require"
@@ -58,6 +61,161 @@ func TestCreateTxStoresWalletCredit(t *testing.T) {
 	_, ok := txIDByHash(t, store, walletID, tx.TxHash())
 	require.True(t, ok)
 	require.True(t, walletUtxoExists(t, store, walletID, wire.OutPoint{
+		Hash: tx.TxHash(), Index: 0,
+	}))
+}
+
+// TestCreateTxCreditBareMultisigMember verifies that CreateTx records a
+// bare-multisig output that the wallet partly owns: the wallet holds one
+// member pubkey address, but not the full multisig output script.
+//
+// This is the SQL counterpart to the kvdb TestCreateTxCreditBareMultisigMember
+// and locks in parity between the two backends. The publisher's ownership
+// filter resolves such an output to the wallet-owned member address and passes
+// it as the credit address; the store must therefore key ownership on the
+// member's own script, not on the full multisig output script that no address
+// row would ever match.
+//
+// Scenario:
+//   - The wallet imports one P2PK member address (script-only on a watch-only
+//     wallet, which the ADR 0012 invariant allows).
+//   - A transaction pays a 1-of-2 bare-multisig output built from that member
+//     plus a foreign key. The full multisig script is NOT a wallet address.
+//
+// Action:
+//   - Insert the transaction through CreateTx with Credits[0] set to the
+//     member address.
+//
+// Assertions:
+//   - CreateTx succeeds (no ErrAddressNotFound) and the credited output exists.
+//   - GetUtxo resolves the UTXO to the imported member address (imported
+//     origin, raw-pubkey type) rather than the multisig output script.
+//
+// Without the fix, the store looks the credit up by the full multisig output
+// script, finds no address row, and CreateTx fails with ErrAddressNotFound.
+func TestCreateTxCreditBareMultisigMember(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+
+	// A script-only import (no private key) requires a watch-only wallet per
+	// the ADR 0012 spendable-wallet invariant.
+	walletID := newWatchOnlyWallet(t, store, "wallet-bare-multisig-credit")
+
+	scope := db.KeyScopeBIP0084
+	importedName := db.DefaultImportedAccountName
+	CreateImportedAccount(t, store, walletID, scope, importedName, true)
+
+	// Build a 1-of-2 bare-multisig script. memberScript is the member's own
+	// P2PK script (what PayToAddrScript(memberAddr) yields), which the wallet
+	// imports as an address. multiSigScript is the full on-chain output script,
+	// which is deliberately never registered as a wallet address.
+	memberAddr, memberScript, multiSigScript := newMultisigScript(t)
+	require.NotEqual(t, memberScript, multiSigScript)
+
+	_, err := store.NewImportedAddress(
+		t.Context(), db.NewImportedAddressParams{
+			WalletID:        walletID,
+			Scope:           scope,
+			AddressType:     db.RawPubKey,
+			PubKey:          memberAddr.ScriptAddress(),
+			ScriptPubKey:    memberScript,
+			EncryptedScript: RandomBytes(48),
+		},
+	)
+	require.NoError(t, err)
+
+	tx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 7000, PkScript: multiSigScript}},
+	)
+
+	// Credits[0] carries the resolved member address, exactly as the publisher
+	// supplies it after filtering ownership by the member script.
+	err = store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       tx,
+		Received: time.Unix(1710004000, 0),
+		Status:   db.TxStatusPending,
+		Credits:  map[uint32]btcutil.Address{0: memberAddr},
+	})
+	require.NoError(t, err)
+
+	outPoint := wire.OutPoint{Hash: tx.TxHash(), Index: 0}
+	require.True(t, walletUtxoExists(t, store, walletID, outPoint))
+
+	// The credit must resolve to the imported member address, not the multisig
+	// output script. Origin and address type prove the resolved address row is
+	// the member import.
+	utxo, err := store.GetUtxo(t.Context(), db.GetUtxoQuery{
+		WalletID: walletID,
+		OutPoint: outPoint,
+	})
+	require.NoError(t, err)
+	require.Equal(t, importedName, utxo.AccountName)
+	require.Equal(t, db.ImportedAccount, utxo.Origin)
+	require.Equal(t, db.RawPubKey, utxo.AddrType)
+}
+
+// TestCreateTxCreditAddrMismatch verifies that CreateTx rejects a non-nil
+// credit address that the credited output does not pay to, instead of
+// recording a UTXO owned by an unrelated address.
+//
+// A non-nil Credits[index] is authoritative for ownership, so the store must
+// confirm the output actually pays that address before trusting it. This is
+// the SQL counterpart to the kvdb TestCreateTxCreditAddrMismatch and keeps the
+// two backends consistent: kvdb validates membership via its own
+// validateCreditAddr, and the SQL backends validate it through the shared
+// ValidateCreditAddrMembership before the credit lookup.
+//
+// Scenario:
+//   - The output pays a wallet-owned derived address.
+//   - Credits[0] instead names an unrelated address the output never pays.
+//
+// Assertions:
+//   - CreateTx fails with ErrInvalidParam and records nothing.
+func TestCreateTxCreditAddrMismatch(t *testing.T) {
+	t.Parallel()
+
+	store := NewTestStore(t)
+	walletID := newWallet(t, store, "wallet-credit-addr-mismatch")
+	createDerivedAccount(t, store, walletID, db.KeyScopeBIP0084, "default")
+
+	// The output pays this wallet-owned address.
+	paidAddr := newDerivedAddress(
+		t, store, walletID, db.KeyScopeBIP0084, "default", false,
+	)
+
+	// The credit names a different address that the output script does not
+	// pay to, simulating a caller mislabeling the credited output.
+	wrongKey, err := btcec.NewPrivateKey()
+	require.NoError(t, err)
+
+	wrongAddr, err := btcutil.NewAddressPubKey(
+		wrongKey.PubKey().SerializeCompressed(),
+		&chaincfg.RegressionNetParams,
+	)
+	require.NoError(t, err)
+
+	tx := newRegularTx(
+		[]wire.OutPoint{randomOutPoint()},
+		[]*wire.TxOut{{Value: 6000, PkScript: paidAddr.ScriptPubKey}},
+	)
+
+	err = store.CreateTx(t.Context(), db.CreateTxParams{
+		WalletID: walletID,
+		Tx:       tx,
+		Received: time.Unix(1710004200, 0),
+		Status:   db.TxStatusPending,
+		Credits:  map[uint32]btcutil.Address{0: wrongAddr},
+	})
+	require.ErrorIs(t, err, db.ErrInvalidParam)
+
+	// The rejected insert must not have created the transaction row or a
+	// UTXO.
+	_, ok := txIDByHash(t, store, walletID, tx.TxHash())
+	require.False(t, ok)
+	require.False(t, walletUtxoExists(t, store, walletID, wire.OutPoint{
 		Hash: tx.TxHash(), Index: 0,
 	}))
 }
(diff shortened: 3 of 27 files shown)


yyforyongyu (commented):
LGTM 🦜 (`deepseek-v4-pro`)

wallet/internal/db/kvdb/utils_test.go
```diff
@@ -306,3 +317,35 @@ func newTestAddressScript(t *testing.T) (btcutil.Address, []byte) {
 
 	return addr, pkScript
 }
+
+// testManagedAddress is a minimal managed address used by kvdb store tests.
+type testManagedAddress struct {
```
yyforyongyu: `testManagedAddress` (32 lines, defined here in `6108c6893`) is fully removed by the last commit of this same stack (`72584bda3 kvdb: replace testManagedAddress with bwmock.ManagedAddress`) in favor of `bwmock.ManagedAddress`. `bwmock.ManagedAddress` already exists in the base branch (`bwtest/mock/managed_address.go` at `impl-wallet-metadata-store`), so the hand-rolled fixture is not bridging a missing dependency — it's a stack-local detour.

This is the introduce-then-remove pattern that `feedback_no_introduce_then_remove.md` rejects: "don't add a type only to delete it in a later commit of the same stack; build toward the end-state shape directly."

Fix: land `kvdb: extend test helpers for CreateTx tests` and `kvdb: implement CreateTx` (the consumers in `6108c6893` + `0da47dfac`) directly against `bwmock.ManagedAddress`, and drop the dedicated `kvdb: replace testManagedAddress with bwmock.ManagedAddress` commit from the stack entirely. Same end state, one fewer commit, no introduce-then-remove churn.

(`claude-opus-4-7[1m]`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -6,18 +6,123 @@ import (
...
+	)
+}
+
+// addCreateTxCredits records wallet-owned outputs for a legacy transaction.
+func (s *Store) addCreateTxCredits(addrmgrNs,
+	txmgrNs walletdb.ReadWriteBucket, txRec *wtxmgr.TxRecord,
+	credits map[uint32]btcutil.Address) error {
+
+	for index, addr := range credits {
+		managedAddr, err := s.addrStore.Address(addrmgrNs, addr)
```
yyforyongyu: `CreateTxParams.Credits` documents the address value as display-only and says ownership is keyed by the credited output script, but this path trusts the supplied address when adding the legacy credit. A caller can therefore credit output index N by passing any wallet address even when `TxOut[N].PkScript` pays somewhere else (or nil), diverging from SQL and corrupting kvdb UTXO ownership. Resolve the managed address from the credited output script, or validate the supplied address matches that script, before calling `AddCredit`. (`gpt-5.5`)

wallet/tx_publisher.go
```diff
@@ -417,54 +390,51 @@ func (w *Wallet) extractTxAddrs(tx *wire.MsgTx) map[uint32][]btcutil.Address {
...
-				err, waddrmgr.ErrAddressNotFound) {
+			info.outputIndices = append(info.outputIndices, index)
+			uniqueAddrs[key] = info
+		}
+	}
 
-				continue
-			}
+	for key, info := range uniqueAddrs {
+		_, err := w.store.GetAddress(ctx, db.GetAddressQuery{
```
yyforyongyu: Switching the ownership check from the extracted address to the whole pkScript drops multi-address scripts. For bare multisig, `ExtractPkScriptAddrs` can return wallet-owned pubkeys, but `GetAddress` by the multisig script will not match those individual address rows, so the transaction is no longer recorded or marked used. Keep resolving each extracted address (or add a Store lookup by address) and use the script only for the later credit write. (`gpt-5.5`)

wallet/tx_publisher.go
```diff
@@ -142,17 +140,17 @@ func (w *Wallet) Broadcast(ctx context.Context, tx *wire.MsgTx,
 	txid := tx.TxHash()
 	log.Errorf("%v: broadcast failed: %v", txid, err)
 
-	// If the tx was rejected for any other reason, then we'll remove it
+	// If the tx was rejected for any other reason, then we'll invalidate it
 	// from the tx store, as otherwise, we'll attempt to continually
 	// re-broadcast it, and the UTXO state of the wallet won't be accurate.
-	removeErr := w.removeUnminedTx(tx)
+	removeErr := w.invalidateUnminedTx(ctx, tx)
```
yyforyongyu, line 158: When `addTxToWallet` finds no wallet-owned outputs, it returns without creating a tx row. A later publish failure now still calls `InvalidateUnminedTx`, so sweep-style transactions with no change turn the original broadcast rejection into an additional `ErrTxNotFound` cleanup failure. Track whether `CreateTx` actually ran, or skip invalidation when nothing was recorded. (`gpt-5.5`)

wallet/tx_publisher.go
```diff
@@ -126,13 +124,13 @@ func (w *Wallet) Broadcast(ctx context.Context, tx *wire.MsgTx,
 	// allow us to track the tx's confirmation status, and also
 	// re-broadcast it upon startup. If any of the subsequent steps fail,
 	// this tx must be removed.
```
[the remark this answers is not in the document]
yyforyongyu: Done — the Broadcast comment now reads "this tx is invalidated via InvalidateUnminedTx", matching the new flow.

wallet/internal/db/kvdb/txstore.go
```diff
@@ -6,18 +6,123 @@ import (
...
+
+	addrmgrNs := tx.ReadWriteBucket(waddrmgr.NamespaceKey)
+	if addrmgrNs == nil {
+		return errMissingAddrmgrNamespace
+	}
+
+	txmgrNs := tx.ReadWriteBucket(wtxmgrNamespaceKey)
+	if txmgrNs == nil {
+		return errMissingTxmgrNamespace
+	}
```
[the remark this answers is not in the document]
yyforyongyu: Done — createTxWithTx only requires/opens the waddrmgr namespace when len(params.Credits) > 0, so a credit-less (sweep) tx can be recorded without the address-manager bucket.

wallet/internal/db/kvdb/utils_test.go
yyforyongyu: Done — the testManagedAddress add + replace commits are dropped entirely; the helper is never introduced (kvdb tests use bwmock.ManagedAddress), so there's no introduce-then-remove.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Done — addCreateTxCredits now validates via validateCreditAddr: it runs ExtractPkScriptAddrs on TxOut[index].PkScript and requires the supplied credit address to be one of the extracted addresses (membership, so bare multisig still works), returning db.ErrInvalidParam on mismatch — a caller can no longer credit an output with an unrelated address.

wallet/tx_publisher.go
yyforyongyu: Done — filterOwnedAddresses now resolves each extracted address by its own script (txscript.PayToAddrScript(addr)) rather than the whole output pkScript, so a bare-multisig output where the wallet owns a member pubkey is recognized and recorded.

wallet/tx_publisher.go
yyforyongyu: Done — addTxToWallet now records debit-only sweeps: when there are no wallet-owned outputs it checks the inputs (spendsWalletOutput → GetUtxo per TxIn) and records the tx (with empty credits) if it spends a wallet UTXO, so InvalidateUnminedTx on a publish failure operates on a recorded row instead of hitting ErrTxNotFound. Covered by TestBroadcastSweepPublishFailsInvalidateSucceeds.

wallet/tx_publisher.go
```diff
@@ -417,54 +390,51 @@ func (w *Wallet) extractTxAddrs(tx *wire.MsgTx) map[uint32][]btcutil.Address {
...
+		}
+	}
 
-				continue
-			}
+	for key, info := range uniqueAddrs {
+		_, err := w.store.GetAddress(ctx, db.GetAddressQuery{
+			WalletID:     w.id,
+			ScriptPubKey: info.scriptPubKey,
+		})
```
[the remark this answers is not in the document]
yyforyongyu: Deferred (perf, not correctness): keeping the per-call GetAddress for now. A batched GetAddresses API can't land here — #1248 is downstream of #1230, which wants the same API, so it must land upstream (prep-utxo-enrichment or earlier, where AddressStore is already extended). The correctness fixes (credit-script validation + multisig ownership) take priority. Leaving open as a tracked follow-up rather than resolving.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🧹
(`claude-opus-4-8[1m]`)

wallet/tx_publisher.go
yyforyongyu: This is still reachable when the transaction is completely wallet-unrelated: `addTxToWallet` returns `nil, nil` after `spendsWalletOutput` returns false, so a later publish failure still calls `InvalidateUnminedTx` for a row that was never created and replaces the broadcast error with `ErrTxNotFound`. Return a recorded bool from `addTxToWallet` and only invalidate when it is true. (`gpt-5.5`)

yyforyongyu: Fixed — `addTxToWallet` now returns a `recorded` bool and `Broadcast` only calls `invalidateUnminedTx` when `recorded == true`. A wallet-unrelated tx (never recorded) whose publish fails now preserves the original broadcast error instead of turning it into an `ErrTxNotFound` cleanup failure. Regression test `TestBroadcastUnrelatedPublishFailsNoInvalidate` added. Resolving.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🌊
(`claude-opus-4-8[1m]`)

wallet/tx_publisher.go
```diff
@@ -417,57 +424,91 @@ func (w *Wallet) extractTxAddrs(tx *wire.MsgTx) map[uint32][]btcutil.Address {
...
+		// resolves the wallet-owned member, which the multisig script
+		// itself would never match.
+		ownScript, err := txscript.PayToAddrScript(info.addr)
+		if err != nil {
+			return nil, err
+		}
 
-		for addr, indices := range uniqueAddrs {
-			ma, err := w.addrStore.Address(addrmgrNs, addr)
+		_, err = w.store.GetAddress(ctx, db.GetAddressQuery{
```
yyforyongyu: Resolving a bare-multisig member here is not enough for SQL-backed wallets: `recordTxAndCredits` only passes the output index, and SQL `CreateTx` inserts the credit by `TxOut[index].PkScript`, not by this member script. That means this can classify the tx as ours, then fail with `ErrAddressNotFound` before broadcasting. Either require the full output script to be representable by the Store, or carry the resolved credit target through the SQL credit write. (`gpt-5.5`)

yyforyongyu: Fixed — SQL `insertCredits` now resolves the credit address via the wallet-owned member (`Credits[index]`) instead of the full output `PkScript`, so a bare-multisig output owned via a member records on SQL rather than failing with `ErrAddressNotFound`; the nil-credit fallback (output-script lookup) is preserved. Regression test added on pg+sqlite. Pushed + CI-green. Resolving.

wallet/tx_publisher.go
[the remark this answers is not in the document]
yyforyongyu: Done — added a batched `GetAddresses` to `db.Store` (kvdb one `walletdb.View`; pg `= ANY`; sqlite `IN`); `filterOwnedAddresses` now makes a single store call instead of one `GetAddress` per output address (benchmark: store_ops/op = 1 across N=1…1024 outputs). Pushed + CI-green. Resolving.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪼
(`claude-opus-4-8[1m]`)

wallet/tx_publisher.go
```diff
@@ -304,83 +330,64 @@ func (w *Wallet) addTxToWallet(tx *wire.MsgTx,
...
-			txHash := txRec.MsgTx.TxHash()
-
-			err := w.txStore.PutTxLabel(txmgrNs, txHash, label)
-			if err != nil {
-				return err
-			}
-		}
+	credits := make(map[uint32]btcutil.Address, len(creditsToWrite))
+	for _, credit := range creditsToWrite {
+		credits[credit.index] = credit.addr
```
yyforyongyu: `creditsToWrite` can contain multiple wallet-owned addresses for the same output, for example a bare multisig script where two member pubkeys are ours. Collapsing it to `map[output]address` overwrites all but one address with map-iteration-dependent choice, so `CreateTx` records/marks only one owned member as used and the other wallet address can remain reusable. Preserve the full owned-address set for usage marking, or explicitly reject/canonicalize multi-owned scripts before this overwrite. (`gpt-5.5`)

wallet/tx_publisher.go
```diff
@@ -417,59 +424,108 @@ func (w *Wallet) extractTxAddrs(tx *wire.MsgTx) map[uint32][]btcutil.Address {
...
 
+// spendsWalletOutput reports whether the transaction spends at least one output
+// owned by the wallet. It asks the Store for each previous outpoint and treats
+// a hit as wallet relevance, so a sweep that pays no wallet-owned outputs is
+// still recognized as ours.
+func (w *Wallet) spendsWalletOutput(ctx context.Context,
+	tx *wire.MsgTx) (bool, error) {
+
+	for _, txIn := range tx.TxIn {
+		_, err := w.store.GetUtxo(ctx, db.GetUtxoQuery{
```
yyforyongyu, line 606: `GetUtxo` only sees the current UTXO set, so a no-change transaction that spends a wallet output already attached to another unmined wallet tx gets `ErrUtxoNotFound` here. `addTxToWallet` then treats it as wallet-unrelated and can broadcast it without recording the tx or hitting the store-level input-conflict path. Use a wallet-owned-outpoint lookup that includes already-spent unmined parents, or record the tx and let `CreateTx`/`MarkInputsSpent` decide the conflict. (`gpt-5.5`)

wallet/internal/db/data_types.go
```diff
@@ -1019,15 +1036,27 @@ type CreateTxParams struct {
...
+	// makes the output ours.
 	//
 	// The output index is the map key, so duplicate credited outputs are
 	// impossible by construction.
 	//
-	// NOTE: The address value is for display only. The database layer still
-	// matches ownership by the output's script_pub_key
-	// (`params.Tx.TxOut[index].PkScript`), which is the canonical key
-	// used by the address schema.
+	// A non-nil address is authoritative for ownership: the store resolves the
```
yyforyongyu, line 1044: Making a non-nil credit address authoritative still leaves the SQL backends without the membership check kvdb added: `creditLookupScript` resolves this address row, then `InsertUtxo` stores `TxOut[index]` even if that output does not pay to the address. A bad caller can therefore corrupt SQL UTXO ownership by passing any wallet address for an unrelated output. Validate each non-nil credit address against `ExtractPkScriptAddrs(TxOut[index].PkScript)` in shared request validation or before the SQL insert. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -6,18 +6,181 @@ import (
...
+
+	_, addrs, _, err := txscript.ExtractPkScriptAddrs(
+		msgTx.TxOut[index].PkScript, chainParams,
+	)
+	if err != nil {
+		return fmt.Errorf("credit output %d: extract script addrs: %w",
+			index, err)
+	}
+
+	want := addr.EncodeAddress()
```
yyforyongyu, line 230: `CreateTxParams.Credits` allows a nil address to mean "use the output script itself", and the SQL backends implement that fallback, but kvdb reaches `addr.EncodeAddress()` here for every credit. A caller using the documented nil-credit path panics instead of recording the credit (or returning a store error). Handle nil before this membership check by resolving the output script path, or reject nil credits in shared validation if kvdb cannot support them. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -6,18 +6,181 @@ import (
...
+
 // errLegacyHeightOverflow reports that one db height cannot fit into the
 // signed legacy wtxmgr height domain.
 var errLegacyHeightOverflow = errors.New("legacy height overflows int32")
 
-// CreateTx is not yet implemented for kvdb.
-func (s *Store) CreateTx(ctx context.Context, _ db.CreateTxParams) error {
-	return notImplemented(ctx, "CreateTx")
+// CreateTx records an unmined transaction through the legacy wtxmgr path.
+func (s *Store) CreateTx(ctx context.Context,
```
yyforyongyu, line 43: we should implement the `CreateTxOps`?

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪼
(`claude-opus-4-8[1m]`)

wallet/internal/db/data_types.go
yyforyongyu: The new helper still over-accepts the bare-multisig case: `scriptPushesData` succeeds for any script that happens to push the member pubkey, not only scripts that actually pay via bare multisig. A caller could mark an `OP_RETURN <pubkey>` or other non-paying script as a wallet credit and `InsertUtxo` would store that output as ours. Please verify the output script is a bare multisig shape before accepting the pushed-pubkey fallback. (`gpt-5.5`)

wallet/tx_publisher.go
```diff
@@ -304,83 +331,81 @@ func (w *Wallet) addTxToWallet(tx *wire.MsgTx,
...
+	})
+	if err == nil {
+		return nil
+	}
 
-		// If the tx has already been recorded, we can return early.
-		if exists {
-			return nil
-		}
+	if !errors.Is(err, db.ErrTxAlreadyExists) {
```
yyforyongyu, line 404: Treating every `ErrTxAlreadyExists` as success is unsafe now that failed/replaced rows are retained. If a prior `Broadcast` invalidated this tx after a transient send failure, a retry reaches this branch, then can publish successfully while the only stored row remains `TxStatusFailed` or `TxStatusReplaced`, so it is not tracked/rebroadcast and the spend edges stay restored. Please only ignore duplicates that are still live, or revive/delete/reinsert retained invalid rows before publishing. (`gpt-5.5`)

wallet/internal/db/data_types.go
yyforyongyu: Done — SQL credit insertion now validates non-nil credit membership before recording the UTXO.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: kvdb `CreateTx` keeps its bespoke unmined-only path with a documented divergence note; the shared `CreateTxWithOps` confirmed/conflict stages do not map to the legacy wtxmgr model.

wallet/tx_publisher.go
yyforyongyu: Done — `recordTxAndCredits` now inspects the existing row on `ErrTxAlreadyExists`: a live row stays idempotent, but a retained failed/replaced row returns the new `ErrTxRetainedInvalid` before publishing, so a retry cannot broadcast while the stored row is invalid.

yyforyongyu:
LGTM 🪼
(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🐚 (`gpt-5.5`)

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪼
(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🦦 (`gpt-5.5`)

yyforyongyu:
LGTM 🪼
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -6,18 +6,240 @@ import (
...
+	req, err := db.NewCreateTxRequest(params)
+	if err != nil {
+		return fmt.Errorf("create tx request: %w", err)
+	}
+
+	if req.Params.Block != nil {
+		return notImplemented(ctx, "CreateTx confirmed")
+	}
+
+	if len(req.Params.Credits) > 0 && s.addrStore == nil {
```
yyforyongyu: `s.addStore` is always present, so there is no need for this check!

wallet/tx_publisher.go
```diff
@@ -18,11 +18,12 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcutil"
+	"github.com/btcsuite/btcd/chaincfg/chainhash"
```
yyforyongyu, line 21: again the commit body in `wallet: route tx recording and ownership filtering through store` is too verbose - keep it short

wallet/tx_publisher.go
```diff
@@ -166,6 +167,13 @@ var (
 	// ErrTxCannotBeNil is returned when a nil transaction is passed to a
 	// function.
 	ErrTxCannotBeNil = errors.New("tx cannot be nil")
+
+	// ErrTxRetainedInvalid is returned when a tx cannot be recorded for
```
yyforyongyu, line 180: we should discuss this:
1. what is the current behavior before this change?
2. should we keep it as a noop if the tx is already found? what happens if this is just a random tx that neither consumes nor produces wallet utxos?

wallet/internal/db/itest/tx_utxo_store_test.go
```diff
@@ -6,8 +6,11 @@ import (
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/btcec/v2"
```
yyforyongyu, line 9: again rework the commit msg body `db/sql: credit by resolved owner address for bare-multisig outputs` - keep them short

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -249,6 +249,92 @@ func (s *Store) GetAddress(ctx context.Context,
 	return info, nil
 }
 
+// GetAddresses resolves a batch of script pubkeys to the subset owned by the
+// wallet using a single read-only walletdb transaction. It restores the
+// single-transaction pattern for output ownership filtering, replacing a
+// per-script GetAddress loop.
+func (s *Store) GetAddresses(ctx context.Context,
```
yyforyongyu: how is this different from `ListAddresses`?

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Removed — the Store/addrStore is always present in production, so this impossible branch (and its test) is gone.

wallet/tx_publisher.go
yyforyongyu: Tightened the commit body.

wallet/internal/db/itest/tx_utxo_store_test.go
yyforyongyu: Tightened the commit body.

wallet/tx_publisher.go
yyforyongyu: Addressed in the record-before-publish design:

- Irrelevant tx (no owned outputs and does not spend a wallet output): `addTxToWallet` returns recorded=false and writes nothing, so the caller does not invalidate a tx that was never stored. A tx that spends a wallet output but credits none is recorded with an empty credit set so it stays tracked.
- Already-recorded tx: `handleExistingTx` reconciles the duplicate against the stored row (not a blind noop) — a retained-invalid row (Failed/Replaced) returns `ErrTxRetainedInvalid` instead of a false success; a still-live row is idempotent; a pending duplicate is promoted to published.

wallet/internal/db/kvdb/addressstore.go
yyforyongyu: Addressed — `GetAddresses` was renamed to `ResolveOwnedAddresses`, which makes the distinction explicit: it resolves a batch of script pubkeys to the wallet-owned subset (output-ownership filtering for tx recording), whereas `ListAddresses` paginates the stored addresses by query.

yyforyongyu:
LGTM 🪼
(`claude-opus-4-8[1m]`)

wallet/tx_publisher.go
```diff
@@ -470,6 +472,105 @@ func (w *Wallet) filterOwnedAddresses(
...
+func (w *Wallet) walletOwnedOutputs(ctx context.Context,
+	txHash chainhash.Hash,
+	cache map[chainhash.Hash]map[uint32]struct{}) (
+	map[uint32]struct{}, error) {
+
+	if owned, ok := cache[txHash]; ok {
+		return owned, nil
+	}
+
+	detail, err := w.store.GetTxDetail(ctx, db.GetTxDetailQuery{
```
yyforyongyu, line 651: should be in newline?

wallet/tx_publisher.go
```diff
@@ -142,17 +145,24 @@ func (w *Wallet) Broadcast(ctx context.Context, tx *wire.MsgTx,
...
+	// rather than overwriting it with cleanup context.
+	if !recorded {
+		return err
+	}
+
+	// If the tx was rejected for any other reason, then we'll invalidate it
 	// from the tx store, as otherwise, we'll attempt to continually
 	// re-broadcast it, and the UTXO state of the wallet won't be accurate.
-	removeErr := w.removeUnminedTx(tx)
+	removeErr := w.invalidateUnminedTx(ctx, tx)
```
yyforyongyu, line 158: When `TestMempoolAccept` is unavailable, this cleanup also runs for `SendRawTransaction` returning `chain.ErrTxAlreadyKnown` or `chain.ErrTxAlreadyConfirmed`. Those cases mean the tx was already accepted/confirmed, but invalidating here retains it as failed; later rebroadcast/confirmation hits the retained-invalid duplicate path and the wallet stops tracking a valid tx. Treat those send errors like the mempool pre-check does before invalidating. (`gpt-5.5`)

wallet/tx_publisher.go
yyforyongyu: Addressed — reflowed the `walletOwnedOutputs` signature to remove the dangling result-list opening, per the function-definition wrapping rules.

wallet/tx_publisher.go
yyforyongyu: Addressed — `publishTx` now treats `ErrTxAlreadyInMempool`, `ErrTxAlreadyKnown`, and `ErrTxAlreadyConfirmed` all as a successful publish (mirroring `checkMempool`), so an already-accepted/confirmed tx stays tracked instead of being invalidated by `Broadcast`. Covered by new `TestPublishTx` cases + `TestBroadcastAlreadyBroadcastedAtPublish`.

yyforyongyu:
LGTM, no correctness findings on current head `53413a32408c79ebdb2ecee61ef6528a78630acf`. (`gpt-5.5`)

wallet/internal/db/interface.go
```diff
@@ -327,6 +327,23 @@ type AddressStore interface {
...
+	// appear in the map: a script absent from the result is simply not owned
+	// by the wallet (the batched equivalent of GetAddress returning
+	// ErrAddressNotFound), which is not an error. Both the wallet ID and the
+	// supplied script set constrain the lookup.
+	//
+	// An empty or nil ScriptPubKeys slice returns an empty, non-nil map
+	// without issuing a backend query or returning an error.
+	ResolveOwnedAddresses(ctx context.Context,
+		query ResolveOwnedAddressesQuery) (
+		map[string]*AddressInfo, error)
```
GustavoStingelin, line 345: nit: fits two lines
```suggestion
	ResolveOwnedAddresses(ctx context.Context,
		query ResolveOwnedAddressesQuery) (map[string]*AddressInfo, error)
```

wallet/internal/sql/pg/sqlc/addresses.sql.go
```diff
@@ -9,6 +9,8 @@ import (
 	"context"
 	"database/sql"
 	"time"
+
+	"github.com/lib/pq"
```
GustavoStingelin, line 13: That is weird. sqlc is generating this with the wrong driver even though we configured it to use the standard database/sql package.

I think we may need to update the sqlc config to use:`sql_package: "pgx/v5"`.

GustavoStingelin (approved):
LGTM! 

I raised an issue related to sqlc, but I think we can resolve that later, as the itests are passing.

wallet/internal/sql/pg/sqlc/addresses.sql.go
yyforyongyu: yeah weird, will address them in a new PR!

Issue #11, opened 2013-10-30
Support pay-to-script-hash notifications


davecgh:
Currently the registered notifications only support pay-to-pubkey-hash.   Support for detecting pay-to-script-hash still needs to be added.

jrick:
This is not an issue on master today. All addresses are encoded with the EncodeAddress method of the btcutil.Address interface, so btcd transaction notifications relevant to any kind of address are supported.

PR #1249, opened 2026-05-28
wallet: SQL-only signer account-secret fallback

yyforyongyu:
## Summary

Read encrypted account private keys through `db.Store` and let
the signer derive child private keys for SQL-only accounts
(accounts that exist on the SQL store but have no legacy
waddrmgr counterpart). Adds the
`GetAccountSecret` contract on `AccountStore` so wallet signers
can resolve a derived address back to its account-level key
material through the store API instead of reaching into the
legacy waddrmgr directly.

Stack position: **#11** of the kvdb-migration stack — a
**sibling** of the main UTXO chain. Stacks directly on #1229
(impl-wallet-metadata-store); no other branch in the stack
depends on this work. Verified: zero references to
`GetAccountSecret` / `keyVault` / `AccountSecret` symbols across
the 9 other branches in the stack.

## Changes

- `db.AccountStore.GetAccountSecret` — new contract that
  resolves an account's encrypted master private key + chain
  code. pg/sqlite implement the lookup via a new sqlc query;
  kvdb implements it through legacy waddrmgr.
- `Wallet.cache` exposes the per-account encrypted secret so the
  signer can derive child keys without rereading the store on
  every signing call.
- `Wallet.signer` derives SQL-only private keys through a small
  `keyVault` wrapper that handles the decrypt + child-derivation
  path; the legacy waddrmgr path stays in place for accounts
  that have it.
- New cross-backend `TestGetAccountSecret` itests on pg, sqlite,
  and kvdb verify the contract.

## Test plan

- [ ] `make sqlc`
- [ ] `GOWORK=off go test ./wallet -run 'Test(ComputeUnlockingScriptSQL|GetPrivKeyForAddressSQL|ResolveDerivedPrivKeyFromStore)'`
- [ ] `GOWORK=off go test ./wallet/internal/db ./wallet/internal/db/sqlite ./wallet/internal/db/pg`
- [ ] `GOWORK=off make itest-db db=sqlite case=TestGetAccountSecret verbose=1`
- [ ] `GOWORK=off make itest-db db=postgres case=TestGetAccountSecret verbose=1`

### DIFF
--- a/wallet/signer.go
+++ b/wallet/signer.go
@@ -13,8 +13,10 @@ import (
 	"github.com/btcsuite/btcd/chaincfg/chainhash"
 	"github.com/btcsuite/btcd/txscript"
 	"github.com/btcsuite/btcd/wire"
+	"github.com/btcsuite/btcwallet/internal/zero"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	"github.com/btcsuite/btcwallet/wallet/internal/keyvault"
 	"github.com/btcsuite/btcwallet/walletdb"
 )
 
@@ -34,6 +36,14 @@ var (
 	// ErrInvalidSignParam is returned when the parameters for the signing
 	// operation are invalid.
 	ErrInvalidSignParam = errors.New("invalid signing parameters")
+
+	// ErrWatchOnlyAccount is returned when account metadata exists but has no
+	// private key material available for signing.
+	ErrWatchOnlyAccount = errors.New("account is watch-only")
+
+	// ErrAccountNotInStore is returned when neither legacy waddrmgr nor the
+	// durable store can resolve the signing account.
+	ErrAccountNotInStore = errors.New("account not in store")
 )
 
 // Signer provides an interface for common, safe cryptographic operations,
@@ -698,7 +708,7 @@ func (w *Wallet) ComputeUnlockingScript(ctx context.Context,
 		return nil, err
 	}
 
-	privKey, err := w.privKeyForOutput(scriptInfo)
+	privKey, err := w.privKeyForOutput(ctx, scriptInfo)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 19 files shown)


yyforyongyu (commented):
LGTM 🔑 (`deepseek-v4-pro`)

wallet/signer.go
```diff
@@ -821,12 +836,13 @@ func (w *Wallet) resolvePrivKey(pubKeyAddr waddrmgr.ManagedPubKeyAddress) (
...
 }
 
 // resolveDerivedPathPrivKey resolves one derived private key through the scoped
 // manager cache or the database-backed fallback.
-func (w *Wallet) resolveDerivedPathPrivKey(keyScope waddrmgr.KeyScope,
+func (w *Wallet) resolveDerivedPathPrivKey(ctx context.Context,
+	keyScope waddrmgr.KeyScope,
 	derivationPath waddrmgr.DerivationPath) (*btcec.PrivateKey, error) {
 
 	// TODO(yy): SQL-only accounts (created via Store.CreateDerivedAccount
```
yyforyongyu, line 848: This `TODO(yy)` is now stale — it describes the work this very commit does. It says the keyVault-backed derivation ("fetch account_secrets.encrypted_priv_key, decrypt via w.keyVault, and derive at branch/index locally") is future work that "the signer-store PR (impl-tx-creator-store) will replace this path" with — but `resolveDerivedPrivKeyFromStore` / `deriveStoredAccountChildKey` added in this commit do exactly that, wired into both miss branches just below. The premise ("SQL-only accounts ... miss both DeriveFromKeyPathCache and the DB-backed DeriveFromKeyPath fallback") is also no longer true after this change.

It's also misattributed: this landed on this branch (off `impl-wallet-metadata-store`), not `impl-tx-creator-store`.

Please drop the TODO (or shrink it to whatever genuinely remains, e.g. the cache-layer typed-error wrapping noted elsewhere) so the comment doesn't point a reader at unfinished work that's actually done right here.

(`claude-opus-4-8[1m]`)

wallet/signer.go
```diff
@@ -895,6 +944,105 @@ func (w *Wallet) resolveDerivedPrivKey(accountManager waddrmgr.AccountStore,
 	return privKey, nil
 }
 
+// resolveDerivedPrivKeyFromStore resolves one derived private key from the
```
yyforyongyu, line 947: Path-based signer APIs still bypass this fallback. `ComputeRawSig`, `SignDigest`, `ECDH`, and `DerivePrivKey` call `fetchManagedPubKeyAddress` directly, so a SQL-only account with no waddrmgr row still fails before `GetAccountSecret` is tried. Route those methods through `resolveDerivedPathPrivKey`/this helper so all private-key derivations can use the store secret. (`gpt-5.5`)

PR #1252, opened 2026-05-28
wallet+AGENTS.md: add agent guidance for wallet subsystems

GustavoStingelin:
This PR adds and refines AGENTS.md guidance for the btcwallet repository and key wallet subsystems. The goal is to give coding agents clearer repo structure, subsystem ownership, workflow constraints, and verification guidance when working in this codebase.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,160 +1,132 @@
 # AGENTS.md
 
-This file is for coding agents working in `btcwallet`.
+Welcome to the `btcwallet` knowledge base. This file defines the global project
+  structure, subsystem mappings, and critical operational constraints for all
+  subagents.
 
-## Scope and Priority
-
+## SCOPE & PRIORITY
 - Follow this file first for repo-specific workflow and style.
-- Then follow `docs/developer/` for deeper rationale.
 - If guidance conflicts, prefer the stricter rule and match nearby code.
 
-## Repo Workflow
+## OVERVIEW
+`btcwallet` is a Bitcoin wallet daemon supporting BIP0032 (HD) and BIP0044. It
+  acts as a client to `btcd`, `bitcoin-core` and `neutrino`, and provides both
+  legacy JSON-RPC and gRPC APIs to wallet clients.
 
+## REPO STRUCTURE & MAP
+| Path                   | Component                 | Description                                                | Sub-Agent Config                |
+|------------------------|:--------------------------|:-----------------------------------------------------------|:--------------------------------|
+| `docs/developer/`      | **Docs**                  | Deep-dive design docs (ADRs) and development guides.       | None                            |
+| `wallet/`              | **Wallet Core**           | Business logic, HD key derivation, transaction building.   | `wallet/AGENTS.md`              |
+| `wallet/internal/db/`  | **Database Layer**        | Storage drivers, migrations, SQL-based relational logic.   | `wallet/internal/db/AGENTS.md`  |
+| `wallet/internal/sql/` | **SQL Assets**            | SQL schema, query, migration, and sqlc codegen.            | `wallet/internal/sql/AGENTS.md` |
+| `rpc/`                 | **RPC**                   | gRPC and legacy JSON-RPC APIs and server implementations.  | None                            |
+| `chain/`               | **Blockchain**            | Sync clients for Neutrino, `btcd` RPC, and `bitcoind` RPC. | None                            |
+| `itest/`               | **Integration Scenarios** | Actual daemon-level test cases and E2E flows.              | None                            |
+| `bwtest/`              | **Harness Layer**         | Test harness definitions and backend wrappers.             | None                            |
+
+## REPO & WORKTREE WORKFLOW
 - Keep the main checkout at the repo root.
 - Place auxiliary worktrees under `<repo-root>/.worktrees/<name>`.
-- Use short issue- or task-oriented worktree names.
-- Do not create sibling worktrees outside the repo root unless asked.
+- Use short, task-oriented worktree names.
+- Don't create sibling worktrees outside the repo root unless asked.
 - Remove finished worktrees with `git worktree remove <path>`.
 - Clean stale worktree metadata with `git worktree prune`.
 
-## Tooling and Environment
+## CRITICAL GOTCHAS & CONSTRAINTS
+- **Go Version:** The workspace root requires Go `1.24.6`. Nested modules keep
+  their own `go.mod` metadata and should be tested in their own module context
+  when touched.
+- **Docker-backed Tooling:** Mutating operations (`make fmt`, `make sql`,
+  `make rpc`, `make lint`) and PostgreSQL integration tests require Docker
+  (`btcwallet-tools` image).
+- **Mutating Checks:** Targets like `make fmt-check`, `make sqlc-check`,
+  `make rpc-check`, and `make tidy-module-check` **mutate files** before
+  checking Git cleanliness. Never run them on a dirty tree unless you want
+  auto-fixes applied.
+- **Generated File Boundaries:** Do not manually edit `*.pb.go` or any code in
+  directories populated by `sqlc` or protobuf generators. Modify source
+  definitions (SQL/Proto) and run `make sql` or `make rpc`.
+- **Test Locations:** Unit tests live alongside code. DB integration tests live
+  in `wallet/internal/db/itest`. E2E integration tests are orchestrated from
+  `itest/`.
+- **ADR** Before changing architecture-sensitive areas, review
+  `docs/developer/adr/README.md` for recorded design decisions, context,
+  tradeoffs, and consequences.
 
-- The authoritative Go version is `1.24.6`.
-- Sources: `go.mod`, `.golangci.yml`, and `.github/workflows/main.yml`.
-- Some older docs mention older Go versions; ignore them.
-- Many maintenance targets use Docker-backed tooling.
-- `make fmt`, `make lint*`, `make sqlc`, and `make protolint` use Docker.
-- PostgreSQL DB integration tests also require Docker via testcontainers.
-- No Cursor rules were found in `.cursor/rules/` or `.cursorrules`.
-- No Copilot instructions were found in `.github/copilot-instructions.md`.
-- There is an additional style summary in `.gemini/styleguide.md`.
-
-## Primary Build, Format, and Codegen Commands
-
-- Build everything: `make build`
-- Install binaries: `make install`
-- CI compile check: `go install -v ./...`
-- Default `make` target is `make build`.
-- `make install` installs `btcwallet`, `cmd/dropwtxmgr`, and `cmd/sweepaccount`.
-- Go format/imports: `make fmt`; verify with `make fmt-check`
-- Lint: `make lint-config-check`, `make lint-check`, `make lint`; optional `workers=4`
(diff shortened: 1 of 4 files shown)


GustavoStingelin:
@yyforyongyu

AGENTS.md
```diff
@@ -1,160 +1,51 @@
...
-- Commit subjects typically look like `subsystem: short description`.
-- Use present tense, keep the subject near 50 chars, and wrap bodies near 72.
-- PRs should include clear test steps and cover positive and negative cases.
-- Insubstantial typo-only changes are discouraged by the contribution guide.
+Welcome to the `btcwallet` knowledge base. This file defines the global project structure, subsystem mappings, and critical operational constraints for all subagents.
+
+## OVERVIEW
+`btcwallet` is a Bitcoin wallet daemon supporting BIP0032 (HD) and BIP0044. It acts as an RPC client to `btcd` and provides both legacy JSON-RPC and experimental gRPC APIs to wallet clients.
+
+## REPO STRUCTURE & MAP
```
yyforyongyu, line 16: should we give an index here so the root agent knows where to look for sub-AGENTS.md files?

yyforyongyu (commented):
Cool I like it❤️

AGENTS.md
GustavoStingelin: Yes, I think that would help.

This lets us load context on demand instead of filling the current context with information that may not be relevant yet.

Since agents usually load the full root `AGENTS.md` into context, keeping the root file lighter and using it mainly as a map should make the instructions easier to navigate.

One dependency here is that the project packages need to be well structured. If the repository layout is clear, both humans and agents can find the right context faster and perform better when working on code.

yyforyongyu: yeah i think for now we can just focus on the `wallet` pkg, so give a map there, for the rest we still need more future work.

GustavoStingelin: ok, reduced the scope; it's also easier to review in parts

wallet/internal/db/AGENTS.md
```diff
@@ -0,0 +1,56 @@
...
+- **Shared Workflows:** Orchestrates complex DB operations using
+  backend-specific "ops" interfaces to minimize logic duplication.
+- **Runtime Helpers:** Handles SQL retries, transaction management, and error
+  classification (`runtime/`, `err/`).
+- **Pagination:** Provides shared cursor-based pagination logic (`page/`).
+
+## WHERE TO LOOK
+| Component             | Path           | Description                                                |
+|:----------------------|:---------------|:-----------------------------------------------------------|
+| **Interfaces**        | `interface.go` | Central `Store` contract and shared error definitions.     |
```
GustavoStingelin, line 21: These tables are good for us humans to read, but maybe it's not worth it for the LLMs, not sure yet what the best approach is

AGENTS.md
```diff
@@ -1,160 +1,140 @@
...
+
+### Testing
+- `make unit`: Run all package unit tests.
+- `make unit pkg=wallet case=TestName`: Targeted unit test.
+- `make unit-race pkg=wallet case=TestName`: Test with race detector enabled.
+- `make itest-db db=[sqlite|postgres]`: Perform database layer integration
+  tests.
+- `make itest chain=[btcd|bitcoind|neutrino]`: Start E2E daemon integration
+  tests.
+- `make itest chain=btcd db=kvdb icase=manager`: Filter E2E tests by case name.
```
yyforyongyu, line 97: The redaction dropped a concrete, **enforced** E2E convention that the old root carried: "case names must follow `component action` and must not use `_`" (plus the `itest/test-logs/` location). That rule is hard-enforced — `itest/main_test.go:114` fails the test on an underscore or non-`component action` name — so an agent writing a new E2E test from this guidance now learns it only by tripping the failure.

The info isn't lost from the repo (it's in `itest/README.md`), but this PR's stated goal is per-subsystem agent guidance, and `itest/` is the one notable subsystem that didn't get a sub-`AGENTS.md` (wallet/, internal/db/, internal/sql/ each did) and whose root coverage got thinner here. Consider either adding an `itest/AGENTS.md` or keeping a one-line pointer to the `itest/README.md` case-naming rule next to these `make itest` commands, so the enforced convention stays discoverable up front.

The rest of the reorg looks good — the new sub-files are accurate (verified the referenced helpers/files/targets exist), and the consolidated COMMANDS/GOTCHAS kept the essentials (incl. the mutating-checks-on-dirty-tree warning).

(`claude-opus-4-8[1m]`)

AGENTS.md
```diff
@@ -1,160 +1,140 @@
...
-
-- Run the narrowest relevant test first.
-- Before handing off a substantial change, run at least package-level tests.
-- For SQL, proto, module, or config changes, run the matching generation/check target.
-- For DB-layer changes, prefer `itest-db` in addition to unit tests.
-- For backend flow or RPC changes, consider `make itest` coverage.
-- If a change claims performance improvement, add or run a benchmark.
+## CRITICAL GOTCHAS & CONSTRAINTS
+- **Go Version:** The workspace root requires Go `1.24.6`. Some submodules still
+  target `1.22`; ignore them in favor of the root `go.mod`.
```
yyforyongyu: Nested module Go versions should not be ignored. This repo has real submodules (`walletdb`, `wtxmgr`, `wallet/txauthor`, etc.) with their own `go.mod`, and the contribution guide requires using/updating those modules in their own context rather than treating the root module as authoritative for them. Please change this to say the root version applies to the root workspace while submodules keep their own module metadata/workflow. (`gpt-5.5`)

wallet/internal/db/AGENTS.md
yyforyongyu: yeah i think it's nice to have it

AGENTS.md
```diff
@@ -1,160 +1,140 @@
...
+| `itest/`               | **Integration Scenarios** | Actual daemon-level test cases and E2E flows.              | None                            |
+| `bwtest/`              | **Harness Layer**         | Test harness definitions and backend wrappers.             | None                            |
+
+## REPO & WORKTREE WORKFLOW
 - Keep the main checkout at the repo root.
 - Place auxiliary worktrees under `<repo-root>/.worktrees/<name>`.
-- Use short issue- or task-oriented worktree names.
-- Do not create sibling worktrees outside the repo root unless asked.
+- Use short, task-oriented worktree names.
+- Don't create sibling worktrees outside the repo root unless asked.
```
yyforyongyu, line 32: is this for token efficiency😂

AGENTS.md
yyforyongyu: yeah think this is valid

AGENTS.md
```diff
@@ -1,160 +1,140 @@
...
+- `make unit-race pkg=wallet case=TestName`: Test with race detector enabled.
+- `make itest-db db=[sqlite|postgres]`: Perform database layer integration
+  tests.
+- `make itest chain=[btcd|bitcoind|neutrino]`: Start E2E daemon integration
+  tests.
+- `make itest chain=btcd db=kvdb icase=manager`: Filter E2E tests by case name.
+
+## CONVENTIONS
+
+### Go Formatting and Layout
```
yyforyongyu: think we can also reference the style guidelines and engineering guidelines under `docs/developer`?

AGENTS.md
GustavoStingelin: LOL. I did not change it myself, but I let the LLM word it on its own. Surprisingly, `do not` costs two tokens, while `don't` is only one token for GPT 5.x.

Maybe that helps a tiny bit with prompting, but I have not reviewed whether every text is using that style. 🤣

[tokenizer](https://platform.openai.com/tokenizer)

image
image

AGENTS.md
GustavoStingelin: reworded

AGENTS.md
GustavoStingelin: added

AGENTS.md
```diff
@@ -1,160 +1,154 @@
...
+  legacy JSON-RPC and gRPC APIs to wallet clients.
+
+## REPO STRUCTURE & MAP
+| Path                   | Component                 | Description                                                | Sub-Agent Config                |
+|------------------------|:--------------------------|:-----------------------------------------------------------|:--------------------------------|
+| `docs/developer/`      | **Docs**                  | Deep-dive design docs (ADRs) and development guides.       | None                            |
+| `wallet/`              | **Wallet Core**           | Business logic, HD key derivation, transaction building.   | `wallet/AGENTS.md`              |
+| `wallet/internal/db/`  | **Database Layer**        | Storage drivers, migrations, SQL-based relational logic.   | `wallet/internal/db/AGENTS.md`  |
+| `wallet/internal/sql/` | **SQL Assets**            | SQL schema, query, migration, and sqlc codegen.            | `wallet/internal/sql/AGENTS.md` |
+| `rpc/walletrpc/`       | **gRPC API**              | gRPC interface definitions and server logic.               | None                            |
```
yyforyongyu: This points agents at `rpc/walletrpc/` for gRPC server logic, but that package only contains generated `api.pb.go` while the handwritten gRPC server lives under `rpc/rpcserver/`; update the map to distinguish `rpc/api.proto`, generated `rpc/walletrpc/`, and server code in `rpc/rpcserver/`. (`gpt-5.5`)

AGENTS.md
```diff
@@ -1,160 +1,154 @@
...
+- E2E test logs are written under `itest/test-logs/`.
+
+## CONVENTIONS
+
+### Go Formatting and Layout
+- Let `make fmt` manage imports. Go files use tab indentation.
+- Keep lines near 80 columns. Markdown files use LF and wrap to 80 characters.
+- Break functions into logical stanzas. Add comments explaining why, not the
+  mechanics.
+- Exported functions need comments starting with the function name.
```
yyforyongyu: This narrows comment guidance to exported functions, but the repo contribution rules require every function and method comment to start with the declaration name, so restore the all-functions requirement to avoid agents adding undocumented unexported helpers. (`gpt-5.5`)

yyforyongyu:

AGENTS.md
yyforyongyu: yeah this is real pain i've had - agents keep ignoring my requests to add comments to all methods and structs, and tests, but only focus on public methods...I also find this section's content is redundant, as it's a weaker version of the `docs/developer/code_formatting_rules.md`.

IMO all we need to do is give an index for the files under https://github.com/btcsuite/btcwallet/tree/sql-wallet/docs/developer, like 📜 Contribution Guidelines, 🎨 Code Formatting Rules, ✅ Unit Testing Guidelines and 🏛️ Engineering Guide

yyforyongyu: Then we should remove all these, so we don't duplicate, which are either covered by the `### Linting & Formatting` above or the `docs/developer`,
```
- Let `make fmt` manage imports. Go files use tab indentation.
- Keep lines near 80 columns. Markdown files use LF and wrap to 80 characters.
...
```

AGENTS.md
GustavoStingelin: It now points only to `rpc/`, so if we want to go deeper into that package later, we should add a dedicated `AGENTS.md` there.

AGENTS.md
GustavoStingelin: Reworded, but I've kept some things in a reminder note; take a look

yyforyongyu (approved):
LGTM 🔢

PR #1256, opened 2026-06-03
db: implement the ApplyScanBatch scan engine

yyforyongyu:
Implements the `ApplyScanBatch` scan engine behind the `db.Store` interface: adds it to the `TxStore` interface and implements it across the `kvdb`, `pg`, and `sqlite` backends.

- `kvdb`: `ApplyScanBatch` orchestration with rollback, plus scan-horizon extension handlers.
- `pg` and `sqlite`: `ApplyScanBatch` implementations, with the shared batch write extracted into an `applyBatchTransaction` helper.
- Idempotent batch replay via a `CanSkipCreateTxDuplicate` helper that treats an `ErrTxAlreadyExists` result as a no-op only for an exact current-shape duplicate, never for terminal history rows.

Covered by `kvdb` unit tests and a cross-backend itest.


yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🐳 (`gemini-3.5-flash`)

yyforyongyu:
LGTM 🧱
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -498,9 +620,157 @@ func (s *Store) InvalidateUnminedTx(_ context.Context,
...
+}
+
+// legacyCreditEntries converts db-native credit addresses into legacy credit
+// entries and marks those addresses as used.
+func (s *Store) legacyCreditEntries(addrmgrNs walletdb.ReadWriteBucket,
+	req db.CreateTxRequest) ([]wtxmgr.CreditEntry, error) {
+
+	credits := make([]wtxmgr.CreditEntry, 0, len(req.Params.Credits))
+	for index, addr := range req.Params.Credits {
+		managedAddr, err := s.addrStore.Address(addrmgrNs, addr)
```
yyforyongyu: Unlike `CreateTx`, this batch path resolves and marks the supplied credit address without first checking that `req.Params.Tx.TxOut[index]` actually pays it. That lets a bad `Credits` map record an unrelated wallet address as owner of any output, corrupting the legacy UTXO set. Please call `validateCreditAddr` (and handle nil credits consistently with `addCreateTxCredit`) before appending the credit. (`gpt-5.5`)

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -104,6 +104,58 @@ func (s *Store) IterWallets(ctx context.Context,
...
+		return nil, fmt.Errorf("kvdb.Store.ListSyncedBlocks: %w",
+			errMissingAddrStore)
+	}
+
+	if query.EndHeight < query.StartHeight {
+		return nil, fmt.Errorf("kvdb.Store.ListSyncedBlocks: %w: end "+
+			"height before start height", db.ErrInvalidParam)
+	}
+
+	blocks := make([]db.Block, 0, query.EndHeight-query.StartHeight+1)
```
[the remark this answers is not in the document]
yyforyongyu: Done - inclusive range length is now converted via the checked `db.Uint32ToInt32` after the `EndHeight >= StartHeight` guard and passed as an `int` (the uint32 cap did compile, but this applies the requested hygiene).

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -357,3 +409,20 @@ func (s *Store) UpdateWalletSecrets(ctx context.Context,
...
+		return waddrmgr.BlockStamp{}, fmt.Errorf("%w: store block "+
+			"height %d exceeds max int32", db.ErrInvalidParam,
+			block.Height)
+	}
+
+	return waddrmgr.BlockStamp{
+		Height:    height,
+		Hash:      block.Hash,
+		Timestamp: block.Timestamp,
+	}, nil
```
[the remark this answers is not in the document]
yyforyongyu: Done - call sites now use the shared `db.BlockStampFromBlock` and the duplicate `kvdbBlockToBlockStamp` helper was removed (including the `ApplyRewind` caller on the stacked branch).

wallet/internal/db/kvdb/txstore.go
```diff
@@ -287,6 +287,128 @@ func (s *Store) UpdateTx(_ context.Context, params db.UpdateTxParams) error {
...
+		return fmt.Errorf("kvdb.Store.ApplyTxBatch: %w",
+			errMissingAddrStore)
+	}
+
+	err := walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
+		addrmgrNs := tx.ReadWriteBucket(waddrmgr.NamespaceKey)
+		if addrmgrNs == nil {
+			return errMissingAddrmgrNamespace
+		}
+
```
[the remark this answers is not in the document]
yyforyongyu: Done - `ApplyTxBatch` now returns early for a truly empty batch (no transactions and no SyncedTo) before opening the walletdb transaction; added a kvdb test.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Done - the batch credit path now validates each non-nil credit address against `TxOut[index]` (the same `validateCreditAddr` policy as the `CreateTx` path) and handles nil credits via the output-script fallback; added mismatch + nil tests.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪨
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -287,6 +287,134 @@ func (s *Store) UpdateTx(_ context.Context, params db.UpdateTxParams) error {
...
+		txmgrNs := tx.ReadWriteBucket(wtxmgrNamespaceKey)
+		if txmgrNs == nil {
+			return errMissingTxmgrNamespace
+		}
+
+		err := s.applyLegacyScanHorizons(addrmgrNs, params.Horizons)
```
yyforyongyu: `ExtendAddresses` updates the scoped manager in-memory address cache and next indexes after writing the bucket, but this runs before the transaction batch and synced-block writes below can still fail. If a later step returns an error, walletdb rolls the derived addresses back while the live manager still thinks the horizon was extended, so later lookups can observe addresses that were never persisted. Snapshot/restore the scoped-manager state on batch rollback, or otherwise ensure no later fallible work runs after a horizon extension mutates memory. (`gpt-5.5`)

yyforyongyu: Fixed. ApplyScanBatch now invalidates the scoped-manager account caches for horizons touched by a failed batch. That forces the next read to reload the persisted rolled-back indices, so an ExtendAddresses side effect cannot leave unpersisted horizon state visible after walletdb rolls back.

yyforyongyu: This still leaves part of the rollback side effect observable. `ExtendAddresses` also inserts every derived address into the scoped manager address cache (`s.addrs`), and `Address` returns from that cache before checking the bucket. Invalidating only the account cache fixes `AccountProperties`, but later `Address`/`MarkUsed` calls can still see addresses that walletdb rolled back. Please also remove the derived addresses from the address cache (and any pending derive-on-unlock entries), or defer those cache inserts until the batch commits. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -498,9 +663,196 @@ func (s *Store) InvalidateUnminedTx(_ context.Context,
...
+}
+
+// applyLegacySyncedBlocks connects a sequence of legacy synced blocks.
+func (s *Store) applyLegacySyncedBlocks(ns walletdb.ReadWriteBucket,
+	blocks []db.Block) error {
+
+	for i := range blocks {
+		block, err := db.BlockStampFromBlock(&blocks[i])
+		if err != nil {
+			return err
```
yyforyongyu, line 1937: `SetSyncedTo` updates the manager's in-memory synced tip immediately after each block write. If this loop applies an earlier block and then a later block returns an error, walletdb rolls the batch back but `addrStore.SyncedTo()` remains advanced to the earlier in-memory tip. That breaks the atomicity promised by `ApplyScanBatch` and can make the next scan/rollback decision start from a block that was never persisted. Snapshot/restore the synced tip on batch failure, or defer the in-memory update until the whole batch commits. (`gpt-5.5`)

yyforyongyu:
**Branch hygiene:** this PR was updated via **merge** this round rather than rebased — its range carries 2 `Merge branch 'impl-psbt-store' into 'impl-runtime-store'` commits. Prior rounds it was a clean linear rebase, and the project preserves commit history (no squash), so these merge commits would land on the target branch.

Please rebase onto the base to drop the `Merge branch …` commits and keep the stack linear. (See the 1258 thread — the same merge-based update also produced a duplicated `stop precreating` commit there.) The code changes themselves look correct.

(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Fixed — ApplyScanBatch rollback now evicts the rolled-back scan-derived addresses from the in-memory address cache and the deriveOnUnlock queue (new EvictDerivedAddresses), so they're no longer observable through Address after a failed batch.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Fixed — the rollback now restores the in-memory synced tip (RestoreSyncedTo), so a mid-batch failure no longer leaves the manager advanced past the persisted (rolled-back) state.

wallet/internal/db/kvdb/txstore.go
```diff
@@ -498,9 +713,226 @@ func (s *Store) InvalidateUnminedTx(_ context.Context,
...
+	if req.Params.Block == nil {
+		err := s.txStore.InsertUnconfirmedTx(txmgrNs, txRec, credits)
+		if err != nil {
+			return fmt.Errorf("insert unconfirmed tx: %w", err)
+		}
+
+		return nil
+	}
+
+	block, err := kvdbTxBlockMeta(req.Params.Block)
```
yyforyongyu: `ExtendAddresses` can derive past `horizon.Index` when the child at the requested horizon is invalid; it keeps going until the next valid child and updates the cached next index/address entries for that later child. This rollback range stops at `horizon.Index+1`, so a failed batch would evict only through the requested index and leave the extra derived address in `s.addrs`/`deriveOnUnlock`. Record the actual post-extension next index, for example by rereading `AccountProperties` after `ExtendAddresses`, and use that as `toIndex`. (`gpt-5.5`)

yyforyongyu:
LGTM 🪨
(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -498,9 +713,226 @@ func (s *Store) InvalidateUnminedTx(_ context.Context,
...
+
+		err = scopedMgr.ExtendAddresses(
+			ns, horizon.Account, horizon.Index, horizon.Branch,
+		)
+		if err != nil {
+			return fmt.Errorf("extend addresses: %w", err)
+		}
+
+		// ExtendAddresses derives through horizon.Index inclusive, so
+		// the post-batch next index is one past it. Record the range so
```
yyforyongyu: `SetSyncedTo` mutates the live manager tip before `txStore.Rollback` runs. If the rollback step returns an error, walletdb rolls the bucket writes back but `addrStore.SyncedTo()` remains rewound, so the next sync decision can observe a fork point that was never persisted. Snapshot the pre-rewind tip and restore it on any `ApplyRewind` failure, like `ApplyScanBatch` does. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -498,9 +713,226 @@ func (s *Store) InvalidateUnminedTx(_ context.Context,
...
+	return nil
+}
+
+// applyLegacyTxNotification records one relevant transaction notification using
+// legacy wtxmgr semantics.
+func (s *Store) applyLegacyTxNotification(addrmgrNs walletdb.ReadWriteBucket,
+	txmgrNs walletdb.ReadWriteBucket, req db.CreateTxRequest) error {
+
+	txRec, err := wtxmgr.NewTxRecordFromMsgTx(
+		req.Params.Tx, req.Received,
```
yyforyongyu: `horizon.Branch` is treated as external for every value except `InternalBranch`. The SQL `ExtendScanHorizon` rejects non-0/1 branch values, so a malformed batch can silently extend and persist the external branch only on kvdb. Reject values other than `ExternalBranch` and `InternalBranch` before selecting counts or calling `ExtendAddresses`. (`gpt-5.5`)

yyforyongyu:
LGTM 🧵 (`gpt-5.5`)

yyforyongyu:
The new `kvdb: implement ApplyScanBatch` commit (`8b1380d5`) has an unwrapped body — it's a single ~213-char line. The repo (and your sibling commits this round, e.g. `aa42ab1e`/`ab87a511`, all wrapped at ~72) follows the standard ~72-char commit-body wrap. Please reflow it.

The code is good — the `ScanHorizon.AccountName` + name-based `applyLegacyScanHorizons` resolution is a coherent fit for imported-xpub HD recovery.

(`claude-opus-4-8[1m]`)

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Addressed in `8b1380d53` (`kvdb: implement ApplyScanBatch`). `applyLegacyScanHorizons` now captures the branch next index before `ExtendAddresses` and re-reads `AccountProperties` afterward, using the actual post-extension next index as the rollback `toIndex`. So when `ExtendAddresses` skips HD-invalid children and advances past `horizon.Index+1`, a failed batch evicts the full extended range instead of stopping at the requested index.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Addressed in `8b1380d53` (`kvdb: implement ApplyScanBatch`). `applyLegacyScanHorizons` now rejects any branch other than `ExternalBranch`/`InternalBranch` with `db.ErrInvalidParam` before reading counts or calling `ExtendAddresses`, matching the SQL `ExtendScanHorizon` validation, so a malformed batch can no longer silently extend the external branch on kvdb only.

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Addressed in `f751e6ab1` (`wallet: restore the in-memory synced tip on a failed rewind`). `DBPutRewind` now snapshots the pre-rewind tip via `addrStore.SyncedTo()` before `SetSyncedTo`/`txStore.Rollback` and restores it with `RestoreSyncedTo` on any rollback failure, mirroring `ApplyScanBatch` synced-tip rollback, so a failed rewind no longer leaves the live manager tip on a never-persisted fork.

wallet/internal/db/kvdb/txstore.go
```diff
@@ -494,9 +709,375 @@ func (s *Store) InvalidateUnminedTx(_ context.Context,
...
+
+		// Rewind the wallet sync tip to the fork point before rolling
+		// transaction state back, so a reorg cannot disconnect blocks
+		// without also moving the sync tip in the same write.
+		err := s.rewindSyncTip(tx, height)
+		if err != nil {
+			return err
+		}
+
+		return s.txStore.Rollback(txmgrNs, height32)
```
yyforyongyu: `rewindSyncTip` has already called `SetSyncedTo`, which updates `addrStore.SyncedTo()` in memory. If `Rollback` fails here, walletdb reverts the bucket write but the live manager stays rewound; the current test only reopens the manager and misses that cache state. Snapshot `SyncedTo()` before the rewind and call `RestoreSyncedTo` on this error path. (`gpt-5.5`)

wallet/internal/db/pg/txstore_batch.go
```diff
@@ -0,0 +1,83 @@
...
+			if err != nil {
+				return fmt.Errorf("validate tx %d: %w", i, err)
+			}
+
+			err = db.CreateTxWithOps(ctx, req, &createTxOps{
+				invalidateUnminedTxOps: invalidateUnminedTxOps{
+					qtx: qtx,
+				},
+			})
+			if errors.Is(err, db.ErrTxAlreadyExists) {
```
yyforyongyu: Blindly skipping `ErrTxAlreadyExists` is not always idempotent now that SQL keeps failed/replaced rows. If a prior publish failure left this hash as `TxStatusFailed` and a later batch observes it confirmed, `CreateTxWithOps` returns `ErrTxAlreadyExists` and this branch leaves the row failed with no block or spend edges. Only skip when the existing row already represents the same live/mined transaction; otherwise revive/update it or return the error. The same guard in the SQL scan-batch path needs the same treatment. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Addressed in the current head — `RollbackToBlock` snapshots `SyncedTo` before the rewind, `rewindSyncTip` reports whether it rewound the live manager, and the rollback-failure path calls `RestoreSyncedTo(preRollbackSyncedTo)`, so the in-memory tip is restored if the bucket write reverts. Covered by `TestRollbackToBlockAtomicRollsBackSyncTipOnRollbackError`.

wallet/internal/db/pg/txstore_batch.go
yyforyongyu: Addressed — the SQL tx-batch and scan-batch paths now gate the duplicate skip on `canSkipBatchDuplicate` -> `db.CanSkipCreateTxDuplicate`, which only skips an exact live duplicate or a same-block published duplicate. A retained failed/replaced/orphaned row is no longer skipped; the original `ErrTxAlreadyExists` is returned so it is reconciled.

yyforyongyu:
LGTM 🪨
(`claude-opus-4-8[1m]`)

wallet/internal/db/txstore_common.go
```diff
@@ -898,6 +898,31 @@ func checkReuseCreateTx(req CreateTxRequest,
...
+// CanSkipCreateTxDuplicate reports whether a batch insert may treat an
+// ErrTxAlreadyExists result as an idempotent no-op. Only the exact current row
+// shape is skippable: unmined duplicates must have the same live status and no
+// block, while mined duplicates must be published in the same block. Terminal
+// history such as failed, replaced, or orphaned rows must not be skipped
+// because doing so would leave stale state in place.
+func CanSkipCreateTxDuplicate(req CreateTxRequest, status TxStatus,
+	block *Block) bool {
+
+	if req.Params.Block == nil {
```
yyforyongyu: This treats matching status/block as an idempotent duplicate, but `CreateTxWithOps` returns `ErrTxAlreadyExists` before running `InsertCredits` or `MarkInputsSpent`. If the same tx was first stored without a later-discovered credit, or before a now-known parent UTXO existed, a batch with the same hash/status/block returns success while leaving the UTXO/spend edge missing. Please only skip after verifying the requested credits and wallet-input spend edges already exist, or replay those idempotent edge writes before returning nil. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -283,6 +283,221 @@ func (s *Store) UpdateTx(_ context.Context, params db.UpdateTxParams) error {
...
+		}
+
+		err := s.applyLegacyTxBatch(
+			addrmgrNs, txmgrNs, params.Transactions,
+		)
+		if err != nil {
+			return err
+		}
+
+		return s.applyLegacyBatchSyncTip(addrmgrNs, params)
```
yyforyongyu: `SetSyncedTo` mutates the live address manager before `walletdb.Update` has committed. If the bucket write succeeds but `walletdb.Update` later returns an error, for example a commit failure, the persisted tx/sync batch rolls back while `addrStore.SyncedTo()` remains advanced; `ApplyScanBatch` restores this exact cache on failure but `ApplyTxBatch` does not. Snapshot `SyncedTo()` before the update and call `RestoreSyncedTo` on any error after the sync tip was advanced. (`gpt-5.5`)

wallet/internal/db/txstore_common.go
yyforyongyu: Addressed. `canSkipBatchDuplicate` / `CanSkipCreateTxDuplicate` now return the existing tx row, and `replayBatchDuplicateEdges` idempotently replays the duplicate's wallet-credit and input-spend edges rather than silently skipping them; exact duplicates that already carry every requested edge are still skipped. Covered by `TestCanSkipCreateTxDuplicate` and `TestApplyTxBatchDuplicate` (sqlite + postgres).

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Addressed. kvdb `ApplyTxBatch` now snapshots `SyncedTo` up front and restores it via `RestoreSyncedTo` if the batch fails after attempting the synced-tip update, so both the in-memory synced tip and the persisted DB state stay rolled back on error. Covered by `TestApplyTxBatch`.

yyforyongyu:
LGTM ✅

(`claude-opus-4-8[1m]`)

wallet/internal/db/pg/txstore_applyscanbatch.go
```diff
@@ -0,0 +1,289 @@
...
+		// created. CreateTxWithOps needs the confirming block row during
+		// PrepareBlock, so creating transactions first would fail with
+		// ErrBlockNotFound.
+		err := applyBatchSyncedBlocks(ctx, qtx, params)
+		if err != nil {
+			return err
+		}
+
+		for i := range params.Transactions {
+			err = applyBatchTransaction(ctx, qtx, params.Transactions[i])
```
yyforyongyu: This uses each transaction's own `WalletID` while the horizons and synced blocks in the same scan batch are applied to `params.WalletID`. A malformed batch can therefore derive addresses and advance the sync tip for wallet A while atomically inserting the discovered transaction into wallet B. Please reject any transaction whose `WalletID` differs from the batch wallet before applying it (and mirror the guard in the other SQL backend). (`gpt-5.5`)

yyforyongyu: Addressed — `ApplyScanBatch` now rejects any batch transaction whose `WalletID` differs from the batch's, via the shared `db.ValidateBatchTransactionsWalletID` guard (checked before `execWrite` in both SQL backends), with a rejection conformance itest (sqlite + postgres) confirming no partial commit.

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` ⚙️

wallet/internal/db/txstore_common.go
```diff
@@ -954,6 +954,31 @@ func canIgnoreUnminedConfirmedDuplicate(req CreateTxRequest,
...
+		return block == nil && status == req.Params.Status
```
yyforyongyu: This predicate ignores the stored label, so a non-idempotent duplicate can be accepted. `CreateTxWithOps` correctly returns `ErrTxAlreadyExists` for a matching hash/status/block with a different label, but `applyBatchTransaction` then calls this helper and returns nil when only status/block match, leaving the original label in place while the batch reports success. Please include the existing label in the duplicate check, or reuse the full idempotence shape, before replaying edges. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -1329,6 +1339,269 @@ func (s *Store) rollbackForkBlock(ns walletdb.ReadBucket,
...
+		// written to the addrmgr bucket before its crediting transaction
+		// is recorded.
+		err := s.applyLegacyScanHorizons(
+			addrmgrNs, params.Horizons, &horizonRollbacks,
+		)
+		if err != nil {
+			return err
+		}
+
+		err = s.applyLegacyTxBatch(
```
yyforyongyu, line 1714: This scan-batch path applies `params.Transactions` without the wallet-ID guard used by `ApplyTxBatch` and the SQL scan-batch implementations. A malformed batch whose transaction has a different `WalletID` is accepted and written while the horizons/synced blocks are applied to the batch wallet, so backends diverge and the batch is not actually wallet-scoped. Please call `db.ValidateBatchTransactionsWalletID(params.WalletID, params.Transactions)` before opening the update. (`gpt-5.5`)

yyforyongyu, line 1714: This path feeds scan transactions straight into applyLegacyTxBatch, unlike ApplyTxBatch which validates wallet IDs/nil txs and topologically sorts parents first. With a child before its in-batch parent, the legacy insert records no spend edge because the parent credit is not present yet, and the parent is later left unspent. Validate and sort params.Transactions before applying them. (`gpt-5.5 xhigh`)

wallet/internal/db/txstore_common.go
yyforyongyu: This treats a duplicate as skippable solely on status/block. The normal idempotence check above also requires matching Label and IsCoinbase; with a different label this returns true and callers replay edges instead of surfacing ErrTxAlreadyExists. Include those fields in the predicate or load enough row data to call the full check. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🦉

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` ⚙️

wallet/internal/db/kvdb/txstore.go
```diff
@@ -1677,6 +1677,306 @@ func (s *Store) rollbackForkBlock(ns walletdb.ReadBucket,
...
+		if err != nil {
+			return fmt.Errorf("extend addresses: %w", err)
+		}
+
+		// ExtendAddresses derives at least through horizon.Index, but it
+		// skips HD-invalid children, so the branch's next index can land
+		// past horizon.Index+1. Re-read the post-extension next index so
+		// the rollback range covers every child this extension may have
+		// cached, not just the requested one.
+		props, err = scopedMgr.AccountProperties(ns, account)
```
yyforyongyu: If this post-extension `AccountProperties` read fails, `ExtendAddresses` has already advanced live address-manager state, but no `scanHorizonRollback` has been appended yet. The bucket write rolls back while the live cache keeps the derived range. Record a rollback range before this read or evict on this error path. (`gpt-5.5 xhigh`)

wallet/internal/db/sqlite/txstore_applyscanbatch.go
```diff
@@ -10,6 +10,92 @@ import (
...
+		// before any relevant transaction confirmed in those blocks is
+		// created. CreateTxWithOps needs the confirming block row during
+		// PrepareBlock, so creating transactions first would fail with
+		// ErrBlockNotFound.
+		err := applyBatchSyncedBlocks(ctx, qtx, params)
+		if err != nil {
+			return err
+		}
+
+		for i := range params.Transactions {
```
yyforyongyu: These scan-batch transactions are applied in caller order. A child that appears before its in-batch parent will not find the parent credit yet, so the spend edge is silently dropped. Sort `params.Transactions` parents-first before this loop, as `ApplyTxBatch` does. (`gpt-5.5 xhigh`)

wallet/internal/db/pg/txstore_applyscanbatch.go
```diff
@@ -11,6 +11,95 @@ import (
...
+		// before any relevant transaction confirmed in those blocks is
+		// created. CreateTxWithOps needs the confirming block row during
+		// PrepareBlock, so creating transactions first would fail with
+		// ErrBlockNotFound.
+		err := applyBatchSyncedBlocks(ctx, qtx, params)
+		if err != nil {
+			return err
+		}
+
+		for i := range params.Transactions {
```
yyforyongyu: These scan-batch transactions are applied in caller order. A child that appears before its in-batch parent will not find the parent credit yet, so the spend edge is silently dropped. Sort `params.Transactions` parents-first before this loop, as `ApplyTxBatch` does. (`gpt-5.5 xhigh`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -1670,6 +1670,330 @@ func (s *Store) rollbackForkBlock(ns walletdb.ReadBucket,
...
+	// range covers every child this extension may have cached, not just the
+	// requested one.
+	props, err = scopedMgr.AccountProperties(ns, account)
+	if err != nil {
+		// ExtendAddresses mutates live caches before this read. If the read
+		// fails, the rollback list has no entry for this extension, so evict
+		// the possible derivation range immediately.
+		if horizon.Index >= fromIndex {
+			scopedMgr.EvictDerivedAddresses(
+				account, horizon.Branch, fromIndex, horizon.Index+1,
```
yyforyongyu: If `ExtendAddresses` skips an HD-invalid child, it can cache later valid children past `horizon.Index+1` before this post-extension read fails. This fallback would then leave those cached addresses/pending unlock entries live after the walletdb rollback; evict through the actual advanced next index, or otherwise cover the full derived range. (`gpt-5.5`)

yyforyongyu: This immediate cleanup still assumes the extension stopped at `horizon.Index+1`. If `ExtendAddresses` skipped invalid children before this `AccountProperties` read failed, the live next index can be higher, so cached children above `horizon.Index+1` survive the rollback. Invalidate the account cache or evict a conservative range in this failure path. (`gpt-5.5 xhigh`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -1716,6 +1716,332 @@ func (s *Store) rollbackForkBlock(ns walletdb.ReadBucket,
...
+	props, err = scopedMgr.AccountProperties(ns, account)
+	if err != nil {
+		// ExtendAddresses mutates live caches before this read. If the read
+		// fails, the rollback list has no entry for this extension, so evict a
+		// conservative range immediately. The extension can skip HD-invalid
+		// children and cache valid children past horizon.Index.
+		if horizon.Index >= fromIndex {
+			scopedMgr.EvictDerivedAddresses(
+				account, horizon.Branch, fromIndex,
+				waddrmgr.MaxAddressesPerAccount+1,
```
yyforyongyu, line 1902: Using `MaxAddressesPerAccount+1` here makes rollback cleanup potentially iterate the entire BIP32 non-hardened range. `EvictDerivedAddresses` re-derives every index in `[fromIndex, toIndex)` when account info is cached, so a post-extension `AccountProperties` failure after a small horizon can try roughly 2.1B derivations while unwinding. Invalidate the account cache or avoid an unbounded eviction range. (`gpt-5.5 xhigh`)

yyforyongyu, line 1902: This fallback can turn a small rolled-back extension into a scan of the entire account range. When the account info is cached, `EvictDerivedAddresses` re-derives every index in `[fromIndex, toIndex)` while holding the manager lock; passing `MaxAddressesPerAccount+1` can therefore run billions of derivations on this error path. Track the attempted post-extension end or evict by scanning cached keys instead of using the global max bound. (`gpt-5.5`)

yyforyongyu:

wallet/internal/db/kvdb/txstore.go
```diff
@@ -1644,6 +1644,332 @@ func (s *Store) rollbackForkBlock(ns walletdb.ReadBucket,
...
+	// address manager's in-memory state synchronously, before this batch
+	// commits, yet walletdb only rolls back the backing bucket writes on
+	// failure. Accumulate the exact in-memory advances so a rolled-back batch
+	// can undo them without clobbering unrelated live state.
+	var (
+		horizonRollbacks []scanHorizonRollback
+		syncTipRestore   batchSyncTipRestore
+	)
+
+	err = walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
```
yyforyongyu, line 1691: This batch has the same pre-commit address-manager cache mutations and commit-failure restore path as the other kvdb Store writes, but it does not hold `s.writeMu` while the update and rollback run. A concurrent Store write can commit overlapping synced-tip or horizon cache state before this failed batch restores, letting this rollback clear state from the committed write. Hold `s.writeMu` across the update and cache rollback here too. (`gpt-5.5 xhigh`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -1644,6 +1644,332 @@ func (s *Store) rollbackForkBlock(ns walletdb.ReadBucket,
...
+	fromIndex uint32
+
+	// toIndex is the branch's next index after the extension, i.e. one past
+	// the last child the extension may have derived.
+	toIndex uint32
+}
+
+// ApplyScanBatch atomically applies recovery scan writes through the legacy
+// walletdb managers.
+func (s *Store) ApplyScanBatch(_ context.Context,
```
yyforyongyu, line 1647: `ApplyScanBatch` mutates the live address-manager cache (horizon next-indices, `applyLegacySyncedBlocks` synced tip) and runs `rollbackScanBatchCaches` -> `RestoreSyncedToIfCurrent` on commit failure, yet unlike `ApplyTxBatch` and `RollbackToBlock` it never takes `s.writeMu`; a concurrent sibling write committing the same tip in the gap can make this failure-path restore revert a legitimately-persisted tip (the ABA `writeMu` exists to prevent). Wrap the `walletdb.Update` + restore in `s.writeMu.Lock()`.

(`claude-opus-4-8[1m]`)

yyforyongyu:
No issues found by gpt-5.5 🧩

yyforyongyu:

yyforyongyu:

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🧱

yyforyongyu:
No issues found by gpt-5.5 🐙

wallet/internal/db/pg/txstore_batch.go
```diff
@@ -72,25 +57,119 @@ func (s *Store) ApplyTxBatch(ctx context.Context,
...
+	skip, txID, skipErr := canSkipBatchDuplicate(ctx, qtx, req)
+	if skipErr != nil {
+		return skipErr
+	}
+
+	if !skip {
+		return err
+	}
+
+	return replayBatchDuplicateEdges(ctx, req, txID, ops)
```
Abdulkbk: `replayBatchDuplicateEdges` will always be a no-op here. the only time this is not a no-op is in the path: `db.CreateTxWithOps` -> `handleExistingCreateTx` -> `canIgnoreUnminedConfirmedDuplicate`, which can return nil without filling the credits. But `canSkipBatchDuplicate` returns `skip=false` in that scenario and we error `if !skip {`.

Abdulkbk (approved):
LGTM 🚢

one non-blocking.

wallet/internal/db/pg/txstore_batch.go
yyforyongyu: Nice catch, now fixed!

PR #1258, opened 2026-06-03
wallet: route runtime startup, sync-tip, and rewind through the Store

yyforyongyu:
Routes the wallet's runtime startup and sync-tip/rollback paths through the runtime `db.Store`, and wires the runtime store into the syncer.

- Startup: route runtime startup setup and pass the runtime store into the syncer.
- Sync tip and rollback: route sync-tip updates and rollback block reads; route manual rewinds through the Store and restore the in-memory synced tip on a failed rewind.
- Wallet-scoped rewind: add `RewindWallet` to the store interface, implemented in `kvdb`, `pg`, and `sqlite`, so a manual rescan can rewind one wallet without deleting shared block rows; also promotes unmined tx duplicates.
- Relevant transactions: route relevant-tx notifications and unmined rebroadcast reads through the Store.

### DIFF
--- a/wallet/db_ops.go
+++ b/wallet/db_ops.go
@@ -501,9 +501,20 @@ func (s *syncer) DBGetSyncedBlocks(_ context.Context, startHeight,
 func (s *syncer) DBPutRewind(_ context.Context,
 	bs waddrmgr.BlockStamp) error {
 
+	var preRewindTip waddrmgr.BlockStamp
+
 	err := walletdb.Update(s.cfg.DB, func(tx walletdb.ReadWriteTx) error {
 		addrmgrNs := tx.ReadWriteBucket(waddrmgrNamespaceKey)
 		txmgrNs := tx.ReadWriteBucket(wtxmgrNamespaceKey)
+
+		// SetSyncedTo below writes the addrmgr bucket and advances the live
+		// manager's in-memory synced tip immediately. If the subsequent
+		// Rollback fails, walletdb rolls the bucket write back but the
+		// in-memory tip stays rewound to a fork point that was never
+		// persisted. Snapshot the pre-rewind tip inside this write transaction,
+		// immediately before the rewind, so a failed update restores the latest
+		// committed live tip.
+		preRewindTip = s.addrStore.SyncedTo()
 
 		err := s.addrStore.SetSyncedTo(addrmgrNs, &bs)
 		if err != nil {
@@ -513,6 +524,12 @@ func (s *syncer) DBPutRewind(_ context.Context,
 		return s.txStore.Rollback(txmgrNs, bs.Height+1)
 	})
 	if err != nil {
+		// walletdb rolled the addrmgr bucket back, but any synced-tip
+		// advance from SetSyncedTo survives in memory. Restore the
+		// pre-rewind tip only if the live manager still points at the
+		// rolled-back rewind target.
+		s.addrStore.RestoreSyncedToIfCurrent(preRewindTip, bs)
+
 		return fmt.Errorf("rollback wallet: %w", err)
 	}
 
--- a/wallet/internal/db/interface.go
+++ b/wallet/internal/db/interface.go
@@ -423,6 +423,13 @@ type TxStore interface {
 	// ApplyScanBatch atomically records recovery scan writes for one wallet.
 	ApplyScanBatch(ctx context.Context, params ScanBatchParams) error
 
+	// RewindWallet atomically detaches one wallet's confirmed transactions at
+	// and above the block after params.Block, and updates only that wallet's
+	// synced tip to params.Block. Unlike RollbackToBlock, this method is
+	// wallet-scoped and must not delete shared block rows or mutate other
+	// wallets' sync states.
+	RewindWallet(ctx context.Context, params RewindWalletParams) error
+
 	// GetTx retrieves a transaction record by its hash. It takes a context
 	// and GetTxQuery, returning a TxInfo struct or an error if the
 	// transaction is not found.
--- a/wallet/internal/db/pg/txstore_batch.go
+++ b/wallet/internal/db/pg/txstore_batch.go
@@ -2,8 +2,12 @@ package pg
 
 import (
 	"context"
+	"database/sql"
+	"errors"
 	"fmt"
 
+	"github.com/btcsuite/btcd/address/v2"
+	"github.com/btcsuite/btcd/txscript/v2"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	"github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
 )
@@ -59,6 +63,11 @@ func (s *Store) ApplyTxBatch(ctx context.Context,
 func applyBatchTransaction(ctx context.Context, qtx *sqlc.Queries,
 	params db.CreateTxParams) error {
 
+	params, err := resolveCreditCandidates(ctx, qtx, params)
+	if err != nil {
+		return fmt.Errorf("resolve credit candidates: %w", err)
+	}
+
 	req, err := db.NewCreateTxRequest(params)
 	if err != nil {
 		return fmt.Errorf("validate tx: %w", err)
@@ -84,7 +93,218 @@ func applyBatchTransaction(ctx context.Context, qtx *sqlc.Queries,
 		}
 	}
 
-	return db.CreateTxWithOps(ctx, req, ops)
+	err = db.CreateTxWithOps(ctx, req, ops)
+
+	return promoteBatchDuplicate(ctx, qtx, req, ops, err)
+}
+
+// resolveCreditCandidates resolves notification credit candidates inside the
+// batch write transaction and promotes owned outputs into Credits.
+func resolveCreditCandidates(ctx context.Context, qtx *sqlc.Queries,
+	params db.CreateTxParams) (db.CreateTxParams, error) {
+
+	if len(params.CreditCandidates) == 0 {
+		return params, nil
+	}
+
+	credits := make(
+		map[uint32]address.Address,
+		len(params.Credits)+len(params.CreditCandidates),
+	)
+	for index, addr := range params.Credits {
+		credits[index] = addr
+	}
+
+	for index, candidates := range params.CreditCandidates {
+		if _, ok := credits[index]; ok {
+			continue
+		}
+
+		if uint64(index) >= uint64(len(params.Tx.TxOut)) {
+			return params, fmt.Errorf("%w: credit candidate index %d: %w",
+				db.ErrInvalidParam, index, db.ErrIndexOutOfRange)
+		}
+
+		outputScript := params.Tx.TxOut[index].PkScript
+
+		owned, err := creditScriptOwned(
+			ctx, qtx, params.WalletID, outputScript,
+		)
+		if err != nil {
+			return params, fmt.Errorf("lookup output script %d: %w", index,
+				err)
+		}
+
+		if owned {
+			credits[index] = nil
+			continue
+		}
+
+		ownedAddr, err := ownedCreditCandidate(
+			ctx, qtx, params.WalletID, outputScript, candidates,
+		)
+		if err != nil {
+			return params, fmt.Errorf("lookup candidate %d: %w", index, err)
+		}
+
+		if ownedAddr != nil {
+			credits[index] = ownedAddr
+		}
+	}
+
+	params.Credits = credits
+	params.CreditCandidates = nil
+
+	return params, nil
+}
+
+// ownedCreditCandidate returns the first candidate address owned by the wallet.
+func ownedCreditCandidate(ctx context.Context, qtx *sqlc.Queries,
+	walletID uint32, outputScript []byte,
+	candidates []address.Address) (address.Address, error) {
+
+	for _, candidate := range candidates {
+		if candidate == nil {
+			return nil, fmt.Errorf("%w: nil credit candidate",
+				db.ErrInvalidParam)
+		}
+
+		err := db.ValidateCreditAddrMembership(candidate, outputScript)
+		if err != nil {
+			return nil, err
+		}
+
+		candidateScript, err := txscript.PayToAddrScript(candidate)
+		if err != nil {
+			return nil, fmt.Errorf("candidate script: %w", err)
+		}
+
+		owned, err := creditScriptOwned(
+			ctx, qtx, walletID, candidateScript,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		if owned {
+			return candidate, nil
+		}
+	}
+
+	return nil, nil //nolint:nilnil // A nil address means no candidate matched.
+}
+
+// creditScriptOwned reports whether walletID has an address row for script.
+func creditScriptOwned(ctx context.Context, qtx *sqlc.Queries, walletID uint32,
+	script []byte) (bool, error) {
+
+	_, err := qtx.GetAddressByScriptPubKey(
+		ctx, sqlc.GetAddressByScriptPubKeyParams{
+			ScriptPubKey: script,
+			WalletID:     int64(walletID),
+		},
+	)
+	if errors.Is(err, sql.ErrNoRows) {
+		return false, nil
+	}
+
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+// promoteBatchDuplicate promotes a pending unmined duplicate to published when
+// the shared create flow rejected that non-idempotent state transition.
+func promoteBatchDuplicate(ctx context.Context, qtx *sqlc.Queries,
+	req db.CreateTxRequest, ops db.CreateTxOps, createErr error) error {
+
+	txID, ok, err := promotableBatchDuplicate(ctx, qtx, req, createErr)
+	if err != nil {
+		return err
+	}
+
+	if !ok {
+		return createErr
+	}
+
+	rows, err := qtx.UpdateTransactionStatusByIDs(
+		ctx, sqlc.UpdateTransactionStatusByIDsParams{
+			WalletID: int64(req.Params.WalletID),
+			Status:   int16(db.TxStatusPublished),
+			TxIds:    []int64{txID},
+		},
+	)
+	if err != nil {
+		return fmt.Errorf("promote duplicate tx status: %w", err)
+	}
+
+	if rows == 0 {
+		return fmt.Errorf("promote duplicate tx %s: %w", req.TxHash,
+			db.ErrTxNotFound)
+	}
+
+	err = ops.InsertCredits(ctx, req, txID)
+	if err != nil {
+		return fmt.Errorf("record promoted duplicate tx credits: %w", err)
+	}
+
+	err = ops.MarkInputsSpent(ctx, req, txID)
+	if err != nil {
+		return fmt.Errorf("record promoted duplicate tx spends: %w", err)
+	}
+
+	return nil
+}
+
+// promotableBatchDuplicate returns the existing tx id when createErr is a
+// pending-unmined duplicate that this batch may promote to published state.
+func promotableBatchDuplicate(ctx context.Context, qtx *sqlc.Queries,
+	req db.CreateTxRequest, createErr error) (int64, bool, error) {
+
+	if createErr == nil || !errors.Is(createErr, db.ErrTxAlreadyExists) {
+		return 0, false, nil
+	}
+
+	row, err := qtx.GetTransactionByHash(
+		ctx, sqlc.GetTransactionByHashParams{
+			WalletID: int64(req.Params.WalletID),
+			TxHash:   req.TxHash[:],
+		},
+	)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return 0, false, nil
+		}
+
+		return 0, false, fmt.Errorf("get duplicate tx: %w", err)
+	}
+
+	status, err := db.ParseTxStatus(int64(row.TxStatus))
+	if err != nil {
+		return 0, false, fmt.Errorf("parse duplicate tx status: %w", err)
+	}
+
+	var block *db.Block
+	if row.BlockHeight.Valid {
+		block, err = buildBlock(
+			row.BlockHeight, row.BlockHash, row.BlockTimestamp,
+		)
+		if err != nil {
+			return 0, false, fmt.Errorf("build duplicate tx block: %w",
+				err)
+		}
+	}
+
+	if !db.CanPromoteUnminedCreateTxDuplicate(
+		req, status, row.IsCoinbase, block,
+	) {
+
+		return 0, false, nil
+	}
+
+	return row.ID, true, nil
 }
 
 // applyBatchSyncTip applies the optional sync-tip update within a batch.
--- a/wallet/syncer.go
+++ b/wallet/syncer.go
@@ -349,7 +349,7 @@ func (s *syncer) checkRollback(ctx context.Context) error {
 		startHeight := max(0, endHeight-batchSize+1)
 
 		// Fetch Local Batch (from wallet's database).
-		localHashes, err = s.DBGetSyncedBlocks(
+		localHashes, err = s.syncedBlockHashes(
 			ctx, startHeight, endHeight,
 		)
 		if err != nil {
@@ -412,6 +412,69 @@ func (s *syncer) checkRollback(ctx context.Context) error {
 	}
 
 	return nil
+}
+
+// syncedBlockHashes returns the wallet's synced block hashes for the inclusive
+// height range.
+func (s *syncer) syncedBlockHashes(ctx context.Context, startHeight,
+	endHeight int32) ([]*chainhash.Hash, error) {
+
+	if s.store == nil {
+		return s.DBGetSyncedBlocks(ctx, startHeight, endHeight)
+	}
+
+	start, err := db.Int64ToUint32(int64(startHeight))
+	if err != nil {
+		return nil, fmt.Errorf("start height %d: %w", startHeight, err)
+	}
+
+	end, err := db.Int64ToUint32(int64(endHeight))
+	if err != nil {
+		return nil, fmt.Errorf("end height %d: %w", endHeight, err)
+	}
+
+	blocks, err := s.store.ListSyncedBlocks(
+		ctx, db.ListSyncedBlocksQuery{
+			StartHeight: start,
+			EndHeight:   end,
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list synced blocks: %w", err)
+	}
+
+	hashes := make([]*chainhash.Hash, len(blocks))
+	for i := range blocks {
+		hashes[i] = &blocks[i].Hash
+	}
+
+	return hashes, nil
+}
+
+// syncedTo returns the wallet's current sync tip from the active runtime
+// backend.
+func (s *syncer) syncedTo(ctx context.Context) (waddrmgr.BlockStamp, error) {
+	if s.store == nil {
+		return s.addrStore.SyncedTo(), nil
+	}
+
+	walletInfo, err := s.store.GetWallet(ctx, s.cfg.Name)
+	if err != nil {
+		return waddrmgr.BlockStamp{}, fmt.Errorf("get wallet sync tip: %w",
+			err)
+	}
+
+	if walletInfo.SyncedTo == nil {
+		return waddrmgr.BlockStamp{}, nil
+	}
+
+	syncedTo, err := db.BlockStampFromBlock(walletInfo.SyncedTo)
+	if err != nil {
+		return waddrmgr.BlockStamp{}, fmt.Errorf("decode wallet sync tip: %w",
+			err)
+	}
+
+	return syncedTo, nil
 }
 
 // findForkPoint compares local and remote block hashes to find the last
--- a/wallet/syncer.go
+++ b/wallet/syncer.go
@@ -15,6 +15,7 @@ import (
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/btcsuite/btcwallet/chain"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	"github.com/btcsuite/btcwallet/wtxmgr"
 )
 
@@ -188,6 +189,12 @@ type syncer struct {
 	// txStore is the transaction manager.
 	txStore wtxmgr.TxStore
 
+	// store is the transitional database store used by migrated runtime paths.
+	store db.Store
+
+	// walletID is the database wallet identifier used by store-backed paths.
+	walletID uint32
+
 	// state tracks the chain synchronization status.
 	state atomic.Uint32
 
@@ -202,17 +209,34 @@ type syncer struct {
 	publisher TxPublisher
 }
 
+// syncerStoreConfig contains store-backed runtime options for the syncer.
+type syncerStoreConfig struct {
+	// store is the transitional database store used by migrated runtime paths.
+	store db.Store
+
+	// walletID is the database wallet identifier used by store-backed paths.
+	walletID uint32
+}
+
 // newSyncer creates a new syncer instance.
 func newSyncer(cfg Config, addrStore waddrmgr.AddrStore,
-	txStore wtxmgr.TxStore, publisher TxPublisher) *syncer {
+	txStore wtxmgr.TxStore, publisher TxPublisher,
+	storeConfigs ...syncerStoreConfig) *syncer {
 
-	return &syncer{
+	s := &syncer{
 		cfg:         cfg,
 		addrStore:   addrStore,
 		txStore:     txStore,
 		scanReqChan: make(chan *scanReq, 1),
 		publisher:   publisher,
 	}
+
+	if len(storeConfigs) > 0 {
+		s.store = storeConfigs[0].store
+		s.walletID = storeConfigs[0].walletID
+	}
+
+	return s
 }
 
 // syncState returns the current synchronization state of the wallet.
--- a/wallet/syncer.go
+++ b/wallet/syncer.go
@@ -399,7 +399,7 @@ func (s *syncer) checkRollback(ctx context.Context) error {
 			}
 
 			// Perform the rollback.
-			return s.DBPutRewind(ctx, waddrmgr.BlockStamp{
+			return s.rewindToBlock(ctx, waddrmgr.BlockStamp{
 				Height:    forkHeight,
 				Hash:      *forkHash,
 				Timestamp: header.Timestamp,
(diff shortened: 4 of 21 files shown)


yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🪴 (`gemini-3.5-flash`)

yyforyongyu:
LGTM 🛰️
(`claude-opus-4-8[1m]`)

wallet/syncer.go
```diff
@@ -1433,6 +1741,214 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+			WalletID:    s.walletID,
+			SkipBalance: true,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("list scan accounts: %w", err)
+		}
+
+		props := make([]*waddrmgr.AccountProperties, 0, len(accounts))
+		for i := range accounts {
+			props = append(props, storeAccountProperties(accounts[i]))
```
yyforyongyu: `ListAccounts` includes imported account rows with NULL account_number, which `AccountRowToInfo` exposes as account 0. Feeding those rows into `RecoveryState.Initialize` makes it derive external/internal lookahead for account 0 even though imported accounts do not have HD branches; legacy full scans build horizons from `ActiveAccounts()` and only watch imports through the active-address/UTXO lists. Please skip `Origin == db.ImportedAccount` for horizon data. (`gpt-5.5`)

wallet/syncer.go
```diff
@@ -1204,7 +1512,7 @@ func (s *syncer) scanBatch(ctx context.Context, syncedTo waddrmgr.BlockStamp,
 		return fmt.Errorf("%w: scan batch empty", ErrScanBatchEmpty)
 	}
 	// Process Batch (Update). We do this in a single DB transaction.
-	return s.DBPutSyncBatch(ctx, results)
+	return s.putSyncBatch(ctx, results)
```
yyforyongyu: In store-backed mode this now commits the batch only through `ApplyScanBatch`, but the next loop still gets the starting point from `s.addrStore.SyncedTo()` in `advanceChainSync`. The SQL store write does not advance that legacy manager, so the wallet can keep starting from the old height and rescan/rewrite the same blocks indefinitely. Please read the tip from the store or keep `addrStore` synchronized before routing scan commits through the store. (`gpt-5.5`)

wallet/syncer.go
```diff
@@ -620,6 +664,259 @@ func (s *syncer) updateSyncTip(ctx context.Context,
...
+				return fmt.Errorf("credit output %d: %w", index,
+					db.ErrInvalidParam)
+			}
+
+			pkScript := match.Rec.MsgTx.TxOut[index].PkScript
+
+			_, err := s.store.GetAddress(ctx, db.GetAddressQuery{
+				WalletID:     s.walletID,
+				ScriptPubKey: pkScript,
+			})
```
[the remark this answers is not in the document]
yyforyongyu: Done - `applyStoreTxBatch` now builds the Credits map from the matched member address (`entry.Address`, like `appendScanTxParams`) instead of re-resolving via `GetAddress(outputScript)`, which dropped bare-multisig member-owned credits. Added a regression test.

wallet/syncer.go
```diff
@@ -1433,6 +1741,214 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+				return nil, fmt.Errorf("scan address script: %w", err)
+			}
+
+			if len(scriptAddrs) == 0 {
+				return nil, fmt.Errorf("scan address script: %w",
+					db.ErrAddressNotFound)
+			}
+
+			addrs = append(addrs, scriptAddrs[0])
+		}
```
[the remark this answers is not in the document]
yyforyongyu: Corrected the comment: it enumerates all active addresses per account, which is a safe superset of legacy `ForEachRelevantActiveAddress` (that watches only internal branches for non-default scopes). The extra wallet-owned addresses are registered as non-lookahead history filters, so the broader set is safe.

wallet/syncer.go
yyforyongyu: Added a skip for imported accounts in `storeScanHorizons` (they have no HD external/internal horizons). kvdb `ListAccounts` already excludes the imported pseudo-account, so this is defense-in-depth and backend-agnostic; imported addresses are still watched via the active-address and UTXO-watch lists.

wallet/syncer.go
yyforyongyu: kvdb already satisfies the suggested fix: `ApplyScanBatch` advances the legacy addrStore tip via `SetSyncedTo`, so `advanceChainSync`'s `SyncedTo()` read makes progress (verified by the passing kvdb itest). Added a comment documenting the coupling and a TODO that a future SQL Store wired into the syncer must either advance the legacy tip or read the tip from the Store.

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu (commented):
LGTM (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🧭
(`claude-opus-4-8[1m]`)

wallet/syncer.go
```diff
@@ -620,6 +664,258 @@ func (s *syncer) updateSyncTip(ctx context.Context,
...
+			// Credit the matched address the notification path
+			// already resolved, exactly like appendScanTxParams.
+			// We must NOT re-resolve via GetAddress(outputScript):
+			// for a bare-multisig output owned through a member
+			// pubkey, PayToAddrScript(member) != outputScript, so a
+			// script lookup misses and the credit would be dropped.
+			// The store enforces ownership when applying the batch
+			// (validateCreditAddr confirms membership and Address
+			// rejects non-wallet addresses).
+			credits[index] = entry.Address
```
yyforyongyu: `prepareTxMatches` extracts every address from every output, not only wallet-owned ones; the legacy path then calls `resolveTxMatches` to drop addresses that are not in the wallet. This store path now puts each extracted address directly into `Credits`, so any relevant transaction that also pays a third party makes `ApplyTxBatch` try to credit that non-wallet output and fail with `ErrAddressNotFound`. Keep the member-address fix, but first resolve `entry.Address` against the wallet (using `PayToAddrScript(entry.Address)`, not the full output script) and only include owned addresses. (`gpt-5.5`)

wallet/syncer.go
```diff
@@ -1110,6 +1417,36 @@ func (s *syncer) dispatchScanStrategy(ctx context.Context,
 	}
 }
 
+// syncedTip returns the wallet's current synced-to block. In store-backed mode
+// it reads the tip from the Store so callers do not depend on the legacy
+// addrStore tip being kept in lockstep by ApplyScanBatch; otherwise it falls
+// back to the legacy addrStore.
+func (s *syncer) syncedTip(ctx context.Context) (waddrmgr.BlockStamp, error) {
```
yyforyongyu: This helper is only used by forward sync; rollback detection and full-rescan rewind decisions still read `s.addrStore.SyncedTo()` directly. In a Store-backed backend that does not mirror `ApplyScanBatch`/`ApplyRewind` into the legacy manager, those paths use a stale height, so `checkRollback` can skip a reorg entirely and `scanWithRewind` can decide there is nothing to rewind. Route those callers through this helper as well, or keep the legacy tip synchronized whenever the Store tip changes. (`gpt-5.5`)

yyforyongyu: Fixed. checkRollback, advanceChainSync, and scanWithRewind now route through syncedTip; in store-backed mode syncedTip reads the current tip from Store.GetWallet instead of depending on the legacy addrStore tip. The legacy addrStore path is still used when no Store is configured.

wallet/syncer.go
yyforyongyu: Fixed. The store-backed sync path now resolves each extracted address against the wallet by PayToAddrScript(entry.Address) and only includes owned addresses in Credits. That preserves the bare-multisig member-address fix while dropping third-party outputs from otherwise relevant transactions. Added tests for the third-party-output case.

wallet/syncer.go
yyforyongyu: The skip should only cover the imported-address pseudo-account, not every ImportedAccount origin. kvdb CreateImportedAccount stores imported xpub/watch-only accounts through NewAccountWatchingOnly, so they have ordinary account numbers plus external/internal key counts that recovery needs for lookahead. With this guard, those accounts never contribute horizons and a restored wallet can miss future derived addresses from an imported xpub. Please distinguish the NULL/imported-address row from imported xpub accounts, for example by skipping only rows without a usable account number or without HD key counts. (`gpt-5.5`)

yyforyongyu:
**Branch hygiene (1255–1258):** this sub-stack was updated via **merge** this round rather than rebased.

1. **Duplicate commit** — `db: stop precreating watch multisig account` appears twice in 1258's history: `dbccb6ab2` and `9d61ff0be` (1257's tip, already in 1258's base). Both remove the same two lines from `tx_utxo_store_test.go`, so the change is committed twice.
2. **Merge commits** — the PR head is a `Merge branch 'impl-runtime-store-2' into impl-runtime-store-3` commit; 1258's range has 3 such merges (1255/1256/1257 have 2 each). Prior rounds these were clean linear rebases.

The project preserves commit history (no squash), so both the merge commits and the duplicate would land on the target branch. Please rebase the stack linearly — drop the `Merge branch …` commits and de-duplicate `stop precreating` — to restore the per-commit cleanliness these PRs had before.

Minor: that commit is test-only (−2 lines) but carries a `db:` prefix and empty body; an `itest:`-scoped title noting the precreation is redundant (NewImportedAddress already registers the member) would read truer.

The code itself looks correct — routed scan/rescan + bare-multisig handling.

(`claude-opus-4-8[1m]`)

wallet/syncer.go
yyforyongyu: storeScanHorizons now skips only the keyless imported-address bucket (reserved name + no account public key); true imported-xpub accounts keep their external/internal horizons, in both the untargeted and targeted paths.

yyforyongyu:
LGTM 🧭
(`claude-opus-4-8[1m]`)

wallet/syncer.go
```diff
@@ -1433,6 +1835,279 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+	return props, nil
+}
+
+// storeAccountProperties converts store account metadata into the recovery
+// state horizon shape.
+func storeAccountProperties(
+	info db.AccountInfo) *waddrmgr.AccountProperties {
+
+	return &waddrmgr.AccountProperties{
+		AccountNumber:        info.AccountNumber,
```
yyforyongyu: SQL imported xpub accounts expose `AccountNumber == 0` because `accounts.account_number` is NULL, but scan horizons are later resolved by `GetAccountByWalletScopeAndNumber`. Keeping a true imported xpub here reports its horizon as account 0, so `ApplyScanBatch` can extend the default derived account or fail instead of extending the imported account. Carry a name/row-id/public-key based identity through scan horizons, or avoid emitting imported-xpub horizons until they can be resolved without the masked account number. (`gpt-5.5`)

wallet/syncer.go
yyforyongyu: This is still not fully fixed. SQL imported-xpub rows still have NULL `account_number` and are exposed as `AccountNumber == 0`, so preserving them now feeds horizon account 0 into `ApplyScanBatch`, which resolves by account number and extends the default derived account or fails instead of extending the imported account. The fix needs a non-masked account identity for imported-xpub horizons, or should avoid emitting those horizons until they can be resolved correctly. (`gpt-5.5`)

yyforyongyu: Still unresolved at the current head: `AccountName` is now emitted, but it is keyed from `RecoveryState` by `(scope, AccountNumber)`. SQL imported-xpub rows expose `AccountNumber == 0`, so they collide with the default account while seeding `accountNames` and branch states; a full scan with both accounts can stamp a default-account horizon with the imported account name and extend the wrong account. The producer needs an identity that remains distinct before the horizon is emitted, such as account name or row identity rather than the masked number. (`gpt-5.5`)

wallet/syncer.go
yyforyongyu: Still unresolved after adding `AccountName`: this assignment uses the masked public account number before the name is recorded. In SQL, the default derived account and every imported-xpub account in the same scope can all arrive here as `AccountNumber == 0`, so `RecoveryState` cannot keep their branch states or account-name lookup distinct and can emit a horizon for the wrong account name. Please carry a non-masked identity into the scan state before flattening horizons. (`gpt-5.5`)

wallet/syncer.go
yyforyongyu: Addressed in `179c6e7a8`. You are right that an imported-xpub account surfacing `AccountNumber == 0` cannot be safely distinguished before emission: `RecoveryState` keys `accountNames`/branch state by account number and derives via `waddrmgr.DeriveAddr(number, ...)` (a by-number cache lookup), so a masked-0 imported account would collide with the default account and derive its addresses. Rather than emit an unidentifiable horizon, `storeScanHorizons` now withholds every `db.ImportedAccount` from recovery horizons via `withholdFromRecoveryHorizons` (both the ListAccounts and GetAccount paths). Imported addresses stay watched through the active-address and UTXO-watch lists, so coverage is unchanged; only the HD lookahead is deferred. `TestStoreScanHorizonsListAccountsSkipsImported` / `...GetAccountSkipsImported` now place a default account and an imported-xpub account both at `AccountNumber == 0` in one scope and assert only the derived horizon is emitted. `ScanHorizon.AccountName` is kept so lookahead can be re-enabled once `RecoveryState` keys by a non-masked identity ((scope, name) or a durable account id) — the tracked follow-up.

wallet/syncer.go
yyforyongyu: Addressed in `179c6e7a8`. This conversion no longer runs for imported accounts: `storeScanHorizons` withholds every `db.ImportedAccount` from recovery horizons (`withholdFromRecoveryHorizons`) precisely because the masked `AccountNumber == 0` cannot keep imported-xpub branch state and name lookup distinct from the default account in `RecoveryState`. Only derived accounts, which hold real distinct numbers, reach `storeAccountProperties` and seed the scan state. Carrying a non-masked identity into the scan state to safely re-enable imported-xpub lookahead is the documented follow-up; the `ScanHorizon.AccountName` plumbing is already in place for it.

yyforyongyu:
Two new commits have unwrapped bodies — `6b4f41fe` (~380-char single line) and `bd7b4f8d` (~355). The repo wraps commit bodies at ~72 (your other commits this round are wrapped), so please reflow both.

The change itself is a solid, deeper fix: classifying store addresses on `HasDerivationPath` rather than account origin correctly treats imported-xpub HD children as HD (Internal from branch, derivation populated, not flagged imported) — matching legacy `keyToManaged` — and the `+157`-line test covers it.

(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🦦 (`gpt-5.5`)

wallet/syncer.go
```diff
@@ -1832,6 +1858,277 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+	}
+
+	props := make([]*waddrmgr.AccountProperties, 0, len(targets))
+	for _, target := range targets {
+		account := target.Account
+
+		info, err := s.store.GetAccount(ctx, db.GetAccountQuery{
+			WalletID:      s.walletID,
+			Scope:         db.KeyScope(target.Scope),
+			AccountNumber: &account,
```
yyforyongyu: Targeted rescans can pass the legacy imported account (`waddrmgr.ImportedAddrAccount`) or an imported-xpub row surfaced as account 0, but this numeric lookup cannot resolve either correctly: SQL imported rows have NULL `account_number`, kvdb rejects the imported pseudo-account by number, and an imported-xpub account 0 resolves to the default derived account. The scan then errors or seeds the wrong horizon before the imported-account skip can run. Resolve imported targets by name/non-masked identity, or skip them before this number-based lookup. (`gpt-5.5`)

yyforyongyu: Addressed by the task-256 fix (landed + green): targeted rescans now resolve each target's durable identity via `resolveScanTargets` BEFORE any number lookup. The legacy `ImportedAddrAccount` bucket is skipped, and every other target is resolved to its `AccountName`, so `storeScanHorizons` looks up by name — an imported-xpub account masked to number 0 is no longer mistaken for the default derived account 0, and imported accounts are withheld from recovery horizons. Covered by `TestStoreScanHorizonsTargetedImportedBucketSkipped`, `TestStoreScanHorizonsTargetedImportedNotResolvedAsDerived`, and the SQL `TestApplyScanBatchResolvesHorizonByName` itest.

yyforyongyu:
LGTM 🧭
(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM, re-checked the targeted imported-account rescan path at `345a1c3575f3d84d57248f39446159b92edc8cb4` and found no new correctness issues. (`gpt-5.5`)

yyforyongyu:
LGTM ✅

(`claude-opus-4-8[1m]`)

wallet/syncer.go
yyforyongyu: Still not fully resolved: withholding every `db.ImportedAccount` avoids the masked-number collision, but it also drops imported-xpub external/internal horizons entirely. Store scan setup then only watches already-persisted active addresses/UTXOs, while the legacy path derives `RecoveryWindow` lookahead for the imported xpub via its real waddrmgr account number, so a rescan can miss payments to unused children beyond the active list. Please keep imported-xpub lookahead keyed by a durable identity (name/row id) or fall back to the legacy horizon path for those accounts. (`gpt-5.5`)

yyforyongyu: Acknowledged — the finding is real: `withholdFromRecoveryHorizons` withholds all imported accounts, dropping HD lookahead for imported-xpub accounts. Deferring to a dedicated, separately-reviewed change rather than folding it into this round: the proper fix needs `RecoveryState` to distinguish accounts by non-masked identity (the current withholding is asserted by tests across several commits, and the imported-xpub horizon path needs the real-backend scan harness that only exists at the tip). Tracked separately.

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔄

wallet/syncer.go
```diff
@@ -500,6 +591,161 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+		credits := make(map[uint32]btcutil.Address, len(match.Entries))
+
+		for _, entry := range match.Entries {
+			index := entry.Credit.Index
+			if uint64(index) >= uint64(len(match.Rec.MsgTx.TxOut)) {
+				return fmt.Errorf("credit output %d: %w", index,
+					db.ErrInvalidParam)
+			}
+
+			pkScript := match.Rec.MsgTx.TxOut[index].PkScript
```
yyforyongyu: This resolves ownership with the full output script but then records `entry.Address` as the credit. For bare-multisig outputs the wallet owns one member address, so `entry.Address` is the member pubkey address while `pkScript` is the multisig script; `GetAddress` misses and the credit is silently dropped even though the transaction was relevant. Resolve `entry.Address` via `PayToAddrScript(entry.Address)`, or otherwise check that address row, before adding it to `Credits`. (`gpt-5.5`)

wallet/syncer.go
```diff
@@ -500,6 +591,161 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+	if s.store == nil {
+		return s.DBPutSyncTip(ctx, block)
+	}
+
+	storeBlock, err := storeBlockFromBlockMeta(block)
+	if err != nil {
+		return err
+	}
+
+	err = s.store.UpdateWallet(
```
yyforyongyu, line 737: In store-backed mode this writes the new tip only to the Store, but rollback/rescan decisions still read `s.addrStore.SyncedTo()` (`checkRollback`, `advanceChainSync`, and `scanWithRewind`). After a block notification, the Store tip advances while the legacy manager remains stale, so later sync logic can compare from the old height or skip a needed rewind. Either mirror the Store tip into `addrStore` here, or route all synced-tip reads through the Store. (`gpt-5.5`)

yyforyongyu, line 737: This Store write advances only the Store wallet row; later code still reads s.addrStore.SyncedTo() in rollback/sync/rescan decisions, so Store-backed runs see a stale legacy tip after a successful update and can rescan from the wrong height or skip needed rollback. Read the tip through the Store in Store mode or keep addrStore synchronized. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -500,6 +591,161 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+				return fmt.Errorf("credit output %d: %w", index,
+					db.ErrInvalidParam)
+			}
+
+			pkScript := match.Rec.MsgTx.TxOut[index].PkScript
+
+			_, err := s.store.GetAddress(
+				ctx, db.GetAddressQuery{
+					WalletID:     s.walletID,
+					ScriptPubKey: pkScript,
```
yyforyongyu: Resolving ownership by the whole output PkScript drops credits where the wallet owns an extracted member address but not the exact output script, such as bare multisig. The legacy path resolves entry.Address; use the address script for entry.Address or an equivalent Store lookup before deciding to skip, otherwise these outputs are omitted from the transaction batch. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔄

yyforyongyu:
No issues found by gpt-5.5 🪴

wallet/syncer.go
```diff
@@ -500,6 +621,209 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
 }
 
+// unminedTxns returns transactions that are still active in the wallet's
+// unmined set.
+func (s *syncer) unminedTxns(ctx context.Context) ([]*wire.MsgTx, error) {
+	if s.store == nil {
+		return s.DBGetUnminedTxns(ctx)
+	}
+
+	infos, err := s.store.ListTxns(
```
yyforyongyu, line 692: `ListTxns(UnminedOnly)` is ordered by store query order, not dependency order. The legacy path returns `wtxmgr.DependencySort`, so rebroadcast can now send a child before its parent and get rejected by peers. Sort the deserialized unmined set parents-first before returning. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -500,6 +621,209 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+		for _, entry := range match.Entries {
+			index := entry.Credit.Index
+			if uint64(index) >= uint64(len(match.Rec.MsgTx.TxOut)) {
+				return fmt.Errorf("credit output %d: %w", index,
+					db.ErrInvalidParam)
+			}
+
+			pkScript := match.Rec.MsgTx.TxOut[index].PkScript
+
+			owned, err := s.creditAddressOwned(
```
yyforyongyu: This ownership read runs before `ApplyTxBatch` opens its write transaction. Legacy notification writes resolve address ownership inside the same walletdb transaction that records the tx, but here an address import or derivation can land between this read and the batch write, causing an owned credit to be skipped. Move the ownership resolution into the store batch transaction. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -1227,7 +1557,7 @@ func (s *syncer) scanWithRewind(ctx context.Context, req *scanReq) error {
 		current.Height, req.startBlock.Height)
 
 	// Rewind the database status.
-	err := s.DBPutRewind(ctx, req.startBlock)
+	err = s.rewindToBlock(ctx, req.startBlock)
```
yyforyongyu: `rewindToBlock` now calls `RollbackToBlock`, whose Store contract is DB-wide and has no wallet ID. A manual rescan for one wallet can therefore roll back confirmed state for every wallet sharing the store. Keep targeted rescan rewind wallet-scoped, or add a wallet-scoped rollback API for this path. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🐙

wallet/db_ops.go
```diff
@@ -501,6 +501,14 @@ func (s *syncer) DBGetSyncedBlocks(_ context.Context, startHeight,
 func (s *syncer) DBPutRewind(_ context.Context,
 	bs waddrmgr.BlockStamp) error {
 
+	// SetSyncedTo below writes the addrmgr bucket and advances the live
+	// manager's in-memory synced tip immediately. If the subsequent
+	// Rollback fails, walletdb rolls the bucket write back but the in-memory
+	// tip stays rewound to a fork point that was never persisted. Snapshot
+	// the pre-rewind tip so a failed update can restore it if no newer live
+	// tip has taken its place.
+	preRewindTip := s.addrStore.SyncedTo()
```
yyforyongyu: This snapshot is taken before acquiring the walletdb write transaction. If another writer advances the live synced tip between this read and `SetSyncedTo`, a rollback failure restores that newer tip back to this stale value. Take the snapshot inside the same walletdb update immediately before the rewind. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -500,6 +621,149 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("list unmined txns: %w", err)
+	}
+
+	txSet := make(map[chainhash.Hash]*wire.MsgTx, len(infos))
+	for i := range infos {
+		info := infos[i]
+		if info.Status != db.TxStatusPending &&
```
yyforyongyu: Including `TxStatusPending` here makes rebroadcast publish transactions that callers explicitly stored before first network publication. The status docs define pending as locally-created and not yet broadcast, but `broadcastUnminedTxns` sends every tx returned from this helper. Restrict rebroadcast to `TxStatusPublished`. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -1226,8 +1496,7 @@ func (s *syncer) scanWithRewind(ctx context.Context, req *scanReq) error {
 	log.Infof("Rewinding sync status from %d to %d for rescan",
 		current.Height, req.startBlock.Height)
 
-	// Rewind the database status.
-	err := s.DBPutRewind(ctx, req.startBlock)
+	err = s.rewindToBlock(ctx, req.startBlock)
```
yyforyongyu: Manual rescans are wallet-scoped, but this now routes them through `rewindToBlock`, whose Store branch calls `RollbackToBlock`. That Store API has no wallet ID and is documented as DB-wide, deleting shared block rows and rewinding every wallet at or above the boundary. Keep user-requested rescans on a wallet-scoped Store update and reserve DB-wide rollback for chain reorg handling. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -992,7 +1256,10 @@ func (s *syncer) advanceChainSync(ctx context.Context) (bool, error) {
 	}
 
 	// Determine our current sync state.
-	syncedTo := s.addrStore.SyncedTo()
+	syncedTo, err := s.syncedTo(ctx)
```
yyforyongyu: This now reads the Store-backed sync tip when a store is configured, but the normal scan commit below still calls `DBPutSyncBatch`, which only advances the legacy addrStore tip. A backend that does not mirror that legacy write will keep returning the same Store tip here after each batch, so the loop starts from the same height and rescans indefinitely. Route scan batches through the Store or keep the read and write on the same backend until `ApplyScanBatch` is wired in. (`gpt-5.5`)

yyforyongyu:

wallet/syncer.go
```diff
@@ -1226,8 +1493,18 @@ func (s *syncer) scanWithRewind(ctx context.Context, req *scanReq) error {
 	log.Infof("Rewinding sync status from %d to %d for rescan",
 		current.Height, req.startBlock.Height)
 
-	// Rewind the database status.
-	err := s.DBPutRewind(ctx, req.startBlock)
+	if s.store == nil {
+		err = s.rewindToBlock(ctx, req.startBlock)
+	} else {
+		err = s.updateSyncTip(ctx, wtxmgr.BlockMeta{
```
yyforyongyu: This store-backed rewind only updates the wallet sync tip, while the legacy path above still uses `DBPutRewind`, which also rolls `wtxmgr` back from `startBlock+1`. After a manual rescan from an earlier height, transactions and UTXOs above the rewind point can remain confirmed in the Store even though the syncer is rescanning that range, so duplicate scan writes may be skipped/replayed against stale post-rewind state instead of rebuilding it. Route this through `rewindToBlock`/`RollbackToBlock` so the sync tip and transaction state rewind atomically. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -500,6 +621,147 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+			)
+		}
+
+		transactions = append(transactions, db.CreateTxParams{
+			WalletID:         s.walletID,
+			Tx:               &match.Rec.MsgTx,
+			Received:         match.Rec.Received,
+			Block:            block,
+			Status:           db.TxStatusPublished,
+			CreditCandidates: creditCandidates,
```
yyforyongyu, line 802: The `CreditCandidates` design here (and the `unminedTxns` `DependencySort`/filter change) belongs in the commits that introduced these paths -- `putTxNotifications` shipped an out-of-band `s.store.GetAddress` ownership read in `e10808496`, and `unminedTxns` in `b17b9cc89` -- not rewritten by the trailing `wallet: fix store-backed syncer paths` commit; fold each fix into its introducing commit.

(`claude-opus-4-8[1m]`)

wallet/syncer.go
```diff
@@ -1226,8 +1498,7 @@ func (s *syncer) scanWithRewind(ctx context.Context, req *scanReq) error {
 	log.Infof("Rewinding sync status from %d to %d for rescan",
 		current.Height, req.startBlock.Height)
 
-	// Rewind the database status.
-	err := s.DBPutRewind(ctx, req.startBlock)
+	err = s.rewindToBlock(ctx, req.startBlock)
```
yyforyongyu: Please fold this fix into the commits that introduced the Store-backed paths instead of leaving it as the trailing `wallet: fix store-backed syncer paths` commit. The stack history currently introduces the broken rewind/checkRollback routing and then fixes it later, which violates the agreed hygiene rule and makes the intermediate commits non-reviewable; squash this change and its tests into the original route/rollback commits. (`gpt-5.5`)

yyforyongyu:

wallet/syncer.go
yyforyongyu: This path is for `Wallet.Resync`, which is a single-wallet operation, but `rewindToBlock` routes Store-backed wallets to `RollbackToBlock`. That Store method is documented DB-wide and has no wallet ID, so on SQL stores it rewinds shared block rows and every wallet sync state at/above this height. A manual resync of one wallet can therefore mutate unrelated wallets that share the same store. Keep this path wallet-scoped, or add/use a wallet-scoped rewind API for manual rescans while reserving `RollbackToBlock` for chain reorgs. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🦔

wallet/syncer.go
```diff
@@ -390,6 +419,123 @@ func (s *syncer) checkRollback(ctx context.Context) error {
...
+// rewindWallet rewinds this wallet's sync and transaction state for a manual
+// rescan without requiring a DB-wide block rollback.
+func (s *syncer) rewindWallet(ctx context.Context,
+	block waddrmgr.BlockStamp) error {
+
+	if s.store == nil {
+		return s.DBPutRewind(ctx, block)
+	}
+
+	storeBlock, err := db.BlockFromBlockStamp(block)
```
yyforyongyu: Manual rescan requests only set `BlockStamp.Height`, so this conversion passes a zero hash/timestamp into `RewindWallet`. The SQL implementations then call `UpdateWalletWithOps`, which `EnsureBlock`s that block before updating `synced_height`; a rescan to a height with no existing row can insert a fake zero-hash epoch block, while a height with a real row fails `ErrBlockMismatch`. Please resolve the requested height to the real block hash/header before calling `RewindWallet`, or make `RewindWallet` take a height and derive the stored block itself. (`gpt-5.5 xhigh`)

wallet/internal/db/interface.go
```diff
@@ -420,6 +420,13 @@ type TxStore interface {
 	// ApplyScanBatch atomically records recovery scan writes for one wallet.
 	ApplyScanBatch(ctx context.Context, params ScanBatchParams) error
 
+	// RewindWallet atomically detaches one wallet's confirmed transactions at
+	// and above the block after params.Block, and updates only that wallet's
+	// synced tip to params.Block. Unlike RollbackToBlock, this method is
+	// wallet-scoped and must not delete shared block rows or mutate other
+	// wallets' sync states.
+	RewindWallet(ctx context.Context, params RewindWalletParams) error
```
yyforyongyu, line 431: `f6da77ff7` bundles the `RewindWallet` Store method (this interface entry + pg/sqlite/kvdb impls + `RewindWalletParams`), its itest, and the wallet-side routing (`rewindWallet`/`scanWithRewind` + tests) into one ~272-prod-line commit, well over the ~200 per-commit cap; split it (db method+impls / db itest / wallet routing).

(`claude-opus-4-8[1m]`)

yyforyongyu, line 431: This `f6da77ff7` commit is not reviewable as an atomic change. It adds the `RewindWallet` Store API, pg/sqlite/kvdb implementations, `RewindWalletParams`, the mock, a DB itest, and wallet-side routing/tests in one commit (~592 insertions across 10 files). Split it into separate DB API/implementation, DB test, and wallet routing/test commits so each intermediate commit has one purpose and can be reviewed independently. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔄

yyforyongyu:
No issues found by gpt-5.5 🦎

wallet/syncer.go
```diff
@@ -500,6 +682,147 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+				return fmt.Errorf("credit output %d: %w", index,
+					db.ErrInvalidParam)
+			}
+
+			creditCandidates[index] = append(
+				creditCandidates[index], entry.Address,
+			)
+		}
+
+		transactions = append(transactions, db.CreateTxParams{
```
yyforyongyu, line 795: A mempool `RelevantTx` can be an echo of a wallet transaction that was already recorded locally, but this batch always replays it as `TxStatusPublished` with an empty label. `ApplyTxBatch` only treats duplicate unmined rows as skippable when status and label match exactly, and unmined duplicates cannot be reused without a block, so an already-pending or labeled wallet tx will propagate `ErrTxAlreadyExists` and fail the sync loop. Preserve the existing pending/label metadata or make the store notification path promote/merge duplicate live unmined rows instead of treating them as hard errors. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🏷️

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` ✨

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔁

wallet/syncer.go
```diff
@@ -500,6 +682,188 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+
+	case err != nil:
+		return 0, "", fmt.Errorf("get tx %v: %w", txHash, err)
+	}
+
+	if info.Block != nil || !db.IsUnminedStatus(info.Status) {
+		return status, label, nil
+	}
+
+	return info.Status, info.Label, nil
```
yyforyongyu: A mempool echo of a locally-created transaction reaches this path with no block, but returning the existing `Pending` state makes `ApplyTxBatch` treat the echo as an idempotent duplicate instead of moving the row to `Published`. The same Store rebroadcast path added above only returns `TxStatusPublished` rows, so the transaction can remain live but never be rebroadcast after it has been observed in the mempool. Please merge this duplicate by promoting `Pending` to `Published` while preserving the label, or otherwise keep the status transition consistent with the rebroadcast filter. (`gpt-5.5`)

yyforyongyu:
No issues found by gpt-5.5 🦀

wallet/internal/db/pg/txstore_batch.go
```diff
@@ -128,12 +268,36 @@ func replayBatchDuplicateEdges(ctx context.Context, req db.CreateTxRequest,
 	return nil
 }
 
-// canSkipBatchDuplicate reports whether an existing transaction row matches the
+// promoteBatchDuplicate marks one pending duplicate row as published while
+// preserving all existing metadata, including the user-visible label.
+func promoteBatchDuplicate(ctx context.Context, qtx *sqlc.Queries,
```
yyforyongyu, line 220: `b9d5852b4 "route relevant tx notifications"` is ~863 weighted (580 prod + 567 test), ~4x the hard cap, and bundles the db-layer credit-candidate resolution + Pending->Published promotion (pg/sqlite/kvdb + `data_types` + `txstore_common`, e.g. this `promoteBatchDuplicate`) with the wallet-layer syncer routing and a 567-line test set under a `wallet:` prefix. Split into a `db:` resolution/promotion commit, a `wallet:` routing commit, and a tests commit.

(`claude-opus-4-8[1m]`)

yyforyongyu: Thanks — the split resolves the structural finding: `db: promote unmined tx duplicates` / `wallet: route relevant tx notifications` / `wallet+db: test relevant tx routing` are now separated, the wallet routing commit is ~92 (under cap), and the db-layer work is no longer under a `wallet:` prefix. The `db:` commit is still ~488 (test commit ~283), so a per-backend pg/sqlite/kvdb split (like the ApplyScanBatch commits) would bring it under the cap if you want strict enforcement — but the bundling/scope issue is fixed. Resolving.

(`claude-opus-4-8[1m]`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` ✂️

Abdulkbk (approved):
LGTM

PR #1260, opened 2026-06-03
wallet: split account methods in multiple files

GustavoStingelin:
This change only splits the account methods across multiple files to prepare for the later ops interface refactor.

It does not change behavior or logic, it only reorganizes the existing code.


GustavoStingelin:
image

Copilot changed its pricing model, and I burned through my $40 plan yesterday, so I can't request more Copilot reviews right now.

yyforyongyu (commented):
LGTM 📁 (`deepseek-v4-pro`)

yyforyongyu:
LGTM 🧩 (`gpt-5.5`)

wallet/internal/db/itest/accountstore_createderived_test.go
```diff
@@ -0,0 +1,397 @@
...
+		)
+		require.NoError(t, err)
+		require.Equal(t, accountName, info.AccountName)
+		require.Equal(t, scope, info.KeyScope)
+	}
+}
+
+// TestCreateDerivedAccountSequentialNumbers verifies that derived accounts
+// within the same scope receive sequential account numbers starting from 0.
+func TestCreateDerivedAccountSequentialNumbers(t *testing.T) {
```
yyforyongyu, line 301: This itest split dropped a CreateDerivedAccount test without relocating it: `TestCreateDerivedAccountMaxAccountNumber` existed in the base `account_store_test.go` but is gone at the PR head, with no replacement anywhere under `wallet/internal/db/itest/` (a tree-wide grep for `MaxAccountNumber` / `ErrMaxAccountNumberReached` in the itest dir is empty). It was removed by `9794b227` — "split itest **GetAccount** tests" — i.e. a different op's split commit, which is why it slipped through rather than landing here alongside the other CreateDerivedAccount itests (`SequentialNumbers`, `Concurrent`, …).

The lost coverage is the end-to-end account-number **overflow boundary** against a real pg+sqlite store: it set the scope counter to `db.MaxAccountNumber`, asserted `CreateDerivedAccount` succeeds at exactly `MaxAccountNumber`, then asserted the next allocation fails with `db.ErrMaxAccountNumberReached`. What survives (`accountstore_createderived_test.go`'s `CreateDerivedAccountWithOps` unit test + `account_derive_test.go`) exercises the overflow only through mocked ops — it never drives the real SQL `AllocateAccountNumber`/counter path, so the cross-backend boundary is now unverified.

Please carry `TestCreateDerivedAccountMaxAccountNumber` into this file so the split stays coverage-preserving.

(`claude-opus-4-8[1m]`)

yyforyongyu: let's fix this in the followup PR cc @GustavoStingelin

yyforyongyu (approved):
LGTM!

wallet/internal/db/itest/accountstore_createderived_test.go
GustavoStingelin: https://github.com/btcsuite/btcwallet/pull/1261

PR #1261, opened 2026-06-04
wallet: readd create max account number itest

GustavoStingelin:
Just re-added an item removed in the last big PR https://github.com/btcsuite/btcwallet/pull/1260.

Also:
> Looks like SQLfluff got an update, and the version is not locked in the Docker image.

so we are pinning it too.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,12 +67,17 @@ DOCKER_TOOLS = docker run \
   $(GIT_VOLUME) \
   btcwallet-tools
 
+# Pin the sqlfluff image by digest to keep linting results stable across
+# environments and over time. The tag alone is mutable, so it could later point
+# to a different build with changed rules or behavior. This digest currently
+# corresponds to sqlfluff 4.2.1 on Docker Hub:
+# https://hub.docker.com/layers/sqlfluff/sqlfluff/4.2.1
 SQLFLUFF = docker run \
 	--rm \
 	--user $$(id -u):$$(id -g) \
     -v $$(pwd):/sql \
     -w /sql \
-    sqlfluff/sqlfluff
+    sqlfluff/sqlfluff@sha256:0b1f131e3e9b4ac10bf45f1b2cb8680e67f74e63f8e6db9f0f0207dbbe3c71d2
 
 GREEN := "\\033[0;32m"
 NC := "\\033[0m"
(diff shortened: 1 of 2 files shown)


GustavoStingelin:
Looks like SQLfluff got an update, and the version is not locked in the Docker image. 

```
 Linting SQL files (postgres migrations).
docker run --rm --user $(id -u):$(id -g) -v $(pwd):/sql -w /sql sqlfluff/sqlfluff lint --config /sql/.sqlfluff --dialect postgres wallet/internal/sql/pg/migrations
== [wallet/internal/sql/pg/migrations/000006_addresses.down.sql] FAIL
L:  12 | P:   1 | PG01 | DROP statement should use CONCURRENTLY to avoid locking
                       | the table. [postgres.excessive_locks]
```

That issue is not relevant here because we are not in production yet, so the migrations will run against an empty database. However, it could become a problem for future migrations on already running SQL databases, so it would be good to improve this later.

I will add a new commit locking the image tag.

yyforyongyu (commented):
LGTM 🎯 (`deepseek-v4-pro`)

Makefile
```diff
@@ -67,12 +67,17 @@ DOCKER_TOOLS = docker run \
...
+# to a different build with changed rules or behavior. This digest currently
+# corresponds to sqlfluff 4.2.1 on Docker Hub:
+# https://hub.docker.com/layers/sqlfluff/sqlfluff/4.2.1
 SQLFLUFF = docker run \
 	--rm \
 	--user $$(id -u):$$(id -g) \
     -v $$(pwd):/sql \
     -w /sql \
-    sqlfluff/sqlfluff
+    sqlfluff/sqlfluff@sha256:0b1f131e3e9b4ac10bf45f1b2cb8680e67f74e63f8e6db9f0f0207dbbe3c71d2
```
yyforyongyu, line 80: This sqlfluff digest-pin is a sound change in its own right (mutable tag → pinned digest, good comment), but it's unrelated to this PR's stated purpose — "readd create max account number itest." The PR otherwise only adds a Go itest (`accountstore_createderived_test.go`) and touches no SQL, so the pin isn't needed for this PR's CI; it's a separate CI-tooling concern (and a different author).

Suggest splitting it into its own PR so the test re-add stays narrowly scoped and independently revertable, matching the repo's scope-matches-title convention. The `06a4bfbc2` itest re-add itself looks correct — it restores the exact dropped coverage (success at `db.MaxAccountNumber`, then `ErrMaxAccountNumberReached` on the next allocation) in the right per-op file.

(`claude-opus-4-8[1m]`)

yyforyongyu:
LGTM 🧭 (`gpt-5.5`)

yyforyongyu (approved):
LGTM🧷

PR #1263, opened 2026-06-10
db+waddrmgr: add scan-batch scaffolding and rollback support

yyforyongyu:
Adds the scan-batch scaffolding — `db.ScanHorizon` / scan-batch query types,
account-name queries, scan-horizon extension helpers, and the
`AddressInfo.HasDerivationPath` population — together with the `waddrmgr`
rollback support it builds on (`RestoreSyncedTo`, `EvictDerivedAddresses`).

Split out of the runtime Store PR (#1256) for reviewability; content-preserving,
no commit changes. Stacked on `prep-runtime-1`.

### DIFF
--- a/wallet/internal/db/data_types.go
+++ b/wallet/internal/db/data_types.go
@@ -1078,6 +1078,20 @@ type CreateTxParams struct {
 	Credits map[uint32]address.Address
 }
 
+// TxBatchParams contains a wallet transaction batch and optional sync-tip
+// update that should be applied atomically.
+type TxBatchParams struct {
+	// WalletID is the ID of the wallet receiving the batch.
+	WalletID uint32
+
+	// Transactions contains the transaction records to apply.
+	Transactions []CreateTxParams
+
+	// SyncedTo optionally records the wallet's new chain sync tip as part of
+	// the same batch.
+	SyncedTo *Block
+}
+
 // UpdateTxState contains one requested transaction-state change.
 type UpdateTxState struct {
 	// Block records the transaction as confirmed in the provided block.
--- a/wallet/internal/db/kvdb/txstore.go
+++ b/wallet/internal/db/kvdb/txstore.go
@@ -289,6 +289,95 @@ func (s *Store) UpdateTx(_ context.Context, params db.UpdateTxParams) error {
 	return nil
 }
 
+// legacyCreditEntries converts db-native credit addresses into legacy credit
+// entries and marks resolved addresses as used.
+//
+// Each non-nil credit is validated against the output script it claims to
+// credit using the same membership check the CreateTx path applies in
+// addCreateTxCredit, so the batch path cannot record a UTXO owned by an
+// address the output does not pay. A nil credit means the caller has no
+// resolved owner, so ownership is keyed on the output's own script: the entry
+// is recorded from the output index alone, with no address-manager lookup or
+// used-marking, matching CreateTx and the SQL backends.
+func (s *Store) legacyCreditEntries(addrmgrNs walletdb.ReadWriteBucket,
+	req db.CreateTxRequest) ([]wtxmgr.CreditEntry, error) {
+
+	chainParams := s.addrStore.ChainParams()
+
+	credits := make([]wtxmgr.CreditEntry, 0, len(req.Params.Credits))
+	for index, addr := range req.Params.Credits {
+		// A nil credit address has no owner to resolve, so key the
+		// credit on the output's own script: record it from the index
+		// with change cleared, skipping the address-manager lookup and
+		// the membership check that has no caller address to validate.
+		if addr == nil {
+			if int(index) >= len(req.Params.Tx.TxOut) {
+				return nil, fmt.Errorf("credit output %d: %w: "+
+					"index out of range", index,
+					db.ErrInvalidParam)
+			}
+
+			credits = append(credits, wtxmgr.CreditEntry{
+				Index:  index,
+				Change: false,
+			})
+
+			continue
+		}
+
+		// Validate the caller-supplied credit against the actual output
+		// script before trusting it, rejecting a mismatch with the same
+		// error CreateTx uses.
+		err := validateCreditAddr(
+			*req.Params.Tx, index, addr, chainParams,
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		managedAddr, err := s.addrStore.Address(addrmgrNs, addr)
+		if waddrmgr.IsError(err, waddrmgr.ErrAddressNotFound) {
+			return nil, fmt.Errorf("credit output %d: %w", index,
+				db.ErrAddressNotFound)
+		}
+
+		if err != nil {
+			return nil, fmt.Errorf("lookup credit address %d: %w",
+				index, err)
+		}
+
+		credits = append(credits, wtxmgr.CreditEntry{
+			Index:  index,
+			Change: managedAddr.Internal(),
+		})
+
+		err = s.addrStore.MarkUsed(addrmgrNs, addr)
+		if err != nil {
+			return nil, fmt.Errorf("mark credit address used %d: %w",
+				index, err)
+		}
+	}
+
+	return credits, nil
+}
+
+// kvdbTxBlockMeta converts store block metadata into legacy transaction block
+// metadata.
+func kvdbTxBlockMeta(block *db.Block) (*wtxmgr.BlockMeta, error) {
+	height, err := db.Uint32ToInt32(block.Height)
+	if err != nil {
+		return nil, fmt.Errorf("convert block height: %w", err)
+	}
+
+	return &wtxmgr.BlockMeta{
+		Block: wtxmgr.Block{
+			Hash:   block.Hash,
+			Height: height,
+		},
+		Time: block.Timestamp,
+	}, nil
+}
+
 // GetTx retrieves one wallet-scoped transaction snapshot through the legacy
 // wtxmgr query path.
 func (s *Store) GetTx(_ context.Context, query db.GetTxQuery) (
--- a/wallet/internal/db/kvdb/txstore.go
+++ b/wallet/internal/db/kvdb/txstore.go
@@ -25,6 +25,18 @@ var _ db.TxStore = (*Store)(nil)
 // signed legacy wtxmgr height domain.
 var errLegacyHeightOverflow = errors.New("legacy height overflows int32")
 
+// kvdbTxStatusBucketKey is a kvdb-owned side bucket under the legacy wtxmgr
+// namespace that records transaction statuses wtxmgr cannot represent.
+var kvdbTxStatusBucketKey = []byte("db-tx-status")
+
+// legacyTxStatusValueLen is the byte width of one encoded db.TxStatus value.
+const legacyTxStatusValueLen = 1
+
+// errLegacyTxStatusUnexpectedSize reports a corrupt status side-bucket value.
+var errLegacyTxStatusUnexpectedSize = errors.New(
+	"legacy tx status bucket value has unexpected byte width",
+)
+
 // CreateTx records an unmined transaction through the legacy wtxmgr path.
 //
 // Unlike the SQL backends, kvdb does NOT route through the shared
@@ -103,6 +115,11 @@ func (s *Store) createTxWithTx(tx walletdb.ReadWriteTx,
 
 	if exists {
 		return db.ErrTxAlreadyExists
+	}
+
+	err = putLegacyTxStatus(txmgrNs, txRec.Hash, params.Status)
+	if err != nil {
+		return fmt.Errorf("put tx status: %w", err)
 	}
 
 	if len(params.Label) != 0 {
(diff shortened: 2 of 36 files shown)


waddrmgr/scoped_manager.go
```diff
@@ -2917,6 +2917,71 @@ func (s *ScopedKeyManager) InvalidateAccountCache(account uint32) {
...
+		filtered = append(filtered, info)
+	}
+
+	// Clear the tail the in-place filter left behind so the dropped entries'
+	// managed addresses can be garbage collected.
+	for i := len(filtered); i < len(s.deriveOnUnlock); i++ {
+		s.deriveOnUnlock[i] = nil
+	}
+
+	s.deriveOnUnlock = filtered
```
yyforyongyu: This removes cached addresses and pending unlock work, but it leaves the same cached account info with the advanced next-index and last-address fields that extendAddresses set before the surrounding transaction committed. If the batch rolls back and the caller invokes this cleanup, AccountProperties/Last*Address can still report the unpersisted horizon until the account cache is invalidated. Invalidate or restore the account cache here as part of the rollback cleanup. (`gpt-5.5`)

yyforyongyu:
LGTM ✅

(`claude-opus-4-8[1m]`)

wallet/internal/db/data_types.go
```diff
@@ -35,6 +35,12 @@ const (
 	// hardened-child cap of 2^31 - 1; the -2 leaves the topmost child
 	// number for the legacy imported-account sentinel.
 	MaxAccountNumber uint32 = (1 << 31) - 2 //nolint:mnd
+
+	// MaxAddressIndex is the largest derived address child index the store may
+	// allocate on a branch. It matches waddrmgr.MaxAddressesPerAccount
+	// (hdkeychain.HardenedKeyStart - 1), so recovery horizon extension rejects
+	// the same out-of-range index the legacy address manager would.
+	MaxAddressIndex uint32 = (1 << 31) - 1 //nolint:mnd
```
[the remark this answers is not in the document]
yyforyongyu: Addressed — reworded the `MaxAddressIndex` doc comment to describe it as the recovery-horizon extension bound (not a global allocation limit), and added `TestExtendScanHorizon*` covering five scenarios.

wallet/internal/db/scan_batch_common.go
```diff
@@ -0,0 +1,226 @@
...
+//     instead of failing, so the SQL path matches the kvdb invalid-child skip.
+//   - After derivation the branch next-index advances to one past the last
+//     derived child, leaving the same terminal counter the legacy path would.
+func ExtendScanHorizon(ctx context.Context, ops ScanHorizonOps,
+	deriveFn AddressDerivationFunc, horizon ScanHorizon) error {
+
+	branch, err := validateHorizon(deriveFn, horizon)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
yyforyongyu: Addressed — reworded the `MaxAddressIndex` doc comment to describe it as the recovery-horizon extension bound (not a global allocation limit), and added `TestExtendScanHorizon*` covering five scenarios.

waddrmgr/scoped_manager.go
yyforyongyu: Addressed — `EvictDerivedAddresses` now also evicts the affected account from the in-memory account cache (`delete(s.acctInfo, account)`) so it reloads from the database on next read; covered by a derived-address-cache eviction test (verified to fail without the fix).

wallet/internal/db/pg/txstore_batch.go
```diff
@@ -0,0 +1,109 @@
...
+			if err != nil {
+				return fmt.Errorf("validate tx %d: %w", i, err)
+			}
+
+			err = db.CreateTxWithOps(ctx, req, &createTxOps{
+				invalidateUnminedTxOps: invalidateUnminedTxOps{
+					qtx: qtx,
+				},
+			})
+			if errors.Is(err, db.ErrTxAlreadyExists) {
```
yyforyongyu: Ignoring every `ErrTxAlreadyExists` drops non-idempotent duplicate observations such as a pending unmined row later reported as published while the batch can still advance sync state, so reconcile valid duplicate status/block transitions or only continue for truly idempotent duplicates. (`gpt-5.5 xhigh`)

wallet/internal/db/sqlite/txstore_batch.go
```diff
@@ -0,0 +1,106 @@
...
+			if err != nil {
+				return fmt.Errorf("validate tx %d: %w", i, err)
+			}
+
+			err = db.CreateTxWithOps(ctx, req, &createTxOps{
+				invalidateUnminedTxOps: invalidateUnminedTxOps{
+					qtx: qtx,
+				},
+			})
+			if errors.Is(err, db.ErrTxAlreadyExists) {
```
yyforyongyu: Ignoring every `ErrTxAlreadyExists` drops non-idempotent duplicate observations such as a pending unmined row later reported as published while the batch can still advance sync state, so reconcile valid duplicate status/block transitions or only continue for truly idempotent duplicates. (`gpt-5.5 xhigh`)

wallet/internal/db/pg/txstore_batch.go
yyforyongyu: Still applies at current head: continuing on every `ErrTxAlreadyExists` still skips a pending-to-published unmined observation, so the row remains pending while the batch can advance sync state. Please update or reconcile non-idempotent duplicates instead of treating all duplicates as no-ops. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_batch.go
yyforyongyu: Still applies at current head: continuing on every `ErrTxAlreadyExists` still skips a pending-to-published unmined observation, so the row remains pending while the batch can advance sync state. Please update or reconcile non-idempotent duplicates instead of treating all duplicates as no-ops. (`gpt-5.5`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔧

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔁

wallet/internal/db/tx_store_common.go
```diff
@@ -1111,6 +1153,10 @@ func CreateTxWithOps(ctx context.Context, req CreateTxRequest,
 	}
 
 	if foundExisting {
+		if checkIdempotentCreateTx(req, *existing) {
```
yyforyongyu: When the same unmined tx is seen again with a different unmined status or label, `checkIdempotentCreateTx` returns false but `checkReuseCreateTx` still returns true because the existing row is an unmined non-coinbase. The code then calls `ConfirmExisting` with `req.Params.Block == nil`; both SQL implementations immediately pass that nil block to `requireBlockMatches`, which dereferences it and panics. Gate reuse on an incoming block or handle unmined duplicate updates explicitly. (`gpt-5.5`)

yyforyongyu: This metadata-only duplicate path skips `InsertCredits` and `MarkInputsSpent`, so a repeated observation that newly marks an output or parent input as wallet-owned is accepted but leaves the UTXO set unreconciled; either compare/reconcile credits and spend edges before returning or reject the duplicate. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🧪

wallet/internal/db/tx_store_common.go
yyforyongyu: Still applies at current head for the pending-to-confirmed reuse path: when an existing unmined row is confirmed, `ConfirmExisting` updates only the block/status and returns, so a duplicate confirmation that newly supplies `Credits` or wallet-owned inputs is accepted without recording those edges. Replay `recordCreateTxEdges` after the confirm (with conflict reconciliation before `MarkInputsSpent`) or reject confirmations that would add edges. (`gpt-5.5`)

wallet/internal/db/tx_store_common.go
```diff
@@ -898,6 +904,42 @@ func checkReuseCreateTx(req CreateTxRequest,
...
+
+	if req.Params.Block == nil {
+		return !existing.HasBlock
+	}
+
+	if !existing.HasBlock || existing.BlockHeight == nil {
+		return false
+	}
+
+	return *existing.BlockHeight == req.Params.Block.Height
```
yyforyongyu: This accepts a confirmed duplicate solely by block height, so the same tx reported with the same height but a different block hash skips `requireBlockMatches` and returns success; validate the incoming block or compare full block metadata before treating it as idempotent. (`gpt-5.5 xhigh`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,378 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	if existing != nil && existing.Block.Height >= 0 {
+		return nil
+	}
+
+	err = s.txStore.InsertUnconfirmedTx(txmgrNs, txRec, credits)
+	if err != nil {
+		return fmt.Errorf("insert unconfirmed tx: %w", err)
+	}
+
+	err = putLegacyTxStatus(txmgrNs, txRec.Hash, req.Params.Status)
```
yyforyongyu, line 632: For an existing unconfirmed tx, `InsertUnconfirmedTx` tolerates the duplicate and this status write then lets a batch change pending to published or mutate the label while advancing the sync tip; compare existing metadata and return `ErrTxAlreadyExists` unless the duplicate is truly idempotent. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🧵

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,482 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+}
+
+// applyUnconfirmedDuplicate handles an unmined batch notification for a
+// transaction hash already present in the legacy store.
+func (s *Store) applyUnconfirmedDuplicate(
+	addrmgrNs walletdb.ReadBucket, txmgrNs walletdb.ReadBucket,
+	existing *wtxmgr.TxDetails, txHash chainhash.Hash,
+	req db.CreateTxRequest) error {
+
+	if existing.Block.Height >= 0 {
```
yyforyongyu, line 647: kvdb treats an unmined batch notification for an already-confirmed tx as a no-op here, but the SQL backend rejects that same case with `ErrTxAlreadyExists` (`checkReuseCreateTx` returns false for `req.Block == nil`, so a confirmed existing row falls through to the duplicate error) — so the same `ApplyTxBatch` input (e.g. a rescan re-observing a confirmed tx in the mempool) succeeds on kvdb but fails and rolls back the batch on SQL, and no conformance itest covers it; reconcile the two backends on this case and add a conformance test.

(`claude-opus-4-8[1m]`)

wallet/internal/db/tx_store_common.go
```diff
@@ -898,6 +907,45 @@ func checkReuseCreateTx(req CreateTxRequest,
...
+		return !existing.HasBlock
+	}
+
+	if !existing.HasBlock || existing.BlockHeight == nil ||
+		existing.BlockHash == nil {
+
+		return false
+	}
+
+	return *existing.BlockHeight == req.Params.Block.Height &&
```
yyforyongyu, line 925: Include the stored block timestamp in this match (or re-run `requireBlockMatches`), because the current height/hash-only idempotent path accepts a duplicate whose timestamp conflicts with the block row instead of returning `ErrBlockMismatch` (`gpt-5.5 xhigh`).

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,482 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	if err != nil {
+		return err
+	}
+
+	block, err := kvdbTxBlockMeta(req.Params.Block)
+	if err != nil {
+		return err
+	}
+
+	err = s.txStore.InsertConfirmedTx(txmgrNs, txRec, block, credits)
```
yyforyongyu, line 520: Guard confirmed duplicates before this call, because `InsertConfirmedTx` only treats hash+block as a duplicate, so the same wallet tx hash can be accepted in a different block and advance the batch instead of returning `ErrTxAlreadyExists` (`gpt-5.5 xhigh`).

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🤝

yyforyongyu:
No issues found by gpt-5.5 🪶

wallet/internal/db/tx_store_common.go
```diff
@@ -1095,6 +1163,54 @@ func insertCreateTx(ctx context.Context, req CreateTxRequest,
...
+	if checkIdempotentCreateTx(req, existing) {
+		if req.Params.Block != nil {
+			err := ops.PrepareBlock(ctx, req)
+			if err != nil {
+				return fmt.Errorf("prepare duplicate block "+
+					"assignment: %w", err)
+			}
+		}
+
+		err := recordCreateTxEdges(ctx, req, existing.ID, ops)
```
yyforyongyu: Reconcile existing children before replaying new credits here, because after a parent was first recorded without a credit, a child could already have treated that outpoint as external and this later replay inserts an unspent UTXO for an output already spent by wallet history (`gpt-5.5 xhigh`).

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,545 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	if params.SyncedTo == nil {
+		return nil
+	}
+
+	block, err := db.BlockStampFromBlock(params.SyncedTo)
+	if err != nil {
+		return err
+	}
+
+	err = s.addrStore.SetSyncedTo(ns, &block)
```
yyforyongyu, line 451: If the surrounding `walletdb.Update` fails after this call, `SetSyncedTo` has already advanced the in-memory tip while the batch writes roll back, so snapshot the previous tip and restore it on update failure. (`gpt-5.5 xhigh`)

wallet/internal/db/pg/txstore_createtx.go
```diff
@@ -400,6 +424,138 @@ func insertCredits(ctx context.Context, qtx *sqlc.Queries,
...
+	if err != nil {
+		return fmt.Errorf("mark existing child %d input %d: %w",
+			childTxID, inputIndex, err)
+	}
+
+	if rowsAffected != 0 {
+		return nil
+	}
+
+	err = ensureSpendConflict(
```
yyforyongyu, line 833: When two active children already spend the late-discovered credit, the first `MarkUtxoSpent` claims the UTXO and the second reaches this conflict path, aborting instead of letting a confirmed child replace the earlier unmined child; collect matching children and run the confirmed/unmined conflict reconciliation before mutating the spend edge. (`gpt-5.5 xhigh`)

wallet/internal/db/sqlite/txstore_createtx.go
```diff
@@ -393,6 +417,130 @@ func insertCredits(ctx context.Context, qtx *sqlc.Queries,
...
+	if err != nil {
+		return fmt.Errorf("mark existing child %d input %d: %w",
+			childTxID, inputIndex, err)
+	}
+
+	if rowsAffected != 0 {
+		return nil
+	}
+
+	err = ensureSpendConflict(
```
yyforyongyu, line 816: When two active children already spend the late-discovered credit, the first `MarkUtxoSpent` claims the UTXO and the second reaches this conflict path, aborting instead of letting a confirmed child replace the earlier unmined child; collect matching children and run the confirmed/unmined conflict reconciliation before mutating the spend edge. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🧪

wallet/internal/db/pg/txstore_createtx.go
```diff
@@ -400,6 +445,244 @@ func insertCredits(ctx context.Context, qtx *sqlc.Queries,
...
+	parentHash := params.Tx.TxHash()
+
+	childSpends, err := collectExistingChildSpends(
+		rows, parentHash, params, txID,
+	)
+	if err != nil {
+		return err
+	}
+
+	for _, spends := range childSpends {
```
yyforyongyu: Because this iterates a snapshot by map group while each group can replace/fail other children, a failed descendant can be re-attached later. Example: child A and confirmed child C spend one late credit, while descendant B also spends another late credit; if A/C’s group runs first, B is marked failed, then B’s own group can still call `MarkUtxoSpent` from the stale snapshot. Reconcile all replacements before attaching any unmined-only spend, or re-check the child is still active before marking it. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_createtx.go
```diff
@@ -393,6 +438,234 @@ func insertCredits(ctx context.Context, qtx *sqlc.Queries,
...
+	parentHash := params.Tx.TxHash()
+
+	childSpends, err := collectExistingChildSpends(
+		rows, parentHash, params, txID,
+	)
+	if err != nil {
+		return err
+	}
+
+	for _, spends := range childSpends {
```
yyforyongyu: This backend copy has the same stale-snapshot ordering: a prior group can replace an unmined root and fail descendant B, then a later group still attaches B as the spender of another late credit from the pre-replacement snapshot. Reconcile replacements before single-child attachments or re-check the child is active before marking it. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
yyforyongyu: Still applies at current head: the update closure still calls `SetSyncedTo` at the end, so a commit/write failure after the closure returns can leave the in-memory tip advanced while the bucket transaction rolls back. Snapshot and restore the previous tip around the update, or defer the in-memory advance until commit success. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_createtx.go
yyforyongyu: Because this loop reconciles each credited outpoint from the pre-mutation child snapshot, a child replaced while handling one output can be reattached as spender of a later output it also spends; compute replacements across all credited outpoints before marking spend edges or skip tx IDs already replaced. (`gpt-5.5 xhigh`)

wallet/internal/db/pg/txstore_createtx.go
yyforyongyu: Because this loop reconciles each credited outpoint from the pre-mutation child snapshot, a child replaced while handling one output can be reattached as spender of a later output it also spends; compute replacements across all credited outpoints before marking spend edges or skip tx IDs already replaced. (`gpt-5.5 xhigh`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,545 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	if err != nil {
+		return fmt.Errorf("insert confirmed tx: %w", err)
+	}
+
+	err = putLegacyTxStatus(txmgrNs, txRec.Hash, req.Params.Status)
+	if err != nil {
+		return fmt.Errorf("put tx status: %w", err)
+	}
+
+	return s.putLegacyTxLabel(txmgrNs, txRec, req.Params.Label)
```
yyforyongyu: When `InsertConfirmedTx` promotes an existing unmined row, this writes the incoming confirmation label over the original label, unlike the SQL promotion path; detect an existing unmined tx and preserve its stored label. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🧭

yyforyongyu:
No issues found by `gpt-5.5 xhigh` 🔎

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🪞

yyforyongyu:
No issues found by gpt-5.5 🦥

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,576 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	params.Transactions = db.SortTxBatchParentsFirst(params.Transactions)
+
+	if s.addrStore == nil {
+		return fmt.Errorf("kvdb.Store.ApplyTxBatch: %w",
+			errMissingAddrStore)
+	}
+
+	var preBatchSyncedTo waddrmgr.BlockStamp
+	if params.SyncedTo != nil {
+		preBatchSyncedTo = s.addrStore.SyncedTo()
```
yyforyongyu: `preBatchSyncedTo` is captured before `walletdb.Update` starts its serialized write transaction. A concurrent batch can commit a newer tip while this call waits; if this update then fails after `SetSyncedTo`, the restore below rewinds memory to the stale pre-lock tip. Capture the snapshot inside the write transaction immediately before the sync-tip update, or restore conditionally only if the current tip is this batch’s attempted tip. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🦫

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,586 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	}
+
+	var syncTipRestore batchSyncTipRestore
+
+	err = walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
+		return s.applyLegacyBatchUpdate(tx, params, &syncTipRestore)
+	})
+	if err != nil {
+		if syncTipRestore.snapshotTaken {
+			s.addrStore.RestoreSyncedTo(syncTipRestore.previous)
```
yyforyongyu: This restore runs after `walletdb.Update` has returned, so it is no longer serialized with other writers. If another batch commits and advances the live tip between this failed update returning and this call, `RestoreSyncedTo` overwrites memory with the stale pre-failure tip while the database has the newer tip; restore conditionally only if the live tip is still this batch’s attempted tip, or serialize the restore with the write path. (`gpt-5.5 xhigh`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,586 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+
+	// Record any in-batch parent before its children. The confirmed path
+	// records a debit only against an already-inserted parent credit, so a
+	// confirmed child applied before its in-batch parent would find no credit
+	// to spend and silently leave the parent output unspent. Sorting parents
+	// first makes the batch order-independent; an already parents-first or
+	// dependency-free batch is returned unchanged.
+	params.Transactions = db.SortTxBatchParentsFirst(params.Transactions)
+
+	if s.addrStore == nil {
```
yyforyongyu: This rejects creditless transaction-only batches when the store has no address manager, even though the `CreateTx` path only requires `addrStore` when credits are present and the SQL batch path can record creditless transactions. Gate the address-manager requirement on `SyncedTo` or non-empty credits, and avoid opening the addrmgr namespace for pure transaction batches. (`gpt-5.5 xhigh`)

wallet/internal/db/tx_store_common.go
```diff
@@ -1095,6 +1205,54 @@ func insertCreateTx(ctx context.Context, req CreateTxRequest,
...
+// handleExistingCreateTx handles a CreateTx request for a transaction hash that
+// already has a wallet-scoped row.
+func handleExistingCreateTx(ctx context.Context, req CreateTxRequest,
+	existing CreateTxExistingTarget, ops CreateTxOps) error {
+
+	if canIgnoreUnminedConfirmedDuplicate(req, existing) {
+		return nil
+	}
+
+	if checkIdempotentCreateTx(req, existing) {
```
yyforyongyu, line 1236: This idempotence check ignores `Credits`. On the replay path, the SQL `InsertCredits` implementations skip already-existing UTXOs before validating the requested credit address, so a duplicate with `Credits[index]` pointing at the wrong wallet address is accepted and the batch can still advance sync state. Compare credit metadata or validate every requested credit before treating the duplicate as idempotent. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔁

wallet/internal/db/pg/txstore_batch.go
```diff
@@ -0,0 +1,104 @@
...
+
+// applyBatchSyncTip applies the optional sync-tip update within a batch.
+func applyBatchSyncTip(ctx context.Context, qtx *sqlc.Queries,
+	params db.TxBatchParams) error {
+
+	if params.SyncedTo == nil {
+		return nil
+	}
+
+	err := ensureBlockExists(ctx, qtx, params.SyncedTo)
```
yyforyongyu, line 102: This only ensures the final `SyncedTo` block exists. A batch can also carry confirmed transactions from earlier same-batch blocks, and `CreateTxWithOps` still requires each confirming block row during `PrepareBlock`. Ensure every transaction `Block` is materialized before creating the tx, not just the sync tip. (`gpt-5.5 xhigh`)

wallet/internal/db/sqlite/txstore_batch.go
```diff
@@ -0,0 +1,101 @@
...
+
+// applyBatchSyncTip applies the optional sync-tip update within a batch.
+func applyBatchSyncTip(ctx context.Context, qtx *sqlc.Queries,
+	params db.TxBatchParams) error {
+
+	if params.SyncedTo == nil {
+		return nil
+	}
+
+	err := ensureBlockExists(ctx, qtx, params.SyncedTo)
```
yyforyongyu, line 102: This only ensures the final `SyncedTo` block exists. A batch can also carry confirmed transactions from earlier same-batch blocks, and `CreateTxWithOps` still requires each confirming block row during `PrepareBlock`. Ensure every transaction `Block` is materialized before creating the tx, not just the sync tip. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🦉

yyforyongyu:

yyforyongyu:

wallet/internal/db/tx_store_common.go
```diff
@@ -898,6 +907,72 @@ func checkReuseCreateTx(req CreateTxRequest,
...
+
+// validateCreateTxCreditRequests validates every caller-supplied credit before
+// an existing tx row is treated as an idempotent duplicate.
+func validateCreateTxCreditRequests(req CreateTxRequest) error {
+	for index, addr := range req.Params.Credits {
+		if addr == nil {
+			continue
+		}
+
+		err := ValidateCreditAddrMembership(
```
yyforyongyu, line 957: This check only proves the requested address appears in the output script; it does not prove the credit is wallet-owned or matches the UTXO already stored for this outpoint. For example, an exact duplicate of a bare-multisig tx already credited to wallet member A can request non-wallet member B; membership passes, then `InsertCredits` skips because `creditExists` is true, so `CreateTx` succeeds even though a fresh insert would reject B with `ErrAddressNotFound`. Validate duplicate credits against the stored UTXO/address before treating them as idempotent. (`gpt-5.5`)

yyforyongyu:

wallet/internal/db/pg/txstore_createtx.go
```diff
@@ -346,75 +371,408 @@ func creditLookupScript(params db.CreateTxParams, index uint32) ([]byte,
...
+	params db.CreateTxParams, spends []existingChildSpend) error {
+
+	confirmed, unmined := splitExistingChildSpends(spends)
+	if len(confirmed) > 1 {
+		return db.ErrTxInputConflict
+	}
+
+	if len(confirmed) == 0 {
+		if len(unmined) != 1 {
+			return db.ErrTxInputConflict
```
yyforyongyu, line 722: This can reject a valid late-credit reconciliation before the later outputs have a chance to replace stale children. For example, if unmined child U spends outputs 0 and 1, unmined child V spends output 0, and confirmed child C spends output 1, processing output 0 first sees two unmined spends and returns `ErrTxInputConflict`; processing output 1 first would replace U with C, leaving V as the only active spender of output 0. Compute all confirmed-vs-unmined replacements across the credited outputs before enforcing the single-unmined-spender check, or skip children scheduled for replacement. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_createtx.go
```diff
@@ -344,64 +369,387 @@ func creditLookupScript(params db.CreateTxParams, index uint32) ([]byte,
...
+	params db.CreateTxParams, spends []existingChildSpend) error {
+
+	confirmed, unmined := splitExistingChildSpends(spends)
+	if len(confirmed) > 1 {
+		return db.ErrTxInputConflict
+	}
+
+	if len(confirmed) == 0 {
+		if len(unmined) != 1 {
+			return db.ErrTxInputConflict
```
yyforyongyu, line 715: This can reject a valid late-credit reconciliation before the later outputs have a chance to replace stale children. For example, if unmined child U spends outputs 0 and 1, unmined child V spends output 0, and confirmed child C spends output 1, processing output 0 first sees two unmined spends and returns `ErrTxInputConflict`; processing output 1 first would replace U with C, leaving V as the only active spender of output 0. Compute all confirmed-vs-unmined replacements across the credited outputs before enforcing the single-unmined-spender check, or skip children scheduled for replacement. (`gpt-5.5`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔧

wallet/internal/db/pg/txstore_batch.go
yyforyongyu: This still continues on every `ErrTxAlreadyExists`. `CreateTxWithOps` uses that sentinel for non-idempotent duplicates too, such as an already-confirmed row or an unmined duplicate that should be rejected, so the batch can silently drop that notification while still advancing the sync tip. Only continue for a distinct/proven idempotent duplicate result. (`gpt-5.5`)

wallet/internal/db/pg/txstore_batch.go
```diff
@@ -0,0 +1,109 @@
...
```
yyforyongyu, line 102: This only materializes the final sync-tip block. A batch may include confirmed transactions from earlier blocks, and `CreateTxWithOps` still calls `PrepareBlock` for each transaction; if that block row is not already stored, the otherwise valid batch fails with `ErrBlockNotFound`. Insert/validate all transaction `Block` values before processing the transactions. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_batch.go
```diff
@@ -0,0 +1,106 @@
...
```
yyforyongyu, line 102: This only materializes the final sync-tip block. A batch may include confirmed transactions from earlier blocks, and `CreateTxWithOps` still calls `PrepareBlock` for each transaction; if that block row is not already stored, the otherwise valid batch fails with `ErrBlockNotFound`. Insert/validate all transaction `Block` values before processing the transactions. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_batch.go
yyforyongyu: This still continues on every `ErrTxAlreadyExists`. `CreateTxWithOps` uses that sentinel for non-idempotent duplicates too, such as an already-confirmed row or an unmined duplicate that should be rejected, so the batch can silently drop that notification while still advancing the sync tip. Only continue for a distinct/proven idempotent duplicate result. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,378 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	if params.SyncedTo == nil {
+		return nil
+	}
+
+	block, err := db.BlockStampFromBlock(params.SyncedTo)
+	if err != nil {
+		return err
+	}
+
+	err = s.addrStore.SetSyncedTo(ns, &block)
```
yyforyongyu, line 451: `SetSyncedTo` mutates the live address-manager tip before the enclosing `walletdb.Update` has committed. If the update fails while committing after this point, the DB rolls back but memory remains advanced, so later sync decisions can skip blocks that were never persisted. Snapshot the previous tip and restore it on update error, or defer the in-memory mutation until commit success. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,378 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+
+	// Record any in-batch parent before its children. The confirmed path
+	// records a debit only against an already-inserted parent credit, so a
+	// confirmed child applied before its in-batch parent would find no credit
+	// to spend and silently leave the parent output unspent. Sorting parents
+	// first makes the batch order-independent; an already parents-first or
+	// dependency-free batch is returned unchanged.
+	params.Transactions = db.SortTxBatchParentsFirst(params.Transactions)
+
+	if s.addrStore == nil {
```
yyforyongyu: This requires an address manager for any non-empty batch, even when the batch has only creditless transactions and no `SyncedTo`. `CreateTx` explicitly supports creditless tx inserts without `addrStore`, and SQL can record those tx-only batches, so kvdb rejects valid inputs. Gate the requirement on `SyncedTo` or non-empty credits, and avoid `ChainParams` lookup for empty credit maps. (`gpt-5.5`)

yyforyongyu:

yyforyongyu:

wallet/internal/db/pg/txstore_batch.go
yyforyongyu, line 102: This only creates the sync-tip block. If a batch contains a confirmed transaction in an earlier block that is not already in the block table, CreateTx will fail with `ErrBlockNotFound` even though the batch carries that block metadata. Materialize each transaction block before inserting the txs. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_batch.go
yyforyongyu, line 102: This only creates the sync-tip block. If a batch contains a confirmed transaction in an earlier block that is not already in the block table, CreateTx will fail with `ErrBlockNotFound` even though the batch carries that block metadata. Materialize each transaction block before inserting the txs. (`gpt-5.5`)

wallet/internal/db/pg/txstore_batch.go
yyforyongyu: Continuing on any `ErrTxAlreadyExists` makes the batch treat non-idempotent duplicates as success, so a conflicting duplicate can be skipped while the sync tip still advances. Route duplicates through CreateTx idempotent replay checks, or only ignore the error after verifying the stored row, credits, and spends match. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
yyforyongyu: This requires an address store for every non-empty batch, even when the batch only contains transactions with no wallet credits and no `SyncedTo` update. That rejects valid tx-only batches that the legacy tx store can apply without address-manager access; gate this on credits or a sync-tip update instead. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_batch.go
yyforyongyu: Continuing on any `ErrTxAlreadyExists` makes the batch treat non-idempotent duplicates as success, so a conflicting duplicate can be skipped while the sync tip still advances. Route duplicates through CreateTx idempotent replay checks, or only ignore the error after verifying the stored row, credits, and spends match. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
yyforyongyu, line 451: SetSyncedTo updates the live address-manager cache before the enclosing `walletdb.Update` commits. If the commit later fails, the DB rolls back but memory stays advanced, so later reads and rollbacks observe a tip that was never committed. Snapshot the prior tip and restore it on update or commit failure. (`gpt-5.5`)

yyforyongyu:

wallet/internal/db/pg/txstore_batch.go
yyforyongyu, line 102: This only creates the sync-tip block. If a batch contains a confirmed transaction in an earlier block that is not already in the block table, CreateTx will fail with `ErrBlockNotFound` even though the batch carries that block metadata. Materialize each transaction block before inserting the txs. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
yyforyongyu, line 451: SetSyncedTo updates the live address-manager cache before the enclosing `walletdb.Update` commits. If the commit later fails, the DB rolls back but memory stays advanced, so later reads and rollbacks observe a tip that was never committed. Snapshot the prior tip and restore it on update or commit failure. (`gpt-5.5`)

wallet/internal/db/pg/txstore_batch.go
yyforyongyu: Continuing on any `ErrTxAlreadyExists` makes the batch treat non-idempotent duplicates as success, so a conflicting duplicate can be skipped while the sync tip still advances. Route duplicates through CreateTx idempotent replay checks, or only ignore the error after verifying the stored row, credits, and spends match. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
yyforyongyu: This requires an address store for every non-empty batch, even when the batch only contains transactions with no wallet credits and no `SyncedTo` update. That rejects valid tx-only batches that the legacy tx store can apply without address-manager access; gate this on credits or a sync-tip update instead. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_batch.go
yyforyongyu, line 102: This only creates the sync-tip block. If a batch contains a confirmed transaction in an earlier block that is not already in the block table, CreateTx will fail with `ErrBlockNotFound` even though the batch carries that block metadata. Materialize each transaction block before inserting the txs. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_batch.go
yyforyongyu: Continuing on any `ErrTxAlreadyExists` makes the batch treat non-idempotent duplicates as success, so a conflicting duplicate can be skipped while the sync tip still advances. Route duplicates through CreateTx idempotent replay checks, or only ignore the error after verifying the stored row, credits, and spends match. (`gpt-5.5`)

wallet/internal/db/pg/txstore_batch.go
```diff
@@ -0,0 +1,126 @@
...
+	err = db.ValidateBatchTransactionsTx(params.Transactions)
+	if err != nil {
+		return err
+	}
+
+	return s.execWrite(ctx, func(qtx *sqlc.Queries) error {
+		// CreateTxWithOps needs each confirming block row during
+		// PrepareBlock, so materialize all batch transaction blocks before
+		// any transaction is created.
+		err := ensureBatchTxBlocks(ctx, qtx, params.Transactions)
```
yyforyongyu, line 37: This review-fix code is still introduced by a tail commit after the backend ApplyTxBatch commits. Fold the block-materialization and duplicate-replay fixes into the commits that introduced the batch/CreateTx paths so the PR does not introduce broken behavior and repair it later. (`gpt-5.5`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -312,12 +329,614 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	}
+
+	var syncTipRestore batchSyncTipRestore
+
+	err = walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
+		return s.applyLegacyBatchUpdate(tx, params, &syncTipRestore)
+	})
+	if err != nil {
+		if syncTipRestore.snapshotTaken {
+			s.addrStore.RestoreSyncedToIfCurrent(
```
yyforyongyu, line 363: This has the same ABA race after the failed update releases the walletdb writer: another batch can commit the same attempted tip before this restore runs, so the value match succeeds and memory is rewound to the stale pre-batch tip. Keep the restore serialized with the write path or gate it on persisted state/version. (`gpt-5.5`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 📦

yyforyongyu:

yyforyongyu:
No issues found by gpt-5.5 🛰️

yyforyongyu:

GustavoStingelin (approved):
LGTM!

PR #1264, opened 2026-06-11
chain: add SubmitPackage to the chain.Interface

ellemouton:
### Summary

Adds `SubmitPackage` to `chain.Interface`, exposing Bitcoin Core's
`submitpackage` RPC so callers can submit a topologically-sorted package of
transactions (unconfirmed parents first, child last) for atomic validation and
acceptance.

This is what lets a **v3/TRUC** transaction package relay: a zero-fee parent
carrying an ephemeral (P2A) anchor is below the minimum relay fee and is
rejected by `sendrawtransaction` on its own, but is accepted when submitted
together with a fee-paying CPFP child whose combined feerate clears policy.
`TestMempoolAccept` already mirrors this kind of pass-through to a Core RPC;
`SubmitPackage` is the broadcasting counterpart.

### Details

- **`BitcoindClient`** forwards to `rpcclient`'s typed `SubmitPackage` method,
  keeping the multi-backend RPC layering intact — the `chain` package calls
  `rpcclient` rather than issuing its own `RawRequest`. Errors are mapped
  through `MapRPCErr` for parity with `SendRawTransaction`.
- The **btcd `RPCClient`** and **`NeutrinoClient`** return `ErrUnimplemented`,
  the same way `TestMempoolAccept` does on neutrino: btcd has no `submitpackage`
  server handler yet, and a neutrino light client has no mempool and so cannot
  validate or atomically accept a package. The method's contract therefore
  stays uniform across backends — it either accepts a package or reports that
  it is unimplemented, and never fabricates a success.
- `maxFeeRate` is the optional per-transaction fee-rate ceiling in BTC/kvB
  (`nil` leaves the node default unchanged; pass a pointer to `0` to disable the
  limit, which a high-feerate CPFP child requires).

### Dependency

The typed `rpcclient.SubmitPackage` method this builds on landed in
btcsuite/btcd#2551 and ships in the tagged **btcd v0.26.0**, which `master`
already depends on — so this PR carries no `replace` directives or dependency
bumps of its own; it just uses the tagged release.

### Testing

`go build ./...` and `go vet ./...` pass. The path was exercised end-to-end
against a regtest `bitcoind` v29: a zero-fee TRUC parent plus a fee-paying CPFP
child submitted as a package is accepted and confirms, where the parent alone
is rejected for `min relay fee not met`.

### DIFF
--- a/chain/bitcoind_client.go
+++ b/chain/bitcoind_client.go
@@ -257,6 +257,30 @@ func (c *BitcoindClient) TestMempoolAccept(txns []*wire.MsgTx,
 	return c.chainConn.client.TestMempoolAccept(txns, maxFeeRate)
 }
 
+// SubmitPackage submits a package of related transactions (topologically
+// sorted, parents first and child last) to bitcoind's mempool for atomic
+// validation and acceptance via the submitpackage RPC. This is what allows a
+// zero-fee v3/TRUC parent to be accepted when paired with a fee-paying CPFP
+// child, which sendrawtransaction (single-tx) rejects.
+//
+// maxFeeRate is the optional per-tx fee-rate ceiling in BTC/kvB (pass a
+// pointer to 0 to disable the limit for high-feerate CPFP children).
+//
+// NOTE: This is part of the chain.Interface interface.
+func (c *BitcoindClient) SubmitPackage(txns []*wire.MsgTx,
+	maxFeeRate *float64) (*btcjson.SubmitPackageResult, error) {
+
+	// The rpcclient SubmitPackage exposes a maxBurnAmount limit too, which
+	// we don't surface on chain.Interface; pass nil to use the node
+	// default.
+	result, err := c.chainConn.client.SubmitPackage(txns, maxFeeRate, nil)
+	if err != nil {
+		return nil, c.MapRPCErr(err)
+	}
+
+	return result, nil
+}
+
 // Notifications returns a channel to retrieve notifications from.
 //
 // NOTE: This is part of the chain.Interface interface.
--- a/chain/btcd.go
+++ b/chain/btcd.go
@@ -204,6 +204,19 @@ func (c *RPCClient) BackEnd() string {
 	return "btcd"
 }
 
+// SubmitPackage is unimplemented for the btcd backend. btcd defines the
+// submitpackage JSON-RPC command type but its RPC server registers no handler
+// for it, so a RawRequest would fail with a method-not-found error. Returning
+// ErrUnimplemented avoids advertising support that does not exist; a btcd
+// server-side submitpackage handler must land before this can do anything.
+//
+// NOTE: This is part of the chain.Interface interface.
+func (c *RPCClient) SubmitPackage(_ []*wire.MsgTx,
+	_ *float64) (*btcjson.SubmitPackageResult, error) {
+
+	return nil, ErrUnimplemented
+}
+
 // Start attempts to establish a client connection with the remote server.
 // If successful, handler goroutines are started to process notifications
 // sent by the server.  After a limited number of connection attempts, this
--- a/chain/neutrino.go
+++ b/chain/neutrino.go
@@ -239,6 +239,25 @@ func (s *NeutrinoClient) TestMempoolAccept(txns []*wire.MsgTx,
 	return nil, ErrUnimplemented
 }
 
+// SubmitPackage is unimplemented for the neutrino backend, mirroring
+// TestMempoolAccept: a light client has no mempool and cannot run the
+// submitpackage RPC, so it can neither validate nor atomically accept a
+// package, nor report whether one was accepted.
+//
+// A zero-fee parent paired with a fee-paying CPFP child can still reach the
+// network over neutrino by broadcasting each transaction individually (parents
+// first) via SendRawTransaction and relying on a peer's P2P 1p1c package
+// relay. That is a best-effort propagation step with no acceptance signal,
+// distinct from submitpackage's submit-for-acceptance contract, so it is left
+// to the caller rather than conflated with this method.
+//
+// NOTE: This is part of the chain.Interface interface.
+func (s *NeutrinoClient) SubmitPackage(_ []*wire.MsgTx,
+	_ *float64) (*btcjson.SubmitPackageResult, error) {
+
+	return nil, ErrUnimplemented
+}
+
 // FilterBlocks scans the blocks contained in the FilterBlocksRequest for any
 // addresses of interest. For each requested block, the corresponding compact
 // filter will first be checked for matches, skipping those that do not report
--- a/chain/interface.go
+++ b/chain/interface.go
@@ -54,6 +54,8 @@ type Interface interface {
 	Notifications() <-chan interface{}
 	BackEnd() string
 	TestMempoolAccept([]*wire.MsgTx, float64) ([]*btcjson.TestMempoolAcceptResult, error)
+	SubmitPackage(txns []*wire.MsgTx,
+		maxFeeRate *float64) (*btcjson.SubmitPackageResult, error)
 	MapRPCErr(err error) error
 }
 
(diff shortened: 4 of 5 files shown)


chain/btcd.go
```diff
@@ -203,6 +203,17 @@ func (c *RPCClient) BackEnd() string {
...
 
+// SubmitPackage submits a package of related transactions (topologically
+// sorted, parents first and child last) to the btcd node via the submitpackage
+// RPC, issued over RawRequest since btcd's rpcclient has no typed method.
+//
+// NOTE: This is part of the chain.Interface interface.
+func (c *RPCClient) SubmitPackage(txns []*wire.MsgTx,
+	maxFeeRate *float64) (*btcjson.SubmitPackageResult, error) {
+
+	return submitPackageRawRequest(c.Client, txns, maxFeeRate)
```
yyforyongyu: This implementation will always fail against a btcd backend because the bumped `btcd` module only registers the `btcjson` command type and has no server handler for `submitpackage`, so either return `ErrUnimplemented` here or add the btcd server RPC before advertising support. (`gpt-5.5`)

yyforyongyu: Returns the `submitPackageRawRequest` error unmapped, whereas `BitcoindClient.SubmitPackage` wraps it with `c.MapRPCErr` (bitcoind_client.go:279), so `errors.Is` against `chain` sentinels (e.g. `ErrTxAlreadyInMempool`) silently fails for the btcd backend — wrap the return in `c.MapRPCErr` for parity.

(`claude-opus-4-8[1m]`)

chain/neutrino.go
```diff
@@ -238,6 +238,47 @@ func (s *NeutrinoClient) TestMempoolAccept(txns []*wire.MsgTx,
...
+			return nil, err
+		}
+
+		results[tx.WitnessHash().String()] = btcjson.SubmitPackageTxResult{
+			TxID: tx.TxHash(),
+		}
+	}
+
+	return &btcjson.SubmitPackageResult{
+		PackageMsg: "success",
```
yyforyongyu: Fabricating `PackageMsg: "success"` locally with no node acceptance signal misrepresents the submit-for-acceptance contract (a caller cannot distinguish "broadcast" from "validated/accepted") and contradicts the PR description, which states neutrino returns `ErrUnimplemented` — either return `ErrUnimplemented` or a result that does not assert `success` for an unconfirmed best-effort broadcast.

(`claude-opus-4-8[1m]`)

yyforyongyu:
We will prioritize reviewing this PR cc @GustavoStingelin

GustavoStingelin (commented):
No additional comments.

chain/neutrino.go
ellemouton: Updated description to match 👍

re keeping the broadcast rather than `ErrUnimplemented`: it just mirrors `SendRawTransaction` on neutrino — no mempool there, so "success" already means "broadcast" not "accepted", and it only errors if a peer sends an explicit `reject`. `SubmitPackage` is the package-shaped version of the same thing, so callers don't have to special-case neutrino. that said, happy to drop the `PackageMsg: "success"` for an empty msg if you'd rather it not imply acceptance.

GustavoStingelin: If neutrino can’t know whether the package was actually accepted, I’d prefer treating the result as inconclusive rather than as a full success. That avoids mixing broadcast-only behavior with Bitcoin Core API success semantics. The client could then explicitly distinguish “success” from “inconclusive,” instead of assuming success when the final state is unknown.

Also, deriving that state from a string seems risky. I would expect the API to expose this in a more structured way, making it harder to break during future code migrations. A plain string comparison is easy to miss or accidentally change.

GustavoStingelin (commented):
I don't know the urgency of this PR, but I've left my thoughts.

chain/bitcoind_client.go
```diff
@@ -256,6 +258,86 @@ func (c *BitcoindClient) TestMempoolAccept(txns []*wire.MsgTx,
 	return c.chainConn.client.TestMempoolAccept(txns, maxFeeRate)
 }
 
+// SubmitPackage submits a package of related transactions (topologically
+// sorted, parents first and child last) to bitcoind's mempool for atomic
+// validation and acceptance via the submitpackage RPC. This is what allows a
+// zero-fee v3/TRUC parent to be accepted when paired with a fee-paying CPFP
+// child, which sendrawtransaction (single-tx) rejects.
+//
+// btcd's rpcclient has no typed submitpackage method, so we issue it via
```
yyforyongyu: why don't we implement it in the btcd's `rpclient` first? Weird that we are making this exception. Atm we have this clean layering that the `rpcclient` manages the multi-backend RPCs and `chain` just calls that.

chain/neutrino.go
```diff
@@ -238,6 +238,47 @@ func (s *NeutrinoClient) TestMempoolAccept(txns []*wire.MsgTx,
...
+// SubmitPackageResult is SYNTHETIC: it is not parsed from any node response.
+// It is fabricated locally to report "success" once every transaction has been
+// handed to the P2P broadcaster, mirroring the best-effort semantics of
+// SendRawTransaction on this backend (which also reports success unless a peer
+// sends an explicit reject). A "success" result therefore means only that the
+// transactions were broadcast, not that the package was validated or accepted
+// anywhere. maxFeeRate is ignored, as there is no local mempool to apply it to.
+//
+// NOTE: This is part of the chain.Interface interface.
+func (s *NeutrinoClient) SubmitPackage(txns []*wire.MsgTx,
```
yyforyongyu: i think the interface method on neutrino is flawed - it now forces the caller to remember which backend it is using, and when it's neutrino, the interface method doesn't validate. We might as well provide a new interface method (e.g., `RelayPackage` or `BroadcastPackage`) that only submits but no validation, that way the caller can at least be aware that the job is not finished - we need another step to validate the submission.

And here i think we should return `ErrUnimplemented` instead, similar to how `TestMempoolAccept` was implemented.

yyforyongyu (changes requested):
I think we should add a typed wrapper for this on the `btcd/rpcclient` side first, similar to `TestMempoolAccept`, and fix the `neutrino` implementation so `SubmitPackage` has the same contract across backends.

ellemouton:
Thanks all 🙏 addressed the comments in the latest push:

- added the typed `SubmitPackage` to `rpcclient` (btcsuite/btcd#2551) — `BitcoindClient` now forwards to that instead of issuing its own `RawRequest`, so the `chain` → `rpcclient` layering stays clean.
- btcd `RPCClient` + `NeutrinoClient` now return `ErrUnimplemented` (no fabricated "success"), so the contract is uniform across backends — same as how `TestMempoolAccept` handles the unsupported case on neutrino.
- bitcoind path maps errors through `MapRPCErr` for parity with `SendRawTransaction`.

`go.mod` temporarily `replace`s btcd with the fork carrying #2551 (with a `TODO` to drop it once that's tagged). PTAL 🙏

GustavoStingelin (approved):
LGTM

yyforyongyu (commented):
Pending the btcd PR, otherwise LGTM🚢

go.mod
```diff
@@ -58,3 +59,9 @@ require (
 // If you change this please run `make lint` to see where else it needs to be
 // updated as well.
 go 1.24.6
+
+// TODO(elle): drop this replace and bump btcd to the tagged release once it
+// is cut. btcsuite/btcd#2551 (the typed rpcclient.SubmitPackage that
+// BitcoindClient.SubmitPackage depends on) is merged to master but not yet
+// tagged, so we pin to the merge commit in the meantime.
+replace github.com/btcsuite/btcd => github.com/btcsuite/btcd v0.25.1-0.20260617155916-de3d460e7626
```
Roasbeef: Can drop this now.

Roasbeef (approved):
LGTM 🙈

PR #1265, opened 2026-06-11
multi: add keyvault lifecycle boundary

GustavoStingelin:
This prepares the wallet/key encryption migration by making the keyvault the
future owner of wallet lock state, unlock timing, and passphrase refresh logic.

Today, legacy `waddrmgr` still owns the working lock/unlock/passphrase behavior.
This PR keeps those paths intact by moving them behind explicitly deprecated
methods, while adding the lifecycle shape that later PRs will implement in the
vault.

- add lock/unlock/passphrase-refresh methods to `keyvault.Vault`
- rename legacy `waddrmgr` lifecycle methods to deprecated entrypoints
- update existing wallet callers, mocks, and tests to keep using the legacy path
- document that keyvault is wallet-scoped in ADR 0010

Follow-up PRs will continue the vault implementation and gradually move callers
off the deprecated `waddrmgr` lifecycle.

### DIFF
--- a/docs/developer/adr/0010-keyvault-encryption-layer.md
+++ b/docs/developer/adr/0010-keyvault-encryption-layer.md
@@ -2,106 +2,185 @@
 
 ## 1. Context
 
-The updated encryption model defined in ADR 0009 and the planned
-cryptographic primitive migration proposed in ADR 0007 require a clear
-boundary between domain logic and the SQL database layer. The
-legacy `waddrmgr` design tightly couples storage, locking, and key derivation,
-which complicates the SQL migration and makes encryption behavior hard to test
-in isolation.
+The encryption model defined in ADR 0009 and the cryptographic primitive
+migration proposed in ADR 0007 require a clear boundary between wallet domain
+logic and database persistence.
 
-To address this, we need a dedicated component that owns lock state, key
-derivation, and encryption, while keeping the database layer strictly
-encryption agnostic.
+The legacy `waddrmgr` design couples storage, locking, key derivation, and
+encryption. This makes the SQL migration harder, spreads encryption behavior
+across persistence code, and makes lock state difficult to test in isolation.
 
-The `db.Store` remains available to other callers for non-cryptographic
+We need a dedicated component that owns lock state, key derivation, secret
+material lifetime, and encryption. The database layer should remain encryption
+agnostic and store encrypted key material as opaque bytes.
+
+The `db.Store` remains available to other wallet code for non-cryptographic
 queries and updates.
 
 ## 2. Decision
 
-We will introduce a dedicated **`wallet/internal/keyvault`** package that
-defines the encryption boundary between domain code and the SQL store layer.
-Keyvault accesses the database through `db.Store`, not by talking directly
-to the SQL backend.
+We will introduce a dedicated `wallet/internal/keyvault` package. This package
+defines the encryption boundary between wallet domain code and the store layer.
+
+A wallet facing `keyvault.Vault` is scoped to exactly one wallet at construction
+time. It receives a `db.Store` and a wallet identifier during wiring, then keeps
+persistence routing internal. Callers that hold a vault do not pass a wallet ID
+to each lock, unlock, encrypt, decrypt, or derivation operation.
+
+Keyvault accesses persistence through `db.Store`. It does not talk directly to
+SQL backends.
 
 ```mermaid
 flowchart TD
-    A[Domain Code] -->|crypto ops| B[keyvault]
-    A -->|non-crypto ops| E[db.Store]
-    B -->|derive| C[btcutil/hdkeychain]
-    B -->|decrypt/encrypt| F[XChaCha20Poly1305</br>XSalsa20Poly1305]
-    B -->|read/write| E
-    B -->|read/write| H[(memory cache)]
-    E --> D[(SQL DB)]
-    E --> G[(kvdb)]
-    classDef store fill:#f8f9fb,stroke:#9aa0a6,stroke-width:1px
-    class G,D store
+    wallet[Wallet]
+
+    subgraph support[Supporting libraries]
+        hd[btcutil/hdkeychain<br/>BIP32 and BIP44 derivation]
+        crypto[AEAD crypto<br/>current and migrated primitives]
+    end
+
+    subgraph vault_boundary[keyvault boundary]
+        vault[keyvault.Vault<br/>wallet scoped lock state<br/>key lifecycle<br/>secret material API]
+        cache[(In memory secret cache<br/>unlocked keys<br/>derived keys)]
+    end
+
+    subgraph store_boundary[db.Store boundary]
+        store[db.Store<br/>multi wallet persistence API]
+
+        subgraph backends[Persistence backends]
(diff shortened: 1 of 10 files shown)


waddrmgr/manager.go
```diff
@@ -397,6 +403,24 @@ type Manager struct {
 	hashedPrivPassphrase [sha512.Size]byte
 }
 
+// Unlock unlocks the manager through the key vault boundary.
+func (m *Manager) Unlock(_ context.Context, _ []byte, _ time.Duration) error {
```
yyforyongyu: Changing this exported lifecycle method to the keyvault signature is an API break rather than a deprecation, because existing callers of `Manager.Unlock(ns, passphrase)` no longer compile; keep the legacy API under its original name and satisfy the vault lifecycle through an adapter or a differently named method. (`gpt-5.5`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔐

docs/developer/adr/0010-keyvault-encryption-layer.md
```diff
@@ -2,106 +2,188 @@
...
 ## 1. Context
 
-The updated encryption model defined in ADR 0009 and the planned
-cryptographic primitive migration proposed in ADR 0007 require a clear
-boundary between domain logic and the SQL database layer. The
-legacy `waddrmgr` design tightly couples storage, locking, and key derivation,
-which complicates the SQL migration and makes encryption behavior hard to test
-in isolation.
+The encryption model defined in ADR 0009 and the cryptographic primitive
+  migration proposed in ADR 0007 require a clear boundary between wallet domain
```
yyforyongyu: some weird spaces here

docs/developer/adr/0010-keyvault-encryption-layer.md
```diff
@@ -2,106 +2,188 @@
...
+
+    store -->|routes by wallet_id| sql
+    store -->|adapts legacy layout| kvdb
 ```
 
+This is a structural boundary diagram, not a runtime call sequence.
+
+The `Wallet` struct holds two related dependencies:
+
+1. A wallet scoped `keyvault.Vault`
```
yyforyongyu, line 71: yeah this is a decision point, whether this should be single-wallet focused or being multi-wallets like db.Store

docs/developer/adr/0010-keyvault-encryption-layer.md
```diff
@@ -2,106 +2,188 @@
...
-   Keyvault can be tested against mock databases, and domain logic can be
-   tested using mock keyvault implementations.
+6. **Better testability**
+   Keyvault can be tested against mock stores, and wallet domain code can be
+     tested against mock vault implementations.
 
-* **Memory safety**
-  Secure memory handling and zeroing are centralized in one place.
+7. **Per wallet isolation**
+   Each wallet has its own vault instance, lock state, cache, timers, and secret
```
yyforyongyu, line 156: yeah i guess this is the strongest pro - one wallet cannot accidentally touch another wallet's key vault.

waddrmgr/interface.go
```diff
@@ -130,10 +130,10 @@ type AddrStore interface {
 	// IsLocked returns whether or not the address managed is locked.
 	IsLocked() bool
 
-	// ChangePassphrase changes either the public or private passphrase to
-	// the provided value depending on the private flag.
-	ChangePassphrase(ns walletdb.ReadWriteBucket, oldPass, newPass []byte,
-		private bool, scryptOptions *ScryptOptions) error
+	// ChangePassphraseDeprecated changes the public/private passphrase to the
+	// provided value depending on the private flag.
+	ChangePassphraseDeprecated(ns walletdb.ReadWriteBucket, oldPass,
```
yyforyongyu: Cool - tho IIRC, this `ChangePassphrase` is actually introduced in this feature branch, so i guess there is no need to deprecate it first?

waddrmgr/interface.go
```diff
@@ -120,8 +120,9 @@ type AddrStore interface {
 	ForEachRelevantActiveAddress(ns walletdb.ReadBucket,
 		fn func(addr btcutil.Address) error) error
 
-	// Unlock derives the master private key from the specified passphrase.
-	Unlock(ns walletdb.ReadBucket, passphrase []byte) error
+	// UnlockDeprecated derives the master private key from the specified
+	// passphrase.
+	UnlockDeprecated(ns walletdb.ReadBucket, passphrase []byte) error
```
yyforyongyu: similar to others, i think we can keep them? My understanding is that, the old wallet will use the `w.addrStore`, and this whole `AddrStore` is the deprecated path, so no need to take the namespaces. The new wallet will access the keyvaule via `w.vault`?

wallet/internal/keyvault/keyvault.go
```diff
@@ -1,17 +1,49 @@
...
+	// Decrypt decrypts ciphertext key material with the selected crypto key
+	// type.
 	Decrypt(keyType waddrmgr.CryptoKeyType, ciphertext []byte) ([]byte, error)
+
+	// RefreshPrivatePassphrase refreshes vault owned runtime passphrase and
+	// crypto state after a successful private passphrase rotation.
+	//
+	// The vault must still be unlocked with the new passphrase when this method
+	// is called.
+	RefreshPrivatePassphrase(passphrase []byte) error
```
yyforyongyu, line 45: hmm when will this method be used?

yyforyongyu (changes requested):

waddrmgr/interface.go
GustavoStingelin: Yes, TBH, I would prefer not to deprecate that too. But to do that, we would need to remove the `var _ Vault = (*waddrmgr.Manager)(nil)` compile check, which I think is more aligned anyway.

waddrmgr/interface.go
GustavoStingelin: same
>Yes, TBH, I would prefer not to deprecate that too. But to do that, we would need to remove the `var _ Vault = (*waddrmgr.Manager)(nil)` compile check, which I think is more aligned anyway.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: Literally, to change the wallet passphrase, we need to handle the internal private material inside the vault. That means getting the crypto keys from the internal memory vault, re-encrypting them with the new passphrase, and then persisting them to the underlying database.

docs/developer/adr/0010-keyvault-encryption-layer.md
GustavoStingelin: removed

waddrmgr/interface.go
GustavoStingelin: Changed the approach; take a look

waddrmgr/interface.go
GustavoStingelin: Changed the approach; take a look

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🧰

yyforyongyu:
No issues found by gpt-5.5 🧭

yyforyongyu (approved):

PR #1266, opened 2026-06-12
Update to btcd v2 modules

guggero:
After merging https://github.com/btcsuite/btcd/pull/1825, we'll need to update all `btcd` dependencies to the new `v2` modules.
We need to start at the edges of the dependency graph, which in this case is `neutrino` and `btcwallet`.

But the graph is very circular, so this depends on:
 - https://github.com/lightningnetwork/lnd/pull/10902
 - https://github.com/lightninglabs/neutrino/pull/364 (circular)


go.sum
```diff
@@ -1,28 +1,13 @@
...
+github.com/btcsuite/btcd v0.26.0-beta.rc1.0.20260604071853-1966c384533d/go.mod h1:7ft7+a/MoJHFouFopCb1zyiR9IWPlrcPVn6K/lJ1dcA=
 github.com/btcsuite/btcd/address/v2 v2.0.0 h1:UVu8Hal6Siu4XastFe+JX5JkeBYONbDUIY5E+SVTs6I=
 github.com/btcsuite/btcd/address/v2 v2.0.0/go.mod h1:htJK1AtaeK3bKNfZY63ep2oN8LbrI6qvmPGe1vekb3I=
-github.com/btcsuite/btcd/btcec/v2 v2.1.0/go.mod h1:2VzYrv4Gm4apmbVVsSq5bqf1Ec8v56E48Vt0Y/umPgA=
-github.com/btcsuite/btcd/btcec/v2 v2.1.3/go.mod h1:ctjw4H1kknNJmRN4iP1R7bTQ+v3GJkZBd6mui8ZsAZE=
 github.com/btcsuite/btcd/btcec/v2 v2.5.0 h1:KioMXOWa76b86sTZZOmbzv/ldaQCmB8KFAyn5PbB8E8=
 github.com/btcsuite/btcd/btcec/v2 v2.5.0/go.mod h1:+K/MYXcLBtHEQjRbjHuJChuybk4LCgjdjgRwil+e+Kk=
-github.com/btcsuite/btcd/btcutil v1.0.0/go.mod h1:Uoxwv0pqYWhD//tfTiipkxNfdhG9UrLwaeswfjfdF0A=
-github.com/btcsuite/btcd/btcutil v1.1.0/go.mod h1:5OapHB7A2hBBWLm48mmw4MOHNJCcUBTwmWH/0Jn8VHE=
-github.com/btcsuite/btcd/btcutil v1.1.5/go.mod h1:PSZZ4UitpLBWzxGd5VGOrLnmOjtPP/a6HaFo12zMs00=
```
Roasbeef, line 15: 🔥

go.mod
```diff
@@ -1,7 +1,7 @@
 module github.com/btcsuite/btcwallet
 
 require (
-	github.com/btcsuite/btcd v0.26.0-beta.rc1
+	github.com/btcsuite/btcd v0.26.0-beta.rc1.0.20260604071853-1966c384533d
```
Roasbeef: Can tag it as v0.26.0 now.

go.mod
```diff
@@ -65,7 +63,9 @@ require (
 
 // If you change this please run `make lint` to see where else it needs to be
 // updated as well.
-go 1.25
+go 1.25.11
+
+replace github.com/lightningnetwork/lnd/tlv => github.com/guggero/lnd/tlv v0.0.0-20260612140810-bdba8e0ddcb2
```
Roasbeef: Is there a corresponding lnd PR here?

wtxmgr/go.mod
```diff
@@ -28,7 +28,7 @@ require (
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
 
-go 1.25
```
Roasbeef: We should make this uniform across the set of packages (some other pkgs require go 1.25 as an example).

go.mod
guggero: Nice, done!

go.mod
guggero: Yes, linked above: https://github.com/lightningnetwork/lnd/pull/10902. Though I'm not sure if we need to turn it into a `v2` module too, to clearly signal that updating has major consequences for a consuming project (as it might require you to update all other btcd modules too).

wtxmgr/go.mod
guggero: Sounds good, done.

guggero:
I addressed your feedback, @Roasbeef, thanks for the review! Not sure how we want to proceed with this one. The "Check commits" fails as indicated in the first commit (since it's scripted and not 100% compilable). Should we ignore that and do "squash and merge" just for this PR? Or should I combine the two commits to make it pass?

Also, I propose merging this one first, then tagging all submodules (probably needs to be in multiple steps since there are multiple levels of dependencies) and getting rid of the replace rules in `go.mod` and then finally merging [the neutrino one](https://github.com/lightninglabs/neutrino/pull/364), pushing a tag there then pinning it here.

Roasbeef (approved):
LGTM 🥙

PR #1267, opened 2026-06-15
wallet+db: fix SQL-wallet review issues

yyforyongyu:
Addresses review findings and regressions surfaced while preparing the
SQL-wallet/runtime-store stack. This is the base branch for #1268.

Highlights:
- Fix taproot PSBT decoration so wallet-decorated taproot inputs do not carry
  both legacy and taproot derivations.
- Keep missing kvdb scopes consistent with `Store.ListAccounts` empty-list
  semantics.
- Snapshot recovery outpoints before same-block lookahead expansion.
- Harden SQL block consistency checks and sparse rollback sync-height rewinds.
- Surface kvdb watch-only wallet state and guard wallet account validation
  against nil extended pubkeys.
- Accept value-typed scoped account and coin-source inputs.
- Update the rollback corruption assertion to check backend-independent error
  context.

Validation:
- `make fmt`
- `make lint`
- `scripts/check-each-commit.sh sql-wallet`
- Focused wallet, kvdb, and DB integration regression tests

PR #1268, opened 2026-06-15
wallet+db: normalize account address identity

yyforyongyu:
Stacked on #1267.

Introduces normalized account and address identity as the base layer for the
runtime-store stack, with ADR coverage and separate SQL, kvdb, wallet, and
integration-test commits split for reviewability.

Highlights:
- Add ADR 0013 for normalized account/address identity.
- Normalize account, address, and UTXO identity so imported accounts and
  imported-xpub children do not fabricate wallet-derived account numbers.
- Update SQL migrations, queries, sqlc output, and SQL adapters for the
  normalized schema.
- Update kvdb adapters to expose the same identity semantics while preserving
  legacy waddrmgr behavior.
- Update wallet account, address, PSBT, signer, tx-creation, and UTXO
  consumers.
- Add regression coverage for account, address, and UTXO identity, including
  legacy kvdb mixed watch-only spendability.

Validation:
- `make fmt`
- `make lint`
- `go test -run '^$' ./wallet/internal/db/... ./wallet`
- `go test ./wallet/internal/db/kvdb`
- Focused wallet and DB integration regression tests

Note: this PR follows the Task 291 review split, so branch-tip validation is
used instead of requiring every individual review commit to be independently
buildable.


wallet/internal/sql/pg/queries/accounts.sql
```diff
@@ -318,7 +318,7 @@ WHERE
             AND key_scopes.purpose = sqlc.arg('purpose')
             AND key_scopes.coin_type = sqlc.arg('coin_type')
     )
-    AND account_number = sqlc.arg(account_number);
+    AND account_number = sqlc.arg('account_number');
```
yyforyongyu: this change should not be in the commit where we introduced the ADR?

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -1,39 +1,22 @@
 -- Migration note: Intentionally NOT idempotent (no "IF NOT EXISTS").
 -- This ensures migration tracking stays accurate and fails loudly if run twice.
 
--- Account origin lookup table - provides standardized descriptions for wallet
--- account origins. This is a reference table that maps the AccountOrigin enum
--- values used in Go code to their human-readable names.
+-- Accounts table stores wallet-level HD account identity under each key scope.
+-- Wallet-derived BIP44 account numbers live in derived_accounts; imported xpub
+-- accounts have an accounts row without a derived_accounts child.
+-- (legacy; dropped in cleanup) account origin lookup table.
```
yyforyongyu: there is no need to care about legacy - we are in a side branch so we can just edit the migrations and queries in place

wallet/internal/sql/pg/queries/accounts.sql
yyforyongyu: The SQL and generated code now live in their own dedicated commits (separate from the ADR-introducing commit) per our commit conventions, after the re-cascade. Resolving.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: Agreed — migrations and queries are edited in place on this feature branch with no legacy-compatibility shims, per this note. Resolving.

wallet/utxo_manager.go
```diff
@@ -231,8 +227,7 @@ func (w *Wallet) buildWalletUtxoFromStore(info *db.UtxoInfo,
 		return nil, err
 	}
 
-	spendable := !w.IsWatchOnly() &&
-		info.Origin != db.ImportedAccount
+	spendable := !w.IsWatchOnly()
```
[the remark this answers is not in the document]
yyforyongyu: Per ADR 0012 (wallet-level watch-only uniformity), spendability is a uniform wallet-level property: `db.UtxoInfo` deliberately carries no per-UTXO/per-address spendable flag and read paths do not consult `account_secrets`. A spendable SQL wallet rejects mixed-mode (watch-only) imports at create, so `spendable := !w.IsWatchOnly()` is correct there; kvdb mixed-mode is grandfathered at the legacy manager layer rather than reclassified per-UTXO. Option (a) (a per-UTXO flag) is specifically disallowed by the ADR, so resolving as wontfix.

wallet/internal/sql/sqlite/queries/addresses.sql
```diff
@@ -95,54 +103,91 @@ LEFT JOIN address_secrets AS s ON a.id = s.address_id
...
+)
+SELECT
+    a.id AS address_id,
+    a.wallet_id,
+    a.scope_id,
+    sqlc.arg('account_id') AS account_id,
+    sqlc.arg('address_branch') AS address_branch,
+    sqlc.arg('address_index') AS address_index
+FROM addresses AS a
+WHERE a.id = sqlc.arg('address_id');
```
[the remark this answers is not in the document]
yyforyongyu: The sole caller (`derivedAddressCreateAddr`) inserts the parent row via `CreateDerivedAddress` and passes the returned `row.ID` straight into `CreateDerivedAddressPath` within the same transaction, so the `INSERT ... SELECT ... WHERE a.id = address_id` always matches the just-created row — a silent 0-row no-op cannot occur by construction. Keeping `:exec`; resolving.

wallet/internal/sql/pg/queries/addresses.sql
```diff
@@ -91,54 +99,91 @@ LEFT JOIN address_secrets AS s ON a.id = s.address_id
...
+)
+SELECT
+    a.id AS address_id,
+    a.wallet_id,
+    a.scope_id,
+    sqlc.arg('account_id') AS account_id,
+    sqlc.arg('address_branch') AS address_branch,
+    sqlc.arg('address_index') AS address_index
+FROM addresses AS a
+WHERE a.id = sqlc.arg('address_id');
```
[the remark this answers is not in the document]
yyforyongyu: The sole caller (`derivedAddressCreateAddr`) inserts the parent row via `CreateDerivedAddress` and passes the returned `row.ID` straight into `CreateDerivedAddressPath` within the same transaction, so the `INSERT ... SELECT ... WHERE a.id = address_id` always matches the just-created row — a silent 0-row no-op cannot occur by construction. Keeping `:exec`; resolving.

wallet/internal/db/pg/block.go
```diff
@@ -44,6 +44,11 @@ func ensureBlockExists(ctx context.Context, qtx *sqlc.Queries,
 		return fmt.Errorf("insert block: %w", err)
 	}
 
+	_, err = requireBlockMatches(ctx, qtx, block)
```
yyforyongyu, line 47: this commit belongs to `review-base-fixes`?

wallet/internal/db/kvdb/walletstore.go
```diff
@@ -83,6 +83,7 @@ func (s *Store) GetWallet(_ context.Context,
 		Birthday:      addrStore.Birthday().UTC(),
 		BirthdayBlock: birthdayBlock,
 		SyncedTo:      syncedTo,
+		IsWatchOnly:   addrStore.WatchOnly(),
```
yyforyongyu, line 86: this commit also belongs to `review-base-fixes`?

wallet/internal/db/pg/txstore_rollback.go
```diff
@@ -66,19 +66,8 @@ func (o rollbackToBlockOps) RewindWalletSyncStateHeights(
...
-			return fmt.Errorf("convert new height: %w", err)
-		}
-	}
-
 	_, err = o.qtx.RewindWalletSyncStateHeightsForRollback(
-		ctx, sqlc.RewindWalletSyncStateHeightsForRollbackParams{
-			RollbackHeight: rollbackHeight,
-			NewHeight:      newHeight,
-		},
+		ctx, rollbackHeight,
```
yyforyongyu, line 70: i am not sure about `db/sql: fix rollback to sparse block height` - how is it possible that we have sparse blocks? i think the wallet will sync every block to be sure it's fully sycned?

wallet/account_manager.go
```diff
@@ -507,6 +507,13 @@ func dbScopeAddrSchema(
 func validateExtendedPubKey(pubKey *hdkeychain.ExtendedKey,
 	isAccountKey bool, chainParams *chaincfg.Params) error {
 
+	// A nil key cannot be validated and would otherwise panic on the
+	// IsPrivate call below.
+	if pubKey == nil {
```
yyforyongyu, line 504: this commit also belongs to `review-base-fixes`

wallet/utxo_manager.go
```diff
@@ -147,7 +147,7 @@ func (w *Wallet) ListUnspent(ctx context.Context,
 
 	log.Debugf("ListUnspent using query: %v", query)
 
-	currentHeight := w.addrStore.SyncedTo().Height
+	currentHeight := w.SyncedTo().Height
```
yyforyongyu: this commit also belongs to `review-base-fixes`

wallet/tx_creator.go
```diff
@@ -309,13 +309,21 @@ func (i *InputsPolicy) validate() error {
 	switch source := i.Source.(type) {
 	// If the source is a scoped account, it must have a non-empty account
 	// name.
+	case ScopedAccount:
```
yyforyongyu, line 312: this commit also belongs to `review-base-fixes`

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -25,6 +25,15 @@ CREATE TABLE addresses (
     -- Reference to the account this address belongs to.
     account_id BIGINT NOT NULL,
 
+    -- Transitional normalized-identity columns (nullable while legacy columns
+    -- remain). scope_id mirrors the owning key scope, is_derived is the shape
+    -- marker, and script_type_id is the renamed type reference.
+    scope_id BIGINT,
+
+    is_derived BOOLEAN,
```
yyforyongyu: missing inline docs

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -25,6 +25,15 @@ CREATE TABLE addresses (
...
+
+    script_type_id SMALLINT,
```
yyforyongyu: missing inline docs

wallet/internal/sql/sqlite/migrations/000006_addresses.up.sql
```diff
@@ -25,6 +25,15 @@ CREATE TABLE addresses (
     -- Reference to the account this address belongs to.
     account_id INTEGER NOT NULL,
 
+    -- Transitional normalized-identity columns (nullable while legacy columns
+    -- remain). scope_id mirrors the owning key scope, is_derived is the shape
+    -- marker, and script_type_id is the renamed type reference.
+    scope_id INTEGER,
+
+    is_derived BOOLEAN,
```
yyforyongyu: missing inline docs

wallet/internal/sql/sqlite/migrations/000006_addresses.up.sql
```diff
@@ -25,6 +25,15 @@ CREATE TABLE addresses (
...
+
+    script_type_id INTEGER,
```
yyforyongyu: missing inline docs

wallet/internal/db/kvdb/accountstore_createimported.go
```diff
@@ -209,14 +213,25 @@ func dryRunImportedAccount(ns walletdb.ReadWriteBucket,
 // waddrmgrScopeAddrSchema converts the account-store schema type into the
 // waddrmgr per-account address schema type.
 func waddrmgrScopeAddrSchema(
-	schema *db.ScopeAddrSchema) *waddrmgr.ScopeAddrSchema {
+	schema *db.ScopeAddrSchema) (*waddrmgr.ScopeAddrSchema, error) {
 
-	if schema == nil {
-		return nil
-	}
+	var result *waddrmgr.ScopeAddrSchema
```
yyforyongyu: we should keep the `if schema == nil` fast exit!

yyforyongyu:
No issues found by gpt-5.5 🧭

wallet/utxo_manager.go
```diff
@@ -231,8 +230,7 @@ func (w *Wallet) buildWalletUtxoFromStore(info *db.UtxoInfo,
 		return nil, err
 	}
 
-	spendable := !w.IsWatchOnly() &&
-		info.Origin != db.ImportedAccount
+	spendable := !w.IsWatchOnly()
```
yyforyongyu, line 244: For kvdb wallets this treats outputs from legacy imported or account-level watch-only rows as spendable because `UtxoInfo` no longer carries the per-address/account spendability that `AddressDetails` enforced, so preserve a kvdb per-row spendable flag or keep those rows non-spendable before setting this value. (`gpt-5.5 xhigh`)

wallet/internal/db/pg/block.go
yyforyongyu: Addressed by moving this block-consistency fix out of PR 1268 and into `review-base-fixes` (#1267).

wallet/internal/db/kvdb/walletstore.go
yyforyongyu: Addressed by moving the kvdb watch-only `GetWallet` fix out of PR 1268 and into `review-base-fixes` (#1267).

wallet/internal/db/pg/txstore_rollback.go
yyforyongyu: Addressed in #1267. The rollback fix is now in `review-base-fixes` with regression coverage. Normal sync should store every block, but `RollbackToBlock` should not require the exact target height to be present when the SQL block table is sparse due to partial/corrupt state or test setup; it rewinds wallet sync state to the highest stored block not above the target.

wallet/account_manager.go
yyforyongyu: Addressed by moving the nil extended pubkey guard into `review-base-fixes` (#1267), with regression coverage.

wallet/utxo_manager.go
yyforyongyu: Addressed by moving the prep fix out of PR 1268 and into `review-base-fixes` (#1267); the later no-op `Wallet.SyncedTo()` accessor-only change was dropped.

wallet/tx_creator.go
yyforyongyu: Addressed by moving value-typed coin-source handling into `review-base-fixes` (#1267), with regression coverage.

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
yyforyongyu: Addressed by adding inline docs in the PostgreSQL address migration for the nullable account identity fields.

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
yyforyongyu: Addressed by adding inline docs in the PostgreSQL address migration for the derived-address path metadata.

wallet/internal/sql/sqlite/migrations/000006_addresses.up.sql
yyforyongyu: Addressed by adding inline docs in the SQLite address migration for the nullable account identity fields.

wallet/internal/sql/sqlite/migrations/000006_addresses.up.sql
yyforyongyu: Addressed by adding inline docs in the SQLite address migration for the derived-address path metadata.

wallet/internal/db/kvdb/accountstore_createimported.go
yyforyongyu: Addressed: kept the explicit `schema == nil` fast exit in `waddrmgrScopeAddrSchema`, with `//nolint:nilnil` documenting the intentional `(nil, nil)` result.

wallet/utxo_manager.go
yyforyongyu: Addressed by adding optional `db.UtxoInfo.Spendable`. kvdb sets it from legacy account/address spendability for mixed-mode imported rows, and wallet UTXO mapping uses it before falling back to wallet-level watch-only. Added `TestGetUtxoUsesStoreSpendableOverride` plus kvdb UTXO spendability coverage.

wallet/internal/db/pg/block.go
yyforyongyu, line 47: Addressed in `review-base-fixes` with `TestUpdateTxRejectsMismatchedConfirmedBlock`, covering the conflicting height/hash path and asserting `ErrBlockMismatch`.

wallet/utxo_manager.go
yyforyongyu: Addressed by dropping the no-op `Wallet.SyncedTo()` accessor-only commit; this stack no longer includes that change.

wallet/account_manager.go
```diff
@@ -507,6 +498,13 @@ func dbScopeAddrSchema(
...
+		return fmt.Errorf("%w: account key cannot be nil",
+			ErrInvalidAccountKey)
```
yyforyongyu, line 506: Addressed by moving these prep fixes into #1267 and rewriting PR 1268; the final PR 1268 stack now passes `scripts/check-each-commit.sh review-base-fixes`. Regression coverage for the nil xpub guard, value-form account input, and sparse rollback lives in `review-base-fixes`.

wallet/internal/sql/pg/queries/accounts.sql
yyforyongyu: Addressed by dropping the no-op quoted `sqlc.arg` commit; the rewritten PR 1268 stack no longer includes it.

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🆔

wallet/internal/db/kvdb/accountstore_createimported.go
```diff
@@ -216,6 +216,8 @@ func waddrmgrScopeAddrSchema(
 	schema *db.ScopeAddrSchema) (*waddrmgr.ScopeAddrSchema, error) {
 
 	if schema == nil {
+		// A nil schema intentionally maps to no per-account override.
```
yyforyongyu: this should be fold into where it's introduced?

wallet/internal/sql/pg/migrations/000005_accounts.down.sql
```diff
@@ -2,9 +2,13 @@
 -- Must succeed even if tables are already dropped or database is in unexpected state.
 DROP TRIGGER IF EXISTS trg_assert_watch_only_account_secrets_insert ON account_secrets;
 DROP TRIGGER IF EXISTS trg_assert_watch_only_account_secrets_update ON account_secrets;
-DROP TRIGGER IF EXISTS trg_assert_account_wallet_id_immutable ON accounts;
+DROP TRIGGER IF EXISTS trg_assert_derived_account_parent_insert ON derived_accounts;
```
yyforyongyu: split this commit into two - one dealing with the `accounts` table, the other with `addresses` table

yyforyongyu:
No issues found by `gpt-5.5 xhigh` 🐙

wallet/internal/db/accountstore_accountinfo.go
```diff
@@ -150,7 +196,7 @@ func AccountPropsRowToInfo[AddrTypeId ~int16 | ~int64, AccOriginId any](
 	}
 
 	externalKeyCount, internalKeyCount, importedKeyCount, err := getKeyCounts(
-		row.ExternalKeyCount, row.InternalKeyCount, row.ImportedKeyCount,
+		row.ExternalKeyCount, row.InternalKeyCount, 0,
```
yyforyongyu, line 228: why 0? we need to add inline comments here

wallet/internal/db/accountstore_accountinfo.go
```diff
@@ -338,7 +375,7 @@ func AccountRowToInfo[AccOriginId ~int16 | ~int64](
...
```
yyforyongyu, line 440: same here, document it

wallet/internal/db/kvdb/accountstore_createimported.go
yyforyongyu: Addressed. The nil-schema fast-exit comment is now folded into `db/kvdb: align account identity`, the commit that changes `waddrmgrScopeAddrSchema`; there is no standalone follow-up commit.

wallet/internal/sql/pg/migrations/000005_accounts.down.sql
yyforyongyu: Addressed. PostgreSQL SQL normalization is now split by domain: `db/sql/pg: normalize account schema`, `db/sql/pg: normalize address schema`, `db/sql/pg: normalize UTXO queries`, plus a generated shared sqlc metadata commit for cross-query wiring. The same account/address/UTXO split was mirrored for SQLite.

wallet/internal/db/accountstore_accountinfo.go
yyforyongyu: Addressed with an inline comment above this conversion. Normalized SQL account rows track only HD external/internal branch counters; individually imported address counts are a legacy kvdb/waddrmgr account property, so SQL account info reports zero for `ImportedKeyCount` here.

wallet/internal/db/accountstore_accountinfo.go
yyforyongyu: Addressed with the same inline comment at this conversion site. The zero documents that normalized SQL account rows do not carry legacy imported-address counts; those remain a kvdb/waddrmgr account property.

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🗂️

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -0,0 +1,113 @@
...
+  The kvdb backend leaves `AccountID` nil.
+- `db.AddressInfo` makes `AccountID` and `AccountNumber` optional. Raw imports
+  have neither; imported-xpub child addresses have an account ID but no BIP44
+  account number.
+- `AddressDerivationParams` carries an optional derived account number. Imported
+  xpub child addresses must not synthesize `0` and accidentally derive wallet
+  seed keys.
+- Low-cardinality account lists may start from `accounts` and batch-hydrate
+  subtype data. Address, UTXO, and balance reads must reject or exclude malformed
+  parent/child shapes instead of silently reclassifying corrupt rows.
```
GustavoStingelin: Why are we specifying low-cardinality here? What about high-cardinality?

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -0,0 +1,113 @@
...
+  or raw imported addresses can appear.
+- Every account/address read path must use the same corruption checks so a
+  parent/child shape mismatch fails loudly instead of leaking through one query.
+- Raw-import compatibility paths are explicit query variants, not ordinary
+  account queries.
+
+## 4. Implementation Notes
+
+- Modify the existing unmerged account and address migrations in place. Do not
+  add new migration numbers for this feature-branch schema rewrite.
```
GustavoStingelin, line 177: Is this only for agents?

I think the ADRs should compare the behavior on master with the target behavior on the sql wallet branch, instead of changing the mental model across commits.

At minimum, before the release, we should clean up the ADRs by merging the final decisions into a clear narrative, without backtracking through intermediate states. We can still keep small notes about alternative designs that were considered, but the main document should describe the final model directly.

wallet/internal/db/itest/accountstore_helpers_test.go
```diff
@@ -90,6 +95,15 @@ func getAccountQueryByNumber(walletID uint32, scope db.KeyScope,
 	}
 }
 
+// accountNumberValue requires a derived account number and returns its value.
+func accountNumberValue(t *testing.T, accountNumber *uint32) uint32 {
```
GustavoStingelin: IMO this way reads cleaner 
```suggestion
func accountNumberNotNil(t *testing.T, accountNumber *uint32) uint32 {
```

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
```diff
@@ -44,15 +14,13 @@ CREATE TABLE accounts (
...
-
     -- Human friendly name for the account.
     account_name TEXT NOT NULL,
 
-    -- Reference to the origin of the account.
-    origin_id SMALLINT NOT NULL,
+    -- Shape marker. TRUE means this account must have a derived_accounts child
+    -- row with a wallet-derived BIP44 account number. Imported xpub accounts
+    -- leave this FALSE.
+    is_derived BOOLEAN NOT NULL,
```
GustavoStingelin: I really like how this boolean field makes the derived meaning explicit, but I am not convinced that the extra table adds enough clarity.

Keeping a nullable `account_number` here seems much simpler. We can enforce the invariant directly with a check constraint: when `is_derived` is true, `account_number` must be non null. That avoids triggers and extra relational machinery.

In the application layer, we can also model `account_number` as a pointer if the `is_derived` boolean is not enough to represent the distinction clearly.

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -8,42 +8,33 @@
...
     -- DB ID of the address, primary key.
     id BIGSERIAL PRIMARY KEY,
 
     -- Reference to the wallet this address belongs to.
     wallet_id BIGINT NOT NULL,
 
-    -- Reference to the account this address belongs to.
-    account_id BIGINT NOT NULL,
+    -- Reference to the key scope this address belongs to.
+    scope_id BIGINT NOT NULL,
```
GustavoStingelin: I think we no longer need this information for addresses that are not wallet-derived. In that case, the address should belong only to the wallet, and we should not assume that we know the parent scope path used for the imported address or account.

If we still need to filter by this later, I think we can model that distinction through the script type column instead.

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
```diff
@@ -8,42 +8,33 @@
...
+
+    -- Shape marker. TRUE means this address must have a derived_addresses child
+    -- row with account ownership and BIP44 path data. Raw imported addresses
+    -- leave this FALSE.
+    is_derived BOOLEAN NOT NULL,
```
GustavoStingelin, line 24: Same as with accounts, we solved one problem with `is_derived`, but I think we are introducing unnecessary complexity with the extra table.

This could be represented more simply with three nullable fields, `account_id`, `address_branch`, and `address_index`, plus a check constraint to enforce the derived address invariant instead of a trigger.

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -0,0 +1,113 @@
...
+  metadata: wallet, scope, name, account-level public key, master fingerprint,
+  and next external/internal derivation indexes.
+- `derived_accounts` stores only wallet-derived BIP44 account numbers. Imported
+  xpub accounts are `accounts` rows with no `derived_accounts` child.
+- `addresses` is the stable address identity table. It holds wallet, scope,
+  script pubkey, script type, creation time, imported public key material, and a
+  structural `is_derived` bit.
+- `derived_addresses` stores HD child ownership and path data: account ID,
+  branch, and index. Raw imported addresses are `addresses` rows with no
+  `derived_addresses` child.
```
GustavoStingelin:
```suggestion
- `accounts` is the stable account identity table. It represents the parent
  account used to group and derive addresses. It holds shared account metadata:
  wallet, scope, name, account public key, master fingerprint, and next
  external/internal derivation indexes.
- `derived_accounts` stores only accounts derived from a wallet scope parent,
  such as BIP44 account numbers. Imported xpub accounts remain `accounts` rows
  with no `derived_accounts` child.
- `addresses` is the stable address identity table. It represents the final
  address known by the wallet. It holds wallet, scope, script pubkey, script
  type, creation time, imported public key material, and structural
  `is_derived`.
- `derived_addresses` stores only addresses derived from an account and records
  account ID, branch, and index. Raw imported addresses remain `addresses` rows
  with no `derived_addresses` child.
```

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -0,0 +1,113 @@
...
+
+## 4. Implementation Notes
+
+- Modify the existing unmerged account and address migrations in place. Do not
+  add new migration numbers for this feature-branch schema rewrite.
+- `db.AccountInfo` exposes SQL `AccountID` and makes `AccountNumber` optional.
+  The kvdb backend leaves `AccountID` nil.
+- `db.AddressInfo` makes `AccountID` and `AccountNumber` optional. Raw imports
+  have neither; imported-xpub child addresses have an account ID but no BIP44
+  account number.
```
GustavoStingelin: I guess this part is a bit misleading. IIUC, the cases are:

* Raw imported accounts do not have an account number. Their child addresses are still derived from the imported xkey, so those addresses need a branch, index, and account parent.

* Raw imported addresses do not have an account parent, derivation parent, branch, or index.

* Derived accounts have a scope parent.

* Derived addresses have an account parent, and that account can be either wallet-derived or imported.

It would be very helpful to have a table covering these cases, either here in this ADR or in another document.

GustavoStingelin (changes requested):
I like the move from `origin` to `is_derived`. That makes the row shape much more explicit and gives the constraints a clearer field to key off.

The part I am less convinced by is moving `account_number` and the address derivation fields into child tables for the final schema. With the fields we have today, this still feels like overnormalization to me. I would rather move in the direction of nullable fields with strong local `CHECK` constraints.

For accounts, a nullable `account_number` is simple to reason about. For addresses, nullable derivation fields such as `account_id`, `branch`, and `index` can express the same invariant locally. The rule stays in one row: if the row is derived, the derivation fields must be present; if it is not derived, they must be null.

The child table version turns that into a relationship across tables. Now we need joins to reconstruct the normal application view, plus triggers or equivalent checks to keep `is_derived` and child table membership consistent. That adds more failure surface and more review surface for a small amount of data.

One thing that makes me lean this way is that the current queries already `LEFT JOIN` the child tables. So the application-facing row is still a nullable projection. The schema mostly moves nullability from base table columns into optional joined rows, while adding extra consistency work around it.

Just to be clear, I do not think this is a correctness bug. Child tables are defensible if we want strict subtype boundaries. But if we want to go that strict, I think it may be cleaner to go further and use tables specialized by function, for example, one table for derived records and another for imported records, without sharing this subtype relationship.

Under the final schema assumption, I would prefer the nullable field direction unless the child tables clearly buy enough clarity to justify the extra joins, triggers, and cross-table reasoning.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Addressed in the ADR rewrite. The low-cardinality note was only meant to justify account list hydration: accounts are small enough to start from `accounts` and hydrate account-number facts. I clarified the contrast with address/UTXO/balance reads, which are higher-cardinality and should stay on selective indexed paths with explicit shape checks.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Agreed. I rewrote the ADR to describe the final SQL-wallet model directly instead of narrating intermediate commit states. The main narrative now focuses on the master/fake identity problem, the target shape, nullable account numbers, accountless raw imported addresses, and the final invariants.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Addressed by adapting this wording to the final model. The main difference is that accounts no longer have a `derived_accounts` child table after the simplification above, and raw imported addresses no longer have scope. The ADR text now separates account identity, derived address path identity, imported xpub accounts, and raw imported addresses.

wallet/internal/sql/pg/migrations/000005_accounts.up.sql
yyforyongyu: Agreed, this was the better tradeoff for accounts. The latest version drops `derived_accounts` and keeps a single `accounts` table with `is_derived` plus nullable `account_number`. That still removes the fake imported-account number `0`, keeps imported xpub accounts stable by `account_id`, and enforces the derived/imported distinction with a simple row-local invariant. Thanks for pushing on this simplification.

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
yyforyongyu: Good catch. Raw imported addresses should not carry a scope because scope is derivation identity, and these rows only represent wallet-local script identity. I removed `addresses.scope_id`; derived address scope now resolves through `derived_addresses.account_id -> accounts.scope_id`. Raw imports are selected as wallet-owned, non-derived addresses, not through a fake scoped imported account.

wallet/address_manager.go
```diff
@@ -674,8 +684,8 @@ func (w *Wallet) ListAddresses(ctx context.Context, accountName string,
 	addresses := w.store.IterAddresses(
 		ctx, db.ListAddressesQuery{
 			WalletID:    w.id,
-			AccountName: accountName,
-			Scope:       db.KeyScope(keyScope),
+			AccountName: &accountName,
```
yyforyongyu: Raw imported addresses are now accountless: the SQL address store only reaches them when both `AccountName` and `Scope` are nil. This always sends `(Scope, AccountName)`, so `ListAddresses(ctx, "imported", ...)` returns an empty page after `ImportPublicKey`/`ImportTaprootScript` even though those imports were just stored. Special-case the reserved imported alias to use the raw-import query shape (and filter by the requested address type if that API still needs it). (`gpt-5.5`)

wallet/internal/sql/pg/queries/utxos.sql
```diff
@@ -148,11 +148,15 @@ SELECT
...
-    ks.coin_type, -- BIP-43 key scope coin type
+    a.is_derived AS address_is_derived,
+    da.address_id AS derived_address_id,
+    da.account_id,
+    acc.is_derived AS account_is_derived,
+    acc.account_number,
+    a.script_type_id AS type_id,
+    ks.purpose, -- script type, used for coin selection
+    ks.coin_type, -- BIP-43 key scope purpose
+    acc.account_name, -- BIP-43 key scope coin type
```
yyforyongyu, line 159: Raw imported addresses have no `derived_addresses`/`accounts` row, so this selected account name is NULL and `UtxoInfo.AccountName` stays empty. The wallet-level `ListUnspent` filter still compares the public `"imported"` alias directly to `AccountName`, which drops every raw-imported UTXO for `UtxoQuery{Account: "imported"}` even though unfiltered reads include them. Either surface the alias for raw-import rows or special-case the alias in the wallet filter; mirror the same behavior for SQLite. (`gpt-5.5`)

wallet/address_manager.go
yyforyongyu: Fixed in `wallet: handle raw imported aliases`. `ListAddresses(ctx, "imported", ...)` now uses the raw-import query shape (`AccountName`/`Scope` nil) and filters the returned raw imported rows by the requested wallet address type, so imported pubkey/script/taproot rows stay accountless in the store but remain reachable through the public alias.

wallet/tx_creator.go
```diff
@@ -1192,6 +1190,49 @@ func (w *Wallet) filterEligibleOutputs(ctx context.Context,
 	return eligible, nil
 }
 
+// outputEligibleForSelection reports whether one wallet UTXO may be used by
+// automatic coin selection for the requested source.
+func (w *Wallet) outputEligibleForSelection(rawImported bool,
+	output *db.UtxoInfo, minconf uint32, currentHeight int32) bool {
+
+	if rawImported && (output.AccountName != "" ||
```
yyforyongyu: This accountless-only check drops kvdb raw imported UTXOs: the kvdb store still returns private-key imported credits with `AccountName == "imported"` and their legacy `KeyScope`, so selecting `ScopedAccount{AccountName: "imported"}` yields an empty eligible set even when spendable imported funds exist. Accept the kvdb imported-alias shape here or normalize kvdb raw-import UTXO metadata before this filter. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🪶

wallet/internal/sql/sqlite/migrations/000005_accounts.up.sql
```diff
@@ -65,64 +37,74 @@ CREATE TABLE accounts (
...
     CHECK (next_internal_index >= 0),
 
-    -- Imported address counter must be non-negative.
-    CHECK (imported_key_count >= 0),
+    -- Account numbers must be non-negative when present.
+    CHECK (account_number IS NULL OR account_number >= 0),
+
+    -- Derived account shape must match account-number presence.
+    CHECK (
+        (is_derived AND account_number IS NOT NULL)
```
yyforyongyu, line 60: Reworked and force-pushed. The former follow-up commits are gone: the helper rename and raw-import alias fixes are folded into the introducing commits, and the DB itest coverage is split back into fixture/account/list/address/UTXO commits.

I also tried keeping the SQL/sqlc changes as a standalone commit, but `scripts/check-each-commit.sh sql-wallet` fails there because regenerated sqlc removes/renames APIs that the existing pg/sqlite adapters still compile against. To keep every commit buildable, the schema/sqlc update and the dependent db/wallet adapter changes are together in `6d31164a0 db+wallet: normalize SQL identity`.

wallet/internal/db/itest/accountstore_helpers_test.go
```diff
@@ -90,6 +95,15 @@ func getAccountQueryByNumber(walletID uint32, scope db.KeyScope,
 	}
 }
 
+// accountNumberNotNil requires a derived account number and returns its value.
+func accountNumberNotNil(t *testing.T, accountNumber *uint32) uint32 {
```
yyforyongyu, line 99: Folded. The helper is now introduced under its final `accountNumberNotNil` name in `e13163ad1 db/itest: cover account creation`; there is no later rename commit in the stack.

wallet/address_manager.go
```diff
@@ -654,35 +664,52 @@ func (w *Wallet) ListAddresses(ctx context.Context, accountName string,
...
 		return nil, err
 	}
 
+	query := db.ListAddressesQuery{
+		WalletID: w.id,
+		Page:     req,
+	}
+
+	var storeAddrType addresstype.StoreType
+	if accountName == db.DefaultImportedAccountName {
```
yyforyongyu: Folded. The raw imported `ListAddresses` and UTXO alias handling are now in `6d31164a0 db+wallet: normalize SQL identity`, with the wallet regression tests included there instead of a follow-up commit.

wallet/tx_creator.go
yyforyongyu: Fixed in `6d31164a0`. The imported-alias filter now accepts both SQL accountless raw imports and kvdb legacy raw imports with `AccountName == "imported"` plus the matching key scope. `TestGetEligibleUTXOsImportedAccountBypassesGetAccount` covers both shapes and excludes a different-scope imported row.

yyforyongyu:
No issues found by gpt-5.5 🧭

wallet/internal/sql/pg/migrations/000006_addresses.up.sql
yyforyongyu: I kept `derived_addresses` after simplifying accounts because the address split carries more weight than the old account split: the base `addresses` row stays pure wallet/script identity, while only derived rows get account, branch, and index facts. A nullable-column design would work too, but the child table keeps raw imports out of derived-path indexes and makes raw-vs-derived shape checks explicit. The ADR now calls out that tradeoff.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Addressed. I added an ADR table for the concrete cases and switched to `imported xpub account` terminology so account-level xpub imports are not confused with raw address imports. The table calls out that imported-xpub child addresses are user-imported but still derived from an account xpub, so they have account/branch/index facts; raw imported addresses have none of those.

wallet/internal/db/itest/accountstore_helpers_test.go
yyforyongyu: Done. The helper is now introduced as `accountNumberNotNil` in the final stack, and the comment matches the assertion it performs.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: > I think the ADRs should compare the behavior on master with the target behavior on the sql wallet branch, instead of changing the mental model across commits.

Reflecting more on this - I don’t think that is quite the right framing. There isn’t really a master implementation to compare against here, since this is SQL-wallet-only schema work.

More importantly, ADRs are moment-in-time decision records. They should capture why we made a decision, what alternatives we considered, and why some approaches were rejected. That history is useful: if we remove the context for discarded approaches, someone can easily revisit the same design later without understanding why we moved away from it.

I agree the ADR should not read like a commit-by-commit narrative, but I do think it should preserve the relevant decision context and rejected alternatives.

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🆔

GustavoStingelin:
@yyforyongyu do you think this is ready for review now?

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -0,0 +1,195 @@
+# ADR 0013: Normalized Account and Address Identity
+
+## 1. Context
+
+The SQL wallet store originally modeled account and address variants by mixing
+shared identity columns with nullable subtype columns:
```
yyforyongyu, line 6: we start with the ambiguity of the meaning of `is_imported` - at the top, the wallet can be imported via seed; at the account level, the account can be imported with xpriv; at the address level, the addr can be imported with a script. This quickly becomes unclear from the code's perspective, as the code only cares whether the item can be derived, but not where it's from. For an imported wallet, its accounts can be derived; for an imported account, its addresses can be derived. Thus we care only about the ability to derive in the db, not the user's context on the being `imported` or not.

So we need to update the ADR for that.

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -0,0 +1,195 @@
...
+- Raw imported addresses no longer require a fake account row.
+- Raw imports no longer carry fake scope or derivation-path identity.
+- Imported-xpub child addresses and wallet-derived child addresses share the
+  same derived-address path model.
+- Derived address queries can start from account/path facts, while raw import
+  queries can start from wallet-local script identity.
+
+Cons:
+
+- Address reads need additional joins and generated query churn.
```
yyforyongyu, line 169: be more specific here? i think some reads actually benefit from the splitting of the tables?

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -0,0 +1,195 @@
...
+- Every address, UTXO, and balance read path must use the same corruption checks
```
yyforyongyu: I don't think so - once we validate and control the data being saved in the tables, there is no need for UTXO and balance read to enforce the corruption checks as it's impossible to corrupt in the first place?

wallet/internal/db/test_helpers_test.go
```diff
@@ -0,0 +1,6 @@
+package db
+
+// testUint32Ptr returns a pointer to v for nullable uint32 assertions.
+func testUint32Ptr(v uint32) *uint32 {
```
yyforyongyu: why do we need these superficial helpers?

wallet/test_helpers_test.go
```diff
@@ -0,0 +1,6 @@
+package wallet
+
+// testUint32Ptr returns a pointer to v for nullable uint32 assertions.
+func testUint32Ptr(v uint32) *uint32 {
```
yyforyongyu: why do we need these superficial helpers?

wallet/internal/db/sqlite/accountstore_helpers.go
```diff
@@ -42,6 +42,14 @@ func derivedAddressGetAccountNumber(
 	return db.DerivedAddressAccountNumber(row.AccountNumber)
 }
 
+// derivedAddressGetAccountIsDerived reports whether the account row is
+// wallet-derived.
+func derivedAddressGetAccountIsDerived(
```
yyforyongyu: why add this thin helper??

wallet/internal/db/pg/accountstore_helpers.go
```diff
@@ -42,6 +42,14 @@ func derivedAddressGetAccountNumber(
 	return db.DerivedAddressAccountNumber(row.AccountNumber)
 }
 
+// derivedAddressGetAccountIsDerived reports whether the account row is
+// wallet-derived.
+func derivedAddressGetAccountIsDerived(
```
yyforyongyu: why add this thin helper??

wallet/internal/db/kvdb/accountstore_createimported.go
```diff
@@ -209,14 +213,26 @@ func dryRunImportedAccount(ns walletdb.ReadWriteBucket,
...
-		return nil
+		// A nil schema intentionally maps to no per-account override.
+		//nolint:nilnil
+		return nil, nil
 	}
 
-	return &waddrmgr.ScopeAddrSchema{
-		ExternalAddrType: waddrmgr.AddressType(schema.ExternalAddrType),
-		InternalAddrType: waddrmgr.AddressType(schema.InternalAddrType),
+	external, err := addresstype.ToWallet(schema.ExternalAddrType, false)
```
yyforyongyu: why adding this check? this is unnecessary for `kvdb`? we should just keep the legacy behavior here

wallet/internal/db/kvdb/addressstore.go
```diff
@@ -190,6 +185,47 @@ func (s *Store) NewImportedAddress(ctx context.Context,
 	return info, nil
 }
 
+// importAddressTx performs the imported-address write within an existing
+// write transaction and returns the resulting address info.
+func (s *Store) importAddressTx(ns walletdb.ReadWriteBucket,
```
yyforyongyu, line 190: i wanna make it clear that the new sql schema only applies to `sqlite` and `pg`, for the `kvdb` backend, we can just assume the imported account is still using the constant and fixed name `imported` - we don't wanna expand our design into kvdb given that it will be deprecated and migrated into sql soon.

wallet/internal/db/kvdb/txstore.go
```diff
@@ -23,6 +24,11 @@ var _ db.TxStore = (*Store)(nil)
 // signed legacy wtxmgr height domain.
 var errLegacyHeightOverflow = errors.New("legacy height overflows int32")
 
+// legacyTxLabelsBucketKey is the private wtxmgr tx-label bucket name. The
+// legacy module exposes label writes but not deletes, while the db.Store
+// contract treats an empty label as clearing the label.
+var legacyTxLabelsBucketKey = []byte("l")
```
yyforyongyu: we should not introduce new bucket for kvdb here - remember, we just wanna leave kvdb's behavior unchanged internally as it's a legacy format

yyforyongyu:
> @yyforyongyu do you think this is ready for review now?

Not yet, let me review first, will request when done!

Abdulkbk:
> Not yet, let me review first, will request when done!

ping me too, started reviewing but paused because of the rebases.

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🧬

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Addressed in the ADR rewrite. The final text no longer claims UTXO or balance reads need corruption checks once write-time constraints make those invalid shapes unrepresentable.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Resolved by adding the concrete case table to the ADR. The table now distinguishes wallet-derived accounts, imported xpub accounts, child addresses derived from either account type, and raw imported accountless addresses.

wallet/internal/db/pg/accountstore_helpers.go
yyforyongyu: Resolved in the rewrite by dropping the thin PostgreSQL helper. The account-store code now keeps that path explicit instead of adding a wrapper with no independent behavior.

wallet/internal/db/kvdb/accountstore_createimported.go
yyforyongyu: Resolved in the rewrite by leaving kvdb internals on the legacy behavior. The normalized account-number validation is now SQL-only and kvdb does not get this additional check.

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🧩

yyforyongyu:
No issues found by gpt-5.5 🧭

GustavoStingelin (approved):
LGTM.

GustavoStingelin:
@Abdulkbk ping

PR #1269, opened 2026-06-17
db+waddrmgr: add scan-horizon scaffolding and rollback eviction

yyforyongyu:
Adds the scan-horizon derivation surface: scan-batch request types, SQL
scan-batch account queries, scan-horizon account types, and the horizon
extension helpers (bounded by `MaxAddressIndex`, skipping `ErrInvalidChild`).
Also adds the `waddrmgr` rollback primitives `RestoreSyncedTo` (restore a
rolled-back sync tip) and `EvictDerivedAddresses` (evict rolled-back derived
horizons), with cache-eviction tests.

Split out of the runtime Store PR for reviewability; content-preserving, no
commit changes. Stacked on the scan-batch scaffolding PR (`prep-runtime-2`,
#1263).

### DIFF
--- a/waddrmgr/interface.go
+++ b/waddrmgr/interface.go
@@ -354,6 +354,13 @@ type AccountStore interface {
 	// InvalidateAccountCache invalidates the account cache.
 	InvalidateAccountCache(account uint32)
 
+	// EvictDerivedAddresses removes the in-memory traces (recent-address
+	// cache entries and pending unlock-derivation entries) of the
+	// addresses derived for the given account branch over the half-open
+	// index range [fromIndex, toIndex), undoing a rolled-back horizon
+	// extension.
+	EvictDerivedAddresses(account, branch, fromIndex, toIndex uint32)
+
 	// ImportPrivateKey imports a private key.
 	ImportPrivateKey(ns walletdb.ReadWriteBucket, wif *btcutil.WIF,
 		bs *BlockStamp) (ManagedPubKeyAddress, error)
--- a/waddrmgr/scoped_manager.go
+++ b/waddrmgr/scoped_manager.go
@@ -2918,6 +2918,39 @@ func (s *ScopedKeyManager) InvalidateAccountCache(account uint32) {
 	delete(s.acctInfo, account)
 }
 
+// EvictDerivedAddresses removes the in-memory traces of the addresses derived
+// for the given account branch over the half-open index range [fromIndex,
+// toIndex) from the live manager. It drops those addresses from the
+// recent-address cache, discards any matching pending deriveOnUnlock entries,
+// and invalidates the cached account info so the bumped next-index and
+// last-address fields are reloaded from the committed state.
+//
+// extendAddresses mutates this in-memory state synchronously, before the
+// surrounding walletdb transaction commits, so a horizon extension that is
+// later rolled back would otherwise leave scan-derived addresses observable
+// through Address, queued for derivation on the next unlock, and reflected in
+// the account's next-index and last-address fields even though they were never
+// persisted. This lets the caller undo exactly those in-memory mutations for a
+// rolled-back batch without disturbing the persisted state.
+func (s *ScopedKeyManager) EvictDerivedAddresses(account, branch, fromIndex,
+	toIndex uint32) {
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	s.evictDerivedAddressCache(account, branch, fromIndex, toIndex)
+	s.evictDeriveOnUnlock(account, branch, fromIndex, toIndex)
+
+	// extendAddresses also advanced this account's cached next-index and
+	// last-address fields. Those were not persisted by the rolled-back
+	// transaction, so drop the cached account entry as well; loadAccountInfo
+	// reloads the committed state from the database on the next read instead
+	// of exposing the uncommitted horizon through AccountProperties or the
+	// last-address getters. The lock is already held, so the delete is inlined
+	// rather than routed through InvalidateAccountCache.
+	delete(s.acctInfo, account)
+}
+
 // NewAddress returns a new address for the given account. The `change`
 // parameter dictates whether a change address (internal) or a receiving
 // address (external) should be generated. The caller is responsible for
@@ -2960,6 +2993,91 @@ func (s *ScopedKeyManager) NewAddress(addrmgrNs walletdb.ReadWriteBucket,
 	addr := addrs[0].Address()
 
 	return addr, nil
+}
+
+// evictDerivedAddressCache removes rolled-back derived addresses from the
+// recent-address cache by scanning the bounded in-memory cache entries.
+func (s *ScopedKeyManager) evictDerivedAddressCache(account, branch, fromIndex,
+	toIndex uint32) {
+
+	var keys []addrKey
+	s.addrs.Range(func(key addrKey, cachedAddr *cachedAddr) bool {
+		if cachedAddr == nil {
+			return true
+		}
+
+		match := derivedAddressMatches(
+			cachedAddr.addr, account, branch, fromIndex, toIndex,
+		)
+		if match {
+			keys = append(keys, key)
+		}
+
+		return true
+	})
+
+	for _, key := range keys {
+		s.addrs.Delete(key)
+	}
+}
+
+// evictDeriveOnUnlock removes rolled-back pending private-key derivations for
+// one account branch and index range.
+func (s *ScopedKeyManager) evictDeriveOnUnlock(account, branch, fromIndex,
+	toIndex uint32) {
+
+	// Drop the pending unlock-derivation entries the rolled-back extension
+	// queued for this account, branch, and range, filtering in place so entries
+	// from other accounts, branches, or unrelated extensions survive.
+	filtered := s.deriveOnUnlock[:0]
+	for _, info := range s.deriveOnUnlock {
+		if pendingDeriveMatches(info, account, branch, fromIndex, toIndex) {
+			continue
+		}
+
+		filtered = append(filtered, info)
+	}
+
+	// Clear the tail the in-place filter left behind so the dropped entries'
+	// managed addresses can be garbage collected.
+	for i := len(filtered); i < len(s.deriveOnUnlock); i++ {
+		s.deriveOnUnlock[i] = nil
+	}
+
+	s.deriveOnUnlock = filtered
+}
+
+// pendingDeriveMatches reports whether one unlock-time derivation request is in
+// the rolled-back account branch and index range.
+func pendingDeriveMatches(info *unlockDeriveInfo, account, branch, fromIndex,
+	toIndex uint32) bool {
+
+	if info == nil || info.managedAddr == nil {
+		return false
+	}
+
+	return info.managedAddr.InternalAccount() == account &&
+		info.branch == branch &&
+		info.index >= fromIndex && info.index < toIndex
+}
+
+// derivedAddressMatches reports whether one cached address belongs to the
+// rolled-back account branch and index range.
+func derivedAddressMatches(addr ManagedAddress, account, branch, fromIndex,
+	toIndex uint32) bool {
+
+	addrWithPath, ok := addr.(ManagedPubKeyAddress)
+	if !ok {
+		return false
+	}
+
+	_, path, ok := addrWithPath.DerivationInfo()
+	if !ok {
+		return false
+	}
+
+	return path.InternalAccount == account && path.Branch == branch &&
+		path.Index >= fromIndex && path.Index < toIndex
 }
 
 // getNextIndex fetches the current next address index for the specified branch
--- a/wallet/db_ops.go
+++ b/wallet/db_ops.go
@@ -27,6 +27,8 @@ var (
 	// ErrMissingTxManager is returned when the transaction manager namespace is
 	// missing from the database.
 	ErrMissingTxManager = errors.New("missing transaction manager namespace")
+
+	errUnknownBranch = errors.New("unknown branch")
 )
 
 // DBCreateWallet initializes the database structure for a new wallet.
@@ -525,12 +527,16 @@ func (s *syncer) DBPutRewind(_ context.Context,
 func (s *syncer) DBPutSyncBatch(ctx context.Context,
 	results []scanResult) error {
 
+	var horizonRollback addrHorizonRollback
+
 	// TODO(yy): build a single SQL query for this.
 	err := walletdb.Update(s.cfg.DB, func(dbtx walletdb.ReadWriteTx) error {
 		addrmgrNs := dbtx.ReadWriteBucket(waddrmgrNamespaceKey)
 
 		// 1. Update Address State (Horizons).
-		err := s.putAddrHorizons(ctx, addrmgrNs, results)
+		var err error
+
+		horizonRollback, err = s.putAddrHorizons(ctx, addrmgrNs, results)
 		if err != nil {
 			return err
 		}
@@ -559,6 +565,8 @@ func (s *syncer) DBPutSyncBatch(ctx context.Context,
 		return nil
 	})
 	if err != nil {
+		horizonRollback.evict()
+
 		return fmt.Errorf("process scan batch: %w", err)
 	}
 
@@ -574,11 +582,15 @@ func (s *syncer) DBPutSyncBatch(ctx context.Context,
 func (s *syncer) DBPutTargetedBatch(ctx context.Context,
 	results []scanResult) error {
 
+	var horizonRollback addrHorizonRollback
+
 	err := walletdb.Update(s.cfg.DB, func(dbtx walletdb.ReadWriteTx) error {
 		addrmgrNs := dbtx.ReadWriteBucket(waddrmgrNamespaceKey)
 
 		// 1. Update Address State (Horizons).
-		err := s.putAddrHorizons(ctx, addrmgrNs, results)
+		var err error
+
+		horizonRollback, err = s.putAddrHorizons(ctx, addrmgrNs, results)
 		if err != nil {
 			return err
 		}
@@ -592,6 +604,8 @@ func (s *syncer) DBPutTargetedBatch(ctx context.Context,
 		return nil
 	})
 	if err != nil {
+		horizonRollback.evict()
+
 		return fmt.Errorf("process rescan batch: %w", err)
 	}
 
@@ -740,13 +754,56 @@ func (s *syncer) filterBranchScopes(_ context.Context, dbtx walletdb.ReadTx,
 	return scopes, nil
 }
 
-// putAddrHorizons aggregates found address horizons from the scan
-// results and updates the address manager state (extends horizons) in the
-// database.
-func (s *syncer) putAddrHorizons(_ context.Context,
-	ns walletdb.ReadWriteBucket, results []scanResult) error {
+// addrHorizonExtension records one in-memory horizon extension that should be
+// evicted if the surrounding walletdb transaction rolls back.
+type addrHorizonExtension struct {
+	scopedMgr waddrmgr.AccountStore
+	account   uint32
+	branch    uint32
+	fromIndex uint32
+	toIndex   uint32
+}
 
-	// Aggregate Horizon Expansion.
+// addrHorizonRollback records successful in-memory horizon extensions that
+// still depend on the surrounding walletdb transaction committing.
+type addrHorizonRollback []addrHorizonExtension
+
+// evict removes all in-memory horizon extensions tracked for a failed batch.
+func (r addrHorizonRollback) evict() {
+	for _, extension := range r {
+		extension.scopedMgr.EvictDerivedAddresses(
+			extension.account,
+			extension.branch,
+			extension.fromIndex,
+			extension.toIndex,
+		)
+	}
+}
+
+// branchNextIndex returns the next child index for an account branch.
+func branchNextIndex(scopedMgr waddrmgr.AccountStore,
+	ns walletdb.ReadBucket, account, branch uint32) (uint32, error) {
+
+	props, err := scopedMgr.AccountProperties(ns, account)
+	if err != nil {
+		return 0, fmt.Errorf("account properties: %w", err)
+	}
+
+	switch branch {
+	case waddrmgr.ExternalBranch:
+		return props.ExternalKeyCount, nil
+
+	case waddrmgr.InternalBranch:
+		return props.InternalKeyCount, nil
+
+	default:
+		return 0, fmt.Errorf("%w: %d", errUnknownBranch, branch)
+	}
+}
+
+// scanBatchHorizons returns the greatest child index discovered for each branch
+// across a scan batch.
+func scanBatchHorizons(results []scanResult) map[waddrmgr.BranchScope]uint32 {
 	batchHorizons := make(map[waddrmgr.BranchScope]uint32)
 	for _, res := range results {
 		for bs, idx := range res.FoundHorizons {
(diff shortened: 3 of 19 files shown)


waddrmgr/interface.go
```diff
@@ -89,6 +89,11 @@ type AddrStore interface {
 	// address manager is synced through at the very least.
 	SyncedTo() BlockStamp
 
+	// RestoreSyncedTo resets the in-memory synced-to block stamp to the
+	// given value without touching the database, undoing synced-tip
+	// advances whose bucket writes were rolled back.
+	RestoreSyncedTo(bs BlockStamp)
+
```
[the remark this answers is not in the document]
yyforyongyu: Intentional. `waddrmgr.AddrStore`/`AccountStore` are the canonical Store contracts for the in-progress SQL-wallet work, which is a pre-release feature branch not yet merged to master; rollback support is part of that contract and every implementation is updated together in-tree. An optional-interface + type-assertion shim would add indirection for an external-compatibility guarantee we do not need pre-release. Resolving as intended.

waddrmgr/interface.go
```diff
@@ -348,6 +353,13 @@ type AccountStore interface {
 	// InvalidateAccountCache invalidates the account cache.
 	InvalidateAccountCache(account uint32)
 
+	// EvictDerivedAddresses removes the in-memory traces (recent-address
+	// cache entries and pending unlock-derivation entries) of the
+	// addresses derived for the given account branch over the half-open
+	// index range [fromIndex, toIndex), undoing a rolled-back horizon
+	// extension.
+	EvictDerivedAddresses(account, branch, fromIndex, toIndex uint32)
+
```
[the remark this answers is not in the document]
yyforyongyu: Intentional. `waddrmgr.AddrStore`/`AccountStore` are the canonical Store contracts for the in-progress SQL-wallet work, which is a pre-release feature branch not yet merged to master; rollback support is part of that contract and every implementation is updated together in-tree. An optional-interface + type-assertion shim would add indirection for an external-compatibility guarantee we do not need pre-release. Resolving as intended.

wallet/internal/sql/sqlite/queries/accounts.sql
```diff
@@ -434,3 +434,20 @@ WHERE
...
     )
 GROUP BY da.account_id;
+
+-- name: AdvanceNextExternalIndex :exec
+-- Advances the external branch's next index to the supplied value during
+-- recovery horizon extension. The MAX guard keeps the counter monotonic so a
+-- slower concurrent writer cannot regress it below an already-recorded index.
+UPDATE accounts
+SET next_external_index = max(next_external_index, sqlc.arg('next_index'))
+WHERE id = sqlc.arg('id');
```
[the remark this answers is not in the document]
yyforyongyu: Confirmed and valid — SQLite generates `NextIndex interface{}` here (vs `int64` on Postgres) because of `max(col, sqlc.arg('next_index'))`. The fix — cast the arg to `INTEGER` (or a CASE) and regenerate sqlc — is deferred to a follow-up so it lands with the next prep-runtime-3 update rather than re-pushing the stack now. Resolving; tracked for that follow-up.

wallet/internal/sql/sqlite/queries/accounts.sql
```diff
@@ -434,3 +434,20 @@ WHERE
...
+
+-- name: AdvanceNextInternalIndex :exec
+-- Advances the internal/change branch's next index to the supplied value
+-- during recovery horizon extension. The MAX guard keeps the counter monotonic
+-- so a slower concurrent writer cannot regress it below an already-recorded
+-- index.
+UPDATE accounts
+SET next_internal_index = max(next_internal_index, sqlc.arg('next_index'))
+WHERE id = sqlc.arg('id');
```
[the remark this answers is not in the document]
yyforyongyu: Confirmed and valid — SQLite generates `NextIndex interface{}` here (vs `int64` on Postgres) because of `max(col, sqlc.arg('next_index'))`. The fix — cast the arg to `INTEGER` (or a CASE) and regenerate sqlc — is deferred to a follow-up so it lands with the next prep-runtime-3 update rather than re-pushing the stack now. Resolving; tracked for that follow-up.

waddrmgr/sync.go
```diff
@@ -69,6 +69,23 @@ func (m *Manager) SetSyncedTo(ns walletdb.ReadWriteBucket, bs *BlockStamp) error
...
+// when several synced blocks are applied inside one transaction and a later
+// step fails, walletdb rolls the bucket back but the in-memory tip stays
+// advanced. Restoring it to the pre-batch stamp keeps the in-memory tip in step
+// with the persisted state.
+func (m *Manager) RestoreSyncedTo(bs BlockStamp) {
+	m.mtx.Lock()
+	defer m.mtx.Unlock()
+
+	m.syncState.syncedTo = bs
+}
```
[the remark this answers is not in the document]
yyforyongyu: Valid — `RestoreSyncedTo` has no dedicated rollback-scenario test yet. Adding one (advance `SetSyncedTo` inside a tx that rolls back, assert `SyncedTo()` is ahead of the persisted value, then `RestoreSyncedTo` to restore the pre-tx stamp) is deferred to a follow-up to avoid re-pushing the stack now. Resolving; tracked for that follow-up.

yyforyongyu:
No issues found by `gpt-5.5 xhigh` 🦦

yyforyongyu:

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🛰️

yyforyongyu:
No issues found by gpt-5.5 🪁

wallet/internal/db/data_types.go
```diff
@@ -1089,6 +1105,51 @@ type TxBatchParams struct {
...
+	Scope KeyScope
+
+	// Account is the account number containing the branch. It is a derived
+	// fast-path identity only: both backends mask an imported account's
+	// number to 0 when emitting the horizon, so a backend resolving the
+	// horizon's owning account MUST prefer AccountName and treat Account as
+	// an optimization that is only trustworthy for derived accounts.
+	Account uint32
+
+	// AccountName is the durable, backend-agnostic identity of the
```
yyforyongyu: Account names are renameable and can be reused after a rename, so treating this field as durable can resolve a stale imported-xpub horizon to a different account; carry the stable `AccountID`/row id instead. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔑

yyforyongyu:
No issues found by gpt-5.5 🪐

wallet/internal/db/pg/txstore_applyscanbatch.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return nil
+}
+
+// horizonAccountByID resolves the horizon's owning account by stable account
+// row ID and verifies that the account belongs to the horizon's key scope.
+func (o scanHorizonOps) horizonAccountByID(ctx context.Context,
+	accountID uint32, scope db.KeyScope) (*db.HorizonAccount, error) {
+
+	row, err := o.qtx.GetAccountPropsById(ctx, int64(accountID))
```
yyforyongyu: This unlocked read lets PostgreSQL scan-batch horizon extension derive from a stale next index while a concurrent address allocation claims the same child, producing a permanent unique-constraint failure; lock the account row or reserve the range before deriving. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🪄

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🛰️

wallet/internal/db/scan_batch_common.go
```diff
@@ -0,0 +1,234 @@
...
+	// against the wrong next-index column.
+	branch := horizon.Branch
+	if branch != externalBranch && branch != internalBranch {
+		return 0, fmt.Errorf("extend horizon: %w: branch %d",
+			ErrInvalidParam, branch)
+	}
+
+	// kvdb caps a single extension at MaxAddressesPerAccount; mirror the bound
+	// so both backends reject the same out-of-range horizon.
+	if horizon.Index > MaxAddressIndex {
```
yyforyongyu: Move this max-index check until after loading the account and applying the `horizon.Index < nextIndex` no-op, because SQL accounts can already have ordinary allocations beyond `MaxAddressIndex` and those covered horizons should not be rejected (`gpt-5.5 xhigh`).

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🛰️

yyforyongyu:
No issues found by gpt-5.5 🧬

yyforyongyu:
No issues found by `gpt-5.5 xhigh` 🧭

yyforyongyu:
No issues found by `gpt-5.5 xhigh` 🧭

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` ♻️

waddrmgr/scoped_manager.go
```diff
@@ -2917,6 +2917,82 @@ func (s *ScopedKeyManager) InvalidateAccountCache(account uint32) {
...
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	// The account info is needed to re-derive the cached addresses' script
+	// addresses, which are the cache keys. A rolled-back extension cannot have
+	// evicted the account from the cache, but guard regardless: with no cached
+	// account there is nothing this manager could have derived.
+	acctInfo, ok := s.acctInfo[account]
+	if !ok {
+		return
```
yyforyongyu: After one branch is evicted, the call below deletes the account from `acctInfo`, so a second rollback eviction for the same account (for example external and internal horizons in one failed scan batch) hits this early return and leaves the later branch’s cached addresses/pending unlock entries in memory. Delay account-cache invalidation until all requested branch ranges are evicted, or make the eviction path still remove per-branch state after the account cache has already been dropped. (`gpt-5.5`)

yyforyongyu:
No issues found by `gpt-5.5 xhigh` 🧩

wallet/internal/sql/sqlite/queries/accounts.sql
```diff
@@ -434,3 +434,20 @@ WHERE
...
+SET next_external_index = max(next_external_index, sqlc.arg('next_index'))
+WHERE id = sqlc.arg('id');
+
+-- name: AdvanceNextInternalIndex :exec
+-- Advances the internal/change branch's next index to the supplied value
+-- during recovery horizon extension. The MAX guard keeps the counter monotonic
+-- so a slower concurrent writer cannot regress it below an already-recorded
+-- index.
+UPDATE accounts
+SET next_internal_index = max(next_internal_index, sqlc.arg('next_index'))
```
yyforyongyu: This has the same unresolved sqlc inference issue as the external-branch update: the uncast next_index argument leaves SQLite with `NextIndex interface{}` instead of an `int64`; cast the argument to `INTEGER` or rewrite the expression before regenerating. (`gpt-5.5`)

wallet/internal/sql/sqlite/queries/accounts.sql
```diff
@@ -434,3 +434,20 @@ WHERE
         OR (acc.is_derived = FALSE AND dacct.account_number IS NULL)
     )
 GROUP BY da.account_id;
+
+-- name: AdvanceNextExternalIndex :exec
+-- Advances the external branch's next index to the supplied value during
+-- recovery horizon extension. The MAX guard keeps the counter monotonic so a
+-- slower concurrent writer cannot regress it below an already-recorded index.
+UPDATE accounts
+SET next_external_index = max(next_external_index, sqlc.arg('next_index'))
```
yyforyongyu: This still makes sqlc infer `NextIndex interface{}` for SQLite, while the PostgreSQL equivalent remains strongly typed as `int64`; cast the next_index argument to `INTEGER` (or use an equivalent CASE) and regenerate so the generated params keep a concrete integer type. (`gpt-5.5`)

wallet/internal/db/sqlite/txstore_applyscanbatch.go
```diff
@@ -0,0 +1,185 @@
...
+
+	return nil
+}
+
+// horizonAccountByID resolves the horizon's owning account by stable account
+// row ID and verifies that the account belongs to the horizon's key scope.
+func (o scanHorizonOps) horizonAccountByID(ctx context.Context,
+	accountID uint32, scope db.KeyScope) (*db.HorizonAccount, error) {
+
+	row, err := o.qtx.GetAccountPropsById(ctx, int64(accountID))
```
yyforyongyu: This lookup only keys on the global account row ID and scope, ignoring `o.walletID`. A scan batch for wallet A can load an account from wallet B; when `horizon.Index` is below that account next index, `ExtendScanHorizon` returns nil before `InsertDerivedAddress` enforces the wallet ID, so the wrong-wallet horizon is accepted. Use a wallet-scoped lookup here. (`gpt-5.5 xhigh`)

wallet/internal/db/pg/txstore_applyscanbatch.go
yyforyongyu: This lookup only keys on the global account row ID and scope, ignoring `o.walletID`. A scan batch for wallet A can load an account from wallet B; when `horizon.Index` is below that account next index, `ExtendScanHorizon` returns nil before `InsertDerivedAddress` enforces the wallet ID, so the wrong-wallet horizon is accepted. Use a wallet-scoped lookup here and keep the row lock. (`gpt-5.5 xhigh`)

yyforyongyu:

yyforyongyu:
No issues found by gpt-5.5 🦆

yyforyongyu:
No issues found by `gpt-5.5 xhigh` 🐳

wallet/internal/db/kvdb/txstore.go
```diff
@@ -311,12 +329,576 @@ func (s *Store) updateLegacyTxLabel(txmgrNs walletdb.ReadWriteBucket,
...
+	params.Transactions = db.SortTxBatchParentsFirst(params.Transactions)
+
+	if s.addrStore == nil {
+		return fmt.Errorf("kvdb.Store.ApplyTxBatch: %w",
+			errMissingAddrStore)
+	}
+
+	var preBatchSyncedTo waddrmgr.BlockStamp
+	if params.SyncedTo != nil {
+		preBatchSyncedTo = s.addrStore.SyncedTo()
```
yyforyongyu: This reverts the #1263 fix by taking the restore snapshot before entering the serialized `walletdb.Update`. Another batch can commit a newer synced tip while this call is waiting; if this update later fails after `SetSyncedTo`, the restore below rewinds memory to the stale pre-lock tip. Keep the snapshot inside the write transaction immediately before `SetSyncedTo`, and only restore that value when this batch actually mutated the in-memory tip. (`gpt-5.5`)

yyforyongyu: Capturing the pre-batch tip here is outside the walletdb write lock. If another writer advances the tip before this update runs and this batch then fails after `SetSyncedTo`, the error path restores this stale value and leaves the live addr manager behind the committed DB state. Capture the snapshot inside the write transaction again. (`gpt-5.5 xhigh`)

wallet/internal/db/kvdb/txstore.go
```diff
@@ -522,9 +1122,244 @@ func (s *Store) InvalidateUnminedTx(_ context.Context,
...
+		return fmt.Errorf("kvdb.Store.RollbackToBlock: height: %w", err)
+	}
+
+	err = walletdb.Update(s.db, func(tx walletdb.ReadWriteTx) error {
+		txmgrNs := tx.ReadWriteBucket(wtxmgrNamespaceKey)
+		if txmgrNs == nil {
+			return errMissingTxmgrNamespace
+		}
+
+		target, err := s.computeRollbackTarget(tx, height, height32)
```
yyforyongyu: Keeping `target` scoped inside the update drops the only state needed to repair the live sync tip if the update fails after `rewindSyncTip` calls `SetSyncedTo`. A commit failure rolls back the DB but leaves the addr manager cache rewound, and the removed restore path/tests used to cover this case. Keep the target visible after the update and restore the prior tip on that failure. (`gpt-5.5 xhigh`)

waddrmgr/scoped_manager.go
```diff
@@ -2917,6 +2917,49 @@ func (s *ScopedKeyManager) InvalidateAccountCache(account uint32) {
...
+// last-address fields are reloaded from the committed state.
+//
+// extendAddresses mutates this in-memory state synchronously, before the
+// surrounding walletdb transaction commits, so a horizon extension that is
+// later rolled back would otherwise leave scan-derived addresses observable
+// through Address, queued for derivation on the next unlock, and reflected in
+// the account's next-index and last-address fields even though they were never
+// persisted. This lets the caller undo exactly those in-memory mutations for a
+// rolled-back batch without disturbing the persisted state.
+func (s *ScopedKeyManager) EvictDerivedAddresses(account, branch, fromIndex,
```
yyforyongyu, line 2935: This eviction primitive is never invoked by the production scan-batch paths. `DBPutSyncBatch` and `DBPutTargetedBatch` still extend horizons first, then later transaction or tip writes can fail and roll back the walletdb transaction after `extendAddresses` has already populated the address cache, pending unlock derivations, and cached next indexes. Wire this into the batch error path around each horizon extension. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🛰️

yyforyongyu:
No issues found by gpt-5.5 🧭

wallet/db_ops.go
```diff
@@ -758,25 +815,63 @@ func (s *syncer) putAddrHorizons(_ context.Context,
...
 
 		err = scopedMgr.ExtendAddresses(
 			ns, bs.Account, maxFoundIndex, bs.Branch,
 		)
 		if err != nil {
-			return fmt.Errorf("extend addresses: %w", err)
+			return rollback, fmt.Errorf("extend addresses: %w", err)
+		}
+
+		toIndex, err := branchNextIndex(
```
yyforyongyu, line 856: If this post-extension `branchNextIndex` read fails, `ExtendAddresses` has already mutated live address-manager caches but the returned rollback list has no entry for this range. Record enough rollback state before any fallible work after `ExtendAddresses`, or evict immediately on this error path. (`gpt-5.5 xhigh`)

wallet/db_ops.go
```diff
@@ -758,25 +815,73 @@ func (s *syncer) putAddrHorizons(_ context.Context,
...
+			scopedMgr, ns, bs.Account, bs.Branch,
+		)
+		if err != nil {
+			// ExtendAddresses mutates live caches before this read. If the
+			// read fails, the caller will not get a rollback entry, so evict
+			// the possible derivation range immediately.
+			if maxFoundIndex >= fromIndex {
+				scopedMgr.EvictDerivedAddresses(
+					bs.Account, bs.Branch, fromIndex,
+					maxFoundIndex+1,
```
yyforyongyu: If `ExtendAddresses` skips an HD-invalid child, it can cache later valid children past `maxFoundIndex+1` before this post-extension read fails. This fallback would then leave those cached addresses/pending unlock entries live after the walletdb rollback; evict through the actual advanced next index, or otherwise cover the full derived range. (`gpt-5.5`)

wallet/internal/db/scan_batch_common.go
```diff
@@ -0,0 +1,233 @@
...
+		// Re-check the recovery bound on every candidate: an invalid-child
+		// skip below can advance index to MaxAddressIndex+1, and without
+		// this guard the loop would derive and persist an out-of-range
+		// child past the bound ExtendScanHorizon enforced for horizon.Index.
+		if index > MaxAddressIndex {
+			return 0, fmt.Errorf("extend horizon: %w",
+				ErrMaxAddressIndexReached)
+		}
+
+		derived, err := deriveFn(ctx, AddressDerivationParams{
```
yyforyongyu, line 205: `AddressDerivationParams` now carries `AccountID`, but this horizon path leaves it nil. Store-backed derivation callbacks that need the stable account row identity lose it here even though `HorizonAccount` has `AccountID`; pass the converted account ID into the params before calling `deriveFn`. (`gpt-5.5 xhigh`)

wallet/db_ops.go
yyforyongyu: This fallback eviction range is too short if `ExtendAddresses` skips an invalid child before the post-extension `AccountProperties` read fails. The live branch next index can advance past `maxFoundIndex+1`, leaving cached children above this bound visible after walletdb rolls back. Invalidate the account cache here or evict a range that covers skipped children. (`gpt-5.5 xhigh`)

wallet/db_ops.go
```diff
@@ -758,25 +815,74 @@ func (s *syncer) putAddrHorizons(_ context.Context,
...
+		)
+		if err != nil {
+			// ExtendAddresses mutates live caches before this read. If the
+			// read fails, the caller will not get a rollback entry, so evict
+			// a conservative range immediately. The extension can skip
+			// HD-invalid children and cache valid children past maxFoundIndex.
+			if maxFoundIndex >= fromIndex {
+				scopedMgr.EvictDerivedAddresses(
+					bs.Account, bs.Branch, fromIndex,
+					waddrmgr.MaxAddressesPerAccount+1,
```
yyforyongyu, line 867: Using `MaxAddressesPerAccount+1` here makes rollback cleanup potentially iterate the entire BIP32 non-hardened range. `EvictDerivedAddresses` re-derives every index in `[fromIndex, toIndex)` when account info is cached, so a post-extension `AccountProperties` failure after a small horizon can try roughly 2.1B derivations while unwinding. Invalidate the account cache or avoid an unbounded eviction range. (`gpt-5.5 xhigh`)

yyforyongyu, line 867: This fallback can turn a small rollback into a scan of the entire account range. When the account info is cached, `EvictDerivedAddresses` calls `evictDerivedAddressCacheWithAccount`, which re-derives every index in `[fromIndex, toIndex)` while holding the manager lock; passing `MaxAddressesPerAccount+1` can therefore run billions of derivations on this error path. Track the attempted post-extension end or evict by scanning cached keys instead of using the global max bound. (`gpt-5.5`)

yyforyongyu:

yyforyongyu:
No issues found by gpt-5.5 🦉

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🛰️

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔭

yyforyongyu:

yyforyongyu:
No issues found by gpt-5.5 🦦

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` ✨

yyforyongyu:
No issues found by gpt-5.5 🪐

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🧩

yyforyongyu:

yyforyongyu:
No issues found by gpt-5.5 🧭

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🪶

wallet/internal/db/data_types.go
```diff
@@ -1089,6 +1105,51 @@ type TxBatchParams struct {
...
+// ScanHorizon records the highest recovered address index for one account
+// branch.
+type ScanHorizon struct {
+	// Scope is the key scope containing the branch.
+	Scope KeyScope
+
+	// AccountID is the stable store-local row identity of the horizon's owning
+	// account. SQL backends resolve horizons from this ID because account names
+	// are mutable and account numbers do not identify imported xpub accounts.
+	AccountID *uint32
```
yyforyongyu, line 1120: This account-ID targeting should be folded into the commit that introduces `ScanHorizon`: the current PR first adds the same API using `AccountName` as the durable target and then rewrites it here, so the stack contains an introduce-then-rework sequence for the horizon identity contract. Please build the initial scan-horizon commit directly around `AccountID` instead of introducing the account-name version and correcting it later. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🧪

yyforyongyu:

wallet/internal/sql/pg/queries/accounts.sql
```diff
@@ -406,3 +429,27 @@ WHERE
         OR (acc.is_derived = FALSE AND acc.account_number IS NULL)
     )
 GROUP BY da.account_id;
+
+-- name: AdvanceNextExternalIndex :exec
+-- Advances the external branch's next index to the supplied value during
+-- recovery horizon extension. The GREATEST guard keeps the counter monotonic
+-- so a slower concurrent writer cannot regress it below an already-recorded
+-- index.
```
GustavoStingelin, line 437: 👀

GustavoStingelin (approved):
LGTM 🏗️

PR #1270, opened 2026-06-17
wallet: route chain-scan writes and load scan data through the Store

yyforyongyu:
Routes the wallet's chain-scan flow through the runtime `db.Store` and loads its scan inputs from it.

- Scan writes: build the Store scan-batch params (horizon merging and flattening, per-output tx params, scan-result block shape), and route filtered-block batches, scan-batch writes, targeted scan-batch writes, and in-scan sync-tip updates through the Store.
- Scan-data loading: load scan horizons from Store accounts, plus scan addresses and watch outputs, and wire that loading into the scan path.
- Stamp account names onto scan horizons.

Tested with filtered-block write-shape, bare-multisig, and third-party-filtering cases, scan-horizon loading, and a check that a confirmed relevant tx omits the sync tip.

### DIFF
--- a/wallet/syncer.go
+++ b/wallet/syncer.go
@@ -906,6 +906,28 @@ func (s *syncer) putSyncBatch(ctx context.Context,
 	return nil
 }
 
+// putTargetedBatch records targeted recovery scan results through the store
+// when configured, falling back to the legacy walletdb path otherwise.
+func (s *syncer) putTargetedBatch(ctx context.Context,
+	results []scanResult) error {
+
+	if s.store == nil {
+		return s.DBPutTargetedBatch(ctx, results)
+	}
+
+	params, err := s.storeScanBatchParams(results, false)
+	if err != nil {
+		return err
+	}
+
+	err = s.store.ApplyScanBatch(ctx, params)
+	if err != nil {
+		return fmt.Errorf("apply targeted scan batch: %w", err)
+	}
+
+	return nil
+}
+
 // mergeScanHorizons records the highest discovered child index per branch scope
 // into the running horizon map.
 func mergeScanHorizons(horizons map[waddrmgr.BranchScope]uint32,
@@ -1834,7 +1856,7 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
 		}
 
 		// Process results (update DB).
-		err = s.DBPutTargetedBatch(ctx, results)
+		err = s.putTargetedBatch(ctx, results)
 		if err != nil {
 			return err
 		}
--- a/wallet/syncer.go
+++ b/wallet/syncer.go
@@ -17,6 +17,7 @@ import (
 	"github.com/btcsuite/btcwallet/chain"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
 	"github.com/btcsuite/btcwallet/walletdb"
 	"github.com/btcsuite/btcwallet/wtxmgr"
 )
@@ -90,6 +91,10 @@ const (
 	//    Funds" errors or extremely inaccurate fee estimates increases,
 	//    making the explicit "Syncing" state a necessary safeguard.
 	syncStateSwitchThreshold = 6
+
+	// scanAddressPageLimit is the store page size used when loading recovery
+	// scan addresses.
+	scanAddressPageLimit = 1000
 )
 
 // syncState represents the synchronization status of the wallet with the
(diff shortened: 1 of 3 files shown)


yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 📦

wallet/syncer.go
```diff
@@ -500,6 +598,419 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
...
+	horizons map[waddrmgr.BranchScope]uint32) []db.ScanHorizon {
+
+	params := make([]db.ScanHorizon, 0, len(horizons))
+	for bs, index := range horizons {
+		name, _ := scanState.AccountName(bs.Scope, bs.Account)
+
+		params = append(params, db.ScanHorizon{
+			Scope:       db.KeyScope(bs.Scope),
+			Account:     bs.Account,
+			AccountName: name,
```
yyforyongyu, line 983: These emitted horizons never set `AccountID`, but the current store scan-horizon resolver rejects a nil account ID before deriving addresses. Any Store-backed scan that discovers a lookahead address will build one of these horizons and then fail `ApplyScanBatch` instead of advancing the horizon. Carry the stable account ID through the recovery state and set it here, or restore a backend resolver for the identity this adapter emits. (`gpt-5.5`)

wallet/syncer_test.go
```diff
@@ -828,6 +832,836 @@ func TestHandleChainUpdate(t *testing.T) {
...
+		RelevantTxs: []*wtxmgr.TxRecord{rec},
+	})
+
+	// Assert: The member-owned credit was written through the store batch API
+	// instead of being dropped.
+	require.NoError(t, err)
+	store.AssertExpectations(t)
+}
+
+// matchOwnedCreditOnlyBatch returns a matcher asserting a filtered block
```
yyforyongyu: This comment group is attached to `matchOwnedCreditOnlyTxShape`, but it starts by documenting `matchOwnedCreditOnlyBatch`; the latter helper is then left without its own declaration comment. Split the comments so each helper has an immediately preceding comment that starts with that helper name. (`gpt-5.5`)

wallet/syncer.go
```diff
@@ -722,15 +758,233 @@ func (s *syncer) putTxNotifications(ctx context.Context,
...
+// properties that seed accountNames, so a miss is not expected in practice;
+// emitting a wrong name would be worse than the number fallback.
+func scanHorizonParams(scanState *RecoveryState,
+	horizons map[waddrmgr.BranchScope]uint32) []db.ScanHorizon {
+
+	params := make([]db.ScanHorizon, 0, len(horizons))
+	for bs, index := range horizons {
+		name, _ := scanState.AccountName(bs.Scope, bs.Account)
+
+		params = append(params, db.ScanHorizon{
```
yyforyongyu, line 979: The emitted horizon never sets AccountID, but both SQL GetHorizonAccount implementations now require it and return ErrInvalidParam when nil. Any scan batch with a found horizon will fail before extending or writing results; carry the Store account ID in the recovery state or resolve it before building ScanHorizon. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 📦

wallet/syncer.go
```diff
@@ -749,15 +784,246 @@ func (s *syncer) putTxNotifications(ctx context.Context,
...
```
yyforyongyu, line 979: These horizons still never set `AccountID`, while both SQL `GetHorizonAccount` implementations reject nil `AccountID` before deriving. Any Store-backed scan that discovers a lookahead address will fail `ApplyScanBatch` instead of advancing the horizon. Carry the Store account ID through the recovery state or resolve it before appending the `ScanHorizon`. (`gpt-5.5`)

wallet/syncer.go
```diff
@@ -749,15 +784,246 @@ func (s *syncer) putTxNotifications(ctx context.Context,
...
+func (s *syncer) scanHorizonAccountID(ctx context.Context,
+	horizon db.ScanHorizon) (*uint32, error) {
+
+	query := db.GetAccountQuery{
+		WalletID:    s.walletID,
+		Scope:       horizon.Scope,
+		SkipBalance: true,
+	}
+	if horizon.AccountName != "" {
+		query.Name = &horizon.AccountName
```
yyforyongyu: This lookup uses the mutable account name captured when scan horizons were built. `RenameAccount` is allowed while the syncer is running, so a rename before this resolution makes the scan batch fail even though the account is unchanged. Carry the stable account ID or resolve it when building the horizon under the same snapshot. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🦊

wallet/syncer.go
```diff
@@ -1482,6 +1760,11 @@ func (s *syncer) waitForEvent(ctx context.Context) error {
 // scanWithRewind rewinds the wallet's sync status to the requested start
 // block.
 func (s *syncer) scanWithRewind(ctx context.Context, req *scanReq) error {
+	// Read the synced tip through syncedTo so a Store-backed backend uses
```
yyforyongyu, line 1877: This Store-backed rewind decision still writes only to the legacy managers below. Because `advanceChainSync` also reads the Store tip after this change, the Store remains at the old height after the manual rewind and the requested full rescan can be skipped on the next sync step. Update the Store wallet-scoped sync/tx state too or keep the store-backed manual rescan on one backend. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -740,15 +771,194 @@ func (s *syncer) putTxNotifications(ctx context.Context,
...
+			len(result.RelevantOutputs) > 0 {
+
+			blockNeeded = true
+		}
+
+		var block *db.Block
+		if blockNeeded {
+			var err error
+
+			block, err = storeBlockFromScanResult(result)
```
yyforyongyu, line 1048: This Store conversion persists `result.meta.Time`, but the full-block scan path still builds `BlockMeta` with only hash and height. Store-backed full-block scans therefore record zero block timestamps here, and `appendScanTxParams` uses the same zero time as each transaction received time. Set `Time` from the fetched block header before converting scan results. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -764,6 +974,17 @@ func storeBlockFromBlockMeta(block wtxmgr.BlockMeta) (*db.Block, error) {
...
 
+// storeBlockFromScanResult converts scan result block metadata into the store
+// block shape.
+func storeBlockFromScanResult(result scanResult) (*db.Block, error) {
+	if result.meta == nil {
+		return nil, fmt.Errorf("scan result is missing block metadata: %w",
+			db.ErrInvalidParam)
+	}
+
+	return storeBlockFromBlockMeta(*result.meta)
```
yyforyongyu, line 1098: This trusts `result.meta.Time`, but the full-block scanner builds `BlockMeta` with only hash and height even though the downloaded `wire.MsgBlock` already has the header timestamp. Store-backed full-block scans and auto fallbacks therefore persist zero timestamps for synced blocks and matching transactions. Copy `block.Header.Timestamp` into the full-block scan metadata before converting scan results. (`gpt-5.5`)

yyforyongyu:

yyforyongyu:

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🛤️

yyforyongyu:
No issues found by gpt-5.5 🐋

yyforyongyu:
No issues found by gpt-5.5 🧭

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 📡

yyforyongyu:

yyforyongyu:
No issues found by gpt-5.5 🐞

wallet/syncer_test.go
```diff
@@ -4043,3 +4902,187 @@ func TestAdvanceChainSync_LargeGap(t *testing.T) {
...
+			Height:    50,
+			Hash:      chainhash.Hash{byte(50)},
+			Timestamp: time.Unix(50, 0).UTC(),
+		}
+
+		// Because the Store tip (100) is above the requested start (50),
+		// the manual rewind rolls back tx state and sync metadata atomically
+		// from the first block above the requested start.
+		store.On(
+			"RollbackToBlock", mock.Anything, uint32(start.Height+1),
```
yyforyongyu: This expectation is stale. `scanWithRewind` now calls `rewindWallet`, and store-backed `rewindWallet` invokes `RewindWallet` with `db.RewindWalletParams`, not `RollbackToBlock`; this test will fail with an unexpected `RewindWallet` call instead of protecting the store-backed rewind path. Update the mock to expect `RewindWallet` and verify the wallet ID plus block. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🧱

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` ↩️

yyforyongyu:
No issues found by gpt-5.5 🦜

wallet/syncer.go
```diff
@@ -1394,7 +1674,7 @@ func (s *syncer) scanBatch(ctx context.Context, syncedTo waddrmgr.BlockStamp,
 		return fmt.Errorf("%w: scan batch empty", ErrScanBatchEmpty)
 	}
 	// Process Batch (Update). We do this in a single DB transaction.
-	return s.DBPutSyncBatch(ctx, results)
+	return s.putSyncBatch(ctx, scanState, results)
```
yyforyongyu, line 1731: This moves full-sync scan writes to `ApplyScanBatch`, but this PR still rebuilds each batch with `loadWalletScanData` -> `DBGetScanData`, which reads horizons, active addresses, and UTXOs from the legacy managers. SQL `ApplyScanBatch` persists newly discovered horizons/UTXOs only to the Store, so the next batch in the same sync can miss the expanded lookahead or newly found spends until the later Store read-path PR is applied. Fold the Store scan-data loading into this commit/PR or keep writes mirrored so the intermediate stack remains bisectable. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔍

yyforyongyu:

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔭

yyforyongyu:
No issues found by gpt-5.5 🪶

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔭

yyforyongyu:
No issues found by gpt-5.5 🐌

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔭

yyforyongyu:
No issues found by gpt-5.5 🦇

wallet/syncer.go
```diff
@@ -766,6 +781,35 @@ func (s *syncer) putTxNotifications(ctx context.Context,
...
+}
+
+// putBlockNotifications records filtered block notifications through the store
+// when configured, falling back to the legacy walletdb path otherwise.
+func (s *syncer) putBlockNotifications(ctx context.Context,
+	matches TxEntries, blockMeta *wtxmgr.BlockMeta) error {
+
+	if s.store == nil {
+		return s.DBPutBlocks(ctx, matches, blockMeta)
+	}
```
GustavoStingelin, line 794: It is not clear to me why we should fall back to the legacy path here. Is this not supposed to use only the `Store` interface?

GustavoStingelin (commented):
I think the code looks good overall. I just left one question about the direction.

wallet/syncer.go
yyforyongyu: Correct - so the fallback is purely test-migration scaffolding and #1271 removes both the fallbacks and the nil-store test construction together. It's a poor man's choice to limit the size of the PRs.

GustavoStingelin (approved):
LGTM

PR #1271, opened 2026-06-17
wallet: resolve targeted rescan by identity and make the syncer Store-mandatory

yyforyongyu:
Resolves targeted-rescan accounts by durable identity and makes the runtime `db.Store` the sole source for the syncer.

- Targeted rescan: resolve accounts by durable identity so an imported-xpub target is keyed by `AccountName` rather than a masked account number (both Store backends collapse an imported account's number to 0, making it indistinguishable from derived account 0); the keyless imported-address bucket is skipped before any Store lookup. Adds the `waddrmgr` lookup this needs.
- Drop the syncer's legacy DB fallbacks and make the Store mandatory.
- Route the manager master-fingerprint read through the Store.
- Convert the syncer rollback/lifecycle, scan, and chain-update/broadcast tests to Store-only.

### DIFF
--- a/wallet/syncer.go
+++ b/wallet/syncer.go
@@ -464,10 +464,6 @@ func (s *syncer) checkRollback(ctx context.Context) error {
 func (s *syncer) rewindToBlock(ctx context.Context,
 	block waddrmgr.BlockStamp) error {
 
-	if s.store == nil {
-		return s.DBPutRewind(ctx, block)
-	}
-
 	rollbackBoundary := int64(block.Height) + 1
 
 	rollbackHeight, err := db.Int64ToUint32(rollbackBoundary)
@@ -554,10 +550,6 @@ func (s *syncer) resolveRewindBlock(
 func (s *syncer) syncedBlockHashes(ctx context.Context, startHeight,
 	endHeight int32) ([]*chainhash.Hash, error) {
 
-	if s.store == nil {
-		return s.DBGetSyncedBlocks(ctx, startHeight, endHeight)
-	}
-
 	start, err := db.Int64ToUint32(int64(startHeight))
 	if err != nil {
 		return nil, fmt.Errorf("start height %d: %w", startHeight, err)
@@ -586,21 +578,15 @@ func (s *syncer) syncedBlockHashes(ctx context.Context, startHeight,
 	return hashes, nil
 }
 
-// syncedTo returns the wallet's current synced-to block. In store-backed mode
-// it reads the tip from the Store so callers do not depend on the legacy
-// addrStore tip being kept in lockstep by ApplyScanBatch; otherwise it falls
-// back to the legacy addrStore.
+// syncedTo returns the wallet's current synced-to block from the Store.
 func (s *syncer) syncedTo(ctx context.Context) (waddrmgr.BlockStamp, error) {
-	if s.store == nil {
-		return s.addrStore.SyncedTo(), nil
-	}
-
 	walletInfo, err := s.store.GetWallet(ctx, s.cfg.Name)
 	if err != nil {
 		return waddrmgr.BlockStamp{}, fmt.Errorf("get wallet sync tip: %w",
 			err)
 	}
 
+	// A nil SyncedTo means the wallet has not been synced to any block yet.
 	if walletInfo.SyncedTo == nil {
 		return waddrmgr.BlockStamp{Height: -1}, nil
 	}
@@ -727,10 +713,6 @@ func (s *syncer) broadcastUnminedTxns(ctx context.Context) error {
 // unminedTxns returns transactions that are still active in the wallet's
 // unmined set.
 func (s *syncer) unminedTxns(ctx context.Context) ([]*wire.MsgTx, error) {
-	if s.store == nil {
-		return s.DBGetUnminedTxns(ctx)
-	}
-
 	infos, err := s.store.ListTxns(
 		ctx, db.ListTxnsQuery{
 			WalletID:    s.walletID,
@@ -763,13 +745,9 @@ func (s *syncer) unminedTxns(ctx context.Context) ([]*wire.MsgTx, error) {
 	return wtxmgr.DependencySort(txSet), nil
 }
 
-// updateSyncTip records the latest synced block for store-backed runtime paths.
+// updateSyncTip records the latest synced block through the store.
 func (s *syncer) updateSyncTip(ctx context.Context,
 	block wtxmgr.BlockMeta) error {
-
-	if s.store == nil {
-		return s.DBPutSyncTip(ctx, block)
-	}
 
 	storeBlock, err := storeBlockFromBlockMeta(block)
 	if err != nil {
@@ -790,13 +768,9 @@ func (s *syncer) updateSyncTip(ctx context.Context,
 }
 
 // putTxNotifications records relevant transaction notifications through the
-// store when configured, falling back to the legacy walletdb path otherwise.
+// store.
 func (s *syncer) putTxNotifications(ctx context.Context,
 	matches TxEntries, blockMeta *wtxmgr.BlockMeta) error {
-
-	if s.store == nil {
-		return s.DBPutTxns(ctx, matches, blockMeta)
-	}
 
 	var block *db.Block
 	if blockMeta != nil {
@@ -811,14 +785,9 @@ func (s *syncer) putTxNotifications(ctx context.Context,
 	return s.applyStoreTxBatch(ctx, matches, block, nil)
 }
 
-// putBlockNotifications records filtered block notifications through the store
-// when configured, falling back to the legacy walletdb path otherwise.
+// putBlockNotifications records filtered block notifications through the store.
 func (s *syncer) putBlockNotifications(ctx context.Context,
 	matches TxEntries, blockMeta *wtxmgr.BlockMeta) error {
-
-	if s.store == nil {
-		return s.DBPutBlocks(ctx, matches, blockMeta)
-	}
 
 	if blockMeta == nil {
 		return fmt.Errorf("filtered block is missing metadata: %w",
@@ -922,13 +891,9 @@ func (s *syncer) txNotificationState(ctx context.Context,
 }
 
 // putSyncBatch records recovery scan results and synced blocks through the
-// store when configured, falling back to the legacy walletdb path otherwise.
+// store.
 func (s *syncer) putSyncBatch(ctx context.Context, scanState *RecoveryState,
 	results []scanResult) error {
-
-	if s.store == nil {
-		return s.DBPutSyncBatch(ctx, results)
-	}
 
 	params, err := s.storeScanBatchParams(scanState, results, true)
 	if err != nil {
@@ -948,14 +913,9 @@ func (s *syncer) putSyncBatch(ctx context.Context, scanState *RecoveryState,
 	return nil
 }
 
-// putTargetedBatch records targeted recovery scan results through the store
-// when configured, falling back to the legacy walletdb path otherwise.
+// putTargetedBatch records targeted recovery scan results through the store.
 func (s *syncer) putTargetedBatch(ctx context.Context,
 	scanState *RecoveryState, results []scanResult) error {
-
-	if s.store == nil {
-		return s.DBPutTargetedBatch(ctx, results)
-	}
 
 	params, err := s.storeScanBatchParams(scanState, results, false)
 	if err != nil {
@@ -1210,10 +1170,6 @@ func (s *syncer) loadFullScanState(
 func (s *syncer) stampRecoveryAccountIDs(ctx context.Context,
 	scanState *RecoveryState,
 	accounts []*waddrmgr.AccountProperties) error {
-
-	if s.store == nil {
-		return nil
-	}
 
 	for _, props := range accounts {
 		accountID, err := s.accountPropertiesAccountID(ctx, props)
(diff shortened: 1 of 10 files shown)


yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔒

wallet/syncer.go
```diff
@@ -1853,6 +1835,328 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+// This reproduces the legacy ForEachRelevantActiveAddress filtering that
+// DBGetScanData relies on: for default key scopes every active address is
+// watched, while for non-default key scopes only internal-branch (change)
+// addresses are watched. The non-default external branches are intentionally
+// skipped because they only ever existed due to a since-fixed bug, and
+// watching them would diverge from the legacy recovery set.
+func (s *syncer) storeScanAddresses(
+	ctx context.Context) ([]btcutil.Address, error) {
+
+	accounts, err := s.store.ListAccounts(ctx, db.ListAccountsQuery{
```
yyforyongyu, line 2166: Listing only materialized accounts omits the reserved imported-address alias. Both SQL backends expose raw imports only when ListAddresses is called with AccountName == "imported", and kvdb hides that pseudo-account from unfiltered ListAccounts, so raw imported addresses are never seeded into the scan set. Explicitly page the imported alias per scope or add a Store API for raw imports. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -1853,6 +1835,328 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+// account's addresses instead of its own.
+//
+// Withholding imported accounts loses no coverage: their addresses stay watched
+// through the active-address and UTXO-watch lists, and only the (currently
+// unsafe) HD recovery lookahead is skipped. The ScanHorizon.AccountName
+// identity is already plumbed through, so imported-xpub horizons can be safely
+// enabled once RecoveryState distinguishes accounts by that non-masked
+// identity rather than by the masked account number.
+func withholdFromRecoveryHorizons(info db.AccountInfo) bool {
+	return info.IsImported
```
yyforyongyu: This also withholds true imported xpub accounts, not just the keyless imported-address bucket. Existing active addresses/UTXOs do not cover lookahead addresses beyond the stored counters, so Store-backed recovery/targeted rescan stops discovering future imported-xpub receives. Keep the keyless bucket out, but preserve an imported-xpub horizon using its non-masked identity/derivation data. (`gpt-5.5 xhigh`)

wallet/syncer.go
```diff
@@ -1853,6 +1835,328 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+		props := make([]*waddrmgr.AccountProperties, 0, len(accounts))
+		for i := range accounts {
+			// Withhold every imported account from the recovery
+			// horizons; see withholdFromRecoveryHorizons for why an
+			// imported account cannot safely seed the recovery state.
+			if withholdFromRecoveryHorizons(accounts[i]) {
+				continue
+			}
+
+			props = append(props, storeAccountProperties(accounts[i]))
```
yyforyongyu: This conversion drops the Store account row ID before the recovery state can emit horizons. The SQL horizon resolver now rejects every horizon with nil `AccountID`, so any lookahead hit makes `ApplyScanBatch` fail instead of extending the branch. Carry the Store account ID through the recovery state and set it on each emitted `ScanHorizon`, or restore a resolver for the identity this path emits. (`gpt-5.5`)

wallet/syncer.go
yyforyongyu: This now withholds true imported-xpub accounts from both full and targeted Store scan horizons. The active-address and UTXO lists only cover already-persisted children, so the recovery window no longer derives unused imported-xpub children and rescans can miss payments to them. Keep skipping only the keyless imported-address bucket, or carry a durable imported-account identity through `RecoveryState` so imported-xpub lookahead can be emitted safely. (`gpt-5.5`)

wallet/syncer.go
```diff
@@ -210,34 +243,22 @@ type syncer struct {
...
-}
-
-// newSyncer creates a new syncer instance.
+// newSyncer creates a new syncer instance. The Store and its wallet ID are
+// mandatory: every migrated runtime path reads and writes through the Store,
+// so there is no nil-store fallback.
 func newSyncer(cfg Config, addrStore waddrmgr.AddrStore,
-	txStore wtxmgr.TxStore, publisher TxPublisher,
-	storeConfigs ...syncerStoreConfig) *syncer {
+	txStore wtxmgr.TxStore, publisher TxPublisher, store db.Store,
```
yyforyongyu, line 257: After making these parameters mandatory, one caller was left on the old four-argument form (`TestLoadWalletScanDataKeepsImportedXpubHorizons`), so the wallet test package no longer compiles. Update that remaining call to pass a mock Store and wallet ID like the neighboring tests. (`gpt-5.5`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔒

wallet/syncer.go
```diff
@@ -1913,6 +1901,445 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+			ctx, accounts[i],
+		)
+		if err != nil {
+			return nil, err
+		}
+
+		addrs = append(addrs, accountAddrs...)
+	}
+
+	for _, scope := range storeAccountScopes(accounts) {
```
yyforyongyu: This only scans the imported alias for scopes already returned by `ListAccounts`. SQL raw imports are stored under the reserved alias with no account row, and `CreateWallet` does not create account rows by itself, so a shell/watch-only wallet that only imports a pubkey or script leaves `accounts` empty and never seeds that import into the recovery watch set. Build the scope set from raw imports too, or explicitly probe the default import scopes independent of `ListAccounts`. (`gpt-5.5`)

yyforyongyu: Raw imported addresses can be the only data in a scope: `CreateImportedAddress` creates an address under the scope without an account row, so `ListAccounts` returns no account for that scope. Deriving imported-address scopes only from `accounts` skips those imported-only scopes and the rescan stops watching their addresses. Include scopes from raw imported addresses or key scopes as well. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by gpt-5.5 🐋

wallet/syncer.go
```diff
@@ -1853,6 +1841,457 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+			return nil, err
+		}
+
+		addrs = append(addrs, accountAddrs...)
+	}
+
+	for _, scope := range storeScanAddressScopes(accounts) {
+		importedAddrs, err := s.storeImportedScanAddresses(ctx, scope)
+		if err != nil {
+			if waddrmgr.IsError(err, waddrmgr.ErrScopeNotFound) {
```
yyforyongyu: This check will not catch the missing-scope case from `storeImportedScanAddresses`: `storeAccountScanAddresses` wraps the `ListAddresses` error, and kvdb `ListAddresses` wraps the scoped-manager error before that. `waddrmgr.IsError` only type-asserts the top-level error, so wrapped `ErrScopeNotFound` falls through and aborts scans on wallets without that scope. Use an `errors.As`-based check for `waddrmgr.ManagerError`. (`gpt-5.5 xhigh`)

yyforyongyu:

yyforyongyu:
No issues found by gpt-5.5 🧭

wallet/syncer.go
```diff
@@ -210,34 +250,22 @@ type syncer struct {
...
-	store db.Store
-
-	// walletID is the database wallet identifier used by store-backed paths.
-	walletID uint32
-}
-
-// newSyncer creates a new syncer instance.
+// newSyncer creates a new syncer instance. The Store and its wallet ID are
+// mandatory: every migrated runtime path reads and writes through the Store,
+// so there is no nil-store fallback.
```
yyforyongyu, line 255: This godoc states the Store is mandatory with "no nil-store fallback", but two `s.store == nil` branches still remain in this file — `scanWithRewind` (~:1769, a dead legacy `rewindToBlock` dispatch) and `stampRecoveryAccountIDs` (~:1060) — which the "drop syncer legacy DB fallbacks" commit missed; since `newSyncer`'s store param is now mandatory and every caller passes a non-nil store, these are unreachable dead legacy branches — drop them (so scanWithRewind's `else` becomes unconditional) to match this godoc.

(`claude-opus-4-8[1m]`)

yyforyongyu:

wallet/syncer_test.go
```diff
@@ -1787,6 +1793,900 @@ func TestScanHorizonParamsStampsAccountIdentity(t *testing.T) {
...
+	require.Len(t, addrs, 1)
+	require.Equal(t, addr.EncodeAddress(), addrs[0].EncodeAddress())
+	require.Len(t, unspent, 1)
+	require.Equal(t, int32(-1), unspent[0].Height)
+	store.AssertExpectations(t)
+}
+
+// TestLoadScanDataLegacyFallback verifies scan-data loading still uses the
+// legacy database path when the transitional store is not configured.
+func TestLoadScanDataLegacyFallback(t *testing.T) {
```
yyforyongyu: `TestLoadScanDataLegacyFallback` is added here to cover the nil-store legacy path, but `99ac41919` deletes it three commits later (the fallback it covers is removed in this same PR) -- drop this commit rather than introducing a test only to remove it within the series.

(`claude-opus-4-8[1m]`)

wallet/syncer.go
```diff
@@ -424,10 +452,6 @@ func (s *syncer) checkRollback(ctx context.Context) error {
 func (s *syncer) rewindToBlock(ctx context.Context,
 	block waddrmgr.BlockStamp) error {
 
-	if s.store == nil {
```
yyforyongyu, line 440: The store-mandatory cleanup drops these `s.store == nil` fallbacks but misses the same now-dead guard in `stampRecoveryAccountIDs` (head ~L1060) and `scanWithRewind` (head ~L1769) -- the store is mandatory after `20e312b2c`, so both branches are unreachable. Remove them too (collapse `scanWithRewind` to the unconditional `updateSyncTip`).

(`claude-opus-4-8[1m]`)

yyforyongyu:
No issues found by gpt-5.5 🦉

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🎯

yyforyongyu:

yyforyongyu:
No issues found by gpt-5.5 🐬

yyforyongyu:

wallet/syncer.go
yyforyongyu, line 440: The store-mandatory cleanup misses one `s.store == nil` fallback: the new `rewindWallet` helper (head ~L480) returns `s.DBPutRewind` when the store is nil, but the store is mandatory after `a700558f5`, so that branch is unreachable. Drop it and call `s.store.RewindWallet` unconditionally.

(`claude-opus-4-8[1m]`)

wallet/syncer.go
yyforyongyu, line 255: This commit makes the Store mandatory and documents that there is no nil-store fallback, but the same PR still leaves the `rewindWallet` `s.store == nil` branch below. That creates a cleanup commit that only partially enforces its invariant; fold the remaining fallback removal into the Store-mandatory cleanup so the commit does not introduce an invariant and then leave an immediate exception behind. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🟢

yyforyongyu:
No issues found by gpt-5.5 🦭

wallet/syncer.go
```diff
@@ -1910,6 +1890,465 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
+
+		seen[scope] = struct{}{}
+		scopes = append(scopes, scope)
+	}
+
+	for i := range accounts {
+		addScope(accounts[i].KeyScope)
+	}
+
+	for _, scope := range waddrmgr.DefaultKeyScopes {
```
yyforyongyu: Raw imported addresses can be created in arbitrary key scopes and SQL deliberately does not materialize an account row for the imported alias. This scope list only includes scopes from `ListAccounts` plus the default scopes, so a raw import in a non-default scope with no account row is never queried by `storeImportedScanAddresses` and full scan initialization will omit that address. Discover raw-import scopes from address/key-scope data as well, or otherwise include all scopes that contain raw imports. (`gpt-5.5 xhigh`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🟩

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🟢

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🟢

yyforyongyu:
No issues found by gpt-5.5 🐧

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🟢

yyforyongyu:
No issues found by gpt-5.5 🦑

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🟢

yyforyongyu:
No issues found by gpt-5.5 🐚

GustavoStingelin (approved):
LGTM

PR #1273, opened 2026-06-18
wallet: impl keyvault lifecycle

GustavoStingelin:
## Change Description
This PR implements the `keyvault.Vault` interface in the `DBVault` struct backed by `db.Store`.
Still uses the old `salsa20poly1305` implemented by the `snacl` package.

## Steps to Test
```
make unit pkg=wallet/internal/keyvault/...
make unit-race pkg=wallet/internal/keyvault/...
```


wallet/internal/keyvault/vault_unlock.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return fmt.Errorf("wallet %d vault Unlock: get secrets: %w",
+			v.walletID, err)
+	}
+
+	state, err := decryptWalletSecrets(secrets, passphrase)
+	if err != nil {
+		return fmt.Errorf("wallet %d vault Unlock: decrypt secrets: %w",
+			v.walletID, err)
+	}
```
GustavoStingelin: Should we explicitly lock the vault when an incorrect passphrase is provided?

wallet/internal/keyvault/vault_cipher.go
```diff
@@ -0,0 +1,56 @@
...
+	keyType waddrmgr.CryptoKeyType) (*snacl.CryptoKey, error) {
+
+	switch keyType {
+	case waddrmgr.CKTPrivate:
+		return &v.unlockedState.cryptoKeyPrivate, nil
+	case waddrmgr.CKTScript:
+		return &v.unlockedState.cryptoKeyScript, nil
+	case waddrmgr.CKTPublic:
+		return nil, fmt.Errorf("public crypto key: %w",
+			errUnsupportedCryptoKeyType)
```
GustavoStingelin, line 83: We will need a compatibility layer to decrypt this type of data from the legacy `kvdb` backend

wallet/internal/keyvault/db_vault_unlock.go
```diff
@@ -0,0 +1,162 @@
...
+	"time"
+
+	"github.com/btcsuite/btcd/btcutil/hdkeychain"
+	"github.com/btcsuite/btcwallet/snacl"
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
+)
+
+// Unlock loads wallet secrets from the store and decrypts them into runtime
+// state using the provided private passphrase.
+func (v *DBVault) Unlock(ctx context.Context, passphrase []byte,
```
yyforyongyu: `e435016b2 impl keyvault Unlock` (~271 weighted added lines: 142 prod + 258 test) and `ff4ba53e6 impl keyvault RefreshPrivatePassphrase` (~221: 151 + 140) exceed the ~200 per-commit cap by bundling a large impl with its full test set; split each methodʼs tests into their own commit — as already done for the encryption vectors in `2c9cd5c8a` — to stay under the cap.

(`claude-opus-4-8[1m]`)

wallet/internal/keyvault/vault_cipher.go
```diff
@@ -0,0 +1,56 @@
...
+var errUnsupportedCryptoKeyType = errors.New("unsupported crypto key type")
+
+// Encrypt encrypts plaintext with the selected unlocked runtime crypto key.
+func (v *DBVault) Encrypt(keyType waddrmgr.CryptoKeyType,
+	plaintext []byte) ([]byte, error) {
+
+	v.mtx.Lock()
+	defer v.mtx.Unlock()
+
+	if v.unlockedState == nil {
```
yyforyongyu, line 22: This guard makes `Encrypt(CKTPublic, ...)` fail with `ErrVaultLocked` before key selection, but `ImportTaprootScript` calls that path to encrypt the encoded tapscript. With the concrete vault, taproot script imports cannot succeed unless the vault has private runtime state, and even then the public key type is rejected below. Handle the public encryptor before the locked check or route the tapscript caller through a public/legacy encryptor. (`gpt-5.5`)

wallet/internal/keyvault/auto_lock_timer.go
```diff
@@ -0,0 +1,61 @@
...
+// callbacks by generation.
+type autoLockTimer struct {
+	// timer holds the active auto lock timer, if one is currently scheduled.
+	// It is stopped and cleared whenever the schedule is canceled or replaced.
+	timer *time.Timer
+
+	// generation identifies the currently valid timer schedule. Each cancel or
+	// reschedule increments it, so callbacks from older timers can detect that
+	// they are stale and return without modifying vault state.
+	generation uint64
```
yyforyongyu: think we can use `atomic`? and this can be `count`? `generation` sounds weird..

yyforyongyu: or use `started` atomic bool? if the intention is to keep the timer only be started once

wallet/internal/keyvault/auto_lock_timer.go
```diff
@@ -0,0 +1,61 @@
...
+func (t *autoLockTimer) schedule(timeout time.Duration, mtx sync.Locker,
+	lock func()) {
+
+	// Snapshot a unique generation for this schedule so the callback can tell
+	// whether it is still current.
+	t.generation++
+	generation := t.generation
+
+	t.timer = time.AfterFunc(timeout, func() {
+		mtx.Lock()
```
yyforyongyu: i would not do this, as this can be a hidden source of deadlock - passing lock is weird, and we don't know what's inside `lock`, so this method cannot guarantee no deadlock. To be go idiomatic, we can use `chan` plus a central control loop

yyforyongyu: That being said, there is already this loop in `mainLoop` on the `Wallet`,
```go
// The auto-lock timer has expired. We trigger a lock with a
		// dummy response channel to avoid nil checks in the handler.
		case <-w.lockTimer.C:
			log.Infof("Auto-lock timeout fired, locking wallet")
			w.handleLockReq(newLockReq())
```

so what is the intended use of this `autoLockTimer`?

wallet/internal/keyvault/db_vault_unlock.go
yyforyongyu: ok can ignore it

wallet/internal/keyvault/vault_unlock.go
yyforyongyu: the vault should already be locked at this stage?

yyforyongyu (changes requested):
Cool very clean commits!

I think the name `DBVault` implies this is a static db-related struct, while I think we wanna build a vault as a service? The intended architecture is,
- wallet level: the user facing front, provides all public APIs
- keyvault: the middle layer, handles locking/unlocking/secret rotation
- db: the raw layer, provides CRUD

The current keyvault impl seems to be missing that - it's not a pure vault for DB, as it handles timers, and it has an implicit lifecycle there, based on the mutex flow. I think what we want here is a full sub-service that takes over `mainLoop` (which atm is just a locking service), and handles auto locking - otherwise, we should leave the timer to be handled by the main loop, and this vault should provide only encrypt/decrypt/rotate functionalities.

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: I don’t quite get it. In the current implementation, the vault may already be either unlocked or locked. If this operation fails, the previous state is preserved.

wallet/internal/keyvault/vault_cipher.go
GustavoStingelin: related to:
> https://github.com/btcsuite/btcwallet/pull/1273#discussion_r3440135040

It's not intended to call this vault with the `CKTPublic`, these callers should be refactored later.

wallet/internal/keyvault/auto_lock_timer.go
GustavoStingelin: We don’t need an atomic here since we already hold the mutex.

I also don’t think a bool works well for this case, because it does not represent whether the callback has started. What we need is a generation or version counter for the lock callback that was created.

The issue is that the callback may be triggered, but then block while waiting for the mutex. Once the mutex is released, that stale callback could still clean the state after a later unlock has already extended the timeout window. The generation check prevents an old callback from clearing the newer state.

wallet/internal/keyvault/auto_lock_timer.go
GustavoStingelin: Look, this is an internal struct and method inside the `keyvault` package, so it is supposed to be called only by the vault functions. The scope here is very small, and the `lock func()` being passed in mostly handles memory zeroing. We can make that clearer in the doc comment.

I did it this way because it is dead simple to reason about. We do not need extra machinery like the channel-based approach, since this is just guarded by a simple mutex.

The lock lives here because it follows the previous ADR statement that the vault should own automatic locking:

> Own lock state and key lifecycle. Keyvault manages unlock state, key material lifetime, auto lock behavior, and secure memory zeroing.

So the current wallet implementation should be removed later.

Otherwise, we can decide to remove that responsibility from `keyvault` and keep it in the wallet, simplifying this PR. That also works, but it increases the current wallet bloat.

GustavoStingelin:
@yyforyongyu I responded in the threads.

What name do you think works better for `keyvault.DBVault`? Maybe `keyvault.WalletVault` or `keyvault.StoreVault`? We already have the `keyvault.Vault` interface, so I think we should avoid making the concrete implementation name too generic.

Also, my next PR will expand the interface to include signing functions. The goal is for the vault to know as little as possible about transactions, while still keeping private keys inside the vault. The design is not done yet, but it will be close to this:

```go
	DerivePubKey(ctx context.Context, ref KeyLocator) (*btcec.PublicKey, error)

	SignECDSA(ctx context.Context, ref KeyLocator,
		digest [32]byte) (*ecdsa.Signature, error)

	SignSchnorr(ctx context.Context, ref KeyLocator,
		digest [32]byte) (*schnorr.Signature, error)

	SignTweakedSchnorr(ctx context.Context, ref KeyLocator,
		digest [32]byte, tweak SchnorrKeyTweak) (*schnorr.Signature, error)
```

So I think we should also keep that in mind when naming it. The vault will not be only a decrypt and encrypt layer.

What are your thoughts?

wallet/internal/keyvault/auto_lock_timer.go
yyforyongyu: > So the current wallet implementation should be removed later.
> Otherwise, we can decide to remove that responsibility from keyvault and keep it in the wallet, simplifying this PR. That also works, but it increases the current wallet bloat.

I think the ADR already says it owns the lifecycle, auto locking behavior, so why not using channels? I don't think there is such a thing as "simple mutex", especially when it's wrapped inside a callback - how do we know the relationship of the callback `lock` and the `mtx` here? they are blackbox, and what if the `lock` also uses `mtx`? I think the Go way is to always use channels first, and avoid using mutexes as much as possible to avoid deadlocks.

wallet/internal/keyvault/auto_lock_timer.go
yyforyongyu: > The issue is that the callback may be triggered, but then block while waiting for the mutex. Once the mutex is released, that stale callback could still clean the state after a later unlock has already extended the timeout window.

So we are queuing requests like `unlock` here? Maybe it's better to explain with a use case I guess

yyforyongyu:
> So I think we should also keep that in mind when naming it. The vault will not be only a decrypt and encrypt layer.

Yeah right, i mean all the secret stuff - if we wanna keep it simple, we can remove the auto locker and let the lifecycle of locking/unlocking be handled by wallet. Or we can expand this to be a component, sort like other components, `Controller`, `Syncer` etc, tho we haven't peeled them off yet. 

> What name do you think works better for keyvault.DBVault? Maybe keyvault.WalletVault or keyvault.StoreVault

Naming indeed can be challenging - I think it depends on what we wanna make keyvault do here - if it's pure encrypt/decrypt/signing, `DBVault` can work(or `StoreVault` to keep the concepts aligned) as it's now just a thin layer above the DB. If we wanna expand it, can rename it to `keyvault.Service`? Also don't have good options i guess.

As for the direction - it's up to you to decide, think there are some implementation details from your local branch😺

GustavoStingelin:
I will create a new commit using the channel approach to try it out, but it looks like it will force us to serialize the requests. With the mutex approach, we can still allow parallelism by using `sync.RWMutex`.

wallet/internal/keyvault/auto_lock_timer.go
GustavoStingelin: The point is that requests should not compete over the vault state. Any operation that observes or changes the vault state should do that under the same mutex, otherwise one goroutine may read one state while another goroutine changes it in the meantime.

A concrete case is:
1. Unlock schedules a timer callback.
2. The timer fires, but the callback blocks waiting for the mutex (bc other op is already doing sth).
3. Before that callback gets the mutex, another unlock refreshes the vault state and extends the timeout window.
4. Once the old callback finally gets the mutex, it could still change the newer state unless we can tell that it belongs to an older timer.

yyforyongyu: It depends on the direction of the `DBVault`. If we wanna keep it thin, we should remove the auto locker here, as it would be conflicting with `mainLoop`. If we wanna extend this to be a full service, that's a different story.

yyforyongyu:
> I will create a new commit using the channel approach to try it out, but it looks like it will force us to serialize the requests. With the mutex approach, we can still allow parallelism by using `sync.RWMutex`.

Ok let's keep it simple - keep `DBVault` thin and small, leave the auto locking and lifecycle management to be handled by `Controller`, and use the mutex to guard simple operations. This means `timeout` should not be part of `keyvault.Vault.Unlock`. The timeout belongs to `Wallet.UnlockRequest/controller` policy, not the passive vault.

GustavoStingelin:
>Ok let's keep it simple - keep DBVault thin and small, leave the auto locking and lifecycle management to be handled by Controller, and use the mutex to guard simple operations. This means timeout should not be part of keyvault.Vault.Unlock. The timeout belongs to Wallet.UnlockRequest/controller policy, not the passive vault.

cool, agreed, pushed this one just as a future reference https://github.com/GustavoStingelin/btcwallet/commit/b9670fb23b18147eb86d1ef5508d0a0f1d217b63#diff-35d82ef33e113def7ecb59fc4ee94faf61459ce701d755589b9673f9794fa0cc

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: changed to no-op when it's unlocked

wallet/internal/keyvault/auto_lock_timer.go
GustavoStingelin: changed to not own autolock

GustavoStingelin:
@yyforyongyu ready for review

wallet/internal/keyvault/vault_unlock.go
```diff
@@ -0,0 +1,163 @@
...
+
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
+	"github.com/btcsuite/btcwallet/snacl"
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
+)
+
+// Unlock loads wallet secrets from the store and decrypts them into runtime
+// state using the provided private passphrase. If the vault is already
+// unlocked, Unlock is a no-op and does not validate the passphrase.
+func (v *WalletVault) Unlock(ctx context.Context, passphrase []byte) error {
```
yyforyongyu: This unlock path is never wired into the wallet lifecycle: `handleUnlockReq` still only calls `DBUnlock`, and `handleLockReq` only locks `addrStore`. The concrete vault installed on `Wallet` therefore keeps `unlockedState == nil`, so existing wallet paths such as account creation that call `w.keyVault.Decrypt(CKTPrivate, ...)` will fail with `ErrVaultLocked` even after a successful wallet unlock. Please either unlock/lock the vault from the controller lifecycle with a Store that can load secrets, or keep the concrete vault out of `Wallet` until that wiring exists. (`gpt-5.5`)

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🔐

wallet/internal/keyvault/keyvault.go
```diff
@@ -37,10 +41,8 @@ type Vault interface {
...
-	// RefreshPrivatePassphrase refreshes vault owned runtime passphrase and
-	// crypto state after a successful private passphrase rotation.
-	//
-	// The vault must still be unlocked with the new passphrase when this method
-	// is called.
-	RefreshPrivatePassphrase(passphrase []byte) error
+	// RefreshPrivatePassphrase rotates persisted wallet secrets to the provided
+	// new private passphrase. The vault must already be unlocked when this
+	// method is called.
+	RefreshPrivatePassphrase(ctx context.Context, passphrase []byte) error
```
yyforyongyu: Q: so this is now basically `ChangePassphrase`?

wallet/internal/keyvault/vault_cipher.go
yyforyongyu: for `kvdb` we will just return an error if it's not supported atm.

wallet/internal/keyvault/vault_islocked.go
```diff
@@ -0,0 +1,9 @@
+package keyvault
+
+// IsLocked reports whether the vault currently has unlocked runtime state.
+func (v *WalletVault) IsLocked() bool {
```
yyforyongyu: I notice that the filenames are mimicking the db store filenames, is this intentional? We previously did this on the db store files because a single file can be too large, so we applied the rules only on the store filenames. It's also nice there because each db ops is very independent. Here I would say it's better to move them into the single `vault.go` as they are closely related.

wallet/internal/keyvault/vault_lock_test.go
```diff
@@ -0,0 +1,167 @@
...
+	require.True(t, vault.IsLocked())
+	require.Nil(t, vault.unlockedState)
+
+	vault.Lock()
+	require.True(t, vault.IsLocked())
+	require.Nil(t, vault.unlockedState)
+}
+
+// TestDBVaultLockWaitsForInFlightUnlock verifies that explicit Lock calls are
+// ordered after an Unlock that has already entered the vault lifecycle.
```
yyforyongyu: I am not following this test - are we testing the mutex behavior here? seems unnecessary?

wallet/internal/keyvault/vault_refreshpassphrase.go
```diff
@@ -0,0 +1,175 @@
...
+			"get secrets: %w", v.walletID, err)
+	}
+
+	updateParams, err := v.makeRotatedWalletSecrets(secrets, passphrase)
+	if err != nil {
+		return fmt.Errorf("wallet %d vault RefreshPrivatePassphrase: "+
+			"rotate secrets: %w", v.walletID, err)
+	}
+
+	// validate that the rotated secrets derive the same runtime keys before
```
yyforyongyu: nit: looks like all the inline comments are not using title case - weird that the agent could also make typos.

wallet/internal/keyvault/vault_refreshpassphrase.go
```diff
@@ -0,0 +1,175 @@
...
+	"context"
+	"fmt"
+
+	"github.com/btcsuite/btcwallet/snacl"
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
+)
+
+// RefreshPrivatePassphrase rotates persisted wallet secrets to the new private
+// passphrase and keeps the existing unlocked runtime state unchanged.
+func (v *WalletVault) RefreshPrivatePassphrase(ctx context.Context,
```
yyforyongyu: If WalletVault is meant to stay dead simple/passive, RefreshPrivatePassphrase should probably not be on the Vault interface. It makes the vault own a higher-level workflow:
- load persisted secrets
- derive a new master private key from the new passphrase
- re-encrypt wallet secrets
- validate them
- write them back to the store

That feels like controller/store orchestration, not a thin vault primitive.

A cleaner split would be:
- Controller handles ChangePassphraseRequest.
- Controller validates policy/state and old passphrase.
- Store transaction updates persisted wallet secret blobs.
- Vault is only told to Lock or maybe re-Unlock after the change if needed.

wallet/internal/keyvault/vault_unlock.go
yyforyongyu: Can ignore given that this is an intermediate PR.

wallet/internal/keyvault/vault_unlock.go
```diff
@@ -0,0 +1,163 @@
...
+	v.mtx.Lock()
+	defer v.mtx.Unlock()
+
+	if v.unlockedState != nil {
+		return nil
```
yyforyongyu: Think this case should be an error?

wallet/internal/keyvault/vault_unlock.go
```diff
@@ -0,0 +1,163 @@
...
+	err = deriveMasterPrivateKey(&masterPrivateKey, passphrase)
+	if err != nil {
+		return nil, err
+	}
+
+	// with the master key, we can start decrypting the other encrypted keys,
+	// including the master HD private key that can spend coins.
+	state := &unlockedState{}
+
+	clearState := true
```
yyforyongyu: This deserves some inline comments to explain why we wanna clear the state, like defensive programming and GC can take time etc.

wallet/internal/keyvault/vault_unlock.go
```diff
@@ -0,0 +1,163 @@
...
+	defer func() {
+		if clearState {
```
yyforyongyu: looking at this again, think there is a cleaner and simpler version, as we don't need to trace the `clearState` to understand the flow,
```go
cryptoKeyPrivate, err := decryptCryptoKey(
	&masterPrivateKey, secrets.EncryptedCryptoPrivKey,
)
if err != nil {
	return nil, fmt.Errorf("crypto key private: %w", err)
}
defer cryptoKeyPrivate.Zero()

cryptoKeyScript, err := decryptCryptoKey(
	&masterPrivateKey, secrets.EncryptedCryptoScriptKey,
)
if err != nil {
	return nil, fmt.Errorf("crypto key script: %w", err)
}
defer cryptoKeyScript.Zero()

// ... decrypt/parse hdRootKey ...

state := &unlockedState{
	cryptoKeyPrivate: cryptoKeyPrivate,
	cryptoKeyScript:  cryptoKeyScript,
	hdRootKey:        hdRootKey,
}

// Transfer ownership to state; don't zero locals after this point.
```

yyforyongyu (changes requested):
Cool very clean now! Left a few comments about the rotation, filenames, typos, and simplification of the zeroing logic.

wallet/internal/keyvault/vault.go
```diff
@@ -46,12 +46,12 @@ type unlockedState struct {
...
 
-// Ensure DBVault implements keyvault.Vault.
-var _ Vault = (*DBVault)(nil)
+// Ensure WalletVault implements keyvault.Vault.
+var _ Vault = (*WalletVault)(nil)
 
 // NewDBVault creates a key-vault bridge scoped to one wallet row.
-func NewDBVault(store db.Store, walletID uint32) *DBVault {
-	return &DBVault{
+func NewDBVault(store db.Store, walletID uint32) *WalletVault {
```
Abdulkbk: should we also update the constructor's name and callsites to `NewWalletVault`?

wallet/internal/keyvault/vault_unlock.go
Abdulkbk: It was documented in struct def as:

> If Unlock is called while the vault is already unlocked, it must be a no-op and must not validate the provided passphrase.

wallet/internal/keyvault/vault_unlock.go
```diff
@@ -0,0 +1,163 @@
...
+			v.walletID, err)
+	}
+
+	state, err := decryptWalletSecrets(secrets, passphrase)
+	if err != nil {
+		return fmt.Errorf("wallet %d vault Unlock: decrypt secrets: %w",
+			v.walletID, err)
+	}
+
+	// After a successful unlockedState construction, clear any existing runtime
```
Abdulkbk: nit: update comment to reflect that `v.clearRuntimeAndLock()` is defensive here (or drop) since we already passed through a nil check above (and we held the mtx lock):
```go
if v.unlockedState != nil {
		return nil
}
```

`v.clearRuntimeAndLock()` is no-op.
```go
func (v *WalletVault) clearRuntimeAndLock() {
	if v.unlockedState != nil {
		v.unlockedState.zero()
		v.unlockedState = nil
	}
}
```

wallet/internal/keyvault/vault_refreshpassphrase.go
Abdulkbk: I agree. Additionally, because this method holds `v.mtx.Lock()`, the same lock used by `Encrypt` and `Decrypt`, a slow `RefreshPrivatePassphrase` will block every `Encrypt` and `Decrypt`.

The same can be said for `Unlock` (i.e., it holds the same lock and performs DB ops), but `Unlock` is typically called once per session.

wallet/internal/keyvault/vault_refreshpassphrase.go
```diff
@@ -0,0 +1,175 @@
...
+	)
+	if err != nil {
+		return db.UpdateWalletSecretsParams{},
+			fmt.Errorf("encrypt crypto key script: %w", err)
+	}
+
+	var encryptedHDRootKey []byte
+	if v.unlockedState.hdRootKey != nil {
+		encryptedHDRootKey, err = v.unlockedState.cryptoKeyPrivate.Encrypt(
+			[]byte(v.unlockedState.hdRootKey.String()),
```
Abdulkbk: Might as well add a todo here because this seems to leak priv material too. Like you did for `decryptWalletSecrets`:
> // TODO(gus): wrap with secret.Do from golang 1.26+....

Abdulkbk (commented):
Nice work and easy to review!

Left some feedback.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: Yes, but I can easy change to `ChangePassphrase` if u think it is clearer

wallet/internal/keyvault/vault_islocked.go
GustavoStingelin: The split was intentional. I wanted to avoid creating a very large file and make it easier to separate private functions by responsibility. That said, I can join everything into a single file if you still prefer.

wallet/internal/keyvault/vault_lock_test.go
GustavoStingelin: Yes, since we have removed autolock, this test no longer makes much sense.

wallet/internal/keyvault/vault_refreshpassphrase.go
GustavoStingelin: I wrote it manually 🤣, will fix the case.

wallet/internal/keyvault/vault_refreshpassphrase.go
GustavoStingelin: not sure, I will elaborate.

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: cool, much simpler this one!

wallet/internal/keyvault/keyvault.go
GustavoStingelin: btw, renamed to be clearer

wallet/internal/keyvault/vault_islocked.go
GustavoStingelin: grouped the files by component to reduce noise. I didn't merge everything into a single file because the upcoming PRs would make it too large

wallet/internal/keyvault/vault_lock_test.go
GustavoStingelin: removed

wallet/internal/keyvault/vault_refreshpassphrase.go
GustavoStingelin: fixed

wallet/internal/keyvault/vault_refreshpassphrase.go
GustavoStingelin: I put passphrase changing in the vault because it touches sensitive decrypted key material in memory and re-encrypts that material with the new passphrase. If we move this workflow completely outside the vault, the vault would need to expose a method that returns or otherwise leaks that sensitive material, which would defeat the purpose of this vault boundary.

One alternative would be for the vault to expose a narrower method that returns only the newly encrypted secrets, leaving the controller responsible for persisting them. That would keep plaintext material inside the vault, but it goes against the current design direction, where `WalletVault` already owns a `db.Store` and persists its own secret updates directly.

The reason for that design should become clearer in the follow-up PRs: the vault will cache sensitive/runtime key state, and letting it coordinate directly with the store avoids extra indirection and unnecessary database reads. External callers generally should not need to know which secret material is already cached inside the vault versus which data must be loaded from the database.

If we think the vault should not hold a `db.Store`, then that changes the whole model I have been building toward. In that case, I think we should revisit the already-merged ADR and realign on what responsibilities the vault should own versus what should stay in the controller/store layer.

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: I intentionally designed this to be a no-op instead of returning an error. Otherwise, the caller would need to enforce its own synchronization to avoid unlocking twice. With a no-op, the caller can decide whether synchronization is necessary.

If you strongly feel this should return an error instead, I can easily change it.

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: I changed the approach based on your previous suggestion to remove that `if`.

Zeroing these values is a common practice to avoid leaving sensitive data in memory. It doesn't force a GC cycle, so we don't need to worry about that at this stage.

Once we can adopt `secret.Do` from Go 1.26+, we'll have stronger guarantees that the memory is properly zeroed and handled securely. As I mentioned in the TODOs/comments, `secret.Do` is specifically designed for this use case and handles the necessary internals. We just need to wait until it becomes stable and we're able to bump our Go version.

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: applied

wallet/internal/keyvault/vault.go
GustavoStingelin: updated

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: nice catch, it was a stale clear, removed now.

wallet/internal/keyvault/vault_refreshpassphrase.go
GustavoStingelin: added as a func TODO in `makeRotatedWalletSecrets`

wallet/internal/keyvault/keyvault.go
```diff
@@ -37,10 +41,8 @@ type Vault interface {
...
-	// RefreshPrivatePassphrase refreshes vault owned runtime passphrase and
-	// crypto state after a successful private passphrase rotation.
-	//
-	// The vault must still be unlocked with the new passphrase when this method
-	// is called.
-	RefreshPrivatePassphrase(passphrase []byte) error
+	// ChangePassphrase rotates persisted wallet secrets to the provided
+	// new private passphrase. The vault must already be unlocked when this
+	// method is called.
+	ChangePassphrase(ctx context.Context, passphrase []byte) error
```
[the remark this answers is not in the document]
GustavoStingelin: adopted

wallet/internal/keyvault/vault.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// cryptoKeyScript is the key used to encrypt and decrypt script material.
+	cryptoKeyScript snacl.CryptoKey
+
+	// hdRootKey is the master HD extended key for the wallet, which can derive
+	// all sub scopes, accounts, addresses, and keys.
+	hdRootKey *hdkeychain.ExtendedKey
+}
+
+// Ensure WalletVault implements keyvault.Vault.
```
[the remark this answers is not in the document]
GustavoStingelin: adopted

wallet/internal/keyvault/vault_lifecycle_test.go
```diff
@@ -0,0 +1,422 @@
...
+	correctPassphrase = []byte("correct-passphrase")
+
+	// wrongPassphrase is used to assert wrong passphrase usage.
+	wrongPassphrase = []byte("wrong-passphrase")
+)
+
+// TestDBVaultLockClearsUnlockedState verifies that Lock returns the vault to
+// the locked state and wipes runtime secrets.
+func TestDBVaultLockClearsUnlockedState(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
GustavoStingelin: renamed

wallet/internal/keyvault/vault_cipher_vector_test.go
```diff
@@ -0,0 +1,208 @@
...
+		scriptCiphertextHex: "808182838485868788898a8b8c8d8e8f90919293949596" +
+			"9715d23fd581b57ca0117f583b2ee0198bf3b37d4770b55c05911d196893871" +
+			"95cb00c2b",
+	},
+}
+
+// TestDBVaultKnownVectorMasterKey verifies that each fixed passphrase and
+// serialized parameters derive the expected master private crypto key.
+func TestDBVaultKnownVectorMasterKey(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
GustavoStingelin: renamed

wallet/internal/keyvault/vault_cipher_test.go
```diff
@@ -0,0 +1,236 @@
...
+
+	"github.com/btcsuite/btcwallet/snacl"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/stretchr/testify/require"
+)
+
+// TestDBVaultEncryptSelectedRuntimeKeys verifies that Encrypt uses the selected
+// runtime crypto key and preserves snacl ciphertext semantics.
+func TestDBVaultEncryptSelectedRuntimeKeys(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
GustavoStingelin: renamed

GustavoStingelin:
Given an ACK, I can squash some commits for better history.

wallet/internal/keyvault/vault_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+		&masterPrivateKey, secrets.EncryptedCryptoScriptKey,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("crypto key script: %w", err)
+	}
+	defer cryptoKeyScript.Zero()
+
+	var hdRootKey *hdkeychain.ExtendedKey
+
+	if len(secrets.EncryptedMasterHdPrivKey) != 0 {
```
yyforyongyu: Q: is this a legal state if `len(secrets.EncryptedMasterHdPrivKey) == 0` yet we have `cryptoKeyPrivate` and `cryptoKeyPrivate`? If yes, explain why that will be the case; if not, we should handle the unexpected state? Either way, we should explain that with some inline docs.

wallet/internal/keyvault/vault_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+	encryptedCryptoKeyScript, err := newMasterPrivateKey.Encrypt(
+		v.unlockedState.cryptoKeyScript[:],
+	)
+	if err != nil {
+		return db.UpdateWalletSecretsParams{},
+			fmt.Errorf("encrypt crypto key script: %w", err)
+	}
+
+	var encryptedHDRootKey []byte
+	if v.unlockedState.hdRootKey != nil {
```
yyforyongyu: why do we need this step? The HD root key is encrypted under `cryptoKeyPrivate`, and `cryptoKeyPrivate` did not change, so the existing `secrets.EncryptedMasterHdPrivKey` could be carried forward instead?

wallet/internal/keyvault/vault_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+
+// Unlock loads wallet secrets from the store and decrypts them into runtime
+// state using the provided private passphrase. If the vault is already
+// unlocked, Unlock is a no-op and does not validate the passphrase.
+func (v *WalletVault) Unlock(ctx context.Context, passphrase []byte) error {
+	v.mtx.Lock()
+	defer v.mtx.Unlock()
+
+	if v.unlockedState != nil {
+		return nil
```
yyforyongyu: I would rather return an error so we can handle this case without asking `IsLocked`, which means the operation could be non-atomic. So in `lnd`, we will return an error when unlocking an already locked wallet, without returning an error here, it means in `lnd` it must first check `IsLocked` then return an error if locked, otherwise run `Unlock`, and its likely another service lock is needed there.

Another case is in bitcoind, it doesn't return an error at RPC iirc, but the timer will be reset. So without returning an error here, in the wallet's lifecycle, we need to perform a similar check when received an unlock request - if already locked -> extend the timeout; otherwise, unlock. And again these two operations are not atomic.

However if we return an error here, we can leave the callsite to decide how to handle it.

wallet/internal/keyvault/vault_unlock.go
yyforyongyu: > Zeroing these values is a common practice to avoid leaving sensitive data in memory. It doesn't force a GC cycle, so we don't need to worry about that at this stage.

What I meant was the level of documentation is important here - we know why we wanna do zeroing here and why we don't wanna rely on GC, but we cannot assume future devs will be aware. Writing explicit and detailed docs in a critical area is always a best practice. Btw this is just for future reference, as currently we don't have the `clearState` anymore.

wallet/internal/keyvault/vault_refreshpassphrase.go
yyforyongyu: yeah holding a mutex too long should always be avoided by a better design, but here i think we are good enough as all these ops are low-frequent

yyforyongyu (changes requested):
Ok a few more things then should be good to go!

wallet/internal/keyvault/vault.go
```diff
@@ -0,0 +1,73 @@
...
+type unlockedState struct {
+	// cryptoKeyPrivate is the key used to encrypt and decrypt private material.
+	cryptoKeyPrivate snacl.CryptoKey
+
+	// cryptoKeyScript is the key used to encrypt and decrypt script material.
+	cryptoKeyScript snacl.CryptoKey
+
+	// hdRootKey is the master HD extended key for the wallet, which can derive
+	// all sub scopes, accounts, addresses, and keys.
+	hdRootKey *hdkeychain.ExtendedKey
```
Abdulkbk, line 49: Since there is no method to access `hdRootKey` on `WalletVault`, callers like those in [account_manager.go](https://github.com/btcsuite/btcwallet/blob/sql-wallet/wallet/account_manager.go#L62-L68) separately fetch the secret seed from the db, call `keyVault.Decrypt` on it to get plaintext, meaning key material leaves the vault boundary.
```go
    plaintext, err := w.keyVault.Decrypt(waddrmgr.CKTPrivate, encrypted)
	if err != nil {
		return nil, fmt.Errorf("decrypt master HD priv: %w", err)
	}

	masterKey, err := hdkeychain.NewKeyFromString(string(plaintext))
	zero.Bytes(plaintext)
```

There is a todo in [account_derive.go](https://github.com/btcsuite/btcwallet/blob/sql-wallet/wallet/account_derive.go#L45C1-L48C68):
> // TODO(yy): Replace masterPrivKey with a KeyLocator once the vault grows
// full secret-management support; callers should hand off a locator and
// let the vault perform the hardened derivation internally so the
// plaintext master HD private key never leaves the vault boundary.

I wonder if it's within the scope of this PR?.

wallet/internal/keyvault/vault_lifecycle.go
Abdulkbk: If this is the case, should we consider removing `hdRootKey` from `unlockedState` altogether, since it's internal and we're only using it here (aside from side checks elsewhere)? After all, the call site has to derive it from the encrypted HD seed themselves atm.

wallet/internal/keyvault/vault_lifecycle.go
```diff
@@ -0,0 +1,353 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("parse master HD private key: %w", err)
+		}
+	}
+
+	state := &unlockedState{
+		cryptoKeyPrivate: cryptoKeyPrivate,
+		cryptoKeyScript:  cryptoKeyScript,
+
+		// Transfer HD root key pointer ownership to state, so we not zero it.
```
Abdulkbk: micro nit: `... so we do not zero it.`

Abdulkbk (commented):
concept ACK

wallet/internal/keyvault/vault.go
GustavoStingelin: In the future, callers will only be allowed to decrypt using `CKTScript`. At that point, this will prevent the HD key from leaking.

wallet/internal/keyvault/vault_lifecycle.go
GustavoStingelin: Yes, this is for the case where we have a watch-only wallet. In that case, we do not have the master HD private key, but we still need to decrypt the script data, so we still need the vault.

I will improve the documentation to make this clearer.

wallet/internal/keyvault/vault_lifecycle.go
GustavoStingelin: Yes, ideally we would not need this step. The issue is that the current store interface strongly requires this field when updating non-watch-only wallets, so it does not support storing the data without the master key.

That is why we are doing it here. To avoid doing this here, we would need a broader refactor on the store side, creating a method specific to this operation with a tighter interface.

If you think that would be a meaningful improvement for these operations, we can open a follow-up PR focused specifically on this step. I would avoid including it in this PR to keep the scope contained for now.

yyforyongyu: yeah `hdRootKey` should stay in `unlockedState` - think it will be used for HD derivation? and just to be clear, what i meant was we can drop the if v.unlockedState.hdRootKey != nil guard entirely and simply do,
```go
encryptedHDRootKey := append(
    []byte(nil), secrets.EncryptedMasterHdPrivKey...,
)
```

GustavoStingelin: Yeah, I'm on it

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: ok, now it's returning error

wallet/internal/keyvault/vault_unlock.go
GustavoStingelin: added some comments, PTAL

wallet/internal/keyvault/vault_lifecycle.go
GustavoStingelin: added a comment

wallet/internal/keyvault/vault_lifecycle.go
GustavoStingelin: ok, just did `EncryptedMasterHdPrivKey: secrets.EncryptedMasterHdPrivKey,` on the struct creation

wallet/internal/keyvault/vault_lifecycle.go
GustavoStingelin: Changed to return an error `ErrVaultUnlocked`

wallet/internal/keyvault/vault_lifecycle.go
GustavoStingelin: fixed

GustavoStingelin:
@yyforyongyu @Abdulkbk Hey guys, thanks for the thoughtful review. I think it's now ready for the final ack.

wallet/internal/keyvault/vault_lifecycle.go
```diff
@@ -28,3 +38,155 @@ func (v *WalletVault) IsLocked() bool {
...
+		&masterPrivateKey, secrets.EncryptedCryptoScriptKey,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("crypto key script: %w", err)
+	}
+	defer cryptoKeyScript.Zero()
+
+	var hdRootKey *hdkeychain.ExtendedKey
+
+	// Watch-only wallets may retain encrypted script crypto material without an
```
Abdulkbk: Just to confirm, is this block reachable for a `watch-only`? It looks like decrypting `cryptoKeyPrivate` above should error out:
```go
    // With the master key, we can start decrypting the other encrypted keys,
	// including the master HD private key that can spend coins.
	cryptoKeyPrivate, err := decryptCryptoKey(
		&masterPrivateKey, secrets.EncryptedCryptoPrivKey,
	)
	if err != nil {
		return nil, fmt.Errorf("crypto key private: %w", err)
	}
	defer cryptoKeyPrivate.Zero()
```

Abdulkbk (approved):
LGTM

wallet/internal/keyvault/vault_lifecycle.go
GustavoStingelin: added a new commit to solve this

GustavoStingelin: This comment helped me find a flaw in the secret storage logic. It was not allowing us to save the keys needed for the watch-only case.

I added one more commit to fix it, PTAL

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` 🛡️

yyforyongyu:
No issues found by gpt-5.5 🦉

yyforyongyu (approved):
LGTM 🚢

PR #1275, opened 2026-06-23
wallet/txauthor+wtxmgr: pin new leaf tags, drop replaces (submodule tag prep, round 2)

Roasbeef:
In this PR, we do the second round of submodule tag prep on top of #1274.
Now that the leaf modules are tagged (`walletdb/v1.6.0`,
`wallet/txrules/v1.3.0`, `wallet/txsizes/v1.3.0`), we can point the level-1
modules at those tags and drop the `../` replace rules they used to build
against the in-tree copies during the btcd v2-module migration.

For `txauthor`, we bump `txrules` and `txsizes` to `v1.3.0` and align the
indirect `btcd` require to `v0.26.0`. For `wtxmgr`, we bump `walletdb` to
`v1.6.0`. Both build and test green against the published tags, so once
this lands they're ready to be tagged `wallet/txauthor/v1.4.0` and
`wtxmgr/v1.6.0`.

## On the wtxmgr btcwallet require

One thing worth calling out: we keep `github.com/btcsuite/btcwallet v0.16.18`
in `wtxmgr/go.mod`. It looks like a stale self-reference, but it isn't, it's
there for the `github.com/btcsuite/btcwallet/build` logging subpackage that
wtxmgr imports. That package only pulls in `btclog`, so it compiles cleanly
against the old tag and no btcd v2 types leak in through it.

This is the one genuinely circular edge in the graph (`wtxmgr -> btcwallet/build
-> root`). Pinning it at `v0.16.18` is harmless: when the root module itself
consumes `wtxmgr`, the main module always wins in MVS, so `v0.16.18` is never
actually selected there. It only matters for standalone wtxmgr builds, where
the old `build` package is perfectly fine. Once the root is tagged we can
optionally bump this in a follow-up, but it's not needed for correctness.

## What's next

After this merges I'll tag `wallet/txauthor/v1.4.0` and `wtxmgr/v1.6.0`,
then pause so we can sort out the neutrino side
([lightninglabs/neutrino#364](https://github.com/lightninglabs/neutrino/pull/364)):
bump its btcwallet submodule requires to the new tags, drop the
`guggero/neutrino` replace, push a neutrino tag, and finally pin that back
in the btcwallet root `go.mod` (which also drops the last replace rules
there).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,9 +12,9 @@ require (
 	github.com/btcsuite/btcd/wire/v2 v2.0.0
 	github.com/btcsuite/btclog v1.0.0
 	github.com/btcsuite/btcwallet/wallet/txauthor v1.3.5
-	github.com/btcsuite/btcwallet/wallet/txrules v1.2.2
-	github.com/btcsuite/btcwallet/wallet/txsizes v1.2.5
-	github.com/btcsuite/btcwallet/walletdb v1.5.1
+	github.com/btcsuite/btcwallet/wallet/txrules v1.3.0
+	github.com/btcsuite/btcwallet/wallet/txsizes v1.3.0
+	github.com/btcsuite/btcwallet/walletdb v1.6.0
 	github.com/btcsuite/btcwallet/wtxmgr v1.5.6
 	github.com/btcsuite/websocket v0.0.0-20150119174127-31079b680792
 	github.com/davecgh/go-spew v1.1.1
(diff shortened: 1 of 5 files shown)


Roasbeef:
Force-pushed a fix for CI. The failure was at the root module: since the root
builds the submodules through its local `./` replace rules, bumping the
submodules' internal requires (txauthor now needs txrules v1.3.0, etc.) raised
the version floor the root sees, leaving the root `go.mod` stale
(`go install` failed with "updates to go.mod needed").

Folded a root `go mod tidy` into the same commit (one self-contained commit so
`check-commit.sh` stays green per-commit): it bumps the root's txrules/txsizes
to v1.3.0 and walletdb to v1.6.0. The replaces stay, these are just floors that
will already be correct when we drop the replaces in the final root-tag step.

Verified locally: `go install`, `go build ./...`, `go vet ./...`, and the
`check-commit.sh` test-compile (`go test -run=_NONE_`) all pass.

PR #1278, opened 2026-07-04
multi: bump stale Go version references

GustavoStingelin:
Update stale Go version references to match the current repository go.mod version: `1.25.11`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,7 +25,7 @@ env:
 
   # If you change this please run `make lint` to see where else it needs to be
   # updated as well.
-  GO_VERSION: 1.24.6
+  GO_VERSION: 1.25.11
 
   BITCOIND_VERSION: '22.0'
   BITCOIND_IMAGE: 'lightninglabs/bitcoin-core'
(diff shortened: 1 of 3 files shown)


yyforyongyu (approved):
LGTM 🗄️

.github/workflows/main.yml
```diff
@@ -25,7 +25,7 @@ env:
 
   # If you change this please run `make lint` to see where else it needs to be
   # updated as well.
-  GO_VERSION: 1.24.6
+  GO_VERSION: 1.25.11
 
   BITCOIND_VERSION: '22.0'
```
Abdulkbk, line 30: are we updating the bitcoind version in this PR or in a follow up?

Abdulkbk (approved):

.github/workflows/main.yml
GustavoStingelin: Not in this one, maybe in a follow-up, but I didn't pay attention to it yet to see if sth will break. Do you wanna take the bitcoind bump?

Abdulkbk: Sure. I will look into it.

PR #1279, opened 2026-07-05
wallet: add sign methods to vault interface

GustavoStingelin:
Adds a digest-level signing surface to the wallet keyvault boundary so future
signing paths can keep private key operations behind the vault.

This PR introduces a shared `wallet/signing` package with:

- sealed key locator variants:
  - full HD derivation address via `waddrmgr.AddrScope`
  - account-row child reference
  - script pubkey reference
- sealed primitive signing request variants:
  - ECDSA
  - compact recoverable ECDSA
  - Schnorr
  - tweaked Schnorr
- sealed signature result variants with `Serialize()` helpers

The keyvault interface is extended with:

- `DerivePubKey`
- `Sign`

`Sign` accepts a sealed `signing.Request` containing a key locator, a
caller-computed 32-byte digest, and the primitive signing mode. Requests do not
carry transactions, PSBTs, scripts, witnesses, sighash types, or Taproot tree
data. Digest construction, transaction assembly, signature serialization format,
and appended sighash-byte handling remain wallet-layer responsibilities.

Tweaked Schnorr signing uses an already-computed additive secp256k1 scalar plus
an even-Y normalization flag. It does not accept Taproot script roots, merkle
roots, or BIP341 tweak input bytes.

`WalletVault` currently returns a generic key-operation unimplemented sentinel
for derivation and signing until the backing implementation is added.

The vault mock was moved into its own `mock/vault` package to avoid an import
cycle with `waddrmgr`.


wallet/internal/keyvault/keyvault.go
```diff
@@ -19,8 +22,77 @@ var ErrVaultLocked = errors.New("vault is locked")
...
+	NormalizeEvenY bool
+}
+
 // Vault manages the lock lifecycle and cryptographic operations for wallet key
 // material.
+// It is intentionally aggregated as the wallet-scoped boundary for lock,
+// encryption, derivation, and signing lifecycle.
+//
+//nolint:interfacebloat
 type Vault interface {
```
GustavoStingelin: I'm thinking of splitting the interface into small ones: `VaultLifecycle`, `VaultCipher`, and `VaultSigner`. WDYT?

GustavoStingelin:
@Abdulkbk ping for review

docs/developer/adr/0010-keyvault-encryption-layer.md
```diff
@@ -30,6 +30,37 @@ to each lock, unlock, encrypt, decrypt, or derivation operation.
 Keyvault accesses persistence through `db.Store`. It does not talk directly to
 SQL backends.
 
+The vault API also includes digest level signing methods. `DerivePubKey`,
```
yyforyongyu: 👍

yyforyongyu:
No issues found by `claude-opus-4-8[1m]` ✍️

wallet/internal/bwtest/mock/vault/vault.go
```diff
@@ -0,0 +1,178 @@
...
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcd/btcec/v2/ecdsa"
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet/internal/keyvault"
+	"github.com/stretchr/testify/mock"
+)
+
+// Vault is a testify-based mock for the wallet key-vault interface.
```
yyforyongyu, line 17: hmm why this change? we usually would name mocked objects as `MockXXX` so it's easier to search, but guess this is less relevant now given that agents do the work.

GustavoStingelin: I did not change the struct name, just moved the files. I can open a follow-up, renaming this.

I guess the LLM just copied from the `mock.Store` style, which I think is not bad either, since it's inside the mock pkg

wallet/internal/keyvault/keyvault.go
```diff
@@ -19,8 +22,77 @@ var ErrVaultLocked = errors.New("vault is locked")
...
+}
+
+// ScriptPubKeyRef identifies a raw imported row by script pubkey.
+type ScriptPubKeyRef struct {
+	// ScriptPubKey is the raw locking script.
+	ScriptPubKey []byte
+}
+
+// SchnorrKeyTweak describes an additive secp256k1 tweak for Schnorr signing.
+type SchnorrKeyTweak struct {
```
yyforyongyu: Can we make the conversion semantics explicit here? The existing signer code models taproot tweaks as raw `[]byte`, while this type adds a fixed 32-byte scalar and `NormalizeEvenY`. That may be the right vault-level shape, but the signer-to-vault adapter needs to define exactly how current taproot tweak bytes map into this descriptor so delegation cannot silently change the tweak being signed with. (gpt-5.5)

wallet/internal/keyvault/keyvault.go
```diff
@@ -49,4 +121,42 @@ type Vault interface {
...
 	// is called.
 	ChangePassphrase(ctx context.Context, newPassphrase []byte) error
+
+	// DerivePubKey derives the public key for the provided key locator.
+	//
+	// The caller provides a wallet key locator, and the vault returns the
+	// corresponding public key without exposing private key material.
+	DerivePubKey(ctx context.Context, ref KeyLocator) (*btcec.PublicKey, error)
+
+	// SignECDSA signs a caller-computed 32-byte digest for the provided key
```
yyforyongyu: Can the PR document the intended layering between this primitive signing surface and the existing intent-based signer API? Keeping `SignDigestIntent` in the signer and unpacking it into `SignECDSA`, `SignCompactECDSA`, `SignSchnorr`, or `SignTweakedSchnorr` makes sense, but the API docs should make clear that the signer owns intent validation and dispatch while the vault intentionally exposes only primitive signing methods. (gpt-5.5)

wallet/internal/keyvault/keyvault.go
```diff
@@ -19,8 +22,77 @@ var ErrVaultLocked = errors.New("vault is locked")
...
+	// AccountKey names a child under a persisted account row, including
+	// imported xkey accounts.
+	AccountKey *AccountKeyRef
+
+	// ScriptPubKey names a raw imported address or private-key row by script
+	// pubkey.
+	ScriptPubKey *ScriptPubKeyRef
+}
+
+// FullHDKeyRef identifies a complete wallet-root derivation path.
```
yyforyongyu: This should probably reuse the existing derivation types instead of flattening them. The flattened form loses the `InternalAccount`/`Account` split and `MasterKeyFingerprint`; `InternalAccount` is what `DeriveFromKeyPath*` uses to load the account row, so this API cannot round-trip the current wallet signing path. Consider modeling this as `waddrmgr.KeyScope` plus `waddrmgr.DerivationPath` or a local wrapper that embeds both so the vault locator preserves the full derivation identity. (gpt-5.5)

wallet/internal/bwtest/mock/vault/vault.go
yyforyongyu: yeah non-blocking, just curious here

wallet/internal/keyvault/keyvault.go
```diff
@@ -19,8 +22,77 @@ var ErrVaultLocked = errors.New("vault is locked")
 // vault was already unlocked.
 var ErrVaultUnlocked = errors.New("vault is already unlocked")
 
+// ErrVaultSigningUnimplemented reports that the vault does not implement key
+// signing yet.
+var ErrVaultSigningUnimplemented = errors.New("vault signing is unimplemented")
+
+// KeyLocator identifies a key using one of the vault's supported reference
+// forms.
+type KeyLocator struct {
```
Abdulkbk: will it make sense to specify what implementations must do when zero or multiple are set and maybe adding a `Validate` helper to enforce how this will be used?

wallet/internal/keyvault/keyvault.go
```diff
@@ -49,4 +121,42 @@ type Vault interface {
 	// private passphrase. The vault must already be unlocked when this method
 	// is called.
 	ChangePassphrase(ctx context.Context, newPassphrase []byte) error
+
+	// DerivePubKey derives the public key for the provided key locator.
+	//
+	// The caller provides a wallet key locator, and the vault returns the
+	// corresponding public key without exposing private key material.
+	DerivePubKey(ctx context.Context, ref KeyLocator) (*btcec.PublicKey, error)
```
Abdulkbk: nit: can specify wether the vault needs be locked/unlocked for the newly added medthods.

wallet/internal/keyvault/vault.go
```diff
@@ -63,6 +67,42 @@ func NewWalletVault(store db.Store, walletID uint32,
 	}
 }
 
+// DerivePubKey reports that WalletVault does not yet expose key derivation.
+func (v *WalletVault) DerivePubKey(context.Context,
+	KeyLocator) (*btcec.PublicKey, error) {
+
+	return nil, ErrVaultSigningUnimplemented
```
Abdulkbk: not important atm, `DerivePubKey` isn't signing, so `ErrVaultSigningUnimplemented` doesn't apply.

wallet/internal/keyvault/keyvault.go
Abdulkbk: I agree, so `//nolint:interfacebloat` can go away

Abdulkbk (approved):
LGTM 🚢

wallet/internal/keyvault/keyvault.go
yyforyongyu: I probably would take the other route to make sure the interface is deep (from one of my favorite books, *A Philosophy of Software Design* - deep modules are the ultimate tool for managing code complexity). It would be sth like `keyvault.Sign`, then there is no need to split. Just a sketch, still need mutex i guess,

```go
type Vault interface {
    ...
	Sign(ctx context.Context, ref KeyLocator,
		req SignRequest) (Signature, error)
}

type SignRequest struct {
	Digest [32]byte
	Type   SignatureType

	CompactECDSA  bool
	CompressedKey bool

	SchnorrTweak *SchnorrKeyTweak
}

type Signature interface {
	Type() SignatureType
	Serialize() []byte
	isSignature()
}
```

GustavoStingelin: Cool, I liked it. The only caveat I see is that it makes the request a bit weak, because it allows callers to construct invalid states that are only caught later by a validation method. That puts some extra burden on the caller to understand which parameter combinations are valid.

Given that, I dug a bit deeper and made it sealed too, similar to `Signature`. Now callers can express their exact intent by construction, which makes it impossible to mix a Schnorr tweak with ECDSA, for example.

Other tradeoff is that callers may have a bit more work if they need access to specific signature internals, like raw `R` and `S` values, because they may need to assert the returned signature type. But I think the generic `Serialize()` method should cover most use cases.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: Changed to the modular scalar to be explicit.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: Created a new `wallet/signing` package so the vault and public signer can share the same API.

Since the vault needs to support the same signing operations exposed by the public signer, this helps avoid duplicated types and extra conversions between equivalent request and response structs.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: It intentionally does not use `InternalAccount` or `MasterKeyFingerprint`, because we do not need them here. Those fields are tied to the legacy `kvdb` flow.

I changed it to use the new `waddrmgr.AddrScope`, which includes the full path.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: It was supposed to use only one reference at a time. I converted it to sealed structures behind an interface, so now this is guaranteed by construction, and callers cannot accidentally provide multiple references.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: added

wallet/internal/keyvault/vault.go
GustavoStingelin: yeah, that's just a temporary error until the next PR. Renamed.

GustavoStingelin:
Added a new `wallet/signing` package with primitives that can be used by the vault now and later reused by the public signer.

This should let us remove some duplicated code later and keep the signing request and response types in one place.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: Adding a bit more detail here: the current shape makes the vault boundary take
the final additive scalar directly.

`SchnorrKeyTweak.Scalar` is a `btcec.ModNScalar`, so callers must pass the
already-computed secp256k1 tweak scalar. It is intentionally not a Taproot script
root, merkle root, or BIP341 tweak input byte string.

So any signer-to-vault adapter remains responsible for converting existing raw
Taproot tweak inputs into that final scalar before constructing the
`TweakedSchnorrRequest`. The vault only applies the scalar tweak and signs the
caller-computed digest.

wallet/internal/keyvault/keyvault.go
GustavoStingelin: Also clarified the layering in the ADR and `Vault.Sign` docs.

The wallet/signer layer owns the higher-level intent work: choosing the signing
intent, constructing the digest, selecting sighash semantics, assembling
transactions/PSBTs/witnesses, deciding serialization format, and appending any
sighash byte.

The vault boundary stays primitive: it receives a sealed `signing.Request` for a
located key and caller-computed 32-byte digest, performs the private-key
operation, and returns a primitive `signing.Signature`.

wallet/signing/key_locator.go
```diff
@@ -0,0 +1,48 @@
...
+type KeyLocator interface {
+	// isKeyLocator seals KeyLocator to variants defined in this package.
+	isKeyLocator()
+}
+
+// FullHDKeyLocator identifies a wallet-root derivation address using the
+// existing waddrmgr address scope type.
+type FullHDKeyLocator struct {
+	// AddrScope identifies the wallet-root derivation address.
+	AddrScope waddrmgr.AddrScope
```
yyforyongyu, line 15: `waddrmgr.AddrScope` still cannot represent the existing public signer path because it omits `DerivationPath.InternalAccount` and `MasterKeyFingerprint`; either keep a locator based on `waddrmgr.KeyScope` plus `waddrmgr.DerivationPath`, or document that this variant is not meant to adapt `wallet.BIP32Path`. (`gpt-5.5`)

wallet/signing/request.go
```diff
@@ -0,0 +1,99 @@
...
+func (ECDSARequest) isRequest() {}
+
+// CompactKeyFormat describes the public-key header semantics for compact ECDSA
+// signatures.
+type CompactKeyFormat uint8
+
+const (
+	// CompactKeyUncompressed requests compact signature header semantics for an
+	// uncompressed public key.
+	CompactKeyUncompressed CompactKeyFormat = iota
```
yyforyongyu: `CompactKeyUncompressed` is the zero value, so a caller that omits `KeyFormat` silently gets an uncompressed recovery header even though the existing signer compact path always passes `compressed=true`; add an unknown zero sentinel that implementations reject, or make the zero value match the expected compressed default. (`gpt-5.5`)

wallet/signing/signature.go
```diff
@@ -0,0 +1,71 @@
...
+	Serialize() []byte
+
+	// isSignature seals Signature to variants defined in this package.
+	isSignature()
+}
+
+// ECDSASignature wraps a normal ECDSA signature result.
+type ECDSASignature struct {
+	// Signature is the generated ECDSA signature.
+	Signature *ecdsa.Signature
```
yyforyongyu: The exported pointer-backed signature variants make `ECDSASignature{}` or `SchnorrSignature{}` non-nil `Signature` values whose `Serialize` returns nil; hide the fields behind constructors or store concrete values so successful signing cannot return an invalid wrapper. (`gpt-5.5`)

wallet/internal/keyvault/keyvault.go
yyforyongyu: yeah correct think we can mimic the current `wallet/signer.go` as it already has everything.

wallet/signing/signature_test.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/btcsuite/btcd/btcec/v2/schnorr"
+	"github.com/stretchr/testify/require"
+)
+
+// TestECDSASignatureSerializeNil checks that a nil ECDSA signature serializes
+// to nil.
+func TestECDSASignatureSerializeNil(t *testing.T) {
+	t.Parallel()
+
+	// Given
```
yyforyongyu: we should fix these?

docs/developer/adr/0010-keyvault-encryption-layer.md
```diff
@@ -30,6 +30,44 @@ to each lock, unlock, encrypt, decrypt, or derivation operation.
...
 
+The vault API also includes a digest level signing method. `DerivePubKey` and
+`Sign` keep private key use inside the same wallet scoped boundary that owns
+lock state and key lifecycle. Callers identify keys with a sealed
+`wallet/signing.KeyLocator` variant: `FullHDKeyLocator` for a full HD
+derivation path, `AccountKeyLocator` for an account row child reference, or
+`ScriptPubKeyLocator` for raw or imported rows without exposing private key
+material or database routing.
+
+Wallet code builds a sealed `wallet/signing.Request` containing both the key
```
yyforyongyu, line 41: is there another consumer of `signing`? I don't see the necessity of making it a public module, nor making it an independent module. We can just put all the code under `wallet/internal/keyvault` as the vault is the only consumer?

wallet/signing/key_locator.go
yyforyongyu: Invalid - take a look again, this is used to sign the new sql path, not the kvdb path. I think the `InternalAccount` is only used in the legacy path, and we will use the legacy keyvault for kvdb?

wallet/internal/keyvault/keyvault.go
yyforyongyu: I am confused about the design - if the `signer.go` and the keyvault are using the same struct, like sign request param to pass info, there must be sth wrong with the abstraction as how can both the public API and the private internal vault require the same set of info? I think it's better to open another PR to show how it's actually used in the `wallet/signer.go` first.

yyforyongyu (changes requested):
Cool. I think the current `wallet/signing` looks premature without another PR to show how it can be applied to `wallet/signer.go` first. It's weird that both the public Signer and the internal keyvault can share the same signing request, tho not necessarily wrong - either the public layer is a pass-through, so a slim, shallow layer, or the internal layer is leaking. So it will be better to show how this new pkg will be applied first, otherwise I think `wallet/signing` should be removed and put everything under the `internal/keyvault`.

There is still the option that we can revert to the previous version where we have individual methods like `SignECDSA`, `SignSchnorr`, etc., but without a concrete implementation first it's hard to tell.

wallet/signing/key_locator.go
GustavoStingelin: I think the legacy keyvault should only be needed during migration. Outside of that, we should use only the new keyvault, so this handling probably belongs in the temporary migration code.

wallet/signing/signature_test.go
GustavoStingelin: u mean removing or renaming to "Arrange, Act, Assert"?

GustavoStingelin:
@yyforyongyu I think the root issue is that the current
`wallet.SignDigest` shape does not line up well with the future keyvault path.
It currently assumes the wallet signer can derive/load the private key itself,
but once signing is delegated to the vault, the public signer should not need
that knowledge.

I actually like the shape where the public signer becomes a thin adapter:

```go
// SignDigest signs a message digest based on the provided request.
func (w *Wallet) SignDigest(ctx context.Context,
	request signing.Request) (signing.Signature, error) {

	if err := w.state.canSign(); err != nil {
		return nil, err
	}

	return w.keyVault.Sign(ctx, request)
}
```
To me, this makes the layering pretty clear: the wallet signer performs
wallet-state authorization, while the vault owns key location and primitive
signing. From that perspective, a shared `signing.Request` does not necessarily
mean the internal layer is leaking; it can also be the boundary object passed
from the public signer into the vault.
That said, I also understand the concern that this may be premature in this PR,
because the current patch does not yet fully migrate wallet/signer.go to that
shape. Without that follow-through, wallet/signing can look like a public
package introduced mainly for the vault.

So I think there are two reasonable directions:
1. Keep the shared `signing.Request` model, where `wallet.SignDigest`
delegates to `keyVault.Sign`.
3. Keep this PR narrower by using explicit vault methods like SignECDSA,
SignCompactECDSA, SignSchnorr, etc., and defer the request/public
signer cleanup to a follow-up PR.

Personally, I like the thin `SignDigest -> keyVault.Sign` shape, but I also
agree the explicit-method version is less speculative for this PR and easier to
review locally. Since neither of us seems strongly attached to the single-method
vault API, I’m leaning toward switching this PR back to the explicit vault
methods and leaving the public signer for a follow-up.

Mainly, I want to avoid continuing to go back and forth on the API shape in this
PR. Which direction sounds better to you?

wallet/signing/request.go
GustavoStingelin: added

wallet/signing/signature.go
GustavoStingelin: changed to value semantics

wallet/signing/signature_test.go
GustavoStingelin: removed, btw

GustavoStingelin:
Converted to draft; planned to rework this after SQL migration.

Issue #1079, opened 2025-09-25
[bug]: test flake in `TestPrunedBlockDispatcherMultipleQueryPeers`


yyforyongyu:
Found in this [build](https://github.com/btcsuite/btcwallet/actions/runs/18001078470/job/51210294960?pr=1077)
```
--- FAIL: TestPrunedBlockDispatcherMultipleQueryPeers (5.01s)
    pruned_block_dispatcher_test.go:635: expected nil err to signal completion
    pruned_block_dispatcher_test.go:186: did not consume all queriedPeer signals
2025-09-25 08:09:23.422 [DBG] CHIO: Received unexpected message *wire.MsgSendAddrV2 from peer 10.0.0.1:18
2025-09-25 08:09:23.422 [DBG] CHIO: Starting worker for peer 10.0.0.1:18
2025-09-25 08:09:23.423 [DBG] CHIO: Failed connecting to peer 10.0.0.1:20: shutting down
FAIL
2025-09-25 08:09:23.423 [DBG] CHIO: Received unexpected message *wire.MsgSendAddrV2 from peer 10.0.0.1:20
```

PR #1280, opened 2026-07-05
rpc: reject raw transactions with trailing bytes

Lrifton92:
### Problem

Three RPC handlers decode a client-supplied transaction with `MsgTx.Deserialize` but never verify that the entire input was consumed. `Deserialize` stops once it has read a complete transaction and leaves any extra bytes untouched, so a transaction encoded with **trailing bytes appended** is accepted rather than rejected as malformed:

- `rpc/legacyrpc` — `signRawTransaction` (JSON-RPC `signrawtransaction`). Bitcoin Core rejects such input with `TX decode failed`, and this handler already uses that exact message for decode failures, so accepting trailing garbage is an oversight.
- `rpc/rpcserver` — `SignTransaction` and `PublishTransaction` (gRPC), same missing check.

### Fix

Read the transaction through a `*bytes.Reader` and, after a successful `Deserialize`, require that no bytes remain, returning each handler's existing error (`DeserializationError` / `codes.InvalidArgument`). Minimal change, existing error paths reused.

### Tests

- `TestSignRawTransactionTrailingBytes` (legacyrpc)
- `TestSignTransactionTrailingBytes`, `TestPublishTransactionTrailingBytes` (rpcserver)

Each serializes a minimal valid transaction, appends a stray byte, and asserts the request is rejected. The trailing-bytes check runs before the wallet is used, so the handlers are exercised directly without a wallet. Verified each test fails without the fix (the malformed tx is accepted past the decode stage) and passes with it. `go build`/`go vet`/`gofmt` clean; both `rpc/legacyrpc` and `rpc/rpcserver` suites pass.

Same class of decoder-strictness fix recently consolidated in btcd (btcsuite/btcd#2558).

### DIFF
--- a/rpc/legacyrpc/methods.go
+++ b/rpc/legacyrpc/methods.go
@@ -1590,9 +1590,19 @@ func signRawTransaction(icmd interface{}, w *wallet.Wallet, chainClient *chain.R
 		return nil, err
 	}
 	var tx wire.MsgTx
-	err = tx.Deserialize(bytes.NewBuffer(serializedTx))
+	txReader := bytes.NewReader(serializedTx)
+	err = tx.Deserialize(txReader)
 	if err != nil {
 		e := errors.New("TX decode failed")
+		return nil, DeserializationError{e}
+	}
+
+	// The transaction must consume the entire input. Any remaining bytes
+	// mean the serialized transaction is malformed (extra trailing data),
+	// which Bitcoin Core also rejects with "TX decode failed" rather than
+	// silently ignoring the excess.
+	if txReader.Len() != 0 {
+		e := errors.New("TX decode failed: unexpected trailing bytes")
 		return nil, DeserializationError{e}
 	}
 
--- a/rpc/legacyrpc/signrawtransaction_test.go
+++ b/rpc/legacyrpc/signrawtransaction_test.go
@@ -1,0 +1,48 @@
+// Copyright (c) 2013-2014 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package legacyrpc
+
+import (
+	"bytes"
+	"encoding/hex"
+	"testing"
+
+	"github.com/btcsuite/btcd/btcjson"
+	"github.com/btcsuite/btcd/wire/v2"
+)
+
+// TestSignRawTransactionTrailingBytes ensures signRawTransaction rejects a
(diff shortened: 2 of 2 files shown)


Lrifton92:
Gentle ping — this is still open and mergeable with no conflicts. It rejects raw transactions that carry trailing bytes after the valid encoding, a small and targeted validation fix. Happy to address any feedback whenever a maintainer has a moment. Thanks!

Lrifton92:
Small process question rather than a nudge on the review itself.

The CI has never run on this PR. The head commit `c73ffbe` has zero check-runs attached, while commits on
master carry 26 — so there is no build or test signal on this branch at all, which is consistent with the
`BLOCKED` merge state. I suspect it's the "approve workflows for outside contributors" gate rather than
anything in the branch.

Could a maintainer approve the workflow run? I'm happy to rebase onto master first if that helps re-trigger
it. The branch itself is mergeable with no unresolved threads.

Lrifton92:
Could a maintainer approve the CI workflow runs for this PR when you get a chance? As an outside contributor the workflows need manual approval, so the build and tests haven't run yet. Happy to address any feedback once CI is green.

Lrifton92:
Friendly ping: this is ready for review, but the CI workflows haven't run yet — for an outside contributor GitHub gates them behind a maintainer approving the run. Could a maintainer kick off the workflow run so there's a green build to review against? The branch is rebased and mergeable, and I'm happy to rebase again or address anything once CI is visible. Thanks!

yyforyongyu (commented):
Approved CI run - given that we have a new wallet coming up and the RPC are mostly deprecated, this will be a low prio PR.

Lrifton92:
Thanks for approving the run. Pushed 2082170f, which fixes the 10 golangci-lint findings from that run in `rpc/rpcserver` (protogetter, wsl_v5, noinlineerr, paralleltest). `golangci-lint run` at the pinned tools version reports 0 issues on `rpc/rpcserver` and `rpc/legacyrpc`, and `go test ./rpc/...` passes.

The `unit-cover` failure was `TestPrunedBlockDispatcherMultipleQueryPeers` in `chain/`, which this PR does not touch; it looks like the known flake tracked in #1079.

Understood on the low priority given the new wallet work.

Lrifton92:
@yyforyongyu when you have a moment, could you approve the CI run for 2082170f? It only contains the lint fixes from the previous run. Understood this is low priority given the RPC deprecation, so no rush.

PR #1281, opened 2026-07-11
docs: inventory SQL port source history

Roasbeef:
In this PR, we freeze the full SQL port source universe before moving code onto the smaller port-first branch.

This draft is the review and attribution artifact for the salvage effort. It isn't part of the final port stack, so we can leave it unmerged once the concrete Stage 1 work has carried the required code and contributor history forward.

The inventory covers all 152 PRs in the series and all 969 commits between `master` and `sql-wallet`. Each commit retains its original SHA, author identity, changed paths, disposition, and destination cut. GitHub still identifies a canonical feeder for 107 commits; the remaining 862 are explicitly unresolved because repeated rebases and force-pushes removed the historical SHA links. We keep that uncertainty visible instead of guessing away contributor history.

The PR also records the existing btcwallet API used by lnd and pins the surviving umbrella and feeder heads. The included generators make the PR inventory, commit map, and API baseline reproducible.

Verification:

- exactly 969 unique commit rows in source ancestry order
- exactly 152 unique PR rows across all four cohorts
- `go test ./scripts`
- shell syntax and whitespace checks
- independent full review


Roasbeef:
Replaced by #1286 after renaming the head branch to remove the codex prefix. The head SHA and artifact contents are unchanged.

PR #1282, opened 2026-07-11
wallet/sql: add SQL build foundation

Roasbeef:
In this PR, we add the SQL build foundation without routing any wallet operation through a new store or changing a public interface.

The source-history and attribution inventory lives separately in draft PR #1281. This PR is the first concrete cut and is based directly on `master`.

The change adds pinned sqlc tooling, generated-code checks, and the first SQLite and PostgreSQL schema slice for block metadata. The generated packages remain internal to the wallet, and both dialects use the same query contract.

This is the first Stage 1 PR in the port-first stack. It is intentionally limited to build support and block schema so later schema reviews can be evaluated independently from runtime and control-flow changes.

Verification:

- `make sqlc-check`
- `go test ./wallet/internal/sql/...`
- no changes to `wallet.Interface`, `waddrmgr`, `wtxmgr`, or loader control flow


Roasbeef:
Replaced by #1287 after renaming the head branch to remove the codex prefix. The head SHA and commit structure are unchanged.

PR #1283, opened 2026-07-11
wallet/sql: preserve transaction store semantics

Roasbeef:
In this PR, we add the port-first transaction, input, credit, and output lease schema for SQLite and PostgreSQL.

This PR is stacked on #1285, which preserves the legacy wallet, scope, account, and address schema.

The long-running branch had moved to a replacement-history state machine that current `wtxmgr` does not expose. This PR keeps the current model instead: active mined and unmined incidences, physical conflict removal, duplicate mined incidences across blocks, and multiple unmined spenders for one outpoint. The rewrite-only replacement table is deferred.

Credits retain the observable change bit and actual transaction output script. A canonical credit-incidence relation keeps duplicate mined incidences from duplicating one UTXO, while an exact mined debit relation preserves incidence-specific spent state. Labels remain keyed by transaction hash and survive incidence deletion. Leases remain keyed by outpoint, use Unix-second expiry, and can exist independently of a credit row. Atomic lease acquisition preserves same-ID renewal, expired takeover, and rejection of a different live lock ID.

This completes the Stage 1 schema foundation. It does not add a store adapter or change wallet control flow.

Verification:

- SQLite and PostgreSQL empty-to-head-to-empty-to-head migration tests
- duplicate incidence, multiple spender, label, credit, output script, and lease parity tests
- `sqlc vet -f sqlc.yaml`
- `make fmt-check tidy-module-check rpc-check sqlc-check`
- unmodified lnd tests for `./lnwallet/btcwallet`, `./keychain`, and `./lnrpc/walletrpc`
- no changes to the existing wallet, waddrmgr, or wtxmgr public surfaces


Roasbeef:
Replaced by #1290 after renaming the head branch to remove the codex prefix. The head SHA and commit structure are unchanged.

PR #1284, opened 2026-07-11
wallet/sql: add database migration foundation

Roasbeef:
In this PR, we add the database connection and migration foundation for the port-first SQL stack.

This PR is stacked on #1282, which adds the build and block-schema foundation.

The SQLite and PostgreSQL pools use bounded connection settings and embedded migrations that can move from empty to head, back to empty, then forward again. PostgreSQL accepts both URL and keyword/value DSNs, enforces TLS without rewriting either form, and runs migrations through a short-lived owned pool so repeated migration calls do not consume the wallet pool. We use the released `github.com/lightningnetwork/lnd/sqldb` module at `v1.0.13` for the shared SQL utility layer and SQLite defaults. btcwallet retains its own schema and rollback-capable migration runner because the concrete lnd stores own lnd-specific schema and global migration ordering.

No wallet method is routed through SQL in this PR. This only establishes the internal database boundary used by the following schema PRs.

Verification:

- SQLite empty-to-head-to-empty-to-head migration test
- PostgreSQL empty-to-head-to-empty-to-head migration test
- `go vet ./wallet/internal/sql/...`
- `make sqlc-check`


Roasbeef:
Replaced by #1288 after renaming the head branch to remove the codex prefix. The head SHA and commit structure are unchanged.

PR #1285, opened 2026-07-11
wallet/sql: preserve legacy wallet schema

Roasbeef:
In this PR, we add the port-first wallet, scope, account, and address schema for SQLite and PostgreSQL.

This PR is stacked on #1284, which adds the database connection and migration foundation.

The schema preserves the current dual-passphrase encryption hierarchy and encrypted public material instead of taking the single-passphrase key-vault redesign. It also preserves watch-only nullability, the explicit per-scope last-account counter, the manager creation time, uint32 account identity, SHA256 address identifiers, and all five legacy address row shapes.

Address usage remains a sticky persisted bit. We do not derive it from surviving UTXOs, since the KV wallet can mark an address used after the relevant transaction history is no longer available. This keeps `LastUnusedAddress` and gap-limit behavior stable across a later migration.

No wallet API, signer interface, PSBT flow, or runtime control path changes here.

Verification:

- SQLite and PostgreSQL migration round trips
- legacy watch-only, scope, account, address, and ciphertext shape tests
- generated sqlc code reproduced cleanly
- full package tests, excluding the environment-dependent local bitcoind startup failure


Roasbeef:
Replaced by #1289 after renaming the head branch to remove the codex prefix. The head SHA and commit structure are unchanged.

PR #1292, opened 2026-07-13
Start SQL Integration Tests

yyforyongyu:
Runs the existing Manager integration test on both backends, so the SQL store is
exercised end to end against a real chain backend rather than only by unit
tests.

Stacked on `impl-runtime-store-6`.

## What lands here

1 commits:

1. `bwtest+itest: select backend, own teardown`
 The harness maps its
`-db` flag to the Manager backend, so the existing `manager` case runs on either
backend without a second test case:

```text
make itest chain=btcd db=kvdb   icase=manager
make itest chain=btcd db=sqlite icase=manager
```

Both pass, each executing exactly one case, `manager_create_wallet`.

## The default stays kvdb

An earlier revision of this PR flipped the global default for newly created
wallets from kvdb to SQLite, which is what the branch name still refers to.
**It does not do that any more** — both commands select the backend explicitly,
and the global default is unchanged. A consistent flip touches the `Makefile`
default and the docs as well as the code, for no capability the itests need.

## Teardown ownership

One cleanup owner stops every registered wallet and *then* closes every Manager,
joining the errors. A Manager owns the database its wallets use, so closing it
first would pull the store out from under a running wallet; `testing.Cleanup`
runs LIFO, which is how the previous duplicate `Stop` inverted that order.

Teardown also checks the on-disk artifact against the requested backend. Without
that check a `db=sqlite` run could pass while the Manager quietly opened kvdb —
which is what made the previous revision's SQLite claim hollow.

`db=postgres` is rejected with a clear message: the harness provisions no DSN.
The PostgreSQL **Store** is retained and green, and the existing shared DB
contract keeps running on PostgreSQL in CI; only the Manager and daemon harness
defer to #321.

### DIFF
--- a/bwtest/harness.go
+++ b/bwtest/harness.go
@@ -178,10 +178,23 @@ func (h *HarnessTest) Subtest(t *testing.T) *HarnessTest {
 func (h *HarnessTest) NewWalletManager() *wallet.Manager {
 	h.Helper()
 
-	// Fixed kvdb until #1292 introduces backend selection: this PR only
-	// moves database ownership to the Manager.
-	//nolint:staticcheck // This boundary intentionally exercises legacy kvdb.
-	backend := wallet.DBBackendKVDB
+	// Switch explicitly: mapping every unknown -db value onto kvdb would let
+	// `db=postgres` pass while exercising kvdb, and the PostgreSQL Manager is
+	// not part of this milestone.
+	var backend wallet.DBBackend
+
+	switch h.dbType {
+	case dbNameKvdb:
+		//nolint:staticcheck // This case intentionally exercises legacy kvdb.
+		backend = wallet.DBBackendKVDB
+
+	case dbNameSQLite:
+		backend = wallet.DBBackendSQLite
+
+	default:
+		h.Fatalf("unsupported -db value %q: the wallet Manager serves "+
+			"kvdb and sqlite", h.dbType)
+	}
 
 	manager, err := wallet.NewManager(h.Context(), wallet.ManagerConfig{
 		Backend:     backend,
--- a/bwtest/harness.go
+++ b/bwtest/harness.go
@@ -3,6 +3,7 @@ package bwtest
 import (
 	"context"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -62,6 +63,10 @@ type HarnessTest struct {
 
 	// dbType is the configured wallet database backend.
 	dbType string
+
+	// managers holds the wallet managers created for this subtest. They are
+	// closed after every registered wallet has stopped.
+	managers []*wallet.Manager
 
 	// mu protects harness state that can be accessed across the main test and
 	// subtests. This includes the wallet registry and idempotent shutdown.
@@ -159,7 +164,7 @@ func (h *HarnessTest) Subtest(t *testing.T) *HarnessTest {
 		// If a test fails, we still try to stop wallets to avoid leaking
 		// goroutines into the next test, but we skip assertions.
 		if st.Failed() {
-			err := st.stopActiveWallets(context.Background())
+			err := st.teardownWallets(context.Background())
 			if err != nil {
 				st.Logf("failed to stop wallets during failed-test cleanup: %v",
 					err)
@@ -172,7 +177,7 @@ func (h *HarnessTest) Subtest(t *testing.T) *HarnessTest {
 			return
 		}
 
-		err := st.stopActiveWallets(context.Background())
+		err := st.teardownWallets(context.Background())
 		require.NoError(st, err, "failed to stop wallets")
 
 		mempool, err := st.getRawMempool()
@@ -220,15 +225,26 @@ func (h *HarnessTest) NewWalletManager() *wallet.Manager {
 	})
 	require.NoError(h, err, "failed to create wallet manager")
 
-	// Temporary: close this Manager directly. #1292 replaces this with the
-	// registry-ordered teardown that stops wallets first.
-	h.Cleanup(func() { _ = manager.Close() })
+	h.managers = append(h.managers, manager)
 
 	// Verify the Manager opened the backend requested by the -db flag without
 	// adding a production accessor solely for the test.
 	h.Cleanup(h.assertBackendArtifact)
 
 	return manager
+}
+
+// teardownWallets stops registered wallets and then closes their Managers.
+func (h *HarnessTest) teardownWallets(ctx context.Context) error {
+	err := h.stopActiveWallets(ctx)
+
+	for _, manager := range h.managers {
+		err = errors.Join(err, manager.Close())
+	}
+
+	h.managers = nil
+
+	return err
 }
 
 // assertBackendArtifact verifies the Manager created the requested database.
@@ -385,7 +401,9 @@ func (h *HarnessTest) Stop() {
 func (h *HarnessTest) stopActiveWallets(ctx context.Context) error {
 	h.Helper()
 
-	for _, w := range h.ActiveWallets() {
+	var stopErr error
+
+	for i, w := range h.ActiveWallets() {
 		if w == nil {
 			// Keep cleanup robust against partially initialized test state. A
 			// caller could register a wallet reference and fail before the
@@ -400,11 +418,13 @@ func (h *HarnessTest) stopActiveWallets(ctx context.Context) error {
 		// legacy fields initialized.
 		err := w.Stop(ctx)
 		if err != nil {
-			return fmt.Errorf("stop wallet: %w", err)
+			stopErr = errors.Join(stopErr, fmt.Errorf(
+				"stop wallet %d: %w", i, err,
+			))
 		}
 	}
 
-	return nil
+	return stopErr
 }
 
 // setUpChainClient creates and starts a chain client for the active harness
--- a/bwtest/harness.go
+++ b/bwtest/harness.go
@@ -2,7 +2,10 @@ package bwtest
 
 import (
 	"context"
+	"encoding/binary"
 	"fmt"
+	"io"
+	"os"
 	"path/filepath"
 	"runtime/debug"
 	"sync"
@@ -15,6 +18,20 @@ import (
 )
 
 const (
+	// sqliteHeaderLen is the length of SQLite's fixed file header string.
+	sqliteHeaderLen = 16
+
+	// sqliteHeaderMagic is the header string every SQLite database starts
+	// with.
+	sqliteHeaderMagic = "SQLite format 3\x00"
+
+	// bboltMagicOffset is the offset of bbolt's magic within the first page
+	// header, bboltMagicLen is its width, and bboltMagic is the value
+	// stored there.
+	bboltMagicOffset = 16
+	bboltMagicLen    = 4
+	bboltMagic       = 0xED0CDAED
+
 	// defaultChainReconnectAttempts is the number of times the chain RPC client
 	// will attempt to reconnect before failing.
 	defaultChainReconnectAttempts = 5
@@ -207,7 +224,72 @@ func (h *HarnessTest) NewWalletManager() *wallet.Manager {
 	// registry-ordered teardown that stops wallets first.
 	h.Cleanup(func() { _ = manager.Close() })
 
+	// Verify the Manager opened the backend requested by the -db flag without
+	// adding a production accessor solely for the test.
+	h.Cleanup(h.assertBackendArtifact)
+
 	return manager
+}
+
+// assertBackendArtifact verifies the Manager created the requested database.
+func (h *HarnessTest) assertBackendArtifact() {
+	h.Helper()
+
+	err := validateBackendArtifact(h.dbType, h.WalletDBPath)
+	require.NoError(h, err)
+}
+
+// validateBackendArtifact verifies that a database file has the requested
+// backend's header.
+func validateBackendArtifact(dbType, dbPath string) error {
+	switch dbType {
+	case dbNameKvdb, dbNameSQLite:
+
+	default:
+		return fmt.Errorf("%w: %s", ErrUnknownDBBackend, dbType)
+	}
+
+	// The path is created and owned by the test harness.
+	f, err := os.Open(dbPath) //nolint:gosec
+	if err != nil {
+		return fmt.Errorf("open wallet database artifact: %w", err)
+	}
+
+	defer func() { _ = f.Close() }()
+
+	// Read enough for both signatures: SQLite's header string, and bbolt's
+	// magic at offset 16 of the first page.
+	header := make([]byte, bboltMagicOffset+bboltMagicLen)
+
+	_, err = io.ReadFull(f, header)
+	if err != nil {
+		return fmt.Errorf("read wallet database header: %w", err)
+	}
+
+	sqlite := string(header[:sqliteHeaderLen]) == sqliteHeaderMagic
+	bbolt := binary.LittleEndian.Uint32(
+		header[bboltMagicOffset:bboltMagicOffset+bboltMagicLen],
+	) == bboltMagic
+
+	var detected string
+	switch {
+	case sqlite:
+		detected = dbNameSQLite
+
+	case bbolt:
+		detected = dbNameKvdb
+
+	default:
+		return fmt.Errorf("unrecognized wallet database artifact: %s",
+			dbPath)
+	}
+
+	if detected != dbType {
+		return fmt.Errorf("db=%s requested but %s is a %s database",
+			dbType, dbPath, detected)
+	}
+
+	return nil
 }
 
 // RegisterWallet registers a wallet with the harness.
--- a/bwtest/harness_test.go
+++ b/bwtest/harness_test.go
@@ -1,0 +1,79 @@
+package bwtest
+
+import (
+	"encoding/binary"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+// TestBackendArtifactValidatesExpectedBackend verifies that backend artifact
+// validation reports positive, mismatched, and malformed headers without
+// depending on testing.T failure state.
+func TestBackendArtifactValidatesExpectedBackend(t *testing.T) {
+	t.Parallel()
+
+	sqliteHeader := make([]byte, bboltMagicOffset+bboltMagicLen)
+	copy(sqliteHeader, sqliteHeaderMagic)
+
+	kvdbHeader := make([]byte, bboltMagicOffset+bboltMagicLen)
+	binary.LittleEndian.PutUint32(
+		kvdbHeader[bboltMagicOffset:], bboltMagic,
+	)
+
+	testCases := []struct {
+		name    string
+		dbType  string
+		header  []byte
+		wantErr string
+	}{
+		{name: "sqlite matches", dbType: dbNameSQLite,
+			header: sqliteHeader},
+		{name: "kvdb matches", dbType: dbNameKvdb,
+			header: kvdbHeader},
+		{
+			name:    "sqlite mismatches kvdb",
+			dbType:  dbNameKvdb,
+			header:  sqliteHeader,
+			wantErr: "is a sqlite database",
+		},
+		{
+			name:    "kvdb mismatches sqlite",
+			dbType:  dbNameSQLite,
+			header:  kvdbHeader,
+			wantErr: "is a kvdb database",
+		},
+		{
+			name:    "unknown header",
+			dbType:  dbNameSQLite,
+			header:  make([]byte, bboltMagicOffset+bboltMagicLen),
+			wantErr: "unrecognized wallet database artifact",
+		},
+		{
+			name:    "truncated header",
+			dbType:  dbNameKvdb,
+			header:  []byte("short"),
+			wantErr: "read wallet database header",
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			dbPath := filepath.Join(t.TempDir(), "wallet.db")
+			err := os.WriteFile(dbPath, testCase.header, 0o600)
+			require.NoError(t, err)
+
+			err = validateBackendArtifact(testCase.dbType, dbPath)
+			if testCase.wantErr == "" {
+				require.NoError(t, err)
+				return
+			}
+
+			require.ErrorContains(t, err, testCase.wantErr)
+		})
+	}
+}
--- a/bwtest/harness_test.go
+++ b/bwtest/harness_test.go
@@ -5,9 +5,103 @@ import (
 	"os"
 	"path/filepath"
 	"testing"
+	"time"
 
+	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/stretchr/testify/require"
 )
+
+// TestTeardownWalletsClosesManagerAfterFailedCreate verifies that centralized
+// teardown releases a Manager even when wallet creation fails.
+func TestTeardownWalletsClosesManagerAfterFailedCreate(t *testing.T) {
+	t.Parallel()
+
+	testCases := []struct {
+		name    string
+		dbType  string
+		backend wallet.DBBackend
+	}{
+		{
+			name:   "kvdb",
+			dbType: dbNameKvdb,
+			//nolint:staticcheck // This case covers legacy kvdb teardown.
+			backend: wallet.DBBackendKVDB,
+		},
+		{
+			name:    "sqlite",
+			dbType:  dbNameSQLite,
+			backend: wallet.DBBackendSQLite,
+		},
+	}
+
+	for _, testCase := range testCases {
+		t.Run(testCase.name, func(t *testing.T) {
+			t.Parallel()
+
+			testManagerTeardownAfterFailedCreate(
+				t, testCase.dbType, testCase.backend,
+			)
+		})
+	}
+}
+
+// testManagerTeardownAfterFailedCreate exercises centralized Manager cleanup
+// for one database backend.
+func testManagerTeardownAfterFailedCreate(t *testing.T, dbType string,
+	backend wallet.DBBackend) {
+
+	t.Helper()
+
+	h := &HarnessTest{
+		T:            t,
+		dbType:       dbType,
+		WalletDBPath: filepath.Join(t.TempDir(), "wallet.db"),
+	}
+
+	manager := h.NewWalletManager()
+	require.NotNil(t, manager)
+
+	// Force Create to fail after the Manager has opened its database.
+	pubPass := []byte("public")
+	cfg := wallet.Config{
+		Name:          "failed-create",
+		PubPassphrase: pubPass,
+	}
+	w, err := manager.Create(cfg, wallet.CreateWalletParams{
+		Mode:              wallet.ModeGenSeed,
+		PubPassphrase:     pubPass,
+		PrivatePassphrase: []byte("private"),
+	})
+	require.ErrorIs(t, err, wallet.ErrMissingParam)
+	require.Nil(t, w)
+
+	// Bound teardown so a missing Manager close fails instead of hanging.
+	done := make(chan error, 1)
+	go func() {
+		done <- h.teardownWallets(t.Context())
+	}()
+
+	select {
+	case err := <-done:
+		require.NoError(t, err)
+
+	case <-time.After(teardownWaitLimit):
+		t.Fatal("teardownWallets did not return: the Manager was " +
+			"probably never closed")
+	}
+
+	// A fresh Manager proves the database was released.
+	reopened, err := wallet.NewManager(t.Context(), wallet.ManagerConfig{
+		Backend:     backend,
+		DataSource:  h.WalletDBPath,
+		ChainParams: h.NetParams(),
+	})
+	require.NoError(t, err, "teardown must release the database")
+	require.NoError(t, reopened.Close())
+}
+
+// teardownWaitLimit bounds teardown so a leaked Manager fails the test.
+const teardownWaitLimit = 30 * time.Second
 
 // TestBackendArtifactValidatesExpectedBackend verifies that backend artifact
 // validation reports positive, mismatched, and malformed headers without
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -1,7 +1,6 @@
 package bwtest
 
 import (
-	"context"
 	"strings"
 	"time"
 
@@ -68,19 +67,11 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
-	// Register so harness-level assertions (MineBlocks and friends) can see
-	// this wallet.
+	// Register before Start, and only register: teardownWallets is the single
+	// cleanup owner. Registering after Start would leave a wallet whose Start
+	// failed unregistered, and a second direct Stop callback here would stop a
+	// successful one twice, out of order with the Manager close.
 	h.RegisterWallet(w)
-
-	// Temporary, until #1292 owns teardown through the registries: stop this
-	// wallet directly. The callback is registered after NewWalletManager's
-	// Close, and testing.Cleanup runs LIFO, so the wallet stops before the
-	// Manager closes the database underneath it.
-	h.Cleanup(func() {
-		// A background context: the test context may already be done by
-		// the time cleanup runs.
-		_ = w.Stop(context.Background())
-	})
 
 	err = w.Start(h.Context())
 	require.NoError(h, err, "failed to start wallet")
--- a/itest/manager_test.go
+++ b/itest/manager_test.go
@@ -3,7 +3,6 @@
 package itest
 
 import (
-	"context"
 	"time"
 
 	"github.com/btcsuite/btcwallet/bwtest"
@@ -37,15 +36,11 @@ func testCreateWallet(h *bwtest.HarnessTest) {
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
-	// Register so harness-level assertions can see this wallet.
+	// Register before Start so teardown owns the wallet even if Start fails.
+	// The harness stops every registered wallet and then closes every
+	// Manager. Do not add another Stop cleanup here; it would only stop the
+	// wallet early and duplicate the harness-owned teardown.
 	h.RegisterWallet(w)
-
-	// Temporary, until #1292 owns teardown through the registries: stop this
-	// wallet directly. Registered after the Manager's Close and run first,
-	// since testing.Cleanup is LIFO.
-	h.Cleanup(func() {
-		_ = w.Stop(context.Background())
-	})
 
 	err = w.Start(h.Context())
 	require.NoError(h, err, "failed to start wallet")
--- a/bwtest/database.go
+++ b/bwtest/database.go
@@ -18,6 +18,9 @@ const (
 	// dbNameKvdb is the identifier used for the kvdb wallet backend.
 	dbNameKvdb = "kvdb"
 
+	// dbNameSQLite is the identifier used for the SQLite wallet backend.
+	dbNameSQLite = "sqlite"
+
 	// kvdbDriver is the walletdb driver name used for kvdb.
 	kvdbDriver = "bdb"
 


bwtest/harness_wallet.go
```diff
@@ -42,6 +42,9 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 		// Use the subtest-scoped DB path and chain client prepared by the
 		// harness.
 		DB: wallet.DBConfig{
+			// Honor the harness-configured backend so db=kvdb parity runs
+			// exercise kvdb rather than silently defaulting to SQLite.
+			Backend: wallet.DBBackend(h.dbType),
```
yyforyongyu: `make itest db=postgres` now selects the Postgres runtime backend here, but the config below only supplies `KVDB.DBPath`, so `Manager.Create` fails validation with missing `DB.Postgres.DSN`; either wire a Postgres DSN/config or reject the unsupported db value before creating the wallet. (`gpt-5.5`)

bwtest/harness_wallet.go
```diff
@@ -42,6 +42,9 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 		// Use the subtest-scoped DB path and chain client prepared by the
 		// harness.
 		DB: wallet.DBConfig{
+			// Honor the harness-configured backend so db=kvdb parity runs
```
yyforyongyu: The harness flag and `make itest` both still default `h.dbType` to `kvdb`, so honoring it here keeps default integration runs and current CI on kvdb instead of exercising SQLite; switch those defaults to SQLite and require explicit `db=kvdb` for parity runs. (`gpt-5.6-sol`)

wallet/signer_test.go
```diff
@@ -1057,6 +1057,67 @@ func TestNewAddressOnSQLOnlyAccount(t *testing.T) {
 	require.NotNil(t, addr)
 }
 
+// TestDerivePubKeySQLOnlyAccount verifies that DerivePubKey resolves a public
```
yyforyongyu: This regression test and its helper wiring cover the SQL-only `DerivePubKey` fallback introduced by `fbc57ec5`, so fold them into that original commit instead of adding them in the SQLite-default commit. (`gpt-5.6-sol`)

wallet/runtime_store_config.go
```diff
@@ -155,15 +155,15 @@ const defaultSQLiteDBName = "wallet.sqlite"
...
-// walletdb store; callers must set Backend to DBBackendSQLite or
-// DBBackendPostgres explicitly to use the SQL runtime store. When the resolved
+// An unset Backend selects DBBackendSQLite so new library-created wallets
+// default to the SQLite runtime store; callers opening an existing kvdb-only
+// wallet must set Backend to DBBackendKVDB explicitly. When the resolved
 // backend is SQLite and no SQLite path was given, a deterministic default is
 // derived next to the legacy kvdb database (DB.KVDB.DBPath) so the runtime
 // store has a stable on-disk location.
 func (c DBConfig) withDefaults() DBConfig {
 	if c.Backend == "" {
```
yyforyongyu: `1f88969a` has no commit body despite changing the default persistence backend and integration harness behavior; add a body of at most 72 columns explaining the migration and explicit kvdb opt-out. (`gpt-5.6-sol`)

wallet/runtime_store_config.go
```diff
@@ -155,15 +155,15 @@ const defaultSQLiteDBName = "wallet.sqlite"
...
-		c.Backend = DBBackendKVDB
+		c.Backend = DBBackendSQLite
```
yyforyongyu: Flipping the unset-`Backend` default from kvdb to SQLite silently switches every existing caller that doesn't set `Backend` (including lnd) to a fresh empty SQLite store derived next to its `wallet.db`, abandoning the kvdb wallet data with no migration; it also isn't needed for the integration tests, since the harness already selects the backend explicitly (`bwtest/harness_wallet.go:45`). Fix: keep the default kvdb and have tests opt into SQLite explicitly, gating any default flip behind shipped migration.

`claude-opus-4-8[1m]`

yyforyongyu:
No issues found by `glm-5.2` :game_die:

Makefile
```diff
@@ -162,7 +162,7 @@ itest-db-race:
 	env CGO_ENABLED=1 GORACE="history_size=7 halt_on_errors=1" $(ITEST_DB_RACE)
 
 #? itest: Run integration tests
-#? itest (vars): chain=btcd|bitcoind|neutrino backend=btcd|bitcoind|neutrino db=kvdb|sqlite|postgres
+#? itest (vars): chain=btcd|bitcoind|neutrino backend=btcd|bitcoind|neutrino db=sqlite|kvdb
```
yyforyongyu: ADR 0008 still documents default btcd+kvdb and a supported `db=postgres` command, contradicting this backend list; update the integration-test documentation with the behavior change. (`gpt-5.6-sol`)

bwtest/harness_wallet.go
```diff
@@ -38,10 +38,33 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
+	// cannot wire. kvdb runs directly off the subtest DB path; sqlite derives
+	// its path from that same kvdb path via the wallet's config defaults.
+	// postgres needs a DSN the harness does not provision, so reject it here
+	// with a clear message rather than failing deep inside Manager.Create.
+	backend := wallet.DBBackend(h.dbType)
+	switch backend {
+	// kvdb and sqlite are both wireable from the subtest DB path.
+	case wallet.DBBackendKVDB, wallet.DBBackendSQLite:
+
+	case wallet.DBBackendPostgres:
```
yyforyongyu: The `8758c4f4` subject is 60 characters, exceeding the project 50-character limit; shorten it and move the rejection detail into the body. (`gpt-5.6-sol`)

itest/main_test.go
```diff
@@ -26,13 +26,14 @@ var (
...
 	//
-	// This flag allows verifying that the wallet functions correctly across all
-	// supported database drivers.
+	// This flag allows verifying that the wallet functions correctly across the
+	// supported database drivers. It defaults to sqlite to match the wallet's
+	// default backend; pass -db=kvdb to run the legacy walletdb parity suite.
 	dbBackend = flag.String(
-		"db", "kvdb",
-		"database backend to use (kvdb, sqlite, postgres)",
+		"db", "sqlite",
```
yyforyongyu: The `db` selection is ignored by the only registered integration case because `testCreateWallet` hard-codes `DBBackendKVDB` and never calls `CreateEmptyWallet`; route this flag into that case so SQLite actually runs. (`gpt-5.6-sol`)

bwtest/harness_wallet.go
```diff
@@ -38,10 +38,33 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 
 	name := "itest-" + strings.ReplaceAll(h.Name(), "/", "_")
 
+	// Resolve the configured backend, rejecting any db value the harness
+	// cannot wire. kvdb runs directly off the subtest DB path; sqlite derives
```
yyforyongyu: The added comments on lines 42, 44, and 63 exceed the mandatory 80-column limit with tabs counted as eight spaces; rewrap them. (`gpt-5.6-sol`)

itest/main_test.go
```diff
@@ -26,13 +26,14 @@ var (
 
 	// dbBackend defines the database backend to be used for the wallet
 	// storage.
-	// Options: "kvdb" (default), "sqlite", "postgres".
+	// Options: "sqlite" (default), "kvdb".
 	//
-	// This flag allows verifying that the wallet functions correctly across all
-	// supported database drivers.
+	// This flag allows verifying that the wallet functions correctly across the
```
yyforyongyu: This added comment block is 84, 82, and 82 columns with tabs counted as eight spaces; rewrap all three lines below the mandatory 80-column limit. (`gpt-5.6-sol`)

wallet/manager_test.go
```diff
@@ -916,10 +916,13 @@ func TestManager_genRootKey(t *testing.T) {
...
 	m := NewManager()
-	cfg := Config{ChainParams: &chainParams}
+	cfg := Config{
+		ChainParams: &chainParams,
+		DB:          DBConfig{Backend: DBBackendKVDB},
+	}
 
-	// With no configured kvdb path there is no legacy wallet to recover, so
-	// genRootKey takes the fresh-generation branch.
+	// With explicit kvdb and no configured path there is no legacy wallet to
```
yyforyongyu: This comment is 81 columns with the leading tab counted as eight spaces; wrap it below the mandatory 80-column limit. (`gpt-5.6-sol`)

bwtest/harness_wallet.go
```diff
@@ -38,10 +38,33 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 
 	name := "itest-" + strings.ReplaceAll(h.Name(), "/", "_")
 
+	// Resolve the configured backend, rejecting any db value the harness
```
yyforyongyu: `8758c4f4` is a separate review-driven follow-up for the backend-default change; fold these harness fixes into the originating commit instead of retaining a fixup commit. (`gpt-5.6-sol`)

wallet/runtime_store_config.go
yyforyongyu: `2ad74219` and `8758c4f4` are substantive backend-default and harness changes with empty commit bodies; add wrapped explanations of what changed and why. (`gpt-5.6-sol`)

wallet/runtime_store_config.go
yyforyongyu: Commit 2ad74219 (`wallet: default new wallets to the sqlite backend`) has an empty commit body; add a body explaining the default flip and its compatibility impact (existing kvdb-only wallets must now set Backend explicitly). `k3`

bwtest/harness_wallet.go
```diff
@@ -38,10 +38,33 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
+	// cannot wire. kvdb runs directly off the subtest DB path; sqlite derives
+	// its path from that same kvdb path via the wallet's config defaults.
+	// postgres needs a DSN the harness does not provision, so reject it here
+	// with a clear message rather than failing deep inside Manager.Create.
+	backend := wallet.DBBackend(h.dbType)
```
yyforyongyu: Commit 8758c4f4 has an empty commit body and a 61-char subject (guideline: <=50); add a body explaining the harness backend resolution and the postgres rejection, and shorten the subject. `k3`

bwtest/harness_wallet.go
yyforyongyu: Addressed in `3d0c50c61`: the harness now resolves the configured backend and rejects unsupported values before creating the wallet. `db=postgres` no longer reaches `Manager.Create` with a missing DSN; instead `harness_wallet.go:55-57` fails with `db=postgres is not supported by the itest harness: no Postgres DSN is provisioned`, and kvdb/sqlite are wired from the subtest DB path.

bwtest/harness_wallet.go
yyforyongyu: Addressed in `3d0c50c61`: both defaults now flip to sqlite. `itest/main_test.go:35-38` defaults the `-db` flag to `sqlite` and `Makefile:178` defaults `db` to `sqlite`, so default itest/CI runs exercise the SQLite backend; `db=kvdb` runs the legacy walletdb parity suite explicitly. The harness honors the resolved backend at `harness_wallet.go:71`.

wallet/runtime_store_config.go
yyforyongyu: Addressed after the history reshape: the backend-default change is now commit `b3250153b` (`wallet: default new wallets to the sqlite backend`), which carries a wrapped body explaining the DBBackendKVDB->DBBackendSQLite flip and the explicit `DBBackendKVDB` opt-out required for existing kvdb-only wallets. The harness/integration changes live in `3d0c50c61`, which also documents its behavior in its body.

wallet/signer_test.go
yyforyongyu: Addressed by the reshape: neither sqlite commit (`b3250153b`, `3d0c50c61`) touches `wallet/signer_test.go`. The SQL-only `DerivePubKey`/`GetPrivKeyForAddress` fallback regression test (`TestGetPrivKeyForAddressSQLDerivedAddress`) and its helpers now live in the store/vault commits `1a71f4172` (`wallet: route signer and passphrase through vault`) and `97229a13d` (`wallet: route GetPrivKeyForAddress through store-derived path`), i.e. the originating (secret/runtime) commits, not the SQLite-default commit.

wallet/runtime_store_config.go
yyforyongyu: By design for this stack. New library-created wallets now default to the SQLite runtime store; callers opening an existing kvdb-only wallet (including lnd) set `Backend: DBBackendKVDB` explicitly to keep using walletdb. This opt-out is now documented in the `b3250153b` commit body and in the `DBConfig`/`withDefaults` godoc (`runtime_store_config.go:57-59,158-163`). Data migration of existing kvdb wallets is intentionally out of scope for this stack. The harness does select its backend explicitly (`3d0c50c61`), but the library default flip is the deliberate behavior change this commit ships.

Makefile
yyforyongyu: Addressed in `3d0c50c61`: `docs/developer/adr/0008-integration-test-framework.md` is updated to match. The default example is now `btcd + sqlite`, the Database Backend section notes the wallet defaults to sqlite (kvdb/postgres supported, but the harness only wires sqlite/kvdb and rejects postgres for lack of a DSN), and the explicit-config example uses `db=kvdb` instead of `db=postgres`.

bwtest/harness_wallet.go
yyforyongyu: Addressed by the reshape: the harness change is now commit `3d0c50c61` with subject `bwtest: default itests to sqlite, reject postgres` (49 chars, <=50). The postgres-rejection and backend-resolution detail is moved into the commit body.

itest/main_test.go
yyforyongyu: Addressed in `3d0c50c61`: `testCreateWallet` (`itest/manager_test.go`) now calls `h.CreateEmptyWallet()`, which routes the wallet's DB backend through the harness-configured `-db` flag (`h.dbType`) instead of hard-coding `DBBackendKVDB`. So `db=sqlite` and `db=kvdb` genuinely exercise their respective backends in the only registered integration case.

bwtest/harness_wallet.go
yyforyongyu: Addressed in `3d0c50c61`: the added/edited comments were rewrapped. All lines in `bwtest/harness_wallet.go` are now <=80 columns with tabs counted as eight spaces (verified: the `defaultPubPass`/`defaultPrivPass` doc, the backend-resolution block at 44-49, and the config comments at 64-70 are all within 80).

itest/main_test.go
yyforyongyu: Addressed in `3d0c50c61`: the `dbBackend` comment block was rewrapped. `itest/main_test.go:31-34` is now 73/74/78/33 columns (tabs=8), all under the 80-column limit.

wallet/manager_test.go
yyforyongyu: The enforced linter (`.golangci.yml` `lll`: line-length 80, tab-width 4) counts this within the 80-column limit, so `make lint` passes. The over-80 reading counts a tab as 8 columns; since tab-width 4 is the gate this repo enforces in CI, the line is left as-is.

bwtest/harness_wallet.go
yyforyongyu: Addressed by the reshape: there is no separate `8758c4f4` review-fixup commit anymore. The harness fixes are folded into the originating harness commit `3d0c50c61` (`bwtest: default itests to sqlite, reject postgres`); the two sqlite commits (`b3250153b`, `3d0c50c61`) are the only sqlite-default changes and no `wip`/fixup commit remains.

wallet/runtime_store_config.go
yyforyongyu: Addressed by the reshape: the two commits now carry wrapped bodies. `b3250153b` (`wallet: default new wallets to the sqlite backend`) explains the DBBackendKVDB->DBBackendSQLite flip and the explicit kvdb opt-out; `3d0c50c61` (`bwtest: default itests to sqlite, reject postgres`) explains the harness backend resolution, postgres rejection, and Makefile/ADR updates. Both bodies wrap at <=72 columns.

wallet/runtime_store_config.go
yyforyongyu: Addressed in `b3250153b` (`wallet: default new wallets to the sqlite backend`): the commit now has a body explaining the default flip from DBBackendKVDB to DBBackendSQLite for new library-created wallets and its compatibility impact (existing kvdb-only wallets must set `DBBackendKVDB` explicitly to keep using walletdb).

bwtest/harness_wallet.go
yyforyongyu: Addressed in `3d0c50c61`: the subject is now `bwtest: default itests to sqlite, reject postgres` (49 chars, <=50), and the commit has a wrapped body explaining the harness backend resolution (kvdb/sqlite from the subtest path) and the postgres rejection (no DSN provisioned), plus the Makefile/ADR updates.

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :sparkles:

yyforyongyu:
No issues found by `k3` 🎉

yyforyongyu:
No issues found by `gpt-5.5` 🫧

itest/manager_test.go
```diff
@@ -3,57 +3,18 @@
...
-		)
-	})
-
-	// Register the wallet so harness helpers can assert global invariants.
-	h.RegisterWallet(w)
+	// Create, start, and register a wallet through the harness helper. This
+	// routes the wallet's DB backend through the harness-configured db flag
+	// (h.dbType), so db=sqlite and db=kvdb genuinely exercise their
+	// respective backends instead of hard-coding one here.
+	h.CreateEmptyWallet()
```
yyforyongyu: Routing this case through `CreateEmptyWallet` masks teardown failures because its later-registered cleanup discards the first `Stop` error and the harness second idempotent stop returns nil; make the helper assert the first error or remove its duplicate cleanup. (`gpt-5.6-sol`)

yyforyongyu: Fixed: removed the duplicate late `t.Cleanup` in `CreateEmptyWallet` that stopped the wallet and discarded the error. The harness `Subtest` cleanup (`stopActiveWallets`, wrapped in `require.NoError`) is now the single teardown owner, so a real first-`Stop` failure is surfaced instead of being masked by the idempotent second stop returning nil.

yyforyongyu:
No issues found by `gpt-5.5` 🛟

bwtest/harness_wallet.go
```diff
@@ -76,13 +104,14 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
-		// canceled by the time cleanup runs.
-		_ = w.Stop(context.Background())
-	})
-
 	// Register the wallet so harness helpers can assert global invariants.
+	//
+	// Registration also hands teardown to the harness: the subtest cleanup
+	// stops every registered wallet via stopActiveWallets and asserts the
+	// error. We deliberately don't register a second Stop cleanup here. A
+	// duplicate cleanup would run first (LIFO) and discard the owning Stop's
```
yyforyongyu: This comment line is 81 columns (limit is 80, counting a tab as 8); rewrap it. `k3`

yyforyongyu: Done — rewrapped the flagged comment (and re-checked the sqlite-added lines) to ≤80 columns at tab=8, folded into its introducing commit.

yyforyongyu:
No issues found by `k3` 🚀

yyforyongyu:
No issues found by `gpt-5.5` 🧭

itest/manager_test.go
```diff
@@ -3,57 +3,18 @@
...
-		// We use a background context here because h.Context() might be
-		// cancelled already.
-		require.NoError(
-			h, w.Stop(context.Background()), "failed to stop wallet",
-		)
-	})
-
-	// Register the wallet so harness helpers can assert global invariants.
-	h.RegisterWallet(w)
+	// Create, start, and register a wallet through the harness helper. This
```
yyforyongyu: Using `CreateEmptyWallet` here contradicts its non-manager godoc and `bwtest/README.md` requirement that manager-focused tests create wallets through the manager API directly; retain direct setup while honoring `h.dbType` or update those contracts. (`gpt-5.6-sol`)

yyforyongyu: Fixed: the manager-focused itest no longer routes through `bwtest.CreateEmptyWallet` (whose godoc + `bwtest/README.md` reserve it as a non-manager helper). It now creates the wallet through the manager API directly while still honoring the harness-selected backend (`h.dbType`), so the SQLite/kvdb selection is preserved without violating the manager-test contract.

yyforyongyu:
No issues found by `gpt-5.5` 🪄

yyforyongyu:
No issues found by `k3` 🎈

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :sunflower:

wallet/runtime_store_config.go
```diff
@@ -157,15 +157,15 @@ const defaultSQLiteDBName = "wallet.sqlite"
...
```
yyforyongyu: Changing the zero-value default to SQLite leaves the inherited `wallet/deprecated.go:7234-7238` and `wallet/runtime_store_config_test.go:22-25` text claiming unset selects KVDB; update those comments and the test name to match this branch. (`gpt-5.6-sol`)

yyforyongyu: Fixed: since this branch flips the zero-value `Backend` default to SQLite, the inherited `deprecated.go` comment and the `runtime_store_config_test.go` text/name (which stated unset → KVDB) are updated to state that an unset `Backend` now defaults to SQLite on this branch.

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :four_leaf_clover:

wallet/deprecated.go
```diff
@@ -7231,9 +7231,9 @@ func OpenWithRetry(db walletdb.DB, pubPass []byte, cbs *waddrmgr.OpenCallbacks,
 		walletDeprecated: deprecated,
 
 		// This deprecated constructor wires the kvdb store directly, so
-		// pin the kvdb backend explicitly. An unset Backend already
-		// resolves to kvdb in this PR, but the backend-dependent legacy
-		// paths (e.g. storeSyncedTo, import mirroring) rely on an
+		// pin the kvdb backend explicitly. An unset Backend now
+		// resolves to sqlite in this PR, and the backend-dependent
+		// legacy paths (e.g. storeSyncedTo, import mirroring) rely on an
```
yyforyongyu: This line is 81 columns (limit is 80, counting a tab as 8), and "in this PR" goes stale once merged; rewrap and drop the PR reference. `k3`

yyforyongyu: Fixed: rewrapped the comment to ≤80 columns (tab=8) and dropped the transient "in this PR" phrasing so it reads correctly post-merge (states the SQLite default directly).

yyforyongyu:
No issues found by `k3` 🎯

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :sparkle:

yyforyongyu:
No issues found by `gpt-5.5` 🪁

itest/manager_test.go
yyforyongyu: Removing the duplicate cleanup makes `stopActiveWallets` the sole teardown path, but it returns on the first `Stop` error and leaves every later registered wallet running; attempt all stops and return the joined errors. (`gpt-5.6-sol`)

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :fireworks:

yyforyongyu:
No issues found by `k3` 🍀

yyforyongyu:
No issues found by `gpt-5.5` 🧵

wallet/runtime_store_config.go
```diff
@@ -0,0 +1,180 @@
...
+
+// withDefaults returns c with the implicit runtime store defaults applied.
+//
+// An unset Backend selects DBBackendSQLite so new library-created wallets
+// default to the SQLite runtime store; callers opening an existing kvdb-only
+// wallet must set Backend to DBBackendKVDB explicitly. When the resolved
+// backend is SQLite and no SQLite path was given, a deterministic default is
+// derived next to the legacy kvdb database (DB.KVDB.DBPath) so the runtime
+// store has a stable on-disk location.
+func (c DBConfig) withDefaults() DBConfig {
```
yyforyongyu: This PR is still based on the pre-restructure impl-runtime-store-6: its head (63691af3) is not an ancestor of the new stack, GitHub reports CONFLICTING, and the PR diff currently shows the entire old 27-commit stack underneath its own 2 commits; rebase onto the new impl-runtime-store-6 (950e3dee) so the diff shrinks back to the two sqlite-default commits. `k3`

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :choo_choo:

wallet/manager.go
```diff
@@ -220,66 +767,371 @@ func (m *Manager) String() string {
...
+		}
+
+		if exists {
+			continue
+		}
+
+		// The account is absent. On an already created wallet this is an
+		// over-create adding a new account, not a partial-import retry to
+		// complete, so reject it instead of mutating the wallet.
+		if preCreated {
```
yyforyongyu: This reintroduces the partial-create wedge that the base fixed: when an earlier attempt created the wallet row and imported only a prefix of `InitialAccounts`, retry sets `preCreated` and rejects the first missing account instead of completing it; keep the durable `create_complete` marker and only reject wallets marked complete. (`gpt-5.5`)

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :compass:

yyforyongyu:
No issues found by `k3` 🌊

wallet/runtime_store_config.go
yyforyongyu: Rebased -- this push puts the branch on the restructured impl-runtime-store-6, so the diff is back to the two sqlite-default commits and GitHub reports the PR mergeable.

wallet/manager.go
yyforyongyu: This was true of the old pre-restructure head (63691af3), but the reshaped branch no longer carries that logic: `preCreated` is gone, the base's durable create-complete marker is kept, only wallets marked complete are rejected with `ErrWalletExists`, and a partial `InitialAccounts` retry completes through the replayable import path (covered by `TestManagerCreateReplaysPartialInitialAccounts` and `TestManagerCreateRetryRequiresExactManifest` at the base). No code change needed on this branch.

yyforyongyu:
No issues found by `k3` 🪙

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :telescope:

yyforyongyu:
No issues found by `gpt-5.5` 🦊

yyforyongyu:
No issues found by `k3` ⚡ (head unchanged at d2717909 since the last review; re-verified, still mergeable)

itest/manager_test.go
yyforyongyu: Still unresolved at the current head: `stopActiveWallets` returns immediately on the first `Stop` error, so later registered wallets are never stopped; attempt every stop and return the joined errors. (`gpt-5.6-sol`)

yyforyongyu: Fixed. `stopActiveWallets` now collects each `Stop` failure into a slice and returns `errors.Join(stopErrs...)`, so every registered wallet is stopped even after one fails rather than leaving later wallets running with their stores open. Folded into `bwtest: default itests to sqlite, reject postgres`, the commit that made this the sole teardown owner.

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :sunrise:

bwtest/harness.go
```diff
@@ -275,13 +278,20 @@ func (h *HarnessTest) stopActiveWallets(ctx context.Context) error {
 		// NOTE: We intentionally don't call the deprecated WaitForShutdown/
 		// ShuttingDown methods here, as modern wallets might not have the
 		// legacy fields initialized.
+		//
+		// Every wallet is stopped even after a failure: this is the sole
```
yyforyongyu: This comment line is 81 columns (limit is 80, counting a tab as 8); rewrap it. `k3`

yyforyongyu: Fixed -- the comment is rewrapped under 80 columns at tab=8, folded into `bwtest: default itests to sqlite, reject postgres`. It was one of 10 overflows my round-2 changes introduced; I audited every added line at tab=8 (CI's `lll` uses `tab-width: 4` and cannot see them) and wrapped all of them, with a full-stack re-audit reporting zero remaining.

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :tulip:

yyforyongyu:
No issues found by `k3` 🦄

yyforyongyu:
No issues found by `gpt-5.5` 🦉

yyforyongyu:
No issues found by `gpt-5.6-sol` on re-review. 🔭

yyforyongyu:
No issues found by `glm-5.2` on re-review of the updated PR. :rainbow:

yyforyongyu:
No issues found by `k3` 🛰️

yyforyongyu:
No issues found by `gpt-5.6-sol` on re-review. 🛰️

yyforyongyu:
No issues found by `k3` 🏔️

bwtest/harness.go
```diff
@@ -275,13 +278,21 @@ func (h *HarnessTest) stopActiveWallets(ctx context.Context) error {
...
 		if err != nil {
-			return fmt.Errorf("stop wallet: %w", err)
+			stopErrs = append(
+				stopErrs, fmt.Errorf("stop wallet: %w", err),
+			)
 		}
 	}
 
-	return nil
+	return errors.Join(stopErrs...)
```
yyforyongyu: Commit `77724271f` describes only backend selection and documentation, but this hunk also changes teardown ownership and aggregates every `Stop` error while the PR description still says “Single commit” although the PR owns two; update both descriptions. (`gpt-5.6-sol`)

Makefile
yyforyongyu: This added Make help line is 91 columns, exceeding the repository’s 80-column limit; split the chain/backend and database options across lines. (`gpt-5.6-sol`)

docs/developer/adr/0008-integration-test-framework.md
```diff
@@ -63,7 +63,7 @@ graph TD
     *   For `bitcoind` tests: A separate `bitcoind` process connects (P2P) to the Miner to sync.
     *   For `neutrino` tests: There is no separate "Chain Node" process; Neutrino connects directly to the Miner.
 *   **`ChainBackend`**: The interface wrapper (`rpcclient` or `neutrino.ChainService`) used by the Wallet to communicate with the Chain Node.
-*   **`Database Backend`**: The storage layer (`kvdb`, `postgres`, `sqlite`).
+*   **`Database Backend`**: The storage layer. The wallet supports `sqlite` (the default), `kvdb`, and `postgres`, but the itest harness currently only wires `sqlite` and `kvdb`; `postgres` is rejected because no DSN is provisioned.
```
yyforyongyu: This added 232-column paragraph violates the 80-column documentation rule and, by saying PostgreSQL is rejected, leaves line 119’s promise to validate the “entire support matrix” false; wrap it and scope coverage to SQLite/kvdb until PostgreSQL is provisioned. (`gpt-5.6-sol`)

bwtest/harness.go
yyforyongyu: Fixed in `dfb34974f`. Both descriptions were understating the commit.

The message now has a second paragraph for the teardown change: `stopActiveWallets` owns the only teardown pass, so it stops every wallet and joins the errors instead of returning on the first one and leaking the rest. The PR description no longer says "Single commit" and names both.

Makefile
yyforyongyu: Fixed in `dfb34974f` — split as suggested, into a chain/backend line (76 columns) and a `db=` line (31). Worth noting it was 103 columns before this branch touched it, so shortening it left a pre-existing violation in place rather than introducing one; either way it is under 80 now.

docs/developer/adr/0008-integration-test-framework.md
yyforyongyu: Fixed in `dfb34974f`. The bullet is wrapped to four lines, and the Coverage promise is now scoped: the chain-backend matrix, plus `sqlite` and `kvdb` for the database matrix until a PostgreSQL DSN is provisioned. Good catch that my change is what falsified it.

Left the rest of the document alone — it is long-lined at base (line 5 is 270 columns, and most of section 2.2 runs 96-141), so reflowing it here would bury this commit in unrelated churn.

yyforyongyu:
No issues found by `k3` 🌋

yyforyongyu:
No issues found by `k3` 🐚 (head unchanged at dfb34974 since the last review; re-verified, still mergeable)

yyforyongyu:
No issues found by `k3` 🚢

yyforyongyu:
No issues found by `gpt-5.6-sol` on final re-review. 🧭

yyforyongyu:
No issues found by `k3` 🏺

bwtest/harness_wallet.go
yyforyongyu: The SQLite default in `Makefile` still does not affect CI because every daemon itest explicitly passes `db=kvdb` at `.github/workflows/main.yml:241`, `:287`, and `:356`, so the new SQLite harness path remains untested; run at least one CI itest with `db=sqlite` or use a backend matrix. (`gpt-5.6-sol`)

yyforyongyu: Correct that CI does not exercise the SQLite harness path at this head — all three daemon itest jobs pin `db=kvdb` at the lines you cite. Not fixing it here, though, because it is already fixed one PR up: #1300 (based on `wallet-default-sqlite`) adds `db: [kvdb, sqlite]` to the btcd, neutrino, and bitcoind jobs and removes the "kvdb only ... Database expansion is planned via a separate matrix" comment — `854cd4a79`, `ci: run itest over sqlite and kvdb`.

This PR touches no workflow file at all, so adding a competing matrix here would duplicate that commit and conflict with the PR stacked directly on top of this one. I did verify the path itself rather than assume it: `make itest chain=btcd db=sqlite` passes locally, so the matrix lands green rather than red.

yyforyongyu:
No issues found by `k3` 🌵

yyforyongyu:
No issues found by `gpt-5` 🐙

wallet/runtime_store_config.go
yyforyongyu: Defaulting to SQLite here makes the fixed `wallet.sqlite` fallback the zero-value path, so Manager wallets whose legacy paths share a directory silently open the same file even though `openRuntimeStore` requires one wallet and pool per file; derive a per-wallet path or add a shared registry before making this the default. (`gpt-5`)

bwtest/harness_wallet.go
```diff
@@ -60,27 +97,34 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
 	manager := wallet.NewManager()
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
+	// Register before starting. Create has already opened the wallet's
+	// store, so registering after Start would leave that store unowned
+	// whenever Start fails: require.NoError exits the test immediately,
+	// stopActiveWallets never sees the wallet, and the leaked handle
+	// poisons later subtests. Stop on a never-started wallet logs and
+	// returns nil, so registering first is always safe.
```
yyforyongyu: The claim that Stop on a never-started wallet "logs and returns nil" is inaccurate for manager-built wallets: Wallet.Stop takes the stopClaimUnstarted branch (wallet/controller.go), which cancels the load-time context and runs the real unstarted teardown via finishStopAsync — closing the stores buildWallet opened, evicting the manager cache entry, and returning any store-close error. Only wallets built by the deprecated OpenWithRetry constructor get a true no-op pre-Start Stop (uninitialized walletState). The register-first conclusion still holds — Stop performing that teardown is exactly what makes registration necessary — but the justification should describe the actual behavior so future readers do not rely on Stop being a no-op here.

(`kimi`)

bwtest/harness_wallet.go
```diff
@@ -60,27 +97,34 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
+	//
+	// Registration also hands teardown to the harness: the subtest cleanup
+	// stops every registered wallet via stopActiveWallets and asserts the
+	// error. We deliberately don't register a second Stop cleanup here. A
+	// duplicate cleanup would run first (LIFO) and discard the owning
+	// Stop's error, and the harness's later Stop would return nil
```
yyforyongyu: This masking rationale is incorrect: a second Stop on an already-stopped wallet does not return nil. Stop resolves to stopClaimNone and fast-paths through waitForStopped to teardownResult(), which returns the store-close error recorded by finishStop — stopResult is stored before the Stopped lifecycle transition (wallet/controller.go). So the harness teardown would still observe and assert the failure even if a duplicate cleanup had run first; nothing is masked by idempotency. The real hazard of a duplicate `_ = w.Stop(...)` cleanup is only that that cleanup itself discards the error. Please reword the justification to match the actual Stop contract.

(`kimi`)

itest/manager_test.go
```diff
@@ -30,28 +33,28 @@ func testCreateWallet(h *bwtest.HarnessTest) {
...
 	}
 
+	manager := wallet.NewManager()
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
+	// Register before starting: Create has already opened the wallet's
+	// store, so a failed Start would exit the test with that store unowned
+	// and leak it past stopActiveWallets. Stop on a never-started wallet
+	// logs and returns nil.
```
yyforyongyu: This repeats the same inaccurate claim as the harness helper: Stop on a never-started, manager-built wallet is not a log-and-return-nil no-op. It takes the stopClaimUnstarted branch (wallet/controller.go) and runs the real unstarted teardown — closing the store Create opened and returning any store-close error. Only deprecated OpenWithRetry wallets no-op. Please correct the comment so the justification matches the actual Stop semantics.

(`kimi`)

bwtest/harness_wallet.go
```diff
@@ -69,6 +69,20 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
 
+	// Register before starting, so a wallet whose Start fails is still torn
+	// down: require.NoError exits the test immediately, and a wallet the
+	// harness never saw keeps its workers and its vault state past the
+	// subtest. Stop on a never-started wallet fails the Started->Stopping
+	// transition, logs, and returns nil, so registering first is safe.
+	//
+	// Registration also hands teardown to the harness: the subtest cleanup
+	// stops every registered wallet via stopActiveWallets and asserts the
+	// error. We deliberately don't register a second Stop cleanup here. A
```
yyforyongyu: This says no second Stop cleanup is registered, but the `h.Cleanup` immediately below still calls `w.Stop` and discards its error before the harness cleanup runs; remove that duplicate cleanup so the asserted harness stop owns teardown. (`gpt-5`)

itest/manager_test.go
```diff
@@ -40,6 +40,15 @@ func testCreateWallet(h *bwtest.HarnessTest) {
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
+	// Register before starting, so a wallet whose Start fails is still torn
+	// down rather than left running past stopActiveWallets. Stop on a
+	// never-started wallet logs and returns nil.
+	//
+	// Registration also hands teardown to the harness, which stops every
+	// registered wallet during subtest cleanup, so we don't add a Stop
```
yyforyongyu: This says the harness is the sole Stop owner, but the cleanup immediately below still stops the wallet first, so remove the duplicate cleanup and let the harness surface the first teardown error. (`gpt-5`)

bwtest/harness_wallet.go
```diff
@@ -69,6 +69,20 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
+	// Register before starting, so a wallet whose Start fails is still torn
+	// down: require.NoError exits the test immediately, and a wallet the
+	// harness never saw keeps its workers and its vault state past the
+	// subtest. Stop on a never-started wallet fails the Started->Stopping
+	// transition, logs, and returns nil, so registering first is safe.
+	//
+	// Registration also hands teardown to the harness: the subtest cleanup
```
yyforyongyu: Registration only gives the harness a `Wallet.Stop`, while the local manager owns the database and is discarded without `Manager.Close`, so this still leaks the database handle; retain the manager and close it after registered wallets stop. (`gpt-5`)

bwtest/harness_wallet.go
```diff
@@ -69,6 +69,20 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
+	// Register before starting, so a wallet whose Start fails is still torn
```
yyforyongyu: The subject of the head commit `672b9476` (`bwtest+itest: correct the wallet-registration rationale`) is 55 characters, exceeding the 50-character limit in `docs/developer/contribution_guidelines.md:231` ("Short (50 chars or less) summary of changes"). Shorten the subject (e.g. `bwtest+itest: fix wallet-registration rationale`, 48 chars) and move any dropped detail into the body. (`kimi`)

bwtest/harness_wallet.go
yyforyongyu: Fixed at `82bec2a02`. The harness is now the single cleanup owner and the rationale matches the code:

- `CreateEmptyWallet` registers the wallet **before** `Start` and adds no Stop callback of its own, so a wallet whose `Start` fails is still owned and a successful one is stopped exactly once;
- `teardownWallets` stops every registered wallet with joined, contextualised errors — it no longer returns on the first failure while later wallets keep running — and only then closes every Manager it handed out;
- the deferred-stop protocol the old rationale described (`stopClaim*`, terminal `stopped`, cache eviction) does not exist any more: `Stop` is base-shaped plus a vault lock, so there is no "never-started" special case left to describe.

bwtest/harness_wallet.go
yyforyongyu: Fixed at `82bec2a02`. The harness is now the single cleanup owner and the rationale matches the code:

- `CreateEmptyWallet` registers the wallet **before** `Start` and adds no Stop callback of its own, so a wallet whose `Start` fails is still owned and a successful one is stopped exactly once;
- `teardownWallets` stops every registered wallet with joined, contextualised errors — it no longer returns on the first failure while later wallets keep running — and only then closes every Manager it handed out;
- the deferred-stop protocol the old rationale described (`stopClaim*`, terminal `stopped`, cache eviction) does not exist any more: `Stop` is base-shaped plus a vault lock, so there is no "never-started" special case left to describe.

itest/manager_test.go
yyforyongyu: Fixed at `82bec2a02`. The harness is now the single cleanup owner and the rationale matches the code:

- `CreateEmptyWallet` registers the wallet **before** `Start` and adds no Stop callback of its own, so a wallet whose `Start` fails is still owned and a successful one is stopped exactly once;
- `teardownWallets` stops every registered wallet with joined, contextualised errors — it no longer returns on the first failure while later wallets keep running — and only then closes every Manager it handed out;
- the deferred-stop protocol the old rationale described (`stopClaim*`, terminal `stopped`, cache eviction) does not exist any more: `Stop` is base-shaped plus a vault lock, so there is no "never-started" special case left to describe.

bwtest/harness_wallet.go
yyforyongyu: Fixed at `82bec2a02`. The harness is now the single cleanup owner and the rationale matches the code:

- `CreateEmptyWallet` registers the wallet **before** `Start` and adds no Stop callback of its own, so a wallet whose `Start` fails is still owned and a successful one is stopped exactly once;
- `teardownWallets` stops every registered wallet with joined, contextualised errors — it no longer returns on the first failure while later wallets keep running — and only then closes every Manager it handed out;
- the deferred-stop protocol the old rationale described (`stopClaim*`, terminal `stopped`, cache eviction) does not exist any more: `Stop` is base-shaped plus a vault lock, so there is no "never-started" special case left to describe.

wallet/runtime_store_config.go
yyforyongyu: Fixed. There is no zero-value backend path any more: `ManagerConfig.Backend` and `DataSource` are both required and an unset backend returns `ErrMissingParam`, so nothing falls back to a fixed `wallet.sqlite` name. What #1292 changes is only the *caller-facing default* — `bwtest` selects SQLite when `-db` is unset — and the harness proves which backend actually opened by checking the on-disk artifact against `h.dbType`, with a positive signature for both (SQLite's header string, bbolt's magic).

bwtest/harness_wallet.go
yyforyongyu: Fixed at `82bec2a02`. The harness is now the single cleanup owner and the rationale matches the code:

- `CreateEmptyWallet` registers the wallet **before** `Start` and adds no Stop callback of its own, so a wallet whose `Start` fails is still owned and a successful one is stopped exactly once;
- `teardownWallets` stops every registered wallet with joined, contextualised errors — it no longer returns on the first failure while later wallets keep running — and only then closes every Manager it handed out;
- the deferred-stop protocol the old rationale described (`stopClaim*`, terminal `stopped`, cache eviction) does not exist any more: `Stop` is base-shaped plus a vault lock, so there is no "never-started" special case left to describe.

itest/manager_test.go
yyforyongyu: Fixed at `82bec2a02`. The harness is now the single cleanup owner and the rationale matches the code:

- `CreateEmptyWallet` registers the wallet **before** `Start` and adds no Stop callback of its own, so a wallet whose `Start` fails is still owned and a successful one is stopped exactly once;
- `teardownWallets` stops every registered wallet with joined, contextualised errors — it no longer returns on the first failure while later wallets keep running — and only then closes every Manager it handed out;
- the deferred-stop protocol the old rationale described (`stopClaim*`, terminal `stopped`, cache eviction) does not exist any more: `Stop` is base-shaped plus a vault lock, so there is no "never-started" special case left to describe.

bwtest/harness_wallet.go
yyforyongyu: Fixed at `82bec2a02`. The harness is now the single cleanup owner and the rationale matches the code:

- `CreateEmptyWallet` registers the wallet **before** `Start` and adds no Stop callback of its own, so a wallet whose `Start` fails is still owned and a successful one is stopped exactly once;
- `teardownWallets` stops every registered wallet with joined, contextualised errors — it no longer returns on the first failure while later wallets keep running — and only then closes every Manager it handed out;
- the deferred-stop protocol the old rationale described (`stopClaim*`, terminal `stopped`, cache eviction) does not exist any more: `Stop` is base-shaped plus a vault lock, so there is no "never-started" special case left to describe.

bwtest/harness_wallet.go
yyforyongyu: Fixed. All 14 subjects in the rebuilt stack are ≤50 columns; this one is now `bwtest+itest: select backend, own teardown` (42). Body lines are all ≤72.

bwtest/harness.go
```diff
@@ -178,10 +201,23 @@ func (h *HarnessTest) Subtest(t *testing.T) *HarnessTest {
...
+	switch h.dbType {
+	case "kvdb":
+		//nolint:staticcheck // This case intentionally exercises legacy kvdb.
+		backend = wallet.DBBackendKVDB
+
+	case "sqlite":
+		backend = wallet.DBBackendSQLite
+
+	default:
+		h.Fatalf("unsupported -db value %q: the wallet Manager serves "+
```
yyforyongyu, line 217: This rejects `db=postgres` while `Makefile:165-178` and `itest/main_test.go:29-35` still advertise it as a supported itest backend, so update those entry points and docs to reject it consistently or provision the PostgreSQL Manager configuration. (`gpt-5`)

bwtest/harness_wallet.go
```diff
@@ -68,20 +66,12 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
-	// Register so harness-level assertions (MineBlocks and friends) can see
-	// this wallet.
+	// Register before Start, and only register: teardownWallets is the single
```
yyforyongyu, line 70: Because this wallet can now be SQLite-backed, `CreateFundedWallet` reaches `NewAddress` at line 94 with no default account—the SQLite Manager create path persists only wallet metadata and secrets—so seed the default account before derivation or reject this helper on SQLite. (`gpt-5`)

bwtest/harness.go
yyforyongyu: This is deliberately out of scope for #1292. This PR is limited to proving the existing manager itest with the kvdb and SQLite Manager backends; PostgreSQL Manager configuration and provisioning are explicitly deferred. The Makefile help and `itest/main_test.go` flag text both predate this commit and are unchanged by it. The explicit rejection here is still necessary: accepting `db=postgres` without a PostgreSQL Manager would produce a false-positive run that actually exercises kvdb. Aligning the repository-wide entry points and documentation, or adding PostgreSQL Manager support, belongs in a focused follow-up rather than expanding this one-commit cutover.

bwtest/harness_wallet.go
yyforyongyu: This is a valid gap for future SQLite itest expansion, but it is not part of this PR. The scoped first SQLite case is `testCreateWallet`; it calls `NewWalletManager` directly and never reaches `CreateFundedWallet`. The helper has no current call sites, and its own documentation says it is intended for future integration tests. Seeding a default account in the harness would also mask wallet-library creation behavior that should be addressed when the next SQLite itest actually needs derivation. Rejecting the unused helper here would likewise add unrelated behavior. Deferring this to that focused follow-up keeps #1292 limited to the already-built first itest.

yyforyongyu:
No issues found by `claude-opus-5[1m]` 🧭

yyforyongyu:
No issues found by gpt-5 🌱 (`gpt-5`)

itest/manager_test.go
```diff
@@ -37,16 +35,13 @@ func testCreateWallet(h *bwtest.HarnessTest) {
 	w, err := manager.Create(cfg, params)
 	require.NoError(h, err, "failed to create wallet")
 
-	// Register so harness-level assertions can see this wallet.
+	// Register before Start so teardown owns the wallet even if Start fails.
+	// The harness stops every registered wallet and then closes every
+	// Manager, so this test adds no Stop of its own: testing.Cleanup runs
+	// LIFO, and a second registration here would close the database out from
```
yyforyongyu: The LIFO mechanism described here does not produce the named hazard. A second Stop cleanup registered in the test body is registered after, and therefore runs before, the harness teardown: the wallet would be stopped early and then stopped again by teardownWallets, not have its database closed out from under it (that would require the extra cleanup to run after the Manager close). And since Stop stores its teardown result, the harness teardown would still surface a first-Stop failure. Please reword the rationale to the actual hazard: a redundant stop outside the single teardown owner. (`kimi`)

yyforyongyu: Fixed in the rewritten #1292 stack. The rationale now states the actual hazard: another cleanup would stop the wallet early and duplicate the harness-owned teardown, rather than close the database underneath it.

yyforyongyu:
No issues found by gpt-5 🍀 (`gpt-5`)

yyforyongyu:
No issues found by `claude-opus-5[1m]` 🫧

yyforyongyu:
No issues found by `kimi` 🎉

bwtest/harness.go
```diff
@@ -178,10 +201,23 @@ func (h *HarnessTest) Subtest(t *testing.T) *HarnessTest {
...
-	// moves database ownership to the Manager.
-	//nolint:staticcheck // This boundary intentionally exercises legacy kvdb.
-	backend := wallet.DBBackendKVDB
+	// Switch explicitly: mapping every unknown -db value onto kvdb would let
+	// `db=postgres` pass while exercising kvdb, and the PostgreSQL Manager is
+	// not part of this milestone.
+	var backend wallet.DBBackend
+
+	switch h.dbType {
+	case "kvdb":
```
Abdulkbk: nit: maybe define  dbNameKvdb = "kvdb" and dbNameSQLite and use them here and down below in `assertBackendArtifact`

bwtest/harness.go
```diff
@@ -190,13 +226,83 @@ func (h *HarnessTest) NewWalletManager() *wallet.Manager {
...
+
+// assertBackendArtifact checks that the file the Manager created matches the
+// backend the -db flag asked for. Both backends are proven positively: SQLite
+// writes a fixed 16-byte header string, bbolt writes its own magic in the page
+// header. Asserting only "not SQLite" would pass on an empty or truncated file.
+//
+// The expectation is derived from h.dbType rather than from the value handed to
+// the Manager, so a mapping bug that opens the wrong backend is caught instead
+// of being confirmed by its own mistake.
+func (h *HarnessTest) assertBackendArtifact() {
```
Abdulkbk, line 251: We should extract the detection logic into a function we can unit-test against mismatched or negative cases without causing the test runner to fail. Then `assertBackendArtifact` can just wrap it.

bwtest/harness_test.go
```diff
@@ -0,0 +1,79 @@
...
+// This is the harness-level counterpart to the Manager unit test: it exercises
+// registration and teardownWallets rather than calling Close directly. The
+// subject is harness ownership, so Create is failed as early and as cheaply as
+// possible, with a config that omits a required parameter.
+func TestTeardownWalletsClosesManagerAfterFailedCreate(t *testing.T) {
+	t.Parallel()
+
+	h := &HarnessTest{
+		T:            t,
+		dbType:       "kvdb",
```
Abdulkbk: We should change this to test for SQLite as well. I think we can acheive that with a table-driven.

Abdulkbk (changes requested):
Reviewed and left few suggestions.

bwtest/harness.go
yyforyongyu: Fixed in 4e1458b9da. I reused the existing `dbNameKvdb` constant, added `dbNameSQLite`, and now use both in the Manager mapping and artifact validation.

bwtest/harness.go
yyforyongyu: Fixed in e7b1decd1e. Artifact validation is now a standalone error-returning function, and table-driven unit coverage exercises matching, mismatched, unknown, and truncated headers without intentionally failing `testing.T`.

bwtest/harness_test.go
yyforyongyu: Fixed in d1023283af. The failed-create teardown test is table-driven across kvdb and SQLite and proves each database can be reopened after centralized Manager cleanup.

Abdulkbk (approved):
LGTM

Issue #1250, opened 2026-05-28
wallet: add public API integration test foundation


yyforyongyu:
## Background

After the SQL store series lands, `wallet/` will have many public APIs routed
through the new store stack. Current coverage is strong at the store and unit
levels, but we are missing a true integration-test foundation that boots a real
wallet with a real SQL backend and exercises the public wallet APIs end-to-end.

This issue tracks adding baseline contract coverage for every non-legacy public
`wallet/` API, then building broader scenario tests on top of that foundation.

## Goals

Add representative happy-path, unhappy-path, and boundary coverage for each
non-legacy public `wallet/` API.

Run these tests against real SQL-backed wallets, starting with sqlite and
extending to postgres where practical.

For APIs that mutate durable state, verify behavior after wallet reload.

Establish a foundation for larger scenario-based integration tests.

## Scope

The primary public components are:

- `Manager`: `Create`, `Load`
- `Controller`: `Start`, `Stop`, `Unlock`, `Lock`, `ChangePassphrase`, `Info`,
  `Resync`, `Rescan`
- `AccountManager`: `NewAccount`, `ListAccounts`, `ListAccountsByScope`,
  `ListAccountsByName`, `GetAccount`, `RenameAccount`, `ImportAccount`
- `AddressManager`: `NewAddress`, `GetUnusedAddress`, `GetAddressInfo`,
  `ListAddresses`, `ImportPublicKey`, `ImportTaprootScript`, `ScriptForOutput`,
  `GetDerivationInfo`
- `UtxoManager`: `ListUnspent`, `GetUtxo`, `LeaseOutput`, `ReleaseOutput`,
  `ListLeasedOutputs`
- `TxCreator`: `CreateTransaction`
- `TxPublisher`: `CheckMempoolAcceptance`, `Broadcast`
- `TxReader`: `GetTx`, `ListTxns`
- `TxWriter`: `LabelTx`
- `PsbtManager`: `DecorateInputs`, `FundPsbt`, `SignPsbt`, `FinalizePsbt`,
  `CombinePsbt`
- `Signer`: `DerivePubKey`, `ECDH`, `SignDigest`, `ComputeUnlockingScript`,
  `ComputeRawSig`
- `UnsafeSigner`: `DerivePrivKey`, `GetPrivKeyForAddress`

## Exclusions

This issue should not target legacy APIs:

- `wallet.Interface`
- methods in `deprecated.go`
- methods suffixed `Deprecated`
[...]

PR #1300, opened 2026-07-27
wallet+itest+bwtest: add Manager and Controller itests

GustavoStingelin:
First slice of the public-API integration-test foundation (#1250): boots a
real SQL-backed wallet on the bwtest harness and exercises the two
lowest-level components end-to-end.

- Make SQLite the default integration-test wallet backend.
- Run integration tests against both SQLite and KVDB in CI.
- Add shared wallet configuration and backend lifecycle helpers.
- Add Manager tests for wallet creation, duplicate creation, loading, and
  missing wallets.
- Add Controller lifecycle tests.
- Normalize missing synced-block errors across SQLite and PostgreSQL.
- Handle missing historical rollback data during wallet reload.
- Add unit and integration regression coverage for rollback behavior.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -211,12 +211,15 @@ jobs:
   ########################
   # Run bwtest integration tests against each supported chain backend.
   #
-  # These jobs currently use kvdb only. Database expansion is planned via a
-  # separate matrix in a future change.
+  # Each job runs over both wallet database backends (kvdb and sqlite).
   ########################
   itest-btcd:
-    name: itest btcd
+    name: itest btcd (${{ matrix.db }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        db: [kvdb, sqlite]
     steps:
       - name: git checkout
         uses: actions/checkout@v5
@@ -240,14 +243,14 @@ jobs:
 
       # The btcd backend job runs btcd for both the shared miner and the chain
       # backend under test.
-      - name: run itest (btcd, kvdb)
-        run: make itest chain=btcd db=kvdb
+      - name: run itest (btcd, ${{ matrix.db }})
+        run: make itest chain=btcd db=${{ matrix.db }}
 
       - name: upload itest logs
         if: ${{ failure() }}
         uses: actions/upload-artifact@v4
         with:
-          name: itest-logs-btcd
+          name: itest-logs-btcd-${{ matrix.db }}
           path: itest/test-logs
           retention-days: 5
 
@@ -257,12 +260,16 @@ jobs:
   # Job flow:
   # - Install btcd `${{ env.BTCD_VERSION_LATEST }}`.
   # - Use btcd as the shared miner for the suite.
-  # - Run `make itest chain=neutrino db=kvdb` so wallets use the in-process
-  #   neutrino backend while blocks/peers come from btcd.
+  # - Run `make itest chain=neutrino db=${{ matrix.db }}` so wallets use the
+  #   in-process neutrino backend while blocks/peers come from btcd.
   ########################
   itest-neutrino:
-    name: itest neutrino
+    name: itest neutrino (${{ matrix.db }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        db: [kvdb, sqlite]
     steps:
       - name: git checkout
         uses: actions/checkout@v5
@@ -286,14 +293,14 @@ jobs:
 
       # Neutrino runs in-process, but it still relies on the shared btcd miner
       # for chain data and peer connectivity.
-      - name: run itest (neutrino, kvdb)
-        run: make itest chain=neutrino db=kvdb
+      - name: run itest (neutrino, ${{ matrix.db }})
+        run: make itest chain=neutrino db=${{ matrix.db }}
 
       - name: upload itest logs
         if: ${{ failure() }}
         uses: actions/upload-artifact@v4
         with:
-          name: itest-logs-neutrino
+          name: itest-logs-neutrino-${{ matrix.db }}
           path: itest/test-logs
           retention-days: 5
 
@@ -304,15 +311,16 @@ jobs:
   # - Install btcd `${{ env.BTCD_VERSION_LATEST }}` as the shared miner.
   # - Run a matrix over bitcoind versions `30` (latest) and `28` (older).
   # - Extract each matrix binary from the docker image and run
-  #   `make itest chain=bitcoind db=kvdb`.
+  #   `make itest chain=bitcoind db=${{ matrix.db }}`.
   ########################
   itest-bitcoind:
-    name: itest bitcoind (v${{ matrix.bitcoind_version }})
+    name: itest bitcoind (v${{ matrix.bitcoind_version }}, ${{ matrix.db }})
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         bitcoind_version: ['30', '28']
+        db: [kvdb, sqlite]
     steps:
       - name: git checkout
         uses: actions/checkout@v5
@@ -355,14 +363,14 @@ jobs:
           docker rm $CONTAINER_ID
           bitcoind --version
 
-      - name: run itest (bitcoind, kvdb)
-        run: make itest chain=bitcoind db=kvdb
+      - name: run itest (bitcoind, ${{ matrix.db }})
+        run: make itest chain=bitcoind db=${{ matrix.db }}
 
       - name: upload itest logs
         if: ${{ failure() }}
         uses: actions/upload-artifact@v4
         with:
-          name: itest-logs-bitcoind-${{ matrix.bitcoind_version }}
+          name: itest-logs-bitcoind-${{ matrix.bitcoind_version }}-${{ matrix.db }}
           path: itest/test-logs
           retention-days: 5
 
--- a/bwtest/harness.go
+++ b/bwtest/harness.go
@@ -324,6 +324,57 @@ func (h *HarnessTest) RegisterWallet(w *wallet.Wallet) {
 	h.mu.Unlock()
 }
 
+// DeregisterWallet releases a wallet from harness ownership.
+//
+// It returns false when the wallet is nil or was not registered. The remaining
+// registrations retain their original order.
+func (h *HarnessTest) DeregisterWallet(w *wallet.Wallet) bool {
+	h.Helper()
+
+	if w == nil {
+		return false
+	}
+
+	h.mu.Lock()
+	defer h.mu.Unlock()
+
+	for i, registered := range h.wallets {
+		if registered != w {
+			continue
+		}
+
+		h.wallets = append(h.wallets[:i], h.wallets[i+1:]...)
+
+		return true
+	}
+
+	return false
+}
+
+// ReleaseManager releases a Manager from harness teardown ownership.
+//
+// It returns false when the Manager is nil or was not registered. The remaining
+// registrations retain their original order.
+func (h *HarnessTest) ReleaseManager(manager *wallet.Manager) bool {
+	h.Helper()
+
+	if manager == nil {
+		return false
+	}
+
+	for i, registered := range h.managers {
+		if registered != manager {
+			continue
+		}
+
+		h.managers = append(h.managers[:i], h.managers[i+1:]...)
+
+		return true
+	}
+
+	return false
+}
+
 // ActiveWallets returns a snapshot of wallets registered with this harness.
 func (h *HarnessTest) ActiveWallets() []*wallet.Wallet {
 	h.Helper()
--- a/bwtest/harness_test.go
+++ b/bwtest/harness_test.go
@@ -11,6 +11,91 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// TestDeregisterWalletRemovesRegisteredWallet verifies that deregistration
+// removes exactly the requested wallet, preserves the order of the remaining
+// registrations, and reports absent wallets without failing.
+func TestDeregisterWalletRemovesRegisteredWallet(t *testing.T) {
+	t.Parallel()
+
+	// Arrange: a harness with three registered wallet identities.
+	h := &HarnessTest{T: t}
+	first := &wallet.Wallet{}
+	removed := &wallet.Wallet{}
+	last := &wallet.Wallet{}
+
+	h.RegisterWallet(first)
+	h.RegisterWallet(removed)
+	h.RegisterWallet(last)
+
+	// Act: the middle wallet is deregistered.
+	require.True(t, h.DeregisterWallet(removed))
+
+	// Assert: only it is removed, ordering is retained, and absence is safe.
+	active := h.ActiveWallets()
+	require.Len(t, active, 2)
+	require.Same(t, first, active[0])
+	require.Same(t, last, active[1])
+	require.False(t, h.DeregisterWallet(removed))
+	require.False(t, h.DeregisterWallet(nil))
+}
+
+// TestReleaseManagerRemovesRegisteredManager verifies that releasing a Manager
+// preserves the remaining teardown order and reports an absent Manager.
+func TestReleaseManagerRemovesRegisteredManager(t *testing.T) {
+	t.Parallel()
+
+	// Arrange: a harness with three Manager registrations.
+	first := &wallet.Manager{}
+	released := &wallet.Manager{}
+	last := &wallet.Manager{}
+	h := &HarnessTest{
+		T:        t,
+		managers: []*wallet.Manager{first, released, last},
+	}
+
+	// Act: the middle Manager is released from teardown ownership.
+	require.True(t, h.ReleaseManager(released))
+
+	// Assert: only it is removed, ordering is retained, and absence is safe.
+	require.Len(t, h.managers, 2)
+	require.Same(t, first, h.managers[0])
+	require.Same(t, last, h.managers[1])
+	require.False(t, h.ReleaseManager(released))
+	require.False(t, h.ReleaseManager(nil))
+}
+
+// TestReleaseManagerTransfersTeardownOwnership verifies that a successfully
+// closed Manager can be removed from teardown without a second close.
+func TestReleaseManagerTransfersTeardownOwnership(t *testing.T) {
+	t.Parallel()
+
+	// Arrange: a Manager owned by a harness.
+	h := &HarnessTest{
+		T:            t,
+		dbType:       "kvdb",
+		WalletDBPath: filepath.Join(t.TempDir(), "wallet.db"),
+	}
+	manager := h.NewWalletManager()
+
+	// Act: the test closes the Manager before releasing teardown ownership.
+	require.NoError(t, manager.Close())
+	require.True(t, h.ReleaseManager(manager))
+
+	// Assert: teardown succeeds without a second close and the database
+	// reopened.
+	require.NoError(t, h.teardownWallets(t.Context()))
+	reopened, err := wallet.NewManager(t.Context(), wallet.ManagerConfig{
+		//nolint:staticcheck // This test intentionally reopens legacy kvdb.
+		Backend:     wallet.DBBackendKVDB,
+		DataSource:  h.WalletDBPath,
+		ChainParams: h.NetParams(),
+	})
+	require.NoError(
+		t, err, "released Manager must leave the database available",
+	)
+	require.NoError(t, reopened.Close())
+}
+
 // TestTeardownWalletsClosesManagerAfterFailedCreate verifies that centralized
 // teardown releases a Manager even when wallet creation fails.
 func TestTeardownWalletsClosesManagerAfterFailedCreate(t *testing.T) {
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -16,8 +16,9 @@ const (
 	// defaultPubPass is the standard public passphrase used by test wallets.
 	defaultPubPass = "public"
 
-	// defaultPrivPass is the standard private passphrase used by test wallets.
-	defaultPrivPass = "private"
+	// TestWalletPrivatePassphrase is the standard private passphrase used by
+	// harness test wallets.
+	TestWalletPrivatePassphrase = "private"
 
 	// defaultWalletRecoveryWindow keeps enough look-ahead addresses for test
 	// cases that derive multiple addresses while scanning historical blocks.
@@ -28,27 +29,25 @@ const (
 	defaultWalletSyncRetryInterval = 500 * time.Millisecond
 )
 
-// CreateEmptyWallet creates, starts, and registers a new wallet instance.
-//
-// This is intended for non-manager integration tests that want a ready-to-use
-// wallet without repeating boilerplate.
-func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
-	h.Helper()
+// TestWalletConfig builds the standard Config and CreateWalletParams for a
+// harness test wallet. Callers retain ownership of the wallet lifecycle.
+func (h *HarnessTest) TestWalletConfig() (wallet.Config,
+	wallet.CreateWalletParams) {
 
-	name := "itest-" + strings.ReplaceAll(h.Name(), "/", "_")
+	h.Helper()
 
 	cfg := wallet.Config{
 		// The chain client is prepared by the harness; the database is
 		// owned by the Manager built below.
-		Chain: h.ChainClient,
+		Chain:       h.ChainClient,
+		ChainParams: h.NetParams(),
 
 		// Keep network and startup behavior deterministic across tests.
-		ChainParams:             h.NetParams(),
 		RecoveryWindow:          defaultWalletRecoveryWindow,
 		WalletSyncRetryInterval: defaultWalletSyncRetryInterval,
 
 		// Use a unique wallet name per test to avoid collisions in logs.
-		Name:          name,
+		Name:          "itest-" + strings.ReplaceAll(h.Name(), "/", "_"),
 		PubPassphrase: []byte(defaultPubPass),
 	}
 
@@ -56,12 +55,24 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
 		// Generate a fresh seed for each test wallet.
 		Mode:              wallet.ModeGenSeed,
 		PubPassphrase:     []byte(defaultPubPass),
-		PrivatePassphrase: []byte(defaultPrivPass),
+		PrivatePassphrase: []byte(TestWalletPrivatePassphrase),
 
 		// Use an old birthday to ensure the wallet can discover historical
 		// blocks when used in tests that pre-mine chain state.
 		Birthday: time.Now().Add(-1 * time.Hour),
 	}
+
+	return cfg, params
+}
+
+// CreateEmptyWallet creates, starts, and registers a new wallet instance.
+//
+// This is intended for non-manager integration tests that want a ready-to-use
+// wallet without repeating boilerplate.
+func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
+	h.Helper()
+
+	cfg, params := h.TestWalletConfig()
 
 	manager := h.NewWalletManager()
 	w, err := manager.Create(cfg, params)
@@ -88,7 +99,7 @@ func (h *HarnessTest) CreateFundedWallet() *wallet.Wallet {
 	w := h.CreateEmptyWallet()
 
 	err := w.Unlock(h.Context(), wallet.UnlockRequest{
-		Passphrase: []byte(defaultPrivPass),
+		Passphrase: []byte(TestWalletPrivatePassphrase),
 	})
 	require.NoError(h, err, "failed to unlock wallet")
 
--- a/itest/controller_test.go
+++ b/itest/controller_test.go
@@ -1,0 +1,177 @@
+//go:build itest
+
+package itest
+
+import (
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/wallet"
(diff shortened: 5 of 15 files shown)


itest/harness_helpers_test.go
```diff
@@ -0,0 +1,69 @@
...
+// walletConfig builds the standard Manager Config and CreateWalletParams for an
+// integration test wallet, wired to the harness-configured chain backend and
+// database path.
+//
+// Manager/controller cases drive wallet.NewManager() directly (rather than the
+// harness CreateEmptyWallet helper) because they need explicit control over the
+// create/start/stop/reload lifecycle. This helper keeps the config literal in
+// one place while honoring the -db flag via h.DBBackend(), so db=sqlite and
+// db=kvdb genuinely exercise their respective backends.
+func walletConfig(h *bwtest.HarnessTest,
```
Abdulkbk: We should consolidate this with what we have in [harness_wallet.go](https://github.com/btcsuite/btcwallet/blob/wallet-default-sqlite/bwtest/harness_wallet.go), to avoid having conflicts when working with wallets in tests. perhaps we should update [CreateEmptyWallet](https://github.com/btcsuite/btcwallet/blob/wallet-default-sqlite/bwtest/harness_wallet.go#L66-L130) to use `walletConfig`

itest/manager_test.go
```diff
@@ -54,6 +31,149 @@ func testCreateWallet(h *bwtest.HarnessTest) {
...
+
+// testManagerLoadMissing verifies that loading a wallet that was never created
+// fails rather than silently returning an empty wallet.
+func testManagerLoadMissing(h *bwtest.HarnessTest) {
+	h.Helper()
+
+	cfg, _ := walletConfig(h, "missingwallet")
+
+	_, err := wallet.NewManager().Load(cfg)
+	require.Error(h, err, "load of never-created wallet should fail")
```
Abdulkbk, line 159: It's unfortunate we can't use `require.ErrorIs` here because of the different errors returned by different backends.

.github/workflows/main.yml
```diff
@@ -208,12 +208,15 @@ jobs:
   ########################
```
Abdulkbk, line 211: Should we also update [itest/README.md](https://github.com/btcsuite/btcwallet/blob/wallet-default-sqlite/itest/README.md), should've been updated in the upstream.

Abdulkbk (commented):
Left a few feedback.

itest/harness_helpers_test.go
```diff
@@ -0,0 +1,69 @@
...
+import (
+	"time"
+
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/wallet"
+)
+
+const (
+	// pubPass is the public passphrase used by itest wallets.
+	pubPass = "public-passphrase"
```
yyforyongyu: These constants and the config literal below duplicate `bwtest.CreateEmptyWallet` (`bwtest/harness_wallet.go:15-32` and `74-110`) but with *different* values (`"public-passphrase"` vs `"public"`), so the suite now carries two competing passphrase conventions and two copies of the same config, and a case that mixes `h.CreateEmptyWallet()` with `privPass` will fail to unlock; export the builder from `bwtest` and have `CreateEmptyWallet` call it too.

(`claude-opus-5`)

itest/harness_helpers_test.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// privPass is the correct private passphrase used by itest wallets.
+	privPass = "private-passphrase"
+
+	// wrongPrivPass is a private passphrase that never matches, used to
+	// exercise unlock failure paths.
+	wrongPrivPass = "wrong-private-passphrase"
```
yyforyongyu: `wrongPrivPass` has no user until `itest: add Controller lifecycle tests` two commits later, so it lands as dead code in this commit (`make unit` per-commit doesn't build the `itest` tag, which is why CI stays green); move the constant into the commit that uses it so each commit is self-contained.

(`claude-opus-5`)

itest/harness_helpers_test.go
```diff
@@ -0,0 +1,69 @@
...
+	cfg := wallet.Config{
+		DB: wallet.DBConfig{
+			Backend: h.DBBackend(),
+			KVDB: wallet.KVDBConfig{
+				DBPath: h.WalletDBPath,
+			},
+		},
+		Chain:                   h.ChainClient,
+		ChainParams:             h.NetParams(),
+		RecoveryWindow:          20,
```
yyforyongyu: `RecoveryWindow: 20` is an unnamed magic number while every other value in this helper is a named constant and `bwtest` already documents the same value as `defaultWalletRecoveryWindow` (`bwtest/harness_wallet.go:24-27`); give it a named constant in the block above.

(`claude-opus-5`)

itest/harness_helpers_test.go
```diff
@@ -0,0 +1,69 @@
...
+		WalletSyncRetryInterval: syncRetryInterval,
+		Name:                    name,
+		PubPassphrase:           []byte(pubPass),
+	}
+
+	params := wallet.CreateWalletParams{
+		Mode:          wallet.ModeGenSeed,
+		PubPassphrase: []byte(pubPass),
+
+		PrivatePassphrase: []byte(privPass),
```
yyforyongyu: The blank lines around `PrivatePassphrase` split it out of the `Mode`/`PubPassphrase` alignment group for no logical reason (compare `bwtest/harness_wallet.go:102-104`, where the three are one stanza); drop them.

(`claude-opus-5`)

itest/manager_test.go
```diff
@@ -3,58 +3,177 @@
...
 	// Register the wallet so harness helpers can assert global invariants.
+	// Registration also hands teardown to the harness, which stops every
+	// registered wallet during subtest cleanup, so we don't add a Stop
+	// cleanup here.
 	h.RegisterWallet(w)
 
+	// Info should reflect a real, running wallet.
+	info, err := w.Info(h.Context())
+	require.NoError(h, err, "failed to query wallet info")
+	require.NotEmpty(h, info.Backend, "wallet info missing backend")
```
yyforyongyu: These `Info` assertions are outside this commit's stated scope ("Cover duplicate create, load/reload durability, and missing load") and are a strictly weaker duplicate of `testControllerInfo` added in the next commit, which asserts the exact backend and chain params (`itest/controller_test.go:126-133`); drop them and let the `Info` case own that coverage.

(`claude-opus-5`)

itest/manager_test.go
```diff
@@ -3,58 +3,177 @@
...
+	require.NoError(h, err, "failed to start wallet")
+
+	// Stop the wallet so it is torn down and evicted from the manager cache,
+	// forcing the retry to consult the durable store. Manage this instance
+	// manually (do not register it) so we control ordering.
+	require.NoError(h, w.Stop(h.Context()), "failed to stop wallet")
+
+	// A second create over the completed wallet must be rejected, both from
+	// the same manager and from a fresh one reading the durable store.
+	_, err = manager.Create(cfg, params)
```
yyforyongyu: Both duplicate-create assertions run after the `Stop()` on line 65 has already evicted the cache entry, so they exercise the same durable `createComplete` rejection twice while `beginCreate`'s live-wallet guard (`wallet/manager.go:1605-1607`) stays uncovered by the suite; assert `manager.Create(cfg, params)` returns `ErrWalletExists` before the `Stop` as well.

(`claude-opus-5`)

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+// testControllerStartStop verifies the wallet lifecycle: a wallet starts and
+// stops cleanly, a stopped instance refuses to restart (ErrWalletStopped), Stop
+// is idempotent, and a fresh Load yields a working instance.
+func testControllerStartStop(h *bwtest.HarnessTest) {
+	h.Helper()
+
+	cfg, params := walletConfig(h, "lifecyclewallet")
+
+	manager := wallet.NewManager()
```
yyforyongyu: This `NewManager`/`Create`/`require`/`Start`/`require` block is repeated verbatim in all six new cases (also lines 58-60 and 112-114 here, plus `itest/manager_test.go:21-26`, `55-60`, `90-95`), which is exactly the boilerplate the shared-helper commit set out to remove; extract a `createWallet(h, name)` helper next to `walletConfig`.

(`claude-opus-5`)

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+// they are forbidden before Start, unlock toggles the locked flag, lock is
+// idempotent, and a wrong passphrase fails without unlocking the wallet.
+func testControllerUnlockLock(h *bwtest.HarnessTest) {
+	h.Helper()
+
+	cfg, params := walletConfig(h, "unlockwallet")
+
+	manager := wallet.NewManager()
+	w, err := manager.Create(cfg, params)
+	require.NoError(h, err, "failed to create wallet")
```
yyforyongyu, line 62: The wallet is created before three failure-prone assertions but is only handed to the harness on line 75, so a failure in between leaves a wallet with open stores that `stopActiveWallets` never sees (`bwtest/harness.go:263-295`); add `defer w.Stop(h.Context())` right after `Create` here and in `testControllerInfo`/the manager cases (`Stop` is idempotent, so the explicit stops still work, and unlike early registration it won't make `MineBlocks` assert sync on a stopped instance).

(`claude-opus-5`)

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+	requireLocked(h, w, false)
+
+	// Lock, then lock again: locking is idempotent.
+	require.NoError(h, w.Lock(h.Context()), "failed to lock wallet")
+	requireLocked(h, w, true)
+	require.NoError(h, w.Lock(h.Context()), "second lock should be a no-op")
+	requireLocked(h, w, true)
+
+	// A wrong passphrase fails and leaves the wallet locked.
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
```
yyforyongyu, line 110: The wrong-passphrase attempt is made while the wallet is already locked, so it cannot catch the documented side effect that a failed unlock on an *unlocked* wallet leaves it locked — `handleUnlockReq` locks the vault before validating (`wallet/controller.go:1190-1204`); add a wrong-passphrase attempt right after the successful unlock on line 82.

(`claude-opus-5`)

itest/manager_test.go
```diff
@@ -3,58 +3,177 @@
...
+		h, h.NetParams(), info.ChainParams, "wallet info chain params mismatch",
+	)
+
 	// Mine a few blocks and require the wallet catches up.
 	h.MineBlocks(5)
 }
+
+// testManagerCreateDuplicate verifies that creating a second wallet with the
+// same name over a fully created wallet is rejected with ErrWalletExists. This
+// exercises the durable CreateComplete path against a real SQL store rather
```
yyforyongyu: This test also runs under `db=kvdb` in the CI matrix added in this PR, where the durable store is bbolt/walletdb, not SQL — "against a real SQL store" overclaims. Suggest rewording to "against the durable store".

(`k3`)

itest/manager_test.go
```diff
@@ -3,58 +3,177 @@
...
+	// cleanup here.
 	h.RegisterWallet(w)
 
+	// Info should reflect a real, running wallet.
+	info, err := w.Info(h.Context())
+	require.NoError(h, err, "failed to query wallet info")
+	require.NotEmpty(h, info.Backend, "wallet info missing backend")
+	require.Equal(
+		h, h.NetParams(), info.ChainParams, "wallet info chain params mismatch",
+	)
```
yyforyongyu: Commit `2d0764aa` ("itest: add Manager create/load tests") also adds these `Info` assertions to `testCreateWallet`, but its body only mentions duplicate-create/load coverage, and `Info` is already exercised more thoroughly in the controller commit (`testControllerInfo`). Consider moving this to the controller commit (or at least mentioning it in the body) to keep the manager/controller split clean.

(`k3`)

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+	require.NoError(h, err, "failed to query wallet info after mining")
+	require.True(h, after.Synced, "wallet should be synced after mining")
+	require.Greater(
+		h, after.SyncedTo.Height, before.SyncedTo.Height,
+		"synced height did not advance after mining",
+	)
+	require.Equal(
+		h, before.BirthdayBlock, after.BirthdayBlock,
+		"birthday block should be static",
+	)
```
yyforyongyu: Both `before.BirthdayBlock` and `after.BirthdayBlock` are served from the same running wallet's in-memory cache (`Info` → `birthdayBlockForInfo`, populated once at `Start` by `verifyBirthday`), so this comparison is a tautology — it cannot fail on a store/backend regression and does not prove the birthday block "stays put". Consider dropping it (reload durability is already covered by `testManagerLoadReload`) or asserting against a store-level read.

(`k3`)

itest/harness_helpers_test.go
yyforyongyu: Minor: the stanza split separates `PubPassphrase` and `PrivatePassphrase` while grouping `Mode` with `PubPassphrase`. Suggest grouping the two passphrase fields in one stanza per the repo's logical-stanza convention.

(`k3`)

itest/manager_test.go
yyforyongyu: This case now runs under the KVDB matrix too, so the `real SQL store` description is false for half the executions; describe the durable backend store instead (`gpt-5.6-sol`).

itest/manager_test.go
```diff
@@ -3,58 +3,177 @@
...
+// exercises the durable CreateComplete path against a real SQL store rather
+// than the in-memory cache guard.
+func testManagerCreateDuplicate(h *bwtest.HarnessTest) {
+	h.Helper()
+
+	cfg, params := walletConfig(h, "dupwallet")
+
+	manager := wallet.NewManager()
+	w, err := manager.Create(cfg, params)
+	require.NoError(h, err, "failed to create wallet")
```
yyforyongyu, line 51: `Create` leaves an open runtime store, so any fatal `Start` or `Stop` assertion before registration leaks it from failed-test cleanup; register every successful `Create` and `Load` immediately while retaining explicit lifecycle stops (`gpt-5.6-sol`).

itest/manager_test.go
```diff
@@ -3,58 +3,177 @@
...
+	h.RegisterWallet(reloaded)
+
+	// The birthday block is durable metadata and must be unchanged. Compare
+	// the block identity (height and hash) rather than the whole BlockStamp:
+	// the Timestamp is a time.Time whose location representation differs by
+	// storage backend on round-trip, which is not a meaningful difference.
+	infoAfter, err := reloaded.Info(h.Context())
+	require.NoError(h, err, "failed to query reloaded wallet info")
+	require.Equal(
+		h, infoBefore.BirthdayBlock.Height, infoAfter.BirthdayBlock.Height,
```
yyforyongyu: `Start` calls `verifyBirthday` before this comparison, so a missing persisted birthday block is reconstructed from the same timestamp and this assertion still passes; verify persistence before startup repair or make recomputation observably different (`gpt-5.6-sol`).

itest/manager_test.go
```diff
@@ -3,58 +3,177 @@
...
+		h, infoBefore.BirthdayBlock.Hash, infoAfter.BirthdayBlock.Hash,
+		"birthday block hash did not survive reload",
+	)
+
+	// The persisted derivation index must carry over: the next address must
+	// differ from the two derived before reload. If the index had reset,
+	// addr3 would collide with addr1.
+	addr3 := newExternalAddress(h, reloaded)
+	require.NotEqual(
+		h, addr1, addr3, "derivation index did not survive reload",
```
yyforyongyu: Checking only inequality permits any incorrect persisted index other than zero or one to pass; assert the exact derivation branch, index, and account through address metadata (`gpt-5.6-sol`).

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+// stops cleanly, a stopped instance refuses to restart (ErrWalletStopped), Stop
+// is idempotent, and a fresh Load yields a working instance.
+func testControllerStartStop(h *bwtest.HarnessTest) {
+	h.Helper()
+
+	cfg, params := walletConfig(h, "lifecyclewallet")
+
+	manager := wallet.NewManager()
+	w, err := manager.Create(cfg, params)
+	require.NoError(h, err, "failed to create wallet")
```
yyforyongyu, line 21: A successful `Create` owns open stores before the later registration, so any intervening fatal assertion bypasses harness teardown; register every created or loaded wallet immediately and keep the explicit lifecycle stops (`gpt-5.6-sol`).

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// The torn-down instance cannot be restarted.
+	err = w.Start(h.Context())
+	require.ErrorIs(
+		h, err, wallet.ErrWalletStopped,
+		"restart of stopped instance not rejected",
+	)
+
+	// Stop is idempotent.
+	require.NoError(h, w.Stop(h.Context()), "second stop should be a no-op")
```
yyforyongyu, line 29: The rejected `Start` performs an intervening lifecycle transition before this second `Stop`, so the test does not directly cover consecutive-stop idempotence; move the second `Stop` immediately after the first (`gpt-5.6-sol`).

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+	requireLocked(h, w, true)
+	require.NoError(h, w.Lock(h.Context()), "second lock should be a no-op")
+	requireLocked(h, w, true)
+
+	// A wrong passphrase fails and leaves the wallet locked.
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(wrongPrivPass),
+		Timeout:    -1,
+	})
+	require.Error(h, err, "unlock with wrong passphrase should fail")
```
yyforyongyu: `require.Error` lets unrelated context or database failures satisfy the wrong-passphrase case while the locked check still passes; expose or map a stable invalid-passphrase sentinel and assert it with `require.ErrorIs` (`gpt-5.6-sol`).

itest/harness_helpers_test.go
yyforyongyu: it's a nit, feel free to ignore

.github/workflows/main.yml
yyforyongyu: yeah feel free to update it in another PR

.github/workflows/main.yml
```diff
@@ -208,12 +208,15 @@ jobs:
...
+  # Each job runs over both wallet database backends (kvdb and sqlite).
   ########################
   itest-btcd:
-    name: itest btcd
+    name: itest btcd (${{ matrix.db }})
     runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        db: [kvdb, sqlite]
```
yyforyongyu, line 222: let's also do postgres? or is it done in another PR?

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+
+	// A fresh Load rebuilds a working instance from the durable store.
+	reloaded, err := manager.Load(cfg)
+	require.NoError(h, err, "failed to reload wallet")
+	require.NoError(
+		h, reloaded.Start(h.Context()), "failed to start reloaded wallet",
+	)
+
+	// Register only the final active instance for harness teardown.
+	h.RegisterWallet(reloaded)
```
yyforyongyu: Q: why do we need this step? so that the harness can clean up the wallet? looks like the harness should just skip the teardown when there is no active wallet?

itest/controller_test.go
```diff
@@ -0,0 +1,159 @@
...
+	require.Equal(
+		h, h.ChainClient.BackEnd(), before.Backend, "info backend mismatch",
+	)
+	require.Equal(
+		h, h.NetParams(), before.ChainParams, "info chain params mismatch",
+	)
+	require.True(h, before.Locked, "freshly started wallet should be locked")
+
+	// Mining advances the synced height while the birthday block stays put.
+	h.MineBlocks(5)
```
yyforyongyu: nit: can we mine 1 block instead? more blocks mined in the test will slow down other tests

itest/manager_test.go
yyforyongyu: yeah we can define sentinel errors, can fix it in another PR.

itest/manager_test.go
yyforyongyu: it's a nit, feel free to ignore

itest/manager_test.go
yyforyongyu: again nit, feel free to ignore

itest/harness_helpers_test.go
yyforyongyu: yeah would be nice to have a single place hosting the config creation. In addition, there are two set of tests really,
- the first set is to check the initialization of the wallet, so mostly about the manager and the controller, which are tested here. There are also tests to check that different config values can lead to rejection, so wallet config validation tests. These are prerequisites before we can move to other tests.
- the second set is to check all other public APIs, like list addr or list utxo - i can imagine that all these tests will share a very similar setup: 1) create a wallet; 2) start the wallet; 3) check wallet is synced; 4) fund wallet; then perform case-specific setups. So we can expand the bwtest to create a common setup wallet method, which i believe is mostly done in `CreateFundedWallet`.

yyforyongyu (changes requested):
I think `sol` and `k3` have found similar issues, which can easily be addressed. Nice work btw, straightforward commits!

.github/workflows/main.yml
GustavoStingelin: Let's do it in another one, bc it looks like we'll need a few more changes

.github/workflows/main.yml
GustavoStingelin: updated

itest/controller_test.go
GustavoStingelin: I kept the manager/controller setup explicit because these cases intentionally exercise Create/Start/Stop/Load ordering.

itest/controller_test.go
GustavoStingelin: Addressed: registration now follows successful `Create` and happens before `Start`, so the harness owns the opened stores even if startup fails.

itest/controller_test.go
GustavoStingelin: Addressed: registration now follows successful `Create` and happens before `Start`, so the harness owns the opened stores even if startup fails.

itest/controller_test.go
GustavoStingelin: Addressed: the wrong-passphrase attempt now follows a successful unlock and verifies that the wallet becomes locked. The test then re-unlocks before checking explicit `Lock` idempotence.

itest/manager_test.go
GustavoStingelin: Addressed: every successful `Create` and `Load` is registered immediately; the explicit `Stop` calls remain where the test needs to verify cache eviction.

itest/harness_helpers_test.go
GustavoStingelin: Addressed: the current tree centralizes wallet setup in `bwtest.TestWalletConfig`; `CreateEmptyWallet` uses it, and the manager/controller cases use the shared builder directly. The old `itest/harness_helpers_test.go` helper is no longer present.

itest/harness_helpers_test.go
GustavoStingelin: Addressed: the duplicate config and passphrase constants were removed. Wallet setup now lives in `bwtest.TestWalletConfig`, which `CreateEmptyWallet` also uses.

itest/harness_helpers_test.go
GustavoStingelin: Addressed: `TestWalletConfig` now uses the named `wallet.MinRecoveryWindow` constant instead of the literal `20`.

itest/harness_helpers_test.go
GustavoStingelin: Addressed: the shared config keeps the public and private passphrase fields together.

itest/controller_test.go
GustavoStingelin: Addressed: the tautological birthday comparison was removed, so the `Info` case no longer claims to prove birthday persistence.

itest/controller_test.go
GustavoStingelin: Addressed: the controller Info case now mines one block instead of five.

itest/harness_helpers_test.go
GustavoStingelin: Addressed: the shared wallet config keeps the public and private passphrase fields together.

itest/manager_test.go
GustavoStingelin: Addressed: the manager create case no longer contains the `Info` assertions; the controller Info case owns backend, chain-parameter, and sync coverage.

itest/controller_test.go
GustavoStingelin: `Create` opens the wallet stores before `Start`, so the harness needs to register the wallet immediately after `Create`. If `Start` fails, `require.NoError` exits the test before any later registration, and the harness would not know about the open stores. `Stop` is safe for a never-started wallet, so registering early gives cleanup ownership without requiring the wallet to be active.

itest/controller_test.go
GustavoStingelin: Addressed: the wrong-passphrase case now uses `require.ErrorIs(err, wallet.ErrInvalidPassphrase)` and keeps the locked-state check, so unrelated failures no longer satisfy it.

itest/manager_test.go
GustavoStingelin: Addressed: the `Info` assertions were removed from the manager case; controller `Info` owns that coverage.

itest/manager_test.go
GustavoStingelin: Addressed: after reload, the test now checks account number, external branch, and exact derivation indices 0, 1, and 2 through `GetAddressInfo`.

itest/controller_test.go
GustavoStingelin: Addressed: the second `Stop` now immediately follows the first, before the restart assertion.

itest/manager_test.go
GustavoStingelin: Addressed: It sets a future effective birthday so reconstruction tracks the chain tip, records the initial birthday block, mines one block, and then reloads. If the persisted birthday block were missing, `verifyBirthday` would reconstruct a different block from the advanced tip; matching the original therefore verifies persistence.

itest/manager_test.go
GustavoStingelin: Addressed: duplicate creation is now checked while the wallet is still cached and again through a fresh manager after the wallet is stopped.

itest/manager_test.go
GustavoStingelin: Addressed: the manager test now describes the durable backend store rather than a SQL store.

GustavoStingelin: Addressed: the test description now refers to the durable store, which is accurate for both KVDB and SQLite runs.

GustavoStingelin:
@yyforyongyu @Abdulkbk  I’ve rebased this branch several times over this week, and the upstream branch’s semantics are still evolving, significantly affecting parts of this test implementation. To avoid repeated rework and asking you to review a moving target, I’ll wait until the upstream work is complete. I’ll then perform a final rebase and review the changes myself before requesting another review.

GustavoStingelin:
@saubyk, this one is missing in c3po

GustavoStingelin:
A lot of things changed during the rebases. The previous address creation tests were moved from this pr.

wallet/syncer.go
```diff
@@ -491,6 +491,32 @@ func (s *syncer) checkRollback(ctx context.Context) error {
 	return nil
 }
 
+// resolveMissingHistoricalBlock accepts a missing historical block when the
+// persisted synced tip matches the chain backend.
+func (s *syncer) resolveMissingHistoricalBlock(listErr error,
```
yyforyongyu, line 496: Further looked into this area, think we can optimize it a bit, but nothing critical so just track it here #1302

yyforyongyu (approved):
LGTM🥏

bwtest/harness.go
```diff
@@ -324,6 +324,57 @@ func (h *HarnessTest) RegisterWallet(w *wallet.Wallet) {
...
+	}
+
+	return false
+}
+
+// ReleaseManager releases a Manager from harness teardown ownership.
+//
+// It returns false when the Manager is nil or was not registered. The remaining
+// registrations retain their original order.
+func (h *HarnessTest) ReleaseManager(manager *wallet.Manager) bool {
```
Abdulkbk, line 358: This looks like a twin of `DeregisterWallet` above, except that it does not hold the Lock.

bwtest/harness_wallet.go
```diff
@@ -28,41 +29,51 @@ const (
...
+// harness test wallet. Callers retain ownership of the wallet lifecycle.
+func (h *HarnessTest) TestWalletConfig() (wallet.Config,
+	wallet.CreateWalletParams) {
 
-	name := "itest-" + strings.ReplaceAll(h.Name(), "/", "_")
+	h.Helper()
 
 	cfg := wallet.Config{
 		// The chain client is prepared by the harness; the database is
 		// owned by the Manager built below.
```
Abdulkbk, line 41: might as well update this comment since no Manager is being built below.

bwtest/harness_test.go
```diff
@@ -11,6 +11,91 @@ import (
...
+
+// TestReleaseManagerTransfersTeardownOwnership verifies that a successfully
+// closed Manager can be removed from teardown without a second close.
+func TestReleaseManagerTransfersTeardownOwnership(t *testing.T) {
+	t.Parallel()
+
+	// Arrange: a Manager owned by a harness.
+	h := &HarnessTest{
+		T:            t,
+		dbType:       "kvdb",
```
Abdulkbk, line 75: niy: use dbNameKvdb, available in the package

Abdulkbk (approved):
LGTM 🚢

All non-blocking

PR #1303, opened 2026-08-06
wallet+bwtest: add PostgreSQL harness

yyforyongyu:
## Change Description

Draft stacked on #1300 (`itests/very-first-itests`).

This change adds PostgreSQL as a first-class `wallet.Manager` backend while
sharing the existing SQL create/load implementation with SQLite. The test
harness starts one PostgreSQL 18 server for the suite, provisions an isolated
database for each subtest, validates the migrated schema, and closes wallets
and Managers before dropping databases and terminating the container.

The four Manager and three Controller integration cases introduced by #1300
run unchanged with `db=postgres`; this PR does not modify `itest/*` or expand
the default CI matrix.

## Steps to Test

```bash
make fmt
/opt/homebrew/bin/go test ./wallet ./bwtest
make itest chain=btcd db=kvdb
make itest chain=btcd db=sqlite
make itest chain=btcd db=postgres
make lint-check
```

### Testing

- [x] Your PR passes all CI checks.
- [x] Tests covering the positive and negative (error paths) are included.
- [ ] Bug fixes contain tests triggering the bug to prevent regressions.

### Code Style and Documentation

- [x] The change is not [insubstantial](https://github.com/btcsuite/btcwallet/blob/master/docs/developer/contribution_guidelines.md#substantial-contributions-only). Typo fixes are not accepted to fight bot spam.
- [x] The change obeys the [Code Documentation and Commenting](https://github.com/btcsuite/btcwallet/blob/master/docs/developer/contribution_guidelines.md#code-documentation-and-commenting) guidelines, and lines wrap at 80.
- [x] Commits follow the [Ideal Git Commit Structure](https://github.com/btcsuite/btcwallet/blob/master/docs/developer/contribution_guidelines.md#ideal-git-commit-structure).
- [x] Any new logging statements use an appropriate subsystem and logging level.

📝 Please see our [Contribution Guidelines](https://github.com/btcsuite/btcwallet/blob/master/docs/developer/contribution_guidelines.md) for further guidance.

PR #1304, opened 2026-08-06
multi: match nested CI base branches

yyforyongyu:
## Summary

- run pull request CI for base branch names containing slashes

## Change Description

GitHub Actions treats the single-star branch glob as matching only branch
names without slashes. Use the recursive glob so stacked pull requests
targeting branches such as itests/very-first-itests trigger CI.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
       - "master"
   pull_request:
     branches:
-      - "*"
+      - "**"
 concurrency:
   # Cancel any previous workflows if they are from a PR or push.
   group: ${{ github.event.pull_request.number || github.ref }}


.github/workflows/main.yml
```diff
@@ -6,7 +6,7 @@ on:
       - "master"
   pull_request:
     branches:
-      - "*"
+      - "**"
```
Abdulkbk, line 9: So basically this removes the restrictive filter we had before. I wonder if that will be the same as what github defaults to.

one nit: commit msg says "multi:"

Abdulkbk (approved):

.github/workflows/main.yml
yyforyongyu: yea so this is changed such that branches off `/` can also trigger the CI, e.g., previous #1303 couldn't trigger the CI due to the base branch name `itests/xxx`. And I don't think we were excluding `/`, this was an oversight.

PR #1305, opened 2026-08-06
wallet: define watch-only and script policy

yyforyongyu:
## Summary

Define explicit watch-only wallet creation and script-import policies.

## Change Description

- Validate every wallet creation mode before backend mutation.
- Restrict watch-only creation to shell mode with public initial accounts.
- Document SQL script-import limits and legacy kvdb behavior.


GustavoStingelin:
similar work on https://github.com/btcsuite/btcwallet/pull/1306

yyforyongyu:
> similar work on #1306

nah it's different, here we actually fix a bug. Will wait for 1306 tho, as we do FIFO style.

yyforyongyu:
Hmm forgot to change the base branch when cleaning it, guess I need to open a new one now.

PR #1306, opened 2026-08-06
itest+wallet: make WatchOnly the sole authority at create time

GustavoStingelin:
`CreateWalletParams.WatchOnly` is what every backend persists, so a create
request whose key material contradicts it is now rejected instead of being
silently reinterpreted:

- A watch-only wallet may not carry a private root key (nothing persists it)
  nor use a seed mode; a spendable wallet requires a private root key.
- The kvdb backend refuses any root key on a watch-only wallet: waddrmgr can
  only represent the rootless shape, and dropping the key would make the same
  request mean two different things across backends.
- SQLite takes `IsWatchOnly` from the params instead of inferring it from the
  key.

### DIFF
--- a/itest/manager_test.go
+++ b/itest/manager_test.go
@@ -15,8 +15,6 @@ import (
 
 // testCreateWallet verifies a wallet can be created, started, and synced.
 func testCreateWallet(h *bwtest.HarnessTest) {
-	h.Helper()
-
 	// This is a manager-focused test, so drive the Manager API directly
 	// rather than the harness's CreateEmptyWallet convenience helper.
 	cfg, params := h.TestWalletConfig()
@@ -45,8 +43,6 @@ func testCreateWallet(h *bwtest.HarnessTest) {
 // testManagerCreateDuplicate verifies that both a live wallet cache entry and
 // a completed wallet in the durable store reject a duplicate creation.
 func testManagerCreateDuplicate(h *bwtest.HarnessTest) {
-	h.Helper()
-
 	cfg, params := h.TestWalletConfig()
 
 	manager := h.NewWalletManager()
@@ -80,8 +76,6 @@ func testManagerCreateDuplicate(h *bwtest.HarnessTest) {
 // testManagerLoadReload verifies Manager cache identity, durable reload, and
 // birthday metadata while preserving lifecycle ownership.
 func testManagerLoadReload(h *bwtest.HarnessTest) {
-	h.Helper()
-
 	cfg, params := h.TestWalletConfig()
 
 	// Keep the effective birthday five days ahead of the chain after the
@@ -153,8 +147,6 @@ func testManagerLoadReload(h *bwtest.HarnessTest) {
 // testManagerLoadMissing verifies that loading a wallet that was never created
 // fails rather than silently returning an empty wallet.
 func testManagerLoadMissing(h *bwtest.HarnessTest) {
-	h.Helper()
-
 	cfg, _ := h.TestWalletConfig()
 
 	manager := h.NewWalletManager()
--- a/itest/manager_test.go
+++ b/itest/manager_test.go
@@ -3,10 +3,13 @@
 package itest
 
 import (
+	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
 	"github.com/btcsuite/btcwallet/bwtest"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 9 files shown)


GustavoStingelin:
Did CI stop running?

GustavoStingelin:
ok, looks like it's an incident with gh actions

itest/manager_test.go
```diff
@@ -158,3 +160,167 @@ func testManagerLoadMissing(h *bwtest.HarnessTest) {
...
+
+// testManagerCreateWatchOnly verifies that a watch-only wallet is created,
+// starts, syncs like a spendable wallet, and stays watch-only across a reload
+// from the durable store.
+//
+// The wallet is rootless: that is the one watch-only shape every backend can
+// represent, and its keyspace arrives later as account-level xpub imports. An
+// XPub root is a SQL-only variant, covered by the Manager unit tests.
+func testManagerCreateWatchOnly(h *bwtest.HarnessTest) {
+	h.Helper()
```
yyforyongyu: I would not do thing - also that's why there is no such thing in `lnd`, IIRC this will push native errors through harness.go and back to the generic runner, which is not very helpful when debugging.

itest/manager_test.go
```diff
@@ -158,3 +160,167 @@ func testManagerLoadMissing(h *bwtest.HarnessTest) {
...
+
+	w, err = manager.Create(cfg, params)
+	require.ErrorIs(
+		h, err, wallet.ErrWalletParams,
+		"initial accounts accepted for a spendable wallet",
+	)
+	require.Nil(h, w, "rejected create returned a wallet")
+
+	// Initial accounts are only honored for a shell wallet.
+	cfg, params = h.TestWalletConfig()
```
yyforyongyu: nit: could do table-driven

wallet/manager.go
```diff
@@ -397,16 +413,52 @@ func (m *Manager) prepareWalletCreation(cfg Config,
...
 
 	return rootKey, nil
 }
 
+// validateWatchOnlyRootKey checks the derived root key against the requested
+// watch-only state.
+//
+// params.WatchOnly is the single source of truth: every backend persists that
+// flag verbatim, so a request whose key material contradicts it is rejected
+// here rather than silently reinterpreted.
```
yyforyongyu, line 432: 👍

wallet/manager_test.go
```diff
@@ -654,6 +654,162 @@ func TestValidateInitialAccountsModeUpfront(t *testing.T) {
...
+	require.NoError(t, err)
+
+	rootPubKey, err := rootKey.Neuter()
+	require.NoError(t, err)
+
+	tests := []struct {
+		name       string
+		watchOnly  bool
+		rootKey    *hdkeychain.ExtendedKey
+		wantErr    bool
```
yyforyongyu: it's unnecessary to have both `wantErr` and `wantErrMsg`? can just use `wantErrMsg` and check for zero values.

yyforyongyu (changes requested):
Cool - going to be more strict about test quality now given that this is a dedicated itest phase. Left a few comments.

itest/manager_test.go
GustavoStingelin: nice call, agreed

itest/manager_test.go
GustavoStingelin: done

itest/manager_test.go
GustavoStingelin: done

wallet/manager_test.go
GustavoStingelin: done

yyforyongyu (approved):
LGTM 🚢

GustavoStingelin:
just rebased

wallet/manager.go
```diff
@@ -299,6 +304,16 @@ func (p *CreateWalletParams) validate() error {
...
 
+	// A seed is private material: both seed modes produce a private root
+	// key that a watch-only wallet cannot hold. Reject the contradiction
+	// here, before the key is generated or derived, so no signing material
+	// is ever created only to be thrown away.
+	if p.WatchOnly && (p.Mode == ModeGenSeed || p.Mode == ModeImportSeed) {
+		return fmt.Errorf("%w: a seed derives a private root key, which "+
+			"a watch-only wallet cannot hold; use ModeImportExtKey "+
+			"with an XPub or ModeShell", ErrWalletParams)
+	}
```
Abdulkbk, line 315: This block isn't covered in the test. I commented it out, and the tests still pass.

itest/manager_test.go
```diff
@@ -150,11 +147,159 @@ func testManagerLoadReload(h *bwtest.HarnessTest) {
...
+	initialAccounts := []wallet.WatchOnlyAccount{{Name: "watchonly"}}
+
+	neuteredRootKey := rootPubKey(h)
+	privateRootKey := masterKey(h)
+
+	manager := h.NewWalletManager()
+
+	testCases := []struct {
+		name   string
+		update func(*wallet.CreateWalletParams)
```
Abdulkbk: we should add `wantErrMsg` as we did in the other test and use `require.ErrorContains` to make stronger assertions.

Abdulkbk (changes requested):
Left a few suggestions

wallet/manager.go
GustavoStingelin: added the test case

itest/manager_test.go
GustavoStingelin: added

Abdulkbk (approved):
LGTM

PR #1307, opened 2026-08-07
itest: add UtxoManager integration tests

Abdulkbk:
Adds integration tests for the wallet's `UtxoManager` interface — `ListUnspent`, `GetUtxo`, `LeaseOutput`, `ReleaseOutput`, `ListLeasedOutputs` — one commit per case, plus a reusable funding helper in `bwtest`.

Part of #1250.

### DIFF
--- a/bwtest/harness_miner.go
+++ b/bwtest/harness_miner.go
@@ -4,9 +4,13 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
+	"github.com/btcsuite/btcd/address/v2"
+	"github.com/btcsuite/btcd/blockchain"
 	"github.com/btcsuite/btcd/btcutil/v2"
 	"github.com/btcsuite/btcd/chainhash/v2"
+	"github.com/btcsuite/btcd/txscript/v2"
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/btcsuite/btcwallet/bwtest/wait"
 	"github.com/stretchr/testify/require"
@@ -54,6 +58,39 @@ const (
 	// miner-funded test transactions.
 	MinerFeeRate = btcutil.Amount(10)
 )
+
+// MineBlockToAddress mines a single block whose coinbase reward pays to addr,
+// and waits for every registered wallet to sync.
+func (h *HarnessTest) MineBlockToAddress(addr address.Address) *wire.MsgBlock {
+	h.Helper()
+
+	if addr == nil {
+		h.Fatalf("nil address")
+	}
+
+	pkScript, err := txscript.PayToAddrScript(addr)
+	require.NoError(h, err, "failed to create coinbase pkscript")
+
+	_, height := h.GetBestBlock()
+	subsidy := blockchain.CalcBlockSubsidy(height+1, h.NetParams())
+
+	block, err := h.miner.GenerateAndSubmitBlockWithCustomCoinbaseOutputs(
+		nil, -1, time.Time{}, []wire.TxOut{{
+			Value:    subsidy,
+			PkScript: pkScript,
+		}},
+	)
+	require.NoError(h, err, "unable to mine block to address")
+	require.NotNil(h, block, "mined block is nil")
+
+	// Ensure all wallets we created in this test have caught up, matching
+	// the behavior of the other mining helpers.
+	for _, w := range h.ActiveWallets() {
+		h.AssertWalletSynced(w)
+	}
+
+	return block.MsgBlock()
+}
 
 // GenerateBlocks generates the specified number of blocks.
 func (h *HarnessTest) GenerateBlocks(num uint32) []*chainhash.Hash {
--- a/bwtest/harness_miner.go
+++ b/bwtest/harness_miner.go
@@ -49,6 +49,10 @@ const (
 	// contains only a coinbase transaction and a single non-coinbase
 	// transaction.
 	coinbaseAndOneTxn = 2
+
+	// MinerFeeRate is the fee rate, in satoshis per byte, used for
+	// miner-funded test transactions.
+	MinerFeeRate = btcutil.Amount(10)
 )
 
 // GenerateBlocks generates the specified number of blocks.
@@ -489,8 +493,17 @@ func (h *HarnessTest) SendOutput(output *wire.TxOut,
 
 	h.Helper()
 
-	txid, err := h.miner.SendOutputs([]*wire.TxOut{output}, feeRate)
-	require.NoError(h, err, "failed to send output")
+	return h.SendOutputs([]*wire.TxOut{output}, feeRate)
+}
+
+// SendOutputs sends funds from the miner in a single transaction.
+func (h *HarnessTest) SendOutputs(outputs []*wire.TxOut,
+	feeRate btcutil.Amount) *chainhash.Hash {
+
+	h.Helper()
+
+	txid, err := h.miner.SendOutputs(outputs, feeRate)
+	require.NoError(h, err, "failed to send outputs")
 
 	return txid
 }
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -1,12 +1,16 @@
 package bwtest
 
 import (
+	"bytes"
+	"fmt"
 	"strings"
 	"time"
 
+	"github.com/btcsuite/btcd/address/v2"
 	"github.com/btcsuite/btcd/btcutil/v2"
 	"github.com/btcsuite/btcd/txscript/v2"
 	"github.com/btcsuite/btcd/wire/v2"
+	"github.com/btcsuite/btcwallet/bwtest/wait"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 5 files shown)


bwtest/harness_miner.go
```diff
@@ -49,8 +53,66 @@ const (
 	// contains only a coinbase transaction and a single non-coinbase
 	// transaction.
 	coinbaseAndOneTxn = 2
+
+	// MinerFeeRate is the fee rate, in satoshis per byte, used for
+	// miner-funded test transactions.
+	//
+	// It is deliberately above 1 sat/byte. rpctest's memWallet derives the fee
+	// from the transaction size measured *before* its change output is
```
yyforyongyu: So this is a bug in `memWallet`?

itest/utxo_manager_test.go
```diff
@@ -0,0 +1,711 @@
...
+
+	output := &wire.TxOut{Value: int64(oneBTC), PkScript: pkScript}
+	txid := h.SendOutput(output, bwtest.MinerFeeRate)
+
+	// The wallet learns about the unconfirmed transaction asynchronously,
+	// so poll until it appears in the listing. The closure only uses locals:
+	// testify runs each tick in its own goroutine, so capturing outer
+	// variables would race, and EventuallyWithT reports the last tick's
+	// errors so a persistent ListUnspent failure names the real cause.
+	require.EventuallyWithT(
```
yyforyongyu: we have `wait.NoError`, think we should default to that unless it doesn't fit here.

itest/utxo_manager_test.go
```diff
@@ -0,0 +1,711 @@
...
+// unknownOutpoint returns an outpoint that was never mined and is therefore
+// unknown to any wallet.
+func unknownOutpoint() wire.OutPoint {
+	return wire.OutPoint{Hash: chainhash.Hash{0xaa}, Index: 0}
+}
+
+// testListUnspent verifies ListUnspent field enrichment, amount ordering,
+// account and confirmation filters, and the pre-start state gate.
+func testListUnspent(h *bwtest.HarnessTest) {
+	h.Helper()
```
yyforyongyu: same here, think we should remove it https://github.com/btcsuite/btcwallet/pull/1306#discussion_r3734034226

itest/utxo_manager_test.go
```diff
@@ -0,0 +1,711 @@
...
+	require.NoError(h, err, "failed to list unspent")
+	require.Len(h, utxos, 3, "unexpected UTXO count")
+
+	// Results are sorted by amount in ascending order.
+	require.Equal(h, btcutil.Amount(oneBTC), utxos[0].Amount)
+	require.Equal(h, btcutil.Amount(twoBTC), utxos[1].Amount)
+	require.Equal(h, btcutil.Amount(threeBTC), utxos[2].Amount)
+
+	for _, utxo := range utxos {
+		amount, ok := funded[utxo.OutPoint]
```
yyforyongyu, line 127: nit: create a new assertion like `h.AssertUTXOEqual`? only if we wanna assert it again in other places tho

itest/utxo_manager_test.go
```diff
@@ -0,0 +1,711 @@
...
+	)
+
+	// GetUtxo reports the same spendability for the same output.
+	utxo, err := w.GetUtxo(h.Context(), coinbase.OutPoint)
+	require.NoError(h, err, "failed to get coinbase utxo")
+	require.False(
+		h, utxo.Spendable, "immature coinbase reported spendable",
+	)
+
+	// NOTE: The complementary case — the same output becoming spendable once
```
yyforyongyu: 👍

itest/utxo_manager_test.go
```diff
@@ -0,0 +1,711 @@
...
+	h.Helper()
+
+	if h.Backend.Name() == "neutrino" {
+		h.Skip("neutrino cannot deliver unconfirmed transactions")
+	}
+
+	w := createWallet(h)
+	require.NoError(h, w.Start(h.Context()), "failed to start wallet")
+
+	const oneBTC = 1 * btcutil.SatoshiPerBitcoin
```
yyforyongyu: maybe we can make these global and reuse them

bwtest/harness_wallet.go
```diff
@@ -91,39 +93,158 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
+		})
+	}
 
-	return w
+	txid := h.SendOutputs(outputs, MinerFeeRate)
+
+	// Confirm the funding transaction and wait for every registered wallet
+	// to sync the mined block.
+	tx := h.AssertTxInMempool(*txid)
+	h.MineBlockWithTx(tx)
```
yyforyongyu, line 180: `MineBlockWithTx` only waits for `SyncedTo().Height`, which does not guarantee that the wallet’s UTXO state is visible yet. I reproduced `utxomanager_list_unspent` failing 6 of 10 identical runs here because the immediate `ListUnspent` returned zero outputs after `FundWallet`. Could we use `wait.NoError` before `FundWallet` returns to wait until every returned outpoint is observable, and similarly wait for the expected confirmation state after the later `MineBlocks` call?

yyforyongyu (changes requested):
Nice docs, easy to follow but also not overwhelming. Left a few comments.

bwtest/harness_miner.go
Abdulkbk: Not really, it's a test-only quirk in btcd. The agent just used it as a justification for choosing a higher value than 1, but the real reason is so transactions safely relay irrespective of the default core fee policy (which was initially 1 sat/vbyte).

itest/utxo_manager_test.go
Abdulkbk: Using it now

itest/utxo_manager_test.go
Abdulkbk: Removed.

Must have missed this when I reviewed the code 😶

bwtest/harness_wallet.go
Abdulkbk: Now using `wait.NoError`

itest/utxo_manager_test.go
```diff
@@ -0,0 +1,690 @@
...
+	// An expired lease is excluded from the listing. Lease the first output
+	// again with a short duration and poll until it lapses.
+	shortLease := wtxmgr.LockID{3}
+
+	_, err = w.LeaseOutput(
+		h.Context(), shortLease, outpoints[0], time.Second,
+	)
+	require.NoError(h, err, "failed to lease output with short duration")
+
+	require.Eventually(
```
yyforyongyu: same here re `wait.NoError`

itest/utxo_manager_test.go
```diff
@@ -0,0 +1,690 @@
...
+	require.Len(h, outpoints, 3, "unexpected funded outpoint count")
+
+	funded := make(map[wire.OutPoint]btcutil.Amount, len(outpoints))
+	for i, op := range outpoints {
+		funded[op] = amounts[i]
+	}
+
+	// The funding transaction is confirmed, so every output reports one
+	// confirmation.
+	utxos, err = w.ListUnspent(h.Context(), wallet.UtxoQuery{
```
yyforyongyu: we need `wait.NoError` here too

yyforyongyu (changes requested):
Left two comments that we should fix. There is a broader change we should make, will create a new task instead and fix it later. Atm let's keep the scope focused.

bwtest/harness_wallet.go
```diff
@@ -91,39 +95,199 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
 	})
 	require.NoError(h, err, "failed to unlock wallet")
+}
+
+// NewWalletAddress derives a fresh receive address from the wallet's default
+// account, unlocking the wallet and ensuring the account exists first.
+func (h *HarnessTest) NewWalletAddress(w *wallet.Wallet) address.Address {
+	h.Helper()
+
+	h.UnlockWallet(w)
```
GustavoStingelin: Do we need to unlock the wallet here? `Wallet.NewAddress` only requires a started wallet and should derive receiving addresses from public account material while locked. Implicitly unlocking changes the caller’s authentication state, leaves the wallet unlocked indefinitely, and can hide an unexpected lock transition that the test should expose. It also makes this helper harder to use for locked or watch-only wallet scenarios.

The operation that may actually require private-key access is the fallback account creation. Could we initialize the standard account explicitly during wallet setup, or limit unlocking to that fallback while preserving the wallet’s previous lock state? Then `NewWalletAddress` and `FundWallet` would not unexpectedly change the wallet’s authentication state.

bwtest/harness_wallet.go
```diff
@@ -91,39 +95,199 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
+// AssertUtxosExist polls until the wallet reports every outpoint as one of its
+// UTXOs.
+//
+// Mining helpers only wait for the wallet's synced height to reach the chain
+// tip. Reaching the tip does not imply the block's outputs have been committed
+// to the wallet's UTXO store, so a query issued immediately after mining can
+// still miss them. Callers that are about to read UTXO state must wait for the
+// outputs themselves rather than for the height alone.
+func (h *HarnessTest) AssertUtxosExist(w *wallet.Wallet,
+	outpoints ...wire.OutPoint) {
```
GustavoStingelin: Nit: could we rename this to `AssertUtxosVisible` and update the documentation accordingly? “Exist” looks ambiguous about whether the parent transaction is still unconfirmed or has already been processed from a block. This helper only guarantees that the wallet’s `GetUtxo` query can observe the outputs, regardless of their confirmation state, so “visible” describes the guarantee more precisely.

bwtest/harness_wallet.go
```diff
@@ -91,39 +95,199 @@ func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
...
+	_, err := w.GetAccount(h.Context(), scope, name)
+	if err == nil {
+		return
+	}
+
+	lookupErr := err
+
+	_, err = w.NewAccount(h.Context(), scope, name)
+	require.NoError(h, err,
+		"failed to create account %q (lookup error: %v)", name, lookupErr)
```
GustavoStingelin: `ensureAccount` currently treats every `GetAccount` error as if the account were missing and then attempts to create it. That could hide a backend/store failure and mutate wallet state after an unrelated lookup error. Could we either create the standard account explicitly during harness setup or only enter the creation path for a confirmed account-not-found error?

GustavoStingelin (changes requested):
I took a closer look at the harness changes because they introduce broader behavior that will be shared across the suite. Left few comments.

bwtest/harness_wallet.go
Abdulkbk: done

bwtest/harness_wallet.go
Abdulkbk: Nice catch, removed unlocking from `NewWalletAddress` and `FundWallet`.

bwtest/harness_wallet.go
Abdulkbk: Now using `ListAccountsByScope` instead.

itest/utxo_manager_test.go
```diff
@@ -0,0 +1,722 @@
...
+	}
+
+	// Every funded outpoint resolves with its funding metadata.
+	for _, f := range funded {
+		utxo, err := w.GetUtxo(h.Context(), f.op)
+		require.NoError(h, err, "failed to get funded utxo")
+		require.Equal(h, f.op, utxo.OutPoint, "outpoint mismatch")
+		require.Equal(h, f.amount, utxo.Amount, "amount mismatch")
+		require.Equal(
+			h, int32(1), utxo.Confirmations, "confirmation mismatch",
```
yyforyongyu, line 394: One thing about asserting these APIs is to think about the internal state of the wallet first. For instance, this still produces a flake, try,
```
for i in {1..10}; do
    make itest chain=btcd db=sqlite nocache=1 \
      GOTEST='GO111MODULE=on go test -run=^TestBtcWallet/btcd/utxomanager_get_utxo$$' \
      || break
  done
```

yyforyongyu (approved):
Left a comment, it's not blocking as I plan to fix it properly via a new PR at a deeper level.

GustavoStingelin (approved):
LGTM!

PR #1310, opened 2026-08-10
wallet: silence local goconst lint errors

GustavoStingelin:
My local environment kept reporting these `goconst` lint errors, although CI does not complain about them.
I spent some time trying to identify the difference, including running make lint through Docker, but couldn’t determine why they only appear locally. Fixing the literals required less effort than continuing the investigation, so this PR reuses the existing default account constant.


yyforyongyu (approved):
LGTM🧷 Simple PR gonna merge it now.

PR #1311, opened 2026-08-10
itest: cover AccountManager lifecycle and contract APIs

GustavoStingelin:
Add end-to-end integration coverage for the wallet `AccountManager` APIs using
real wallet stores.

The new scenarios cover:

- Derived account creation across every default key scope
- Account listing and lookup by scope and by name, with scope-local names
- Renaming: isolation from identically named accounts in other scopes, reuse of
  the freed name, and rejection of a rename onto a name taken in the same scope
- Reads that do not depend on lock state, and durability across a wallet reload
- State gates and rejected mutations, which must leave public state unchanged
- Reserved-account rules for `imported` and `default`
- Account numbers: per-scope allocation, advancement, and reload stability

### DIFF
--- a/itest/account_manager_test.go
+++ b/itest/account_manager_test.go
@@ -1,0 +1,615 @@
+//go:build itest
+
+package itest
+
+import (
+	"bytes"
+	"fmt"
+	"strconv"
+
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// accountManagerExpectation records the published identity of a derived
+// account, which is what every read surface is then required to report for it.
+//
+// Names are only unique within a key scope, so a row is identified by the name
+// and scope together rather than by the name alone.
+type accountManagerExpectation struct {
+	name      string
+	scope     waddrmgr.KeyScope
+	publicKey []byte
+}
+
+// String identifies an expectation in assertion messages so that a failure
+// raised inside a loop over several accounts names the one it came from.
+func (e accountManagerExpectation) String() string {
+	return fmt.Sprintf("account %q in scope %s", e.name, e.scope)
+}
+
+// defaultKeyScopes are the key scopes a wallet initializes on creation. Every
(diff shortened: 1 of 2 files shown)


itest/account_manager_test.go
```diff
@@ -0,0 +1,1096 @@
...
+// raised inside a loop over several accounts names the one it came from.
+func (e accountManagerExpectation) String() string {
+	return fmt.Sprintf("account %q in scope %s", e.name, e.scope)
+}
+
+// defaultKeyScopes are the key scopes a wallet initializes on creation. Every
+// derived-account case runs against all four so a scope-specific regression,
+// such as the taproot schema BIP0086 resolves through its own scope mapping,
+// cannot hide behind the two scopes that happen to be exercised elsewhere.
+var defaultKeyScopes = []waddrmgr.KeyScope{
```
yyforyongyu, line 38: we should use `waddrmgr.DefaultKeyScopes` to make sure we only have a single source of truth.

itest/account_manager_test.go
```diff
@@ -0,0 +1,1096 @@
...
+	waddrmgr.KeyScopeBIP0044,
+	waddrmgr.KeyScopeBIP0049Plus,
+	waddrmgr.KeyScopeBIP0084,
+	waddrmgr.KeyScopeBIP0086,
+}
+
+// testAccountManagerLifecycle verifies derived account creation across every
+// default key scope, scope-local names, rename isolation, read results that do
+// not depend on lock state, and durable state after a fresh wallet reload.
+func testAccountManagerLifecycle(h *bwtest.HarnessTest) {
```
yyforyongyu, line 48: `testAccountManagerLifecycle` uses,
  → `requireAccountManagerState`
  → `requireDerivedAccountsByName`
  → `requireDerivedAccountKeyEqual`
  → `parseDerivedAccountKey`

This is hard to follow as I must jump through several functions to understand one assertion. We should perform more linear assertion here so it's easier to follow.

yyforyongyu (changes requested):
I don't think this PR is following the plan, as we wanna assert one method in one test clearly, similar to what we've done in #1307 or #1300. I paused the review after checking the first methods, and it needs more work.

GustavoStingelin:
I’ve been working on this since last week, which is why it is less “plan like.”

Because of that, the approach here is structurally different from what the plan proposes. I’ll close this PR and open new ones following the plan statements, since that will be easier than reworking the current changeset.

Thanks for reviewing.

PR #1312, opened 2026-08-10
itest: cover AccountManager account imports

GustavoStingelin:
Stacked on #1311, which this PR depends on for its wallet setup and assertion
helpers. Review and merge that one first; the base changes to `sql-wallet` once
it lands.

Adds end-to-end coverage for importing accounts through `AccountManager` 
from deterministic keys:

- Dry-run imports: reported row matches the persisted one, and the transaction
  is rolled back so nothing stays queryable
- Persisted imports: published metadata through every read surface, and
  durability across a wallet reload
- Rejections, each required to fail as `ErrInvalidAccountKey` and leave no row:
  nil, private, wrong depth, non-hardened, and wrong-network keys
- Duplicate account names within the derived scope
- Scope routing driven by the key's version bytes and the requested address
  type, with the resulting address schema surviving a reload
- Watch-only shells rejecting derived account creation without a partial row
- The spendable-wallet import boundary, where the two backends disagree, so
  only the invariant both must honor is asserted: no partial account state

### DIFF
--- a/itest/account_manager_import_test.go
+++ b/itest/account_manager_import_test.go
@@ -533,6 +533,64 @@ func testAccountManagerImportRouting(h *bwtest.HarnessTest) {
 	}
 }
 
+// testAccountManagerImportSpendable pins what a spendable wallet does when an
+// xpub-only account is imported into it.
+//
+// ADR 0012 states that a wallet is uniformly watch-only or uniformly spendable,
+// and the two backends disagree on enforcing it for this operation: sqlite
+// rejects the import outright, while kvdb accepts it and publishes the account
+// as imported and watch-only inside an otherwise spendable wallet. Until that
+// is settled, the assertion covers the part both must honor either way, which
+// is that no partial account state is left behind: the import either fails and
+// publishes nothing, or succeeds and is fully queryable.
+func testAccountManagerImportSpendable(h *bwtest.HarnessTest) {
+	const importName = "account manager spendable import"
+
+	ctx := h.Context()
+	keys := deterministicImportAccountKeys(h)
+	_, _, w := newAccountManagerWallet(h)
+
+	require.False(h, w.IsWatchOnly(), "the wallet is not spendable")
+
+	imported, err := w.ImportAccount(
+		ctx, importName, keys.candidate, keys.fingerprint,
+		waddrmgr.WitnessPubKey, false,
+	)
+	if err != nil {
+		requireAccountManagerNameAbsent(
+			h, w, importName, "rejected import into a spendable wallet",
+		)
+
+		return
+	}
+
+	// The import was accepted, so the account has to be complete and reachable
+	// through every read surface.
+	require.NotNil(h, imported, "the accepted import returned no account")
+	require.True(
+		h, imported.IsImported,
+		"the imported account is not reported as imported",
+	)
+	require.True(
+		h, imported.IsWatchOnly,
+		"an xpub-only account in a spendable wallet is not watch-only",
+	)
+	require.Nil(
+		h, imported.AccountNumber,
+		"the imported account was given a derived account number",
+	)
+	require.Equal(
+		h, []byte(keys.candidate.String()), imported.PublicKey,
+		"the imported account did not round-trip its key",
+	)
+
+	requireImportedAccountState(h, w, importAccountExpectation{
+		name:        importName,
+		publicKey:   bytes.Clone([]byte(keys.candidate.String())),
+		fingerprint: keys.fingerprint,
+	})
+}
+
 // importRoutingKey derives a distinct account-level public key and stamps it
 // with the requested HD version.
 //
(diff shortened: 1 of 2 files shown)


yyforyongyu (changes requested):
Thanks for adding this coverage. I am pausing the detailed review because this PR inherits the same structural problems as #1311, and I think the stack needs to be reorganized before line-by-line review will be productive.

Please treat the following as review requirements, not optional suggestions.

## Declare the roadmap scope explicitly

Please update the PR title or description to state exactly which roadmap task or tasks this PR claims to complete. A PR may complete one task or multiple tasks, but the signal must be unambiguous. Please include links to the claimed tasks and map their acceptance criteria to the tests in this PR.

If this is only part of a task, say that explicitly instead of presenting the PR as completing it. Identify what is covered here, what remains, and which PR will provide the remaining coverage. If this PR is stacked on #1311, also state which requirements are owned by #1311 and which are owned here.

At present, the intended roadmap result is difficult to determine:

- #1311 mixes NewAccount and RenameAccount mutation coverage with broad query coverage, reserved-account behavior, lock behavior, and account-number behavior.
- This PR adds ImportAccount separately and depends on the broad setup and assertion helpers introduced by #1311.
- The boundaries between completed requirements, partial coverage, and incidental behavior are not visible from the PR structure.

Please reorganize the stack around the declared roadmap acceptance criteria. If a single PR claims multiple tasks, it must complete each named task and still keep the tests separated by public API method.

## Follow the integration-test style established in #1307

Please use #1307 as the style precedent.

My preference is one registered integration test per public API method. For AccountManager, the intended cases are:

Mutation APIs:

- `account manager new account`
- `account manager rename account`
- `account manager import account`

Query APIs:

- `account manager list accounts`
- `account manager list accounts by scope`
- `account manager list accounts by name`
- `account manager get account`

“One test per API method” means one method is the subject of each registered test. The test may use another API for fixture setup or a narrow postcondition, but it should not start validating that other API's complete contract.

For example, an ImportAccount test may call GetAccount after a reload to prove that the imported account persisted. It should not then validate ListAccounts, ListAccountsByScope, ListAccountsByName, and GetAccount together. Those query contracts belong to their own method-specific tests.

A method-specific test may cover several variants of that method, as #1307 does. For example, the ImportAccount test can cover dry run, successful import, invalid keys, duplicate names, lifecycle gates, and persistence. Those cases should remain in a clear linear flow or use a data-only table where the setup structure is identical.

## Keep the tests linear

The public API call and its assertions should be visible in the test body.

Please avoid assertion chains such as:

`testAccountManagerImportAccount` → `requireImportedAccountState` → several public query methods → repeated assertion helpers

The same problem exists in #1311:

`testAccountManagerLifecycle` → `requireAccountManagerState` → `requireDerivedAccountsByName` → `requireDerivedAccountKeyEqual` → `parseDerivedAccountKey`

A reviewer should not have to jump through several functions to understand what one API call is expected to return.

Helpers are appropriate for:

- constructing a wallet fixture;
- constructing deterministic keys;
- reopening a Wallet and Manager;
- waiting for asynchronous state;
- pure value transformations, such as normalizing a serialized account key.

Helpers should not:

- call several public AccountManager methods;
- assert a complete scenario;
- hide the API method under test;
- duplicate the contracts owned by another test.

For key comparisons, a value-producing helper such as `canonicalAccountKey` would be preferable to `requireDerivedAccountKeyEqual`. The actual assertion should remain visible:

```go
wantKey := canonicalAccountKey(h, importedKey)

account, err := w.GetAccount(ctx, scope, name)
require.NoError(h, err)
require.Equal(h, name, account.AccountName)
require.Equal(h, scope, waddrmgr.KeyScope(account.KeyScope))
require.Equal(h, wantKey, canonicalAccountKey(h, account.PublicKey))
```

## Keep test tables data-only

When several inputs exercise the same method with the same setup, a table-driven test is appropriate. The table should contain only data: name, input, expected result, and expected error.

Do not put setup or execution closures such as `func() error` into table rows. That hides the operation under test and conflicts with the repository's data-only test-table guidance.

If different cases require materially different setup, keep them as explicit linear phases or separate registered tests.

## Keep comments useful

The comments in #1307 are a good model. Comments should explain why a setup step or assertion is necessary, for example:

- why a wallet must be locked or stopped before the call;
- why a control value prevents a vacuous pass;
- why a retry is required for asynchronous state;
- why reopening proves durability rather than an in-memory result;
- why a specific error classification is part of the public contract.

Please avoid comments that merely narrate the next line or compensate for an overly indirect helper structure.

## Do not weaken backend-neutral contracts

The integration tests should assert one public AccountManager contract across the supported stores.

`testAccountManagerImportSpendable` currently accepts either success or failure depending on the backend. That characterizes a disagreement, but it does not define or test a stable public contract. Similarly, matching `watch(ing)?-only` against an error string depends on backend implementation text rather than a stable error identity.

Please use `errors.Is` with a public sentinel when the API defines one. If the backends disagree and no stable public contract exists, call that discrepancy out separately rather than writing an integration test that accepts both outcomes and presenting it as contract coverage.

The same applies to #1311's default-account test: the AccountManager documentation says default accounts exist in every default scope, while the test treats “none exist” as valid and skips the assertions. The test should not silently weaken the documented contract.

## Use existing sources of truth

Do not redeclare the default scopes. Use:

```go
waddrmgr.DefaultKeyScopes
```

If a test requires a private copy to prevent mutation, clone that slice rather than repeating its contents.

An independent test oracle is different from duplicated production configuration. For example, deterministic HD-version test vectors can be appropriate when they intentionally verify production routing rather than derive their expected result from the code under test. Please keep that distinction explicit in the comments.

## Keep mutation coverage narrow

The mutation tests should cover:

- NewAccount success, maintained rejection cases, lifecycle gates, and reload;
- RenameAccount success, maintained rejection cases, lifecycle gates, and reload;
- ImportAccount success, dry run, maintained invalid-input cases, duplicate identity/name, lifecycle gates, and reload;
- rejected mutations leaving no partial public account;
- real SQLite Wallet operation with backend-neutral registration.

Use GetAccount only as the narrow mutation postcondition where needed.

Do not include:

- full query-surface consistency checks;
- exact account-number allocation behavior;
- unrelated query filtering or ordering;
- production fixes;
- tests that accept contradictory backend outcomes.

The watch-only NewAccount case belongs in the NewAccount test, not in an ImportAccount-specific PR. Import routing may be a data-only ImportAccount case if it is part of the maintained current contract, but it should not become a separate broad scenario with query-contract assertions.

## Keep query coverage separate

The query tests should build a deterministic fixture through the public mutation APIs and then test each query method independently.

The fixture should contain the account shapes required by the task, including created and imported accounts and the documented default-account behavior.

Each query test should cover only its own API's relevant success, absence, filter, lifecycle, and reopen behavior. The query suite should compare the complete public view before and after reopening rather than only checking that selected expected rows are present.

GustavoStingelin:
I’ve been working on this since last week, which is why it is less “plan like.”

Because of that, the approach here is structurally different from what the plan proposes. I’ll close this PR and open new ones following the plan statements, since that will be easier than reworking the current changeset.

Thanks for reviewing.

PR #1313, opened 2026-08-11
wallet: define semantic account identity

yyforyongyu:
## Summary

Implements Task 341 — define semantic account identity.

## Change Description

Add wallet-owned `AccountNumber` and `MasterFingerprint` types plus an opaque
`AccountSelector` with constructors for name and number selection. Keep shape
validation package-private and leave key-scope support and account existence
checks to consuming operations and the Store.

### DIFF
--- a/wallet/account_identity.go
+++ b/wallet/account_identity.go
@@ -1,0 +1,65 @@
+// Copyright (c) 2026 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package wallet
+
+import (
+	"errors"
+
+	"github.com/btcsuite/btcwallet/waddrmgr"
+)
+
+var errInvalidAccountSelector = errors.New(
+	"exactly one of account name or account number must be provided",
+)
+
+// AccountNumber is the BIP44 account index within a key scope. A nil pointer
+// means the account has no BIP44 number, while a non-nil zero identifies
+// account zero.
+type AccountNumber uint32
+
+// MasterFingerprint is the BIP32 master key fingerprint. A nil pointer means
+// the fingerprint is absent, while a non-nil zero is a present-zero
+// fingerprint.
+type MasterFingerprint uint32
+
+// AccountSelector identifies an account by its portable wallet semantics.
+// Callers construct selectors with NewAccountSelectorByName or
+// NewAccountSelectorByNumber.
+type AccountSelector struct {
+	keyScope      waddrmgr.KeyScope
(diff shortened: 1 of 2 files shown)


GustavoStingelin (approved):

wallet/account_identity.go
```diff
@@ -0,0 +1,50 @@
...
+
+package wallet
+
+import (
+	"errors"
+
+	"github.com/btcsuite/btcwallet/waddrmgr"
+)
+
+var errInvalidAccountSelector = errors.New(
```
Abdulkbk, line 13: The error is returnede in `Validate(), an exported method, we should export it too.

wallet/account_identity.go
```diff
@@ -0,0 +1,50 @@
...
+// means the account has no BIP44 number, while a non-nil zero identifies
+// account zero.
+type AccountNumber uint32
+
+// MasterFingerprint is the BIP32 master key fingerprint. A nil pointer means
+// the fingerprint is absent, while a non-nil zero is a present-zero
+// fingerprint.
+type MasterFingerprint uint32
+
+// AccountSelector identifies an account by its portable wallet semantics. A
```
Abdulkbk: But `Validate()` does not inspect KeyScope. Either we do that or we update add a comment why.

nit: `... contains a KeyScope...` instead of `a key scope`, just like the other 2 fields.

Abdulkbk (changes requested):
Left some suggestions

wallet/account_identity.go
yyforyongyu: Why do we need to inspect the key scope here? an error will be returned by the db when it's not found?

wallet/account_identity.go
yyforyongyu: make sense, making the method private instead!

yyforyongyu: Resolved in ce0349206b: AccountSelector.validate and errInvalidAccountSelector are both package-private. External callers construct selectors through NewAccountSelectorByName or NewAccountSelectorByNumber, and consuming wallet operations can invoke validate internally.

wallet/account_identity.go
yyforyongyu: Resolved in ce0349206b: validate now documents that it checks selector shape only. KeyScope support and account existence remain the consuming operation and Store responsibility, preserving the database as the authoritative owner of those errors.

Abdulkbk (approved):
LGTM

PR #1314, opened 2026-08-11
wallet: enforce synchronization test boundary

yyforyongyu:
## Summary

Implements roadmap Task 433 by making `Wallet.Info().Synced` a reliable boundary for block-derived wallet assertions.

## Change Description

Report the committed wallet sync tip, publish connected-block progress only after the filtered update commits, and make the harness wait for both `Synced` and the miner’s exact tip. UTXO integration checks now perform one-shot reads after that boundary instead of retrying wallet results.

### DIFF
--- a/bwtest/harness_assertions.go
+++ b/bwtest/harness_assertions.go
@@ -15,7 +15,8 @@ var (
 )
 
 // AssertWalletSynced polls until the wallet reports it is synced to the
-// miner's best known height.
+// miner's exact best block. Once this returns, wallet reads for that block may
+// be asserted without another retry.
 func (h *HarnessTest) AssertWalletSynced(w *wallet.Wallet) {
 	h.Helper()
 
@@ -24,16 +25,23 @@ func (h *HarnessTest) AssertWalletSynced(w *wallet.Wallet) {
 	}
 
 	err := wait.NoError(func() error {
-		syncedTo := w.SyncedTo()
+		info, err := w.Info(h.Context())
+		if err != nil {
+			return fmt.Errorf("get wallet info: %w", err)
+		}
 
-		_, bestHeight, err := h.miner.Client.GetBestBlock()
+		bestHash, bestHeight, err := h.miner.Client.GetBestBlock()
 		if err != nil {
 			return fmt.Errorf("get best block: %w", err)
 		}
 
-		if syncedTo.Height != bestHeight {
-			return fmt.Errorf("%w: wallet=%d chain=%d", ErrWalletNotSynced,
-				syncedTo.Height, bestHeight)
+		if !info.Synced || info.SyncedTo.Height != bestHeight ||
+			!info.SyncedTo.Hash.IsEqual(bestHash) {
+
+			return fmt.Errorf("%w: synced=%v wallet=%v:%d "+
+				"chain=%v:%d", ErrWalletNotSynced, info.Synced,
+				info.SyncedTo.Hash, info.SyncedTo.Height, bestHash,
+				bestHeight)
 		}
 
 		return nil
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -470,10 +470,20 @@ func (w *Wallet) ChangePassphrase(ctx context.Context,
 // and dynamic synchronization state.
 //
 // This is part of the Controller interface.
-func (w *Wallet) Info(_ context.Context) (*Info, error) {
+func (w *Wallet) Info(ctx context.Context) (*Info, error) {
 	err := w.state.validateStarted()
 	if err != nil {
 		return nil, err
+	}
+
+	walletInfo, err := w.store.GetWallet(ctx, w.cfg.Name)
+	if err != nil {
+		return nil, fmt.Errorf("get wallet info: %w", err)
+	}
+
+	syncedTo, err := db.OptionalBlockStampFromBlock(walletInfo.SyncedTo)
+	if err != nil {
+		return nil, fmt.Errorf("decode wallet sync tip: %w", err)
 	}
 
 	info := &Info{
@@ -482,7 +492,7 @@ func (w *Wallet) Info(_ context.Context) (*Info, error) {
 		ChainParams:      w.cfg.ChainParams,
 		Locked:           !w.state.isUnlocked(),
 		Synced:           w.state.isSynced(),
-		SyncedTo:         w.SyncedTo(), //nolint:contextcheck // Legacy API.
+		SyncedTo:         syncedTo,
 		IsRecoveryMode:   w.state.isRecoveryMode(),
 		RecoveryProgress: 0,
 	}
--- a/wallet/syncer_test.go
+++ b/wallet/syncer_test.go
@@ -923,12 +923,9 @@ func TestHandleChainUpdate(t *testing.T) {
 		store, walletID,
 	)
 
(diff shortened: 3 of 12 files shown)


bwtest/harness_assertions_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/btcsuite/btcd/chainhash/v2"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// TestWalletSynced verifies the public sync flag, hash, and height must all
+// match the backend tip.
+func TestWalletSynced(t *testing.T) {
```
yyforyongyu: we should not write test code to test `harness_assertions.go`? why having that? we will end up in a loop of `writing test to check test code`?

bwtest/harness_assertions.go
```diff
@@ -24,21 +26,36 @@ func (h *HarnessTest) AssertWalletSynced(w *wallet.Wallet) {
...
+			info.SyncedTo.Height, bestHash, bestHeight)
 	}, defaultTestTimeout)
 	if err != nil {
 		h.Fatalf("wallet sync timeout: %v", err)
 	}
 }
+
+// walletSynced reports whether the public wallet state matches the backend
+// tip.
+func walletSynced(info *wallet.Info, bestHash *chainhash.Hash,
```
yyforyongyu: remove this thin wrapper, it's useless.

bwtest/harness_miner.go
```diff
@@ -369,27 +369,45 @@ func (h *HarnessTest) SpawnTempMiner() *HarnessTest {
...
+
+	return th
+}
+
+// WaitForMinerSync waits for an external miner peer to advertise the harness
+// miner's exact best hash and height.
+func (h *HarnessTest) WaitForMinerSync(peer *HarnessTest) {
+	h.Helper()
+
+	if peer == nil || peer.miner == nil {
```
yyforyongyu: this is a test, so in the test, there is no need to assert nil - we should just let it panic instead.

itest/controller_test.go
```diff
@@ -167,6 +173,149 @@ func testControllerInfo(h *bwtest.HarnessTest) {
 	)
 }
 
+// testControllerSyncBoundary verifies that one exact synchronization boundary
+// makes every wallet-owned view of a newly mined payment immediately visible.
+func testControllerSyncBoundary(h *bwtest.HarnessTest) {
+	w := createWallet(h)
+	require.NoError(h, w.Start(h.Context()), "failed to start wallet")
+	h.AssertWalletSynced(w)
```
yyforyongyu: These three lines will be commonly used in our later test? maybe create a new method `createAndStartWallet` and perform the assertion inside that method?

itest/controller_test.go
```diff
@@ -167,6 +173,149 @@ func testControllerInfo(h *bwtest.HarnessTest) {
...
+
+		outpoint = wire.OutPoint{
+			Hash:  *txid,
+			Index: uint32(i),
+		}
+
+		break
+	}
+
+	require.Equal(h, *txid, outpoint.Hash, "funding output not found")
```
yyforyongyu: why do we even need this assertion? what properties or invariants does assert?

bwtest/harness_miner.go
```diff
@@ -369,27 +369,45 @@ func (h *HarnessTest) SpawnTempMiner() *HarnessTest {
...
+
+	// Disconnect the temp miner so it can mine an alternative chain.
+	h.DisconnectFromMiner(th)
+
+	return th
+}
+
+// WaitForMinerSync waits for an external miner peer to advertise the harness
+// miner's exact best hash and height.
+func (h *HarnessTest) WaitForMinerSync(peer *HarnessTest) {
```
yyforyongyu: i don't understand why we need to refactor these area when it's unrelated?

bwtest/harness_assertions.go
```diff
@@ -14,8 +15,9 @@ var (
 	ErrWalletNotSynced = errors.New("wallet not synced")
 )
 
-// AssertWalletSynced polls until the wallet reports it is synced to the
-// miner's best known height.
+// AssertWalletSynced polls until the wallet reports the miner's exact best
```
yyforyongyu: This method should be renamed to `WaitForWalletSynced` based on our new rule?

itest/controller_test.go
```diff
@@ -167,6 +173,149 @@ func testControllerInfo(h *bwtest.HarnessTest) {
...
+		MinConfs: 0,
+		MaxConfs: maxConfsLimit,
+	})
+	require.NoError(h, err, "failed to list synchronized outputs")
+	require.Len(h, utxos, 1, "unexpected synchronized output count")
+	require.Equal(h, outpoint, utxos[0].OutPoint, "listed outpoint mismatch")
+	require.Equal(h, int32(1), utxos[0].Confirmations,
+		"listed confirmation mismatch")
+
+	utxo, err := w.GetUtxo(h.Context(), outpoint)
```
yyforyongyu: I highly doubt that we even need this test - the invariant of the `Synced` is constantly being checked in other tests already, e.g.,
- when creating the test wallet, we return the wallet only after the wallet is synced
- then in other tests like test list utxos or get utxo, if the `Synced` invariant is violated, those tests will fail

The only test i see suitable is, check that the `Synced` will become false when there is a new block. Also, this test is checking too many things, a violation of one test one API.

wallet/controller.go
```diff
@@ -467,22 +467,40 @@ func (w *Wallet) ChangePassphrase(ctx context.Context,
...
+	}
+
+	synced := w.state.isSynced()
+
+	syncedTo, err := w.syncedTo(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	synced = synced && w.state.isSynced() &&
```
yyforyongyu: why do we need all these extra checks, won't `w.state.isSynced()` be enough??

wallet/syncer_test.go
```diff
@@ -1702,6 +1703,57 @@ func TestProcessFilteredBlockUsesStore(t *testing.T) {
...
+	tests := []struct {
+		name         string
+		current      int32
+		notification int32
+		apply        bool
+	}{
+		{name: "stale", current: 102, notification: 101},
+		{name: "duplicate", current: 102, notification: 102},
+		{name: "gap", current: 100, notification: 102},
+		{name: "next", current: 100, notification: 101, apply: true},
```
yyforyongyu: there really is no need to force every test case using the table driven? if only this case has `apply` to be true, why not just create a standalone test?

wallet/syncer_test.go
```diff
@@ -1702,6 +1703,57 @@ func TestProcessFilteredBlockUsesStore(t *testing.T) {
...
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			t.Parallel()
+
+			store := &walletmock.Store{}
+			expectSyncedTip(store, waddrmgr.BlockStamp{
+				Height: test.current,
+			})
+
+			if test.apply {
```
yyforyongyu: if we remove `test.apply` then we can remove the crazy nesting below

GustavoStingelin (approved):
LGTM

Abdulkbk (approved):
LGTM

PR #1315, opened 2026-08-11
chain: synchronize bitcoind event tests

yyforyongyu:
## Summary

Fix a long-running `TestBitcoindEventsRPC/NotifyBlocks` flake where the test
expects `FilteredBlockConnected` but receives `BlockDisconnected`.

The fixture starts bitcoind while it is still synchronizing with the miner.
Previously, it constructed and started the event source before verifying that
bitcoind had reached the miner's tip. The RPC poller could therefore snapshot
an older height while a newly started client independently sampled a newer
best block. Replaying the intervening synchronization blocks into that client
made its already-known tip look like a competing chain transition, causing the
reorganization path to emit `BlockDisconnected` before the notification for
the block generated by the test. Constructing the ZMQ source early had the
equivalent risk of queuing synchronization blocks on its subscriptions.

## Change Description

Snapshot the miner's best hash and height, then use a temporary RPC client to
wait until bitcoind matches that exact tip before constructing either event
source. This gives the RPC poller and each client the same stable initial chain
view and ensures the ZMQ subscriptions cannot accumulate historical sync
events.

Record the block generated after `NotifyBlocks` and require both
`FilteredBlockConnected` and `BlockConnected` to reference that exact hash.
This directly verifies the notification contract without retries, suppressed
events, or weakened ordering assertions.

The failure reproduced under the CI-equivalent coverage command before the
change. After the fix, 20 repeated RPC coverage runs, 20 repeated ZMQ runs,
the full `chain` package, and repository lint checks pass.

### DIFF
--- a/chain/bitcoind_events_test.go
+++ b/chain/bitcoind_events_test.go
@@ -99,11 +99,10 @@ func runBitcoindEventsTests(t *testing.T, rpcPolling bool) {
 		t.Run(test.name, func(t *testing.T) {
 			// Initialize a fresh miner for the test case.
 			miner1 := setupMiner(t)
-			addr := miner1.P2PAddress()
 
 			// Initialize a fresh bitcoind client for EVERY test
 			// case.
-			btcClient := setupBitcoind(t, addr, rpcPolling)
+			btcClient := setupBitcoind(t, miner1, rpcPolling)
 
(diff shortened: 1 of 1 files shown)


GustavoStingelin (approved):
LGTM

PR #1316, opened 2026-08-11
itest: add TxCreator integration tests

Abdulkbk:
Implements Task 434 — cover transaction creation contracts.

## Task 434 acceptance criteria mapped to tests

| Criterion | Case |
| --- | --- |
| Automatic coin selection | `txcreator select coins` (named account, largest-first), `txcreator default account` (no source named, resolved under BIP86) |
| Caller-selected inputs | `txcreator manual inputs` (selection bypassed), `txcreator coin source` (restricted candidate list) |
| Transactions with and without change | `txcreator select coins` (change to an internal address), `txcreator omit change` (dust remainder dropped) |
| Representative fee and output boundaries | `txcreator output boundaries` (dust edge derived from the relay policy, smallest payable amount, `MaxSatoshi` and one above it), `txcreator reject intent` (zero, maximum and above-maximum fee rates; negative output), `txcreator multiple outputs` |
| Insufficient funds | `txcreator reject inputs` (over balance, and a leased coin skipped by selection), `txcreator coin source` (confirmation bound) |
| Locked or foreign inputs | `txcreator reject inputs` (leased, unknown and on-chain foreign outputs, through both manual and candidate paths) |
| Invalid outputs | `txcreator reject intent` |
| Bad change source | `txcreator reject intent` (unnamed and unknown change accounts) |
| Stopped and locked wallet states | `txcreator wallet state` (forbidden before Start and after Stop; a locked wallet still authors) |
| Watch-only wallet state | Not covered — see below |
| Backend neutral for PostgreSQL | Full suite passes on SQLite, PostgreSQL and kvdb |
| Test-only, defines no future behavior | No production file is touched |

## Harness changes

All additive, in `bwtest`, with the existing entry points kept as wrappers:

- `WalletFixture` and `NewWallet` — one parameterized wallet preparation path (address type, funding amounts, unlock, unstarted), so component tests stop growing their own creation and funding policy.
- `FundWalletOfType` and `NewWalletAddressOfType` — funding under a chosen key scope, which the implicit default-account path needs, since the wallet resolves accounts per scope while the harness funded BIP84 only.
- `ForeignOutpoint` — the miner's change output as a property of the funding fixture; `testGetUtxo` now uses it instead of deriving it inline.
- Shared fixtures moved to `itest/fixtures_test.go` so no component's test file owns values another depends on.

Two existing-test changes follow from those, so each policy has one owner:
`testGetUtxo` uses `ForeignOutpoint` rather than its own inline copy, and the
seven UtxoManager cases drop their local `createWallet` for
`NewWallet(WalletFixture{Unstarted: true})`.

## Synchronization boundary

The final commit removes the explicit `AssertWalletSynced` calls from the
cases. Mining already asserts that every registered wallet is synced, which
funding does, so the calls asserted nothing further; over full-suite
PostgreSQL runs the known scan-completeness failure appeared at the same rate
either way, with the numbers recorded in that commit's body. Reverting that
one commit restores them.

## not covered

- **Watch-only state.** A funded watch-only wallet requires shell mode with public initial accounts, which #1320 currently restricts.

Part of #1250.

### DIFF
--- a/bwtest/README.md
+++ b/bwtest/README.md
@@ -34,28 +34,36 @@ The `bitcoind` backend uses ZMQ for block/tx notifications.
 
 ## Wallet Helpers
 
-`bwtest` includes convenience helpers for tests that do not want to directly
-exercise the wallet manager:
-
-- `(*HarnessTest).CreateEmptyWallet`
-- `(*HarnessTest).CreateFundedWallet`
-- `(*HarnessTest).FundWallet`
-
-Example usage:
+`bwtest` owns wallet creation, funding and lock policy so component tests do
+not each define their own. `(*HarnessTest).NewWallet` is the single
+parameterized entry point; it takes a `WalletFixture` describing what the case
+needs and returns the wallet with the outpoints funding produced:
 
 ```go
 func testFoo(t *bwtest.HarnessTest) {
-	t.CreateEmptyWallet()
+	w, outpoints := t.NewWallet(bwtest.WalletFixture{
+		AddrType: waddrmgr.WitnessPubKey,
+		Amounts:  []btcutil.Amount{oneBTC, twoBTC},
+		Unlocked: true,
+	})
 
-	// Now add tests that need a started wallet instance.
-}
-
-func testBar(t *bwtest.HarnessTest) {
-	t.CreateFundedWallet()
-
-	// Now add tests that need a wallet with spendable funds.
+	// Now add tests that need a wallet with two spendable coins.
 }
 ```
+
+`WalletFixture` carries the funding address type, the funding amounts, whether
+to unlock, and whether to leave the wallet unstarted. A case that selects the
+funded coins derives its key scope from the same `AddrType` with `KeyScope()`,
+so the funded scope has one authority.
+
+Two convenience wrappers remain for cases that need nothing else:
+
+- `(*HarnessTest).CreateEmptyWallet`
+- `(*HarnessTest).CreateFundedWallet`
+
+Funding is also available on its own through `(*HarnessTest).FundWallet` and
+`(*HarnessTest).FundWalletOfType`, and addresses through
+`(*HarnessTest).NewWalletAddress` and `(*HarnessTest).NewWalletAddressOfType`.
 
 Manager-focused tests should continue to create wallets through the manager API
 directly.
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -131,35 +131,63 @@ func (h *HarnessTest) UnlockWallet(w *wallet.Wallet) {
 	require.NoError(h, err, "failed to unlock wallet")
 }
 
-// NewWalletAddress derives a fresh receive address from the wallet's default
-// account, ensuring the account exists first.
+// NewWalletAddress derives a fresh receive address of the default funding
+// type from the wallet's default account, ensuring the account exists first.
 func (h *HarnessTest) NewWalletAddress(w *wallet.Wallet) address.Address {
 	h.Helper()
 
-	scope, err := fundingAddrType.KeyScope()
-	require.NoError(h, err, "failed to resolve funding address scope")
+	return h.NewWalletAddressOfType(w, fundingAddrType)
+}
+
+// NewWalletAddressOfType derives a fresh receive address of the requested type
+// from the wallet's default account in that type's key scope, ensuring the
+// account exists first.
+//
+// Callers that need funds under a particular key scope must use this instead
+// of NewWalletAddress, because the wallet resolves accounts per scope: coins
+// held under one scope's default account are invisible to selection from
+// another.
+func (h *HarnessTest) NewWalletAddressOfType(w *wallet.Wallet,
+	addrType waddrmgr.AddressType) address.Address {
+
+	h.Helper()
+
+	scope, err := addrType.KeyScope()
+	require.NoError(h, err, "failed to resolve address scope")
 
 	h.ensureAccount(w, scope, waddrmgr.DefaultAccountName)
 
 	addr, err := w.NewAddress(
-		h.Context(), waddrmgr.DefaultAccountName, fundingAddrType, false,
+		h.Context(), waddrmgr.DefaultAccountName, addrType, false,
 	)
 	require.NoError(h, err, "failed to create address")
 
 	return addr
 }
 
-// FundWallet pays one output per amount to fresh wallet addresses in a single
-// miner transaction, confirms it, and returns the wallet outpoints in the
-// order of the amounts argument.
+// FundWallet pays one output per amount to fresh wallet addresses of the
+// default funding type in a single miner transaction, confirms it, and returns
+// the wallet outpoints in the order of the amounts argument.
 func (h *HarnessTest) FundWallet(w *wallet.Wallet,
+	amounts ...btcutil.Amount) []wire.OutPoint {
+
+	h.Helper()
+
+	return h.FundWalletOfType(w, fundingAddrType, amounts...)
+}
+
+// FundWalletOfType pays one output per amount to fresh wallet addresses of the
+// requested address type in a single miner transaction, confirms it, and
+// returns the wallet outpoints in the order of the amounts argument.
+func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
+	addrType waddrmgr.AddressType,
 	amounts ...btcutil.Amount) []wire.OutPoint {
 
 	h.Helper()
 
 	outputs := make([]*wire.TxOut, 0, len(amounts))
 	for _, amount := range amounts {
-		addr := h.NewWalletAddress(w)
+		addr := h.NewWalletAddressOfType(w, addrType)
 
 		pkScript, err := txscript.PayToAddrScript(addr)
 		require.NoError(h, err, "failed to create pkscript")
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -291,6 +291,42 @@ func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
 	return outpoints
 }
 
+// ForeignOutpoint returns an outpoint of the funding transaction that the
+// funded wallet does not own.
+//
+// Funding pays one output per requested amount plus a single change output
+// back to the miner, so exactly one index in the transaction's first
+// len(funded)+1 outputs is not wallet-owned. That output exists on chain and is
+// spendable by the miner, which makes it the fixture's foreign coin.
+func (h *HarnessTest) ForeignOutpoint(funded []wire.OutPoint) wire.OutPoint {
+	h.Helper()
+
+	require.NotEmpty(h, funded, "no funded outpoints")
+
+	owned := make(map[uint32]struct{}, len(funded))
+	for _, outpoint := range funded {
+		owned[outpoint.Index] = struct{}{}
+	}
+
+	for index := range len(funded) + 1 {
+		//nolint:gosec // A funding transaction's output count is small.
+		outputIndex := uint32(index)
+
+		if _, ok := owned[outputIndex]; ok {
+			continue
+		}
+
+		return wire.OutPoint{
+			Hash:  funded[0].Hash,
+			Index: outputIndex,
+		}
+	}
+
+	h.Fatalf("funding transaction has no foreign output")
+
+	return wire.OutPoint{}
+}
+
 // ensureAccount makes sure an account exists for the given key scope, creating
 // it when the backend did not seed it at wallet creation.
 func (h *HarnessTest) ensureAccount(w *wallet.Wallet,
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -92,15 +92,14 @@ type WalletFixture struct {
 }
 
 // NewWallet creates, registers and prepares a wallet as the fixture describes,
-// returning it alongside the outpoints funding produced, in the order of the
-// requested amounts.
+// returning it alongside what its funding transaction created.
 //
 // A funded fixture returns after the funding block's synchronization check,
 // which mining performs for every registered wallet, so its coins are readable
 // without a further check; deriving an address afterwards does not need one
 // either. An unfunded fixture mines nothing and so provides no such boundary.
 func (h *HarnessTest) NewWallet(fixture WalletFixture) (*wallet.Wallet,
-	[]wire.OutPoint) {
+	WalletFunding) {
 
 	h.Helper()
 
@@ -121,7 +120,7 @@ func (h *HarnessTest) NewWallet(fixture WalletFixture) (*wallet.Wallet,
 			h, fixture.Amounts, "funding needs a started wallet",
 		)
 
-		return w, nil
+		return w, WalletFunding{}
 	}
 
 	err = w.Start(h.Context())
@@ -132,7 +131,7 @@ func (h *HarnessTest) NewWallet(fixture WalletFixture) (*wallet.Wallet,
 	}
 
 	if len(fixture.Amounts) == 0 {
-		return w, nil
+		return w, WalletFunding{}
 	}
 
 	return w, h.FundWalletOfType(w, fixture.AddrType, fixture.Amounts...)
@@ -224,11 +223,23 @@ func (h *HarnessTest) NewWalletAddressOfType(w *wallet.Wallet,
 	return addr
 }
 
+// WalletFunding describes the outputs a funding transaction created.
+type WalletFunding struct {
+	// WalletOutpoints are the outputs the funded wallet owns, in the order
+	// of the amounts that were requested.
+	WalletOutpoints []wire.OutPoint
+
+	// ForeignOutpoints are the transaction's remaining outputs, which
+	// exist on chain but belong to the miner rather than to the funded
+	// wallet. Funding pays its own change here, so there is normally one.
+	ForeignOutpoints []wire.OutPoint
+}
+
 // FundWallet pays one output per amount to fresh wallet addresses of the
 // default funding type in a single miner transaction, confirms it, and returns
-// the wallet outpoints in the order of the amounts argument.
+// what the funding transaction created.
 func (h *HarnessTest) FundWallet(w *wallet.Wallet,
-	amounts ...btcutil.Amount) []wire.OutPoint {
+	amounts ...btcutil.Amount) WalletFunding {
 
 	h.Helper()
 
@@ -237,10 +248,13 @@ func (h *HarnessTest) FundWallet(w *wallet.Wallet,
 
 // FundWalletOfType pays one output per amount to fresh wallet addresses of the
 // requested address type in a single miner transaction, confirms it, and
-// returns the wallet outpoints in the order of the amounts argument.
+// returns what the funding transaction created.
+//
+// The transaction's outputs are classified here, where the transaction itself
+// is known, so no caller has to infer which of them the wallet owns.
 func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
 	addrType waddrmgr.AddressType,
-	amounts ...btcutil.Amount) []wire.OutPoint {
+	amounts ...btcutil.Amount) WalletFunding {
 
 	h.Helper()
 
@@ -267,7 +281,11 @@ func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
 	// Locate each wallet output's index within the funding transaction so
 	// callers receive ready-to-use outpoints in the order of the amounts
 	// argument.
-	outpoints := make([]wire.OutPoint, 0, len(outputs))
+	owned := make(map[uint32]struct{}, len(outputs))
+	funding := WalletFunding{
+		WalletOutpoints: make([]wire.OutPoint, 0, len(outputs)),
+	}
+
 	for _, output := range outputs {
 		index := -1
 
@@ -282,49 +300,34 @@ func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
 		require.NotEqual(h, -1, index,
 			"funding output missing from transaction")
 
-		outpoints = append(outpoints, wire.OutPoint{
-			Hash:  *txid,
-			Index: uint32(index), //nolint:gosec
-		})
+		outputIndex := uint32(index) //nolint:gosec
+		owned[outputIndex] = struct{}{}
+
+		funding.WalletOutpoints = append(
+			funding.WalletOutpoints, wire.OutPoint{
+				Hash:  *txid,
+				Index: outputIndex,
+			},
+		)
 	}
 
-	return outpoints
-}
-
-// ForeignOutpoint returns an outpoint of the funding transaction that the
-// funded wallet does not own.
-//
-// Funding pays one output per requested amount plus a single change output
-// back to the miner, so exactly one index in the transaction's first
-// len(funded)+1 outputs is not wallet-owned. That output exists on chain and is
-// spendable by the miner, which makes it the fixture's foreign coin.
-func (h *HarnessTest) ForeignOutpoint(funded []wire.OutPoint) wire.OutPoint {
-	h.Helper()
-
-	require.NotEmpty(h, funded, "no funded outpoints")
-
-	owned := make(map[uint32]struct{}, len(funded))
-	for _, outpoint := range funded {
-		owned[outpoint.Index] = struct{}{}
-	}
-
-	for index := range len(funded) + 1 {
-		//nolint:gosec // A funding transaction's output count is small.
-		outputIndex := uint32(index)
-
+	// Whatever the transaction pays elsewhere belongs to the miner, and is
+	// reported as it stands rather than assumed to exist.
+	for i := range tx.TxOut {
+		outputIndex := uint32(i) //nolint:gosec
 		if _, ok := owned[outputIndex]; ok {
 			continue
 		}
 
-		return wire.OutPoint{
-			Hash:  funded[0].Hash,
-			Index: outputIndex,
-		}
+		funding.ForeignOutpoints = append(
+			funding.ForeignOutpoints, wire.OutPoint{
+				Hash:  *txid,
+				Index: outputIndex,
+			},
+		)
 	}
 
-	h.Fatalf("funding transaction has no foreign output")
-
-	return wire.OutPoint{}
+	return funding
 }
 
 // ensureAccount makes sure an account exists for the given key scope, creating
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -95,10 +95,10 @@ type WalletFixture struct {
 // returning it alongside the outpoints funding produced, in the order of the
 // requested amounts.
 //
-// The returned wallet is not a synchronization boundary. Funding and address
-// derivation are both wallet side effects, so a case that reads wallet state
-// must cross AssertWalletSynced after its own last funding or derivation, not
-// merely after this call.
+// A funded fixture returns after the funding block's synchronization check,
+// which mining performs for every registered wallet, so its coins are readable
+// without a further check; deriving an address afterwards does not need one
+// either. An unfunded fixture mines nothing and so provides no such boundary.
 func (h *HarnessTest) NewWallet(fixture WalletFixture) (*wallet.Wallet,
 	[]wire.OutPoint) {
 
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -67,11 +67,41 @@ func (h *HarnessTest) TestWalletConfig() (wallet.Config,
 	return cfg, params
 }
 
-// CreateEmptyWallet creates, starts, and registers a new wallet instance.
+// WalletFixture describes the wallet a test case needs. It is the harness's
+// parameterized wallet preparation request, so a component test states what it
+// needs instead of growing its own creation, funding and lock policy.
+type WalletFixture struct {
+	// AddrType is the address type funding is paid to. A test that selects
+	// those coins derives its key scope from this same value with
+	// KeyScope(), so the funded scope has a single authority.
+	//
+	// This field is always honored, including its zero value, PubKeyHash.
+	AddrType waddrmgr.AddressType
+
+	// Amounts funds the wallet with one confirmed output per amount, in
+	// the order given. Funding needs a running wallet, so it cannot be
+	// combined with Unstarted.
+	Amounts []btcutil.Amount
+
+	// Unlocked unlocks the wallet once it has started.
+	Unlocked bool
+
+	// Unstarted returns the wallet before Start, for cases asserting
+	// behavior that is only observable while the wallet is not running.
+	Unstarted bool
+}
+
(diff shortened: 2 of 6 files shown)


itest/tx_creator_test.go
```diff
@@ -218,6 +218,18 @@ func assertChangeOutput(h *bwtest.HarnessTest, w *wallet.Wallet,
 	require.Equal(h, addrType, info.AddrType, "change address type mismatch")
 }
 
+// assertInputSourceError asserts the error reports that the wallet could not
+// raise enough input value. The concrete error is unexported, so the typed
+// txauthor interface is the stable public assertion available to a caller.
+func assertInputSourceError(h *bwtest.HarnessTest, err error, msg string) {
```
yyforyongyu: Don't do this in itest - if we think an assertion is useful, build in on the assertion harness, otherwise use pure `require.XXX` to unwrap the fields we are checking in the test, this way we can easily read what's been asserted here without jumping around.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,929 @@
...
+var relayFeeRate = btcunit.NewSatPerKVByte(txrules.DefaultRelayFeePerKb)
+
+// fundedScope is the key scope backing the harness funding address type. The
+// wallet resolves accounts per scope, so a case that funds this scope must
+// select from it too.
+var fundedScope = waddrmgr.KeyScopeBIP0084
+
+// setupTxWallet creates a started, unlocked and synced wallet funded with one
+// output per amount under the default funding address type.
+func setupTxWallet(h *bwtest.HarnessTest,
```
yyforyongyu: why do we need a customized wallet creation method for every component we touch? we should instead rely on the `harness`, that is what the harness is for, such that we only have one or two wallet creation method and use it everywhere in the code - if there are customization needed, the wallet creation method should take a params. E.g., we have this in `lnd`: https://github.com/lightningnetwork/lnd/blob/8464116a051a06a23ecdba71f35509bb14db9183/lntest/node/harness_node.go#L95 which can be complicated given the intrinsic complexity in lnd, but we should not add a new way to create test wallets whenever we start testing a new component.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,929 @@
...
+	for _, outpoint := range outpoints {
+		h.AssertUtxoConfirmations(w, outpoint, 1)
+	}
+
+	return w, outpoints
+}
+
+// scopedDefault returns the funded scope's default account as a coin source
+// and change source.
+func scopedDefault(scope waddrmgr.KeyScope) *wallet.ScopedAccount {
```
yyforyongyu: why do we need this thin method?

itest/tx_creator_test.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	authored, err := w.CreateTransaction(h.Context(), intent)
+	require.NoError(h, err, "failed to create transaction")
+
+	assertAuthoredTx(h, authored, intent.Outputs)
+
+	// Largest-first selection covers the payment with the two BTC coin on
+	// its own.
+	assertTxInputs(h, authored.Tx, outpoints[1])
```
yyforyongyu: This is also suffering from the same issue `Keep the tests linear` mentioned here https://github.com/btcsuite/btcwallet/pull/1312#pullrequestreview-4903720360
itests need to be linear so we can easily validate what's asserted here.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,344 @@
...
+// the requested outputs survive unchanged, the reported input metadata
+// describes the transaction's own inputs, and the inputs cover the outputs
+// with a fee left over.
+func assertAuthoredTx(h *bwtest.HarnessTest, authored *txauthor.AuthoredTx,
+	outputs []wire.TxOut) {
+
+	h.Helper()
+
+	require.NotNil(h, authored, "nil authored tx")
+	require.NotNil(h, authored.Tx, "nil unsigned tx")
```
yyforyongyu: we should stop doing these kinda `nil` checks in the test - this is a test env, if it panics we can just fix it.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+
+	var spent btcutil.Amount
+	for _, output := range tx.TxOut {
+		spent += btcutil.Amount(output.Value)
+	}
+
+	require.Greater(
+		h, authored.TotalInput, spent, "authored tx pays no fee",
+	)
```
yyforyongyu, line 215: This helper is checking too much: it combines several distinct contracts: transaction shape, input metadata, output preservation, change accounting, and fee behavior. This makes individual tests appear more focused than they really are and obscures which assertions are essential to each named test.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,344 @@
...
+}
+
+// containsOutput reports whether the outputs contain one with the same value
+// and script as want.
+func containsOutput(outputs []*wire.TxOut, want *wire.TxOut) bool {
```
yyforyongyu: When naming variables, we need to be mindful that this method is exposed to all tests under `itest`, so we would prefer verbose and explicit names.

yyforyongyu: Also if this method is used only once, it's premature abstraction to have it now, unless we have upcoming tests to use it?

yyforyongyu (changes requested):

yyforyongyu:
## Detailed review summary

I reviewed the commits from `6a12d9b9bfd13263877d2eaba44aae558c323248`
through the current head, `d996ccec16d3b5cfa41029133cf7b0c9c04a653f`.
As requested, this assessment ignores the expected rebase conflicts.

The intended TxCreator coverage is valuable, but the current test structure
introduces a real backend failure and makes the contracts difficult to audit.
The main direction should be: keep wallet lifecycle and synchronization in
`bwtest`, keep each itest linear, and keep each test's subject API and relevant
assertions visible together.

### 1. The TxCreator tests race wallet recovery

`setupTxWalletOfType` in `itest/tx_creator_test.go:48-71` waits for two
observable conditions:

1. `AssertWalletSynced`, which compares the wallet and chain heights.
2. `AssertUtxoConfirmations`, which observes each outpoint through `GetUtxo`.

Those conditions can both pass while the Store-backed syncer is still
replaying recovery work or moving between synced and syncing states. In
addition, `paymentOutput` derives another wallet address after setup, so a
setup-only synchronization check is not necessarily the final synchronization
boundary before `CreateTransaction`.

This is reproducible on the current checkout:

- `testCreateTransactionSelectCoins` fails on btcd/kvdb at the
  `CreateTransaction` call with `insufficient funds available to construct
  transaction`.
- `testCreateTransactionManualInputs` can fail at its second
  `CreateTransaction` call with `operation forbidden in current state: wallet
  is currently syncing`.
- The same focused nine-case suite passed locally on SQLite and PostgreSQL,
  demonstrating that the tests are not currently enforcing one
  backend-neutral public contract.

The harness should own an authoritative readiness boundary that waits for a
stable synced, non-recovery state and for the Store query used by selection to
observe the funded inputs. That boundary must be crossed after the last
address/funding side effect that can trigger recovery, not merely after the
initial funding transaction.

### 2. Wallet setup belongs to the harness

`setupTxWallet` and `setupTxWalletOfType` create another component-specific way
to construct and prepare a wallet. If every component introduces its own
creation method, wallet lifecycle, funding, locking, synchronization, and
cleanup policy will gradually diverge.

`bwtest.HarnessTest` should remain the single owner of those side effects. If
tests need different funding address types, lock states, or funding amounts,
the existing harness entry point should accept explicit parameters or a small
configuration value. TxCreator tests should consume that common facility
rather than define a TxCreator-specific wallet constructor.

This is both a SOLID ownership issue and an SSOT issue: the harness should have
one reason to change when wallet setup semantics change, instead of every
component test carrying its own partial setup policy.

### 3. Keep the itests linear and keep assertions visible

The component-local assertion helpers make the individual test names look
focused while hiding most of their actual contracts:

- `assertAuthoredTx` checks transaction non-nilness, input existence, metadata
  lengths, positional input values and scripts, `TotalInput`, output count,
  change-index bounds, requested-output preservation, and the existence of a
  positive fee. These are several independent contracts, not one assertion.
- `assertTxInputs` hides the central assertion in the coin-selection and
  manual-input tests.
- `assertInputSourceError` hides the stable `ErrorAs` contract that rejection
  tests are meant to document.
- `assertChangeOutput`, `assertWalletUtxos`, and `assertLeasedOutputs` similarly
  need to be evaluated under the same rule: if they are authoritative,
  reusable integration-harness assertions, put them on `bwtest.HarnessTest`;
  otherwise keep the relevant `require` calls in the test.

For example, `testCreateTransactionSelectCoins` should visibly show:

1. the funded coins and intended policy;
2. the `CreateTransaction` call;
3. the exact selected input;
4. the requested output and change contract;
5. the wallet-visible UTXO and lease state.

A reviewer should not have to jump through several helpers to discover which
of those behaviors the case actually asserts. Apply the same linear
Arrange/Act/Assert shape to every TxCreator case, and assert only the contract
named by that case plus the minimum invariants needed to make that assertion
meaningful.

The defensive `require.NotNil` checks inside `assertAuthoredTx` are also
unnecessary here. After `require.NoError`, a nil API result is a test defect;
allowing its use to panic makes the broken contract obvious rather than adding
generic checks to every success path.

### 4. Remove thin and premature abstractions

`scopedDefault` only constructs a two-field `wallet.ScopedAccount`. It does not
isolate a side effect, define a new domain operation, or materially reduce
complexity, and it hides the account and scope at call sites where those values
matter. Construct the value explicitly.

`containsOutput` is used only through `assertAuthoredTx`. If
`assertAuthoredTx` is removed, the helper has no reason to remain. If a shared
package-level helper is genuinely needed later, its names should be explicit
because every test file in package `itest` can use it; names such as `outputs`
and `want` are too context-dependent for package-level test vocabulary.

The same callable-design rule applies throughout this file: add a helper only
when it removes real duplication, owns a side effect, materially reduces
complexity, or represents a genuine test-domain operation.

### 5. Funding address type and scope have two authorities

`bwtest/harness_wallet.go:110-112` defines `fundingAddrType` as the canonical
address type used by `FundWallet`, while `itest/tx_creator_test.go:33-36`
independently hardcodes `fundedScope = waddrmgr.KeyScopeBIP0084` and
`setupTxWallet` independently hardcodes `waddrmgr.WitnessPubKey`.

The divergence is concrete in `testCreateTransactionWalletState`: it funds via
`h.FundWallet`, whose type is owned by the harness, but selects and derives
change via the separately maintained `fundedScope`. Changing the harness
default could silently make the test fund one scope and select another.

Use one explicit funding address type and derive its key scope with
`KeyScope()` at the use site, or expose the harness's configured funding type
as part of its result. Do not maintain the type and scope as separate facts in
different packages.

### 6. Foreign-output derivation is duplicated

`foreignOutpoint` in `itest/tx_creator_test.go:719-744` builds an owned-index
map and scans the funding transaction's output range for the miner's change
output. `testGetUtxo` in `itest/utxo_manager_test.go:411-436` already contains
the same algorithm inline.

Make one package-level operation the canonical owner and use it from both
tests, or move the operation into the funding harness if identifying the
non-wallet output is a supported property of the fixture. Keeping two copies
allows the assumptions about the funding transaction to diverge.

Relatedly, `tx_creator_test.go` currently consumes generic fixture constants
and helpers such as `oneBTC`, `twoBTC`, `threeBTC`, `maxConfsLimit`,
`leaseDuration`, and `unknownOutpoint` from `utxo_manager_test.go`. Truly shared
fixtures should have neutral ownership; otherwise each component test should
state its relevant values explicitly instead of depending on another
component's test file.

### Test-function impact

| Test function | Main issues |
| --- | --- |
| `testCreateTransactionSelectCoins` | Recovery race; primary selection and transaction assertions hidden behind helpers; duplicated funding scope. |
| `testCreateTransactionMultipleOutputs` | Recovery race; the named multiple-output contract is buried inside `assertAuthoredTx`; duplicated funding scope. |
| `testCreateTransactionManualInputs` | Recovery race, including observed `ErrStateForbidden`; both manual-selection contracts hidden behind helpers; duplicated funding scope. |
| `testCreateTransactionDefaultAccount` | Recovery race after funding/address derivation; input, output, and change contracts hidden behind helpers. |
| `testCreateTransactionCoinSource` | Recovery race; candidate selection hidden by `assertTxInputs`; duplicated funding scope. |
| `testCreateTransactionOmitChange` | Recovery race; `assertAuthoredTx` overlaps the case's direct no-change assertions; duplicated funding scope. |
| `testCreateTransactionRejectIntent` | Recovery can affect its successful maximum-fee boundary; stable error and success contracts are hidden by helpers. |
| `testCreateTransactionRejectInputs` | Recovery race; duplicated `foreignOutpoint`; duplicated funding scope. |
| `testCreateTransactionWalletState` | Recovery race; concrete split authority between harness funding type and `fundedScope`; success behavior hidden by helpers. |
| `testGetUtxo` | Contains the original foreign-output derivation duplicated by the new TxCreator helper. |

### Expected rewrite direction

1. Add or extend one harness-owned, parameterized wallet preparation path.
2. Give the harness a durable wallet-ready assertion that covers sync state,
   recovery state, and Store-visible funded inputs.
3. Use one explicit funding address type and derive the corresponding scope
   rather than mirroring it in `itest`.
4. Rewrite each TxCreator test linearly with its relevant `require` assertions
   next to `CreateTransaction`.
5. Remove thin/component-local helpers that do not own side effects or real
   reusable contracts.
6. Consolidate the foreign-output fixture logic and give shared fixtures
   neutral ownership.
7. Re-run the focused TxCreator suite on every claimed wallet backend,
   especially kvdb, and add the missing commit bodies while retaining valid
   signatures.

Abdulkbk:
Converting to draft, will request for reviews once it's back in shape.

itest/tx_creator_test.go
Abdulkbk: now using require.XXX inline

itest/tx_creator_test.go
Abdulkbk: Removed, added a wallet fixture.

itest/tx_creator_test.go
Abdulkbk: We don't, removed, doing it inline.

itest/tx_creator_test.go
Abdulkbk: removed

itest/tx_creator_test.go
Abdulkbk: fixed

bwtest/README.md
```diff
@@ -34,28 +34,36 @@ The `bitcoind` backend uses ZMQ for block/tx notifications.
...
-exercise the wallet manager:
-
-- `(*HarnessTest).CreateEmptyWallet`
-- `(*HarnessTest).CreateFundedWallet`
-- `(*HarnessTest).FundWallet`
-
-Example usage:
+`bwtest` owns wallet creation, funding and lock policy so component tests do
+not each define their own. `(*HarnessTest).NewWallet` is the single
+parameterized entry point; it takes a `WalletFixture` describing what the case
```
yyforyongyu, line 39: 👍

bwtest/README.md
```diff
@@ -34,28 +34,36 @@ The `bitcoind` backend uses ZMQ for block/tx notifications.
...
-}
-```
+Two convenience wrappers remain for cases that need nothing else:
+
+- `(*HarnessTest).CreateEmptyWallet`
+- `(*HarnessTest).CreateFundedWallet`
+
+Funding is also available on its own through `(*HarnessTest).FundWallet` and
+`(*HarnessTest).FundWalletOfType`, and addresses through
+`(*HarnessTest).NewWalletAddress` and `(*HarnessTest).NewWalletAddressOfType`.
```
yyforyongyu, line 66: It just occurred to me that we are missing an `AGENTS.md` here, will create it in another PR.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,897 @@
...
+	needsBothCoins := deriveWalletPayment(
+		h, w, txCreatorFundingType, twoBTC+halfBTC,
+	)
+
+	// This one exceeds everything the wallet holds.
+	beyondBalance := deriveWalletPayment(
+		h, w, txCreatorFundingType, oneBTC+twoBTC+halfBTC,
+	)
+
+	h.AssertWalletSynced(w)
```
yyforyongyu: So there is an occuring pattern of `deriveWalletPayment` + `h.AssertWalletSynced`, I wonder if we should put `AssertWalletSynced` inside `deriveWalletPayment` instead, or make `deriveWalletPayment` a method on `h`?

yyforyongyu: Actually, since #1314 is merged, I don't think we need this `AssertWalletSynced` check anymore - the wallet won't be returned unless it's already synced from `NewWallet`, in which the mining part will assert the wallet being synced.

yyforyongyu (changes requested):
Wow very nice! Left a few comments, also gonna post agentic reviews soon!

itest/tx_creator_test.go
```diff
@@ -0,0 +1,897 @@
...
+		Outputs: []wire.TxOut{payment},
+		Inputs: &wallet.InputsPolicy{
+			Strategy: wallet.CoinSelectionLargest,
+			MinConfs: 1,
+			Source:   account,
+		},
+		ChangeSource: account,
+		FeeRate:      relayFeeRate,
+	})
+	require.NoError(h, err, "failed to create transaction")
```
yyforyongyu, line 83: Under concurrent harness load, this case can reach this line after `AssertWalletSynced` has observed the exact funding block reported by the miner and still fail with `insufficient funds available to construct transaction` (3 of 4 concurrent btcd/SQLite runs; 20 isolated runs passed). `ApplyTxBatch` commits the matched transactions/UTXOs and `SyncedTo` atomically, so adding another outpoint poll here would only hide a scan-completeness violation. Please root-cause why the scan can advance the tip without materializing this wallet-owned output (for example, the startup watch-list/recovery-state timing) and preserve `AssertWalletSynced` as the readiness contract.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,897 @@
...
+	for _, txIn := range authored.Tx.TxIn {
+		spent = append(spent, txIn.PreviousOutPoint)
+	}
+
+	require.ElementsMatch(
+		h, []wire.OutPoint{outpoints[0], outpoints[2]}, spent,
+		"unexpected manual inputs",
+	)
+	require.Equal(
+		h, btcutil.Amount(oneBTC+threeBTC), authored.TotalInput,
```
yyforyongyu, line 264: `AuthoredTx` is expected to return signing metadata positionally aligned with `Tx.TxIn`, but this multi-input case only checks outpoints and `TotalInput`. Please assert each `PrevScripts[i]` and `PrevInputValues[i]` against `GetUtxo(Tx.TxIn[i].PreviousOutPoint)`. The existing single-input value check does not cover script metadata or multi-input ordering.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,897 @@
...
+		expectedErr: wallet.ErrNilTxIntent,
+	}, {
+		name: "no outputs",
+		intent: &wallet.TxIntent{
+			Inputs:  &wallet.InputsPolicy{Source: account},
+			FeeRate: relayFeeRate,
+		},
+		expectedErr: wallet.ErrNoTxOutputs,
+	}, {
+		name: "dust output",
```
yyforyongyu: This checks an arbitrary dust value, not the policy edges. Please cover one satoshi below the script-specific dust threshold (rejected) and the smallest non-dust amount (accepted), plus `btcutil.MaxSatoshi + 1` yielding `txrules.ErrAmountExceedsMax` and ideally `MaxSatoshi` being accepted. This is needed to catch off-by-one regressions in `txrules.CheckOutput`.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,897 @@
...
+		},
+		expectedErr: wallet.ErrAccountNotFound,
+	}}
+
+	for _, tc := range testCases {
+		_, err := w.CreateTransaction(h.Context(), tc.intent)
+		require.ErrorIs(
+			h, err, tc.expectedErr, "%s not rejected", tc.name,
+		)
+	}
```
yyforyongyu, line 652: The no-state-change check occurs only after the successful maximum-fee authoring below and only inspects UTXOs and leases. A rejected intent could derive and persist a change address and still pass. Please snapshot addresses, UTXOs, and leases immediately before this rejection loop and compare them immediately after it, before the successful authoring call.

bwtest/README.md
```diff
@@ -34,28 +34,36 @@ The `bitcoind` backend uses ZMQ for block/tx notifications.
...
-`bwtest` includes convenience helpers for tests that do not want to directly
-exercise the wallet manager:
-
-- `(*HarnessTest).CreateEmptyWallet`
-- `(*HarnessTest).CreateFundedWallet`
-- `(*HarnessTest).FundWallet`
-
-Example usage:
+`bwtest` owns wallet creation, funding and lock policy so component tests do
+not each define their own. `(*HarnessTest).NewWallet` is the single
```
yyforyongyu, line 38: This now declares `NewWallet` the single parameterized fixture entry point, but `itest/utxo_manager_test.go` still has a separate `createWallet` implementation of the same create/register/unstarted policy and seven UTXO-manager cases still call it. Please migrate those callers to `h.NewWallet(WalletFixture{Unstarted: true})` and remove `createWallet` so this policy has one owner.

yyforyongyu (changes requested):

itest/tx_creator_test.go
yyforyongyu: ok it's hard to read, but what it meant was,


The expected sequence is:
1. The test derives wallet-owned funding addresses.
2. A transaction pays those addresses.
3. The miner confirms that transaction in a block.
4. The wallet scans the block.
5. The wallet stores the transaction’s outputs as spendable UTXOs.
6. The wallet updates SyncedTo to that block.
7. AssertWalletSynced returns.
8. CreateTransaction selects those UTXOs.

In the failing runs, this appears to happen instead:
1. The wallet scans or otherwise advances through the funding block.
2. It updates SyncedTo to that block.
3. AssertWalletSynced succeeds.
4. The funding outputs are nevertheless absent from ListUTXOs.
5. CreateTransaction returns “insufficient funds.

yyforyongyu: So this appears to be a deeper issue that #1314 claimed to fix - I think we should create a followup task instead rather than expanding our scope here.

bwtest/README.md
Abdulkbk: fixed in https://github.com/btcsuite/btcwallet/pull/1316/changes/c4e5833c3c6bc1c50b1353d21725b74cdc735b88

itest/tx_creator_test.go
Abdulkbk: fixed in 68d5b38e2642a7c52a4ac51a2b98068d50c0d833

itest/tx_creator_test.go
Abdulkbk: added a test covering the edge 2183ecc5dd5dd9274c73ae0e2aaedb76ed6fb195

itest/tx_creator_test.go
Abdulkbk: fixed in 1e02cab0

itest/tx_creator_test.go
Abdulkbk: Removed`AssertWalletSynced` in 0178fe6bbc7e0164b6fd9c64bb543768c128d485

bwtest/harness_wallet.go
```diff
@@ -67,11 +67,41 @@ func (h *HarnessTest) TestWalletConfig() (wallet.Config,
...
+}
+
+// NewWallet creates, registers and prepares a wallet as the fixture describes,
+// returning it alongside the outpoints funding produced, in the order of the
+// requested amounts.
 //
-// This is intended for non-manager integration tests that want a ready-to-use
-// wallet without repeating boilerplate.
-func (h *HarnessTest) CreateEmptyWallet() *wallet.Wallet {
+// The returned wallet is not a synchronization boundary. Funding and address
```
yyforyongyu: This doc is no longer accurate - maybe mention that once the funding tx is mined, the mining process will assert that the wallet is synced. We can say that funded fixtures return after the funding block synchronization check, while unfunded fixtures provide no such boundary; address derivation alone does not require another sync check.

itest/tx_creator_test.go
```diff
@@ -0,0 +1,1039 @@
...
+	)
+	require.Equal(
+		h, payment.Value, authored.Tx.TxOut[0].Value,
+		"payment value changed",
+	)
+}
+
+// testCreateTransactionRejectIntent verifies the stable errors a malformed or
+// unresolvable intent produces, that the fee rate bound admits its own limit,
+// and that a rejection leaves the wallet's coins untouched.
```
yyforyongyu: So if we wanna go with option 1, we need to update here, basically add that `addresses being untouched` is also asserted here,
```diff
-// testCreateTransactionRejectInputs verifies that inputs the wallet cannot                                          
-// spend are refused, that selection skips a leased coin rather than spending                                        
-// it, and that a refusal leaves the wallet's coins and leases untouched.                                            
+// testCreateTransactionRejectInputs verifies that inputs the wallet cannot                                          
+// spend are refused, that selection skips a leased coin rather than spending                                        
+// it, and that a refusal leaves the wallet's addresses, coins and leases                                            
+// untouched. 
```

itest/tx_creator_test.go
```diff
@@ -0,0 +1,1039 @@
...
+			FeeRate:      relayFeeRate,
+		},
+		expectedErr: wallet.ErrAccountNotFound,
+	}}
+
+	// A refused intent must change nothing the wallet exposes, including
+	// its derived addresses: a rejection that still allocated a change
+	// address would leak one on every failed attempt. Snapshot before the
+	// loop and compare straight after it, before anything succeeds.
+	addressesBefore, err := w.ListAddresses(
```
yyforyongyu: There are two coherent choices.

  ### Option 1: Keep and complete the address-state contract

  Retain the assertion that rejected transactions must not derive or persist change addresses.

  To make it correct:

  - Give every rejection case an explicit ChangeSource in the monitored BIP84 account—especially empty manual inputs,   
    duplicate manual inputs, and empty coin source.

  - Keep the BIP84 before/after snapshot in testCreateTransactionRejectIntent.
  - Add the same snapshot to testCreateTransactionRejectInputs, because those failures reach deeper into transaction    
    preparation.

  This establishes a broader contract:

  > Any failed CreateTransaction call leaves address derivation state unchanged.

  Tradeoff: this expands Task 434 into address-index side-effect behavior and constrains future implementations.        

  ### Option 2: Remove the address-state contract

  Delete the addressesBefore and addressesAfter blocks from testCreateTransactionRejectIntent, and do not add them to   
  testCreateTransactionRejectInputs.

  The tests would continue covering:

  - Stable intent-validation errors.
  - Invalid and unknown change sources.
  - Successful change-address generation in testCreateTransactionSelectCoins.
  - Implicit Taproot change in testCreateTransactionDefaultAccount.
  - Preservation of coins and leases after rejection.

  Tradeoff: the suite would not detect a rejected transaction consuming a change-address index, but that behavior is not required by Task 434.

yyforyongyu: I'd favor option 2 to not increase the scope, but it's up to you - as long as the stated contract in the test doc matches what's being asserted, we should be good here.

yyforyongyu (changes requested):
Cool one last thing about the godoc, plus an option we need to take! Otherwise it's good to go!

itest/tx_creator_test.go
Abdulkbk: Went with option 2 so as not to expand the scope too much.

itest/tx_creator_test.go
Abdulkbk: fixed

bwtest/harness_wallet.go
Abdulkbk: fixed

yyforyongyu (approved):
LGTM!

itest/tx_creator_test.go
```diff
@@ -0,0 +1,1024 @@
...
+	// txCreatorFundingType is the address type the TxCreator cases fund and
+	// select from. Each case derives its key scope from this value, so the
+	// funded scope is never stated twice.
+	txCreatorFundingType = waddrmgr.WitnessPubKey
+)
+
+// relayFeeRate is the fee rate the TxCreator cases author at. It is the
+// default relay fee, one satoshi per virtual byte, which keeps the fee
+// negligible against the funded amounts.
+var relayFeeRate = btcunit.NewSatPerKVByte(txrules.DefaultRelayFeePerKb)
```
GustavoStingelin: nit: as it is a package-visible helper, I guess it should live in the fixtures file?

bwtest/harness_wallet.go
```diff
@@ -204,6 +291,42 @@ func (h *HarnessTest) FundWallet(w *wallet.Wallet,
...
+		return wire.OutPoint{
+			Hash:  funded[0].Hash,
+			Index: outputIndex,
+		}
+	}
+
+	h.Fatalf("funding transaction has no foreign output")
+
+	return wire.OutPoint{}
+}
```
GustavoStingelin: Could we return the foreign outputs as part of the funding result instead of inferring one here?

This helper only receives the wallet-owned outpoints, so it cannot verify the funding transaction's actual outputs. In particular, with `N` funded outpoints, `owned` contains at most `N` indices, while the loop examines `N+1` candidate indices (`0` through `N`). It is therefore guaranteed to find a missing index, making the final `Fatalf` effectively unreachable.

For example, if the transaction contained only the `N` wallet outputs and no miner change, the helper would return index `N`. That index does not exist in the transaction, but tests expecting `ErrUnknownOutput` or `ErrUtxoNotEligible` could still pass. The intended "existing on-chain output not owned by this wallet" case would silently become another nonexistent-output case.

Since `FundWalletOfType` has the complete funding transaction, could it classify the actual outputs and return something like:

```go
type WalletFunding struct {
	WalletOutpoints  []wire.OutPoint
	ForeignOutpoints []wire.OutPoint
}
```
Callers that specifically require one miner change output can then assert:
```go
require.Len(h, funding.ForeignOutpoints, 1)
foreign := funding.ForeignOutpoints[0]
```
This keeps funding-transaction knowledge in the harness while avoiding assumptions about the transaction's output count and layout.

GustavoStingelin (changes requested):
I left just two comments

itest/tx_creator_test.go
Abdulkbk: fixed in https://github.com/btcsuite/btcwallet/pull/1316/changes/72287bc6a2b448a9f00fee4bdfcd7a60a7f004ac

bwtest/harness_wallet.go
Abdulkbk: Added https://github.com/btcsuite/btcwallet/pull/1316/changes/6f9f10263fb1ca60bfb84b4e3aac44bfbfd3d332 to fix this

GustavoStingelin (approved):
Thanks, LGTM!

GustavoStingelin:
@yyforyongyu pls re-ack

PR #1318, opened 2026-08-11
wallet: filter recovery blocks outside database transaction

### LINKED ISSUE #1317: Recovery holds the wallet database writer during FilterBlocks I/O
## Description

`Wallet.recovery` currently invokes `chain.Interface.FilterBlocks` while a
`walletdb.Update` transaction is active. A chain backend is allowed to perform
remote I/O in `FilterBlocks` (neutrino and Esplora-backed implementations do),
so a slow or stalled chain request holds the wallet's single database writer
for its full duration. Unrelated key derivation and persistence operations then
block behind recovery and cannot observe their caller's context cancellation.

This was reproduced on a physical iPhone with Wavelength's Esplora-backed
btcwallet integration. An invoice request timed out while deriving its receive
authentication key.

The blocked goroutine was:

```text
bbolt.(*DB).beginRWTx
walletdb/bdb.(*db).Update
BtcWalletKeyRing.DerivePrivKey
BtcWalletKeyRing.ECDH
```

The writer-owning goroutine was:

```text
net/http.(*http2ClientConn).roundTrip
EsploraChainService.FilterBlocks
btcwallet.(*Wallet).recoverScopedAddresses
btcwallet.(*Wallet).recovery.func2
walletdb/bdb.(*db).Update
```

The same build usually succeeds in the iOS simulator because the remote scan
finishes quickly enough there to hide the transaction boundary.

## Root cause

`Wallet.recovery` wraps both `recoverScopedAddresses` and the synced-to update
in one write transaction. `recoverScopedAddresses` expands address horizons,
calls `FilterBlocks`, applies results, and can repeat that process several
times. The remote filtering step therefore runs while the exclusive writer is
held.

## Expected behavior

Recovery should not hold a wallet database transaction during chain backend
I/O. It should retain the existing atomic commit of recovered addresses,
transactions, and synced-to state.

## Proposed fix

Split each recovery batch into two phases:

1. Derive recovery horizons under short read transactions and run all
   `FilterBlocks` requests without a database transaction, accumulating the
   responses and advancing only in-memory recovery state.
2. In one write transaction, apply the accumulated address/transaction
   discoveries and advance the synced-to marker.

A regression test can have `FilterBlocks` attempt a concurrent
`walletdb.Update`; it must acquire the writer before returning.

Downstream report with the full device evidence:
https://github.com/lightninglabs/wavelength/issues/1134

bhandras:
## The problem

During seed recovery, the wallet scans old blocks in batches. It used to keep a wallet database write transaction open while the chain backend ran `FilterBlocks`.

`FilterBlocks` can wait on the network. If it is slow, the open transaction blocks every other wallet database writer. On a physical iPhone this blocked invoice authentication key work long enough for invoice creation to time out.

## The fix

This PR splits each recovery batch into three steps:

1. Read the wallet's address state in a short read transaction and build a complete filter request.
2. Close that transaction, then ask the chain backend to scan the blocks.
3. After every backend call for the batch succeeds, save all recovered addresses, transactions, and sync markers in one write transaction.

Operations that can advance address indexes, including address creation and transaction funding, are held for the duration of a recovery batch. This keeps the address range in the request stable while the backend scans it. Other database writers, including key operations unrelated to address indexes, remain available.

## Safety rule

Recovery still has one all-or-nothing database update per batch:

- If a backend call fails, the wallet does not save any discovery from that batch and does not move its sync marker forward.
- If the final database update fails, walletdb rolls back the recovered addresses, transactions, and sync marker together.
- If the process stops before that update, the stored sync marker is unchanged, so recovery scans the batch again after restart.

Filter responses remain in memory until the batch commits, so peak recovery memory grows with matching blocks and transactions within one batch. This bounded trade-off preserves the single atomic commit boundary.

The old database transaction never locked the chain backend or froze its view of the chain. Moving backend I/O outside that transaction therefore does not weaken reorg handling. Requests still name the exact blocks to scan, and the existing rollback and rescan paths are unchanged.

## What does not change

- The recovery window, block batches, address derivation, and watched outpoints are unchanged.
- The `chain.Interface` API and its request and response types are unchanged.
- The wallet database schema is unchanged.
- A failed scan is still an error; it is never treated as an empty result.
- Address-index operations could already wait behind the old wallet-wide database writer. They now wait on a narrow address-index lock, while unrelated writers can continue.

## Backend safety

This change is not specific to lwwallet or Esplora.

The built-in btcd, bitcoind, and Neutrino backends all receive a self-contained `FilterBlocksRequest` containing the block list, address maps, and watched outpoints. They return a self-contained response and do not receive or use a wallet database transaction. The implementations differ only in how they query the chain: Neutrino and btcd use compact-filter-based paths, while bitcoind fetches blocks through Bitcoin Core. Bitcoin Core notification mode—RPC polling or ZMQ—does not change this recovery call.

An external `chain.Interface` implementation, such as an Esplora-backed wallet, has the same contract. The PR changes when btcwallet holds its own locks; it does not change what any backend must do.

## Tests

New backend-neutral tests verify that:

- an unrelated wallet database writer can complete while `FilterBlocks` is running;
- real address creation is serialized until the batch is committed, then completes normally;
- the locking tests cannot pass without invoking `FilterBlocks`;
- discoveries returned by multiple backend calls are committed with the final sync marker;
- an error from a later backend call leaves earlier discoveries and the sync marker uncommitted.

Local checks:

- `golangci-lint run ./wallet`
- `go test ./wallet -run 'TestRecovery(FilterBlocks|Batch)' -count=10`
- focused recovery tests with the race detector, repeated five times
- the complete wallet unit suite

The original failure was also retested on an Esplora-backed Wavelength build on a physical iPhone. Two consecutive invoice requests completed in under one second instead of timing out at 20 seconds.

## Compatibility and rollout

There are no API, database, protocol, configuration, or backend changes. Existing wallets and all current chain backends use the same recovery data and commit boundary. The visible behavior change is that unrelated wallet writes no longer wait for recovery network I/O.

Fixes #1317.

Related downstream report: https://github.com/lightninglabs/wavelength/issues/1134

### DIFF
--- a/wallet/recovery_internal_test.go
+++ b/wallet/recovery_internal_test.go
@@ -1,0 +1,474 @@
+package wallet
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/btcsuite/btcd/address/v2"
+	"github.com/btcsuite/btcd/chainhash/v2"
+	"github.com/btcsuite/btcd/txscript/v2"
+	"github.com/btcsuite/btcd/wire/v2"
+	"github.com/btcsuite/btcwallet/chain"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/walletdb"
+	"github.com/btcsuite/btcwallet/wtxmgr"
+	"github.com/stretchr/testify/require"
+)
+
+// newRecoveryTestState creates the recovery state and scoped-manager map used
+// by a production recovery run, including state already persisted in walletdb.
+func newRecoveryTestState(t *testing.T, w *Wallet) (*RecoveryManager,
+	map[waddrmgr.KeyScope]*waddrmgr.ScopedKeyManager) {
+
+	t.Helper()
+
+	recoveryMgr := NewRecoveryManager(
+		w.recoveryWindow, recoveryBatchSize, w.chainParams,
+	)
+
+	scopedMgrs := make(
+		map[waddrmgr.KeyScope]*waddrmgr.ScopedKeyManager,
+	)
+	for _, scopedMgr := range w.Manager.ActiveScopedKeyManagers() {
+		scopedMgrs[scopedMgr.Scope()] = scopedMgr
+	}
+
+	err := walletdb.View(w.db, func(tx walletdb.ReadTx) error {
+		txMgrNS := tx.ReadBucket(wtxmgrNamespaceKey)
+
+		credits, err := w.TxStore.UnspentOutputs(txMgrNS)
+		if err != nil {
+			return err
+		}
+
+		addrMgrNS := tx.ReadBucket(waddrmgrNamespaceKey)
+
+		return recoveryMgr.Resurrect(addrMgrNS, scopedMgrs, credits)
+	})
+	require.NoError(t, err)
+
+	return recoveryMgr, scopedMgrs
+}
+
+// newRecoveryTestBlock returns matching filter and sync records for a block.
+func newRecoveryTestBlock(height int32) (wtxmgr.BlockMeta,
+	*waddrmgr.BlockStamp) {
+
+	var hash chainhash.Hash
+
+	hash[0] = byte(height)
+
+	timestamp := time.Unix(int64(height), 0)
+	blockMeta := wtxmgr.BlockMeta{
+		Block: wtxmgr.Block{
+			Hash:   hash,
+			Height: height,
+		},
+		Time: timestamp,
+	}
+	blockStamp := &waddrmgr.BlockStamp{
+		Hash:      hash,
+		Height:    height,
+		Timestamp: timestamp,
+	}
+
+	return blockMeta, blockStamp
+}
+
+// newRecoveryTestResponse builds a filter response for the highest external
+// address in a scope and a transaction that pays to that address.
+//
+//nolint:ireturn
+func newRecoveryTestResponse(req *chain.FilterBlocksRequest,
+	scope waddrmgr.KeyScope, value int64) (*chain.FilterBlocksResponse,
+	address.Address, uint32, *wire.MsgTx, error) {
+
+	var (
+		recoveredAddr  address.Address
+		recoveredIndex uint32
+	)
+	for scopedIndex, candidate := range req.ExternalAddrs {
+		if scopedIndex.Scope != scope {
+			continue
+		}
+
+		if recoveredAddr != nil && scopedIndex.Index < recoveredIndex {
+			continue
+		}
+
+		recoveredAddr = candidate
+		recoveredIndex = scopedIndex.Index
+	}
+
+	if recoveredAddr == nil {
+		return nil, nil, 0, nil, errors.New("missing recovery address")
+	}
+
+	pkScript, err := txscript.PayToAddrScript(recoveredAddr)
+	if err != nil {
+		return nil, nil, 0, nil, err
+	}
+
+	relevantTx := &wire.MsgTx{
+		TxIn:  []*wire.TxIn{{}},
+		TxOut: []*wire.TxOut{wire.NewTxOut(value, pkScript)},
+	}
+	response := &chain.FilterBlocksResponse{
+		BatchIndex: 0,
+		BlockMeta:  req.Blocks[0],
+		FoundExternalAddrs: map[waddrmgr.KeyScope]map[uint32]struct{}{
+			scope: {recoveredIndex: {}},
+		},
+		RelevantTxns: []*wire.MsgTx{relevantTx},
+	}
+
+	return response, recoveredAddr, recoveredIndex, relevantTx, nil
+}
+
+// TestRecoveryFilterBlocksDoesNotHoldWriter verifies that a chain backend can
+// obtain a wallet database write transaction while FilterBlocks is running.
+// A network-backed backend may block for an unbounded amount of time, so
+// holding the writer across this call would also block every unrelated wallet
+// operation that needs to persist state.
+func TestRecoveryFilterBlocksDoesNotHoldWriter(t *testing.T) {
+	t.Parallel()
+
+	w, cleanup := testWallet(t)
+	defer cleanup()
+
+	recoveryMgr, scopedMgrs := newRecoveryTestState(t, w)
+
+	height := w.Manager.SyncedTo().Height + 1
+	blockMeta, blockStamp := newRecoveryTestBlock(height)
+	batch := []wtxmgr.BlockMeta{blockMeta}
+	blocks := []*waddrmgr.BlockStamp{blockStamp}
+
+	errWriterHeld := errors.New("wallet writer held during FilterBlocks")
+	filterCalls := 0
+	chainClient := &mockChainClient{}
+	chainClient.filterBlocksFunc = func(*chain.FilterBlocksRequest) (
+		*chain.FilterBlocksResponse, error) {
+
+		filterCalls++
+
+		writerDone := make(chan error, 1)
+		go func() {
+			writerDone <- walletdb.Update(
+				w.db, func(walletdb.ReadWriteTx) error {
+					return nil
+				},
+			)
+		}()
+
+		select {
+		case err := <-writerDone:
+			return nil, err
+
+		case <-time.After(time.Second):
+			return nil, errWriterHeld
+		}
+	}
+
+	err := w.recoverBatch(
+		chainClient, batch, blocks, recoveryMgr.State(), scopedMgrs,
+	)
+	require.NoError(t, err)
+	require.Equal(t, 1, filterCalls)
+}
+
+// TestRecoveryFilterBlocksSerializesAddressCreation verifies that a recovery
+// request's address horizon stays fixed until the batch is committed. Address
+// creation may continue as soon as the backend call returns.
+func TestRecoveryFilterBlocksSerializesAddressCreation(t *testing.T) {
+	t.Parallel()
+
+	w, cleanup := testWallet(t)
+	defer cleanup()
+
+	recoveryMgr, scopedMgrs := newRecoveryTestState(t, w)
+
+	height := w.Manager.SyncedTo().Height + 1
+	blockMeta, blockStamp := newRecoveryTestBlock(height)
+	batch := []wtxmgr.BlockMeta{blockMeta}
+	blocks := []*waddrmgr.BlockStamp{blockStamp}
+
+	errAddressNotSerialized := errors.New(
+		"address creation completed before recovery commit",
+	)
+	addressDone := make(chan error, 1)
+	filterCalls := 0
+	expectedSync := *blockStamp
+
+	chainClient := &mockChainClient{}
+	chainClient.notifyReceivedFunc = func([]address.Address) error {
+		if w.Manager.SyncedTo() != expectedSync {
+			return errAddressNotSerialized
+		}
+
+		return nil
+	}
+	chainClient.filterBlocksFunc = func(*chain.FilterBlocksRequest) (
+		*chain.FilterBlocksResponse, error) {
+
+		filterCalls++
+
+		addressGoroutineStarted := make(chan struct{})
+		go func() {
+			close(addressGoroutineStarted)
+
+			_, err := w.NewAddress(
+				waddrmgr.DefaultAccountNum,
+				waddrmgr.KeyScopeBIP0084,
+			)
+			addressDone <- err
+		}()
+
+		<-addressGoroutineStarted
+
+		select {
+		case err := <-addressDone:
+			if err != nil {
+				return nil, err
+			}
+
+			return nil, errAddressNotSerialized
+
+		case <-time.After(100 * time.Millisecond):
+		}
+
+		return nil, nil //nolint:nilnil
+	}
+	w.chainClient = chainClient
+
+	err := w.recoverBatch(
+		chainClient, batch, blocks, recoveryMgr.State(), scopedMgrs,
+	)
+	require.NoError(t, err)
+	require.Equal(t, 1, filterCalls)
+	require.Equal(t, expectedSync, w.Manager.SyncedTo())
+
+	select {
+	case err := <-addressDone:
+		require.NoError(t, err)
+
+	case <-time.After(time.Second):
+		t.Fatal("address creation remained blocked after recovery commit")
+	}
+}
+
+// TestRecoveryBatchCommitsResults verifies that discoveries from multiple
+// backend calls and the synced-to marker are committed together after the
+// backend finishes filtering.
+func TestRecoveryBatchCommitsResults(t *testing.T) {
+	t.Parallel()
+
+	w, cleanup := testWallet(t)
+	defer cleanup()
+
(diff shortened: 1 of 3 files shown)


bhandras:
Physical-device validation passed with the recovery I/O change in this PR integrated into Wavelength's Esplora-backed lightweight wallet and rebuilt through gomobile for an iPhone 12 Pro.

The device test used the preceding PR revision. The current head additionally serializes derived-address creation across each recovery batch and adds backend-neutral safety tests; it does not put the wallet database writer back around `FilterBlocks`.

Before the fix, invoice creation repeatedly blocked in `BtcWalletKeyRing.ECDH -> walletdb.Update` for the full 20-second mobile deadline while recovery owned the writer in `FilterBlocks`. It then failed with `DeadlineExceeded` / local `RST_STREAM CANCEL`.

After the fix, two consecutive 2,000-sat invoice requests completed:

| Attempt | Total receive time | Receive-auth ECDH stage |
|---|---:|---:|
| 1 | 984 ms | 69 ms |
| 2 | 882 ms | 2.6 ms |

Both invoices were returned successfully. There was no deadline, goroutine dump, or RST_STREAM error. The second request is especially useful because it rules out a one-off favorable startup result: the key writer remained available across repeated recovery-era wallet operations.

The complete captured device console is stored locally, and the relevant stages were emitted by the same diagnostic Wavelength build used to identify the original writer owner.

Lrifton92 (commented):
Read this against `master` rather than the diff alone, since most of the risk here is in the interaction between the new split and the code that was left untouched. The three-phase structure holds up, and the crash-safety argument in the description is accurate: the sync marker still only advances inside the same `walletdb.Update` that persists the discoveries, so an interrupted batch is simply rescanned. CI is green including `unit-race`.

The test deserves a specific mention: having `filterBlocksFunc` assert that `newAddrMtx.TryLock()` *fails* during filtering (`wallet/recovery_internal_test.go:199`) turns the locking invariant into a real regression test rather than a comment. That is the part most likely to be broken by a future refactor, and it is now covered.

Three observations, none blocking.

**1. Peak memory per batch is no longer O(1).**

`recoverScopedAddresses` now accumulates every response for the batch (`wallet/wallet.go:937`) and hands the whole slice to `applyRecoveryResponses` (`:998`). Previously each `*chain.FilterBlocksResponse` was consumed and released inside the loop before the next `FilterBlocks` call.

Each retained response holds `RelevantTxns []*wire.MsgTx` plus the found-address and outpoint maps (`chain/interface.go:101`), and the `goto expandHorizons` loop (`:967`) produces one response per block that matched, bounded by `recoveryBatchSize = 2000` (`:52`). So for a wallet with dense history, peak retention during a batch is now proportional to (matched blocks × transactions per block) instead of a single response.

That is a real trade-off rather than a defect, but it is worth being deliberate about given the motivating environment is a physical iPhone — peak RSS during recovery is precisely the constrained resource there, and it is being traded for the DB availability win. If you want to bound it without giving up the all-or-nothing rule, flushing at a block boundary and advancing `SetSyncedTo` only up to that block would keep the crash-safety property intact: the marker still never moves ahead of committed work, there are simply more commit points.

**2. The split silently depends on `ReportFound` being idempotent.**

`reportFoundAddresses` (`:976`) advances the in-memory branches during the scan so the next `expandScopeHorizons` sees the new horizon. Then `extendFoundAddresses`, called from `applyRecoveryResponses`, runs the same `ReportFound(index)` loops again over the same indexes.

This is correct today, and I checked it before raising it: `ReportFound` no-ops when `index < brs.nextUnfound` (`wallet/recovery.go:362`), and the `invalidChildren` pruning is idempotent too. But nothing in the new code says the second pass is expected to be a no-op. If `ReportFound` ever grows a counter or a side effect, the horizons corrupt silently and only under recovery. A one-line comment at the `reportFoundAddresses` call site, or skipping the re-report in the commit phase, would pin the invariant down.

**3. The description understates what the batch serializes.**

`newAddrMtx` guards more than derived-address creation: `txToOutputs` (`wallet/createtx.go:176`), `FundPsbt` (`wallet/psbt.go:179`), `CurrentAddress` (`wallet/wallet.go:1859`), `NewAddress` (`:3393`), `NewChangeAddress` (`:3461`) and `ImportAccountDryRun` (`wallet/import.go:317`). So for the duration of a batch the wallet also cannot build or fund a transaction.

This is not a regression — the old code held a wallet-wide `walletdb.Update` across `FilterBlocks`, which blocked those paths just as hard — and the change is strictly an improvement for every unrelated writer. But "normal derived-address creation is held" reads narrower than what actually happens, and that distinction matters to anyone deciding whether to run recovery on a live wallet.

Separately, narrowing `expandScopeHorizons` to `walletdb.ReadBucket` (`:1039`) is a nice touch: it makes "the scan phase cannot write" a compile-time property instead of a convention.

wallet/recovery_internal_test.go
```diff
@@ -0,0 +1,433 @@
...
+)
+
+// newRecoveryTestState creates the recovery state and scoped-manager map used
+// by a production recovery run, including state already persisted in walletdb.
+func newRecoveryTestState(t *testing.T, w *Wallet) (*RecoveryManager,
+	map[waddrmgr.KeyScope]*waddrmgr.ScopedKeyManager) {
+
+	t.Helper()
+
+	const recoveryWindow = 1
```
yyforyongyu: `newRecoveryTestState` owns a separate recovery bootstrap and hard-codes a window of 1 while production owns `w.recoveryWindow` and initialization in `Wallet.recovery`, so the tests can drift from the configured path; extract a production-owned initializer shared by both or construct the test manager from `w.recoveryWindow`.

wallet/recovery_internal_test.go
```diff
@@ -0,0 +1,433 @@
...
+
+		case <-time.After(time.Second):
+			return nil, errWriterHeld
+		}
+	}
+
+	err := w.recoverBatch(
+		chainClient, batch, blocks, recoveryMgr.State(), scopedMgrs,
+	)
+	require.NoError(t, err)
```
yyforyongyu, line 176: Because the callback contains the only observation and this line only asserts `recoverBatch` success, this test passes if `FilterBlocks` is never invoked; record the callback count or observed result and assert it here, and apply the same fix to the serialization test.

wallet/recovery_internal_test.go
```diff
@@ -0,0 +1,433 @@
...
+	blockMeta, blockStamp := newRecoveryTestBlock(height)
+	batch := []wtxmgr.BlockMeta{blockMeta}
+	blocks := []*waddrmgr.BlockStamp{blockStamp}
+
+	errAddressLockAvailable := errors.New("address lock available")
+	chainClient := &mockChainClient{}
+	chainClient.filterBlocksFunc = func(*chain.FilterBlocksRequest) (
+		*chain.FilterBlocksResponse, error) {
+
+		if w.newAddrMtx.TryLock() {
```
yyforyongyu: `TryLock` tests the private implementation while `NewAddress` is invoked only after recovery returns, so this does not prove address creation remains blocked through the final commit; start `NewAddress` during the controlled filter call, assert it remains blocked through commit, and then assert completion.

yyforyongyu (commented):
Blocking review findings are attached inline. The separate `ReportFound` SSOT finding is already covered by existing review discussion and is not duplicated here.

wallet/recovery_internal_test.go
Lrifton92: You are right, and this lands on something I got wrong in my own review.

I singled out this test as the part that "turns the locking invariant into a real regression test rather than a comment". That praise was too generous, and your point exposes exactly why: the only observation lives inside `filterBlocksFunc`, while the test body asserts nothing beyond `require.NoError`. It is a test that can pass without ever exercising the mechanism it is named after.

Worth making the failure mode concrete, because it is reachable rather than theoretical: `recoverScopedAddresses` returns early on `len(batch) == 0`, so any future change that leaves the batch empty — or any rewiring of the mock — silently skips `FilterBlocks` entirely and both `TestRecoveryFilterBlocksDoesNotHoldWriter` and `TestRecoveryFilterBlocksSerializesAddressCreation` stay green while proving nothing. That is precisely the regression these tests exist to catch, so the vacuous-pass window is the worst possible place to have a blind spot. Recording the callback invocation and asserting it in the body, as you suggest, closes it.

Your `:199` point stands on its own too. `TryLock` inside the filter callback does establish that the lock is held during the backend call, which is the part I was reacting to, but `NewAddress` only runs after `recoverBatch` has returned — so nothing in the current shape distinguishes "held through the commit" from "released as soon as filtering ended". Given that the whole design rests on the horizon staying fixed until the batch is committed, that is the assertion the test should be making.

No disagreement on `:26` either. Thanks for the sharper read.

bhandras: Fixed. Both lock tests now count and assert `FilterBlocks` invocations, so they cannot pass if the callback is skipped.

wallet/recovery_internal_test.go
bhandras: Fixed. The test now starts `NewAddress` from the controlled filter callback, verifies it cannot complete before the recovery sync commit, and verifies it completes afterward. It no longer probes `newAddrMtx`.

wallet/recovery_internal_test.go
bhandras: Fixed. The helper now builds the manager with `w.recoveryWindow`, matching the wallet’s configured production value.

bhandras:
@yyforyongyu @Lrifton92 Review points are addressed, folded into the single commit, and CI is green. Please re-review when you have a moment.

Lrifton92 (commented):
Thanks for the rewrite — both points from my previous review are addressed, and I verified that
against the new revision rather than taking it on trust:

- Every test now carries a `filterCalls` counter with an explicit `require.Equal`, so none of them can
  pass vacuously if `FilterBlocks` is never invoked.
- `TryLock` is gone. `TestRecoveryFilterBlocksSerializesAddressCreation` now calls `w.NewAddress`
  for real and proves both halves of the invariant: it does not complete during the scan, and it does
  unblock once the batch commits. That is the property that actually matters.

I built the branch and ran `go test -count=5 -run TestRecovery ./wallet/` locally; the assertions pass
consistently. I could not run `-race` in my environment, so I have not verified the tests under the
race detector.

One thing I would like your take on before this lands.

### `newAddrMtx` is now held across unbounded network I/O, and it guards more than address creation

`recoverBatch` takes `w.newAddrMtx` and holds it across `recoverScopedAddresses`, which is where the
`chainClient.FilterBlocks` call happens. The doc comment describes this as "Address creation is
serialized across the scan", but `newAddrMtx` is also taken by:

- `txToOutputs` (`wallet/createtx.go`) — the transaction construction path
- `FundPsbt` (`wallet/psbt.go`)
- `ImportAccountDryRun` (`wallet/import.go`)
- `CurrentAddress`, `NewAddress`, `NewChangeAddress` (`wallet/wallet.go`)

So for the duration of a remote `FilterBlocks` call — over a batch of up to `recoveryBatchSize = 2000`
blocks — callers cannot build or fund a transaction either, not just derive a fresh address. On a
slow or stalled backend that is the same unbounded-wait failure mode the PR is fixing, just moved to a
narrower lock.

I think this is still a clear net improvement over holding the walletdb writer, and it may well be an
acceptable trade during recovery. But two things would help:

1. Widen the doc comment on `recoverBatch` (and the one on the `newAddrMtx` field) to say that the
   send path is serialized too, so the next reader does not have to trace the callers to find out.
2. If blocking `txToOutputs` for a full batch is not intended, the horizon could be pinned by
   snapshotting the request's address set rather than by holding the mutex across the network call.

### Minor

- `filterResponses` now accumulates every `FilterBlocksResponse` — including `RelevantTxns` — for a
  whole batch before the commit, where the previous code folded each response into the open
  transaction as it arrived. It is bounded by one batch, but peak memory during recovery of a
  heavily-used wallet now scales with matches per batch. Worth a line in the commit message.
- In `TestRecoveryFilterBlocksSerializesAddressCreation`, `close(addressStarted)` runs before
  `w.NewAddress` is called, so `<-addressStarted` only proves the goroutine was scheduled. The real
  proof is the 100 ms window below it — which is fine, just noting the channel name promises more
  than it delivers.

bhandras:
@Lrifton92 Good catch on the lock's full caller set. For transaction construction and funding, the unbounded wait does remain. The important difference is its blast radius: the old walletdb writer blocked every writer, including the ECDH/auth path from #1317; this lock blocks only operations that can advance address indexes.

I do not think moving `newAddrMtx` off `FilterBlocks` is safe with only the existing request snapshot. The request copies the address maps, so concurrent derivation cannot mutate that in-flight request. But the recovery state is resurrected from persisted key counts only once. If a slow backend call overlaps more than `recoveryWindow` index advances, the persisted count can move beyond the copied horizon. A nil response can then let recovery commit the batch sync marker without scanning those newly known indexes for that historical range. That can miss historical activity from another wallet using the same seed.

Serializing only the final commit fixes the address manager's `OnCommit` ordering, but not that scan-coverage gap. Removing the narrow lock safely would need horizon versioning plus retry/re-scan, or an equivalent refresh protocol. That is a broader change than this PR.

I kept the batch lock and folded the useful review points into `88fc8b3`:

- `newAddrMtx` and `recoverBatch` now state that transaction construction and funding are serialized across each recovery batch.
- `addressStarted` is renamed to `addressGoroutineStarted`, matching what the channel proves.
- The commit message records the bounded whole-batch response retention.

Validation: focused recovery tests passed 10 times, the focused race run passed, the full wallet suite passed, and `golangci-lint run ./wallet` reports 0 issues.

Lrifton92 (approved):
Your reasoning on the scan-coverage gap convinces me, and I'd rather say why than just approve.

I had been thinking of the request's copied address maps as the whole invariant, so "snapshot the horizon
instead of holding the lock" looked equivalent. It isn't. The part I was missing is that the recovery state
is resurrected from persisted key counts exactly once: if a slow backend call overlaps more than
`recoveryWindow` index advances, the persisted count moves past the copied horizon, and a nil response then
lets the batch sync marker commit over a range those indexes were never scanned for. Missing historical
activity from another wallet on the same seed is a much worse failure than a bounded wait on
`txToOutputs`, and it fails silently. Agreed that horizon versioning plus retry belongs in its own PR.

The blast-radius framing is fair too — this lock is strictly narrower than the walletdb writer it replaces,
including for the ECDH path from #1317.

One suggestion, and it's the reason I'm writing this out rather than just approving: that argument currently
lives only in this thread. The code says *what* is serialized, not *why it cannot be narrowed*. Someone
looking at `recoverBatch` in six months sees a mutex held across a network call — which is exactly the shape
of thing people "fix" — and nothing on the page tells them that dropping it reintroduces a silent
scan-coverage hole. A couple of lines above the `w.newAddrMtx.Lock()` capturing the persisted-count-vs-copied-
horizon reason would make that durable. Entirely your call whether it's worth the diff.

LGTM otherwise. The tests now assert `filterCalls` on every path, the rollback test covers the partial-batch
case, and the renamed channel says what it actually proves.

bhandras:
@yyforyongyu ptal 🙏

yyforyongyu (approved):
LGTM

GustavoStingelin (approved):
LGTM

PR #1319, opened 2026-08-12
wallet+db: record SQL database identity

yyforyongyu:
## Summary

Record the accepted durable SQL database identity design.

## Change Description

Define the identity tuple, signet challenge handling, read-only preflight, native first-open exclusion, metadata bootstrap, and forward-migration order for SQLite and PostgreSQL. This is an ADR-only change with no Store, schema, migration, generated, or test changes.

### DIFF
--- a/docs/developer/adr/0014-sql-database-identity.md
+++ b/docs/developer/adr/0014-sql-database-identity.md
@@ -1,0 +1,327 @@
+# ADR 0014: Durable SQL Database Identity
+
+## Status
+
+- **Status:** Accepted
+- **Date:** 2026-08-19
+
+## Relationships
+
+- **Amends:** None.
+- **Supersedes:** None.
+- **Amended by:** None.
+- **Superseded by:** None.
+
+## 1. Problem
+
+Migration version alone does not prove that a SQL namespace belongs to the
+btcwallet role wallet or the requested Bitcoin network. A foreign,
+wrong-network, malformed, dirty, or newer namespace must be rejected before
+btcwallet changes durable state.
+
+SQLite owns a complete database file. PostgreSQL owns the fixed `btcwallet`
+schema within a database that may also contain lnd or other application data.
+
+## 2. Context
+
+The current SQLite path enables WAL before constructing a migration driver,
+which may create its default migration table. The PostgreSQL path creates its
+default migration table and wallet objects through the ambient `search_path`.
+Neither path validates durable ownership before mutation. The SQLite driver
+lock is process-local, and neither driver provides the required identity-first,
+caller-owned atomic transaction.
+
+### Constraints
+
+- Identity must be validated before persistent pragmas, migration metadata,
+  migrations, sqlc queries, or wallet access can mutate the namespace.
+- Only an empty unmarked namespace is adoptable. Populated unmarked and legacy
+  namespaces require a separate conversion decision.
+- Concurrent first openers must use database-native exclusion, not a
+  process-local correctness lock.
+- One Store open performs one read-only preflight and one locked recheck.
+  Ordinary wallet queries perform no identity checks.
+- Identity is independent of ordinary migration progress and runtime wallet
+  parameters.
+- This decision adds no sidecar lock, create manifest, recovery state machine,
+  Store implementation, schema change, generated output, or migration.
+
+## 3. Decision
+
+### Identity contract
+
+Every role-wallet SQL namespace contains exactly one immutable identity row:
+
+| Field | Required value and encoding |
+| --- | --- |
+| `schema_family` | Exact case-sensitive UTF-8 text `btcwallet-sql` |
+| `identity_generation` | Integer `1` |
+| `genesis_hash` | Exact 32 bytes from `chaincfg.Params.GenesisHash[:]` |
+| `network_magic` | Mathematical unsigned 32-bit `Params.Net` value |
+| `signet_challenge_digest` | Trusted 32-byte signet digest; NULL otherwise |
+
+Genesis bytes use `chainhash.Hash` array order, not the reversed display order
+from `Hash.String()`. PostgreSQL stores network magic in `BIGINT`, preserving
+the full unsigned uint32 value. Comparisons are exact and byte-for-byte.
+
+Generation 1 is the only supported generation. A future generation requires a
+superseding decision with an explicit compatibility and transition path; it is
+not an ordinary schema migration.
+
+The identity excludes `Params.Name`, coinbase maturity, address and WIF
+prefixes, Bech32 HRP, HD private and public versions, and BIP44 coin type. Those
+values affect presentation, policy, import/export, or per-wallet and per-scope
+behavior; their existing subsystems remain responsible for them.
+
+### Signet identity
+
+All signets share one genesis block, and network magic contains only four bytes
+of the challenge digest. The full identity input is:
+
+```text
+DoubleSHA256(byte(len(challenge)) || challenge)
+```
+
+This is the one-byte-length-prefixed serialization used by
+`chaincfg.CustomSignetParams`. Manager configuration must carry the trusted
+32-byte result in returned array order, derived from the same challenge used to
+construct those params. Before database access, the little-endian uint32 from
+its first four bytes must equal `uint32(Params.Net)`.
+
+A network with the fixed signet genesis is signet regardless of `Params.Name`
+and requires the digest. A non-signet requires NULL. Missing or internally
+inconsistent input is rejected before database access. An
+equal-magic/different-challenge stored identity is rejected during preflight
+before mutation.
+
+### Metadata ownership
+
+PostgreSQL stores every role-wallet object in schema `btcwallet` and never
+depends on the ambient `search_path`. SQLite treats the complete file as its
+namespace. Their identity tables are:
+
+| Backend | Identity table |
+| --- | --- |
+| SQLite | `btcwallet_database_identity` |
+| PostgreSQL | `btcwallet.database_identity` |
+
+The columns are `id`, `schema_family`, `identity_generation`, `genesis_hash`,
+`network_magic`, and `signet_challenge_digest`. `id` is the primary key,
+constrained to 1. Required text is nonempty, integer fields enforce their
+ranges, and byte fields enforce their lengths. SQLite uses `INTEGER` for the
+three integers, `TEXT` for family, and `BLOB` for hashes. PostgreSQL uses
+`SMALLINT` for `id`, `BIGINT` for generation and magic, `TEXT` for family, and
+`BYTEA` for hashes.
+
+After bootstrap the table has exactly one row selected by `id = 1`. A wrong
+shape, zero or multiple rows, a NULL required field, or an invalid value is
+malformed and foreign.
+
+Ordinary migration state remains separate:
+
+| Backend | Migration-state table |
+| --- | --- |
+| SQLite | `btcwallet_schema_migrations` |
+| PostgreSQL | `btcwallet.schema_migrations` |
+
+Both tables use golang-migrate's standard `version` and `dirty` columns. An
+empty table is the sole clean no-version state. After the first migration the
+table contains exactly one row with the highest applied positive version and a
+logical Boolean dirty value. SQLite uses `INTEGER` and 0/1; PostgreSQL uses
+`BIGINT` and `BOOLEAN`. When identity exists, a missing migration-state
+relation, NULL, zero, invalid Boolean, or multiple rows are malformed. Identity
+generation never derives from or tracks migration version.
+
+The existing golang-migrate database drivers own transaction boundaries and
+cannot satisfy the atomic initialization contract. The initialization executor
+must execute migration bodies and maintain the standard version row through
+the caller-owned pinned transaction without creating metadata or starting an
+independent transaction first.
+
+### Preflight and classification
+
+Requested identity is validated in memory before database access. Each open
+then performs a logically non-mutating preflight:
+
+- SQLite reads `sqlite_schema` and recognized metadata through a read-only
+  connection, ignoring only SQLite-owned internal objects.
+- PostgreSQL reads the fixed schema through catalog queries. Before first
+  adoption, it also scans non-system schemas for these reserved legacy table
+  names: `blocks`, `wallets`, `wallet_secrets`, `wallet_sync_states`,
+  `address_types`, `key_scopes`, `key_scope_secrets`, `accounts`,
+  `account_secrets`, `addresses`, `derived_addresses`, `address_secrets`,
+  `transactions`, `utxos`, `tx_replacements`, and `utxo_leases`.
+
+Any reserved table name outside the fixed schema rejects first adoption,
+regardless of shape, so partial or malformed unqualified wallets fail closed.
+An unrelated migration-state table alone is not a footprint. After a valid
+identity exists in `btcwallet`, objects outside that schema are irrelevant.
+
+Logically non-mutating means no schema, identity, migration-state, wallet,
+journal-mode, or other logical database change. Backend-internal coordination
+required to read existing state is permitted.
+
+| Observed state | Outcome after locked recheck |
+| --- | --- |
+| Empty and unmarked | Bootstrap identity, then migrate |
+| Matching, clean, current | Open without schema work |
+| Matching, clean, old or no-version | Apply pending forward migrations |
+| Populated and unmarked | Reject without durable mutation |
+| Foreign or malformed | Reject without durable mutation |
+| Matching but dirty | Reject without durable mutation |
+| Matching but newer than the binary supports | Reject without mutation |
+| Concurrent matching openers | Serialize and converge |
+| Concurrent conflicting openers | Winner commits; loser rejects |
+
+For SQLite, empty means no user object. For PostgreSQL, it means the fixed
+schema is absent or has no user object and no reserved legacy name exists
+elsewhere. Any object in an unmarked fixed schema is populated. An identity
+table with no row is malformed, not empty. Legacy or populated-unmarked state
+is never adopted automatically.
+
+`Current` means the highest applied version equals the binary's highest
+supported schema version. `Old` means the table is empty or holds a lower
+positive version. `Newer` means `btcwallet_schema_migrations.version` on SQLite
+or `btcwallet.schema_migrations.version` on PostgreSQL exceeds that supported
+version. Tuple mismatch includes wrong family, generation, genesis, magic, or
+signet digest.
+
+### Atomic initialization
+
+Every candidate that passes preflight repeats the complete classification on
+one context-bounded pinned connection in one transaction under native
+exclusion:
+
+- SQLite starts the transaction with `BEGIN IMMEDIATE`.
+- PostgreSQL begins an explicit transaction-scoped `READ COMMITTED`
+  transaction and makes `pg_advisory_xact_lock` its first statement. This does
+  not change database, role, session, or runtime Store defaults. The lock is
+  local to the current database and released at transaction end.
+
+The PostgreSQL signed bigint lock key is the big-endian two's-complement int64
+formed from the first eight bytes of
+`SHA256(ASCII("btcwallet-sql-identity-migration"))`:
+`-1270285209963246243`.
+
+Initialization then follows this order:
+
+1. Repeat catalog, metadata-shape, tuple, and migration-state classification.
+2. Roll back immediately if the result is a rejection case.
+3. If empty, create the fixed PostgreSQL schema when applicable, create the
+   identity and migration-state relations, and insert the identity row. Leave
+   migration state empty.
+4. For new or matching-old state, apply only pending forward migrations in
+   numeric order and maintain version/dirty state in the same transaction.
+5. Commit. Failure rolls back bootstrap, migrations, and migration state.
+6. For SQLite only, enable WAL after commit.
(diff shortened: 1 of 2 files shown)


docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,501 @@
...
+own. After identity and migrations commit, runtime setup may enable WAL in the
+accepted database.
+
+## 2. Decision
+
+### Identity tuple
+
+Every role-wallet SQL database has exactly one immutable identity tuple:
+
+1. Schema family: the exact UTF-8 string `btcwallet-role-wallet`.
```
yyforyongyu: just use `btcwallet-sql`

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,501 @@
...
+2. Identity generation: the integer `1`.
+3. Genesis hash: 32 canonical bytes.
+4. Network magic: an unsigned 32-bit integer.
+5. Behavior digest: a 32-byte version-1 digest.
```
yyforyongyu: why this field?

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,501 @@
...
+The behavior digest covers maintained Wallet behavior that is not already
+represented by genesis, magic, or the signet digest. It is SHA-256 of the
+following concatenation, with no separators or terminator other than the one
+explicit zero byte:
+
+1. ASCII bytes `btcwallet-role-wallet/behavior/v1`.
+2. One zero byte.
+3. The network-name byte length as uint16 big-endian.
+4. The exact UTF-8 bytes of `Params.Name`.
+5. `Params.CoinbaseMaturity` as uint16 big-endian.
```
yyforyongyu: this is an over design - all these properties should be handled by the account/address tables? and for the current known network configs, these values are already set?

GustavoStingelin:
Good cleanup. I’ll review it again later. I didn’t submit the previous review because there were too many things that needed elaboration.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+  `search_path`, and the wallet migrations create unqualified objects there.
+  This does not give the role wallet a fixed ownership namespace.
+
+The database identity must therefore be checked before either migration helper
+is constructed, before ordinary migrations run, and before wallet queries are
+created or executed. SQLite also needs this check before any persistent pragma
+such as `journal_mode=WAL`.
+
+This decision defines the contract that the durable-identity implementation
+must follow. It does not implement that contract.
```
GustavoStingelin: I wonder if we could remove this from the ADR. Since the ADR is mainly about documenting the decision and the contract, this implementation scope note feels a bit out of place here.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+
+## 2. Decision
+
+### Identity tuple
+
+Every role-wallet SQL database has exactly one immutable identity tuple:
+
+1. Schema family: the exact UTF-8 string `btcwallet-sql`.
+2. Identity generation: the integer `1`.
+3. Genesis hash: 32 canonical bytes.
```
GustavoStingelin: `identity_generation` versions the whole tuple, but the current classification treats any mismatch as foreign before migrations run. Could we explain how a future generation would be introduced without stranding generation-1 databases? If that transition is intentionally future work, it may be clearer to say generation 1 is the only supported value and a future generation requires a superseding decision and compatibility path.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+signed 64-bit column so every uint32 value is representable without changing
+its value.
+
+The tuple deliberately excludes `Params.Name`, coinbase maturity, address and
+WIF prefixes, Bech32 HRP, HD private and public versions, and BIP44 coin type.
+Those values affect presentation, import and export, wallet policy, or
+per-wallet and per-scope behavior; they do not identify the owning SQL schema
+or Bitcoin chain. Account, address, key-scope, and runtime validation retain
+responsibility for them. Changing one of these values does not make an
+otherwise matching database foreign.
```
GustavoStingelin: 👍🏼

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+Manager configuration must therefore carry the trusted full challenge digest
+as separate identity input. A network with the fixed signet genesis is
+classified as signet regardless of `Params.Name` and must have a 32-byte
+challenge digest. A non-signet must not supply that digest. Missing signet
+input and equal-magic/different-challenge signets are rejected before
+mutation.
+
+### Owned metadata
+
+PostgreSQL reserves schema `btcwallet` for every role-wallet object, including
```
GustavoStingelin: 👍🏼

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+the identity and migration state. Runtime behavior must not depend on an
+ambient `search_path`; catalog reads and metadata references use the fixed
+schema, and the ordinary migration executor explicitly targets it.
+
+SQLite treats the complete database file as its role-wallet namespace. Its
+metadata names carry the btcwallet prefix to avoid looking like the old
+default migration marker.
```
GustavoStingelin: Maybe we should also define the default SQLite database file location and name?

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+- `network_magic`: an integer in `[0, 4294967295]`.
+- `signet_challenge_digest`: NULL or exactly 32 bytes.
+
+SQLite uses `INTEGER`, `TEXT`, and `BLOB` storage. PostgreSQL uses `SMALLINT`
+for `id`, `BIGINT` for identity generation and network magic, `TEXT` for the
+family, and `BYTEA` for byte fields. `id` is the primary key and is constrained
+to 1. Database constraints enforce every range and length above. After
+bootstrap the table contains exactly one row, always selected by `id = 1`. A
+table with zero or multiple identity rows, the wrong shape, NULL in a required
+field, or an invalid constraint value is malformed and foreign.
```
GustavoStingelin: so, the pk id is just to try to enforce in the database side to not allow the inclusion of other rows? and maybe have a stable identifier for updates/inserts?

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+
+The ordinary migration-state tables are separate:
+
+- SQLite: `btcwallet_schema_migrations`.
+- PostgreSQL: `btcwallet.schema_migrations`.
+
+Both have `id`, `version`, and `dirty`. The defaulted primary-key `id` is
+constrained to 1. Version is a positive SQLite `INTEGER` or PostgreSQL
```
GustavoStingelin: The migration-state shape differs from golang-migrate’s standard `(version, dirty)` table, and the current drivers also own their transactions.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+  namespace does not itself make the database populated.
+- **Matching and current:** the identity tuple matches exactly, migration
+  state is well-formed and clean, and its version equals the binary's supported
+  version. It is reopenable.
+- **Matching and old:** the identity tuple matches exactly, migration state is
+  well-formed and clean, and it has no version or a lower positive supported
+  version. It is reopenable and needs forward migration.
+- **Populated and unmarked:** any user object exists without the identity
+  relation. It is rejected, including the current unqualified wallet or
+  default `schema_migrations` table in `public`.
```
GustavoStingelin: To confirm the tradeoff: scanning all non-system schemas means an existing LND schema would make the PostgreSQL database non-adoptable even when the `btcwallet` schema is absent. If a dedicated object-empty database is intentional, could we state why schema-local ownership is not sufficient? That would make the operational cost and benefit clear.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+- **Foreign or malformed:** the identity relation has the wrong shape or row
+  count, any tuple field differs, or migration state is absent or malformed.
+  It is rejected.
+- **Dirty:** matching identity with dirty migration state. It is rejected.
+- **Newer:** matching identity with a migration version greater than the
+  binary supports. It is rejected.
```
GustavoStingelin: `Newer` could be a little more explicit: matching identity, but `btcwallet_schema_migrations.version` is greater than the highest schema version supported by this binary.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+
+An identity table with no row is malformed, not empty. A version-zero row is
+malformed; an empty, correctly shaped migration-state table is the sole
+no-version representation. No existing populated-unmarked database is adopted,
+even when its objects resemble the current btcwallet schema. Legacy adoption
+requires a separate, explicit future decision.
```
GustavoStingelin: a bit confusing text here

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+  `pg_advisory_xact_lock` with the signed bigint key defined below. The lock is
+  scoped to the current database and is released by transaction end.
+
+The PostgreSQL lock input is this exact byte sequence:
+
+```text
+ASCII("btcwallet-sql") || NUL ||
+ASCII("btcwallet") || NUL ||
+ASCII("identity-migration")
+```
```
GustavoStingelin: This NUL-separated preimage seems more structured than necessary for one stable advisory-lock namespace. Would `SHA256("btcwallet-sql-identity-migration")` be simpler? If the separated components are intentional for collision-resistant composition or future reuse, could we explain that benefit here?

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+   applicable, create the identity and migration-state relations, and insert
+   the one requested identity row. The migration-state relation remains empty
+   until an ordinary migration is applied.
+4. For either a newly marked database or a matching old database, execute only
+   pending forward migrations in numeric order against the fixed btcwallet
+   namespace. Maintain the singleton version/dirty state in the same
+   transaction.
+5. Commit the initialization transaction.
+6. Only after commit, enable SQLite WAL mode and construct runtime Store/sqlc
+   queries. Wallet reads and writes may begin after that setup succeeds.
```
GustavoStingelin: Could we split this final step by backend? For SQLite, enable WAL after commit; for both backends, construct Store/sqlc queries. Keeping the backend-specific WAL step separate from shared setup would make the initialization order easier to scan.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+schema work. A migration failure rolls back the transaction, including fresh
+metadata bootstrap or any migrations attempted by that transaction. This
+atomicity is part of initialization; it does not introduce a general recovery
+state machine.
+
+Matching concurrent openers serialize and converge on the same marked,
+migrated database. Conflicting first openers may both observe an empty database
+during read-only preflight, but only one can bootstrap it. The loser acquires
+the exclusion afterward, sees the winner's identity during the mandatory
+recheck, and rejects without changing the winner's database.
```
GustavoStingelin: it also works for sqlite? but by locking the file? driver do it natively?

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+**Genesis plus network magic.** Custom signets share the fixed signet genesis,
+and different challenges can collide on the four-byte magic derived from their
+digest prefixes.
+
+**Digesting runtime behavior parameters.** Network name, coinbase maturity,
+address and WIF prefixes, Bech32 HRP, HD versions, and coin type affect
+presentation, import and export, wallet policy, or per-scope configuration.
+Binding them into database identity would reject otherwise valid persisted data
+after a runtime-configuration change. Their existing subsystems remain
+responsible for validating and interpreting them.
```
GustavoStingelin: This repeats the identity-tuple rationale from lines 75-81 almost verbatim. Since this is the alternatives section, could we shorten it to the specific rejection reason and refer back to the earlier rationale? That would preserve the decision while reducing repeated text.

docs/developer/adr/0014-sql-database-identity.md
GustavoStingelin: why not just sha256("btcwallet-sql-identity-migration")?

GustavoStingelin: as i think it is just a pseudo random number to identify our software and not colide with others advisory locks, and it should be stable across all future releases. so.  KISS

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+### Locked bootstrap and migration order
+
+All initialization work uses one context-bounded, pinned database connection
+and one database transaction. The native exclusions are:
+
+- SQLite: start the write transaction with `BEGIN IMMEDIATE`. SQLite permits
+  only one writer, so competing initializers wait at this boundary. No
+  process-local mutex is a correctness dependency.
+- PostgreSQL: after beginning the transaction, acquire
+  `pg_advisory_xact_lock` with the signed bigint key defined below. The lock is
```
GustavoStingelin: Could we make the transaction isolation explicit here? If the DSN or server default is `REPEATABLE READ` or `SERIALIZABLE`, this lock query can establish the losing opener’s snapshot before it waits. After the winner commits, the loser may still classify the old empty snapshot and fail during bootstrap instead of seeing the identity and converging. I think the simplest fix is to require a `READ COMMITTED` initialization transaction and make this advisory lock its first statement.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+contains exactly one row holding the highest applied version and its dirty
+state. Identity generation never derives from, tracks, or substitutes for this
+migration version.
+
+### Read-only preflight and classification
+
+The requested identity is validated entirely in memory first. Each open then
+performs a side-effect-free preflight:
+
+- An existing SQLite file is inspected through a read-only connection. The
```
GustavoStingelin: One SQLite nuance: a read-only open is not always free of filesystem side effects for an existing WAL database. SQLite (including the pinned modernc driver) may create or open the `-shm` coordination file while reading the WAL. Could we scope “side-effect-free” to no logical database mutation, journal-mode change, or schema/wallet writes, and explicitly allow transient WAL coordination files? Keeping the absolute wording would otherwise require a custom non-creating VFS strategy.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -0,0 +1,469 @@
...
+0a03cf400fc1ba4f477200d58687ea1f128648d7b20aaacf38c590fc728433df
+```
+
+and the network magic is `0x40cf030a`.
+
+`chaincfg.Params` retains the truncated magic but not the original challenge.
+Manager configuration must therefore carry the trusted full challenge digest
+as separate identity input. A network with the fixed signet genesis is
+classified as signet regardless of `Params.Name` and must have a 32-byte
+challenge digest. A non-signet must not supply that digest. Missing signet
```
GustavoStingelin: Could we make the cross-field signet validation explicit here? Because the full digest is supplied separately from `chaincfg.Params`, a stale or mismatched digest could otherwise be persisted alongside a different `Params.Net`. I suggest requiring `littleEndian(digest[0:4]) == uint32(Params.Net)` and stating that the digest must be derived from the same trusted challenge used to construct `CustomSignetParams`. That keeps the durable identity internally consistent.

GustavoStingelin (changes requested):
looking well designed! left some comments

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: yeah nice catch! will fix

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: def, will update

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: will update, but i think the `shm` detail is too implementation specific, will just say that we need to ensure the identity before claiming the db as its own.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Good question - given that it's unlikely we run a dedicated pg instance for the wallet, will change the design to focus on wallet-related schema instead.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: tbh i think by default all transactions will share the same isolation level, which is the default `READ COMMITTED`.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: yeah an overdesign, will remove

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: generation 1 is now the only supported value, and any future generation requires a superseding decision with an explicit compatibility and transition path.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: removed the redundant implementation-scope note. The scope boundary remains at the end of the ADR.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Thanks!

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Thanks!

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: PostgreSQL ownership is now schema-local, so btcwallet may share a database with lnd. First adoption checks the fixed btcwallet schema plus a deterministic reserved list for legacy unqualified wallet tables.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: migration state now uses the standard version and dirty columns. The ADR also requires migration bodies and version updates to use the caller-owned pinned initialization transaction rather than the current driver-owned transaction boundaries.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: Newer now names the backend-specific migration table and compares its version with the highest schema version supported by the binary.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: simplified the lock input to the single ASCII literal btcwallet-sql-identity-migration and recorded its recomputed SHA-256 digest and signed bigint key.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: SQLite WAL activation is now a separate backend-specific step, followed by shared Store and sqlc construction for both backends.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: shortened the rejected alternative to its specific reason and referred back to the identity-tuple rationale.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: preflight is now defined as logically non-mutating, with no schema, identity, migration-state, wallet, or journal-mode change, while permitting backend-internal read coordination.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: PostgreSQL initialization now requires a transaction-scoped READ COMMITTED transaction with the advisory lock as its first statement. Runtime Store and session defaults remain unchanged.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Addressed in 96fd99b65: the digest must derive from the same trusted challenge used for CustomSignetParams, and its first four bytes must reproduce Params.Net before database access.

GustavoStingelin (approved):
LGTM!

PR #1320, opened 2026-08-12
wallet: define watch-only and script policy

yyforyongyu:
## Summary

Implements roadmap Task 332

Define explicit watch-only wallet creation and script-import policies.

This replaces #1305, which GitHub closed after its base branch was deleted.

## Change Description

- Validate every wallet creation mode before backend mutation.
- Restrict watch-only creation to shell mode with public initial accounts.
- Document SQL script-import limits and legacy kvdb behavior.

### DIFF
--- a/wallet/manager.go
+++ b/wallet/manager.go
@@ -34,12 +34,12 @@ const (
 	// (CreateWalletParams.Seed).
 	ModeImportSeed
 
-	// ModeImportExtKey indicates creating a wallet from an extended key
-	// (CreateWalletParams.RootKey).
+	// ModeImportExtKey indicates creating a spendable wallet from an extended
+	// private root key (CreateWalletParams.RootKey).
 	ModeImportExtKey
 
-	// ModeShell indicates creating an empty wallet shell (no root key).
-	// Intended for importing specific Account XPubs.
+	// ModeShell indicates creating a watch-only wallet shell without a root
+	// key. Its optional initial accounts contain account XPubs.
 	ModeShell
 )
 
@@ -71,10 +71,8 @@ type CreateWalletParams struct {
 	// Seed is required for ModeImportSeed. Ignored for others.
 	Seed []byte
 
-	// RootKey is required for ModeImportExtKey and must be nil for every
-	// other mode, which reject a key they would not use. It must agree with
-	// WatchOnly: an XPrv for a spendable wallet, an XPub for a watch-only
-	// one.
+	// RootKey is the required ModeImportExtKey XPrv. XPub-root creation is
+	// unsupported; use a ModeShell wallet and import account XPubs.
 	RootKey *hdkeychain.ExtendedKey
 
 	// InitialAccounts is optional and names watch-only accounts to import
@@ -82,11 +80,7 @@ type CreateWalletParams struct {
 	// honored for ModeShell.
 	InitialAccounts []WatchOnlyAccount
 
-	// WatchOnly requests a watch-only wallet, and is the sole authority on
-	// whether the created wallet is watch-only. The root key must agree with
-	// it: a watch-only wallet takes an XPub root (ModeImportExtKey) or no
-	// root at all (ModeShell), while a spendable wallet requires a private
-	// root key.
+	// WatchOnly requests a watch-only wallet and is valid only for ModeShell.
 	WatchOnly bool
 
 	// Birthday is the wallet's birthday.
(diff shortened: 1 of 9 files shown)


wallet/manager.go
```diff
@@ -360,6 +335,28 @@ func (p *CreateWalletParams) validate() error {
...
+
+	default:
+		return fmt.Errorf("%w: unknown mode %v", ErrWalletParams, p.Mode)
+	}
+
+	return nil
+}
+
+// validateInitialAccountKeys requires public keys for shell account imports.
+func validateInitialAccountKeys(accounts []WatchOnlyAccount) error {
```
GustavoStingelin, line 352: Could we run the complete initial-account validation before `backend.create`? This helper only rejects nil and private keys, while `importAccountInternal` validates the XPub network, account depth, hardened index, and address-type mapping later, after the wallet has been persisted and published in `m.wallets`.

I reproduced this with a public depth-0 XPub on both SQLite and kvdb: `Create` returned `invalid account key`, but after closing and reopening the manager, `Load` succeeded. That leaves a durable wallet even though creation was reported as failed.

wallet/manager_test.go
```diff
@@ -232,117 +233,207 @@ func TestManagerCreateError(t *testing.T) {
...
+				chainParams: &chainParams,
+			}
+			cfg := Config{
+				Chain: &bwmock.Chain{}, Name: "test-wallet",
+				RecoveryWindow: MinRecoveryWindow,
+			}
+
+			_, err := m.Create(cfg, tc.params)
+			require.ErrorIs(t, err, ErrWalletParams)
+			require.False(t, backend.createCalled)
```
GustavoStingelin: The valid and invalid rows currently exercise different APIs: valid rows call the private `params.validate()` helper, while invalid rows call `Manager.Create`. That can hide integration gaps; for example, the valid shell fixture uses a depth-0 master XPub that passes `params.validate()` but is rejected by the real account-import path, after wallet creation has begun.

Could every row call `Manager.Create` and use a dedicated sentinel from `recordingManagerBackend.create`? Valid rows can then assert the sentinel and `createCalled == true`, while invalid rows keep asserting `ErrWalletParams` and `createCalled == false`. The valid shell fixture should also derive a real depth-3 account XPub.

GustavoStingelin (changes requested):

wallet/manager.go
yyforyongyu: Yeah this is the atomic wallet creation i mentioned previously, which is tracked and broken down into multiple tasks. Also pre-existing, so out of this task's scope. For instance check task 365 and 329.

wallet/manager_test.go
yyforyongyu: Nice catch! Will update.

yyforyongyu: Updated in a9f1aea39909: every policy row now calls `Manager.Create`; accepted rows assert a dedicated backend sentinel and `createCalled`, while rejected rows assert `ErrWalletParams` and no backend call. The shell fixture now derives a real depth-3 account XPub.

GustavoStingelin (approved):
LGTM

wallet/manager.go
```diff
@@ -294,27 +262,17 @@ func (w *Wallet) importInitialAccounts(ctx context.Context,
 //
 //nolint:cyclop
 func (p *CreateWalletParams) validate() error {
```
Abdulkbk, line 264: With the new changes to `validate`, `validateWatchOnlyRootKey` seems redundant, since the `Create` call invokes `validate` first and already covers watch-only cases. I think we can do away with it and also `TestValidateWatchOnlyRootKey`, Then collapse `prepareWalletCreation` to just `deriveRootKey`.

wallet/manager.go
```diff
@@ -360,6 +335,28 @@ func (p *CreateWalletParams) validate() error {
...
+}
+
+// validateInitialAccountKeys requires public keys for shell account imports.
+func validateInitialAccountKeys(accounts []WatchOnlyAccount) error {
+	for i, account := range accounts {
+		switch {
+		case account.XPub == nil:
+			return fmt.Errorf("%w: account %d needs XPub", ErrWalletParams, i)
+		case account.XPub.IsPrivate():
+			return fmt.Errorf("%w: account %d needs XPub", ErrWalletParams, i)
```
Abdulkbk: we're returning the same error fofr 2 different cases?

Abdulkbk (changes requested):

wallet/manager.go
yyforyongyu: Updated in dca260f936e2: `Manager.Create` now calls `deriveRootKey` directly after exhaustive parameter validation. I removed `prepareWalletCreation`, `validateWatchOnlyRootKey`, and the redundant standalone test; the public `Manager.Create` policy matrix retains the pre-mutation coverage.

wallet/manager.go
yyforyongyu: Updated in dca260f936e2 and f19d8b77274d: a nil key still reports `needs XPub`, while a private key now reports `needs XPub, not XPrv`. Both wrap `ErrWalletParams`, and the policy matrix asserts the distinct diagnostics before any backend call.

Abdulkbk (approved):
LGTM

PR #1322, opened 2026-08-12
docs: separate live watches from recovery

yyforyongyu:
## Summary

- Separate live-tip watch registration from explicit historical recovery.
- Clarify targeted rescan state and current wallet-wide matching.

## Change Description

Implements roadmap Task 400

Imports may register addresses with the chain backend for new activity, but do
not start a historical scan. Historical recovery remains explicit. Targeted
`Rescan(...)` leaves global `SyncedTo` unchanged while requested accounts
limit derivation horizons rather than the wallet-wide match set.

### DIFF
--- a/docs/developer/adr/0004-targeted-rescan-vs-rewind.md
+++ b/docs/developer/adr/0004-targeted-rescan-vs-rewind.md
@@ -1,5 +1,13 @@
 # ADR 0004: Targeted Rescan vs. Global Rewind
 
+> **Superseded in part:** ADR 0005 supersedes Section 2.2's automatic import
+> trigger. Imports may register a live watch after persistence, but they do not
+> start historical recovery; callers invoke `Rescan(...)` explicitly. Section
+> 2.2's account-exclusive target description is also superseded: requested
+> accounts limit derivation horizons, while recovery-relevant addresses and
+> UTXOs are loaded wallet-wide. A targeted rescan still does not rewind the
+> global `SyncedTo` watermark.
+
 ## 1. Context
 
 In `btcwallet`, discovering missing transactions has historically required a "Rescan." The legacy implementation treated all rescans as a "Rewind":
--- a/docs/developer/adr/0005-no-auto-rescan-on-import.md
+++ b/docs/developer/adr/0005-no-auto-rescan-on-import.md
@@ -2,19 +2,28 @@
 
 ## 1. Context
 
-When importing new keys, addresses, or accounts into a wallet (e.g., via `ImportPrivateKey` or `ImportAccount`), the wallet needs to scan the blockchain history to discover any existing funds associated with these new credentials.
+When importing new keys, addresses, or accounts into a wallet, the wallet needs
+to scan the blockchain history to discover any existing funds associated with
+these new credentials.
 
 A common pattern in some wallet implementations is to automatically trigger a rescan immediately upon import. However, this approach introduces several issues:
+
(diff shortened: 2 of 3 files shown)


docs/developer/adr/0005-no-auto-rescan-on-import.md
```diff
@@ -2,44 +2,227 @@
 
 ## 1. Context
 
-When importing new keys, addresses, or accounts into a wallet (e.g., via `ImportPrivateKey` or `ImportAccount`), the wallet needs to scan the blockchain history to discover any existing funds associated with these new credentials.
+Importing a key, address, script, or account adds wallet state that needs two
```
yyforyongyu: old doc is clear enough? why do we need to touch existing ADRs?

docs/developer/adr/0004-targeted-rescan-vs-rewind.md
```diff
@@ -2,57 +2,153 @@
...
 
-This "Global Rewind" approach is problematic for modern use cases like importing a single private key or account.
-*   **Disruption:** It forces the entire wallet to be "unsynced" for minutes or hours, blocking critical operations like creating transactions, even though the existing keys are perfectly up-to-date.
-*   **Inefficiency:** It re-scans the chain for *all* wallet addresses, not just the imported ones.
+1. Move the wallet's global `SyncedTo` height back to the start block.
+2. Put the wallet in the `Syncing` state.
+3. Reprocess every block from that height for every wallet target.
 
-We need a mechanism to scan for specific keys without disrupting the global wallet state.
+A global rewind is unnecessarily disruptive when the caller needs to recover
```
yyforyongyu: this only enriches the doc, but all these are redudant

yyforyongyu:
Review findings for `8cbea66576486eecba6d6dfada9c5e671d5fbf1d`:

- `docs/developer/adr/0005-no-auto-rescan-on-import.md:12-17` lacks the PR's stated accepted, pending, and ready live-watch states plus its atomic database-obligation, asynchronous-delivery, and restart timeline; add the authoritative state-transition contract and mark unimplemented portions accordingly.
- `docs/developer/adr/0004-targeted-rescan-vs-rewind.md:38` still says targeted rescans contain only requested targets, contradicting the wallet-wide matching documented at `docs/developer/scanning_sync_architecture.md:74-76` and `:95-100`; expand the supersession notice or align both documents.
- `docs/developer/adr/0005-no-auto-rescan-on-import.md:16` says imports only persist data and return immediately, while `wallet/address_manager.go:844-856` and `:894-915` synchronously call `NotifyReceived` after persistence; separate database acceptance from live delivery or align the maintained import paths.

docs/developer/adr/0005-no-auto-rescan-on-import.md
```diff
@@ -5,26 +5,35 @@
 When importing new keys, addresses, or accounts into a wallet (e.g., via `ImportPrivateKey` or `ImportAccount`), the wallet needs to scan the blockchain history to discover any existing funds associated with these new credentials.
```
Abdulkbk: so `ImportPrivateKey(..., rescan=true)` (though deprecated) will still build a rescan job and fire it. Maybe we should not cite it here or mention that it's an exception.

Abdulkbk (changes requested):
I left one suggestion

docs/developer/adr/0005-no-auto-rescan-on-import.md
yyforyongyu: Done!

Abdulkbk (approved):
LGTM

PR #1324, opened 2026-08-13
wallet: restore imported XPub derivation

yyforyongyu:
## Summary

- Restore semantic public-key derivation for imported extended public-key
  accounts.
- Accept selector-bearing parameters directly in Signer.DerivePubKey without
  exposing backend account identifiers.
- Preserve private-signing rejection for imported-XPub children.

## Change Description

Implements roadmap Task 337

### DIFF
--- a/wallet/signer.go
+++ b/wallet/signer.go
@@ -693,8 +693,9 @@ func (w *Wallet) ComputeUnlockingScript(ctx context.Context,
 // privKeyForOutput returns the private key needed to sign for the given
 // wallet-controlled output.
 //
-// Derived addresses resolve through the account-level secret; imported
-// addresses have no derivation path and resolve through their own encrypted
+// Derived addresses resolve through the account-level secret. A derived child
+// without wallet-seed derivation metadata, such as an imported-XPub child,
+// cannot sign. Only raw imported addresses resolve through their own encrypted
 // private key material in the store.
 func (w *Wallet) privKeyForOutput(ctx context.Context,
 	scriptInfo OutputScriptInfo) (
@@ -702,6 +703,10 @@ func (w *Wallet) privKeyForOutput(ctx context.Context,
 
 	if canUseAddressInfoDerivation(scriptInfo.AddressInfo) {
 		return w.privKeyForAddressInfo(ctx, scriptInfo.AddressInfo)
+	}
+
+	if !scriptInfo.Imported {
+		return nil, ErrNoAssocPrivateKey
 	}
 
 	return w.resolveImportedAddrPrivKey(ctx, scriptInfo.scriptPubKey())
--- a/wallet/signer.go
+++ b/wallet/signer.go
@@ -45,12 +45,26 @@ var (
 	ErrAccountNotInStore = errors.New("account not in store")
 )
 
+// DerivePubKeyParams identifies an account and the unhardened child key to
+// derive from its extended public key.
+type DerivePubKeyParams struct {
+	// Account identifies the account by portable wallet semantics.
+	Account AccountSelector
+
+	// Branch is the account child branch to derive.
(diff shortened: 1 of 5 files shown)


wallet/signer.go
```diff
@@ -468,26 +479,67 @@ var _ SpendDetails = (*LegacySpendDetails)(nil)
...
+// checked numeric adapter for PublicKeyDeriver: Account may be zero when
+// unspecified, but a present Account must match InternalAccount's hardened
+// derivation child.
 func (w *Wallet) DerivePubKey(ctx context.Context, path BIP32Path) (
 	*btcec.PublicKey, error) {
 
+	account := path.DerivationPath.InternalAccount
+	accountChild := path.DerivationPath.Account
+
+	if accountChild != 0 {
```
yyforyongyu: add inline comments here explaining this check

wallet/public_derivation_external_test.go
```diff
@@ -0,0 +1,54 @@
...
+	"testing"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// publicKeyDeriver records whether an exported derivation request was invoked.
+type publicKeyDeriver struct {
```
yyforyongyu: we should always use `mock.Mock`!!

wallet/signer.go
```diff
@@ -45,6 +45,16 @@ var (
 	ErrAccountNotInStore = errors.New("account not in store")
 )
 
+// PublicKeyDeriver derives public keys from semantic wallet account identity.
+// It supports accounts without a BIP44 account number, such as imported XPub
+// accounts, without exposing backend account identifiers.
+type PublicKeyDeriver interface {
```
yyforyongyu: why do we need this to be a public interface?

wallet/signer.go
```diff
@@ -468,26 +479,67 @@ var _ SpendDetails = (*LegacySpendDetails)(nil)
...
+		Branch: path.DerivationPath.Branch,
+		Index:  path.DerivationPath.Index,
+	})
+}
+
+// DerivePubKeyForAccount derives a public child key from a semantically
+// selected account. The account XPub is read from the durable store, then the
+// branch and child index are derived in memory. Public derivation requires a
+// started wallet but remains available while the wallet is locked.
+func (w *Wallet) DerivePubKeyForAccount(ctx context.Context,
```
yyforyongyu: don't add a new method, instead, fix it in `DerivePubKey`

yyforyongyu: Addressed by removing `DerivePubKeyForAccount`. Semantic account selection is implemented directly in `Signer.DerivePubKey` through `DerivePubKeyParams`.

wallet/signer.go
yyforyongyu: Addressed by removing the legacy BIP32 numeric adapter and its account-field consistency check. `Signer.DerivePubKey` now accepts `DerivePubKeyParams` with the Task 341 semantic account selector.

wallet/signer.go
yyforyongyu: Addressed by removing `PublicKeyDeriver`. `Signer` is now the only public component for this derivation operation.

wallet/public_derivation_external_test.go
yyforyongyu: Addressed. The external `Signer` stub now embeds `mock.Mock`, registers the exact selector-bearing parameters, asserts expectations, and safely checks the configured public-key result type.

GustavoStingelin (approved):
LGTM!

PR #1321, opened 2026-08-12
pg: use native pgx for PostgreSQL storage

yyforyongyu:
## Summary

Use native pgx for normal PostgreSQL store reads, writes, and transactions
while keeping SQLite and migration tooling on `database/sql`.

## Change Description

Implements roadmap Task 350

- generate PostgreSQL queries for `pgx/v5` and adapt PostgreSQL values at the
  backend boundary
- make the shared transaction runtime driver-neutral while preserving retry,
  health, fatal-error, rollback, and ambiguous-commit behavior
- retain migration and test-administration compatibility through the pgx
  pool's `database/sql` adapter

## Status

This PR is deprioritized and is not currently being pursued. Please do not
review it.

PR #1326, opened 2026-08-14
itest: add AccountManager mutation integration tests

GustavoStingelin:
Implements roadmap Task 429 by adding daemon-level integration coverage for the
public AccountManager mutation APIs:

- `NewAccount`
- `RenameAccount`
- `ImportAccount`

## Test structure

The account-manager suite has 15 independently registered cases in
`itest/account_manager_test.go`: five creation, five rename, and five import
cases. Successful mutations, rejection and rollback checks, and lifecycle gates
are isolated instead of sharing one long scenario. Rejection tables contain
only inputs and expected results.

Every case runs unchanged against kvdb, SQLite, and PostgreSQL. There is no
backend branch anywhere in the suite; backend-specific code is confined to
harness setup and teardown.

## Coverage

The scenarios cover:

- complete public `AccountInfo` results from mutation, lookup, and wallet
  reload;
- the absent, present-zero, and present-nonzero forms of both optional
  identities — a present-zero `AccountNumber` via the default account and a
  present-zero `MasterKeyFingerprint` via a zero-fingerprint import — so a
  backend cannot collapse a present zero into an absent value;
- rejection of duplicate, invalid-name, invalid-key, reserved-name, and
  missing-source requests, with the public account set left unchanged;
- locked, stopped, and watch-only lifecycle gates, each reopening the wallet
  afterwards to prove the rejected call left no partial account behind;
- derived, default, and imported-XPub renames, including both reserved
  imported-name directions;
- persisted and dry-run imports; and
- complete serialized account-key comparisons after normalizing only the
  network version bytes.

## Scope

Stable public error identities for AccountManager rejections are owned by
Task 378 and are **not** established here. The duplicate-account and
rename-not-found mappings that an earlier revision of this PR carried have been
removed; the negative cases assert rejection and unchanged state only, and will
be tightened to `errors.Is`/`errors.As` in the dedicated Task 378 PR.

`AccountID` is intentionally outside this public mutation coverage. Account
query coverage remains scoped to Task 430.

## Supporting fixes

Two production fixes remain, both required for the shared contract to hold on
every backend:

- **Imported account address schemas** convert explicitly between the database
  and `waddrmgr` enums instead of casting ordinals. The two enums share names
  but not values, so the cast silently rewrote the schema of an imported BIP44
  account. This mismatch pre-exists the sql-wallet base and has no introducing
  commit; see the commit body for detail.
- **kvdb renames imported-XPub accounts by name**, while still refusing the
  reserved raw-import pseudo-account, so rename behaves identically on kvdb and
  SQL.

Carrying these here is a one-time exception, taken to avoid further churn in an
already-restacked branch. Production bugs discovered during future itest work
should be fixed in separate, focused PRs rather than folded into the test PR
that happened to surface them.

## Harness

The shared `bwtest` fixture supports rootless and seeded watch-only wallets and
keeps reload metadata private in one manager-keyed ownership registry. Wallets
are registered before start, reload replaces the sole wallet and manager it
consumes, and teardown stops wallets before closing managers.

## Testing

```bash
make lint-check
make unit
make itest chain=btcd db=kvdb icase="account manager"
make itest chain=btcd db=sqlite icase="account manager"
make itest chain=btcd db=postgres icase="account manager"
```

### DIFF
--- a/bwtest/harness.go
+++ b/bwtest/harness.go
@@ -38,6 +38,11 @@ const (
 	defaultChainReconnectAttempts = 5
 )
 
+// walletRegistration tracks every wallet owned by one Manager. Each map value
+// is a copied reload configuration, or nil for wallets registered through
+// RegisterWallet.
+type walletRegistration map[*wallet.Wallet]*wallet.Config
+
 // HarnessTest is the integration test harness.
 type HarnessTest struct {
 	*testing.T
@@ -68,16 +73,12 @@ type HarnessTest struct {
 	// parent harness and its subtests.
 	postgresServer *postgresTestServer
 
-	// managers holds the wallet managers created for this subtest. They are
-	// closed after every registered wallet has stopped.
-	managers []*wallet.Manager
+	// wallets owns the wallet registrations for each Manager.
+	wallets map[*wallet.Manager]walletRegistration
 
 	// mu protects harness state that can be accessed across the main test and
 	// subtests. This includes the wallet registry and idempotent shutdown.
 	mu sync.Mutex
-
-	// wallets is the set of wallets created by a test case.
-	wallets []*wallet.Wallet
 
 	// stopped prevents stopping shared infrastructure more than once.
 	stopped bool
@@ -239,7 +240,14 @@ func (h *HarnessTest) NewWalletManager() *wallet.Manager {
 	})
 	require.NoError(h, err, "failed to create wallet manager")
 
-	h.managers = append(h.managers, manager)
+	h.mu.Lock()
+
+	if h.wallets == nil {
+		h.wallets = make(map[*wallet.Manager]walletRegistration)
+	}
+
+	h.wallets[manager] = make(walletRegistration)
+	h.mu.Unlock()
 
 	// Verify the Manager opened the backend requested by the -db flag without
 	// adding a production accessor solely for the test.
@@ -252,11 +260,22 @@ func (h *HarnessTest) NewWalletManager() *wallet.Manager {
 func (h *HarnessTest) teardownWallets(ctx context.Context) error {
 	err := h.stopActiveWallets(ctx)
 
-	for _, manager := range h.managers {
+	h.mu.Lock()
+
+	managers := make([]*wallet.Manager, 0, len(h.wallets))
+	for manager := range h.wallets {
+		managers = append(managers, manager)
+	}
+
+	h.mu.Unlock()
+
+	for _, manager := range managers {
 		err = errors.Join(err, manager.Close())
 	}
 
-	h.managers = nil
+	h.mu.Lock()
+	h.wallets = nil
+	h.mu.Unlock()
 
 	return err
 }
@@ -330,26 +349,52 @@ func validateFileBackendArtifact(dbType, dbPath string) error {
 	return nil
 }
 
-// RegisterWallet registers a wallet with the harness.
+// RegisterWallet registers a wallet with its harness-owned manager.
 //
 // Registered wallets are automatically included in harness-level assertions,
 // such as MineBlocks.
-func (h *HarnessTest) RegisterWallet(w *wallet.Wallet) {
+func (h *HarnessTest) RegisterWallet(manager *wallet.Manager,
+	w *wallet.Wallet) {
+
 	h.Helper()
 
-	if w == nil {
-		h.Fatalf("cannot register nil wallet")
-	}
+	h.registerWallet(manager, w, nil)
+}
+
+// registerWallet records a wallet under manager and copies reloadConfig when
+// provided.
+func (h *HarnessTest) registerWallet(manager *wallet.Manager, w *wallet.Wallet,
+	reloadConfig *wallet.Config) {
+
+	h.Helper()
 
 	h.mu.Lock()
-	h.wallets = append(h.wallets, w)
-	h.mu.Unlock()
+	defer h.mu.Unlock()
+
+	registration, ok := h.wallets[manager]
+	require.True(h, ok, "manager is not owned by this harness")
+	require.NotNil(h, w, "cannot register nil wallet")
+
+	for _, registered := range h.wallets {
+		_, exists := registered[w]
+		require.False(
+			h, exists,
+			"wallet is already registered with this harness",
+		)
+	}
+
+	var config *wallet.Config
+	if reloadConfig != nil {
+		copiedConfig := *reloadConfig
+		config = &copiedConfig
+	}
+
+	registration[w] = config
 }
 
 // DeregisterWallet releases a wallet from harness ownership.
 //
-// It returns false when the wallet is nil or was not registered. The remaining
-// registrations retain their original order.
+// It returns false when the wallet is nil or was not registered.
 func (h *HarnessTest) DeregisterWallet(w *wallet.Wallet) bool {
 	h.Helper()
 
@@ -360,12 +405,12 @@ func (h *HarnessTest) DeregisterWallet(w *wallet.Wallet) bool {
 	h.mu.Lock()
 	defer h.mu.Unlock()
 
-	for i, registered := range h.wallets {
-		if registered != w {
+	for _, registration := range h.wallets {
+		if _, ok := registration[w]; !ok {
 			continue
 		}
 
-		h.wallets = append(h.wallets[:i], h.wallets[i+1:]...)
+		delete(registration, w)
 
 		return true
 	}
@@ -375,8 +420,8 @@ func (h *HarnessTest) DeregisterWallet(w *wallet.Wallet) bool {
 
 // ReleaseManager releases a Manager from harness teardown ownership.
 //
-// It returns false when the Manager is nil or was not registered. The remaining
-// registrations retain their original order.
+// It returns false when the Manager is nil, was not registered, or has any
+// registered wallets.
 func (h *HarnessTest) ReleaseManager(manager *wallet.Manager) bool {
 	h.Helper()
 
@@ -384,25 +429,33 @@ func (h *HarnessTest) ReleaseManager(manager *wallet.Manager) bool {
 		return false
 	}
 
-	for i, registered := range h.managers {
-		if registered != manager {
-			continue
-		}
+	h.mu.Lock()
+	defer h.mu.Unlock()
 
-		h.managers = append(h.managers[:i], h.managers[i+1:]...)
-
-		return true
+	registration, ok := h.wallets[manager]
+	if !ok || len(registration) != 0 {
+		return false
 	}
 
-	return false
+	delete(h.wallets, manager)
+
+	return true
 }
 
 // ActiveWallets returns a snapshot of wallets registered with this harness.
+// The order is unspecified.
 func (h *HarnessTest) ActiveWallets() []*wallet.Wallet {
 	h.Helper()
 
 	h.mu.Lock()
-	wallets := append([]*wallet.Wallet(nil), h.wallets...)
+
+	var wallets []*wallet.Wallet
+	for _, registration := range h.wallets {
+		for w := range registration {
+			wallets = append(wallets, w)
+		}
+	}
+
 	h.mu.Unlock()
 
 	return wallets
@@ -495,14 +548,7 @@ func (h *HarnessTest) stopActiveWallets(ctx context.Context) error {
 
 	var stopErr error
 
-	for i, w := range h.ActiveWallets() {
-		if w == nil {
-			// Keep cleanup robust against partially initialized test state. A
-			// caller could register a wallet reference and fail before the
-			// assignment completes.
-			continue
-		}
-
+	for _, w := range h.ActiveWallets() {
 		// The modern Wallet controller's Stop method is idempotent.
 		//
 		// NOTE: We intentionally don't call the deprecated WaitForShutdown/
@@ -510,9 +556,9 @@ func (h *HarnessTest) stopActiveWallets(ctx context.Context) error {
 		// legacy fields initialized.
 		err := w.Stop(ctx)
 		if err != nil {
-			stopErr = errors.Join(stopErr, fmt.Errorf(
-				"stop wallet %d: %w", i, err,
-			))
+			stopErr = errors.Join(
+				stopErr, fmt.Errorf("stop wallet: %w", err),
+			)
 		}
 	}
 
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -85,6 +85,15 @@ type WalletFixture struct {
 	// combined with Unstarted.
 	Amounts []btcutil.Amount
 
+	// WatchOnly creates a rootless watch-only shell wallet when
+	// InitialAccounts is empty.
+	WatchOnly bool
+
+	// InitialAccounts seeds a watch-only shell wallet with the provided
+	// accounts. A nil or empty slice creates no accounts, and a non-empty
+	// slice implies WatchOnly.
+	InitialAccounts []wallet.WatchOnlyAccount
+
 	// Unlocked unlocks the wallet once it has started.
 	Unlocked bool
 
@@ -106,6 +115,10 @@ func (h *HarnessTest) NewWallet(fixture WalletFixture) (*wallet.Wallet,
 	h.Helper()
 
 	cfg, params := h.TestWalletConfig()
+	if fixture.WatchOnly || len(fixture.InitialAccounts) != 0 {
+		params.Mode, params.WatchOnly = wallet.ModeShell, true
+		params.InitialAccounts = fixture.InitialAccounts
+	}
 
 	manager := h.NewWalletManager()
 	w, err := manager.Create(cfg, params)
@@ -115,7 +128,7 @@ func (h *HarnessTest) NewWallet(fixture WalletFixture) (*wallet.Wallet,
 	// cleanup owner. Registering after Start would leave a wallet whose Start
 	// failed unregistered, and a second direct Stop callback here would stop a
 	// successful one twice, out of order with the Manager close.
-	h.RegisterWallet(w)
+	h.registerWallet(manager, w, &cfg)
 
 	if fixture.Unstarted {
 		require.Empty(
@@ -137,6 +150,73 @@ func (h *HarnessTest) NewWallet(fixture WalletFixture) (*wallet.Wallet,
 	}
 
 	return w, h.FundWalletOfType(w, fixture.AddrType, fixture.Amounts...)
+}
+
+// ReloadWallet consumes current after shutting it down and returns a fresh
+// registered, started, and locked replacement loaded from the same store.
+// The current wallet must be the Manager's only registered wallet because
+// reload closes and replaces that Manager. Callers must use the returned wallet
+// for any later reload.
+func (h *HarnessTest) ReloadWallet(current *wallet.Wallet) *wallet.Wallet {
+	h.Helper()
+
+	require.NotNil(h, current, "cannot reload nil wallet")
+
+	h.mu.Lock()
+
+	var (
+		manager           *wallet.Manager
+		reloadConfig      *wallet.Config
+		registeredWallets int
+	)
+
+	for candidate, registration := range h.wallets {
+		config, ok := registration[current]
+		if !ok {
+			continue
+		}
+
+		manager = candidate
+		reloadConfig = config
+		registeredWallets = len(registration)
+
+		break
+	}
+
+	h.mu.Unlock()
+
+	require.NotNil(h, manager, "wallet is not registered with this harness")
+	require.NotNil(h, reloadConfig, "wallet is not reloadable")
+	require.Equal(
+		h, 1, registeredWallets,
+		"wallet manager has sibling registered wallets",
+	)
+
+	cfg := *reloadConfig
+
+	ctx := h.Context()
+	require.NoError(
+		h, current.Stop(ctx), "failed to stop wallet before reload",
+	)
+	require.True(
+		h, h.DeregisterWallet(current), "failed to deregister wallet",
+	)
+	require.NoError(
+		h, manager.Close(), "failed to close wallet manager",
+	)
+	require.True(
+		h, h.ReleaseManager(manager), "failed to release wallet manager",
+	)
+
+	manager = h.NewWalletManager()
+	w, err := manager.Load(cfg)
+	require.NoError(h, err, "failed to reload wallet")
+	require.NotSame(h, current, w, "reload returned the original wallet")
+
+	h.registerWallet(manager, w, &cfg)
+	require.NoError(h, w.Start(ctx), "failed to start reloaded wallet")
+
+	return w
 }
 
 // CreateEmptyWallet creates, starts, and registers a new wallet instance.
--- a/itest/account_manager_test.go
+++ b/itest/account_manager_test.go
@@ -1,0 +1,358 @@
+//go:build itest
+
+package itest
+
+import (
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// importedAccountKeys holds deterministic public and private account material
+// used to exercise imported-account contracts without sharing wallet state.
+// accountPrivateKey is present only so a case can assert private keys are
+// refused.
+type importedAccountKeys struct {
+	scope                waddrmgr.KeyScope
+	addrType             waddrmgr.AddressType
+	accountKey           *hdkeychain.ExtendedKey
+	otherAccountKey      *hdkeychain.ExtendedKey
+	accountPrivateKey    *hdkeychain.ExtendedKey
+	masterKeyFingerprint uint32
+}
+
+// canonicalAccountKey returns a derived account key under the network default
+// public version so backend-specific serialized versions compare consistently.
+func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
+	h.Helper()
+
+	parsed, err := hdkeychain.NewKeyFromString(string(key))
+	require.NoError(h, err, "failed to decode account public key")
+	normalized, err := parsed.CloneWithVersion(
+		h.NetParams().HDPublicKeyID[:],
+	)
+	require.NoError(
+		h, err, "failed to normalize account public key version",
+	)
+
+	return []byte(normalized.String())
+}
+
+// deterministicImportedAccountKeys derives distinct BIP84 account XPubs from a
+// fixed root so imported-account cases are reproducible.
+func deterministicImportedAccountKeys(
+	h *bwtest.HarnessTest) importedAccountKeys {
+
+	h.Helper()
+
+	scope := waddrmgr.KeyScopeBIP0084
+	root, err := hdkeychain.NewMaster(
+		[]byte{
+			0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
+			0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f, 0x10,
+		},
+		h.NetParams(),
+	)
+	require.NoError(h, err, "failed to derive deterministic root key")
+
+	defer root.Zero()
+
+	purpose, err := root.Derive(hdkeychain.HardenedKeyStart + scope.Purpose)
+	require.NoError(h, err, "failed to derive BIP84 purpose key")
+
+	defer purpose.Zero()
+
+	coinType, err := purpose.Derive(
+		hdkeychain.HardenedKeyStart + h.NetParams().HDCoinType,
+	)
+	require.NoError(h, err, "failed to derive BIP84 coin type key")
+
+	defer coinType.Zero()
+
+	// Neuter hands the public key the same chain code and parent
+	// fingerprint slices the private key holds, so the two account private
+	// keys below must outlive this call. Zeroing either one rewrites the
+	// XPub being returned.
+	accountPrivateKey, err := coinType.Derive(hdkeychain.HardenedKeyStart)
+	require.NoError(h, err, "failed to derive imported account private key")
+	accountKey, err := accountPrivateKey.Neuter()
+	require.NoError(h, err, "failed to derive imported account public key")
+
+	otherPrivateKey, err := coinType.Derive(hdkeychain.HardenedKeyStart + 1)
+	require.NoError(h, err, "failed to derive second imported private key")
+	otherAccountKey, err := otherPrivateKey.Neuter()
+	require.NoError(h, err, "failed to derive second imported public key")
+
+	return importedAccountKeys{
+		scope:                scope,
+		addrType:             waddrmgr.WitnessPubKey,
+		accountKey:           accountKey,
+		otherAccountKey:      otherAccountKey,
+		accountPrivateKey:    accountPrivateKey,
+		masterKeyFingerprint: purpose.ParentFingerprint(),
+	}
+}
+
+// testAccountManagerCreateAccount verifies that a new derived account's
+// returned view matches an immediate read and survives a wallet reload.
+func testAccountManagerCreateAccount(h *bwtest.HarnessTest) {
+	const accountName = "account manager created"
+
+	scope := waddrmgr.KeyScopeBIP0084
+	ctx := h.Context()
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+
+	created, err := w.NewAccount(ctx, scope, accountName)
+
+	require.NoError(h, err, "failed to create derived account")
+	require.Equal(h, accountName, created.AccountName)
+	require.NotNil(
+		h, created.AccountNumber, "derived account has no number",
+	)
+	require.Zero(h, created.ExternalKeyCount)
+	require.Zero(h, created.InternalKeyCount)
+	require.Zero(h, created.ImportedKeyCount)
+	require.Zero(h, created.ConfirmedBalance)
+	require.Zero(h, created.UnconfirmedBalance)
+	require.False(h, created.IsImported, "derived account is imported")
+	require.False(h, created.IsWatchOnly, "derived account is watch-only")
+	require.NotZero(
+		h, created.CreatedAt, "derived account has no creation time",
+	)
+	require.Equal(h, scope, created.KeyScope)
+	require.Equal(
+		h, waddrmgr.WitnessPubKey, created.AddrSchema.ExternalAddrType,
+	)
+	require.Equal(
+		h, waddrmgr.WitnessPubKey, created.AddrSchema.InternalAddrType,
+	)
+	normalizedPublicKey := canonicalAccountKey(h, created.PublicKey)
+	require.NotEmpty(
+		h, normalizedPublicKey, "derived account has no public key",
+	)
+	require.NotNil(
+		h, created.MasterKeyFingerprint,
+		"derived account has no master key fingerprint",
+	)
+	require.NotZero(
+		h, *created.MasterKeyFingerprint,
+		"derived account has no master key fingerprint",
+	)
+	want := *created
+	want.PublicKey = normalizedPublicKey
+	got, err := w.GetAccount(ctx, scope, accountName)
+	require.NoError(h, err, "failed to read created account")
+
+	gotInfo := *got
+	gotInfo.PublicKey = canonicalAccountKey(h, gotInfo.PublicKey)
+	require.Equal(h, want, gotInfo)
+
+	w = h.ReloadWallet(w)
+	durable, err := w.GetAccount(ctx, scope, accountName)
+	require.NoError(h, err, "failed to read account after reload")
+
+	durableInfo := *durable
+	durableInfo.PublicKey = canonicalAccountKey(h, durableInfo.PublicKey)
+	require.Equal(h, want, durableInfo)
+}
+
+// testAccountManagerCreateAccountSequence verifies that derived account numbers
+// are allocated contiguously within a key scope, that each scope allocates from
+// its own counter, and that the counter survives a wallet reload.
+func testAccountManagerCreateAccountSequence(h *bwtest.HarnessTest) {
+	const (
+		otherScopeName = "account manager sequence other"
+		resumedName    = "account manager sequence next"
+	)
+
+	names := []string{
+		"account manager sequence one",
+		"account manager sequence two",
+		"account manager sequence three",
+	}
+
+	ctx := h.Context()
+	scope := waddrmgr.KeyScopeBIP0084
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+
+	numbers := make([]wallet.AccountNumber, 0, len(names))
+	for _, name := range names {
+		created, err := w.NewAccount(ctx, scope, name)
+		require.NoError(h, err, "failed to create %q", name)
+		require.NotNil(h, created.AccountNumber, "%q has no number", name)
+
+		numbers = append(numbers, *created.AccountNumber)
+	}
+
+	for i := 1; i < len(numbers); i++ {
+		require.Equal(
+			h, numbers[i-1]+1, numbers[i],
+			"%q did not take the next account number", names[i],
+		)
+	}
+
+	// A second scope allocates from its own counter, so its first
+	// user-created account repeats the first scope's starting number.
+	other, err := w.NewAccount(ctx, waddrmgr.KeyScopeBIP0044, otherScopeName)
+	require.NoError(h, err, "failed to create other scope account")
+	require.NotNil(h, other.AccountNumber, "other scope has no number")
+
+	require.Equal(
+		h, numbers[0], *other.AccountNumber,
+		"second scope did not allocate from its own counter",
+	)
+
+	// The counter is persistent, so the next account continues the sequence
+	// rather than reusing a number.
+	w = h.ReloadWallet(w)
+	h.UnlockWallet(w)
+
+	resumed, err := w.NewAccount(ctx, scope, resumedName)
+	require.NoError(h, err, "failed to create account after reload")
+	require.NotNil(
+		h, resumed.AccountNumber, "resumed account has no number",
+	)
+	require.Equal(
+		h, numbers[len(numbers)-1]+1, *resumed.AccountNumber,
+		"account number did not resume the sequence after reload",
+	)
+}
+
+// testAccountManagerRejectAccountCreation verifies rejected names cannot alter
+// an existing account or its scope's account count.
+func testAccountManagerRejectAccountCreation(h *bwtest.HarnessTest) {
+	const (
+		sourceName         = "account manager rejection source"
+		afterRejectionName = "account manager after rejection"
+	)
+
+	scope := waddrmgr.KeyScopeBIP0084
+	ctx := h.Context()
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+
+	_, err := w.NewAccount(ctx, scope, sourceName)
+	require.NoError(h, err, "failed to create rejection source")
+
+	existing, err := w.GetAccount(ctx, scope, sourceName)
+	require.NoError(h, err, "failed to read rejection source")
+
+	wantExisting := *existing
+	wantExisting.PublicKey = canonicalAccountKey(h, wantExisting.PublicKey)
+	accounts, err := w.ListAccountsByScope(ctx, scope)
+	require.NoError(h, err, "failed to list source scope accounts")
+
+	wantCount := len(accounts)
+
+	// These rejections have no stable public error identity yet, so the
+	// rows assert rejection and unchanged state only.
+	testCases := []struct {
+		name        string
+		accountName string
+	}{
+		{
+			name:        "duplicate source name",
+			accountName: sourceName,
+		},
+		{
+			name:        "empty name",
+			accountName: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		_, err := w.NewAccount(ctx, scope, tc.accountName)
+
+		require.Error(h, err, "%s was accepted", tc.name)
+
+		current, err := w.GetAccount(ctx, scope, sourceName)
+		require.NoError(
+			h, err, "failed to read source account after rejection",
+		)
+
+		currentInfo := *current
+		currentInfo.PublicKey = canonicalAccountKey(
+			h, currentInfo.PublicKey,
+		)
+		require.Equal(h, wantExisting, currentInfo)
+
+		accounts, err = w.ListAccountsByScope(ctx, scope)
+		require.NoError(
+			h, err, "failed to list accounts after rejection",
+		)
+		require.Len(
+			h, accounts, wantCount, "rejection changed account count",
+		)
+	}
+
+	// The account number is allocated before the insert, so only the
+	// rolled-back transaction keeps a rejected call from consuming one.
+	next, err := w.NewAccount(ctx, scope, afterRejectionName)
+	require.NoError(h, err, "failed to create account after rejections")
+	require.NotNil(
+		h, next.AccountNumber, "account after rejections has no number",
+	)
+	require.Equal(
+		h, *wantExisting.AccountNumber+1, *next.AccountNumber,
+		"rejected creation consumed an account number",
+	)
+}
+
+// testAccountManagerEnforceAccountCreationLifecycle verifies NewAccount admits
+// neither locked nor stopped wallets.
+func testAccountManagerEnforceAccountCreationLifecycle(h *bwtest.HarnessTest) {
+	const (
+		lockedName  = "account manager locked"
+		stoppedName = "account manager stopped"
+	)
+
+	scope := waddrmgr.KeyScopeBIP0084
+	ctx := h.Context()
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+
+	accounts, err := w.ListAccounts(ctx)
+	require.NoError(h, err, "failed to list accounts before rejections")
+
+	wantCount := len(accounts)
+
+	require.NoError(h, w.Lock(ctx), "failed to lock wallet")
+
+	_, err = w.NewAccount(ctx, scope, lockedName)
+
+	require.Error(h, err, "locked wallet created an account")
+	require.NoError(h, w.Stop(ctx), "failed to stop wallet")
+
+	_, err = w.NewAccount(ctx, scope, stoppedName)
+
+	require.ErrorIs(h, err, wallet.ErrStateForbidden)
+
+	// An admission rejected before or after the wallet stops must leave no
+	// partial account, so neither name resolves across a reopen and the
+	// account set is unchanged.
+	w = h.ReloadWallet(w)
+	_, err = w.GetAccount(ctx, scope, lockedName)
+	require.Error(h, err, "locked rejection created an account")
+	_, err = w.GetAccount(ctx, scope, stoppedName)
+	require.Error(h, err, "stopped rejection created an account")
+
+	accounts, err = w.ListAccounts(ctx)
+	require.NoError(h, err, "failed to list accounts after rejections")
+	require.Len(h, accounts, wantCount, "rejection changed account count")
+}
+
+// testAccountManagerRejectWatchOnlyAccountCreation verifies a rootless wallet
+// cannot derive account key material.
+func testAccountManagerRejectWatchOnlyAccountCreation(h *bwtest.HarnessTest) {
+	const accountName = "watch-only"
+
+	ctx := h.Context()
+	w, _ := h.NewWallet(bwtest.WalletFixture{WatchOnly: true})
+	require.True(h, w.IsWatchOnly(), "watch-only fixture is not watch-only")
+
+	_, err := w.NewAccount(ctx, waddrmgr.KeyScopeBIP0084, accountName)
+
+	require.Error(h, err, "watch-only wallet created an account")
+	_, err = w.GetAccount(ctx, waddrmgr.KeyScopeBIP0084, accountName)
+	require.Error(h, err, "rejected watch-only account name resolves")
+}
--- a/itest/account_manager_test.go
+++ b/itest/account_manager_test.go
@@ -3,12 +3,18 @@
 package itest
 
 import (
+	"github.com/btcsuite/btcd/btcutil/v2"
 	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
 	"github.com/btcsuite/btcwallet/bwtest"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet"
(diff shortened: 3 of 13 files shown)


itest/account_manager_fixtures_test.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+// newAccountManagerWallet creates, registers, starts, and unlocks one wallet,
+// and returns it with the handles reloadAccountManagerWallet needs.
+//
+// Account mutations derive from the master HD private key, so the wallet is
+// unlocked. Registration happens before Start so harness teardown owns the
+// wallet even when Start fails.
+func newAccountManagerWallet(h *bwtest.HarnessTest) (wallet.Config,
```
yyforyongyu: cool let's merged #1316 first then we can rebase and use that new wallet fixture to create the testing wallet.

itest/account_manager_fixtures_test.go
```diff
@@ -0,0 +1,84 @@
...
+	h.RegisterWallet(w)
+	require.NoError(h, w.Start(h.Context()), "failed to start wallet")
+	h.UnlockWallet(w)
+
+	return cfg, manager, w
+}
+
+// reloadAccountManagerWallet stops a wallet and its Manager, hands both back to
+// the harness, and reopens the wallet from the same data source.
+func reloadAccountManagerWallet(h *bwtest.HarnessTest, cfg wallet.Config,
```
yyforyongyu: can also move this to be a method on `h`, sth like `h.ReloadWallet`?  waiting on #1316 tho

itest/account_manager_import_test.go
```diff
@@ -0,0 +1,196 @@
...
+		accountName = "account manager imported"
+		dryRunName  = "account manager dry run"
+		invalidName = ""
+	)
+
+	ctx := h.Context()
+	scope := waddrmgr.KeyScopeBIP0084
+	addrType := waddrmgr.WitnessPubKey
+
+	seed := []byte{
```
yyforyongyu: nit: add some docs explaining that we are preparing two private keys for the tests

itest/account_manager_import_test.go
```diff
@@ -0,0 +1,196 @@
...
+	w, err := manager.Create(cfg, params)
+	require.NoError(h, err, "failed to create watch-only shell wallet")
+
+	h.RegisterWallet(w)
+	require.NoError(h, w.Start(ctx), "failed to start watch-only shell wallet")
+	require.True(h, w.IsWatchOnly(), "watch-only fixture is not watch-only")
+	walletInfo, err := w.Info(ctx)
+	require.NoError(h, err, "failed to inspect watch-only wallet")
+	require.True(h, walletInfo.Locked, "watch-only wallet was not locked")
+	require.NoError(h, w.Lock(ctx), "failed to lock watch-only wallet")
```
yyforyongyu: yeah so L59 to L80 will be gone once we use the new methods from #1316

itest/account_manager_import_test.go
```diff
@@ -0,0 +1,196 @@
...
+	require.Equal(
+		h, uint8(waddrmgr.WitnessPubKey),
+		uint8(imported.AddrSchema.ExternalAddrType),
+	)
+	require.Equal(
+		h, uint8(waddrmgr.WitnessPubKey),
+		uint8(imported.AddrSchema.InternalAddrType),
+	)
+	require.Equal(h, publicKey, imported.PublicKey)
+	require.Equal(h, fingerprint, imported.MasterKeyFingerprint)
```
yyforyongyu: so could we create an `expectedImported` struct and perform `require.Equal` check on these two structs instead?

itest/account_manager_import_test.go
```diff
@@ -0,0 +1,196 @@
...
+	)
+	require.ErrorIs(h, err, wallet.ErrInvalidAccountKey)
+	require.Nil(h, invalid)
+
+	_, err = w.GetAccount(ctx, scope, "account manager invalid private")
+	require.Error(h, err, "private-key import left a row")
+
+	// ImportAccount is admitted while started and rejects only after Stop.
+	require.NoError(h, w.Stop(ctx), "failed to stop wallet")
+	invalid, err = w.ImportAccount(
```
yyforyongyu: Given that this is the occurring pattern of `ImportAccount` + `GetAccount` + assertion, feels like this can be table-driven? Or is that too much?

yyforyongyu (changes requested):
Looking good! Think we will perform a rebase after #1316 - also will leave some agentic reviews as some gaps are found.

itest/account_manager_import_test.go
yyforyongyu: Guess this can damage the linear flow - so up to you.

itest/account_manager_new_test.go
```diff
@@ -0,0 +1,151 @@
...
+	account, err := w.NewAccount(ctx, scope, accountName)
+	require.NoError(h, err, "failed to create account")
+	require.Equal(
+		h, accountName, account.AccountName, "new account name mismatch",
+	)
+	require.Equal(
+		h, scope, waddrmgr.KeyScope(account.KeyScope),
+		"new account scope mismatch",
+	)
+	require.NotNil(
```
yyforyongyu: Could we assert `AccountID` on the created, queried, and reloaded account? The contract says it is always populated, but kvdb currently omits it from `CreateDerivedAccountRow` while SQL populates it; the test should expose that divergence and this PR should fix the kvdb return value.

itest/account_manager_rename_test.go
```diff
@@ -0,0 +1,178 @@
...
+	require.NoError(h, err, "failed to create duplicate target account")
+	require.NotNil(h, target.AccountNumber)
+	require.NotEmpty(h, target.PublicKey)
+	require.False(h, target.IsImported)
+	wantTargetKey := canonicalAccountKey(h, target.PublicKey)
+
+	// A rename moves the name only. The account number and derived key identify
+	// the keyspace the caller already holds addresses for, so both must survive
+	// unchanged, and the old name must stop resolving.
+	err = w.RenameAccount(ctx, scope, sourceName, renamedName)
```
yyforyongyu: Could we cover both renaming the reserved `imported` account and renaming another account to `imported`? These are explicit `RenameAccount` validation rules; please assert the stable errors and verify that neither source nor target state changes.

itest/account_manager_import_test.go
```diff
@@ -0,0 +1,196 @@
...
+	require.True(h, imported.IsImported)
+	require.True(h, imported.IsWatchOnly)
+	require.Nil(h, imported.AccountNumber)
+	require.Zero(h, imported.ExternalKeyCount)
+	require.Zero(h, imported.InternalKeyCount)
+	require.Zero(h, imported.ImportedKeyCount)
+	require.Zero(h, imported.ConfirmedBalance)
+	require.Zero(h, imported.UnconfirmedBalance)
+	require.NotZero(h, imported.CreatedAt)
+	require.Equal(
```
yyforyongyu: `waddrmgr.AddressType` and `db.AddressType` do not share a numeric contract, so casting both to `uint8` duplicates the mapping and currently passes by coincidence; please assert the resulting public address behavior or compare through the authoritative conversion.

itest/account_manager_fixtures_test.go
```diff
@@ -0,0 +1,84 @@
...
+// waddrmgr, which re-exports it with the key scope's HD version
+// (cloneKeyWithVersion: BIP84 on testnet becomes vpub), while the SQL stores
+// return the bytes the wallet derived from the master key, which keep the
+// network's default version (tpub). Both encode the same point, so comparing
+// the point is the only assertion that holds on every backend today.
+//
+// This normalization applies to derived accounts only: waddrmgr deliberately
+// leaves imported and watch-only account keys as the caller supplied them, so
+// the import test compares the serialized key byte for byte.
+func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
```
yyforyongyu: Converting the extended key to an EC point discards its chain code, depth, parent fingerprint, and child index, allowing a different derivation to pass; please normalize only the version with `CloneWithVersion` and compare the complete serialized extended key.

itest/account_manager_new_test.go
```diff
@@ -0,0 +1,151 @@
...
+		h, account.IsImported, "derived account is reported as imported",
+	)
+	wantKey := canonicalAccountKey(h, account.PublicKey)
+
+	// A rejected duplicate must leave the account that owns the name exactly
+	// as it was: re-derivation under the same name would silently replace a
+	// key the caller may already have handed out addresses for. There is no
+	// maintained public sentinel for the collision, so only the rejection and
+	// the untouched row are asserted.
+	duplicate, err := w.NewAccount(ctx, scope, accountName)
```
yyforyongyu: These duplicate cases use `require.Error`, so unrelated failures pass and kvdb and SQL expose different error forms; please normalize duplicate-account failures to one stable public identity and assert it with `require.ErrorIs` or the existing typed-error mechanism.

itest/account_manager_new_test.go
```diff
@@ -0,0 +1,151 @@
...
+	invalid, err := w.NewAccount(ctx, scope, invalidName)
+	require.Error(h, err, "empty account name was accepted")
+	require.Nil(h, invalid, "invalid account returned account info")
+
+	invalidPostcondition, err := w.GetAccount(ctx, scope, invalidName)
+	require.Error(h, err, "empty account name resolved after rejection")
+	require.Nil(h, invalidPostcondition)
+
+	// Deriving m/purpose'/coin'/account' needs the master HD private key, which
+	// only an unlocked wallet holds, and a stopped wallet admits no operation
```
yyforyongyu: The comment says rejection occurs before consuming an account number, but the test only checks that no row exists; please create a valid account after the rejected calls and assert that its account number is contiguous.

itest/account_manager_import_test.go
GustavoStingelin: I tried this locally, but the whole-struct version becomes more complicated than the field assertions because `ImportAccount` returns `*wallet/internal/db.AccountInfo`. The `itest` package cannot import or name that internal type, and fields such as `CreatedAt` are generated at runtime.

The closest implementation I found requires a mirrored projection:

```diff
-require.Equal(h, accountName, imported.AccountName)
-require.Equal(h, scope, waddrmgr.KeyScope(imported.KeyScope))
-require.True(h, imported.IsImported)
-require.True(h, imported.IsWatchOnly)
-require.Nil(h, imported.AccountNumber)
-require.Zero(h, imported.ExternalKeyCount)
-require.Zero(h, imported.InternalKeyCount)
-require.Zero(h, imported.ImportedKeyCount)
-require.Zero(h, imported.ConfirmedBalance)
-require.Zero(h, imported.UnconfirmedBalance)
-require.NotZero(h, imported.CreatedAt)
-require.Equal(
-    h, uint8(waddrmgr.WitnessPubKey),
-    uint8(imported.AddrSchema.ExternalAddrType),
-)
-require.Equal(
-    h, uint8(waddrmgr.WitnessPubKey),
-    uint8(imported.AddrSchema.InternalAddrType),
-)
-require.Equal(h, publicKey, imported.PublicKey)
-require.Equal(h, fingerprint, imported.MasterKeyFingerprint)
+type importedAccountView struct {
+    AccountName          string
+    KeyScope             waddrmgr.KeyScope
+    IsImported           bool
+    IsWatchOnly          bool
+    AccountNumber        *uint32
+    ExternalKeyCount     uint32
+    InternalKeyCount     uint32
+    ImportedKeyCount     uint32
+    ConfirmedBalance     int64
+    UnconfirmedBalance   int64
+    Created              bool
+    ExternalAddrType     uint8
+    InternalAddrType     uint8
+    PublicKey            []byte
+    MasterKeyFingerprint uint32
+}
+expectedImported := importedAccountView{
+    AccountName:          accountName,
+    KeyScope:             scope,
+    IsImported:           true,
+    IsWatchOnly:          true,
+    Created:              true,
+    ExternalAddrType:     uint8(waddrmgr.WitnessPubKey),
+    InternalAddrType:     uint8(waddrmgr.WitnessPubKey),
+    PublicKey:            publicKey,
+    MasterKeyFingerprint: fingerprint,
+}
+actualImported := importedAccountView{
+    AccountName:          imported.AccountName,
+    KeyScope:             waddrmgr.KeyScope(imported.KeyScope),
+    IsImported:           imported.IsImported,
+    IsWatchOnly:          imported.IsWatchOnly,
+    AccountNumber:        imported.AccountNumber,
+    ExternalKeyCount:     imported.ExternalKeyCount,
+    InternalKeyCount:     imported.InternalKeyCount,
+    ImportedKeyCount:     imported.ImportedKeyCount,
+    ConfirmedBalance:     int64(imported.ConfirmedBalance),
+    UnconfirmedBalance:   int64(imported.UnconfirmedBalance),
+    Created:              !imported.CreatedAt.IsZero(),
+    ExternalAddrType:     uint8(imported.AddrSchema.ExternalAddrType),
+    InternalAddrType:     uint8(imported.AddrSchema.InternalAddrType),
+    PublicKey:            imported.PublicKey,
+    MasterKeyFingerprint: imported.MasterKeyFingerprint,
+}
+require.Equal(h, expectedImported, actualImported)
```

This duplicates the production shape, is longer, and can silently omit future `AccountInfo` fields. Copying `imported` and normalizing dynamic fields would be shorter, but it would derive the expected value from the result under test.

For those reasons, I think the explicit assertions are clearer here and produce more precise failures. Did you have a cleaner struct-comparison pattern in mind, or would you be okay with keeping the field-by-field assertions?

itest/account_manager_import_test.go
GustavoStingelin: Addressed: the import test now documents the two deterministic private account keys and their roles before neutering them for watch-only use.

itest/account_manager_new_test.go
GustavoStingelin: Addressed: kvdb now returns the derived account ID, and the integration test asserts it on the created, queried, and reloaded account.

itest/account_manager_rename_test.go
GustavoStingelin: Addressed: the rename test covers both reserved imported-alias directions, asserts stable errors, and verifies that the source and imported account state remain unchanged.

itest/account_manager_import_test.go
GustavoStingelin: Addressed: db-to-waddrmgr schema conversion is explicit and covered by a BIP44 kvdb regression, and the itest now checks external/internal P2WPKH behavior instead of enum ordinals.

itest/account_manager_fixtures_test.go
GustavoStingelin: Addressed: `canonicalAccountKey` now uses `CloneWithVersion` to normalize only the public HD version and compares the full serialized extended key, preserving the chain code, depth, parent fingerprint, and child index.

itest/account_manager_import_test.go
GustavoStingelin: I kept the field-by-field assertions. `ImportAccount` returns `*wallet/internal/db.AccountInfo`, which `itest` cannot name directly; a mirrored projection would be longer, duplicate the production shape, and still need dynamic `CreatedAt` normalization. The enum ordinal checks were removed separately; the test now checks external/internal P2WPKH behavior instead.

itest/account_manager_new_test.go
GustavoStingelin: Addressed: duplicate derived-account failures now share the stable public `waddrmgr.ErrDuplicateAccount` identity, and the new-account itest asserts it with `waddrmgr.IsError` while preserving the untouched row.

itest/account_manager_fixtures_test.go
GustavoStingelin: Addressed: the standard unlocked-wallet cases now use the shared harness-managed fixture instead of a local constructor.

itest/account_manager_fixtures_test.go
GustavoStingelin: Addressed: the local reload helper was removed, and the tests now use `h.ReloadWallet` so the harness owns the wallet and manager lifecycle.

itest/account_manager_new_test.go
GustavoStingelin: Addressed: after the rejected duplicate, invalid-name, locked, and stopped calls, the test creates another valid account and asserts that it receives the immediate next account number.

itest/account_manager_import_test.go
GustavoStingelin: Addressed: the manual watch-only manager setup is gone. The import test now seeds its watch-only account through `h.NewManagedWallet` with `InitialAccounts` and reuses the returned managed wallet for `h.ReloadWallet`.

itest/account_manager_import_test.go
GustavoStingelin: I kept this flow linear. I tried a table, but the cases have different contracts: duplicate requests must preserve the existing row, invalid inputs must leave no row, and only key failures have a stable public error identity. Encoding those differences required branching and made the test longer, while the later rename, stop, and reload checks still depend on the sequence.

GustavoStingelin (commented):
I made a few corrections to the code; it should still be easy to review, since they were simple adjustments.

bwtest/harness_wallet.go
```diff
@@ -91,6 +100,18 @@ type WalletFixture struct {
 	Unstarted bool
 }
 
+// ManagedWallet bundles a wallet with the Config and Manager needed to replace
+// its current lifecycle generation from the same persistent store.
+//
+// Tests with custom CreateWalletParams may construct a ManagedWallet directly.
+// Its Config, Manager, and Wallet must describe the same generation, and the
+// Manager and Wallet must remain registered with the harness until reload.
+type ManagedWallet struct {
```
yyforyongyu: it seems to me that the only purpose of `ManagedWallet` is to provide a way to restart the wallet? if so, i think it's too heavy to introduce a `NewManagedWallet` just for this, plus a `ManagedWallet` struct, which leaks the lifecycle into the tests. Instead, why don't we just save it in `RegisterWallet` when called inside `NewWallet`? We can rearrange the existing fields to track the manager/wallets/configs data?
https://github.com/btcsuite/btcwallet/blob/a6865a37e0493df62a7d895a47269022fe62ed66/bwtest/harness.go#L73
https://github.com/btcsuite/btcwallet/blob/a6865a37e0493df62a7d895a47269022fe62ed66/bwtest/harness.go#L80

itest/account_manager_import_test.go
```diff
@@ -0,0 +1,217 @@
...
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
+	"github.com/btcsuite/btcd/txscript/v2"
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// testAccountManagerImportAccount verifies one account xpub import through the
+// public AccountManager API, including rename, rollback, and durable reload.
```
yyforyongyu: can we break each into standalone tests? like rename/rollback/reload, why do we keep them in a single test?

itest/account_manager_fixtures_test.go
```diff
@@ -0,0 +1,36 @@
+//go:build itest
```
yyforyongyu: let's use a similar pattern like the `utxo_manager_test.go` and `tx_creator_test.go`, and put all in `account_manager_test.go`?

wallet/internal/db/kvdb/accountstore_createderived.go
```diff
@@ -141,6 +141,12 @@ func (o *createDerivedAccountOps) CreateDerivedAccount(_ context.Context,
 		derived.PublicKey, derived.EncryptedPrivateKey,
 	)
 	if err != nil {
+		if waddrmgr.IsError(err, waddrmgr.ErrDuplicateAccount) {
+			return db.CreateDerivedAccountRow{},
```
yyforyongyu: 👍

wallet/internal/db/kvdb/accountstore_renameaccount.go
```diff
@@ -42,9 +42,13 @@ func (s *Store) RenameAccount(_ context.Context,
 			return err
 		}
 
-		err = assertRenameableAccount(ns, scopedMgr, acctNum)
-		if err != nil {
-			return err
+		if params.AccountNumber != nil {
```
yyforyongyu, line 47: add some inline comments here explaining what's going on here? like `if else if` - why this check, etc.

wallet/internal/db/kvdb/accountstore_createderived.go
```diff
@@ -183,6 +189,7 @@ func (o *createDerivedAccountOps) CreateDerivedAccount(_ context.Context,
 	}
 
 	return db.CreateDerivedAccountRow{
+		AccountID: accountNumber,
```
yyforyongyu: Is this caught by the itest? Seems out of scope? Also I believe there are multiple tasks targeting this area, like 360/377/298 etc.

yyforyongyu (changes requested):
Cool that we are fixing bugs now! So there are two aspects we need to improve here,
- the itest cases are still large, i wonder if we could make them more atomic.
- we should load the roadmap tasks when creating bug fix PRs - the reason we have that roadmap is to avoid duplicate efforts, so it'd be great to link the tasks if this PR happens to fix some of them, or remove the unnecessary fix that's out of scope.

Will send agentic reviews soon for more detailed suggestions.

bwtest/harness_wallet.go
yyforyongyu: Suggested shape: keep the reload state entirely private to `bwtest` and use one ownership registry:

```go
type walletRegistration struct {
    instance     *wallet.Wallet
    reloadConfig *wallet.Config
}

wallets map[*wallet.Manager]*walletRegistration
```

Key the map by Manager because `NewWalletManager` must register ownership before `Create` or `Load`; if either operation fails before producing a wallet, teardown still closes the opened Manager. The Manager is only the map key, so it is not tracked in two places.

For manager-focused tests, change lifecycle registration to `RegisterWallet(manager, w)`. Those tests already hold both values and still invoke the raw Manager API themselves; registration only transfers cleanup ownership. It fills the existing map entry without adding reload metadata.

`NewWallet` also has the exact Manager and Config, so it should attach the wallet and a private Config copy to that same entry while continuing to return only `*wallet.Wallet`:

```go
w, funding := h.NewWallet(fixture)
w = h.ReloadWallet(w)
```

Keep `reloadConfig` private to the harness rather than adding `Wallet.Config()`. The production `Config` contains runtime handles and mutable or sensitive fields, while configuration observability belongs in a sanitized `Wallet.Info` shape and is outside this task.

`ReloadWallet(w)` finds the entry containing `w`, stops the wallet, closes and removes the old Manager entry, creates a newly registered Manager, loads the saved Config, attaches the replacement wallet to the new entry, starts it, and returns the replacement. If `Load` fails, the empty entry installed by `NewWalletManager` still gives teardown ownership of the new Manager.

`DeregisterWallet` clears the wallet from its entry but leaves the Manager owned; `ReleaseManager` removes the entry only after its wallet has been deregistered. `ActiveWallets` derives the non-nil wallet instances from the map. Teardown makes two passes over this one map: stop all attached wallets first, then close all Manager keys.

This removes `ManagedWallet` and `NewManagedWallet` and makes `wallets` the single lifecycle source of truth. Registration order should not be a harness contract; if deterministic iteration is ever required, derive it at the read boundary rather than maintaining a second ownership collection.

itest/account_manager_import_test.go
yyforyongyu: Suggested breakdown, all in one `itest/account_manager_test.go` with focused registered cases:

**NewAccount**
- `account manager create account` — create one valid derived account, assert every public `AccountInfo` field and the narrow `GetAccount` postcondition, then immediately reload and compare the complete result.
- `account manager reject account creation` — data-only duplicate and invalid-name rows with stable error identities and the complete public account set unchanged.
- `account manager enforce account creation lifecycle` — locked and stopped admission only.
- `account manager reject watchonly account creation` — keep the existing focused watch-only fixture and case.

**RenameAccount**
- `account manager rename derived account` — one accepted derived rename, old-name absence and new-name complete result, then immediate reload.
- `account manager rename imported account` — one accepted imported-XPub rename and immediate reload.
- `account manager reject account rename` — data-only duplicate target, invalid target, unknown source, and reserved-name rows; assert stable errors and unchanged relevant source and target state without querying backend-specific pseudo-account representation.
- `account manager enforce account rename lifecycle` — locked metadata rename succeeds and is reloaded; stopped rename is rejected.

**ImportAccount**
- `account manager import account` — one persistent import, complete mutation and `GetAccount` result, then immediate reload.
- `account manager preview account import` — dry-run result plus absence postcondition; no reload because no mutation commits.
- `account manager reject account import` — data-only duplicate-name or identity and invalid-name or key rows with stable errors and unchanged account state.
- `account manager enforce account import lifecycle` — stopped admission only.

Shared setup stays in this file: deterministic imported keys and one complete `AccountInfo` assertion helper. Use `h.NewWallet` and `h.ReloadWallet(w)` only. Keep table rows data-only, and remove address coverage, `AccountID`, exact account-number sequencing, backend-conditioned assertions, and the second harness handoff. Every accepted persistent mutation should cross close and reopen before another accepted mutation can overwrite its durable postcondition.

wallet/internal/db/kvdb/accountstore_createderived.go
GustavoStingelin: it came from https://github.com/btcsuite/btcwallet/pull/1326#discussion_r3783228392

>Could we assert AccountID on the created, queried, and reloaded account? The contract says it is always populated, but kvdb currently omits it from CreateDerivedAccountRow while SQL populates it; the test should expose that divergence and this PR should fix the kvdb return value.

GustavoStingelin:
@yyforyongyu

> the itest cases are still large, i wonder if we could make them more atomic.

I would say they are larger mostly because we are not using any assertion helpers right now, so we end up repeating the same long assertions multiple times.

By more atomic, do you mean splitting some of the longer tests into smaller individual cases? That would make each test smaller, but the total amount of test code would probably remain about the same.

wallet/internal/db/kvdb/accountstore_createderived.go
yyforyongyu: You are right — this was added in response to my earlier review. That request was mistaken: I conflated the internal `CreateDerivedAccountRow` contract with the public `AccountInfo` contract required by Task 429. Task 429 explicitly excludes `AccountID`, and Task 377 removes backend routing identity from the public DTO. Please revert the created, queried, and reloaded `AccountID` assertions and the KVDB return change from this PR. If the internal Store row needs that invariant, it should be justified and tested with its owning internal task rather than through this public mutation itest. Sorry for the churn.

yyforyongyu:
Yes, by more atomic I mean splitting the longer flows into focused registered cases. The concern is reviewability and contract isolation, not reducing the total number of lines. Right now each function combines successful mutations, multiple rejection contracts, lifecycle states, and durability checks, so a reviewer has to reason about one long stateful sequence and state established much earlier in the function. Separate cases give each requirement a clear name and a small Arrange/Act/Assert scope, isolate failures, and let every successful mutation cross its reopen boundary before another mutation changes the state. A shared helper is still useful for the complete repeated `AccountInfo` comparison, but it should not be used to preserve the large sequential flows or hide the behavior under test. I posted the concrete case breakdown here: https://github.com/btcsuite/btcwallet/pull/1326#discussion_r3809478048

yyforyongyu:
I will be extremely detailed here about itest cases, speaking of the pains we've had from the `lnd`'s itests.

The best way is, really, to read the test cases once the code is written and see if it's easy to follow first. There are multiple aspects here, comparing a test case testing 10 items vs 1 item,
- when a test fails, how quickly can we locate the callsite and bugs? 
- when reading the test, how easily will we lose track of the initial state of the test?
- when asserting, how can we be sure a previous assertion flow won't alter the state we try to assert now?
 
Having assertion helpers can be nice, only if they are done right, while it doesn't justify testing multiple cases in one test. The best way to reduce complexity is to keep it simple.

bwtest/harness_wallet.go
GustavoStingelin: Removed `ManagedWallet` and `NewManagedWallet`; reload state stays private to `bwtest`. The single ownership registry is `map[*wallet.Manager]walletRegistration`, where each registration maps its wallets to an optional copied `*wallet.Config`. `NewWallet` registers its wallet before `Start` with a reload config; raw `RegisterWallet` entries have no reload config. `ReloadWallet` requires that wallet to be the manager’s sole registration, stops it, releases the old manager, and registers the loaded replacement under a new manager. `ActiveWallets` derives from this registry, and teardown stops all wallets before closing managers. No lifecycle wrapper or registration-order contract is exposed to tests.

itest/account_manager_import_test.go
GustavoStingelin: Split the combined flows into 12 independently registered cases in `itest/account_manager_test.go`: four creation, four rename, and four import cases. Each successful persistent mutation owns its immediate `GetAccount` and reload verification; rejection tables remain data-only; and locked or stopped admission is isolated in the lifecycle cases.

itest/account_manager_fixtures_test.go
GustavoStingelin: Consolidated the deterministic key setup, shared helpers, and all 12 create, rename, and import cases into `itest/account_manager_test.go`, matching the component-file pattern used by `utxo_manager_test.go` and `tx_creator_test.go`. The four split account-manager test files were removed.

wallet/internal/db/kvdb/accountstore_renameaccount.go
GustavoStingelin: Added inline rationale at the branch: public `AccountNumber` input represents a BIP44 account and therefore rejects waddrmgr-internal imported-account numbers, while name lookup may resolve imported XPub accounts and rejects only the reserved raw-import pseudo-account. The extracted guard also documents that imported rows have no public BIP44 account number.

wallet/internal/db/kvdb/accountstore_createderived.go
GustavoStingelin: Reverted the `AccountID` scope from this PR: the created, queried, and reloaded assertions are gone, and KVDB no longer includes `AccountID` in its `CreateDerivedAccountRow` result. The public account-mutation tests contain no backend routing identity; any internal row invariant remains with its owning task.

GustavoStingelin:
Disclosure: I am the AI coding agent working on behalf of @GustavoStingelin on this PR.

I reviewed the current #1326 scope specifically for duplicated roadmap ownership, rather than textual merge conflicts:

- Task 350 / #1321 owns the native pgx migration; #1326 contains only narrow account-store behavior fixes required by the Task 429 mutation tests.
- Task 368 / #1328 owns production wallet lifecycle serialization; the #1326 changes are private `bwtest` fixture and reload support.
- Task 453 / #1336 owns concurrent Manager assembly serialization; #1326 does not implement Manager locking or publication.
- Task 379 / #1331 owns `Manager.Load` not-found translation, and Task 337 / #1324 owns imported-XPub signer derivation; neither behavior is duplicated here.
- The earlier `AccountID` work overlapped Task 377 and was out of scope, so both its assertions and KVDB return change were removed.
- Task 430 query coverage remains explicitly outside #1326.

I found no remaining #1326 behavior proven to duplicate Tasks 298 or 360. The current scope is Task 429 mutation integration coverage plus the minimum supporting fixes exposed by those tests.

wallet/internal/db/kvdb/accountstore_renameaccount.go
```diff
@@ -42,13 +43,29 @@ func (s *Store) RenameAccount(_ context.Context,
...
+				return err
+			}
+		} else if acctNum == waddrmgr.ImportedAddrAccount {
+			// Name lookups may select imported xpub accounts.
+			// Only the reserved raw-import pseudo-account remains
+			// unrenameable.
+			return db.ErrAccountNotFound
 		}
 
 		err = scopedMgr.RenameAccount(ns, acctNum, params.NewName)
```
yyforyongyu, line 59: `RenameAccount(oldName == newName)` appears to expose a backend divergence: KVDB returns a duplicate error while SQL reports success. The current Task 429 itests do not cover this case and this PR does not fix it, so let's not expand PR 1326. Please track and fix this in a separate focused bug-fix PR after 1326.

itest/account_manager_test.go
```diff
@@ -0,0 +1,826 @@
...
+	)
+	require.Equal(
+		h, waddrmgr.WitnessPubKey, created.AddrSchema.InternalAddrType,
+	)
+	normalizedPublicKey := canonicalAccountKey(h, created.PublicKey)
+	require.NotEmpty(
+		h, normalizedPublicKey, "derived account has no public key",
+	)
+	require.NotNil(
+		h, created.MasterKeyFingerprint,
```
yyforyongyu, line 142: Task 429 explicitly requires absent, present-zero, and present-nonzero optional identities, but the current suite only exercises `AccountNumber == nil` for imported accounts and non-zero values for derived account numbers and fingerprints. Could we add the two non-nil zero cases without expanding behavior: rename the default account and assert `AccountNumber != nil && *AccountNumber == 0` through `GetAccount` and reload; import (or preview) with fingerprint zero and assert `MasterKeyFingerprint != nil && *MasterKeyFingerprint == 0` through the relevant postcondition and reload. This catches the zero-to-nil collapsing that Task 341's contract is meant to prevent.

itest/account_manager_test.go
```diff
@@ -0,0 +1,826 @@
...
+	require.NoError(h, w.Lock(ctx), "failed to lock wallet")
+
+	_, err := w.NewAccount(ctx, scope, lockedName)
+
+	require.Error(h, err, "locked wallet created an account")
+	require.NoError(h, w.Stop(ctx), "failed to stop wallet")
+
+	_, err = w.NewAccount(ctx, scope, stoppedName)
+
+	require.ErrorIs(h, err, wallet.ErrStateForbidden)
```
yyforyongyu, line 334: These lifecycle rejection cases stop at the error assertion. Task 429's definition of done requires every rejection to leave public state unchanged, so a call that partially writes and then returns the expected error could still pass. After locked or stopped `NewAccount`, could we unlock or reload and assert both requested names are absent and the account set is unchanged? Likewise, after stopped `RenameAccount`, reload and assert the source is unchanged and target absent; after stopped `ImportAccount`, reload and assert the target is absent and the account set unchanged. The watch-only case already performs the narrow absence check.

itest/account_manager_test.go
```diff
@@ -0,0 +1,826 @@
...
+			accountName: "",
+		},
+	}
+
+	for _, tc := range testCases {
+		_, err := w.NewAccount(ctx, scope, tc.accountName)
+
+		require.Error(h, err, "%s was accepted", tc.name)
+
+		if tc.wantSentinel != nil {
```
yyforyongyu: These cases show that PR 1326 does not establish stable public identities for every `AccountManager` rejection. That complete public error boundary is owned by Task 378. PR 1326 currently implements only part of Task 378—the duplicate-account and rename-not-found mappings—while retaining internal/legacy error exposure. Since we want to merge Task 429 before doing Task 378, please remove or extract the partial Task 378 production changes from this PR and keep these negative tests focused on rejection and unchanged state for now. We can tighten them to `ErrorIs`/`ErrorAs` in the dedicated Task 378 PR.

wallet/account_manager.go
```diff
@@ -240,6 +242,23 @@ func (w *Wallet) accountInfoFromStore(
...
 }
 
+// mapDuplicateAccountErr converts a Store account-name conflict into the stable
+// public duplicate-account identity shared by NewAccount, RenameAccount, and
+// ImportAccount, adding the rejected name as selector context.
+//
+// The result satisfies errors.Is(err, ErrDuplicateAccount) on every maintained
+// backend. It is still returned as a top-level waddrmgr.ManagerError because
+// waddrmgr.IsError type asserts instead of unwrapping, so existing callers of
+// that check keep working; Task 378 can drop the legacy envelope without
```
yyforyongyu: we should not mention local context in the code as it will be confusing for future devs

itest/account_manager_test.go
```diff
@@ -0,0 +1,346 @@
...
+	)
+
+	scope := waddrmgr.KeyScopeBIP0084
+	ctx := h.Context()
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+
+	_, err := w.NewAccount(ctx, scope, sourceName)
+	require.NoError(h, err, "failed to create rejection source")
+
+	w = h.ReloadWallet(w)
```
yyforyongyu: why do we need to `ReloadWallet` here?

itest/account_manager_test.go
```diff
@@ -344,3 +344,263 @@ func testAccountManagerRejectWatchOnlyAccountCreation(h *bwtest.HarnessTest) {
...
+		renamedName = "account manager derived renamed"
+	)
+
+	scope := waddrmgr.KeyScopeBIP0084
+	ctx := h.Context()
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	_, err := w.NewAccount(ctx, scope, sourceName)
+	require.NoError(h, err, "failed to create derived rename source")
+
+	w = h.ReloadWallet(w)
```
yyforyongyu: same question here - why do we need to `ReloadWallet` after `NewAccount`? I think the durability of the `NewAccount` is already tested in `testAccountManagerCreateAccount`.

wallet/internal/db/kvdb/accountstore_createimported.go
```diff
@@ -105,7 +105,17 @@ func (s *Store) putImportedAccount(ns walletdb.ReadWriteBucket,
 	pubKey *hdkeychain.ExtendedKey,
 	walletIsWatchOnly bool) (*db.AccountInfo, error) {
 
-	addrSchema := waddrmgrScopeAddrSchema(params.AddrSchema)
+	var addrSchema *waddrmgr.ScopeAddrSchema
```
yyforyongyu, line 108: Let's document in the commit body that this fixes a pre-existing `sql-wallet` imported-account schema mismatch already identified by Task 294: the `waddrmgr` schema ordinal was cast directly to the database enum even though the two enums do not share all numeric values. We'll keep this fix here as a one-time exception to avoid more stack churn, but please also note in the PR description that future production bugs discovered during itest work should be handled in separate focused PRs.

yyforyongyu (changes requested):
Much easier to follow! I think once these comments are addressed we should be good to go.

wallet/internal/db/kvdb/accountstore_renameaccount.go
GustavoStingelin: Agreed, not expanding this PR. `RenameAccount(oldName == newName)` is neither covered nor fixed here; I'll track it and open a separate focused bug-fix PR after 1326 merges.

itest/account_manager_test.go
GustavoStingelin: Added, as two new registered cases.

- `account manager rename default account` asserts `AccountNumber != nil && *AccountNumber == 0` at the source, after the rename through `GetAccount`, and again after reload.
- `account manager import account zero fingerprint` imports with fingerprint `0` and asserts `MasterKeyFingerprint != nil && *MasterKeyFingerprint == 0` on the mutation result, the `GetAccount` postcondition, and after reload. The wallet is seeded with a second imported account carrying a non-zero fingerprint, so present-zero is distinguishable from present-nonzero *within the same wallet* rather than from a backend-wide default.

One adjustment was needed to make the first one portable. A fresh SQL wallet has no accounts at all, while kvdb seeds a `default` per scope at genesis, so `GetAccount(scope, "default")` fails outright on SQL:

```
kvdb   fresh:  default @ 44/0=0, 49/0=0, 84/0=0, 86/0=0
sqlite fresh:  (no accounts)
```

The case therefore funds the wallet and derives the scope with `KeyScope()`, the same funded-scope authority `tx_creator_test.go` uses. That is the only scope guaranteed to hold account number 0 on all three backends — kvdb pre-seeds it, SQL creates it lazily on first use. No backend branch.

To confirm the cases actually catch the collapse, I mutated `accountInfoFromStore` to turn a present zero into `nil` for both identities and verified each case fails for the right reason. Mutation reverted.

itest/account_manager_test.go
GustavoStingelin: Added to all three lifecycle cases. Each captures the account set while the wallet is still started, then reopens after the rejection:

- **creation** — neither `lockedName` nor `stoppedName` resolves, account count unchanged;
- **rename** — target absent, source still compares equal to its complete expected result;
- **import** — target absent, account count unchanged.

I used reload rather than unlock everywhere, since it proves nothing reached disk rather than only that the in-memory view looks right.

The baselines use `ListAccounts` rather than `ListAccountsByScope`: no account is ever created successfully in these cases, so on SQL the key scope may not exist and a scope-filtered query would not be safe there.

Verified by injecting a leftover account before the rejections — the previous version **passed** with the row present, the new one fails with `locked rejection created an account`. That is exactly the hole you described.

itest/account_manager_test.go
GustavoStingelin: Done — six commits removed from this PR:

```
db: define duplicate account error
db: normalize SQL duplicate account errors
kvdb: normalize duplicate account errors
wallet: map duplicate account errors
wallet: map rename account not-found error
wallet+db: normalize duplicate rename account errors
```

`git log -S` confirms nothing on the branch now introduces `ErrDuplicateAccount`, `MapAccountNameConflictErr`, `mapDuplicateAccountErr` or `RenameConflictErr`. `wallet/account_manager.go`, `wallet/wallet.go`, the pg/sqlite stores and the SQL queries are untouched by this PR. Only two production commits remain, neither of them error-boundary work: the imported-schema conversion and the kvdb imported-XPub rename.

The removed commits are preserved on a stacked branch ready for the Task 378 PR, along with a commit that tightens the negative cases back to `errors.Is`/`errors.As`.

The rejection tables here now assert rejection and unchanged state only — that includes dropping the `waddrmgr.ManagerError` / `ErrorCode` assertions, which were the internal/legacy exposure.

**One thing worth your call:** I kept `wallet.ErrInvalidAccountKey` on the import nil-key / private-key rows, and `wallet.ErrStateForbidden` in the lifecycle cases. Both sentinels already exist on `sql-wallet` and are already returned there — they are input validation and lifecycle guards rather than backend error translation, so they do not depend on anything this PR does and Task 378 will not change them. Happy to strip those too if you would rather the negative tests carry no identity assertions at all.

wallet/account_manager.go
GustavoStingelin: Removed. That particular comment went away with the function when the Task 378 mappings came out of this PR, but the same practice was in three other places I had written, so those are gone too: two `Task 378` notes in the rejection tables and a `Task 388` note in the import table here, plus a `Task 343` note on the stacked branch.

Each keeps its technical content and loses only the tracker number — for example `// These rejections have no stable public error identity yet, so the rows assert rejection and unchanged state only.` rather than naming the task.

`git grep -iE 'task [0-9]{3}'` is clean across the code on both branches.

itest/account_manager_test.go
GustavoStingelin: Removed — you're right, it earned nothing. Its only effect was locking the wallet, which the very next statement undid:

```go
w = h.ReloadWallet(w)   // returns a locked wallet
...
h.UnlockWallet(w)       // undo it again
```

Both are gone, and the baseline snapshot now comes from `GetAccount` on the live wallet, which is the same public view.

itest/account_manager_test.go
GustavoStingelin: Removed, agreed — `testAccountManagerCreateAccount` already proves `NewAccount` durability, so re-proving it here was redundant setup. The **trailing** reload stays, since proving the rename is durable is what this case exists for.

One side effect worth naming: the rename previously ran on a locked wallet purely as an accident of that reload. Locked-rename admission is still covered deliberately, by `testAccountManagerEnforceAccountRenameLifecycle`.

That lifecycle case also reloads right after `NewAccount` and would draw the same question, but there the reload *is* the mechanism for reaching the locked state. Its explanation sat ten lines below, next to the rename; I moved it to the reload itself and made it explicit about what it is not doing:

```go
// Reload to reach the started but locked state this rename must be
// admitted in, not to re-check that the source persisted.
```

wallet/internal/db/kvdb/accountstore_createimported.go
GustavoStingelin: Documented in the commit body, and the PR description now carries the note that production bugs found during future itest work belong in separate focused PRs rather than the test PR that surfaced them.

Two adjustments while writing it:

**Direction.** Task 294 names the `waddrmgr`→db cast, but that direction is already explicit at the `sql-wallet` base through `db.ScopeAddrSchemaFromWaddrmgr`. What actually survived is the db→`waddrmgr` cast in kvdb's imported-account path — same enum pair, same root cause, opposite direction — so the body describes that one.

**Reference.** Per your other comment about local context, I left the task number out: the roadmap repo isn't public and I found no btcwallet issue describing the mismatch, so `Task 294` would be unresolvable to a future reader. The body cites #1250 instead, which explains why a production fix rides in a test PR, and states the facts directly — pre-exists the `sql-wallet` base, no introducing commit, with the concrete `waddrmgr.PubKeyHash=0` vs `db.RawPubKey=0` example.

GustavoStingelin:
Short churn timeline
- Aug 14 — Narrow, nearly ready  
PR opens for Task 429 mutation integration tests. First review: “Looking good!”
- Aug 14–18 — Scope expands  
Reviews add harness changes, AccountID, stable error mapping, schema fixes, and broader assertions. The test PR starts changing production code.
- Aug 19 — Scope drift noticed  
Reviewer asks to check roadmap ownership and remove unrelated fixes. Tests are also requested to be split into smaller cases.
- Aug 19–21 — Major rewrite  
Four test files become one file with 12 cases. ManagedWallet is introduced, then removed. AccountID is reverted—the reviewer acknowledges the original request was mistaken.
- Aug 21–23 — Scope expands again  
Six error-normalization commits are added, overlapping Task 378.
- Aug 24 — Scope reduced again  
The six Task 378 commits are extracted. Redundant reloads and assertions are removed; missing Task 429 cases are added.

GustavoStingelin:
Review agents should read the linked roadmap task first and only request changes required by its acceptance criteria or needed to fix a regression introduced by the PR. 
Valid concerns owned by another task could still be mentioned, but clearly marked as non-blocking follow-ups and linked to the owning task.

Likewise, authors' agents should treat review comments as proposals rather than automatically implementing them. Before making a requested change, they should check the roadmap and confirm that it belongs in the current PR. This should help prevent future PRs from absorbing adjacent tasks and later having to revert or extract that work.

itest/account_manager_test.go
```diff
@@ -0,0 +1,1001 @@
...
+	wantTarget := *target
+	wantTarget.PublicKey = canonicalAccountKey(h, wantTarget.PublicKey)
+
+	// These rejections have no stable public error identity yet, so the
+	// rows assert rejection and unchanged state only.
+	testCases := []struct {
+		name             string
+		oldName          string
+		newName          string
+		wantTargetAbsent bool
```
yyforyongyu: Could we remove `wantTargetAbsent` and give every row the same assertion path? Capture the complete public account set with `ListAccounts` before the table, then after each rejected `RenameAccount` call compare it with `require.ElementsMatch`. That makes the table uniformly data-only and also proves the reserved source was not altered without querying backend-specific pseudo-account details.

itest/account_manager_test.go
```diff
@@ -0,0 +1,1001 @@
...
+	wantExisting := *existing
+	accounts, err := w.ListAccountsByScope(ctx, keys.scope)
+	require.NoError(h, err, "failed to list imported account scope")
+
+	wantCount := len(accounts)
+	testCases := []struct {
+		name         string
+		accountName  string
+		accountKey   *hdkeychain.ExtendedKey
+		wantSentinel error
```
yyforyongyu: Could we split this into two registered cases so the table rows no longer select different assertions? Keep duplicate-name and empty-name rows in `account manager reject account import`, uniformly asserting `require.Error` plus an unchanged complete account set; move the nil/private-key rows to `account manager reject invalid import key`, uniformly asserting `require.ErrorIs(..., wallet.ErrInvalidAccountKey)` plus the same unchanged-state check. This removes `wantSentinel` and its conditional without changing production behavior.

itest/account_manager_test.go
yyforyongyu: One concrete shape would be:

```go
wantAccounts, err := w.ListAccounts(ctx)
require.NoError(h, err, "failed to list accounts before rejections")

testCases := []struct {
    name    string
    oldName string
    newName string
}{
    // Keep the existing rows, without wantTargetAbsent.
}

for _, tc := range testCases {
    err := w.RenameAccount(ctx, scope, tc.oldName, tc.newName)
    require.Error(h, err, "%s was accepted", tc.name)

    gotAccounts, err := w.ListAccounts(ctx)
    require.NoError(
        h, err, "failed to list accounts after %s", tc.name,
    )
    require.ElementsMatch(
        h, wantAccounts, gotAccounts,
        "%s changed the account set", tc.name,
    )
}
```

The complete before/after comparison replaces `wantTargetAbsent` and the
per-row branch while checking that no source, target, or reserved account
changed.

itest/account_manager_test.go
yyforyongyu: One concrete split would be:

```go
// testAccountManagerRejectAccountImport verifies name-based import
// rejections leave the complete account set unchanged.
func testAccountManagerRejectAccountImport(h *bwtest.HarnessTest) {
    // Keep the current deterministic keys and seeded-wallet setup.

    wantAccounts, err := w.ListAccounts(ctx)
    require.NoError(h, err, "failed to list accounts before rejections")

    testCases := []struct {
        name        string
        accountName string
        accountKey  *hdkeychain.ExtendedKey
    }{
        {
            name:        "duplicate name, same key",
            accountName: existingName,
            accountKey:  keys.accountKey,
        },
        {
            name:        "duplicate name, other key",
            accountName: existingName,
            accountKey:  keys.otherAccountKey,
        },
        {
            name:        "empty name",
            accountName: "",
            accountKey:  keys.otherAccountKey,
        },
    }

    for _, tc := range testCases {
        _, err := w.ImportAccount(
            ctx, tc.accountName, tc.accountKey,
            keys.masterKeyFingerprint, keys.addrType, false,
        )
        require.Error(h, err, "%s was accepted", tc.name)

        gotAccounts, err := w.ListAccounts(ctx)
        require.NoError(h, err, "failed to list accounts after rejection")
        require.ElementsMatch(h, wantAccounts, gotAccounts)
    }
}

// testAccountManagerRejectInvalidImportKey verifies invalid key material is
// rejected with the public sentinel and leaves the account set unchanged.
func testAccountManagerRejectInvalidImportKey(h *bwtest.HarnessTest) {
    // Create an independent wallet, then capture wantAccounts as above.

    testCases := []struct {
        name       string
        accountKey *hdkeychain.ExtendedKey
    }{
        {
            name:       "nil key",
            accountKey: nil,
        },
        {
            name:       "private key",
            accountKey: keys.accountPrivateKey,
        },
    }

    for _, tc := range testCases {
        _, err := w.ImportAccount(
            ctx, "account manager invalid key", tc.accountKey,
            keys.masterKeyFingerprint, keys.addrType, false,
        )
        require.ErrorIs(h, err, wallet.ErrInvalidAccountKey)

        gotAccounts, err := w.ListAccounts(ctx)
        require.NoError(h, err, "failed to list accounts after rejection")
        require.ElementsMatch(h, wantAccounts, gotAccounts)
    }
}

// Register the second contract independently in list_on_test.go.
{
    Name:     "account manager reject invalid import key",
    TestFunc: testAccountManagerRejectInvalidImportKey,
},
```

Each registered case now has one uniform assertion contract, so
`wantSentinel` and its conditional disappear.

yyforyongyu (changes requested):
Cool two more comments then i think it should be good to go!

itest/account_manager_test.go
GustavoStingelin: cool! done

itest/account_manager_test.go
GustavoStingelin: cool! done

bwtest/harness.go
```diff
@@ -330,26 +349,51 @@ func validateFileBackendArtifact(dbType, dbPath string) error {
...
+
 	h.Helper()
 
-	if w == nil {
-		h.Fatalf("cannot register nil wallet")
-	}
+	h.registerWallet(manager, w, nil)
+}
+
+// registerWallet records w under manager and copies reloadConfig when provided.
```
Abdulkbk: why do we shorten wallet to `w`?

itest/account_manager_test.go
```diff
@@ -0,0 +1,1005 @@
...
+
+		gotAccounts, err := w.ListAccounts(ctx)
+		require.NoError(h, err, "failed to list accounts after rejection")
+		require.ElementsMatch(
+			h, wantAccounts, gotAccounts, "%s changed the account set", tc.name,
+		)
+	}
+}
+
+func testAccountManagerRejectInvalidImportKey(h *bwtest.HarnessTest) {
```
yyforyongyu, line 920: missing godoc

itest/account_manager_test.go
```diff
@@ -0,0 +1,1005 @@
...
+	require.Zero(
+		h, *durable.MasterKeyFingerprint,
+		"fingerprint is not zero after reload",
+	)
+	require.Equal(h, want, *durable)
+}
+
+// testAccountManagerPreviewAccountImport verifies a dry run returns portable
+// identity fields without materializing an account.
+func testAccountManagerPreviewAccountImport(h *bwtest.HarnessTest) {
```
yyforyongyu, line 811: This test actually exposes an issue that we don't have a solution yet, and is tracked here: https://github.com/yyforyongyu/btcwallet-sql-roadmap/issues/17

I don't think it should block this PR tho.

yyforyongyu (approved):
LGTM!

itest/account_manager_test.go
Abdulkbk: +1

bwtest/harness.go
```diff
@@ -330,26 +349,51 @@ func validateFileBackendArtifact(dbType, dbPath string) error {
 	return nil
 }
 
-// RegisterWallet registers a wallet with the harness.
+// RegisterWallet registers w with its harness-owned manager.
```
Abdulkbk: nit: `registers wallet with...`

Abdulkbk (changes requested):
just the godoc yy pointed out and 2 nits

itest/account_manager_test.go
GustavoStingelin: added

bwtest/harness.go
GustavoStingelin: Good point, updated the comments to avoid leaning on the abbreviated parameter name.

I kept the parameter as `w` because short names are not uncommon in Go when the type/context is local and obvious, like `w io.Writer` or `ctx context.Context`. In this specific case, `wallet` would also conflict with the imported `wallet` package.

bwtest/harness.go
GustavoStingelin: updated the comment

Abdulkbk (approved):
lgtm

GustavoStingelin:
thanks guys

PR #1328, opened 2026-08-15
wallet: make Manager the lifecycle authority

yyforyongyu:
## Summary

Make Manager the sole public lifecycle owner for its managed Wallets and
owned database. Startup publishes the complete durable Wallet set, creation
returns a started Wallet, and shutdown drains owned work before closing storage.

## Change Description

- `Manager.Start(ctx)` uses the existing backend `listWallets` mechanism to
  assemble and start every durable Wallet in stable Wallet-ID order. An empty
  database returns an empty slice. A failed or canceled startup publishes no
  Wallet handles and completes cleanup before returning.
- `Manager.Create(params)` is admitted only while the Manager is running. It
  persists, initializes, and starts the candidate before publishing its handle.
  Accepted creation completes before Stop can close shared storage. Caller
  context support remains deferred until wallet creation uses one atomic
  database transaction; the Create signature is unchanged.
- `Manager.Stop()` waits for startup, accepted creation, Wallet operations,
  and Wallet workers to settle, then closes the owned database. It also closes
  constructor-owned resources when called before Start.
- Shutdown and admitted startup failure are terminal. Reopening durable state
  requires a fresh Manager and its Start result; old Wallet handles remain
  stopped.
- Remove the superseded public Wallet/Controller lifecycle methods and Manager
  Load/Close entry points. Migrate maintained callers, harness ownership,
  benchmarks, and tests to Manager Start/Create/Stop. Legacy deprecated
  methods retain their behavior.
- Reuse the existing Manager mutex, Wallet registry, backend discovery, and
  private Wallet shutdown, with one atomic startup guard. No lifecycle worker,
  detached operation, cached result, or restart API is introduced. The modern
  kvdb backend retains its single-Wallet constraint.

The implemented contract deliberately simplifies the original lifecycle
proposal:

| Operation | Original proposal | Implemented behavior |
| --- | --- | --- |
| Start during startup or while running | Join the attempt or return the same Wallet pointers | Reject with `ErrStateForbidden`; no joining or result replay |
| Start after admitted startup failure | Replay the startup error until Stop | Failure drains candidates and closes storage immediately; a later eligible Start returns `ErrManagerStopped` |
| Stop racing startup | Cancel Manager-owned startup at a shutdown cutoff | Wait for the accepted synchronous Start to settle, then drain its published Wallets if it succeeded |
| Context cancellation | Caller can stop waiting while Manager-owned work continues | Start uses its caller context and completes failure cleanup before returning; Stop has no context parameter and waits for shutdown to finish |
| Concurrent or repeated Stop | Join or replay one recorded aggregate shutdown result | Calls serialize; the call performing cleanup returns its errors, and subsequent calls return nil |
| Shared chain source | Manager owns shared chain lifecycle | The supplied ChainSource remains caller-owned; Manager does not start or stop it |

An eligible Start with an already canceled context returns `ctx.Err()` without
admitting startup. The active-start guard is checked first, so a duplicate Start
still rejects while starting, running, or stopping. There is no Stop-then-Start
restart on the same Manager and no separate TearDown method.

**Create failure boundary:** publication is all-or-nothing, but durable creation
and runtime startup are not one transaction. Account-import or startup failure
can occur after persistence. In that case Create returns no Wallet, drains the
candidate and the existing active set, closes the backend, and returns the
primary and cleanup errors. A later Stop is a no-op. Earlier validation failures
leave the Manager usable. This PR adds no rollback, automatic retry, duplicate
reconciliation, or guarantee that an error means no durable Wallet exists.
Reopening also does not automatically finish partially imported initial
accounts. Exact creation reconciliation remains separate work.

Validation covers ordered/empty startup, startup failure and cancellation
cleanup, duplicate Start rejection, terminal shutdown and blocking cleanup,
Create/Stop serialization, and post-persistence startup failure. Package and
focused race checks pass, as does the Manager integration matrix across kvdb,
SQLite, and PostgreSQL with the supported chain backends. See the PR checks
for the current CI result.

## Notes

Builds on #1356, now merged into `sql-wallet`, which remains this PR's base.
Review scope begins after that prerequisite: backend discovery and assembly
provided by #1356 are reused here, not introduced by this PR.


wallet/controller.go
```diff
@@ -177,39 +177,57 @@ type Controller interface {
 //
 // This is part of the Controller interface.
 func (w *Wallet) Start(startCtx context.Context) error {
-	// 1. Attempt to transition from Stopped to Starting.
+	w.lifecycleMu.Lock()
```
yyforyongyu: This design is ugly and hard to follow - instead, we should take a step back and think about how each component handles its own internal state and boundary. For the wallet, there is no need to support restart, instead, its lifecycle should be managed by a higher component like the manager.

wallet/manager_test.go
```diff
@@ -33,8 +35,6 @@ func TestManagerBuildsWalletsFromRuntimePolicy(t *testing.T) {
 
 	// Arrange: Configure two exact Store creations and a distinctive Manager
 	// snapshot shared by both identity-only creation requests.
-	const secondWalletName = "second"
```
yyforyongyu, line 36: unnecessary change?

wallet/manager.go
```diff
@@ -30,6 +34,31 @@ var (
 	ErrDatabaseIdentityMismatch = errors.New("database identity mismatch")
 )
 
+// managerState identifies the one terminal lifecycle owned by Manager.
+type managerState uint8
```
yyforyongyu: why do we need these many different states?

wallet/manager.go
```diff
@@ -137,6 +164,21 @@ type LoadWalletParams struct {
 type Manager struct {
 	sync.RWMutex
 
+	// state is the aggregate lifecycle state guarded by the Manager lock.
+	state managerState
+
+	// startDone publishes the fixed ordered pointers and primary error only
+	// after successful startup or complete candidate cleanup.
+	startDone      chan struct{}
+	startedWallets []*Wallet
```
yyforyongyu: manager already has `wallets`, so why creating another one?

wallet/manager.go
```diff
@@ -137,6 +164,21 @@ type LoadWalletParams struct {
...
+	startErr       error
+	startCancel    context.CancelFunc
```
yyforyongyu: i don't think we should track all these start/stop errors or context? the errors should just be returned at the callsite and the context should be supplied by the caller?

wallet/manager.go
```diff
@@ -137,6 +164,21 @@ type LoadWalletParams struct {
 type Manager struct {
 	sync.RWMutex
 
+	// state is the aggregate lifecycle state guarded by the Manager lock.
+	state managerState
```
yyforyongyu: the idea is that we only assume one callsite calling `Start/Stop`, so all we need is to track whether it's started or not?

wallet/manager.go
```diff
@@ -187,6 +196,25 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
...
+	return &Manager{
+		wallets: make(map[string]*Wallet),
+		backend: backend,
+		config:  cfg,
+	}, nil
+}
+
+// newManagerBackend opens the configured database for construction or a Start
+// after Stop, preserving the same identity checks for either entry point.
+func newManagerBackend(ctx context.Context,
```
yyforyongyu: why do we introduce this method?

wallet/manager.go
```diff
@@ -205,19 +233,75 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
...
-	return &Manager{
-		wallets: make(map[string]*Wallet),
-		backend: backend,
-		config:  cfg,
-	}, nil
+		wallet := newManagedWallet(cfg, walletData)
+		wallets = append(wallets, wallet)
+
+		err = wallet.Start(ctx)
+		if err != nil {
```
yyforyongyu, line 229: if there is an error starting the wallet, we should stop the other wallets or?

wallet/manager.go
```diff
@@ -205,19 +233,75 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
...
+			return wallets, err
+		}
+	}
+
+	return wallets, nil
+}
+
+// stopLocked drains the owned Wallets and closes storage last. The caller
+// holds the Manager write lock until the guard can safely admit another Start.
+func (m *Manager) stopLocked() error {
```
yyforyongyu: we can just combine this with `stopWallet` and call it `stopWallets`? there is no locking behavior so this name will be misleading

wallet/common_test.go
```diff
@@ -128,33 +151,45 @@ func testSQLiteManager(tb testing.TB) *Manager {
...
 			WalletSyncRetryInterval: initialBackoff,
 			AutoLockDuration:        defaultLockDuration,
 		},
 	}
+
+	m.started.Store(true)
+
+	// The fixture owns required-call verification. Stop joins workers before
+	// assertions run, while the caller retains ownership of the mock Store.
+	tb.Cleanup(func() { store.AssertExpectations(tb) })
```
yyforyongyu: use multiple pls, for all `Cleanup` methods introduced in this PR

wallet/manager.go
```diff
@@ -182,6 +180,7 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
 	}
 
 	cfg.ChainParams = chainParams
+	cfg.SignetChallengeDigest = slices.Clone(cfg.SignetChallengeDigest)
```
yyforyongyu: this is unrelated? also it should be put in its own commit even if we wanna fix it? for scope wise concern, remove it

wallet/manager.go
```diff
@@ -346,6 +345,76 @@ func (m *Manager) Close() error {
...
 
+// Start starts every durable Wallet and returns the complete active set in
+// stable Wallet-ID order. Another Start while starting, running, or stopping
+// returns ErrStateForbidden. A caller may Start again after Stop completes.
+// Cancellation or failure cleans up all candidates before returning an error.
+func (m *Manager) Start(ctx context.Context) ([]*Wallet, error) {
+	// Reject without waiting for the lock held by startup or shutdown. Claim
+	// under that lock below so Stop cannot overtake an admitted startup.
+	if m.started.Load() {
+		return nil, ErrStateForbidden
```
yyforyongyu: use `fmt.Errorf` to wrap the error and mention that `already started`?

wallet/manager.go
```diff
@@ -346,6 +345,76 @@ func (m *Manager) Close() error {
...
+		return nil, err
+	}
+
+	if m.started.Swap(true) {
+		return nil, ErrStateForbidden
+	}
+
+	// Stop releases the backend, so a later ordinary Start opens it through
+	// the same constructor path and assembles fresh Wallet runtime instances.
+	if m.backend == nil {
```
yyforyongyu, line 343: how come the backend can be nil here?? we should just let it panic if the manager is not even initialized!!

wallet/manager.go
```diff
@@ -346,6 +345,76 @@ func (m *Manager) Close() error {
...
+
+	wallets, err := m.startWallets(ctx)
+	if err == nil {
+		err = ctx.Err()
+	}
+
+	if err != nil {
+		// No candidate is published on failure; join its work before closing
+		// shared storage and permitting another startup attempt.
+		cleanupErr := m.stopWallets(wallets)
```
yyforyongyu: ok i see the wallets are stopped on error so nvm.

wallet/manager.go
```diff
@@ -346,6 +345,76 @@ func (m *Manager) Close() error {
...
+
+		return nil, errors.Join(err, cleanupErr, m.stopLocked())
+	}
+
+	for _, wallet := range wallets {
+		m.wallets[wallet.cfg.Name] = wallet
```
yyforyongyu: i think this can be moved to `startWallets`? nits tho

wallet/controller.go
yyforyongyu: The Manager now owns Start/Stop and backend reopening. Individual Wallet instances have private, terminal lifecycle methods; starting again after Manager.Stop assembles fresh instances.

wallet/manager_test.go
yyforyongyu: Removed the incidental constant hoist. secondWalletName stays local to TestManagerBuildsWalletsFromRuntimePolicy.

wallet/manager.go
yyforyongyu: Removed the lifecycle state enum. Manager admission uses one atomic started flag plus its existing mutex to serialize resource changes.

wallet/manager.go
yyforyongyu: Removed the persistent startedWallets collection. The existing wallets map owns the active set; startup uses only a local slice for ordered results and cleanup of unpublished candidates.

wallet/manager.go
yyforyongyu: Removed retained contexts and historical error results. Start runs synchronously with the caller context and returns errors locally. Stop completes cleanup before returning its local errors.

wallet/manager.go
yyforyongyu: Manager now uses one atomic started flag. Start rejects while it is set; Stop drains all owned work and closes storage before clearing it, allowing a later Start.

wallet/manager.go
yyforyongyu: Kept newManagerBackend because both NewManager and Start after a completed Stop open the configured backend. It reuses the backend selection and database identity checks for those two paths.

wallet/manager.go
yyforyongyu: Yes. A failed Start drains all partial candidates and closes shared storage before returning. The merged stopWallets helper now owns that complete cleanup, including joined cleanup errors.

wallet/manager.go
yyforyongyu: Combined them into stopWallets. It drains unpublished candidates and the owned set, closes storage last, then clears the map and started flag. Its documentation states that callers hold the Manager write lock.

wallet/common_test.go
yyforyongyu: Expanded all 12 Cleanup callbacks introduced or changed by this PR into multiline closures. Callback bodies and cleanup order are unchanged.

wallet/manager.go
yyforyongyu: Kept this one-line copy because Stop closes the backend and a later Start reuses the retained configuration to reopen it. Without the copy, mutating the caller-owned digest changes that database identity and reopening fails. TestManagerRetainsSignetIdentity covers this failure; the copy and regression are colocated with Manager Start/Stop.

wallet/manager.go
yyforyongyu: Both Start rejection paths now return fmt.Errorf("manager already started: %w", ErrStateForbidden). The existing repeated-Start test checks the diagnostic while retaining ErrorIs coverage.

wallet/manager.go
yyforyongyu: The backend is nil after an ordinary completed Stop: Stop closes and clears it. This branch lets a subsequent Start reopen the backend using the retained configuration; it is not support for an uninitialized Manager.

wallet/manager.go
yyforyongyu: Confirmed. Failed startup completes candidate cleanup and storage closure before returning; the merged cleanup helper preserves that behavior.

wallet/manager.go
yyforyongyu: Moved the final cancellation check and complete-set publication into startWallets, after every Wallet starts successfully. Error paths return unpublished candidates for cleanup.

wallet/manager.go
yyforyongyu: Manager.Stop is now terminal, so backend reopening is removed. Backend construction is back in NewManager and the extra newManagerBackend helper is gone.

wallet/manager.go
yyforyongyu: Manager.Stop is now terminal. Removed the Signet digest copy and TestManagerRetainsSignetIdentity because same-Manager reopening is no longer supported. Existing durable-reopen coverage uses a fresh Manager.

wallet/manager.go
yyforyongyu: Start no longer reopens a nil backend. After cleanup closes it, a live-context Start returns ErrManagerStopped; a caller must construct a fresh Manager to reopen the database. Repeated Stop returns immediately without changing admission state.

itest/manager_test.go
```diff
@@ -112,23 +112,23 @@ func testManagerStartConcurrent(h *bwtest.HarnessTest) {
...
+	loaded, err := manager.Start(h.Context())
+	require.NoError(h, err, "failed to start wallet manager")
+	require.Empty(h, loaded, "empty store loaded wallets")
 
+	// Act: Create through the active Manager; this call must start its
+	// runtime before the harness can observe synchronization.
 	w, err := manager.Create(params)
-	require.NoError(h, err, "failed to create wallet")
 
-	// Register before Start so teardown owns the wallet even if Start fails.
```
yyforyongyu, line 96: it seems like we've added lots of unnecessary changes in the docs?

yyforyongyu: Restored concise existing explanations around unchanged Controller, UTXO, and Manager assertions, and removed redundant summaries. Kept the comments needed to explain Manager-owned startup and terminal shutdown. These changes are folded into their respective component commits.

itest/tx_publisher_test.go
```diff
@@ -415,66 +415,50 @@ func testBroadcastRejected(h *bwtest.HarnessTest) {
...
 func testCheckMempoolAcceptanceWalletState(h *bwtest.HarnessTest) {
+	// Arrange: Use a running fixture, then retain its stopped pointer across
+	// reload. The replacement stays registered for harness-owned cleanup.
 	w, _ := h.NewWallet(bwtest.WalletFixture{
-		AddrType:  txPublisherFundingType,
-		Unstarted: true,
+		AddrType: txPublisherFundingType,
 	})
 
+	h.ReloadWallet(w)
```
yyforyongyu: what is the point of restarting the wallet immediately after the creation?

wallet/manager.go
```diff
@@ -254,12 +317,66 @@ func translateDatabaseIdentityError(err error) error {
...
+		return nil, errors.Join(err, m.stopWallets(wallets))
+	}
+
+	return wallets, nil
+}
+
+// Stop terminally drains every Wallet and closes the owned database.
+// It waits for startup, accepted creation, and any other Stop to finish.
+// Cancellation does not shorten this wait. After shutdown, Stop is a no-op.
+func (m *Manager) Stop(_ context.Context) error {
```
yyforyongyu: why passing a context but never uses it? might as well just remove it?

wallet/manager.go
```diff
@@ -287,8 +404,54 @@ func validateManagedWalletName(name string) error {
 	return nil
 }
 
-// Create persists and assembles a Wallet with Manager-owned runtime policy.
+// Create persists, starts, and publishes one Wallet while Manager is running.
 func (m *Manager) Create(params CreateWalletParams) (*Wallet, error) {
+	m.Lock()
```
yyforyongyu, line 412: just do `defer m.Unlock`?

wallet/manager.go
```diff
@@ -306,39 +469,31 @@ func (m *Manager) Create(params CreateWalletParams) (*Wallet, error) {
...
-			context.Background(), params.InitialAccounts,
-		)
-		if err != nil {
-			return nil, err
-		}
-	}
+// failCreateLocked drains a committed but unpublished candidate and the live
+// set before returning the Create and cleanup errors. The caller holds the
+// Manager write lock, preventing storage closure ahead of accepted work.
+func (m *Manager) failCreateLocked(w *Wallet,
```
yyforyongyu: why having this useless shim?

wallet/manager.go
```diff
@@ -287,8 +404,54 @@ func validateManagedWalletName(name string) error {
...
+	if err != nil {
+		m.Unlock()
+
+		return nil, err
+	}
+
+	// If we are in shell mode and have initial accounts, we import them now.
+	if params.Mode == ModeShell && len(params.InitialAccounts) > 0 {
+		err = w.importInitialAccounts(
+			context.Background(), params.InitialAccounts,
```
yyforyongyu, line 434: so why start/stop takes a context but not `Create`?

itest/tx_publisher_test.go
yyforyongyu: The replacement runtime was unnecessary: these assertions use only the stopped original Wallet. The four terminal-pointer cases now retain their Manager and call Stop directly. ReloadWallet remains where its returned Wallet is used to check persistence or reopening.

wallet/manager.go
yyforyongyu: Removed the context parameter from Manager.Stop and updated its callers. Shutdown still waits for admitted work and Wallet workers to finish before closing storage. Start retains its context for synchronous startup work.

wallet/manager.go
yyforyongyu: Create now defers m.Unlock immediately after locking. Every return, including candidate cleanup, remains under the same lock until the operation has settled.

wallet/manager.go
yyforyongyu: Removed failCreateLocked. The two failure paths directly join the creation error with stopWallets cleanup; Create owns the deferred unlock.

wallet/manager.go
yyforyongyu: Kept Create(params) unchanged and added the requested TODO(yy) to introduce a context parameter once wallet creation uses a single atomic database transaction. Caller cancellation during the current multi-stage creation could otherwise return an error after persistence. Start uses its context for opening existing Wallets; Stop now has no context parameter.

wallet/manager.go
```diff
@@ -220,6 +225,44 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
 	}, nil
 }
 
+// stopWallets drains unpublished candidates and owned Wallets before closing
+// storage. Candidates must be disjoint from m.wallets, and the caller holds the
+// Manager write lock until cleanup clears the started guard.
+func (m *Manager) stopWallets(wallets []*Wallet) error {
```
yyforyongyu, line 251: why does this take a slice of wallets when we only use `m.wallets`?

wallet/manager.go
```diff
@@ -220,6 +225,44 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
...
+	for _, wallet := range m.wallets {
+		wallets = append(wallets, wallet)
+	}
+
+	sort.Slice(wallets, func(i, j int) bool {
+		return wallets[i].id < wallets[j].id
```
yyforyongyu, line 258: unnecessary sort?

wallet/manager.go
```diff
@@ -220,6 +225,44 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
...
+			errs = append(errs, fmt.Errorf(
+				"stop wallet %q: %w", wallet.cfg.Name, err,
+			))
+		}
+	}
+
+	// Clear runtime ownership after shutdown has settled. The nil backend
+	// keeps later Start calls from reopening a terminal Manager.
+	err := errors.Join(errs...)
+	if m.backend != nil {
```
yyforyongyu: `m.Stop` already checks this so this is no longer needed?

wallet/manager.go
```diff
@@ -262,6 +305,25 @@ func (m *Manager) Close() error {
...
+// Stop terminally drains every Wallet and closes the owned database.
+// It waits for startup, accepted creation, and any other Stop to finish.
+// After shutdown, Stop is a no-op.
+func (m *Manager) Stop() error {
+	m.Lock()
+	defer m.Unlock()
+
+	// Preserve terminal admission on repeated no-op shutdown calls.
+	if m.backend == nil {
+		return nil
```
yyforyongyu, line 371: add a warning log here?

wallet/manager.go
```diff
@@ -406,8 +404,53 @@ func validateManagedWalletName(name string) error {
...
+	if params.Mode == ModeShell && len(params.InitialAccounts) > 0 {
+		err = w.importInitialAccounts(
+			context.Background(), params.InitialAccounts,
+		)
+		if err != nil {
+			return nil, errors.Join(err, m.stopWallets([]*Wallet{w}))
+		}
+	}
+
+	err = w.Start(context.Background())
```
yyforyongyu: we need to consider attach a run context on the manager and use it to start the wallet - a future task tho, as we need to differentiate wallet creation ctx and wallet runtime ctx

wallet/manager.go
yyforyongyu: Failed candidates stay out of m.wallets. The slice lets cleanup stop those unpublished Wallets before closing storage, including Wallets that started before a later Wallet failed. Added a short comment explaining this.

wallet/manager.go
yyforyongyu: Kept the sort so cleanup errors stay in Wallet-ID order; ranging over m.wallets alone would make that order vary. Added a short comment explaining why.

wallet/manager.go
yyforyongyu: Removed the redundant check. All callers enter this helper with a non-nil backend while holding the Manager lock.

wallet/manager.go
yyforyongyu: Kept this path quiet. Repeated Stop is a supported no-op, and normal fixture cleanup calls it after explicit shutdown. Actual cleanup failures still return to the caller.

wallet/manager.go
yyforyongyu: Agreed, this belongs in future work. A Manager run context should parent each Wallet's lifetime context separately from Create/Start request contexts. Passing it only to wallet.start would affect startup work; Wallet currently creates its runtime lifetime from Background.

wallet/manager.go
```diff
@@ -287,8 +403,53 @@ func validateManagedWalletName(name string) error {
...
+		return nil, err
+	}
+
+	// If we are in shell mode and have initial accounts, we import them now.
+	if params.Mode == ModeShell && len(params.InitialAccounts) > 0 {
+		err = w.importInitialAccounts(
+			context.Background(), params.InitialAccounts,
+		)
+		if err != nil {
+			return nil, errors.Join(err, m.stopWallets([]*Wallet{w}))
```
Abdulkbk, line 436: A wallet key for the wrong network (or an unsupported address type) in InitialAccounts bypasses validation. The import then fails after the wallet row is saved, stopping every other running wallet and shutting down the Manager. We should run the same key checks the import uses in createWallet before the row is saved.

Abdulkbk (changes requested):
One suggestion, otherwise good ot go.

wallet/manager.go
yyforyongyu: Moved the existing key/network and address-type checks before wallet persistence. Invalid initial accounts now return without shutting down the Manager. The regression covers both inputs and verifies the existing Wallet remains usable and corrected input can create the same wallet name.

Abdulkbk (approved):
LGTM

PR #1329, opened 2026-08-15
wallet: make account results wallet-owned

yyforyongyu:
## Summary

Make AccountManager results wallet-owned and publicly implementable.

## Change Description

Add wallet.AccountInfo, preserve optional account and fingerprint presence across SQL and kvdb, copy mutable result fields, and cover backend conversion behavior.

Implements roadmap Task 377

### DIFF
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -10,6 +10,7 @@
 package wallet
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"errors"
@@ -24,6 +25,7 @@ import (
 	"github.com/btcsuite/btcwallet/internal/zero"
 	"github.com/btcsuite/btcwallet/netparams"
 	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet/internal/addresstype"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
 )
 
@@ -115,26 +117,26 @@ type AccountManager interface {
 	// NewAccount creates a new account for a given key scope and name. The
 	// provided name must be unique within that key scope.
 	NewAccount(ctx context.Context, scope waddrmgr.KeyScope, name string) (
-		*db.AccountInfo, error)
+		*AccountInfo, error)
 
 	// ListAccounts returns a list of all accounts managed by the wallet.
-	ListAccounts(ctx context.Context) ([]db.AccountInfo, error)
+	ListAccounts(ctx context.Context) ([]AccountInfo, error)
 
 	// ListAccountsByScope returns a list of all accounts for a given key
 	// scope.
 	ListAccountsByScope(ctx context.Context, scope waddrmgr.KeyScope) (
-		[]db.AccountInfo, error)
+		[]AccountInfo, error)
 
 	// ListAccountsByName searches for accounts with the given name across
 	// all key scopes. Because names are not globally unique, this may
 	// return multiple results.
 	ListAccountsByName(ctx context.Context, name string) (
-		[]db.AccountInfo, error)
+		[]AccountInfo, error)
 
 	// GetAccount returns the snapshot for a specific account, looked up
 	// by its key scope and unique name within that scope.
 	GetAccount(ctx context.Context, scope waddrmgr.KeyScope, name string) (
-		*db.AccountInfo, error)
+		*AccountInfo, error)
 
 	// RenameAccount renames an existing account. To uniquely identify the
 	// account, the key scope must be provided. The new name must be unique
@@ -154,7 +156,7 @@ type AccountManager interface {
 	ImportAccount(ctx context.Context, name string,
 		accountKey *hdkeychain.ExtendedKey,
 		masterKeyFingerprint uint32, addrType waddrmgr.AddressType,
-		dryRun bool) (*db.AccountInfo, error)
+		dryRun bool) (*AccountInfo, error)
 }
 
 // A compile time check to ensure that Wallet implements the interface.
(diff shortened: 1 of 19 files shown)


wallet/internal/db/kvdb/accountstore_getaccount.go
```diff
@@ -207,20 +207,27 @@ func effectiveAddrSchema(scopeSchema waddrmgr.ScopeAddrSchema,
...
+		return nil, fmt.Errorf("read master fingerprint: %w", err)
 	}
 
 	if ok {
-		return persisted, nil
+		return &persisted, nil
 	}
 
-	return props.MasterKeyFingerprint, nil
+	if accountIsImported && props.AccountPubKey != nil {
```
yyforyongyu, line 227: add an inline comment above explaining this check - why do we wanna use the fingerprint here when it's imported?

wallet/internal/db/accountstore_accountinfo.go
```diff
@@ -124,6 +124,23 @@ func optionalAccountNumber(accountNumber sql.NullInt64) (*uint32, error) {
 	return result, nil
 }
 
+// optionalMasterFingerprint converts a nullable SQL master fingerprint to a
+// pointer without collapsing an absent value into a present zero.
+func optionalMasterFingerprint(
```
yyforyongyu, line 129: why creating this method, that ends up returning `nil, nil` that violates the linter?

wallet/internal/db/accountstore_accountinfo.go
```diff
@@ -230,12 +247,9 @@ func AccountPropsRowToInfo[AddrTypeId ~int16 | ~int64](
 		return nil, fmt.Errorf("coin type: %w", err)
 	}
 
-	var fingerprint uint32
-	if row.MasterFingerprint.Valid {
-		fingerprint, err = Int64ToUint32(row.MasterFingerprint.Int64)
-		if err != nil {
-			return nil, fmt.Errorf("master fingerprint: %w", err)
-		}
+	fingerprint, err := optionalMasterFingerprint(row.MasterFingerprint)
```
yyforyongyu, line 251: i don't think this method is necessary?

wallet/account_manager.go
```diff
@@ -237,6 +296,11 @@ func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
 		fingerprint = props.MasterKeyFingerprint
 	}
 
+	var fingerprintResult *uint32
+	if !isImported || props.AccountPubKey != nil {
```
yyforyongyu, line 322: same here - explain the logic - why overwriting the fingerprintResult when it's imported?

wallet/account_manager.go
```diff
@@ -279,28 +343,36 @@ func (w *Wallet) ListAccounts(ctx context.Context) ([]db.AccountInfo, error) {
...
 func (w *Wallet) listAccountInfos(ctx context.Context,
-	query db.ListAccountsQuery) ([]db.AccountInfo, error) {
+	query db.ListAccountsQuery) ([]AccountInfo, error) {
 
 	infos, err := w.cache.ListAccounts(ctx, query)
 	if err != nil {
 		return nil, err
 	}
 
+	if infos == nil {
```
yyforyongyu, line 378: this check is unnecessary? when it's nil, the below `for` loop will be noop?

wallet/account_manager_external_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
+)
+
+// externalAccountManager is an external-package implementation proving every
+// AccountManager method can be named without importing wallet/internal/db.
+type externalAccountManager struct{}
```
yyforyongyu: remove this - i don't see it's testing anything

wallet/internal/db/kvdb/accountstore_getaccount.go
yyforyongyu: Addressed in be77fe517: added an inline explanation that imported XPub rows own the importer-supplied waddrmgr fingerprint, while the keyless imported-address pseudo-account keeps the optional absent.

wallet/internal/db/accountstore_accountinfo.go
yyforyongyu: Addressed in be77fe517: the helper no longer returns a literal nil, nil or needs a nilnil suppression. It now returns an explicit result variable so SQL NULL remains absent. I kept the helper because both SQL row converters use it.

wallet/internal/db/accountstore_accountinfo.go
yyforyongyu: Kept this shared helper intentionally. Both AccountPropsRowToInfo and AccountRowToInfo use it to preserve SQL NULL versus present-zero and apply the same uint32 validation; inlining would duplicate that conversion logic.

wallet/account_manager.go
yyforyongyu: Addressed in be77fe517: added inline comments explaining that imported accounts use their waddrmgr fingerprint instead of the wallet-seed fingerprint, and that AccountPubKey distinguishes an imported XPub from the keyless pseudo-account.

wallet/account_manager.go
yyforyongyu: Kept this check intentionally. Ranging over nil is a no-op, but the following make call would turn a nil Store result into a non-nil empty slice. Task 377 requires preserving nil list results, so removing the check would change observable API behavior.

wallet/account_manager_external_test.go
yyforyongyu: Kept this compile-time test intentionally. The authoritative Task 377 verification requires an external-package implementation of AccountManager using only public packages; this assertion fails to compile if an internal db type leaks back into the interface.

yyforyongyu: Follow-up: removed wallet/account_manager_external_test.go in 8bfb5151d per author direction. The remaining behavior is covered by focused wallet unit tests and the planned SQLite/PostgreSQL database itests; no Task-377-specific daemon E2E was added because this Go interface is not an RPC boundary.

yyforyongyu: Restored in 4819f3a1a per the latest review direction. The external-package compile assertion implements wallet.AccountManager using only public wallet, waddrmgr, hdkeychain, and context types, so any future internal Store-type leak fails compilation.

yyforyongyu:
Addressed the latest review pass in 4819f3a1a:

- derived account DTOs now always use the Wallet-cached canonical root fingerprint, with a stale nonzero Store-value regression fixture;
- restored the external-package AccountManager compile assertion using public types only;
- centralized the BIP32 fingerprint calculation in wallet/internal/db for wallet and kvdb callers;
- corrected the propertiesToAccountInfo boundary documentation; and
- rewrote the signed stack so the persistent canonicalization helper replaces the temporary per-call fallback blocks.

Focused wallet, internal/db, kvdb, and bwtest packages pass, as does fmt-check.

wallet/account_manager_external_test.go
yyforyongyu: Correction: removed wallet/account_manager_external_test.go in cc8c565da, following the authoritative Task 377 specification and the final author direction. Focused in-package converter tests remain the intended coverage.

GustavoStingelin (approved):
LGTM!

PR #1330, opened 2026-08-16
wallet: expose transaction status

yyforyongyu:
## Summary

- Expose wallet-owned transaction validity through the public TxReader API.
- Preserve SQL terminal audit history while documenting the narrower kvdb result.

## Change Description

- Add TxStatus and populate TxDetail.Status through one centralized conversion.
- Table-test pending, published, failed, replaced, and orphaned SQL states.
- Verify kvdb reports only its retained pending and published facts.
- Implements roadmap Task 339

### DIFF
--- a/wallet/tx_reader.go
+++ b/wallet/tx_reader.go
@@ -30,6 +30,11 @@ var (
 	// TxDetailInfo contract and omits the parsed transaction. Callers can
 	// match this error with errors.Is.
 	ErrNilTxDetailMsgTx = errors.New("tx detail MsgTx is nil")
+
+	// ErrInvalidTxStatus is returned when the store supplies a persisted
+	// transaction status that the public wallet API cannot represent. Callers
+	// can match this error with errors.Is.
+	ErrInvalidTxStatus = errors.New("invalid tx status")
 )
 
 // TxReader provides an interface for querying tx history.
@@ -300,12 +305,18 @@ func (w *Wallet) buildTxDetailFromStore(txDetails *db.TxDetailInfo,
 		return nil, ErrNilTxDetailMsgTx
 	}
 
+	status, err := txStatusFromDB(txDetails.Status)
+	if err != nil {
+		return nil, fmt.Errorf("convert tx status: %w", err)
+	}
+
 	msgTx := txDetails.MsgTx
 
 	details := buildBasicTxDetail(
 		txDetails.Hash, txDetails.SerializedTx, txDetails.Label,
 		txDetails.Received, msgTx,
 	)
+	details.Status = status
 
 	w.populateBlockDetails(details, txDetails.Block, currentHeight)
 	w.calculateValueAndFeeFromStore(details, txDetails, msgTx)
@@ -313,6 +324,32 @@ func (w *Wallet) buildTxDetailFromStore(txDetails *db.TxDetailInfo,
 	w.populatePrevOuts(details, msgTx, txDetails.OwnedInputs)
 
 	return details, nil
+}
+
+// txStatusFromDB converts the internal store status into the public wallet
+// status without coupling their enum values.
+func txStatusFromDB(status db.TxStatus) (TxStatus, error) {
+	switch status {
+	case db.TxStatusPending:
+		return TxStatusPending, nil
+
+	case db.TxStatusPublished:
+		return TxStatusPublished, nil
+
+	case db.TxStatusReplaced:
+		return TxStatusReplaced, nil
+
+	case db.TxStatusFailed:
+		return TxStatusFailed, nil
+
+	case db.TxStatusOrphaned:
+		return TxStatusOrphaned, nil
+
+	default:
+		return TxStatusUnknown, fmt.Errorf(
+			"%w: %d", ErrInvalidTxStatus, status,
+		)
+	}
 }
 
 // buildBasicTxDetail builds the common non-wallet-relative fields for one tx
--- a/wallet/tx_reader.go
+++ b/wallet/tx_reader.go
@@ -26,9 +26,10 @@ var (
 	// store.
 	ErrTxNotFound = errors.New("tx not found")
 
-	// errNilTxDetailMsgTx is returned when a store detail row violates the
-	// TxDetailInfo contract and omits the parsed transaction.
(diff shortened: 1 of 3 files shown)


wallet/tx_reader.go
```diff
@@ -29,6 +29,10 @@ var (
 	// errNilTxDetailMsgTx is returned when a store detail row violates the
 	// TxDetailInfo contract and omits the parsed transaction.
 	errNilTxDetailMsgTx = errors.New("tx detail MsgTx is nil")
+
+	// errInvalidTxStatus is returned when the store supplies a transaction
+	// status that the public wallet API does not recognize.
+	errInvalidTxStatus = errors.New("invalid tx status")
```
yyforyongyu: if this is used by public APIs, this error should be exported?

wallet/tx_reader_test.go
```diff
@@ -19,6 +19,23 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+var txReaderStatusCases = []struct {
```
yyforyongyu: explain why we wanna define a global var here in the godoc, as typically we define it in the tests?

wallet/tx_reader_test.go
```diff
@@ -232,6 +298,39 @@ func TestListTxnsSuccess(t *testing.T) {
 	require.Equal(t, expectedTxDetail, details[0])
 }
 
+// TestListTxnsStatus tests that ListTxns preserves every persisted transaction
+// status through the wallet-owned public contract.
+func TestListTxnsStatus(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range txReaderStatusCases {
```
yyforyongyu: actually i think it's better to move the def into the tests so it's easier to follow

wallet/tx_reader_test.go
```diff
@@ -232,6 +300,58 @@ func TestListTxnsSuccess(t *testing.T) {
...
+	t.Parallel()
+
+	// Cover every store status so the public conversion cannot silently drop
+	// a state.
+	testCases := []struct {
+		name        string
+		storeStatus db.TxStatus
+		wantStatus  TxStatus
+	}{
+		{name: "pending", storeStatus: db.TxStatusPending,
```
yyforyongyu: use newlines like this:
```go
{
    name:...
    storeStatus:
    wantStatus:
}, {...
```

and rename `storeStatus` to be `dbStatus`

wallet/tx_reader.go
```diff
@@ -29,6 +29,11 @@ var (
 	// errNilTxDetailMsgTx is returned when a store detail row violates the
 	// TxDetailInfo contract and omits the parsed transaction.
 	errNilTxDetailMsgTx = errors.New("tx detail MsgTx is nil")
```
Abdulkbk: Let's make this exported too `ErrNilTxDetailMsgTx`, like we did with `ErrInvalidTxStatus`

wallet/tx_reader.go
```diff
@@ -88,6 +93,36 @@ type BlockDetails struct {
 	Timestamp int64
 }
 
+// TxStatus describes the wallet-relative validity of a transaction.
+//
+// SQL-backed readers can return every defined status because they retain
+// terminal transaction attempts for audit history. The kvdb backend only
+// returns TxStatusPending or TxStatusPublished because its legacy store does
+// not retain terminal attempts.
+type TxStatus uint8
```
Abdulkbk, line 103: `db.TxStatus` has a `.String()`, let's mirror that here as well

wallet/tx_reader.go
```diff
@@ -88,6 +93,36 @@ type BlockDetails struct {
...
+
+const (
+	// TxStatusPending identifies a locally-authored transaction that the
+	// wallet has not published successfully.
+	TxStatusPending TxStatus = iota
```
Abdulkbk: We should add `TxStatusUnknown` as the first here so that unsupported store statuses cannot silently convert to the public pending status

Abdulkbk (changes requested):
Left some suggestions

wallet/tx_reader.go
yyforyongyu: Addressed in ec2b27b9c: exported ErrNilTxDetailMsgTx with caller-facing documentation and updated all production and regression-test references.

wallet/tx_reader.go
yyforyongyu: Addressed in ec2b27b9c: added TxStatus.String() with the canonical status names and an unknown fallback, plus exhaustive table-driven coverage.

wallet/tx_reader.go
yyforyongyu: Addressed in ec2b27b9c: added TxStatusUnknown as the public zero value. The existing database conversion still fails closed on unsupported store values; this change makes the public enum zero-value semantics explicit.

wallet/tx_reader_test.go
```diff
@@ -19,6 +19,64 @@ import (
...
+
+	testCases := []struct {
+		name   string
+		status TxStatus
+		want   string
+	}{
+		{
+			name:   "unknown",
+			status: TxStatusUnknown,
+			want:   "unknown",
```
yyforyongyu, line 35: use `txStatusUnknownString`?

wallet/tx_reader_test.go
```diff
@@ -19,6 +19,64 @@ import (
...
+			status: TxStatusFailed,
+			want:   "failed",
+		}, {
+			name:   "orphaned",
+			status: TxStatusOrphaned,
+			want:   "orphaned",
+		}, {
+			name:   "unsupported",
+			status: TxStatus(255),
+			want:   "unknown",
```
yyforyongyu, line 59: use `txStatusUnknownString`?

wallet/tx_reader_test.go
yyforyongyu: Addressed in b86e9f2c: the TxStatusUnknown expectation now uses txStatusUnknownString.

wallet/tx_reader_test.go
yyforyongyu: Addressed in b86e9f2c: the unsupported-status expectation now uses txStatusUnknownString.

Abdulkbk (approved):
lgtm

PR #1331, opened 2026-08-16
wallet: translate load not-found errors

yyforyongyu:
## Summary

Implements roadmap Task 379

## Change Description

- Expose wallet-owned `ErrWalletNotFound` from `Manager.Load`.
- Translate SQLite and pristine-kvdb misses without leaking the database
  sentinel.
- Preserve other backend errors and add public-boundary regression coverage.

### DIFF
--- a/itest/manager_test.go
+++ b/itest/manager_test.go
@@ -214,11 +214,19 @@ func testManagerLoadReload(h *bwtest.HarnessTest) {
 // testManagerLoadMissing verifies that loading a wallet that was never created
 // fails rather than silently returning an empty wallet.
 func testManagerLoadMissing(h *bwtest.HarnessTest) {
+	// Arrange a manager whose durable store has never contained the named
+	// wallet from the standard test configuration.
 	cfg, _ := h.TestWalletConfig()
-
 	manager := h.NewWalletManager()
-	_, err := manager.Load(cfg)
-	require.Error(h, err, "load of never-created wallet should fail")
+
+	// Act by asking the manager to load the never-created wallet.
+	w, err := manager.Load(cfg)
+
+	// Assert that the public missing-wallet contract is returned without a
+	// partially assembled wallet.
+	require.ErrorIs(h, err, wallet.ErrWalletNotFound,
+		"load of never-created wallet should report wallet not found")
+	require.Nil(h, w, "load of never-created wallet should return no wallet")
 }
 
 // testManagerCreateWatchOnly verifies that a watch-only wallet is created,
--- a/wallet/internal/db/kvdb/driver_test.go
+++ b/wallet/internal/db/kvdb/driver_test.go
@@ -5,31 +5,69 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/chaincfg/v2"
+	"github.com/btcsuite/btcwallet/waddrmgr"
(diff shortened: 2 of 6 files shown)


wallet/manager_backend.go
```diff
@@ -41,6 +42,16 @@ type walletData struct {
 	isWatchOnly       bool
 }
 
+// managerLoadError translates a backend wallet miss to the public Manager
+// contract while preserving all unrelated backend failures unchanged.
+func managerLoadError(walletName string, err error) error {
```
yyforyongyu: why do we need this thin method helper when this is only used once??

wallet/manager_load_external_test.go
```diff
@@ -0,0 +1,38 @@
+package wallet_test
```
yyforyongyu: dont add an arbitrary test file!

Lrifton92 (commented):
The translation at the Manager boundary is the right call and worth stating explicitly: `db.ErrWalletNotFound` lives under `wallet/internal/db`, so an external caller cannot reference it and `errors.Is` against it is unusable outside the module. Re-exporting a public `wallet.ErrWalletNotFound` and mapping at the edge is what makes the error actionable, and wrapping with `fmt.Errorf("wallet %q: %w", ...)` keeps `errors.Is` working while carrying the name. CI is green across all 19 checks including the kvdb/sqlite/postgres itest matrix.

Splitting "both namespaces missing" from "one namespace missing" is also a real improvement — an absent wallet and a half-written one are genuinely different conditions and used to collapse into `errMissingAddrmgrNamespace`.

### The new doc promise does not hold on kvdb

`Load` now documents:

> If the named wallet does not exist, Load returns ErrWalletNotFound.

But `wallet/internal/db/interface.go` documents the kvdb adapter as not participating in name lookup at all:

```go
// GetWallet retrieves information about a wallet given its name. SQL
// multi-wallet backends return ErrWalletNotFound when the wallet name is
// unknown. The legacy kvdb backend is a single-wallet adapter and echoes
// the requested name without validating it.
```

So on kvdb the new sentinel does not mean "no wallet by that name" — it means "this database contains no wallet at all", which is what the `addrMgrBucket == nil && txMgrBucket == nil` check actually tests. Load a kvdb store that holds wallet `alice` while asking for `bob` and you get `alice` back under the requested name, with no error.

That is presumably intended, since it matches the existing single-wallet adapter contract. The issue is only that `Load`'s comment now states the strong SQL-shaped guarantee without the caveat, while `GetWallet` and `UpdateWallet` both carry it two lines apart in the same interface. A caller writing `errors.Is(err, wallet.ErrWalletNotFound)` to decide whether to create a wallet will behave differently on the two backends, and the doc as written suggests it will not.

Mirroring the existing wording — something like "SQL backends return it for an unknown name; the legacy kvdb backend returns it only when the database holds no wallet" — would make the two contracts agree.

### One question on the empty-database case

A store whose file exists but whose buckets were never written — creation interrupted between opening the database and the first `Update` — now reports `ErrWalletNotFound` rather than a corruption error. I think that is the behaviour you want, since `Create` is the natural recovery from there and the alternative is a wallet that can neither be loaded nor recreated. Worth confirming it is deliberate rather than incidental, because it is the one case where "not found" and "damaged" genuinely overlap.

Abdulkbk (approved):
lgtm

PR #1332, opened 2026-08-16
wallet: translate missing address errors

yyforyongyu:
## Summary

Expose a wallet-owned not-found contract for address-manager lookups.

## Change Description

- Translate normalized Store address misses to `wallet.ErrAddressNotFound` while preserving address context.
- Preserve raw legacy manager errors and private-key lookup behavior.
- Cover translation and negative paths with the existing wallet test mocks.

Implements roadmap Task 380.

### DIFF
--- a/wallet/address_manager.go
+++ b/wallet/address_manager.go
@@ -46,6 +46,10 @@ var (
 	// script.
 	ErrDerivationPathNotFound = errors.New("derivation path not found")
 
+	// ErrAddressNotFound is returned when an address is not known to the
+	// wallet.
+	ErrAddressNotFound = errors.New("address not found")
+
 	// ErrUnknownAddrType is an error returned when a wallet function is
 	// called with an unknown address type.
 	ErrUnknownAddrType = errors.New("unknown address type")
@@ -186,7 +190,8 @@ type AddressManager interface {
 		address.Address, error)
 
 	// GetAddressInfo returns detailed information about a managed address. If
-	// the address is not known to the wallet, an error is returned.
+	// the address is not known to the wallet, the returned error wraps
+	// ErrAddressNotFound for errors.Is matching.
 	GetAddressInfo(ctx context.Context, a address.Address) (AddressInfo, error)
 
 	// ListAddresses lists all addresses for a given account, including
@@ -680,7 +685,9 @@ func nextUnusedStoreAddress(storeAddr db.AddressInfo,
 	return addr, true, nil
 }
 
-// GetAddressInfo returns detailed information regarding a wallet address.
+// GetAddressInfo returns detailed information regarding a wallet address. If
+// the address is not known to the wallet, the returned error wraps
+// ErrAddressNotFound for errors.Is matching.
 func (w *Wallet) GetAddressInfo(ctx context.Context, a address.Address) (
 	AddressInfo, error) {
 
@@ -701,6 +708,14 @@ func (w *Wallet) GetAddressInfo(ctx context.Context, a address.Address) (
 		},
 	)
 	if err != nil {
+		// Store implementations normalize backend-specific misses before
+		// this boundary. Translate only that normalized form so raw legacy
+		// manager errors remain unexpected errors.
+		if errors.Is(err, db.ErrAddressNotFound) {
+			return AddressInfo{}, fmt.Errorf("%w: addr=%v",
+				ErrAddressNotFound, a)
+		}
+
 		return AddressInfo{}, err
 	}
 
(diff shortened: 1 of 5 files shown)


wallet/signer.go
```diff
@@ -1246,7 +1246,7 @@ func (w *Wallet) privKeyForAddress(ctx context.Context,
 
 		return w.resolveImportedAddrPrivKey(ctx, scriptPubKey)
 
-	case errors.Is(err, db.ErrAddressNotFound):
+	case errors.Is(err, ErrAddressNotFound):
```
yyforyongyu, line 1249: this changes more than it needs? previously we won't match `waddrmgr.ErrAddressNotFound` but we do, is this intended?

wallet/address_manager_api_test.go
```diff
@@ -0,0 +1,136 @@
+// Copyright (c) 2026 The btcsuite developers
```
yyforyongyu: remove this file - we should not add arbitrary test files, instead, stick to the current file pattern.

wallet/address_manager_api_test.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/btcsuite/btcd/chainhash/v2"
+	"github.com/btcsuite/btcd/wire/v2"
+	bwmock "github.com/btcsuite/btcwallet/bwtest/mock"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// addressLookupChain supplies the deterministic genesis data required while a
+// real-backend Wallet starts for the public address lookup test.
+type addressLookupChain struct {
```
yyforyongyu: why do we even need this? don't we have all the mocks defined already?

wallet/signer.go
yyforyongyu: Addressed in fa147273a: the wallet mapper now translates only the normalized db.ErrAddressNotFound sentinel. The signer regression also proves that a raw waddrmgr.ErrAddressNotFound remains an unexpected Store failure.

wallet/address_manager_api_test.go
yyforyongyu: Addressed in fa147273a: wallet/address_manager_api_test.go was removed, and the regression coverage now lives in the existing address_manager_test.go and signer_test.go files.

wallet/address_manager_api_test.go
yyforyongyu: Addressed in fa147273a: the custom mock was removed and the tests use the existing mockWalletDeps Store mock.

wallet/address_manager.go
```diff
@@ -680,7 +685,22 @@ func nextUnusedStoreAddress(storeAddr db.AddressInfo,
 	return addr, true, nil
 }
 
-// GetAddressInfo returns detailed information regarding a wallet address.
+// mapAddressLookupError translates the normalized Store address-miss sentinel
+// into the wallet-owned public contract while preserving the queried address
+// as diagnostic context. Store implementations normalize backend-specific
+// misses before this boundary, so raw legacy manager errors deliberately remain
+// unexpected errors here.
+func mapAddressLookupError(a address.Address, err error) error {
```
yyforyongyu: i don't think we need this thin wrapper?

yyforyongyu: Addressed in a7b215ede: removed the single-use helper and inlined the normalized db.ErrAddressNotFound translation in Wallet.GetAddressInfo, while preserving raw legacy manager errors as unexpected failures.

wallet/address_manager.go
```diff
@@ -701,6 +708,15 @@ func (w *Wallet) GetAddressInfo(ctx context.Context, a address.Address) (
 		},
 	)
 	if err != nil {
+		// Store implementations normalize backend-specific misses before
+		// this boundary. Translate only that normalized form so raw legacy
+		// manager errors remain unexpected errors.
+		if errors.Is(err, db.ErrAddressNotFound) {
+			return AddressInfo{}, fmt.Errorf(
+				"%w: addr=%v", ErrAddressNotFound, a,
+			)
```
GustavoStingelin: nit: Should we still take care of the formatting rules?
```suggestion
		if errors.Is(err, db.ErrAddressNotFound) {
			return AddressInfo{}, fmt.Errorf("%w: addr=%v",
				ErrAddressNotFound, a)
```

GustavoStingelin (approved):
LGTM!

wallet/address_manager.go
yyforyongyu: done!

PR #1334, opened 2026-08-17
itest: add TxPublisher integration tests

Abdulkbk:
Implements Task 435 — cover transaction publication contracts.

`CheckMempoolAcceptance` and `Broadcast` had unit coverage against a mocked
chain and no integration coverage at all. This adds six cases in
`itest/tx_publisher_test.go`.

## Task 435 acceptance criteria mapped to tests

| Criterion | Case |
| --- | --- |
| Accepted spend | `txpublisher check acceptance`, `txpublisher broadcast transaction` |
| Representative policy rejection | `txpublisher reject acceptance` (dust → `chain.ErrDust`), `txpublisher reject broadcast` |
| Malformed or conflicting | `txpublisher reject acceptance` (unknown input → `chain.ErrMissingInputs`, duplicate input → `chain.ErrDuplicateInput`) |
| Already-known behavior | `txpublisher broadcast known` |
| Stopped Wallet state | `txpublisher wallet state` (both methods, before Start and after Stop; nil transaction on a running wallet) |
| Acceptance check never publishes | `txpublisher check acceptance` (mempool, wallet records and coins all unchanged) |
| Broadcast observable before and after mining | `txpublisher broadcast transaction` |
| kvdb, SQLite and PostgreSQL, no skip branch | Full suite passes on all three; no case contains database-conditional code |
| Test-only | No production file is touched |

## Harness changes

Both additive, in `bwtest`:

- `SpendFixture` and `SignSpend` — authors and signs a transaction from named
  wallet coins, using the inputs and outputs verbatim. Not built on
  `CreateTransaction`/`FundPsbt`, which enforce the creation-time policy Task
  434 owns and reject the dust fixture outright.
- `MempoolAcceptanceSupported` — one authority for which chain backends answer
  the acceptance check.

## Backend neutrality

The rejections were chosen so one identity holds on every backend. Three
candidates were measured and discarded:

- **Minimum relay fee** — btcd reports a mempool minimum, bitcoind a relay
  minimum. Dust carries the policy requirement instead.
- **Double spend** — bitcoind enables full replace-by-fee by default, so it
  judges the second transaction as a replacement bid, accepting it when it pays
  more and reporting an insufficient fee when it pays less, while btcd reports a
  mempool conflict. `MaxTxInSequenceNum` does not prevent this. Duplicate inputs
  cover the malformed requirement instead.
- **No outputs** — btcd rejects on serialized size before reaching the
  empty-output rule, mapping to `ErrUndefined`.

Neutrino has no mempool and reports the acceptance check as unimplemented. The
two cases whose subject is that check assert `chain.ErrUnimplemented` there
rather than skipping, so every backend gets a definite assertion. Everything
else is backend-neutral with no split, including the dust refusal, which
neutrino's peer returns over the wire.

Part of #1250.

### DIFF
--- a/bwtest/README.md
+++ b/bwtest/README.md
@@ -68,6 +68,43 @@ Funding is also available on its own through `(*HarnessTest).FundWallet` and
 Manager-focused tests should continue to create wallets through the manager API
 directly.
 
+## Signed Transactions
+
+`(*HarnessTest).SignSpend` authors and signs a transaction from a wallet's own
+coins. It takes a `SpendFixture` naming the inputs and the outputs, and returns
+the signed transaction:
+
+```go
+tx := t.SignSpend(w, bwtest.SpendFixture{
+	Inputs:  []wire.OutPoint{funding.WalletOutpoints[0]},
+	Outputs: []wire.TxOut{{Value: amount, PkScript: pkScript}},
+})
+```
+
+The inputs and outputs are used verbatim, in the order given, and every input is
+final. Unlike `CreateTransaction` no coin selection, change or output policy is
+applied, so a case can author a transaction the network is required to reject —
+a dust payment, a fee the relay will not carry — which is what publication tests
+need. Cases that mean to test coin selection or change should use
+`CreateTransaction` instead.
+
+The wallet must be started, synced and unlocked, and it must own every named
+input.
+
+To build a rejected output without writing down an amount that drifts from
+policy, `(*HarnessTest).DustThreshold` returns the smallest amount an output
+paying a given script may carry; anything below it is dust.
+
+## Chain Backend Capabilities
+
+Not every backend answers every question. `ChainBackend.SupportsMempoolAcceptance`
+reports whether the backend under test runs the mempool acceptance check, and
+each backend answers for the client it constructs so the answer cannot drift
+from the implementation. Neutrino is a light client with no mempool and reports
+the check as unimplemented instead of returning a verdict, so cases reach it as
+`t.Backend.SupportsMempoolAcceptance()` and assert a definite result on both
+sides of the split rather than skipping a backend.
+
 ## Fast Scrypt
 
 `bwtest` sets `waddrmgr.DefaultScryptOptions` to `waddrmgr.FastScryptOptions` via
--- a/bwtest/backend.go
+++ b/bwtest/backend.go
@@ -26,6 +26,16 @@ type ChainBackend interface {
 	// Name returns the name of the backend ("btcd", "bitcoind", "neutrino").
 	Name() string
 
+	// SupportsMempoolAcceptance reports whether the clients this backend
+	// creates answer the mempool acceptance test. A backend that keeps no
+	// mempool reports the check as unimplemented instead of returning a
+	// verdict.
+	//
+	// Each backend answers for the client it constructs, so the answer
+	// cannot drift from the implementation the way a central lookup by
+	// name would.
+	SupportsMempoolAcceptance() bool
+
 	// Start launches the chain backend.
 	Start() error
 
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -2,16 +2,16 @@ package bwtest
 
 import (
 	"bytes"
-	"strings"
-	"time"
-
 	"github.com/btcsuite/btcd/address/v2"
 	"github.com/btcsuite/btcd/btcutil/v2"
+	"github.com/btcsuite/btcd/mempool"
 	"github.com/btcsuite/btcd/txscript/v2"
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/stretchr/testify/require"
+	"strings"
+	"time"
 )
 
 const (
@@ -328,6 +328,26 @@ func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
 	}
 
 	return funding
+}
+
+// DustThreshold returns the smallest amount an output paying pkScript may
+// carry without the network refusing it as dust. Any amount below it is dust.
+//
+// The limit depends on the output's script, so it is taken from the relay
+// policy rather than written down as a number that would silently drift from
+// it. The policy states dust as the comparison
+// value*1000/threshold < relayFee, which first holds at the threshold itself,
+// so the threshold is the smallest amount that clears it at the default relay
+// fee. A case paying a different relay fee cannot use this.
+//
+// The script is assumed spendable, which is what paying a wallet address
+// produces. An unspendable script is dust at any amount and has no threshold.
+func (h *HarnessTest) DustThreshold(pkScript []byte) btcutil.Amount {
+	h.Helper()
+
+	output := wire.TxOut{PkScript: pkScript}
+
+	return btcutil.Amount(mempool.GetDustThreshold(&output))
 }
 
 // ensureAccount makes sure an account exists for the given key scope, creating
--- a/bwtest/harness_wallet.go
+++ b/bwtest/harness_wallet.go
@@ -2,16 +2,18 @@ package bwtest
 
 import (
 	"bytes"
+	"strings"
+	"time"
+
 	"github.com/btcsuite/btcd/address/v2"
 	"github.com/btcsuite/btcd/btcutil/v2"
 	"github.com/btcsuite/btcd/mempool"
+	"github.com/btcsuite/btcd/psbt/v2"
 	"github.com/btcsuite/btcd/txscript/v2"
 	"github.com/btcsuite/btcd/wire/v2"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/stretchr/testify/require"
-	"strings"
-	"time"
 )
 
 const (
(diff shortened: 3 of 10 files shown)


itest/tx_publisher_test.go
```diff
@@ -0,0 +1,501 @@
...
+
+// spendFee is the fee in satoshis that the publication fixtures leave behind.
+// It clears the relay minimum by a wide margin while staying far below the
+// backend's default maximum fee rate of 10,000 sat/vbyte, above which an
+// otherwise valid transaction is rejected as paying absurdly much.
+const spendFee = 10_000
+
+// dustAmount is below the relay's dust threshold for every output type the
+// wallet produces, so a payment of it is refused as dust.
+const dustAmount = 100
```
yyforyongyu: let's do sth like this rather than defining an arbitrary amount? https://github.com/btcsuite/btcwallet/blob/a6865a37e0493df62a7d895a47269022fe62ed66/itest/tx_creator_test.go#L723-L725

maybe we can create sth on the `harness` such that it can always give us a dust amount

itest/tx_publisher_test.go
```diff
@@ -0,0 +1,501 @@
...
+		err := w.CheckMempoolAcceptance(h.Context(), rejection.tx)
+		require.ErrorIs(
+			h, err, want, "unexpected verdict for %s",
+			rejection.name,
+		)
+
+		h.AssertTxNotInMempool(rejection.tx.TxHash())
+	}
+
+	// A refused transaction leaves the coin it named still spendable.
```
yyforyongyu, line 197: 👍

bwtest/README.md
```diff
@@ -68,6 +68,37 @@ Funding is also available on its own through `(*HarnessTest).FundWallet` and
...
+
+`(*HarnessTest).SignSpend` authors and signs a transaction from a wallet's own
+coins. It takes a `SpendFixture` naming the inputs, the outputs and the input
+sequence, and returns the signed transaction:
+
+```go
+tx := t.SignSpend(w, bwtest.SpendFixture{
+	Inputs:   []wire.OutPoint{funding.WalletOutpoints[0]},
+	Outputs:  []wire.TxOut{{Value: amount, PkScript: pkScript}},
+	Sequence: wire.MaxTxInSequenceNum,
```
yyforyongyu: there is no such field?

yyforyongyu (changes requested):
Again very easy to follow!

itest/tx_publisher_test.go
```diff
@@ -0,0 +1,501 @@
...
+	// And the coin it named is untouched.
+	utxo, err := w.GetUtxo(h.Context(), spent)
+	require.NoError(h, err, "refused transaction consumed the coin")
+	require.True(h, utxo.Spendable, "coin is no longer spendable")
+}
+
+// testPublishWalletState verifies the lifecycle gate both publication methods
+// share: neither is available before the wallet starts or after it stops, and a
+// running wallet refuses a nil transaction rather than acting on it.
+func testPublishWalletState(h *bwtest.HarnessTest) {
```
yyforyongyu: This case covers both `CheckMempoolAcceptance` and `Broadcast`, across unstarted, running, and stopped states, and it also tests nil arguments. That combines several public contracts and interleaves multiple Act/Assert phases. Can we split this by API and behavior? For Task 435, separate stopped-wallet cases may be sufficient; the pre-start and nil checks could remain unit coverage or become distinct integration cases.

bwtest/backend.go
```diff
@@ -46,6 +46,17 @@ type ChainBackend interface {
...
 }
 
+// MempoolAcceptanceSupported reports whether the chain backend under test
+// answers the mempool acceptance test.
+//
+// Neutrino is a light client with no mempool, so it reports the check as
+// unimplemented instead of returning a verdict. Publication test cases assert
+// an equally definite, but different, result against it, and this is the single
+// place that decides which backends have a mempool.
+func (h *HarnessTest) MempoolAcceptanceSupported() bool {
```
yyforyongyu: This derives a capability from the backend name even though the actual behavior is owned by the chain implementation, so the two can drift. The tests also accept `ErrUnimplemented` for neutrino, but the public `TxPublisher.CheckMempoolAcceptance` documentation does not describe that backend-dependent contract. Could we expose the capability through the backend contract and document the possible error on the public API?

bwtest/harness_wallet.go
```diff
@@ -330,6 +332,93 @@ func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
...
 }
 
+// DustThreshold returns the smallest amount an output paying pkScript may
+// carry without the network refusing it as dust. Any amount below it is dust.
+//
+// The limit depends on the output's script, so it is derived from the relay
+// policy itself rather than written down as a number that would silently drift
+// from it.
+func (h *HarnessTest) DustThreshold(pkScript []byte) btcutil.Amount {
+	h.Helper()
```
GustavoStingelin, line 348: Could this use `mempool.GetDustThreshold(&output)` directly? For the spendable scripts used by these tests, it appears to return the same threshold at `DefaultRelayFeePerKb`. If this loop handles another policy case, please document it.

bwtest/neutrino.go
```diff
@@ -32,6 +32,15 @@ func NewNeutrinoBackend(t *testing.T, _ string) *NeutrinoBackend {
...
 }
 
+// SupportsMempoolAcceptance reports that neutrino does not answer the mempool
+// acceptance test. A light client keeps no mempool, so its TestMempoolAccept
+// returns chain.ErrUnimplemented for every transaction.
+//
+// NOTE: This is part of the ChainBackend interface.
+func (n *NeutrinoBackend) SupportsMempoolAcceptance() bool {
+	return false
+}
```
GustavoStingelin, line 42: 👍🏼

bwtest/harness_wallet.go
```diff
@@ -330,6 +332,93 @@ func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
...
+
+	// The wallet supplies each input's previous output and derivation path
+	// from its own records, so the fixture only has to name the outpoints.
+	// Unknown inputs are an error rather than a silently unsigned input,
+	// which would surface later as an opaque finalization failure.
+	packet, err = w.DecorateInputs(h.Context(), packet, false)
+	require.NoError(h, err, "failed to decorate psbt inputs")
+
+	err = w.FinalizePsbt(h.Context(), packet)
+	require.NoError(h, err, "failed to finalize psbt")
```
GustavoStingelin, line 419: Non-blocking: `FinalizePsbt` also signs wallet-owned inputs before finalization. A short comment here could clarify why no separate `SignPsbt` call is needed.

itest/tx_publisher_test.go
Abdulkbk: Added to harness bc9fe4b7292ed9dbcdb6b2c7e6f70a37e653d261

bwtest/README.md
Abdulkbk: Stale doc, now fixed

itest/tx_publisher_test.go
Abdulkbk: splitted

bwtest/backend.go
Abdulkbk: Done, exposed it through backend contract and also document the public API

bwtest/harness_wallet.go
```diff
@@ -330,6 +332,93 @@ func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
...
+// carry without the network refusing it as dust. Any amount below it is dust.
+//
+// The limit depends on the output's script, so it is derived from the relay
+// policy itself rather than written down as a number that would silently drift
+// from it.
+func (h *HarnessTest) DustThreshold(pkScript []byte) btcutil.Amount {
+	h.Helper()
+
+	output := wire.TxOut{PkScript: pkScript}
+	for output.Value = 1; ; output.Value++ {
```
yyforyongyu: we can just do `return btcutil.Amount(mempool.GetDustThreshold(txOut))`

yyforyongyu (changes requested):
One last thing, otherwise good to go!

GustavoStingelin (changes requested):
looking very good, left just a few comments

bwtest/harness_wallet.go
GustavoStingelin: left this one too, haha

Abdulkbk: done

bwtest/harness_wallet.go
Abdulkbk: done

bwtest/harness_wallet.go
Abdulkbk: added

bwtest/harness_wallet.go
```diff
@@ -330,6 +332,98 @@ func (h *HarnessTest) FundWalletOfType(w *wallet.Wallet,
...
+// fee. A case paying a different relay fee cannot use this.
+//
+// The script is assumed spendable, which is what paying a wallet address
+// produces. An unspendable script is dust at any amount and has no threshold.
+func (h *HarnessTest) DustThreshold(pkScript []byte) btcutil.Amount {
+	h.Helper()
+
+	output := wire.TxOut{PkScript: pkScript}
+
+	return btcutil.Amount(mempool.GetDustThreshold(&output))
```
GustavoStingelin, line 352: Is it worth removing this function now that it's so simplified? Creating a whole function just to save one line of code might be overkill.

yyforyongyu (approved):
LGTM!

bwtest/harness_wallet.go
yyforyongyu: I think it's fine - also depends on how often it's used as 2x reducation is good. But more importantly, I think it's there such that the agent won't miss using the correct way to calculate dust, which previously it ended up using a `for` loop

Abdulkbk: Yes, I think we should leave it too.

GustavoStingelin: ok

GustavoStingelin (approved):

GustavoStingelin:
@Abdulkbk my rebase from ui removed ur sigs, could u rebase this one?

PR #1335, opened 2026-08-17
wallet: accept bounded recovery windows

Abdulkbk:
Implements Task 342 — accept bounded recovery windows.

`Config.validate` rejected every `RecoveryWindow` below 20 and bounded it from
above not at all. The floor was artificial: `RecoveryWindow` has exactly one
non-test consumer, `syncer.go`'s `NewRecoveryState`, and recovery already
operates with windows far smaller than 20. Nothing between the configuration
and the scan reads the field, so a caller asking for a window of one — or for
none — was refused a request the syncer would have served unchanged.

## What changed

`wallet/wallet.go` only:

- `MinRecoveryWindow = 20` is replaced by `MaxRecoveryWindow = 100_000`.
- `Config.validate` rejects `> MaxRecoveryWindow` with the existing
  wallet-owned `ErrInvalidParam` contract, and accepts everything below it.
- The `RecoveryWindow` field comment states the accepted range.

The check keeps its position in `validate`, which `Manager.Create` and
`Manager.Load` call before `deriveRootKey` and `backend.create` — so the
rejection lands before any Store or chain access rather than depending on one.

## Task 342 acceptance criteria mapped to tests

| Criterion | Test |
| --- | --- |
| Accepts 0, 1, 2, 19, 20, `MaxRecoveryWindow` unchanged | `TestConfigValidateRecoveryWindow` (subtest per value) |
| Maximum-plus-one returns the wallet-owned invalid-parameter error | `TestConfigValidateRecoveryWindow/rejects_the_first_excess_window` — `ErrorIs(ErrInvalidParam)` |
| ...without Store or chain access | `TestManagerCreate_ExcessRecoveryWindow` — `recordingManagerBackend.createCalled` is false, and a `bwmock.Chain` with no registered calls passes `AssertExpectations` |
| Existing small-window recovery fixtures remain valid | Full `wallet` package suite passes; `bwtest`'s window of 20 is untouched |
| RPC retains its separate negative-value check | No RPC file is touched. `RecoveryWindow` stays `uint32`, so the unsigned configuration never sees a negative value |

## Verification

- `go test ./wallet/...` — all packages pass.
- `golangci-lint run ./wallet/...` — 0 issues.

### DIFF
--- a/wallet/wallet.go
+++ b/wallet/wallet.go
@@ -60,11 +60,11 @@ const (
 	// locking.
 	defaultLockDuration = 10 * time.Minute
 
-	// MinRecoveryWindow is the minimum allowed value for the RecoveryWindow
-	// configuration parameter. This value ensures that a sufficient number
-	// of addresses are scanned during wallet recovery to avoid missing
-	// funds due to gaps in the address chain.
-	MinRecoveryWindow = 20
+	// MaxRecoveryWindow is the largest accepted value for the
+	// RecoveryWindow configuration parameter. The bound is public so a
+	// caller can reject an expensive request on its own, before the wallet
+	// reaches its Store or the chain backend to serve it.
+	MaxRecoveryWindow = 100_000
 )
 
 var (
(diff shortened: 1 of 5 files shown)


wallet/wallet.go
```diff
@@ -187,7 +187,9 @@ type Config struct {
 	// ChainParams defines the network parameters (e.g. mainnet, testnet).
 	ChainParams *chaincfg.Params
 
-	// RecoveryWindow specifies the address lookahead for recovery.
+	// RecoveryWindow specifies the address lookahead for recovery. Any
+	// value from zero through MaxRecoveryWindow is accepted; zero scans
+	// only the addresses the wallet already knows.
```
yyforyongyu: Let's make a stronger godoc here to cover,
- Values are used exactly as supplied.
- btcwallet applies no default.
- Callers must choose any desired default.
- Zero disables lookahead derivation but still synchronizes persisted targets.

Was thinking whether we should give a default value, like 20, but then think it's cleaner this way, and we let the upper layer handle the defaults, so we need to provide a verbose and detailed doc here.

yyforyongyu (changes requested):

yyforyongyu (approved):
LGTM🧷

PR #1336, opened 2026-08-18
wallet: serialize Manager wallet assembly

yyforyongyu:
## Change Description

Implements roadmap Task 453.

Serialize every Manager `Create` and cold `Load` operation through one
Manager-wide lock. The lock spans durable creation, or the Load cache
decision, through backend assembly and cache installation. Concurrent
same-name Loads therefore return one exact Wallet pointer, and Create cannot
be overtaken between durable commit and runtime installation.

The cache remains `map[string]*Wallet` and contains only fully assembled
Wallets. Different wallet names intentionally assemble sequentially; there
are no per-name entries, channels, transient nil values, or result latches.

## Steps to Test

1. Run: `make unit pkg=wallet`
2. Run the concurrent Manager Load integration test for each backend:
   - `make itest chain=btcd db=kvdb icase="manager load concurrent"`
   - `make itest chain=btcd db=sqlite icase="manager load concurrent"`
   - `make itest chain=btcd db=postgres icase="manager load concurrent"`
3. Run: `make lint-check`

### Testing

### DIFF
--- a/itest/list_on_test.go
+++ b/itest/list_on_test.go
@@ -28,6 +28,10 @@ var allTestCases = []*testCase{
 		TestFunc: testManagerLoadReload,
 	},
 	{
+		Name:     "manager load concurrent",
+		TestFunc: testManagerLoadConcurrent,
+	},
+	{
 		Name:     "manager load missing",
 		TestFunc: testManagerLoadMissing,
 	},
--- a/itest/manager_test.go
+++ b/itest/manager_test.go
@@ -3,6 +3,7 @@
 package itest
 
 import (
+	"sync"
 	"time"
 
 	"github.com/btcsuite/btcwallet/bwtest"
@@ -10,6 +11,74 @@ import (
 	"github.com/btcsuite/btcwallet/wallet"
 	"github.com/stretchr/testify/require"
 )
+
+// testManagerLoadConcurrent verifies that every supported Manager backend gives
+// concurrent cold Loads one exact runtime Wallet pointer.
+func testManagerLoadConcurrent(h *bwtest.HarnessTest) {
+	// Arrange. The setup Manager creates durable state, then closes so the
+	// test Manager opens the same backend with an empty runtime cache. Using
+	// two Managers is required to exercise a true cold Load on kvdb as well
+	// as SQL.
+	cfg, params := h.TestWalletConfig()
+	setupManager := h.NewWalletManager()
+
+	created, err := setupManager.Create(cfg, params)
+	require.NoError(h, err, "failed to create wallet")
+	require.NoError(h, created.Stop(h.Context()), "failed to stop wallet")
+	require.NoError(h, setupManager.Close(), "failed to close setup manager")
+	require.True(
+		h, h.ReleaseManager(setupManager),
+		"failed to release setup manager",
+	)
+
+	manager := h.NewWalletManager()
+
+	// Multiple callers contend for cold assembly. The exact number is not
+	// significant as long as it is greater than one.
+	const numCallers = 8
+
+	type result struct {
+		// wallet is the exact runtime returned to one caller.
+		wallet *wallet.Wallet
+
+		// err is the Load result returned with wallet.
+		err error
+	}
+
+	results := make(chan result, numCallers)
+	start := make(chan struct{})
+
+	var ready sync.WaitGroup
+
+	ready.Add(numCallers)
+
+	for range numCallers {
+		go func() {
+			ready.Done()
+			<-start
+
+			w, err := manager.Load(cfg)
+			results <- result{wallet: w, err: err}
+		}()
+	}
+
+	// Act. Release every caller into the empty Manager cache together.
+	ready.Wait()
+	close(start)
+
+	// Assert. One caller assembles the runtime and every other caller returns
+	// that exact installed pointer.
+	first := <-results
+	require.NoError(h, first.err, "failed to load wallet")
+	installed := first.wallet
+	h.RegisterWallet(installed)
+
+	for range numCallers - 1 {
+		result := <-results
+		require.NoError(h, result.err, "failed to load wallet")
+		require.Same(h, installed, result.wallet, "load rebuilt wallet")
+	}
+}
 
 // testCreateWallet verifies a wallet can be created, started, and synced.
 func testCreateWallet(h *bwtest.HarnessTest) {
--- a/wallet/manager.go
+++ b/wallet/manager.go
@@ -107,10 +107,11 @@ type CreateWalletParams struct {
 type Manager struct {
 	sync.RWMutex
 
-	// wallets holds the active wallets keyed by their unique name, published
-	// as soon as they are assembled. A ModeShell Create imports its initial
-	// accounts after publishing, so a wallet can be observed here before
-	// that import finishes.
+	// wallets holds the active wallets keyed by their unique name. The
+	// Manager lock serializes Create and Load through assembly and cache
+	// installation. A ModeShell Create imports its initial accounts after
+	// installation, so a wallet can be observed here before that import
+	// finishes.
 	wallets map[string]*Wallet
 
 	// backend owns the database and resolves the storage dependencies for
@@ -209,16 +210,18 @@ func (m *Manager) Create(cfg Config,
 		return nil, err
 	}
 
+	m.Lock()
+
 	data, err := m.backend.create(
 		context.Background(), cfg, params, rootKey,
 	)
 	if err != nil {
+		m.Unlock()
+
 		return nil, err
 	}
 
 	w := newManagedWallet(cfg, data)
-
-	m.Lock()
 	m.wallets[cfg.Name] = w
 	m.Unlock()
 
@@ -366,10 +369,6 @@ func validateInitialAccountKeys(accounts []WatchOnlyAccount) error {
 // Load loads an existing wallet from the provided configuration. It opens the
 // database, initializes the wallet structure, and registers it with the manager
 // for tracking.
-//
-// TODO(yy): concurrent Load calls for the same name are not serialized, so two
-// callers racing on one wallet can both open a store. Call it from one
-// goroutine per wallet until that is fixed.
 func (m *Manager) Load(cfg Config) (*Wallet, error) {
 	// The Manager owns the network for every wallet it serves; see Create.
 	cfg.ChainParams = m.chainParams
@@ -379,12 +378,12 @@ func (m *Manager) Load(cfg Config) (*Wallet, error) {
 		return nil, err
 	}
 
-	// A wallet already published under this name is returned as is; callers
-	// serialize Create and Load, so a miss means this call builds it.
-	m.RLock()
-	existingW, ok := m.wallets[cfg.Name]
-	m.RUnlock()
+	m.Lock()
+	defer m.Unlock()
 
+	// A wallet already installed under this name is returned as is. The
+	// Manager lock makes a miss the sole assembly owner until installation.
+	existingW, ok := m.wallets[cfg.Name]
 	if ok {
 		return existingW, nil
 	}
@@ -395,10 +394,7 @@ func (m *Manager) Load(cfg Config) (*Wallet, error) {
 	}
 
 	w := newManagedWallet(cfg, data)
-
-	m.Lock()
 	m.wallets[cfg.Name] = w
-	m.Unlock()
 
 	return w, nil
 }


itest/manager_test.go
```diff
@@ -11,6 +12,59 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+// testManagerWalletPublication verifies that every supported Manager backend
+// gives concurrent cold Loads one exact runtime Wallet pointer.
+func testManagerWalletPublication(h *bwtest.HarnessTest) {
+	cfg, params := h.TestWalletConfig()
+
+	creator := h.NewWalletManager()
```
yyforyongyu: name it `manager` instead of `creator`? and explain why we need call this `NewWalletManager` twice?

itest/manager_test.go
```diff
@@ -11,6 +12,59 @@ import (
...
+	created, err := creator.Create(cfg, params)
+	require.NoError(h, err, "failed to create wallet")
+	require.NoError(h, created.Stop(h.Context()), "failed to stop wallet")
+	require.NoError(h, creator.Close(), "failed to close creator")
+	require.True(h, h.ReleaseManager(creator), "failed to release creator")
+
+	manager := h.NewWalletManager()
+
+	const callers = 8
```
yyforyongyu: explain this number, and name it `numCallers`?

itest/manager_test.go
```diff
@@ -11,6 +12,59 @@ import (
...
+
+	type result struct {
```
yyforyongyu, line 40: also add inline comments explaining this struct

itest/manager_test.go
```diff
@@ -11,6 +12,59 @@ import (
...
+
+	ready.Wait()
+	close(start)
+
+	first := <-results
+	require.NoError(h, first.err, "failed to load wallet")
+	published := first.wallet
+	h.RegisterWallet(published)
+
+	for range callers - 1 {
```
yyforyongyu: and where are the AAA comments? like details?

wallet/manager.go
```diff
@@ -121,6 +122,14 @@ type Manager struct {
 	chainParams *chaincfg.Params
 }
 
+// walletRuntimeEntry holds one current Wallet or an in-flight publication
```
yyforyongyu: what is `publication`? i think use `creation` is more accurate?

wallet/manager.go
```diff
@@ -171,7 +180,11 @@ func (m *Manager) String() string {
 	defer m.RUnlock()
 
 	names := make([]string, 0, len(m.wallets))
-	for name := range m.wallets {
+	for name, entry := range m.wallets {
+		if entry.wallet == nil {
```
yyforyongyu: why should we silently skip the `nil` wallet?

wallet/manager.go
```diff
@@ -121,6 +122,14 @@ type Manager struct {
 	chainParams *chaincfg.Params
 }
 
+// walletRuntimeEntry holds one current Wallet or an in-flight publication
+// result. Lifecycle state is deliberately not part of this entry.
+type walletRuntimeEntry struct {
+	wallet          *Wallet
+	publicationDone chan struct{}
+	publicationErr  error
```
yyforyongyu: why do we need this entry?

wallet/manager.go
```diff
@@ -379,28 +389,54 @@ func (m *Manager) Load(cfg Config) (*Wallet, error) {
...
-	existingW, ok := m.wallets[cfg.Name]
-	m.RUnlock()
+	return m.publishWallet(cfg, func() (*walletData, error) {
+		return m.backend.load(context.Background(), cfg)
+	})
+}
+
+// publishWallet joins or owns publication for cfg.Name. The data function is
+// invoked only by the publication owner, so no losing Wallet is assembled.
+func (m *Manager) publishWallet(cfg Config,
```
yyforyongyu: i think we are overcomplicating things here - there is no need to support concurrent creation of wallets, all we need is to allow create wallets, using a lock, and do it sequentially?

wallet/manager.go
yyforyongyu: Addressed through the redesign. The publication abstraction and terminology are gone; the code and tests now describe synchronous Manager wallet assembly and cache installation.

wallet/manager.go
yyforyongyu: Addressed. `Create` and cold `Load` now serialize through the Manager lock, including different wallet names. The lock spans durable Create or the Load cache decision through assembly and cache installation; the per-name gate, channels, waiters, and latched results were removed.

wallet/manager.go
yyforyongyu: Addressed through the redesign. Task 453 keeps `map[string]*Wallet`, never inserts an in-flight or nil value, and `Manager.String` no longer needs to skip transient entries.

itest/manager_test.go
yyforyongyu: Addressed. The integration case now has explicit Arrange, Act, and Assert comments, including the reason for two Managers and the cold-cache boundary being tested.

itest/manager_test.go
yyforyongyu: Addressed. The integration test now uses `setupManager` for durable setup and `manager` for the cold-load assertion. Its Arrange comment explains that the first Manager must close so the second opens the same backend with an empty runtime cache on both kvdb and SQL.

itest/manager_test.go
yyforyongyu: Addressed. The constant is now `numCallers`, with a comment explaining that the exact value is not significant and only needs to be greater than one to contend cache admission.

wallet/manager.go
yyforyongyu: Addressed through the redesign. `walletRuntimeEntry` is no longer part of Task 453. The cache stores only fully assembled Wallet pointers; Task 368 introduces a runtime entry later because it has actual lifecycle state to own.

itest/manager_test.go
yyforyongyu: Addressed. Both fields in the local result struct now have inline comments describing the exact Wallet pointer and associated Load error.

wallet/manager_assembly_test.go
```diff
@@ -0,0 +1,311 @@
...
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
+	bwmock "github.com/btcsuite/btcwallet/bwtest/mock"
+	"github.com/stretchr/testify/require"
+)
+
+var errAssemblyTest = errors.New("assembly failed")
+
+// assemblyTestBackend exposes controlled Manager assembly behavior.
+type assemblyTestBackend struct {
+	createFn func(context.Context, Config, CreateWalletParams,
```
yyforyongyu: we should use `mock.Mock`

wallet/manager_assembly_test.go
```diff
@@ -0,0 +1,311 @@
...
+	cached := m.wallets[cfg.Name]
+	m.RUnlock()
+
+	require.Same(t, installed, cached)
+	stopAssemblyWallet(installed)
+}
+
+// testManagerAssemblyDifferentNames proves one Manager admits assembly
+// sequentially even when wallet names differ.
+func testManagerAssemblyDifferentNames(t *testing.T) {
```
yyforyongyu: these should be moved into `manager_test.go`? and these tests seem to be duplicating the `testManagerWalletAssembly` itest?

itest/list_on_test.go
```diff
@@ -27,6 +27,10 @@ var allTestCases = []*testCase{
 		Name:     "manager load reload",
 		TestFunc: testManagerLoadReload,
 	},
+	{
+		Name:     "manager wallet assembly",
+		TestFunc: testManagerWalletAssembly,
```
yyforyongyu: we should name it consistently - so sth like `manager load concurrent`

wallet/manager_assembly_test.go
yyforyongyu: Follow-up: the new unit-test suite was removed, so assemblyTestBackend no longer exists and there is no custom interface mock to maintain.

wallet/manager_assembly_test.go
yyforyongyu: Follow-up: removed the new unit-test suite entirely. Concurrent cold-Load pointer identity remains covered by the real-backend integration test.

itest/list_on_test.go
yyforyongyu: Renamed the case to "manager load concurrent" and the function to testManagerLoadConcurrent in cb61f0c94.

wallet/manager_test.go
```diff
@@ -960,3 +962,295 @@ func TestManagerCreateHonoursCreatePubPassphrase(t *testing.T) {
...
+
+// assemblyTestBackend implements managerBackend.
+var _ managerBackend = (*assemblyTestBackend)(nil)
+
+// create runs the configured creation behavior.
+func (b *assemblyTestBackend) create(ctx context.Context, cfg Config,
+	params CreateWalletParams, rootKey *hdkeychain.ExtendedKey) (
+	*walletData, error) {
+
+	args := b.MethodCalled("create", ctx, cfg, params, rootKey)
```
yyforyongyu: we should use `args := m.Called`??

Lrifton92 (commented):
The race is real and the fix closes it. Worth saying that `testManagerLoadConcurrent` is genuinely discriminating rather than decorative: if two callers both assembled, each would return the runtime it built, so `require.Same` against `first` would fail — the assertion actually proves single assembly rather than just a consistent final state. Three things on the approach.

### The lock is global, the documented race was per-name

The TODO this removes reads "concurrent Load calls **for the same name** are not serialized". The fix serializes every name: `Load` now holds the exclusive Manager lock across `m.backend.load(...)` (`wallet/manager.go:381`) and `Create` holds it across `m.backend.create(...)` (line 213). Both of those open stores, so wallet assembly across the whole Manager is now strictly one at a time, and a slow or hanging store open for one wallet blocks `Load`, `Create` and `String` for every other.

What makes this worth raising rather than shrugging at is where it lands. `NewManager`'s own comment says "the legacy kvdb backend remains single-wallet until it is removed; SQL stores distinguish wallets by ID" — so the coarse lock is free on kvdb, which only ever holds one wallet, and costs exactly on the SQL backends that multi-wallet support exists for. A per-name guard (an in-flight map keyed by `cfg.Name`, or `singleflight`) would close the same race without the global chokepoint.

If the global lock is a deliberate simplification for now, that is a defensible call — but the new comment on `wallets` says "The Manager lock serializes Create and Load through assembly and cache installation" without noting it serializes across *all* names, and that is the part a future reader needs.

### `Create` and `Load` now have different lock discipline

`Load` takes `m.Lock(); defer m.Unlock()`. `Create` takes `m.Lock()` at line 213 and unlocks manually on two paths — line 219 on the backend error, line 226 after installation. Both are correct today. But `Create` keeps going after the unlock (the `ModeShell` import), so it cannot simply `defer`, and that asymmetry is the fragile kind: any early return added between 213 and 226 later leaks the lock and wedges the Manager permanently, with no test that would catch it. Extracting the locked span into a small helper that can `defer` would make it structurally safe instead of safe-by-inspection.

### Does `Create` need to consult the cache before installing?

`Create` never reads `m.wallets[cfg.Name]` before `m.wallets[cfg.Name] = w`. Under the new lock two concurrent `Create` calls for one name no longer interleave, but the second still overwrites the first's entry, orphaning a runtime the first caller is holding and may already have started — and the same applies to a `Create` following a `Load` of that name.

I could not settle from here whether the backends make that unreachable by failing the second `create` on a uniqueness constraint; if they do, this is a non-issue and worth a line in the comment saying so. If they do not, then "installation" clobbering a live entry seems in scope for a PR about making installation safe.

GustavoStingelin (approved):
LGTM!

yyforyongyu:
@Lrifton92 Thanks for the review. This PR intentionally follows Task 453's narrow scope:

- One Manager-wide lock, including sequential assembly across different wallet names, is an explicit task requirement; per-name gates, `singleflight`, channels, and operation registries are expressly excluded.
- The current `Create` paths both release the lock correctly. Extracting a helper solely for a hypothetical future early return would add structure without fixing a current defect.
- `Create` cannot overwrite an existing runtime: kvdb rejects creation once its store is loaded, while SQLite and PostgreSQL enforce unique wallet names at the database boundary; the backend-neutral Manager duplicate-create integration case also covers this behavior.

Given those task boundaries and existing guarantees, I do not plan code changes for these points.

Lrifton92:
Makes sense — I hadn't seen Task 453's explicit scope boundaries, and the duplicate-create guarantees at the database boundary do cover my third point. Thanks for taking the time to walk through each one.

PR #1337, opened 2026-08-19
chain: stop JitterTicker goroutine cleanly

zloglevel:
## Change Description

Fix `JitterTicker` shutdown handling in `chain/jitter.go`.

Previously, `Stop` only closed the quit channel. The `start` goroutine continued looping after receiving the quit signal. Since the quit channel remains readable after being closed, the goroutine could enter the shutdown branch again and block forever while draining an already-stopped timer.

Calling `Stop` more than once also panicked because the quit channel was closed repeatedly.

This change:

- Makes `Stop` idempotent with `sync.Once`.
- Drains the timer channel without blocking.
- Returns from the ticker goroutine after shutdown.
- Adds regression tests for shutdown, repeated `Stop` calls, post-stop ticks, and timer/stop concurrency.

## Steps to Test

```shell
go test ./chain -run '^TestJitterTickerStop' -count=100
go test -race ./chain -run '^TestJitterTickerStop' -count=20
go vet ./chain
```

### Testing

### DIFF
--- a/chain/jitter.go
+++ b/chain/jitter.go
@@ -4,11 +4,15 @@ import (
 	"errors"
 	"math"
 	"math/rand"
+	"sync"
 	"time"
 )
 
 // JitterTicker is a ticker that adds jitter to the tick duration.
 type JitterTicker struct {
+	// stopOnce ensures the quit channel is closed at most once.
+	stopOnce sync.Once
+
 	// C is a read-only channel that receives ticks.
 	C <-chan time.Time
 
@@ -80,7 +84,9 @@ func calculateMinMax(d time.Duration, scaler float64) (int64, int64) {
 
 // Stop stops the ticker.
 func (jt *JitterTicker) Stop() {
-	close(jt.quit)
+	jt.stopOnce.Do(func() {
+		close(jt.quit)
+	})
 }
 
 // start starts the ticker.
@@ -103,10 +109,9 @@ func (jt *JitterTicker) start() {
 			}
 
 		case <-jt.quit:
-			// Stop the timer and clean the channel when it stops.
-			if !timer.Stop() {
-				<-timer.C
-			}
+			// Stop the timer before exiting the ticker goroutine.
+			timer.Stop()
+			return
 		}
 	}
 }
--- a/chain/jitter_test.go
+++ b/chain/jitter_test.go
@@ -96,3 +96,76 @@ func TestJitterTicker(t *testing.T) {
 		require.True(t, diff < 121*time.Millisecond, "diff: %v", diff)
 	}
 }
+
+// newTestJitterTicker starts a ticker and returns a channel that is closed
+// when the ticker's start goroutine exits.
+func newTestJitterTicker(d time.Duration) (*JitterTicker, <-chan struct{}) {
+	minimum, maximum := calculateMinMax(d, 0)
+	ticker := &JitterTicker{
+		c:        make(chan time.Time, 1),
+		duration: d,
+		min:      minimum,
+		max:      maximum,
+		quit:     make(chan struct{}),
+	}
+	ticker.C = (<-chan time.Time)(ticker.c)
+
+	done := make(chan struct{})
+	go func() {
+		ticker.start()
+		close(done)
+	}()
+
+	return ticker, done
+}
+
+func TestJitterTickerStop(t *testing.T) {
+	t.Parallel()
+
+	ticker, done := newTestJitterTicker(time.Hour)
+	ticker.Stop()
+
+	select {
+	case <-done:
+	case <-time.After(time.Second):
+		t.Fatal("ticker start goroutine did not exit")
+	}
+}
+
+func TestJitterTickerStopIdempotent(t *testing.T) {
+	t.Parallel()
+
+	ticker, done := newTestJitterTicker(time.Hour)
+
+	require.NotPanics(t, func() {
+		ticker.Stop()
+		ticker.Stop()
+	})
+
+	select {
+	case <-done:
+	case <-time.After(time.Second):
+		t.Fatal("ticker start goroutine did not exit")
+	}
+}
+
+func TestJitterTickerStopConcurrentWithTick(t *testing.T) {
+	t.Parallel()
+
+	// A zero duration keeps the timer ready while Stop is called, exercising
+	// the select interleaving between a timer tick and the quit signal.
+	ticker, done := newTestJitterTicker(0)
+	select {
+	case <-ticker.C:
+	case <-time.After(time.Second):
+		t.Fatal("ticker did not produce a tick")
+	}
+
+	ticker.Stop()
+
+	select {
+	case <-done:
+	case <-time.After(time.Second):
+		t.Fatal("ticker start goroutine did not exit")
+	}
+}


Lrifton92 (commented):
Nice catch — I reproduced this locally and the diagnosis holds exactly as described.

For the record, the mechanism. Instrumenting the real `start()` on master
(`eee8e99`), driven through `NewJitterTicker(10*time.Millisecond, 0)`, then
calling `Stop()` after a few ticks:

```
tour 6 : timer.Stop() = true      <- quit branch: no drain, and no return
tour 7 : timer.Stop() = false     <- quit is closed, so it is selected again
  -> enters <-timer.C  (blocking)
```

The goroutine never comes back out of that receive — I let it sit for 3s and the
statement after the drain never printed. Two loop iterations, then a permanent
block.

Both `JitterTicker` call sites in the package take the ticker with
`defer ticker.Stop()` — `chain/bitcoind_rpc_events.go:270` and
`chain/bitcoind_zmq_events.go:454` — so every poller teardown leaks one
`start()` goroutine plus the timer it holds, which accumulates across bitcoind
reconnects on a long-running node.

I also checked that the tests are real regression tests rather than tests that
merely pass. Applying only `chain/jitter_test.go` on top of unmodified master,
all four fail: `TestJitterTickerStopIdempotent` panics at `jitter.go:83`
(`close of closed channel`) and the other three time out on "ticker start
goroutine did not exit". With the full PR,
`go test ./chain -run '^TestJitterTickerStop' -count=100` and `go vet ./chain`
are clean. I could not run `-race` here (no cgo toolchain on this machine), so
that part is on CI.

Three notes below, none of them blocking.

---

**1. With the `return` in place, the drain is dead code in this module**

`go.mod` declares `go 1.25.11`, so the Go 1.23 timer semantics apply: timer
channels are unbuffered and "no stale values prepared before [a `Stop`] call
will be sent or received after the call".

That changes which branch is reachable here, and not in the direction I first
assumed. When the timer has fired and nobody has received from `timer.C`,
`Stop()` returns **`true`**, not `false`:

```
len(timer.C)=0 cap(timer.C)=0
timer.Stop() = true
```

which is also what `tour 6` above shows. So once the second iteration is
eliminated by the `return`, the only way to reach `!timer.Stop()` is a timer
that was already stopped — and in that case there is provably nothing left to
receive:

```
1st Stop() = true
2nd Stop() = false
non-blocking drain: nothing to receive (default taken)
```

In other words the `select { case <-timer.C: default: }` can never take its
first case under this module's Go directive; it only does anything under
`GODEBUG=asynctimerchan=1`. `timer.Stop()` followed by `return` would be
equivalent and matches the post-1.23 idiom. I am happy either way — but if you
keep the drain, a one-line comment saying it is there for the
`asynctimerchan=1` fallback would save the next reader the same investigation.

**2. `TestJitterTickerStopNoTicks` asserts a property that does not hold**

The test uses `time.Hour`, so the timer can never have fired and the
`select ... default` on `ticker.C` is vacuous — it would pass against any
implementation. Its only real content is the `done` assertion, which duplicates
`TestJitterTickerStop`.

That matters because the property the name claims is actually false. `jt.c` is
buffered with capacity 1 and nothing drains it, so a tick produced before
`Stop()` stays readable afterwards. With the PR applied, a 5ms duration and a
50ms sleep before stopping:

```
RESULT: stale tick readable after Stop() : 2026-08-19 11:54:17.9475178 +0200
```

Two options, and I do not have a strong opinion on which is right for this
type: either make the test honest (short duration, let a tick land, `Stop()`,
then assert whatever the intended contract is), or decide that `JitterTicker`
should match `time.Ticker`'s post-1.23 guarantee and drain `jt.c` on shutdown.
The second is a behaviour change and probably out of scope here, but it seems
worth being explicit about which contract you intend, since the test name
currently promises the stronger one.

**3. Minor: keep the test helper coupled to the constructor**

`newTestJitterTicker` hardcodes `min: int64(d), max: int64(d)`, which is what
`calculateMinMax(d, 0)` returns today. Calling `calculateMinMax(d, 0)` directly
would keep the helper from silently drifting if that function ever changes. (I
realise the helper cannot simply call `NewJitterTicker`, since it needs to wrap
`start()` to observe the goroutine exit — that duplication looks justified.)

Also `TestJitterTickerStopConcurrentWithTick` passes `0` as the duration, so
`rand()` returns 0 and the timer refires immediately; between the first tick and
`Stop()` the goroutine spins. It is short-lived and harmless, but the tick/quit
interleaving the name suggests is opportunistic rather than forced — worth a
comment, or a small non-zero duration.

---

Unrelated to the code: your checklist has "lines wrap at 80" and "Ideal Git
Commit Structure" left unticked, but the PR already satisfies both — the widest
added line is 76 columns with tabs expanded, and it is a single commit with a
conventional `chain: ...` headline. You can tick them.

zloglevel:
> Nice catch — I reproduced this locally and the diagnosis holds exactly as described.
> 
> For the record, the mechanism. Instrumenting the real `start()` on master (`eee8e99`), driven through `NewJitterTicker(10*time.Millisecond, 0)`, then calling `Stop()` after a few ticks:
> 

@Lrifton92 
Thanks for the detailed review. I updated the PR accordingly:

- Removed the unnecessary timer channel drain and now stop the timer before returning from start().
- Removed TestJitterTickerStopNoTicks because jt.c is buffered and an already queued tick may remain readable after Stop().
- Updated newTestJitterTicker to derive min and max through calculateMinMax(d, 0).
- Added a comment explaining why the zero-duration timer is used in the concurrent stop test.

yyforyongyu:
Approved CI run

zloglevel:
> Approved CI run

@yyforyongyu @Lrifton92 I fixed the CI lint errors by adding t.Parallel() to the tests and renaming the variables that conflicted with predeclared identifiers. Could you please re-run the CI checks and take another look at the updated PR?

Lrifton92:
Re-reviewed the updated PR. My position is unchanged: the lint fixes do not touch anything the review was about.

I diffed the two revisions rather than re-reading the whole thing — `chain/jitter.go` is **byte-identical** between `028e9dd` and `ff575bd`, so the substantive fix (dropping the drain, stopping the timer before returning from `start()`) is exactly what I looked at before. The only delta is in `chain/jitter_test.go`: `min`/`max` renamed to `minimum`/`maximum`, and `t.Parallel()` added to `TestJitterTickerStop`, `TestJitterTickerStopIdempotent` and `TestJitterTickerStopConcurrentWithTick`.

No objection to the `t.Parallel()` additions. I did look at whether the zero-duration ticker was a problem there, since its `start()` goroutine spins hot while it is alive and the two sibling tests use one-second deadlines — but the spin window is only the few milliseconds between `newTestJitterTicker(0)` and `Stop()`, so it is not a realistic flake source.

On re-running CI: that one is not mine to give. The run on `ff575bd` is sitting in `action_required`, so the workflow approval from the previous head did not carry over — it needs a maintainer again. @yyforyongyu.

zloglevel:
> Re-reviewed the updated PR. My position is unchanged: the lint fixes do not touch anything the review was about.
> 
> I diffed the two revisions rather than re-reading the whole thing — `chain/jitter.go` is **byte-identical** between `028e9dd` and `ff575bd`, so the substantive fix (dropping the drain, stopping the timer before returning from `start()`) is exactly what I looked at before. The only delta is in `chain/jitter_test.go`: `min`/`max` renamed to `minimum`/`maximum`, and `t.Parallel()` added to `TestJitterTickerStop`, `TestJitterTickerStopIdempotent` and `TestJitterTickerStopConcurrentWithTick`.
> 
> No objection to the `t.Parallel()` additions. I did look at whether the zero-duration ticker was a problem there, since its `start()` goroutine spins hot while it is alive and the two sibling tests use one-second deadlines — but the spin window is only the few milliseconds between `newTestJitterTicker(0)` and `Stop()`, so it is not a realistic flake source.
> 
> On re-running CI: that one is not mine to give. The run on `ff575bd` is sitting in `action_required`, so the workflow approval from the previous head did not carry over — it needs a maintainer again. @yyforyongyu.

@yyforyongyu Hi, Could you please review this PR at your convenience?

zloglevel:
@Abdulkbk @yyforyongyu @Lrifton92  Thanks for your review!  If there's anything need I to do, please feel free to let me know.

PR #1338, opened 2026-08-19
wallet/txsizes: fix virtual transaction size accounting

GustavoStingelin:
Corrects `CompactSize` output counting and mixed legacy/witness input accounting in `EstimateVirtualSize`. 
Adds serialization-based regression tests for boundary cases.

Part of btcwallet SQL roadmap task 355 (https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/355-transaction-size-accounting.md).


wallet/txsizes/size_serialization_test.go
```diff
@@ -0,0 +1,132 @@
...
+					),
+				})
+			}
+
+			actual := EstimateVirtualSize(
+				test.p2pkhIns, test.p2trIns, test.p2wpkhIns,
+				test.nestedP2WPKHIns, estimateOutputs, test.changeScriptSize,
+			)
+			expected := serializedVirtualSize(tx)
+			if actual != expected {
```
yyforyongyu: we should use `require.Equal` here

wallet/txsizes/size.go
```diff
@@ -212,28 +213,23 @@ func EstimateVirtualSize(numP2PKHIns, numP2TRIns, numP2WPKHIns, numNestedP2WPKHI
 		outputCount++
 	}
 
-	// Version 4 bytes + LockTime 4 bytes + Serialized var int size for the
-	// number of transaction inputs and outputs + size of redeem scripts +
-	// the size out the serialized outputs and change.
+	// baseSize includes the 4-byte version and 4-byte locktime.
```
yyforyongyu: I don't think we should delete the old godoc? and given this is a foundamental pkg, we need detailed docs here explaining the calculation.

wallet/txsizes/size.go
```diff
@@ -198,10 +198,11 @@ func EstimateSerializeSize(inputCount int, txOuts []*wire.TxOut, addChangeOutput
 // EstimateVirtualSize returns a worst case virtual size estimate for a
 // signed transaction that spends the given number of P2PKH, P2TR, P2WPKH and
 // (nested) P2SH-P2WPKH outputs, and contains each transaction output
-// from txOuts. The estimate is incremented for an additional P2PKH
-// change output if addChangeOutput is true.
+// from txOuts. A positive changeScriptSize adds a change output with a script
+// of that size.
 func EstimateVirtualSize(numP2PKHIns, numP2TRIns, numP2WPKHIns, numNestedP2WPKHIns int,
 	txOuts []*wire.TxOut, changeScriptSize int) int {
+	inputCount := numP2PKHIns + numP2TRIns + numP2WPKHIns + numNestedP2WPKHIns
```
yyforyongyu, line 219: add a newline above?

wallet/txsizes/size.go
```diff
@@ -198,10 +198,11 @@ func EstimateSerializeSize(inputCount int, txOuts []*wire.TxOut, addChangeOutput
 // EstimateVirtualSize returns a worst case virtual size estimate for a
 // signed transaction that spends the given number of P2PKH, P2TR, P2WPKH and
 // (nested) P2SH-P2WPKH outputs, and contains each transaction output
-// from txOuts. The estimate is incremented for an additional P2PKH
-// change output if addChangeOutput is true.
+// from txOuts. A positive changeScriptSize adds a change output with a script
```
yyforyongyu: let's also add to the commit msg body about the issue it fixed given this is rather an important fix

wallet/txsizes/size_serialization_test.go
```diff
@@ -0,0 +1,132 @@
...
+	return wire.NewTxIn(
+		&wire.OutPoint{}, bytes.Repeat([]byte{0xff}, signatureScriptSize),
+		witness,
+	)
+}
+
+// serializedVirtualSize returns the BIP 141 vsize of tx: its weight, which
+// counts every base byte four times and every witness byte once, divided by
+// the witness scale factor and rounded up.
+func serializedVirtualSize(tx *wire.MsgTx) int {
```
yyforyongyu: think we already have `mempool.GetTxVirtualSize(btcutil.NewTx(tx))`

yyforyongyu (changes requested):
Cool pending a rebase, plus a few comments re docs.

wallet/txsizes/size_serialization_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+	tests := []struct {
+		name             string
+		p2pkhIns         int
+		p2trIns          int
+		p2wpkhIns        int
+		nestedP2WPKHIns  int
+		outputCount      int
+		changeScriptSize int
+	}{
```
yyforyongyu, line 37: The table only contains inputs, so the expected contract for each case is not visible. Please add a `wantVSize` field with the expected value for every row, compute `got := int64(EstimateVirtualSize(...))`, validate the constructed transaction with `mempool.GetTxVirtualSize(btcutil.NewTx(tx))`, and use `require.Equal` to compare both results with `test.wantVSize`.

wallet/txsizes/size_serialization_test.go
```diff
@@ -0,0 +1,132 @@
...
+// the witness scale factor and rounded up.
+func serializedVirtualSize(tx *wire.MsgTx) int {
+	weight := tx.SerializeSizeStripped()*3 + tx.SerializeSize()
+	return (weight + blockchain.WitnessScaleFactor - 1) /
+		blockchain.WitnessScaleFactor
+}
+
+// TestEstimateVirtualSizeSerializationAccounting verifies that size estimates
+// match the BIP 141 virtual size of serialized transactions.
+func TestEstimateVirtualSizeSerializationAccounting(t *testing.T) {
```
yyforyongyu: This test currently combines the CompactSize output boundary, repeated witness inputs, mixed legacy/witness inputs, and Taproot coverage under one broad contract. Please split it into focused tests such as `TestEstimateVirtualSizeOutputCount` and `TestEstimateVirtualSizeWitnessAccounting`, keeping each table limited to one behavior.

wallet/txsizes/size_serialization_test.go
GustavoStingelin: It's a separate module that does not have the testify dependency yet. Should we add it?

wallet/txsizes/size_serialization_test.go
GustavoStingelin: The test is comparing the serialized transaction size with the result of `EstimateVirtualSize`. Why should we use `GetTxVirtualSize` from the transaction if you are proposing to add a `wantVSize` value to the test cases? Those options look mutually exclusive to me.

Maybe `GetTxVirtualSize` should have its own table test with explicit expected `wantVSize` values, since it currently has no direct tests, even on the btcd master branch.

wallet/txsizes/size_serialization_test.go
GustavoStingelin: added

wallet/txsizes/size.go
GustavoStingelin: rephrased the comments

wallet/txsizes/size.go
GustavoStingelin: added

wallet/txsizes/size.go
GustavoStingelin: added

wallet/txsizes/size_serialization_test.go
GustavoStingelin: changed to use the GetTxVirtualSize from btcd mempool pkg

wallet/txsizes/size_serialization_test.go
GustavoStingelin: splited

yyforyongyu (approved):
LGTM!

PR #1340, opened 2026-08-22
build: preserve semantic version separators

pucedoteth:
btcd hit this in its own copy of `version.go` and fixed it in [65db493](https://github.com/btcsuite/btcd/commit/65db49397a59fb10fd5a954df19bda33ce4e5e5d) ("version: preserve semantic version separators", btcsuite/btcd#2578): SemVer separates pre-release identifiers and build metadata with periods, but `semanticAlphabet` omitted the period, so `normalizeVerString` silently collapsed `beta.rc1` into `betarc1`.

`btcwallet/build/version.go` carries the same code and still has the old alphabet:

```
btcd/version.go:14           ...uvwxyz-."
btcwallet/build/version.go:14 ...uvwxyz-"
```

btcd's test, run unchanged against this copy, fails:

```
--- FAIL: TestNormalizeVerString
    version_test.go:25: unexpected normalized version: got "betarc1", want "beta.rc1"
```

This is reachable rather than latent. `normalizeVerString` is applied to `appBuild`, which the comment above it documents as settable at build time:

```go
// appBuild is defined as a variable so it can be overridden during the build
// process with '-ldflags "-X main.appBuild foo' if needed.
```

SemVer build metadata is dot-separated — `exp.sha.5114f85` is an example from the spec itself — so a btcwallet built with that stamp reports `0.15.1-alpha+expsha5114f85`, with the separators dropped and the metadata no longer round-tripping.

The change is the one-character fix from 65db493 plus that commit's test, with a build metadata case added for the `appBuild` path.

`go build ./...`, `go test ./build/`, `gofmt` and `go vet` are clean.

Two things I noticed but left alone to keep the diff to the fix:

- That doc comment says `-X main.appBuild`, but this copy lives in package `build`, so the flag is really `-X github.com/btcsuite/btcwallet/build.appBuild`. It was accurate before the code moved out of `main`.
- btcsuite/btcd#2596 applies the same fix to `btcd/cmd/btcctl/version.go`, which is a third copy of this file that was also missed.

---

*Disclosure: written with AI assistance; the analysis, fix and test were verified locally.*

### DIFF
--- a/build/version.go
+++ b/build/version.go
@@ -11,7 +11,7 @@ import (
 )
 
 // semanticAlphabet
-const semanticAlphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-"
+const semanticAlphabet = "0123456789ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz-."
 
 // These constants define the application version and follow the semantic
 // versioning 2.0.0 spec (http://semver.org/).
@@ -21,13 +21,19 @@ const (
 	appPatch uint = 1
 
 	// appPreRelease MUST only contain characters from semanticAlphabet
-	// per the semantic versioning spec.
+	// per the semantic versioning spec.  That is necessary but not
+	// sufficient: semanticAlphabet is a per-character filter and cannot
+	// reject a malformed arrangement of valid characters, such as a
+	// trailing or doubled '.', so the value must also be a well-formed
+	// dot-separated list of non-empty identifiers.
 	appPreRelease = "alpha"
 )
 
 // appBuild is defined as a variable so it can be overridden during the build
 // process with '-ldflags "-X main.appBuild foo' if needed.  It MUST only
-// contain characters from semanticAlphabet per the semantic versioning spec.
+// contain characters from semanticAlphabet per the semantic versioning spec,
+// and like appPreRelease must also be a well-formed dot-separated list of
+// non-empty identifiers, which the per-character filter cannot enforce.
 var appBuild string
 
 // Version returns the application version as a properly formed string per the
--- a/build/version_test.go
+++ b/build/version_test.go
@@ -1,0 +1,46 @@
+// Copyright (c) 2026 The btcsuite developers
+// Use of this source code is governed by an ISC
+// license that can be found in the LICENSE file.
+
+package build
+
+import "testing"
+
+// TestNormalizeVerString ensures semantic version separators are preserved
+// while characters outside the alphabet are still stripped.
+func TestNormalizeVerString(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+		in   string
+		want string
+	}{
+		{
+			name: "dotted pre-release identifier",
+			in:   "beta.rc1",
+			want: "beta.rc1",
+		},
+		{
+			// appBuild may be set to a dotted value via -ldflags at
+			// build time.
+			name: "dotted build metadata",
+			in:   "exp.sha.5114f85",
+			want: "exp.sha.5114f85",
+		},
+		{
+			// Guards against the filter being dropped entirely: an
+			// identity function satisfies the cases above.
+			name: "character outside the alphabet",
+			in:   "beta!rc1",
+			want: "betarc1",
+		},
+	}
+
+	for _, test := range tests {
+		if got := normalizeVerString(test.in); got != test.want {
+			t.Fatalf("%s: normalizeVerString(%q) = %q, want %q",
+				test.name, test.in, got, test.want)
+		}
+	}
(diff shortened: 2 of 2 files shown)


Lrifton92 (commented):
Correct fix, and there is a stronger argument for it than the PR description makes: btcwallet is the odd one out in its own family. Both sibling projects already allow the dot —

- `btcsuite/btcd` `version.go:14`
- `lightningnetwork/lnd` `build/version.go:38`

both use `"0123456789...xyz-."`, identical to what this PR lands. So this is not a behaviour change so much as btcwallet catching up, which also means `-X build.appBuild=exp.sha.5114f85` currently produces a different version string in btcwallet than the exact same flag does in btcd. Worth putting in the commit message.

### The test cannot fail if the filter is deleted

```go
tests := []string{"beta.rc1", "exp.sha.5114f85"}
for _, version := range tests {
    if got := normalizeVerString(version); got != version {
```

Every case asserts `normalizeVerString(x) == x`, which is equally true of an identity function. Removing the `strings.ContainsRune` check entirely — or dropping the whole loop and returning `str` — passes this test unchanged. It pins the new half of the behaviour and leaves the old half unguarded.

One invalid-character case fixes that and costs a line, something like `{in: "beta!rc1", want: "betarc1"}` alongside the pass-through ones. Then the test distinguishes "dots survive" from "nothing is filtered", which is the actual claim.

### Minor, and inherited rather than introduced

Allowing `.` in a character filter also lets malformed input through: `appBuild = "beta."` now yields `0.15.1+beta.`, and `".."` yields `0.15.1+..`, neither of which is a valid semver string — where previously both were stripped to empty and the suffix was dropped. SemVer requires dot-*separated* identifiers that are each non-empty, and a per-character allowlist cannot express that.

I would not hold the PR for it, because btcd and lnd have exactly the same naive filter with no further validation, so tightening here would make btcwallet inconsistent again in the other direction. But since `appBuild` is set from build scripts via `-ldflags`, a typo now propagates into `Version()` instead of being neutralised. If you want to acknowledge it cheaply, the comments on `appPreRelease` and `appBuild` currently say the value "MUST only contain characters from semanticAlphabet per the semantic versioning spec" — that is now necessary but no longer sufficient, and saying so would save the next reader the same detour.

pucedoteth:
All three addressed — thanks, the second one was a real hole.

**Test could not fail if the filter were removed.** Correct, and I verified it: every case I wrote asserted `normalizeVerString(x) == x`, which an identity function satisfies. The test is now table-driven with your `beta!rc1` → `betarc1` case alongside the pass-through ones.

I mutation-tested the result rather than eyeballing it. Replacing the body with `return str` only proves a build failure (unused imports), so I neutralised the check in place instead — `if true || strings.ContainsRune(...)` — which still compiles:

```
--- FAIL: TestNormalizeVerString
    version_test.go:42: character outside the alphabet: normalizeVerString("beta!rc1") = "beta!rc1", want "betarc1"
```

Restored, it passes. Both halves are now pinned.

**btcwallet is the family outlier.** Confirmed and moved into the commit message: `btcd/version.go:14` and `lnd/build/version.go:38` are both already `"…xyz-."`, so the same `-ldflags` stamp yields a different version string in btcwallet than in btcd today. That framing is better than mine — it is catching up, not changing behaviour.

**Per-character filter cannot express dot-separated identifiers.** Also confirmed:

```
normalizeVerString("beta.") = "beta."
normalizeVerString("..")    = ".."
```

Both previously stripped to empty and dropped the suffix. I took the cheap option you suggested and left the validation alone, since btcd and lnd carry the identical naive filter and tightening only here would re-introduce the divergence in the other direction. The comments on `appPreRelease` and `appBuild` now say the alphabet is necessary but not sufficient, and that the value must also be a well-formed dot-separated list of non-empty identifiers.

`go build ./...`, `go test ./build/`, `gofmt` and `go vet` are clean.

Lrifton92 (approved):
The three points from my previous review are addressed on `a827f70`, and I checked each one against the diff rather than the description:

1. `semanticAlphabet` now carries the dot, matching `btcd` `version.go:14` and `lnd` `build/version.go:38` — so `-X build.appBuild=exp.sha.5114f85` finally yields the same string across the three projects.
2. The comments on `appPreRelease` and `appBuild` now state the real invariant: a per-character filter is necessary but not sufficient, since it cannot reject a malformed arrangement of valid characters (trailing or doubled dot). That is the right place for it — callers setting these via ldflags are the ones who have to keep the value well-formed.
3. The test is table-driven and keeps the negative case (`beta!rc1` -> `betarc1`), so it still fails if the filter is dropped entirely. Without it, an identity function would pass.

One thing left open on purpose, and I do not think it should block: nothing here rejects `beta..rc1` or a trailing dot, so `Version()` can still emit a string that is not valid semver. That is a wider contract change than this PR, and the new comments say so explicitly.

LGTM.

PR #1341, opened 2026-08-22
wallet+db: resolve postgres secret commits

yyforyongyu:
## Summary

Resolve roadmap Task 421 at the backend boundary that can produce an uncertain wallet-secret commit: the PostgreSQL Store.

## Why this belongs in PostgreSQL

ChangePassphrase relies on the Store.UpdateWalletSecrets contract: return nil when the update persisted and an error when it failed. Callers should not need to understand database transport failures or reconcile transaction state themselves.

SQLite commits in process and does not surface runtime.ErrAmbiguousTxCommit. PostgreSQL can lose the connection while waiting for the COMMIT result, leaving the client unable to tell whether the server persisted the update. The PostgreSQL runtime reports only that lost acknowledgement as an ambiguous commit.

The PostgreSQL Store therefore settles that backend-specific condition before returning. This keeps the database abstraction intact, avoids adding reconciliation behavior to unaffected SQLite and keyvault call sites, and ensures ChangePassphrase still receives the binary Store result it expects.

## Change Description

When PostgreSQL UpdateWalletSecrets receives runtime.ErrAmbiguousTxCommit, perform one detached, bounded read of the persisted wallet-secret row. Return success only when all four stored fields exactly match the attempted new tuple. If the read fails or the row differs, return an ordinary PostgreSQL Store error that retains the underlying commit and verification failures without leaking the runtime ambiguity marker.

This intentionally refines the task specification suggested implementation. It does not add WalletVault old/new/indeterminate states, mutation retries, generic reconciliation machinery, or SQLite handling. The Task 421 durability issue is instead resolved at its sole source while preserving the existing Store API.

## Verification

Focused PostgreSQL Store unit tests cover ordinary success and failure, exact persisted matches, partial mismatches, unavailable reads, caller cancellation, bounded verification, and secret-buffer cleanup. Existing keyvault tests and the UpdateWalletSecrets integration cases pass for both SQLite and PostgreSQL.

Implements roadmap Task 421

### DIFF
--- a/wallet/internal/db/pg/walletstore_updatewalletsecrets.go
+++ b/wallet/internal/db/pg/walletstore_updatewalletsecrets.go
@@ -1,24 +1,132 @@
 package pg
 
 import (
+	"bytes"
 	"context"
 	"database/sql"
 	"errors"
 	"fmt"
+	"time"
 
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	dbruntime "github.com/btcsuite/btcwallet/wallet/internal/db/runtime"
 	"github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
 )
 
-// UpdateWalletSecrets updates the secrets for the wallet.
+// walletSecretsCommitVerificationTimeout bounds the independent read used when
+// PostgreSQL loses the database connection after sending COMMIT. The server may
+// already have made the transaction durable even though the client missed its
+// acknowledgement, so a short limit keeps verification through another pooled
+// connection from stalling the caller.
+const walletSecretsCommitVerificationTimeout = 5 * time.Second
+
+// UpdateWalletSecrets updates the wallet secrets through one PostgreSQL
+// transaction. If the database connection is lost while awaiting the COMMIT
+// acknowledgement, the Store never retries the mutation; it performs one
+// detached, bounded read through another pooled connection and returns success
+// only when every attempted field is visible. A failed read or mismatch returns
+// an ordinary Store error without exposing the runtime ambiguity marker.
 func (s *Store) UpdateWalletSecrets(ctx context.Context,
 	params db.UpdateWalletSecretsParams) error {
 
-	return s.execWrite(ctx, func(qtx *sqlc.Queries) error {
+	err := s.execWrite(ctx, func(qtx *sqlc.Queries) error {
 		return db.UpdateWalletSecretsWithOps(
 			ctx, params, updateWalletSecretsOps{q: qtx},
 		)
 	})
+
+	return verifyWalletSecretsCommit(
+		ctx, params, walletSecretsCommitVerificationTimeout, err,
+		s.GetWalletSecrets,
+	)
+}
+
+// verifyWalletSecretsCommit applies the binary Store contract to a completed
+// PostgreSQL write. Definite outcomes pass through unchanged; a lost commit
+// reply triggers one bounded read that proves success only from an exact match
+// with the attempted persisted tuple.
+func verifyWalletSecretsCommit(ctx context.Context,
+	params db.UpdateWalletSecretsParams, verificationTimeout time.Duration,
+	updateErr error, getWalletSecrets func(context.Context,
+		uint32) (*db.WalletSecrets, error)) error {
+
+	if updateErr == nil ||
+		!errors.Is(updateErr, dbruntime.ErrAmbiguousTxCommit) {
+
+		return updateErr
+	}
+
+	// Detach verification from the failed mutation context because a canceled
+	// connection must not prevent another pooled connection from observing the
+	// durable row. The explicit timeout keeps that independent read bounded.
+	verificationCtx, cancel := context.WithTimeout(
+		context.WithoutCancel(ctx), verificationTimeout,
+	)
+	defer cancel()
+
+	persisted, readErr := getWalletSecrets(
+		verificationCtx, params.WalletID,
+	)
+	if persisted != nil {
+		// The Store owns this verification-only read, so destroy every returned
+		// credential buffer after comparison rather than leaking a second copy.
+		defer clearWalletSecrets(persisted)
+	}
+
+	commitErr := walletSecretsCommitError(updateErr)
+	if readErr != nil {
+		return errors.Join(
+			commitErr,
+			fmt.Errorf("verify wallet secrets after commit: %w", readErr),
+		)
+	}
+
+	if walletSecretsMatchUpdate(persisted, params) {
+		return nil
+	}
+
+	return fmt.Errorf("verify wallet secrets after commit: "+
+		"persisted values differ: %w", commitErr)
+}
+
+// walletSecretsMatchUpdate requires all verifier and ciphertext fields to
+// equal the attempted PostgreSQL update, preventing a partial or unrelated
+// row from converting an uncertain commit into success.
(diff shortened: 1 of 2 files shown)


wallet/internal/db/itest/walletvault_passphrase_behavior_test.go
```diff
@@ -0,0 +1,117 @@
+//go:build itest
+
```
yyforyongyu: these new behavior tests are out of scope?

wallet/internal/keyvault/keyvault.go
```diff
@@ -19,6 +19,44 @@ var ErrVaultLocked = errors.New("vault is locked")
 // vault was already unlocked.
 var ErrVaultUnlocked = errors.New("vault is already unlocked")
 
+// ErrPassphraseChangeNotCommitted reports that reconciliation proved the
+// persisted wallet-secret tuple still contains the exact pre-rotation facts.
+// Callers may safely report this as a definitive failed change rather than an
+// ambiguous SQL commit.
+var ErrPassphraseChangeNotCommitted = errors.New(
```
yyforyongyu: call it `ErrCommitPassphrase`?

wallet/internal/keyvault/keyvault.go
```diff
@@ -19,6 +19,44 @@ var ErrVaultLocked = errors.New("vault is locked")
...
+	"vault passphrase change did not commit",
+)
+
+// PassphraseChangeIndeterminateError reports that an ambiguous passphrase
+// commit could not be resolved to exactly the old or new credential tuple.
+// Its wrapped causes retain the original commit error and the reconciliation
+// failure for diagnostics, while errors.As lets the wallet translate this
+// keyvault-owned result without another Store read.
+type PassphraseChangeIndeterminateError struct {
```
yyforyongyu: call it `ErrChangePassphrase`? or `ChangePassphraseError`?

wallet/internal/keyvault/vault_lifecycle.go
```diff
@@ -259,15 +347,72 @@ func (v *WalletVault) ChangePassphrase(ctx context.Context,
 			"rotate secrets: %w", v.walletID, err)
 	}
 
-	err = v.store.UpdateWalletSecrets(ctx, updateParams)
-	if err != nil {
+	newFacts := walletSecretFactsFromUpdate(updateParams)
+	defer newFacts.zero()
+
+	err = v.store.UpdateWalletSecrets(ctx, newFacts.updateParams(v.walletID))
```
yyforyongyu: isn't this a db tx already? why do we need to reconcile? and it is not reconciling strictly speaking? as we just compare new-vs-old and return an error?

wallet/internal/keyvault/vault_lifecycle.go
```diff
@@ -259,15 +347,72 @@ func (v *WalletVault) ChangePassphrase(ctx context.Context,
...
+	if err == nil {
+		return nil
+	}
+
+	if !errors.Is(err, dbruntime.ErrAmbiguousTxCommit) {
 		return fmt.Errorf("wallet %d vault ChangePassphrase: "+
 			"update secrets: %w", v.walletID, err)
 	}
 
+	err = v.reconcilePassphraseCommit(ctx, err, &oldFacts, &newFacts)
```
yyforyongyu: ok i think this is the wrong layer to fix it - if we think `Commit` is unreliable, we should fix it there, but not breaking the abstraction provided by `Commit`.

wallet/internal/db/itest/walletvault_passphrase_behavior_test.go
yyforyongyu: Agreed. The added behavior itests were out of scope and have been removed. Coverage now lives in focused PostgreSQL Store unit tests, while the existing SQLite and PostgreSQL TestUpdateWalletSecrets integration test remains unchanged.

wallet/internal/keyvault/keyvault.go
yyforyongyu: Removed this sentinel entirely rather than renaming it. The reworked implementation adds no commit-specific keyvault API.

wallet/internal/keyvault/vault_lifecycle.go
yyforyongyu: Agreed on the layer. The keyvault comparison and reconciliation code has been removed. Only pg.Store.UpdateWalletSecrets handles the PostgreSQL ambiguous commit marker, using one bounded exact-row read to decide whether the attempted tuple persisted.

wallet/internal/keyvault/keyvault.go
yyforyongyu: Removed this exported error type entirely rather than renaming it. Keyvault now retains its existing error surface.

wallet/internal/keyvault/vault_lifecycle.go
yyforyongyu: Agreed that the shared transaction abstraction must not leak into keyvault. The shared runtime and keyvault call sites are unchanged; the narrow policy is contained in the PostgreSQL Store method that has the attempted row values needed to resolve a lost commit acknowledgment.

wallet/internal/db/pg/walletstore_updatewalletsecrets.go
```diff
@@ -1,26 +1,143 @@
...
 )
 
+// walletSecretsCommitVerificationTimeout limits the independent PostgreSQL
+// read used to confirm a wallet-secret update after the commit reply is lost.
+const walletSecretsCommitVerificationTimeout = 5 * time.Second
+
+// walletSecretsCommitOps isolates the PostgreSQL write and verification read
+// so the Store can keep commit uncertainty below the db.Store boundary while
+// focused tests exercise every policy branch without simulating SQL drivers.
+type walletSecretsCommitOps interface {
```
yyforyongyu: db Ops is meant to be created when it's shared by different db backends? here the only consumer is postgres so we don't need it?

yyforyongyu: Agreed. Removed the PostgreSQL-local Ops interface. Store.UpdateWalletSecrets now performs the single write directly and passes only the completed error plus its GetWalletSecrets method to the verifier; focused tests retain strict read and no-read expectations.

wallet/internal/db/pg/walletstore_updatewalletsecrets.go
```diff
@@ -1,24 +1,126 @@
...
+	"time"
 
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	dbruntime "github.com/btcsuite/btcwallet/wallet/internal/db/runtime"
 	"github.com/btcsuite/btcwallet/wallet/internal/sql/pg/sqlc"
 )
 
-// UpdateWalletSecrets updates the secrets for the wallet.
+// walletSecretsCommitVerificationTimeout limits the independent PostgreSQL
+// read used to confirm a wallet-secret update after the commit reply is lost.
```
yyforyongyu: add more docs explaining this hardening - it is for db conn lost.

yyforyongyu: Expanded the PostgreSQL Store and timeout documentation to explain that the database connection can be lost after COMMIT is sent, the transaction may already be durable despite the missing acknowledgement, verification uses one detached bounded read through another pooled connection, and the mutation is never retried.

Abdulkbk (approved):
LGTM

PR #1342, opened 2026-08-22
wallet+db: define account creation identity

yyforyongyu:
## Summary

Implements roadmap Task 387.

## Change Description

Define root-derived, arbitrary external, and path-bound external account
identity across ADR 0012 and ADR 0013. Record staged request ownership,
the retained ordinary-import boundary, and the NoChainSync policy
independently from wallet custody.

### DIFF
--- a/docs/developer/adr/0013-normalized-account-address-identity.md
+++ b/docs/developer/adr/0013-normalized-account-address-identity.md
@@ -20,26 +20,28 @@ looked like members of an account solely because SQL needed a non-null
 
 The recovery follow-up needs immutable SQL identity for imported xpub scan
 horizons. Account names and nullable account numbers are not suitable for that
-purpose: names can be renamed, and imported xpubs do not have BIP44 account
-numbers.
+purpose: names can be renamed, and unnumbered caller-supplied XPub accounts do
+not have account numbers.
 
 The word "imported" is overloaded across wallet layers. A wallet may be
 imported from a seed while its accounts remain derivable from that seed. An
-account xpub may be imported while its child addresses remain derivable from the
-xpub. A raw script or public-key address import is not derivable at all. The SQL
-schema therefore models whether a row has derivation identity, not where the
-user originally obtained the material.
+account XPub may be imported while its child addresses remain derivable from
+the XPub. A raw script or public-key address import is not derivable at all.
+The SQL schema therefore keeps account-key provenance separate from whether an
+address row has derivation identity.
 
 ## 2. Decision
 
 Normalize account and address identity around real persisted identity only:
 
-| Case | Parent | Number | Path |
+| Case | Parent | Number | Child path |
 | --- | --- | --- | --- |
-| Wallet-derived account | `accounts` row | BIP44 number | N/A |
-| Imported xpub account | `accounts` row | none | N/A |
-| Wallet-derived child address | derived account | has number | branch/index |
-| Imported-xpub child address | imported xpub account | none | branch/index |
+| Wallet-root-derived account | `accounts` row | required | N/A |
+| Caller-supplied XPub, unnumbered | `accounts` row | none | N/A |
+| Caller-supplied XPub, numbered | `accounts` row | required | N/A |
+| Root-derived child address | root account | required | branch/index |
+| Unnumbered-XPub child address | caller-supplied account | none | branch/index |
+| Numbered-XPub child address | caller-supplied account | required | branch/index |
 | Raw imported address | none | none | none; no scope |
 
 The SQL primitive is `is_derived` rather than `is_imported`. An imported-xpub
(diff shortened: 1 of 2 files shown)


docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -20,54 +20,144 @@ looked like members of an account solely because SQL needed a non-null
...
+Each branch must be derivable from a single account key. The supported forms
+are P2PKH (`PubKeyHash`), P2SH-P2WPKH (`NestedWitnessPubKey`), P2WPKH
+(`WitnessPubKey`), and key-path P2TR (`TaprootPubKey`); other forms are not
+account schemas.
+
+Equal normalized account-XPub payloads are admitted within or across scopes
+only when both accounts' effective branch schemas derive disjoint scripts.
+
+Existing schema fields, including `coin_pub_key`, `key_scope_secrets`, and all
+other columns, remain; this ADR prescribes no migration or source change.
```
GustavoStingelin: The sentence `this ADR prescribes no migration or source change` is broader than the constraint to retain `coin_pub_key`, `key_scope_secrets`, and the other existing columns. Supporting the third account shape still requires updating the unmerged account migrations to permit `is_derived = false` with a non-null `account_number`, along with the affected queries, comments, and validators. Could this instead say that no existing columns are removed and that the existing unmerged migrations and related source are updated for the third account shape?

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -20,54 +20,144 @@ looked like members of an account solely because SQL needed a non-null
...
+| `AddressSchema` | `*waddrmgr.ScopeAddrSchema` | optional |
+| `AccountNumber` | `*AccountNumber` | optional |
+| `AccountPubKey` | `*hdkeychain.ExtendedKey` | optional |
+| `MasterFingerprint` | `*MasterFingerprint` | optional |
+| `DiscoveryMode` | `DiscoveryMode` | required |
+
+Without `AccountPubKey`, an absent `AccountNumber` requests sequential
+wallet-root derivation and a present value requests the exact account,
+including account zero. A present `AccountPubKey` requires a present
+`AccountNumber` and creates a path-bound external account. An absent
```
GustavoStingelin: How should exact-number creation interact with `next_account_number` and the existing account-gap policy? If exact account `N` is inserted without advancing the counter, sequential creation eventually allocates `N`, hits the unique `(scope_id, account_number)` index, and rolls back the same transaction that incremented the counter; every retry can then repeat the collision. Please specify an atomic `max(counter, N+1)` rule, occupied-number behavior, maximum-number exhaustion, and whether exact requests are subject to the 100-unused-account restriction.

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -20,54 +20,144 @@ looked like members of an account solely because SQL needed a non-null
...
+
+| Field | Type | Presence |
+| --- | --- | --- |
+| `Name` | `string` | required |
+| `Scope` | `waddrmgr.KeyScope` | required |
+| `AddressSchema` | `*waddrmgr.ScopeAddrSchema` | optional |
+| `AccountNumber` | `*AccountNumber` | optional |
+| `AccountPubKey` | `*hdkeychain.ExtendedKey` | optional |
+| `MasterFingerprint` | `*MasterFingerprint` | optional |
+| `DiscoveryMode` | `DiscoveryMode` | required |
```
GustavoStingelin: `DiscoveryMode` is required here and described as immutable, but it has no enum values, zero-value behavior, validation rules, or persistence target, and this is its only occurrence in the repository. Please define its observable semantics and storage/recovery behavior, or remove it from the final request contract.

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -20,54 +20,144 @@ looked like members of an account solely because SQL needed a non-null
...
+assertion; any external/internal branch mismatch is a conflict and the request
+fails before mutation.
+
+Each branch must be derivable from a single account key. The supported forms
+are P2PKH (`PubKeyHash`), P2SH-P2WPKH (`NestedWitnessPubKey`), P2WPKH
+(`WitnessPubKey`), and key-path P2TR (`TaprootPubKey`); other forms are not
+account schemas.
+
+Equal normalized account-XPub payloads are admitted within or across scopes
+only when both accounts' effective branch schemas derive disjoint scripts.
```
GustavoStingelin: Could we make this admission rule executable? Define "normalized account-XPub payload" as an exact canonical tuple (script derivation depends on public key plus chain code, not version/depth/fingerprint/child metadata), define the wallet-local comparison domain, and give a branch/schema disjointness matrix. Because schema is persisted per key scope, equal payloads in one scope share the same effective schema and should be an explicit prohibition; across scopes the rule needs to state which external/internal combinations are disjoint.

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -20,54 +20,144 @@ looked like members of an account solely because SQL needed a non-null
...
+network and do not infer scope or schema. Discovery mode is explicit and
+immutable.
+
+#### Scope And Branch Schema
+
+An omitted address schema reuses an existing scope's persisted schema or the
+canonical schema for a standard scope. Creating a new custom scope requires an
+explicit schema. Supplying a schema for an existing scope is an equality
+assertion; any external/internal branch mismatch is a conflict and the request
+fails before mutation.
```
GustavoStingelin: This equality assertion requires behavior that the current scope-reuse path does not provide: `EnsureKeyScopeWithOps` returns an existing scope ID before inspecting `schemaOverride`, so a conflicting supplied schema is silently ignored. `CreateImportedAccount` forwards `AddrSchema` into that path, while the parameter comment calls it an override "for this imported account" even though the schema is persisted per scope. Please call out the existing-scope comparison before mutation and update the parameter contract to describe a per-scope assertion.

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -20,54 +20,144 @@ looked like members of an account solely because SQL needed a non-null
...
+`EnsureExactAccount` operation is introduced.
+
+`ImportAccount` carries name, scope, optional address schema, account XPub,
+optional master fingerprint, discovery mode, and dry-run. It creates only an
+arbitrary numberless external account, and dry-run performs the same validation
+without persistence.
+
+The caller-declared scope is authoritative for every supplied-XPub request.
+Extended-key version bytes are checked only for the configured Bitcoin
+network and do not infer scope or schema. Discovery mode is explicit and
```
GustavoStingelin: This reverses the current `ImportAccount` contract. Today it accepts `addrType` and calls `keyScopeFromPubKey`, which derives scope/schema from extended-key version bytes plus the address type; the proposed request instead carries authoritative `Scope`/`AddressSchema` and treats version bytes as a network check only. Please record that API and validation migration in the implementation notes, including the compatibility boundary, rather than leaving it under the adjacent "no migration or source change" statement.

docs/developer/adr/0013-normalized-account-address-identity.md
```diff
@@ -20,54 +20,144 @@ looked like members of an account solely because SQL needed a non-null
...
-| Case | Parent | Number | Path |
+| Case | Parent | Number | Child path |
 | --- | --- | --- | --- |
-| Wallet-derived account | `accounts` row | BIP44 number | N/A |
-| Imported xpub account | `accounts` row | none | N/A |
-| Wallet-derived child address | derived account | has number | branch/index |
-| Imported-xpub child address | imported xpub account | none | branch/index |
+| Wallet-root-derived account | `accounts` row | required | N/A |
+| Arbitrary external XPub account | `accounts` row | none | N/A |
+| Path-bound external XPub account | `accounts` row | required | N/A |
```
GustavoStingelin: Could we narrow these names to local descriptive terms? BIP 44 already uses `external` for the receive chain, so `external account` is easy to confuse with that branch.

Suggested labels:

- **Wallet-root-derived**
- **Caller-supplied XPub, unnumbered**
- **Caller-supplied XPub, numbered**

These labels describe the btcwallet account-key source and account-number presence; they are not BIP-defined account categories.

GustavoStingelin: Could the ADR make the root relationship explicit for `Caller-supplied XPub, numbered`?

A uniformly watch-only wallet may still have a master public key, but that public key cannot authenticate the hardened account ancestry of the supplied XPub. If the wallet controls the private root, Task 343 should derive the requested account from that root rather than import a supplied XPub. Otherwise, the Task 391 XPub is caller-supplied and is not required or proven to belong to the wallet root; its `account_number` is caller-declared metadata.

That leaves a namespace decision: two independent caller-supplied roots in one uniformly watch-only wallet can declare the same `(scope, account_number)`. Should the second be rejected as an occupied wallet/scope-global slot, require a separate wallet, or use a stronger root/provenance namespace?

My take is that `Caller-supplied XPub, numbered` should not consume or constrain wallet-root-derived account allocation.

GustavoStingelin (changes requested):
Could the ADR explicitly describe whether a Path-bound external XPub account can coexist in the same SQL wallet with a Wallet-root-derived account?

ADR 0012 appears to make the current public XPub-only path mutually exclusive with wallet-root-derived spendable accounts: supplied XPub accounts are accepted only in a uniformly watch-only wallet, while wallet-root-derived accounts come from generated or imported private root material. If that is the intended model, please state the mutual exclusion explicitly.

If same-wallet coexistence is intended, the ADR should instead define the custody path and numbered namespace. In particular, does (scope, account_number) identify one wallet-wide slot across roots and provenances, and does a caller-supplied numbered account consume a slot that would otherwise be allocated from the wallet root?

Even without wallet-root coexistence, two path-bound XPub accounts from independent roots can declare the same (scope, account_number). The ADR should specify whether the second account is rejected or whether another immutable selector distinguishes them.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Will make it explicit

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: I think that is an implementation detail that belongs to task 388, will provide a highlevel version tho as i don't wanna expand the ADR to be too long.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: cool yeah i like it, will change.

yyforyongyu: Not sure about `consume or constrain` part (you mean `(scope, account)` slot?) - but yeah, one is public, one is private wallet-root, so in no way can a public XPub consume the latter. So non-spendable wallet cannot coexist with a spendable wallet.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Updated this to preserve the existing columns while requiring account-shape constraints and affected validation/read paths to admit `is_derived=false` with a non-null account number.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Updated the ADR to use wallet-root-derived and caller-supplied XPub, unnumbered/numbered terminology.

yyforyongyu: Clarified that public-only caller-supplied and wallet-root-derived accounts cannot coexist under uniform SQL custody. Numbered supplied accounts share the scope-local namespace, so a duplicate `(scope, account_number)` is rejected.

docs/developer/adr/0013-normalized-account-address-identity.md
yyforyongyu: Defined normalization as public key plus chain code within one wallet, and clarified corresponding external/internal branch overlap and same-scope conflicts.

docs/developer/adr/0013-normalized-account-address-identity.md
GustavoStingelin: Yes, by “consume or constrain” I meant the (scope, account_number) slot.

GustavoStingelin (approved):
LGTM 👍🏼

PR #1344, opened 2026-08-22
chain: add live-tip-only watches

yyforyongyu:
## Summary

Implements roadmap Task 401.

Adds the context-aware `WatchAddrsFromTip` chain operation while keeping this
PR limited to the Wallet-owned interface and existing backend notification
mechanisms.

## Change Description

`NotifyReceived` is a live registration primitive for btcd, but not for every
backend. Neutrino can start a birthday-based scan or add addresses to an
already-active historical rescan, allowing matches from blocks before the
registration call. That is a correctness concern for callers that require a
strict live-tip boundary.

Btcd delegates to its existing receive-notification RPC without starting a
rescan. Both bitcoind event modes extend their existing in-memory live filter.
The shared conformance test proves those RPC-backed paths ignore a payment
mined before registration, tolerate repeated registration, deliver one future
payment, and remain quiet through a later block barrier.

Native Neutrino support is intentionally not implemented in this PR. Enforcing
a strict live-tip boundary safely requires a substantially larger rescan
lifecycle design, including coordination with birthday and active historical
rescans. That work is deferred to a dedicated follow-up because its size and
complexity are outside the deliberately minimal scope of this PR.

For interface compatibility only, Neutrino checks pre-call cancellation and
delegates to the unchanged `NotifyReceived` path. A `TODO(yy)` documents that
this temporary shim retains birthday-scan and active-rescan widening behavior
until the deferred Neutrino live-watch design is implemented.

### DIFF
--- a/chain/bitcoind_events_test.go
+++ b/chain/bitcoind_events_test.go
@@ -3,6 +3,7 @@ package chain
 import (
 	"fmt"
 	"os/exec"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -95,7 +96,6 @@ func runBitcoindEventsTests(t *testing.T, rpcPolling bool) {
 	}
 
 	for _, test := range tests {
-		test := test
 		t.Run(test.name, func(t *testing.T) {
 			// Initialize a fresh miner for the test case.
 			miner1 := setupMiner(t)
@@ -106,6 +106,77 @@ func runBitcoindEventsTests(t *testing.T, rpcPolling bool) {
 
 			test.testFn(t, miner1, btcClient)
 		})
+	}
+}
+
+// TestBitcoindEventsWatchAddrsFromTip applies the live-tip contract.
+//
+//nolint:paralleltest // Btcd's rpctest ports can collide across subtests.
+func TestBitcoindEventsWatchAddrsFromTip(t *testing.T) {
+	backends := []struct {
+		name       string
+		rpcPolling bool
+	}{
+		{
+			name:       "ZMQ",
+			rpcPolling: false,
+		},
+		{
+			name:       "RPC",
+			rpcPolling: true,
+		},
+	}
+
+	for _, backend := range backends {
+		t.Run(backend.name, func(t *testing.T) {
+			// Arrange: mine history before bitcoind starts and select its
+			// transport from data, excluding stale ZMQ transaction events.
+			miner := setupMiner(t)
+			addr := mineHistoricalWatchAddr(t, miner)
+			client := setupBitcoind(t, miner, backend.rpcPolling)
+
+			// Act: prove no hidden rescan with an isolated client, then run
+			// the production contract on the selected transport.
+			assertWatchAddrsFromTipDoesNotRescan(t, client, addr)
+			testWatchAddrsFromTipConformance(
+				t, miner, client, addr,
+			)
+
+			// Assert: one future notification drains through a later block.
+		})
+	}
+}
+
+// assertWatchAddrsFromTipDoesNotRescan uses a handler-free test client.
+func assertWatchAddrsFromTipDoesNotRescan(t *testing.T,
+	client *BitcoindClient, addr address.Address) {
+
+	t.Helper()
+
+	// Arrange: start only the sibling queue so registration owns the wait
+	// group and every tracked goroutine belongs to the operation under test.
+	probe, err := client.chainConn.NewBitcoindClient()
+	require.NoError(t, err)
+	probe.notificationQueue.Start()
+
+	atomic.StoreUint32(&probe.notifyBlocks, 1)
+	defer probe.Stop()
+
+	// Act: wait for registration work, then enqueue its FIFO marker.
+	err = probe.WatchAddrsFromTip(t.Context(), []address.Address{addr})
+	require.NoError(t, err)
+	probe.WaitForShutdown()
+
+	probe.notificationQueue.ChanIn() <- ClientConnected{}
+
+	// Assert: reach the marker without a historical rescan completion.
+	for ntfn := range probe.Notifications() {
+		if _, ok := ntfn.(ClientConnected); ok {
+			return
+		}
+
+		_, rescanned := ntfn.(*RescanFinished)
+		require.False(t, rescanned)
 	}
 }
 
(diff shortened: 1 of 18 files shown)


chain/btcd.go
```diff
@@ -336,6 +336,20 @@ func (c *RPCClient) Rescan(startHash *chainhash.Hash, addrs []address.Address,
...
+// registration so a canceled delivery attempt cannot introduce new watches.
+func (c *RPCClient) WatchAddrsFromTip(ctx context.Context,
+	addrs []address.Address) error {
+
+	err := ctx.Err()
+	if err != nil {
+		return err
+	}
+
+	return c.Client.NotifyReceived(addrs) //nolint:staticcheck
```
yyforyongyu, line 354: explain why we skip this linter

chain/interface.go
```diff
@@ -53,6 +53,9 @@ type Interface interface {
 	Rescan(*chainhash.Hash, []address.Address,
 		map[wire.OutPoint]address.Address) error
 	NotifyReceived([]address.Address) error
+	// WatchAddrsFromTip registers addresses for future notifications
```
yyforyongyu: add a newline above

chain/neutrino.go
```diff
@@ -586,6 +594,131 @@ func (s *NeutrinoClient) NotifyReceived(addrs []address.Address) error {
...
+	s.clientMtx.Lock()
+
+	if !s.started {
+		s.clientMtx.Unlock()
+
+		return errors.New("can't watch addresses when the chain client " +
+			"is not started")
+	}
+
+	if s.scanning {
```
yyforyongyu: we should add verbose inline docs explaining what is going on here?

yyforyongyu: Addressed in 53630930aaf2. The inline documentation now explains the completion boundary, the consistent rescan/channel snapshot under clientMtx, why clientMtx is released before waiting, and why rescanMtx remains held through the additive update.

chain/interface.go
yyforyongyu: Addressed in 63651f6a0872 by separating the new documented interface method with a blank line.

chain/btcd.go
yyforyongyu: Addressed in 5d5ab8b4cb5f. The inline comment now explains that staticcheck recommends LoadTxFilter, while this adapter deliberately keeps NotifyReceived because its handlers consume OnRecvTx and OnRedeemingTx rather than the newer relevant-transaction path.

chain/neutrino.go
```diff
@@ -586,6 +594,138 @@ func (s *NeutrinoClient) NotifyReceived(addrs []address.Address) error {
...
+	if err != nil {
+		return err
+	}
+
+	// Serialize with legacy rescan replacement so the rescan that reaches
+	// the boundary is also the one that receives the address update.
+	s.rescanMtx.Lock()
+	defer s.rescanMtx.Unlock()
+
+	s.clientMtx.Lock()
```
yyforyongyu: pay extreme attention to this lock inside a lock as it's the common source of deadlock - validate this is necessary change.

yyforyongyu: Validated and updated. Every dual-lock path acquires rescanMtx before clientMtx, completion callbacks acquire only clientMtx, and clientMtx is released before the boundary wait, so there is no reverse lock edge. Holding rescanMtx remains necessary to pin one rescan generation through the historical-tip wait and additive update. The validation also exposed incomplete and early-terminated generation states; the branch now fails fast for incomplete state and wakes the existing notification handler when a new rescan result channel is installed, with regression coverage proving terminal errors release rescanMtx. Opus, Kimi, and GPT all reviewed the final revision and returned LGTM.

yyforyongyu: Validated and corrected in the rewritten implementation. WatchAddrsFromTip now holds rescanMtx and clientMtx only while snapshotting the current generation and registering pending address references, then releases both before waiting for the historical scan to reach its tip. After the boundary closes, applyPendingWatchAddrs reacquires rescanMtx, verifies the captured generation, releases clientMtx during the existing rescan.Update handoff, and rechecks the generation before committing the active-address state. This avoids holding either client mutex across the potentially long wait while preventing an address update from being applied to a replacement rescan. The regression tests cover concurrent legacy registration and cancellation after an explicit pending-registration barrier.

chain/bitcoind_events_test.go
```diff
@@ -109,6 +111,45 @@ func runBitcoindEventsTests(t *testing.T, rpcPolling bool) {
...
 
+// TestBitcoindEventsWatchAddrsFromTip applies the shared contract visibly to
+// both maintained bitcoind notification transports.
+//
+// Each subtest starts both bitcoind and btcd. They remain sequential because
+// btcd's current rpctest setup can assign the same RPC ports to concurrent
+// harnesses.
+//
+//nolint:paralleltest
+func TestBitcoindEventsWatchAddrsFromTip(t *testing.T) {
```
Abdulkbk, line 115: `waitForExclusiveRelevantTx` and `waitForNoRelevantTxUntilBlock` check for `RelevantTx`, but `BitcoindClient` never emits that for confirmed trx, it dispatches through `FilteredBlockConnected`,  which the helpers are blind to. 

I replaced the body of `BitcoindClient.WatchAddrsFromTip`  with a full rescan from genesis, and the test still passes. 

```golang
func (c *BitcoindClient) WatchAddrsFromTip(ctx context.Context,
	addrs []address.Address) error {

	err := ctx.Err()
	if err != nil {
		return err
	}

	c.updateWatchedFilters(addrs)

	if err := c.NotifyBlocks(); err != nil {
		return err
	}

	// MUTATION: deliberately trigger a full historical rescan.
	genesis, err := c.GetBlockHash(0)
	if err != nil {
		return err
	}

	return c.Rescan(genesis, addrs, nil)
}
```

But btcd `TestRPCClientWatchAddrsFromTip` catches the mutation.

Abdulkbk (changes requested):

chain/bitcoind_events_test.go
yyforyongyu: Nice catch, this is now fixed!

Abdulkbk (approved):
LGTM

PR #1110, opened 2025-11-04
 [draft]: Introduce SQL schema for wallet

yyforyongyu:
Open this PR to track rebase conflicts.

TODO: add descriptions and link PRs and issues.

PR #1345, opened 2026-08-24
wallet+db: add durable database identity

yyforyongyu:
## Stack

This is a stacked PR based on [#1110](https://github.com/btcsuite/btcwallet/pull/1110) (`sql-wallet`). Review the `sql-wallet..task-db-identity` range; this PR should merge only after #1110.

## Summary

Implements roadmap Task 340's minimal durable database identity startup gate.

## Scope

This PR intentionally limits the persisted identity to the genesis hash, network magic, and full signet challenge digest. It initializes that identity in empty SQLite or PostgreSQL storage, checks it once on later Store startup, and stops startup when the stored identity does not match.

It does not implement every requirement currently listed in Task 340. Schema-family and generation fields, a behavior-parameter digest and immutable Manager network snapshot, PostgreSQL legacy-namespace discovery and empty-schema adoption, migration-state preflight, and concurrent-first-open serialization are deferred. Those changes expand runtime network ownership, compatibility and migration policy, or multi-opener coordination beyond this minimal startup gate and require separate design decisions. Simultaneous Managers targeting one database are unsupported by this reduced design.

## Change Description

Stores a durable SQL database identity derived from the wallet chain parameters when an empty SQLite or PostgreSQL database is initialized.

On later startup, the wallet verifies that identity before WAL configuration, migrations, or wallet access. Startup stops for a mismatched or malformed identity and for a populated database that has no identity marker.


wallet/internal/db/itest/database_identity_concurrency_test.go
```diff
@@ -0,0 +1,115 @@
+//go:build itest
+
+//nolint:lll,wsl_v5 // Concurrency choreography stays adjacent and readable.
```
yyforyongyu: we should not skip linters arbitrarily

wallet/internal/db/itest/database_identity_concurrency_test.go
```diff
@@ -0,0 +1,115 @@
...
+// concurrentOpenResult keeps goroutine outcomes and requested identity
+// together so assertions can determine which conflicting opener committed.
+type concurrentOpenResult struct {
+	identity db.DatabaseIdentity
+	err      error
+}
+
+// raceDatabaseOpen releases all requested identities simultaneously and waits
+// for every Store to close, giving each goroutine a bounded completion path.
+func raceDatabaseOpen(t *testing.T, database identityTestDatabase,
```
yyforyongyu: i don't see why we need to test concurrency for db identity??

wallet/internal/db/itest/database_identity_fixture_pg_test.go
```diff
@@ -0,0 +1,188 @@
+//go:build itest && test_db_postgres
```
yyforyongyu: db itest is meant to be backend agnostic, if that cannot be guaranteed, it is a design issue and we should fix that.

also we should not add arbitrary test files - we should respect the current filename structure.

wallet/internal/db/pg/database_identity.go
```diff
@@ -0,0 +1,150 @@
...
+	"context"
+	"fmt"
+
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
+)
+
+const (
+	// namespaceObjectsQuery includes relations, routines, and standalone types
+	// so an apparently empty fixed schema cannot hide a foreign user object.
+	namespaceObjectsQuery = `SELECT
```
yyforyongyu: how come we have raw sql queries in the code???

wallet/internal/db/sqlite/database_identity.go
```diff
@@ -0,0 +1,590 @@
...
+// databaseInspection retains the accepted namespace's migration version so a
+// locked initializer can apply only the forward bodies that remain pending.
+type databaseInspection struct {
+	state   databaseIdentityState
+	version int64
+}
+
+// databaseCatalog captures the ownership objects and their exact SQLite
+// catalog shapes so classification stays separate from query mechanics.
+type databaseCatalog struct {
```
yyforyongyu: i still think this PR is too large - we don't need `databaseIdentityState` all these different states? and why all these catalog?

wallet/internal/db/itest/sqlite_test.go
```diff
@@ -50,6 +57,55 @@ func NewTestStoreWithDerive(t *testing.T,
...
+				return nil, err
+			}
+
+			return store, nil
+		},
+		seed: func(seed databaseIdentitySeed) error {
+			statements := map[databaseIdentitySeed]string{
+				// This near-prefix name verifies that the catalog treats
+				// SQLite's internal "sqlite_" prefix literally.
+				identitySeedPopulated: "CREATE TABLE sqliteXforeign " +
```
yyforyongyu: what is this `seed` for? and why do we need to use raw queries here?

wallet/internal/db/itest/store_test.go
```diff
@@ -5,9 +5,28 @@ package itest
 import (
 	"testing"
 
+	"github.com/btcsuite/btcd/chaincfg/v2"
+	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	"github.com/stretchr/testify/require"
 )
 
+type databaseIdentityStore interface {
```
yyforyongyu: all these are missing godocs - i wonder why we need these?

wallet/internal/db/itest/store_test.go
```diff
@@ -22,3 +41,102 @@ func TestNewTestStore(t *testing.T) {
...
+	t.Parallel()
+
+	tests := []struct {
+		name       string
+		initialize bool
+		seed       databaseIdentitySeed
+		mismatch   bool
+		wantErr    error
+	}{
+		{"first initialization", false, identitySeedNone, false, nil},
```
yyforyongyu: malformed, we should always do,
```go
{
   name: xxx
   initialize: xxx
   field: value
...
}
```

wallet/internal/db/itest/pg_test.go
```diff
@@ -255,54 +256,146 @@ func NewTestStore(t *testing.T) *pg.Store {
...
+	t.Helper()
+
+	ctx := t.Context()
 	container, err := GetPostgresContainer(ctx)
 	require.NoError(t, err, "failed to get postgres container")
-
 	connStr, err := container.ConnectionString(ctx, "sslmode=disable")
 	require.NoError(t, err, "failed to get connection string")
-
-	// Connect to the default database to create our test database.
```
yyforyongyu: why remove this?

wallet/internal/db/common_test.go
```diff
@@ -37,3 +39,73 @@ func TestNilIfEmptyBytes(t *testing.T) {
...
+	reg := &chaincfg.RegressionNetParams
+	bad := ErrInvalidDatabaseIdentity
+
+	tests := []struct {
+		name   string
+		params *chaincfg.Params
+		digest []byte
+		want   error
+	}{
+		{name: "ordinary", params: reg},
```
yyforyongyu: write it in newlines,
```
{
   name: xxx
   params: xxx
}, {...

File: wallet/internal/db/database_identity.go [hunk 11]
```diff
@@ -0,0 +1,102 @@
...
+
+// ErrInvalidDatabaseIdentity reports invalid pre-Store identity input.
+var ErrInvalidDatabaseIdentity = errors.New("invalid database identity")
+
+// ErrDatabaseIdentityMismatch reports failure to establish stored identity.
+var ErrDatabaseIdentityMismatch = errors.New("database identity mismatch")
+
+// DatabaseIdentity copies the tuple so mutation cannot redirect SQL startup.
+type DatabaseIdentity struct {
+	genesisHash    chainhash.Hash
```
yyforyongyu: all these are missing inline docs - also, do we need to export this identity? should this struct be private?

sqlc.yaml
```diff
@@ -3,7 +3,9 @@
 version: "2"
 sql:
   - engine: "postgresql"
-    schema: "wallet/internal/sql/pg/migrations"
+    schema:
+      - "wallet/internal/sql/pg/migrations"
+      - "wallet/internal/sql/pg/schema"
```
yyforyongyu: find a better name and add inline docs explaining this?

wallet/manager.go
```diff
@@ -137,17 +137,29 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
...
+
 	var backend managerBackend
 
 	switch cfg.Backend {
 	case DBBackendKVDB:
 		backend, err = newKVDBManagerBackend(cfg)
 
 	case DBBackendSQLite:
-		backend, err = newSQLiteManagerBackend(ctx, cfg)
+		backend, err = newSQLiteManagerBackend(ctx, cfg, identity)
```
yyforyongyu, line 194: should the `identity` be part of the `cfg`, so a field there?

wallet/internal/db/itest/database_identity_concurrency_test.go
yyforyongyu: Addressed in the rewrite. The concurrency test file and its linter suppression were removed; the current implementation does not add arbitrary nolint directives.

wallet/internal/db/itest/database_identity_concurrency_test.go
yyforyongyu: Agreed. The concurrency machinery and tests were removed. The current tests cover only first initialization, matching reopen, mismatch rejection, and populated-unmarked rejection.

wallet/internal/db/itest/database_identity_fixture_pg_test.go
yyforyongyu: Addressed. The arbitrary fixture files were removed. Shared behavior stays backend-agnostic, while unavoidable backend setup remains in the established pg_test.go and sqlite_test.go files.

wallet/internal/db/pg/database_identity.go
yyforyongyu: Addressed. Production identity DDL and queries now live in checked SQL sources and generated sqlc methods; handwritten Go no longer contains raw identity queries.

wallet/internal/db/sqlite/database_identity.go
yyforyongyu: Addressed by rebuilding the design. databaseIdentityState and the broad catalog/state machinery are gone; startup now has only the minimal empty, matching, mismatch, and populated-unmarked decisions.

wallet/internal/db/itest/sqlite_test.go
yyforyongyu: Addressed. The variable is now fixtureDB. The remaining raw DDL is test-only and intentionally creates foreign pre-Store state that the production Store API cannot create, so the startup rejection path can be tested.

wallet/internal/db/itest/store_test.go
yyforyongyu: Addressed by removing the obsolete helper declarations and the added store_test.go machinery. The remaining task-added declarations have substantive declaration-leading documentation.

wallet/internal/db/itest/store_test.go
yyforyongyu: Addressed. The obsolete positional table was removed, and the remaining identity validation cases use multiline keyed struct literals.

wallet/internal/db/itest/pg_test.go
yyforyongyu: Restored. The extracted PostgreSQL fixture helper now preserves the baseline setup/cleanup explanations and the require.NotNil check for the admin connection.

wallet/internal/db/common_test.go
yyforyongyu: Addressed. All seven database-identity validation cases are now multiline keyed literals with each supplied field on its own line.

wallet/internal/db/database_identity.go
yyforyongyu: Addressed the missing documentation: every invariant-bearing field now has inline rationale. DatabaseIdentity remains exported because wallet.Manager constructs it for the sibling pg/sqlite Config types; all fields remain private so callers cannot forge the validated tuple.

sqlc.yaml
yyforyongyu: Addressed. Both generic schema directories were renamed to bootstrap, and sqlc.yaml now explains that these DDL sources are listed for type-checking while Store startup applies them before numbered migrations.

wallet/manager.go
yyforyongyu: The derived identity is intentionally not duplicated in ManagerConfig. ManagerConfig owns the source inputs (ChainParams and the optional signet digest); NewManager derives one validated value and passes it to the concrete pg/sqlite Config, whose Identity field owns the Store boundary.

yyforyongyu: The derived identity is intentionally not duplicated in ManagerConfig. ManagerConfig owns the source inputs (ChainParams and the optional signet digest); NewManager derives one validated value and passes it to the concrete pg/sqlite Config, whose Identity field owns the Store boundary.

wallet/internal/db/database_identity.go
```diff
@@ -0,0 +1,101 @@
...
+
+	"github.com/btcsuite/btcd/chaincfg/v2"
+	"github.com/btcsuite/btcd/chainhash/v2"
+)
+
+// ErrInvalidDatabaseIdentity reports invalid pre-Store identity input.
+var ErrInvalidDatabaseIdentity = errors.New("invalid database identity")
+
+// ErrDatabaseIdentityMismatch reports rejected persisted identity state.
+var ErrDatabaseIdentityMismatch = errors.New("database identity mismatch")
```
yyforyongyu: group the def? so
```go
var (
 xxx = yyy
)
```

wallet/internal/db/database_identity.go
```diff
@@ -0,0 +1,101 @@
...
+
+// DatabaseIdentity copies the tuple so mutation cannot redirect SQL startup.
+type DatabaseIdentity struct {
```
yyforyongyu, line 24: why do we have a public struct while all its fields are private?

wallet/internal/db/database_identity.go
```diff
@@ -0,0 +1,101 @@
...
+		return fmt.Errorf("%w: missing genesis hash",
+			ErrInvalidDatabaseIdentity)
+	}
+
+	if i.isSignet != i.hasSignetInput {
+		return fmt.Errorf("%w: inconsistent signet digest",
+			ErrInvalidDatabaseIdentity)
+	}
+
+	if i.hasSignetInput && binary.LittleEndian.Uint32(
```
yyforyongyu, line 74: exit early on `if !i.hasSignetInput`?

wallet/internal/db/database_identity.go
```diff
@@ -99,3 +101,37 @@ func (i DatabaseIdentity) Matches(genesisHash []byte, networkMagic int64,
 				signetDigest, i.signetDigest[:],
 			))
 }
+
+// VerifyStoredDatabaseIdentity applies the backend-neutral singleton and tuple
+// contract while leaving operational query failures distinguishable.
+func VerifyStoredDatabaseIdentity(ctx context.Context,
```
yyforyongyu, line 108: ok should we instead build a db ops for it?

wallet/internal/db/database_identity.go
yyforyongyu: Done. The two database identity sentinel definitions now share one var block.

wallet/internal/db/database_identity.go
yyforyongyu: DatabaseIdentity is intentionally opaque. It must be exported because wallet constructs it and the sibling PostgreSQL and SQLite packages receive it across package boundaries. Keeping its fields private forces construction through NewDatabaseIdentity and prevents callers from forging or mutating an already validated tuple.

wallet/internal/db/database_identity.go
yyforyongyu: Done. Values now returns early when no signet digest is present, making the SQL NULL case explicit.

wallet/internal/db/database_identity.go
yyforyongyu: I kept the two callbacks as the smaller adapter boundary. The database guide discourages ops interfaces for thin reads, and the generated PostgreSQL and SQLite GetDatabaseIdentity methods return different row types. A named ops interface would therefore require two extra adapter types and additional mock surface without changing the shared singleton invariant owned here.

docs/developer/adr/0014-sql-database-identity.md
```diff
@@ -105,171 +93,73 @@ namespace. Their identity tables are:
...
-- SQLite starts the transaction with `BEGIN IMMEDIATE`.
-- PostgreSQL begins an explicit transaction-scoped `READ COMMITTED`
-  transaction and makes `pg_advisory_xact_lock` its first statement. This does
-  not change database, role, session, or runtime Store defaults. The lock is
-  local to the current database and released at transaction end.
-
-The PostgreSQL signed bigint lock key is the big-endian two's-complement int64
-formed from the first eight bytes of
-`SHA256(ASCII("btcwallet-sql-identity-migration"))`:
-`-1270285209963246243`.
```
GustavoStingelin: The intent here is to simplify the adr?

wallet/internal/db/itest/store_test.go
```diff
@@ -22,3 +71,93 @@ func TestNewTestStore(t *testing.T) {
...
+		append([]byte{byte(len(challenge))}, challenge...),
+	)
+	otherDigest := append([]byte(nil), digest...)
+	otherDigest[len(otherDigest)-1] ^= 1
+	signet, signetErr := db.NewDatabaseIdentity(&chaincfg.SigNetParams, digest)
+	otherSignet, otherErr := db.NewDatabaseIdentity(
+		&chaincfg.SigNetParams, otherDigest,
+	)
+	require.NoError(
+		t, errors.Join(regtestErr, testErr, signetErr, otherErr),
```
GustavoStingelin: What is the intended benefit of joining these independent constructor errors before passing them to `require.NoError`? Since each identity is a setup prerequisite, would it be clearer to assert immediately after each constructor instead? Sequential `require.NoError` calls fail fast and identify the exact setup step that failed; the tradeoff is that only the first failure is reported. If collecting all independent failures is intentional, separate `assert.NoError` calls would make that behavior more explicit. I think `errors.Join` is more useful when an operation and its cleanup must both be attempted and returned together, as in `openIdentityStore`.

wallet/internal/sql/pg/queries/database_identity.sql
```diff
@@ -0,0 +1,77 @@
...
+        FROM pg_catalog.pg_namespace AS n
+        WHERE n.nspname = 'btcwallet'
+    ) AS schema_exists,
+    exists(
+        SELECT 1 FROM identity_relation
+    ) AS identity_exists,
+    coalesce(
+        (
+            SELECT
+                array_agg(a.attname::TEXT ORDER BY a.attnum) = ARRAY[
```
GustavoStingelin, line 26: Should `identity_readable` verify the constraints required by ADR 0014, rather than only column names, types, and nullability? A pre-existing table with these four columns and one matching row, but without the primary key, `CHECK (id = 1)`, or value-length/range constraints, appears to pass this probe and the subsequent count/value checks. SQLite similarly verifies the primary key but not the `CHECK` constraints. Should startup reject that lookalike schema, or should the ADR clarify that only the current row shape and values constitute the marker contract?

wallet/internal/db/itest/store_test.go
```diff
@@ -3,11 +3,60 @@
...
+	identity db.DatabaseIdentity) (io.Closer, error) {
+
+	t.Helper()
+
+	store, err := fixture.open(identity)
+	if store == nil {
+		return nil, errors.Join(err, errors.New("nil Store"))
+	}
+
+	return store, errors.Join(err, store.Close())
```
GustavoStingelin, line 61: Could we make this helper's ownership semantics more explicit? `openIdentityStore` closes every non-nil store before returning it, so callers receive an already-closed `io.Closer` solely to test whether the backend returned one. The name and return type suggest the resource remains usable. Would returning a boolean alongside the error, or renaming this to indicate open-and-close behavior, make misuse less likely?

wallet/internal/db/itest/store_test.go
```diff
@@ -3,11 +3,60 @@
...
+	malformations map[string][]string
+}
+
+// mutateIdentity applies test-only SQL and always closes its temporary handle.
+func (f databaseIdentityFixture) mutateIdentity(t *testing.T,
+	statements []string) {
+
+	t.Helper()
+
+	fixtureDB, err := sql.Open(f.driver, f.source)
```
GustavoStingelin: Could we register cleanup immediately after `sql.Open`? A failing `require.NoError` inside the statement loop calls `FailNow`, so execution never reaches the explicit `Close` below and the handle may remain open on the failure path. A `t.Cleanup` or deferred close would preserve cleanup even when setup fails.

wallet/internal/db/common_test.go
```diff
@@ -37,3 +41,121 @@ func TestNilIfEmptyBytes(t *testing.T) {
...
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			// Act: Build the tuple checked before SQL connection setup.
+			_, err := NewDatabaseIdentity(tc.params, tc.digest)
+
+			// Assert: The sentinel appears exactly for invalid input rows.
+			require.ErrorIs(t, err, tc.want)
```
GustavoStingelin: Could the success cases use `require.NoError` explicitly instead of relying on `require.ErrorIs(err, nil)`? The latter works, but branching on `want == nil`, as `TestManagerConfigValidate` does, makes the expected success path and resulting diagnostics clearer.

docs/developer/adr/0014-sql-database-identity.md
yyforyongyu: Yes! I simplified it further to find a minimal design

wallet/internal/db/itest/store_test.go
```diff
@@ -3,11 +3,60 @@
...
+
+// openIdentityStore closes every successful Store so rejection leaks nothing.
+func openIdentityStore(t *testing.T, fixture databaseIdentityFixture,
+	identity db.DatabaseIdentity) (io.Closer, error) {
+
+	t.Helper()
+
+	store, err := fixture.open(identity)
+	if store == nil {
+		return nil, errors.Join(err, errors.New("nil Store"))
```
GustavoStingelin: why join instead of error ftm?

GustavoStingelin (changes requested):
just some quality comments

wallet/internal/db/itest/store_test.go
yyforyongyu: Yeah updated!

wallet/internal/sql/pg/queries/database_identity.sql
yyforyongyu: There is no path we can create that case? The designed flow is,

1. Wallet creates the identity table with the proper constraints.
2. Wallet writes the singleton identity.
3. Later startups read and compare that identity.

wallet/internal/db/itest/store_test.go
yyforyongyu: ok decided to go with the minimal option and named it `openAndCloseIdentityStore`

wallet/internal/sql/pg/queries/database_identity.sql
GustavoStingelin: Yes, I think it's not possible for normal operation, thanks for updating the ADR

GustavoStingelin (approved):
LGTM!

PR #1346, opened 2026-08-24
itest: add TxReader integration tests

Abdulkbk:
Task 436 of the [#1250][issue] TxReader lane, following #1316 (TxCreator)
and #1334 (TxPublisher).

### What

Nine `txreader ...` cases covering `GetTx` and `ListTxns` through real kvdb,
SQLite and PostgreSQL wallets: empty history, a missing transaction, a received
and an authored transaction field by field, the mined transition, confirmation
counting, the supported list ranges, agreement between the two readers, and a
full durable reopen.

Two harness additions make that possible:

- `WalletFunding` now reports the funding transaction and its confirming block.
  The harness already had both and discarded them, and the wallet did not author
  that transaction, so a case cannot reconstruct it from the amounts it asked
  for.
- `(*HarnessTest).ReopenWallet` owns the close-and-reopen lifecycle. The Config
  and Manager a reload needs are built inside `NewWallet` and never handed out,
  so a case reproducing the sequence would become a second owner of wallet
  lifecycle policy.

`spendFee` moves to `itest/fixtures_test.go`, since the reader cases need it and
a component's test file should not be the source of another's constants.

### One gap, deliberately not faked

Task 436 asks for pending and published `TxStatus` fixtures. Published is
covered on all three backends; **pending has no public producer**. Task 339
exports `TxStatusPending` and converts it faithfully, but every write path
records published — `recordTxAndCredits` in `wallet/tx_publisher.go` and
`txNotificationState` in `wallet/syncer.go` — and `LabelTx` touches only the
label. The remaining `TxStatusPending` mentions in `package wallet` are read
branches that promote a row "another workflow inserted", and that workflow does
not exist yet. Task 339 reached pending by writing the store directly, which
this task excludes.

The gap is recorded in a comment in `itest/tx_reader_test.go`. Happy to add the
case once a producer lands, or to write it against the store if you would rather
have the coverage now.

### Testing

Full suite green on kvdb, SQLite and PostgreSQL (btcd), plus a shuffled run
(`-shuffleseed=436`) and the focused cases on neutrino. Every commit builds,
vets and lints on its own. `bitcoind` is not installed on this machine, so that
chain backend is untested locally.

[issue]: https://github.com/btcsuite/btcwallet/issues/1250

### DIFF
--- a/itest/tx_reader_test.go
+++ b/itest/tx_reader_test.go
@@ -344,3 +344,85 @@ func testGetTxUnmined(h *bwtest.HarnessTest) {
 	// confirm the transaction now that the unmined view has been read.
 	h.MineBlockWithTx(tx)
 }
+
+// testGetTxMined verifies that confirming a transaction adds the block that
+// contains it and the confirmations it earned, and changes nothing else about
+// how the wallet reports it.
+func testGetTxMined(h *bwtest.HarnessTest) {
+	w, funding := h.NewWallet(bwtest.WalletFixture{
+		AddrType: txReaderFundingType,
+		Amounts:  []btcutil.Amount{oneBTC},
+		Unlocked: true,
+	})
+
+	addr := h.NewWalletAddressOfType(w, txReaderFundingType)
+
+	pkScript, err := txscript.PayToAddrScript(addr)
+	require.NoError(h, err, "failed to create payment pkscript")
+
+	tx := h.SignSpend(w, bwtest.SpendFixture{
+		Inputs: []wire.OutPoint{funding.WalletOutpoints[0]},
+		Outputs: []wire.TxOut{{
+			Value:    oneBTC - spendFee,
+			PkScript: pkScript,
+		}},
+	})
+
+	txid := tx.TxHash()
+
+	err = w.Broadcast(h.Context(), tx, "")
+	require.NoError(h, err, "failed to broadcast transaction")
+
+	// Read the unconfirmed view first, so the comparison below is against
+	// what confirmation actually changed rather than against a second
+	// description of the same transaction.
+	unmined, err := w.GetTx(h.Context(), txid)
+	require.NoError(h, err, "failed to get the published transaction")
+
+	block := h.MineBlockWithTx(tx)
+	_, height := h.GetBestBlock()
+
+	mined, err := w.GetTx(h.Context(), txid)
+	require.NoError(h, err, "failed to get the confirmed transaction")
+
+	// The block that confirmed it is now reported, and the mined block is
+	// the chain tip, so the transaction has exactly one confirmation.
+	require.Equal(
+		h, block.BlockHash(), mined.Block.Hash, "unexpected block hash",
+	)
+	require.Equal(h, height, mined.Block.Height, "unexpected block height")
+	require.Equal(
+		h, block.Header.Timestamp.Unix(), mined.Block.Timestamp,
+		"unexpected block timestamp",
+	)
+	require.Equal(h, int32(1), mined.Confirmations, "unexpected confirmations")
+
+	// The wallet still reports a time at which it saw the transaction, but
+	// which time that is does not survive confirmation on every backend, so
+	// it is held out of the comparison below rather than asserted.
+	//
+	// kvdb rebuilds the stored record from the confirming notification,
+	// through InsertConfirmedTx, so the record takes that notification's
+	// timestamp. Its own confirmedBatchLabel keeps the stored label across
+	// the same transition, but nothing does that for the received time. The
+	// SQL backends keep the time the wallet first saw the transaction.
+	// Which of the two the public reader should report is a question for
+	// the store, not for this case to settle by picking one.
+	require.False(h, mined.ReceivedTime.IsZero(), "no received time")
+
+	// Nothing else about the transaction changed. Comparing the whole
+	// result with only the confirmation facts removed leaves no field that
+	// could have been quietly rewritten.
+	unconfirmed := *mined
+	unconfirmed.Block = nil
+	unconfirmed.Confirmations = 0
+
+	// Carried over rather than compared: this is the exemption explained
+	// above, not an assertion that the two times agree.
+	unconfirmed.ReceivedTime = unmined.ReceivedTime
+
+	require.Equal(
+		h, unmined, &unconfirmed,
+		"confirmation altered the transaction beyond its block",
+	)
+}
--- a/itest/tx_reader_test.go
+++ b/itest/tx_reader_test.go
@@ -1,0 +1,61 @@
+//go:build itest
+
+package itest
+
+import (
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/stretchr/testify/require"
+)
+
+// txReaderFundingType is the address type the TxReader cases fund and spend.
+// The script class the readers report for a wallet-owned output follows from
+// it, so the funded type is never stated twice.
+const txReaderFundingType = waddrmgr.WitnessPubKey
+
+// unminedHeight is the height that stands for "no confirming block" in the
+// range ListTxns takes. A negative start includes the unmined transactions
+// before the confirmed ones, a negative end includes them after, and both
+// negative asks for the unmined transactions alone.
+const unminedHeight = -1
+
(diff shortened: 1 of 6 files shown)


itest/tx_reader_test.go
```diff
@@ -0,0 +1,746 @@
...
+
+// testListTxnsEmptyHistory verifies that a wallet holding no transactions
+// reports none, whichever part of the range a caller asks about.
+func testListTxnsEmptyHistory(h *bwtest.HarnessTest) {
+	// An unfunded wallet has received nothing and published nothing, so
+	// every range below covers a history that is genuinely empty rather
+	// than merely filtered out.
+	w, _ := h.NewWallet(bwtest.WalletFixture{AddrType: txReaderFundingType})
+
+	_, tip := h.GetBestBlock()
```
GustavoStingelin, line 51: Non-blocking question: The chain is already non-empty because harness setup mines unrelated coinbase transactions. Would it still be worth adding an unrelated regular transaction, for example one paying another wallet, to make this case explicitly cover wallet-relevance filtering? Or would that be better covered by sync tests?

itest/tx_reader_test.go
```diff
@@ -0,0 +1,746 @@
...
+	require.Equal(h, txReaderStatus, detail.Status, "unexpected status")
+	require.Empty(h, detail.Label, "unexpected label")
+	require.False(h, detail.ReceivedTime.IsZero(), "no received time")
+
+	// The wallet gained exactly what it was paid.
+	require.Equal(h, btcutil.Amount(oneBTC), detail.Value, "unexpected value")
+
+	// It cannot state the fee, because the inputs are the payer's and their
+	// values are not in the wallet's records.
+	require.Zero(h, detail.Fee, "unexpected fee")
```
GustavoStingelin, line 136: Non-blocking and not for this PR:

The API uses `0` for both "fee unknown because not all input values are available" and "fee is known to be zero." That ambiguity becomes more relevant with BIP 431/TRUC package relay, which permits a zero-fee presigned parent to be relayed with a fee-paying child and enables zero-fee Lightning commitment transactions.

Should `Fee` and `FeeRate` eventually expose whether their values are known, for example with optional values or a `FeeKnown` field?

What do you guys think about it?

itest/tx_reader_test.go
```diff
@@ -0,0 +1,746 @@
...
+		h, btcutil.Amount(-spendFee), detail.Value, "unexpected value",
+	)
+	require.Equal(h, btcutil.Amount(spendFee), detail.Fee, "unexpected fee")
+
+	weight := btcunit.NewWeightUnit(uint64(
+		blockchain.GetTransactionWeight(btcutil.NewTx(tx)),
+	))
+	require.Equal(h, weight, detail.Weight, "unexpected weight")
+
+	feeRate := btcunit.CalcSatPerVByte(btcutil.Amount(spendFee), weight.ToVB())
```
GustavoStingelin, line 315: Nice to see the new `btcunit` types used here.

itest/tx_reader_test.go
```diff
@@ -0,0 +1,746 @@
...
+
+	// The funding block is the tip, so the transaction in it is confirmed
+	// exactly once.
+	detail, err := w.GetTx(h.Context(), txid)
+	require.NoError(h, err, "failed to get the funding transaction")
+	require.Equal(
+		h, int32(1), detail.Confirmations, "unexpected initial confirmations",
+	)
+
+	h.MineEmptyBlocks(blocksMined)
```
GustavoStingelin, line 453: Follow-up idea: would it be useful to cover a published transaction remaining unmined while the chain tip advances, asserting that `Block` stays nil and `Confirmations` stays zero?

The existing unmined case checks those fields at the current tip. `MineEmptyBlocks` cannot exercise this scenario because it requires an empty mempool, so this would need dedicated harness support to mine a block that deliberately excludes the transaction.

GustavoStingelin (commented):
Looking very good!

pending a rebase

Abdulkbk:
Ohh forgot to rebase after https://github.com/btcsuite/btcwallet/pull/1326 was merged...

yyforyongyu (changes requested):
Yeah let's do a rebase first

itest/tx_reader_test.go
Abdulkbk: I think it will be better suited for sync tests. Plus, there is https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/449-multi-wallet-sql-isolation-scenario.md for a multi-wallet scenario that covers SQL only (since kvdb manager supports one wallet per db).

itest/tx_reader_test.go
Abdulkbk: I think we can differentiate using `.IsOurs`. `Fee` is suppressed exactly when not every input is ours (`wallet/tx_reader.go:416`, `len(OwnedInputs) != len(msgTx.TxIn)`), and `PrevOut.IsOurs` is populated from that same `OwnedInputs` set. So "the fee is known" is precisely "every `PrevOuts[i].IsOurs` is true".

For the TRUC case:
- zero-fee presigned parent, all inputs ours → `Fee: 0`, every `IsOurs` true → known, and genuinely zero

- received transaction with external inputs → `Fee: 0`, some `IsOurs` false → unknown

itest/tx_reader_test.go
yyforyongyu: Good idea - i think we don't wanna do scenario based tests here per scope limit, but we can record these ideas in the roadmap repo.

itest/tx_reader_test.go
yyforyongyu: yeah let's track this issue in the roadmap repo?

itest/tx_reader_test.go
```diff
@@ -0,0 +1,746 @@
...
+			h, isOurs, output.IsOurs, "unexpected output %d ownership", i,
+		)
+
+		// What the reader reports about the payer's change beyond its
+		// script, amount and disowned state is the payer's business: the
+		// script it pays to is the miner's choice, so pinning the class
+		// or the address count here would make this case fail for a
+		// change in how the miner pays itself. Only the outputs the
+		// wallet was funded through follow from the funded address type.
+		if !isOurs {
```
yyforyongyu, line 230: Task 436 asks us to compare every exported transaction-result field, but this `continue` means `Type` and `Addresses` are never asserted for foreign outputs even though `populateOutputs` fills them for every output. Could we derive the expected class and addresses from `tx.TxOut[i].PkScript` with `txscript.ExtractPkScriptAddrs(..., h.NetParams())` and assert them before this ownership branch? That avoids hard-coding the miner change format while still covering both exported fields.

itest/tx_reader_test.go
```diff
@@ -0,0 +1,746 @@
...
+// testListTxnsAgreesWithGetTx verifies that the two readers describe the same
+// transaction identically, so a caller's choice between them cannot change what
+// it learns.
+func testListTxnsAgreesWithGetTx(h *bwtest.HarnessTest) {
+	// wantTxns is how many transactions the fixture below leaves the wallet
+	// holding: the funding transaction, the one it mined, and the one still
+	// in the mempool.
+	const wantTxns = 3
+
+	w, funding := h.NewWallet(bwtest.WalletFixture{
```
yyforyongyu: This repeats the same fund-two-coins, mine-one-spend, leave-one-unmined history built above and again in the durable-reopen case, so the amounts, fee, ordering, and expected count have three independent owners. Could we extract only this Arrange logic into a file-local `newTxReaderHistory` fixture that returns the wallet, funding transaction, mined transaction and height, and unmined transaction, while leaving each `ListTxns`, `GetTx`, or `ReloadWallet` act and its assertions visible in the individual test? The expected count can then be derived from the returned transaction identities instead of maintaining separate constants.

yyforyongyu (changes requested):
Only two comments, otherwise good to go!

itest/tx_reader_test.go
GustavoStingelin: Tracked in https://github.com/yyforyongyu/btcwallet-sql-roadmap/issues/18.

GustavoStingelin (approved):
LGTM!

GustavoStingelin:
@Abdulkbk I guess it is missing ask yy review again?

yyforyongyu (approved):
LGTM!

PR #1347, opened 2026-08-24
itest: add TxWriter label integration tests

Abdulkbk:
Addresses Task 437 of the [#1250](https://github.com/btcsuite/btcwallet/issues/1250)
integration-test lane. `LabelTx` is the whole `TxWriter` interface, and nothing
proved its contract through a real Store-backed wallet.

Seven cases, registered as `txwriter ...`, run unchanged on kvdb, SQLite and
PostgreSQL:

| Case | Contract |
|---|---|
| `txwriter label transaction` | A written label is what both maintained readers report. |
| `txwriter replace label` | A second label replaces the first, as documented. |
| `txwriter label boundaries` | The shortest and longest labels every backend stores survive unaltered. |
| `txwriter reject unknown transaction` | Labeling a transaction the wallet does not hold fails with `ErrTxNotFound`, creates nothing, and leaves the stored label alone. |
| `txwriter reject oversize label` | A label past the limit is refused and the previous one survives. |
| `txwriter label survives reload` | The label is still reported after the wallet is stopped, its database closed, and the same wallet loaded again. |
| `txwriter wallet state` | Labeling before `Start` and after `Stop` returns `ErrStateForbidden`. |

Rebased onto `sql-wallet` after #1346. The durability case crosses the
close-and-reload lifecycle through `bwtest.(*HarnessTest).ReloadWallet`, which
landed with that PR; the harness commit this branch used to carry was the same
helper under an earlier name and is dropped, so nothing here touches `bwtest`.

## Two gaps this exposed

Both are in production code and are left for their own tasks rather than fixed
inside a test change.

**1. The empty label has no single behavior.** `db.UpdateTxParams` documents the
empty string as valid and as clearing the label, and the SQL backends implement
that. kvdb refuses it: `UpdateTx` goes through `wtxmgr.PutTxLabel`, which
returns `ErrEmptyLabel` for a zero-length label. So the same call succeeds on
two backends and fails on the third, and no backend-neutral case can be written
for it at all. Clearing a label is not otherwise reachable through the public
API, so today a caller cannot portably remove one.

**2. An over-long label is refused three different ways.** Every backend
enforces the 500-character limit, but each reports its own violation:
`wtxmgr.ErrLabelTooLong` on kvdb, a failed `CHECK` constraint on SQLite, and an
overlong `VARCHAR(500)` value on PostgreSQL. `LabelTx` publishes no identity of
its own, so `txwriter reject oversize label` can only assert that some error
came back. A `wallet.ErrLabelTooLong` — with the length check in `LabelTx`,
above the stores — would make that case an `ErrorIs` like every other rejection
in the suite.

A related detail for whoever takes that on: `wtxmgr` counts the limit in bytes
while both SQL schemas count it in characters, so a 200-character label of
3-byte runes is stored by the SQL backends and refused by kvdb. The boundary
case uses ASCII deliberately for that reason.

## Testing

`txwriter` cases pass on every supported database and both available chain
backends:

- `make itest chain=btcd db=sqlite icase='txwriter'`
- `make itest chain=btcd db=kvdb icase='txwriter'`
- `make itest chain=btcd db=postgres icase='txwriter'`
- `make itest chain=bitcoind db=sqlite icase='txwriter'`

The full suite passes shuffled on both sqlite (`-shuffleseed=42`) and kvdb
(`-shuffleseed=7`), 66 cases, no failures and no skips.

### DIFF
--- a/itest/tx_writer_test.go
+++ b/itest/tx_writer_test.go
@@ -191,3 +191,55 @@ func testLabelTxRejectUnknown(h *bwtest.HarnessTest) {
 		h, kept, detail.Label, "refused label displaced the stored one",
 	)
 }
+
+// testLabelTxRejectOversize verifies that the first label longer than the
+// wallet stores is refused by every backend, and that the refusal leaves the
+// transaction's existing label in place rather than truncating it to fit.
+//
+// The rejection is asserted only as an error. Every backend enforces the limit
+// itself, so the caller receives whichever of three unrelated values that
+// backend produces: kvdb reports wtxmgr.ErrLabelTooLong, SQLite a violated
+// CHECK constraint, and PostgreSQL an overlong value for its column type. There
+// is no wallet error identity to match on, and matching any one backend's text
+// would tie this case to a detail no contract promises. That missing identity
+// is a gap in LabelTx, not in this case.
+//
+// The identities LabelTx does publish are ruled out instead, so the case still
+// fails if the write is ever refused for a reason other than the label it was
+// given.
+func testLabelTxRejectOversize(h *bwtest.HarnessTest) {
+	const kept = "rent for march"
+
+	w, funding := h.NewWallet(bwtest.WalletFixture{
+		AddrType: txWriterFundingType,
+		Amounts:  []btcutil.Amount{oneBTC},
+	})
+
+	txHash := funding.WalletOutpoints[0].Hash
+
+	err := w.LabelTx(h.Context(), txHash, kept)
+	require.NoError(h, err, "failed to label the funding transaction")
+
+	// One character past the longest label testLabelTxBoundaries stores, so
+	// length is the only thing wrong with it.
+	oversize := strings.Repeat("x", wtxmgr.TxLabelLimit+1)
+
+	err = w.LabelTx(h.Context(), txHash, oversize)
+
+	require.Error(h, err, "label longer than the limit was accepted")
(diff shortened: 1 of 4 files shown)


Abdulkbk:
Found some gaps while working on this (described in the PR description). I will be creating task for them.

itest/tx_writer_test.go
```diff
@@ -0,0 +1,332 @@
...
+// immaterial here; naming one keeps the fixture explicit instead of resting on
+// a zero value that means something to other components.
+const txWriterFundingType = waddrmgr.WitnessPubKey
+
+// wholeHistoryStart and wholeHistoryEnd are the ListTxns range covering every
+// transaction a wallet holds: every confirmed block from genesis, and then the
+// unmined transactions, which is what the negative end height asks for.
+const (
+	wholeHistoryStart = 0
+	wholeHistoryEnd   = -1
```
yyforyongyu: this duplicates `unminedHeight`, maybe we can move these to `fixture_test.go` and unify them

yyforyongyu (approved):
LGTM! Left one nit - also think we should create a followup PR to fix the gaps found?

itest/tx_writer_test.go
Abdulkbk: done

GustavoStingelin (approved):
LGTM!

PR #1348, opened 2026-08-24
txauthor+wallet: validate transaction output and fee arithmetic

Abdulkbk:
Task 357 — validate transaction output and fee arithmetic.

Transaction construction did every amount computation with raw `btcutil.Amount`
arithmetic: `SumOutputValues` wrapped silently, the target-plus-fee addition was
unchecked and formed twice per iteration, the sufficiency and change
subtractions were unchecked, and `txrules.FeeForSerializeSize` accepted a
non-positive rate and clamped an unrepresentable fee to `MaxSatoshi` instead of
reporting it.

`wallet/txauthor` now owns the checked primitives:

- `CheckOutputs` — rejects a nil element, a negative or over-`MaxSatoshi` value,
  and an aggregate above `MaxSatoshi`, returning the checked total for reuse. A
  nil set stays valid and totals zero, so a change-only sweep still authors.
- `CheckedFeeForSerializeSize` — rejects a rate at or below zero and a negative
  size before multiplying, reports a product overflow, and requires the rounded
  fee to land in `0..MaxSatoshi`.
- Internal checked add and subtract behind both.

`NewUnsignedTransaction` carries their results through construction: the output
set and the fee rate are validated before the first `fetchInputs` call, the
target is formed once and reused for the sufficiency test, and change is
derived from the sufficiency remainder rather than re-derived from the input
total. `SumOutputValues` stays exported for external callers but is deprecated
and no longer used.

TxCreator translates the result once, at the shared authoring boundary, so
`CreateTransaction` and `FundPsbt` both report the wallet's own sentinels.
Output value violations reuse the `txrules` errors `validateTxIntent` already
reports; only `ErrOutputTotalExceedsMax`, `ErrFeeOutOfRange` and
`ErrInvalidFeeRate` are new. Conditions no caller can observe are not named at
the wallet level and pass through unchanged, alongside `InputSourceError`.

Target-plus-fee is checked for **overflow only**, never against `MaxSatoshi`: a
single output of exactly `MaxSatoshi` must still reach the input source and fail
for want of coins, which `txcreator output boundaries` asserts.

### Notes for review

- **Behaviour change on the deprecated interface.** Nothing between
  `CreateSimpleTx` / `SendOutputs` / `SendOutputsWithInput` /
  `FundPsbtDeprecated` and `txauthor` bounded the fee rate, so a zero rate used
  to author a zero-fee transaction and now returns an error. The in-tree legacy
  RPC always passes `txrules.DefaultRelayFeePerKb` and is unaffected, but
  external callers set their own rate. Pinned by
  `TestTxToOutputsRejectsNonPositiveFeeRate`.
- **What is reachable from the public API.** `validateTxIntent` front-runs the
  per-output bounds and the fee-rate bounds, so the aggregate output total is
  the only class a public wrapper reaches today; it has no intent-level
  counterpart. `ErrFeeOutOfRange` is reachable once `DefaultMaxFeeRate` is
  raised — it is a mutable package variable meant to become configurable, and at
  9e15 sat/kvb any transaction over ~234 vbytes prices a fee above `MaxSatoshi`.
  The rest is defence for the internal boundary.
- **`btcunit` overflow, separate defect.** `CalcSatPerKVByte` computes
  `fee * kilo` in int64, so `NewSatPerKVByte` silently wraps to a negative rate
  above ~9.22e15 sat/kvb (`NewSatPerKVByte(3e16).Val()` returns
  `-6893488147419104`). Outside this task's scope; needs its own roadmap task.
- The suggested fix names one add helper; the Details require every sufficiency
  and change subtraction to be checked, so its subtraction counterpart ships
  with it. Both stay unexported — every call site is bounded by the checks
  around it and cannot currently fail, which is recorded at the loop.
- The root module gains a `replace` for the local `txauthor`, matching the ones
  `wtxmgr` and `wallet/txsizes` already carry. Without it none of this reaches
  TxCreator. It is load-bearing, as `wtxmgr`'s already is on this branch, and
  wants converting to a tagged release before `sql-wallet` lands.
- `countInputTypes` is extracted so the added error handling does not push
  `NewUnsignedTransaction` past the `funlen` limit.

### Testing

Nested-module and wallet tests cover nil, negative, individually over-maximum,
checked-sum overflow, aggregate-over-maximum, zero and negative direct fee
rates, negative size, fee-product overflow, a rounded fee above `MaxSatoshi`,
target-plus-fee overflow, the smallest valid positive rate, and the exact
maximum boundaries; every failure case asserts both callbacks went unused.
Translation is covered per class, through `FundPsbt`, and for message
stability.

Full `go test ./wallet/...`, the `txauthor` module tests, and the complete itest
suite pass. golangci-lint reports 0 issues on the root module.

### DIFF
--- a/wallet/tx_creator.go
+++ b/wallet/tx_creator.go
@@ -82,6 +82,27 @@ var (
 	// errNegativeHeight is returned when a height expected to be non-negative
 	// is negative.
 	errNegativeHeight = errors.New("negative height")
+
+	// The two errors below name the same conditions as their txauthor
+	// counterparts, deliberately including their names. They exist so a
+	// caller can match an authoring failure without importing the nested
+	// module; the originating txauthor error stays in the chain for anyone
+	// who wants the finer distinction.
+
+	// ErrOutputTotalExceedsMax is returned when a transaction's outputs are
+	// individually valid but sum to more than the maximum representable
+	// amount. Unlike the per-output bounds, this one has no counterpart in
+	// validateTxIntent, so it is reported from the authoring boundary.
+	ErrOutputTotalExceedsMax = errors.New(
+		"total output amount exceeds maximum value",
+	)
+
+	// ErrFeeOutOfRange is returned when the fee a transaction would pay is
+	// not a representable amount. DefaultMaxFeeRate keeps this out of reach
+	// at its default value, but that bound is a mutable package variable
+	// intended to become configurable: raised far enough, an ordinary
+	// multi-input transaction can price a fee above MaxSatoshi.
+	ErrFeeOutOfRange = errors.New("transaction fee out of range")
 )
 
 var (
@@ -677,7 +698,7 @@ func (w *Wallet) authorTransaction(outputs []wire.TxOut,
 		txOutputs, feeSatPerKb, inputSource, changeSource,
 	)
 	if err != nil {
-		return nil, err
+		return nil, translateAuthorError(err)
 	}
 
 	// Randomize the position of the change output, if one was created. This
@@ -688,6 +709,80 @@ func (w *Wallet) authorTransaction(outputs []wire.TxOut,
 	}
 
 	return tx, nil
+}
+
+// authorError gives an authoring failure a wallet-facing identity without
+// restating it. Its message is the originating error's alone, while errors.Is
+// resolves both the wallet sentinel, through Is, and everything the originating
+// error already matched, through Unwrap. Wrapping with a second %w verb would
+// match both too, but would render the wallet sentinel's text and the txauthor
+// error's text one after the other, and for most of these classes those say the
+// same thing twice.
+//
+// This mirrors AmbiguousTxCommitError in wallet/internal/db/runtime.
+type authorError struct {
+	// sentinel is the wallet-facing identity this failure reports as.
+	sentinel error
+
+	// cause is the originating txauthor error, which supplies the message.
+	cause error
+}
+
+// Error returns the originating error's message.
+func (e *authorError) Error() string {
+	return e.cause.Error()
+}
+
+// Unwrap returns the originating txauthor error.
+func (e *authorError) Unwrap() error {
+	return e.cause
+}
+
+// Is reports whether target is the wallet sentinel this failure carries.
+func (e *authorError) Is(target error) bool {
+	return errors.Is(e.sentinel, target)
+}
+
+// translateAuthorError maps the checked-arithmetic errors the txauthor module
+// reports onto the wallet's own error vocabulary, so callers of the public
+// wrappers match on one set of sentinels. It is the single translation point
+// for both CreateTransaction and FundPsbt.
+//
+// Conditions validateTxIntent already names keep that identity, which costs no
+// new API and leaves one wallet-level error per condition however it was
+// reached: the txrules errors for an output value, and ErrMissingFeeRate for a
+// rate that is not positive. Only the output total has no intent-level
+// counterpart and needs a sentinel of its own.
+//
+// Everything else is returned untouched: an InputSourceError, so "cannot fund
+// this" stays distinguishable from "will not author this", and the checked
+// arithmetic classes no caller can observe. A nil output cannot occur, because
+// authorTransaction takes its outputs by value; the add and subtract guards
+// cannot fire, because CheckOutputsAmount bounds the target before every
+// addition and the sufficiency checks bound the rest. Naming those in the
+// wallet would be exported API with nothing behind it.
+func translateAuthorError(err error) error {
+	switch {
+	case errors.Is(err, txauthor.ErrOutputValueNegative):
+		return &authorError{txrules.ErrAmountNegative, err}
+
+	case errors.Is(err, txauthor.ErrOutputValueExceedsMax):
+		return &authorError{txrules.ErrAmountExceedsMax, err}
+
+	case errors.Is(err, txauthor.ErrOutputTotalExceedsMax):
+		return &authorError{ErrOutputTotalExceedsMax, err}
+
+	case errors.Is(err, txauthor.ErrFeeRateNotPositive):
+		return &authorError{ErrMissingFeeRate, err}
+
+	case errors.Is(err, txauthor.ErrFeeOverflow),
+		errors.Is(err, txauthor.ErrFeeOutOfRange):
+
+		return &authorError{ErrFeeOutOfRange, err}
+
+	default:
+		return err
+	}
 }
 
 // determineChangeSource determines the source for the transaction's change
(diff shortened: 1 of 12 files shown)


wallet/txauthor/amounts.go
```diff
@@ -0,0 +1,191 @@
...
+)
+
+// addAmounts returns the sum of two amounts, reporting ErrAmountOverflow
+// rather than wrapping when the sum is not representable.
+func addAmounts(a, b btcutil.Amount) (btcutil.Amount, error) {
+	sum := a + b
+
+	// A sum that moved the wrong way relative to the sign of the addend is
+	// the signed-overflow signature.
+	if (b > 0 && sum < a) || (b < 0 && sum > a) {
```
yyforyongyu, line 73: 👍

wallet/txauthor/amounts.go
```diff
@@ -0,0 +1,191 @@
...
+// MaxSatoshi.
+//
+// A nil or empty set is valid and totals zero: a sweep authors a transaction
+// whose only output is change, and has no non-change outputs to declare.
+//
+// The running sum is added through addAmounts as defence in depth. Because
+// every accepted element is non-negative and bounded by MaxSatoshi, the
+// aggregate bound below is reached long before the sum could overflow, so that
+// guard is not expected to fire on any set built from accepted elements.
+func CheckOutputs(outputs []*wire.TxOut) (btcutil.Amount, error) {
```
yyforyongyu: nit: `CheckOutputsAmount`

wallet/txauthor/amounts.go
```diff
@@ -0,0 +1,191 @@
...
+			return 0, fmt.Errorf("%w: index %d has %d",
+				ErrOutputValueNegative, i, output.Value)
+		}
+
+		if output.Value > btcutil.MaxSatoshi {
+			return 0, fmt.Errorf("%w: index %d has %d",
+				ErrOutputValueExceedsMax, i, output.Value)
+		}
+
+		sum, err := addAmounts(total, btcutil.Amount(output.Value))
```
yyforyongyu: The aggregate limit is checked only after the loop, so enough individually valid outputs can overflow int64 here first. For example, 4,393 outputs at MaxSatoshi wrap the accumulator and return ErrAmountOverflow instead of ErrOutputTotalExceedsMax; the wallet translator then leaks that nested error. Please reject before addition, for example by checking total > MaxSatoshi-value, and add a many-output regression test.

wallet/txauthor/author.go
```diff
@@ -18,6 +18,10 @@ import (
 )
 
 // SumOutputValues sums up the list of TxOuts and returns an Amount.
+//
+// Deprecated: this sum is unchecked. It ignores a nil element, admits a
```
yyforyongyu: This documentation says a nil element is ignored, but SumOutputValues dereferences txOut on line 27 and therefore panics. Please remove the claim or make the deprecated function actually skip nil entries.

wallet/tx_creator.go
```diff
@@ -690,6 +716,79 @@ func (w *Wallet) authorTransaction(outputs []wire.TxOut,
...
+		return &authorError{txrules.ErrAmountNegative, err}
+
+	case errors.Is(err, txauthor.ErrOutputValueExceedsMax):
+		return &authorError{txrules.ErrAmountExceedsMax, err}
+
+	case errors.Is(err, txauthor.ErrOutputTotalExceedsMax):
+		return &authorError{ErrOutputTotalExceedsMax, err}
+
+	case errors.Is(err, txauthor.ErrFeeRateNotPositive):
+		return &authorError{ErrInvalidFeeRate, err}
```
yyforyongyu: validateTxIntent already rejects every nonpositive public fee rate as ErrMissingFeeRate before authorTransaction runs. As a result, ErrInvalidFeeRate is a second wallet sentinel for the same condition and is not observable through CreateTransaction or FundPsbt. Please keep one wallet-level identity and map the txauthor cause to ErrMissingFeeRate.

wallet/txauthor/author_test.go
```diff
@@ -222,3 +228,160 @@ func TestNewUnsignedTransaction(t *testing.T) {
...
+			require.ErrorIs(t, err, tc.wantErr)
+			require.Nil(t, tx)
+			require.Zero(t, sources.inputCalls)
+			require.Zero(t, sources.changeCalls)
+		})
+	}
+}
+
+// TestNewUnsignedTransactionNilOutputs verifies that a transaction with no
+// non-change outputs still authors. A sweep pays its entire input value to a
```
yyforyongyu: This fixture funds exactly the requested fee and then asserts no outputs, so it only proves a nil slice is accepted; it does not exercise the documented change-only sweep. Please fund above the fee and assert one change output, or retitle and reword the test to match its actual coverage.

wallet/tx_creator_test.go
```diff
@@ -732,6 +734,246 @@ func TestAuthorTransaction(t *testing.T) {
...
+// running-sum overflow cannot occur once the aggregate bound is enforced; and
+// no SatPerKVByte the wallet can build is large enough to price a fee out of
+// range. Those are covered in the nested module and, for the translation
+// itself, by TestTranslateAuthorError.
+func TestAuthorTransactionChecksAmounts(t *testing.T) {
+	t.Parallel()
+
+	script := append([]byte{0x00, 0x14}, make([]byte, 20)...)
+
+	testCases := []struct {
```
yyforyongyu: These negative, oversized, and zero-rate cases are already screened by validateTxIntent, while TestTranslateAuthorError separately covers their mappings. Only aggregate total exercises a distinct authorTransaction boundary. Could this test retain aggregate plus callback-ordering and avoid duplicating the other two layers?

wallet/txauthor/amounts_test.go
```diff
@@ -0,0 +1,344 @@
...
+		size    int
+		want    btcutil.Amount
+		wantErr error
+	}{{
+		name: "an ordinary rate and size",
+		rate: 1e3,
+		size: 1000,
+		want: 1e3,
+	}, {
+		name: "a fee rounds down to the whole satoshi",
```
yyforyongyu: The rate 1000 and size 1500 produce exactly 1500 after division; no fractional satoshi is discarded. Please choose a product with a non-zero remainder so this case actually verifies truncation.

yyforyongyu (changes requested):
Very nice! Did a commit by commit review and left some comments - re the scope, initially I thought we could instead just fix `TxCreator` and ignore all the other deprecated usage, such that we only need to fix it in `TxCreator` - but given that is defined in the task already, plus it's surprisingly easy to review, will go with this route instead!

wallet/txauthor/amounts.go
Abdulkbk: changed

wallet/txauthor/author.go
Abdulkbk: updated the comment

yyforyongyu (approved):
LGTM!

PR #1349, opened 2026-08-24
itest: add Controller credential integration tests

GustavoStingelin:
## Summary

Implements roadmap Task 427 by adding Controller credential integration tests:
https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/427-controller-lifecycle-credential-itests.md

Also fixes the empty-new-private-passphrase behavior found while adding the
invalid credential coverage: `Vault.ChangePassphrase` now rejects an empty
`PrivateNew` for both SQL and kvdb, and RPC maps that stable error to
`InvalidArgument`.

## Change Description

Adds focused controller cases for the credential contracts:

- private `ChangePassphrase` succeeds while locked, rejects the old
  passphrase, accepts the new passphrase, and keeps the rotated credential
  authoritative after a durable reload;
- private `ChangePassphrase` succeeds while unlocked without changing the
  public unlocked state, then rejects the old passphrase after re-lock;
- lifecycle gates reject passphrase changes before `Start` and after `Stop`
  with `wallet.ErrStateForbidden`, while proving rejected mutations do not
  install the proposed credential;
- wrong current credentials are rejected with `wallet.ErrInvalidPassphrase`
  while locked and while unlocked, and the original credential still works;
- empty new private credentials are rejected with `wallet.ErrEmptyPassphrase`
  at the public Controller/Wallet boundary, and the original credential still
  works; and
- a positive `Unlock` timeout returns the wallet to the public locked state.

The empty-passphrase rejection is enforced inside the SQL and kvdb Vault
`ChangePassphrase` implementations so both backends share the same contract.

Every controller case is registered once and runs unchanged across kvdb,
SQLite and PostgreSQL. Assertions stay at the public Controller/Wallet
boundary; the tests do not reach into stores directly and do not branch by
backend.

This branch uses the `ReloadWallet` helper introduced by #1326. The branch has
been rebased onto `sql-wallet`, so the helper now comes from the base branch
rather than from a cherry-picked copy in this PR.

## Testing

```bash
make itest db=sqlite icase='controller (change passphrase|timed unlock)'
make itest db=kvdb icase='controller (change passphrase|timed unlock)'
make itest db=postgres icase='controller (change passphrase|timed unlock)'
```

### DIFF
--- a/itest/controller_credential_test.go
+++ b/itest/controller_credential_test.go
@@ -1,0 +1,261 @@
+//go:build itest
+
+package itest
+
+import (
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// testControllerChangePassphraseLocked verifies private passphrase rotation
+// while locked and durable credential enforcement after reload.
+func testControllerChangePassphraseLocked(h *bwtest.HarnessTest) {
+	const (
+		oldPassphrase = bwtest.TestWalletPrivatePassphrase
+		newPassphrase = "controller-new-private-passphrase"
+	)
+
+	w, _ := h.NewWallet(bwtest.WalletFixture{})
+	requireLocked(h, w, true)
+
+	err := w.ChangePassphrase(h.Context(), wallet.ChangePassphraseRequest{
+		ChangePrivate: true,
+		PrivateOld:    []byte(oldPassphrase),
+		PrivateNew:    []byte(newPassphrase),
+	})
+	require.NoError(h, err, "failed to change private passphrase")
+	requireLocked(h, w, true)
+
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(oldPassphrase),
+		Timeout:    -1,
+	})
+	require.ErrorIs(
+		h, err, wallet.ErrInvalidPassphrase,
+		"old passphrase should be rejected after change",
+	)
+	requireLocked(h, w, true)
+
+	// A zero timeout uses the default auto-lock duration; -1 keeps the wallet
+	// unlocked until Lock so credential assertions cannot race a timer.
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(newPassphrase),
+		Timeout:    -1,
+	})
+	require.NoError(h, err, "new passphrase should unlock wallet")
+	requireLocked(h, w, false)
+
+	// ReloadWallet crosses the durable close-and-reopen boundary and returns
+	// a fresh started wallet that must require the rotated credential.
+	w = h.ReloadWallet(w)
+	requireLocked(h, w, true)
+
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(oldPassphrase),
+		Timeout:    -1,
+	})
+	require.ErrorIs(
+		h, err, wallet.ErrInvalidPassphrase,
+		"old passphrase should stay rejected after reload",
+	)
+	requireLocked(h, w, true)
+
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(newPassphrase),
+		Timeout:    -1,
+	})
+	require.NoError(h, err, "new passphrase should unlock reloaded wallet")
+	requireLocked(h, w, false)
+}
+
+// testControllerChangePassphraseUnlocked verifies private passphrase rotation
+// while unlocked preserves the public unlocked state.
+func testControllerChangePassphraseUnlocked(h *bwtest.HarnessTest) {
+	const (
+		oldPassphrase = bwtest.TestWalletPrivatePassphrase
+		newPassphrase = "controller-unlocked-private-passphrase"
+	)
+
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	requireLocked(h, w, false)
+
+	err := w.ChangePassphrase(h.Context(), wallet.ChangePassphraseRequest{
+		ChangePrivate: true,
+		PrivateOld:    []byte(oldPassphrase),
+		PrivateNew:    []byte(newPassphrase),
+	})
+	require.NoError(
+		h, err, "failed to change private passphrase while unlocked",
+	)
+	requireLocked(h, w, false)
+
+	require.NoError(h, w.Lock(h.Context()), "failed to lock wallet")
+	requireLocked(h, w, true)
+
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(oldPassphrase),
+		Timeout:    -1,
+	})
+	require.ErrorIs(
+		h, err, wallet.ErrInvalidPassphrase,
+		"previous passphrase should be rejected after unlocked change",
+	)
+	requireLocked(h, w, true)
+
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(newPassphrase),
+		Timeout:    -1,
+	})
+	require.NoError(
+		h, err, "unlocked change passphrase should unlock wallet",
+	)
+	requireLocked(h, w, false)
+}
+
+// testControllerChangePassphraseLifecycle verifies lifecycle state gates for
+// passphrase rotation.
+func testControllerChangePassphraseLifecycle(h *bwtest.HarnessTest) {
+	const (
+		oldPassphrase  = bwtest.TestWalletPrivatePassphrase
+		nextPassphrase = "controller-next-private-passphrase"
+	)
+
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unstarted: true})
+
+	err := w.ChangePassphrase(h.Context(), wallet.ChangePassphraseRequest{
+		ChangePrivate: true,
+		PrivateOld:    []byte(oldPassphrase),
+		PrivateNew:    []byte(nextPassphrase),
+	})
+	require.ErrorIs(
+		h, err, wallet.ErrStateForbidden,
+		"change passphrase before start not rejected",
+	)
+
+	require.NoError(h, w.Start(h.Context()), "failed to start wallet")
+	requireLocked(h, w, true)
+
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(nextPassphrase),
+		Timeout:    -1,
+	})
+	require.ErrorIs(
+		h, err, wallet.ErrInvalidPassphrase,
+		"pre-start rejection should not install new passphrase",
+	)
+	requireLocked(h, w, true)
+
+	err = w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(oldPassphrase),
+		Timeout:    -1,
+	})
+	require.NoError(
+		h, err, "old passphrase should survive pre-start rejection",
+	)
+	requireLocked(h, w, false)
+
+	require.NoError(h, w.Lock(h.Context()), "failed to lock wallet")
+	requireLocked(h, w, true)
+
+	require.NoError(h, w.Stop(h.Context()), "failed to stop wallet")
+
+	err = w.ChangePassphrase(h.Context(), wallet.ChangePassphraseRequest{
+		ChangePrivate: true,
+		PrivateOld:    []byte(oldPassphrase),
+		PrivateNew:    []byte(nextPassphrase),
+	})
+	require.ErrorIs(
+		h, err, wallet.ErrStateForbidden,
+		"change passphrase after stop not rejected",
+	)
+
+	require.NoError(h, w.Start(h.Context()), "failed to restart wallet")
+	requireLocked(h, w, true)
+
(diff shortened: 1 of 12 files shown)


yyforyongyu (changes requested):
Let's do a rebase first? Also posting some agentic findings now

itest/controller_credential_test.go
```diff
@@ -0,0 +1,251 @@
...
+	)
+
+	w, _ := h.NewWallet(bwtest.WalletFixture{})
+	requireLocked(h, w, true)
+
+	err := w.ChangePassphrase(h.Context(), wallet.ChangePassphraseRequest{
+		ChangePrivate: true,
+		PrivateOld:    []byte(oldPassphrase),
+		PrivateNew:    []byte(newPassphrase),
+	})
```
yyforyongyu, line 30: Task 427 requires invalid-new-credential coverage, but every new `ChangePassphrase` request supplies a valid non-empty new passphrase; please add a backend-neutral public rejection case that asserts the stable error identity and proves the original credential remains authoritative, or reconcile the roadmap requirement before merging if the public API intentionally has no invalid private new credential.

itest/controller_credential_test.go
```diff
@@ -0,0 +1,251 @@
...
+
+	err = w.ChangePassphrase(h.Context(), wallet.ChangePassphraseRequest{
+		ChangePrivate: true,
+		PrivateOld:    []byte(bwtest.TestWalletPrivatePassphrase),
+		PrivateNew:    []byte(nextPassphrase),
+	})
+	require.ErrorIs(
+		h, err, wallet.ErrStateForbidden,
+		"change passphrase after stop not rejected",
+	)
```
yyforyongyu, line 175: This lifecycle test asserts `ErrStateForbidden` before `Start` and after `Stop` but never checks persisted credential state, so it can pass if either rejected mutation partially changes the wallet; please restart after each rejection and verify the proposed credential is rejected while the original still unlocks.

itest/controller_credential_test.go
```diff
@@ -0,0 +1,251 @@
...
+	requireLocked(h, w, true)
+
+	// Use a positive timeout long enough that the immediate unlocked
+	// assertion observes the post-Unlock state before the timer can fire.
+	err := w.Unlock(h.Context(), wallet.UnlockRequest{
+		Passphrase: []byte(bwtest.TestWalletPrivatePassphrase),
+		Timeout:    time.Second,
+	})
+	require.NoError(h, err, "failed to unlock wallet")
+	requireLocked(h, w, false)
```
yyforyongyu, line 348: This requires a database-backed `Info` call to observe the unlocked state before a one-second auto-lock timer fires, which can fail under scheduler or database pauses despite correct behavior; please use a larger positive timeout while retaining bounded observations of both the unlocked and eventual locked states.

GustavoStingelin:
I think I only did the rebase in my mind 🤣

itest/controller_credential_test.go
Abdulkbk: Creation rejects empty on both backends, but `ChangePassphrase` does not. Also, `errEmptyPassphrase` isn't exported, so we can't test it here. Perhaps we should create a task to reject an empty passphrase when rotating, unless we have a use case for it.

itest/controller_credential_test.go
Abdulkbk: +1

itest/controller_credential_test.go
```diff
@@ -0,0 +1,214 @@
...
+		ChangePrivate: true,
+		PrivateOld:    []byte(bwtest.TestWalletPrivatePassphrase),
+		PrivateNew:    []byte(nextPassphrase),
+	})
+	require.ErrorIs(
+		h, err, wallet.ErrStateForbidden,
+		"change passphrase before start not rejected",
+	)
+
+	require.NoError(h, w.Start(h.Context()), "failed to start wallet")
```
Abdulkbk, line 140: According to task https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/368-wallet-startup-shutdown.md:
> Remove Wallet.Start, Wallet.Stop, Controller.Start, Controller.Stop, Manager.StartWallet, Manager.StopWallet, Manager.Load, Manager.Unload, and Manager.Close from the maintained public surface. Retain no wrapper, deprecated alias, restart generation, or parallel lifecycle interface solely for compatibility.

It seems we're removing `w.Start` and `w.Stop`, which are used in this test. Do we leave as is then remove when task 368 lands?. I'm unsure

Abdulkbk (changes requested):

itest/controller_credential_test.go
GustavoStingelin: Hmm, good question. I would leave the tests here, so when we pick that one, we can explicitly see the tests failing and adjust them

itest/controller_credential_test.go
GustavoStingelin: Added this in the latest push. ChangePassphrase now rejects an empty PrivateNew at the Vault boundary for both SQL and kvdb via wallet.ErrEmptyPassphrase/keyvault.ErrEmptyPassphrase, RPC maps it to InvalidArgument, and the controller itest asserts the public error while proving the original credential still unlocks.

itest/controller_credential_test.go
GustavoStingelin: Updated. The timed unlock case now uses a larger positive timeout while still checking both states: first that the wallet becomes observable as unlocked, then that it locks again after the timeout.

itest/controller_credential_test.go
GustavoStingelin: Updated. The pre-start and post-stop rejection paths now verify the proposed passphrase is not accepted and that the original private passphrase remains authoritative after the rejected mutation.

itest/controller_credential_test.go
GustavoStingelin: Leaving this as-is for this PR. The case documents the current public lifecycle behavior required by Task 427; when Task 368 removes that lifecycle surface, the failing test will make the required test migration explicit in that task instead of broadening this PR.

itest/controller_credential_test.go
GustavoStingelin: How large should it be? how to calculate it?

itest/controller_credential_test.go
```diff
@@ -0,0 +1,327 @@
...
+	})
+	require.NoError(
+		h, err, "old passphrase should survive unlocked rejection",
+	)
+	requireLocked(h, w, false)
+}
+
+// testControllerTimedUnlock verifies that a positive Unlock timeout returns the
+// Wallet to its public locked state.
+func testControllerTimedUnlock(h *bwtest.HarnessTest) {
```
yyforyongyu: nit: `testControllerTimedUnlock` -> `testControllerUnlockTimeout`

wallet/internal/db/kvdb/vault.go
```diff
@@ -141,43 +139,69 @@ func (v *LegacyWalletVault) ChangePassphrase(ctx context.Context,
 		return err
 	}
 
-	changePublic := params.PublicOld != nil
+	changePublic, changePrivate, err := validateChangePassphraseParams(params)
```
yyforyongyu: The empty-private-new rule now has separate implementations here and in WalletVault, and their precedence can already differ because kvdb checks context before validation; move the backend-neutral rule into ChangePassphraseParams.Validate in keyvault, document ErrEmptyPassphrase on the request and Vault contract, and have both implementations call that single validation before any secret read or mutation.

rpc/rpcserver/server.go
```diff
@@ -82,16 +82,18 @@ func errorCode(err error) codes.Code {
 		err = e.Err
 	}
 
-	switch err {
-	case wallet.ErrLoaded:
+	switch {
+	case errors.Is(err, wallet.ErrEmptyPassphrase):
```
yyforyongyu: This PR exercises the new Controller Go API, but this hunk changes the legacy gRPC service. Its ChangePassphrase handler still calls the deprecated Wallet.ChangePrivatePassphrase and Wallet.ChangePublicPassphrase methods rather than Controller.ChangePassphrase, so wallet.ErrEmptyPassphrase cannot reach this translator. Please remove this RPC mapping, restore the unrelated sentinel comparisons, and remove the RPC claim from the PR description; legacy gRPC migration or parity should be handled separately.

yyforyongyu: in other words - we should not need to touch any of the RPCs as they are using the old deprecated methods intentionally

itest/controller_credential_test.go
yyforyongyu: yeah as long as it's less than the polling timeout 30s we are good.

yyforyongyu (changes requested):
Cool very close! My main comment is we should not need to touch the `rpc` package as they are using deprecated APIs

itest/controller_credential_test.go
GustavoStingelin: renamed

wallet/internal/db/kvdb/vault.go
GustavoStingelin: now, we have just one method for struct validation

rpc/rpcserver/server.go
GustavoStingelin: removed

wallet/internal/keyvault/keyvault.go
```diff
@@ -90,3 +96,19 @@ type ChangePassphraseParams struct {
...
+// Validate verifies backend-neutral passphrase rotation requirements.
+func (p ChangePassphraseParams) Validate() error {
+	publicSelected := p.PublicOld != nil
+	privateSelected := p.PrivateOld != nil
+
+	if !publicSelected && !privateSelected {
+		return ErrEmptyPassphrase
+	}
+
+	if privateSelected && (len(p.PrivateOld) == 0 || len(p.PrivateNew) == 0) {
```
yyforyongyu: This shared rule rejects an empty current private credential, but creation does not enforce the same invariant: `CreateWalletParams.validate` (including PR #1353) does not check `PrivatePassphrase`, and SQL `validateGenesisInputs` returns success immediately for watch-only wallets even when it is empty. That allows a ModeShell SQL wallet to be created in a state that can never pass this rotation validation. Since SQL wallets are not deployed and empty credentials should be forbidden, please reject an empty private passphrase when creating SQL watch-only wallets and add negative coverage. Otherwise this validation must allow an empty old credential; creation and rotation need one consistent contract.

yyforyongyu (approved):
LGTM! Found one thing we should fix - so in sql we allow creating wallets using empty passphrase, which should be unified here. Feel free to fix it here or in your next PR.

wallet/internal/keyvault/keyvault.go
Abdulkbk: @GustavoStingelin do you plan to address this here or in a follow up?

wallet/internal/db/kvdb/vault.go
```diff
@@ -134,50 +134,61 @@ func (v *LegacyWalletVault) crypt(op func(waddrmgr.CryptoKeyType,
 func (v *LegacyWalletVault) ChangePassphrase(ctx context.Context,
 	params keyvault.ChangePassphraseParams) error {
 
+	err := params.Validate()
+	if err != nil {
+		return err
```
Abdulkbk: in [vault_lifecycle.go](https://github.com/btcsuite/btcwallet/pull/1349/changes#diff-e77bd142c8926ddf934bfa2ecc1b7c683978b924505629aefc199cf59a28a17aR226-R231), the error carries some additional context:

```golang
	err := params.Validate()
	if err != nil {
		return fmt.Errorf("wallet %d vault ChangePassphrase: %w",
			v.walletID, err)
	}
```

wallet/internal/keyvault/keyvault.go
```diff
@@ -90,3 +96,19 @@ type ChangePassphraseParams struct {
...
 	PrivateNew []byte
 }
+
+// Validate verifies backend-neutral passphrase rotation requirements.
+func (p ChangePassphraseParams) Validate() error {
+	publicSelected := p.PublicOld != nil
+	privateSelected := p.PrivateOld != nil
+
+	if !publicSelected && !privateSelected {
+		return ErrEmptyPassphrase
```
Abdulkbk: Isn't `ErrEmptyPassphrase` overloaded in this func?

Abdulkbk (commented):
Looks good, pending your response to https://github.com/btcsuite/btcwallet/pull/1349/changes#r3928005485

wallet/internal/keyvault/keyvault.go
GustavoStingelin: In a follow-up

wallet/internal/db/kvdb/vault.go
GustavoStingelin: added formatted error to kvdb too

wallet/internal/keyvault/keyvault.go
GustavoStingelin: added more context to each if branch, formatting the error

wallet/internal/keyvault/keyvault.go
GustavoStingelin: follow up pr https://github.com/btcsuite/btcwallet/pull/1364

Abdulkbk (approved):
LGTM

PR #1351, opened 2026-08-26
wallet: retain output leases through spend confirmation

bhandras:
## Change Description

Add an opt-in output lease mode that releases a locked output only after its
spending transaction reaches a requested confirmation depth.

The existing lease is wall-clock based. It prevents concurrent coin selection
while an operation is being assembled, but btcwallet removes it as soon as a
confirmed spend is observed. That is correct for ordinary callers. It is not
enough for callers that must preserve the identity of a previously broadcast
transaction across a shallow chain reorganization. If the spend is disconnected,
the input becomes available again and another wallet transaction can consume it.

This PR keeps the default behavior unchanged. A caller opts in through
`LeaseOutputWithOptions` and `wtxmgr.WithReleaseAfterSpend(confirmations)`.

The retained lease follows this lifecycle:

1. Before the output is spent, the normal wall-clock expiry applies.
2. When btcwallet observes a confirmed spend, it records the spend height and
   disables wall-clock expiry for that lease.
3. A rollback that disconnects the spend clears its confirmation progress while
   keeping the lease active.
4. A later confirmation records the new height.
5. The lease is removed only when that spend reaches the requested depth, or
   when its owner explicitly calls `ReleaseOutput`.

This owns the invariant at the wallet boundary. Callers do not need a separate
UTXO tracker or a loop that repeatedly renews leases while blocks connect and
disconnect.

### Storage compatibility

The locked-output value keeps its existing `(lock ID, expiry)` prefix and appends
the release depth and observed spend height. Legacy rows remain readable and
decode with a zero release depth, which preserves the existing behavior. No
migration is required.

Compatibility is one-way while a retained lease is active. Older software can
decode the prefix but applies time-only release semantics. Do not downgrade
while a retained lease protects a spend below its configured confirmation
depth.

The root module temporarily replaces the standalone `wtxmgr` dependency with
the in-tree module. That replace can be removed after a `wtxmgr` version
containing the first commit is published.

### Compatibility and limits

- `LeaseOutput` is unchanged and delegates to the new options method with no
  options.
- A zero release depth preserves the existing lease behavior.
- Explicit release remains available after the output is spent.
- The protection horizon is the requested confirmation depth. Once the lease is
  released, a reorganization at or beyond that depth is outside this mechanism's
  guarantee.

## Steps to Test

```bash
cd wtxmgr
go test ./...

cd ..
go test ./wallet
```

`TestOutputLockReleaseAfterSpendAcrossReorg` covers confirmation, rollback,
wall-clock expiry while disconnected, reconfirmation at a new height, and final
release. `TestDeserializeLegacyLockedOutput` covers the upgrade path.

Local result:

- `go test ./...` in `wtxmgr`: pass
- `go test ./wallet`: pass

### Testing

- [x] Tests covering the positive and negative paths are included.
- [x] The reorganization regression is covered by a test.

- [x] The change is substantial and scoped to output lease lifetime.
- [x] Exported additions are documented and formatted with `gofmt`.
- [x] Commits follow the ideal structure and are signed.
- [x] No logging statements were added.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,5 +1,9 @@
 module github.com/btcsuite/btcwallet
 
+// TODO: Remove this replace after publishing a wtxmgr release that contains
+// release-after-spend output leases.
+replace github.com/btcsuite/btcwallet/wtxmgr => ./wtxmgr
+
 require (
 	github.com/btcsuite/btcd v0.26.0
 	github.com/btcsuite/btcd/address/v2 v2.0.0
--- a/wallet/chainntfns.go
+++ b/wallet/chainntfns.go
@@ -33,56 +33,6 @@ func (w *Wallet) handleChainNotifications() {
 		return
 	}
 
-	catchUpHashes := func(w *Wallet, client chain.Interface,
-		height int32) error {
-		// TODO(aakselrod): There's a race condition here, which
-		// happens when a reorg occurs between the
-		// rescanProgress notification and the last GetBlockHash
-		// call. The solution when using btcd is to make btcd
-		// send blockconnected notifications with each block
-		// the way Neutrino does, and get rid of the loop. The
-		// other alternative is to check the final hash and,
-		// if it doesn't match the original hash returned by
-		// the notification, to roll back and restart the
-		// rescan.
-		log.Infof("Catching up block hashes to height %d, this"+
-			" might take a while", height)
-		err := walletdb.Update(w.db, func(tx walletdb.ReadWriteTx) error {
-			ns := tx.ReadWriteBucket(waddrmgrNamespaceKey)
-
-			startBlock := w.Manager.SyncedTo()
-
-			for i := startBlock.Height + 1; i <= height; i++ {
-				hash, err := client.GetBlockHash(int64(i))
-				if err != nil {
-					return err
-				}
-				header, err := chainClient.GetBlockHeader(hash)
-				if err != nil {
-					return err
-				}
-
-				bs := waddrmgr.BlockStamp{
-					Height:    i,
-					Hash:      *hash,
-					Timestamp: header.Timestamp,
-				}
-				err = w.Manager.SetSyncedTo(ns, &bs)
-				if err != nil {
-					return err
-				}
-			}
-			return nil
-		})
-		if err != nil {
-			log.Errorf("Failed to update address manager "+
-				"sync state for height %d: %v", height, err)
-		}
-
-		log.Info("Done catching up block hashes")
-		return err
-	}
-
 	waitForSync := func(birthdayBlock *waddrmgr.BlockStamp) error {
 		// We start with a retry delay of 0 to execute the first attempt
 		// immediately.
@@ -194,7 +144,7 @@ func (w *Wallet) handleChainNotifications() {
 			// The following require some database maintenance, but also
 			// need to be reported to the wallet's rescan goroutine.
 			case *chain.RescanProgress:
-				err = catchUpHashes(w, chainClient, n.Height)
+				err = w.catchUpHashes(chainClient, n.Height)
 				notificationName = "rescan progress"
 				select {
 				case w.rescanNotifications <- n:
@@ -202,7 +152,7 @@ func (w *Wallet) handleChainNotifications() {
 					return
 				}
 			case *chain.RescanFinished:
-				err = catchUpHashes(w, chainClient, n.Height)
+				err = w.catchUpHashes(chainClient, n.Height)
 				notificationName = "rescan finished"
 				w.SetChainSynced(true)
 				select {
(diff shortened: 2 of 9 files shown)


go.mod
```diff
@@ -1,5 +1,7 @@
 module github.com/btcsuite/btcwallet
 
+replace github.com/btcsuite/btcwallet/wtxmgr => ./wtxmgr
```
GustavoStingelin, line 5: Non-blocking: release reminder

wallet/chainntfns.go
```diff
@@ -253,6 +254,13 @@ func (w *Wallet) connectBlock(dbtx walletdb.ReadWriteTx, b wtxmgr.BlockMeta) err
 		return err
 	}
```
GustavoStingelin: This maturity cleanup only runs for live `BlockConnected` notifications. `RescanFinished` advances the synced height through `catchUpHashes` without applying maturity cleanup, so a lease that reaches its requested depth while the wallet is offline remains indefinitely when reconnect finishes at a static tip. Please run `DeleteMaturedLockedOutputs` at successful rescan completion and cover that path with a static-tip reconnect test.

wtxmgr/tx.go
```diff
@@ -129,9 +129,29 @@ type TxRecord struct {
...
+}
+
+// LockOutputOption configures the lifetime of an output lock.
+type LockOutputOption func(*lockOutputOptions)
+
+type lockOutputOptions struct {
+	releaseAfterSpendConfs uint32
+}
+
+// WithReleaseAfterSpend keeps an output locked until its spending transaction
```
GustavoStingelin: The PR description says normal wall-clock expiry still applies before spend, but exported `WithReleaseAfterSpend` says it keeps the output locked until the confirmation count, which reads as unconditional. The newly exported `ConfirmedSpendHeight` also exposes `0`, `-1`, and positive states without documentation. Please document the pre-spend expiry condition and these sentinel meanings so external callers can correctly interpret the contract.

wtxmgr/tx_test.go
```diff
@@ -2901,3 +2901,154 @@ func TestOutputLocks(t *testing.T) {
...
+
+		if err := store.Rollback(ns, spendBlock.Height); err != nil {
+			return err
+		}
+		if err := store.RemoveUnminedTx(ns, spendRec); err != nil {
+			return err
+		}
+
+		assertLocked(t, ns, creditOutpoint, store.clock.Now(), true)
+		assertUtxos(t, store, ns, nil)
```
GustavoStingelin, line 3113: `assertUtxos(..., nil)` is a no-op here because the helper only loops over expected entries. If the intent is to verify that `RemoveUnminedTx` restored the funding output, please pass the expected outpoint or assert the exact result length and contents; otherwise remove this misleading assertion.

GustavoStingelin (changes requested):
just left small suggestions

go.mod
bhandras: Added a TODO beside the temporary replace so its removal is tied to publishing the wtxmgr release.

wtxmgr/tx_test.go
bhandras: Fixed the shared helper to require the exact UTXO count. This nil expectation now asserts zero spendable UTXOs; the funding credit is restored, but the retained lease intentionally keeps it out of UnspentOutputs.

wtxmgr/tx.go
bhandras: Updated the exported documentation. It now states that wall-clock expiry applies before the first confirmed spend, documents zero as wall-clock-only behavior, and defines the 0, -1, and positive ConfirmedSpendHeight states.

GustavoStingelin (approved):
👍🏼

wtxmgr/tx.go
```diff
@@ -1300,7 +1358,9 @@ func (s *Store) LockOutput(ns walletdb.ReadWriteBucket, id LockID,
 	}
 
 	expiry := s.clock.Now().Add(duration)
-	if err := lockOutput(ns, id, op, expiry); err != nil {
+	if err := lockOutput(
+		ns, id, op, expiry, opts.releaseAfterSpendConfs, 0,
```
yyforyongyu: Correctness: every renewal writes `spendHeight` as `0`, so after rollback restores the credit and records `-1`, a same-ID renewal re-enables wall-clock expiry and may delete the retained lease before reconfirmation; fetch the existing record and preserve its spend state and release depth unless explicitly overridden.

wallet/chainntfns.go
```diff
@@ -253,6 +254,13 @@ func (w *Wallet) connectBlock(dbtx walletdb.ReadWriteTx, b wtxmgr.BlockMeta) err
...
 
+	err = w.TxStore.DeleteMaturedLockedOutputs(
```
yyforyongyu, line 265: Correctness: maturity cleanup is only added to live `connectBlock`, while `catchUpHashes` advances an offline wallet directly to the rescan tip without calling it, so a lease that matured offline remains locked until another block; run the same cleanup at the validated catch-up tip and cover offline catch-up with a regression test.

yyforyongyu (commented):
Two correctness issues from the completed review:

wallet/chainntfns.go
bhandras: Fixed. Offline hash catch-up now runs maturity cleanup at the validated catch-up height in the same database transaction as the sync-tip update. The regression matures a retained lease through catchUpHashes without a live block notification.

wtxmgr/tx.go
bhandras: Fixed. Same-ID renewals now preserve the persisted spend state and release depth when the option is omitted; an explicit WithReleaseAfterSpend value replaces only the depth. The regression covers renewal immediately after rollback and an explicit override.

wtxmgr/tx.go
```diff
@@ -1299,8 +1360,28 @@ func (s *Store) LockOutput(ns walletdb.ReadWriteBucket, id LockID,
...
+	// Preserve its release depth as well unless the caller explicitly
+	// replaces it.
+	var (
+		releaseAfterSpendConfs = opts.releaseAfterSpendConfs
+		spendHeight            int32
+	)
+
+	existingID, _, existingReleaseConfs, existingSpendHeight, exists :=
+		fetchLockedOutput(ns, op)
+	if exists && existingID == id {
```
yyforyongyu: Correctness: this carry-over condition accepts any persisted same-ID record, even when the lease has expired and `isLocked` above is false. Because expired rows are removed lazily and callers can reuse a `LockID`, a later option-free `LeaseOutput` can inherit the stale `WithReleaseAfterSpend` depth and retain its next confirmed spend unexpectedly. Preserve the old policy only for an active same-owner lease (`isLocked && lockedID == id`), and add a regression covering reuse after expiry.

yyforyongyu (commented):
One correctness issue found on the updated head:

wtxmgr/tx.go
bhandras: Fixed. State carry-over now requires an active lease owned by the same ID. An expired lazily retained row is treated as a new lease even when the LockID is reused, so option-free reuse resets the release depth and spend state. The regression expires a retained lease, reuses its ID, and verifies the next confirmed spend is not retained.

bhandras:
@yyforyongyu The latest finding is fixed and folded, and CI is green. Please re-review when you have a chance.

yyforyongyu (approved):
LGTM

wtxmgr/tx_test.go
marshalokos: It need to be out for spend and full functionality in all financial transactions

marshalokos:
Approved

PR #1352, opened 2026-08-29
wallet+waddrmgr+wtxmgr: add RemoveAccount

kaloudis:
## Change Description

Adds the ability to **remove a watch-only account** previously registered through
`ImportAccount`, together with every address derived from it and every unspent
output paying to one of those addresses.

Motivation: imported accounts (e.g. via lnd's `WalletKit.ImportAccount`) can
currently never be deleted — they clutter account listings and consume resources
forever. Unblocks [lightningnetwork/lnd#8654](https://github.com/lightningnetwork/lnd/issues/8654);
the corresponding lnd RPC is already implemented against the API added here
(`wallet.(*Wallet).RemoveAccount(scope waddrmgr.KeyScope, name string) error`).

The removal spans both stores atomically, in one `walletdb.Update`:

1. **`wtxmgr.(*Store).RemoveCredits(ns, []wire.OutPoint)`** (new): removes
   unspent credits — mined and unmined — by outpoint, decrementing the mined
   balance for each removed mined credit. Deleting the waddrmgr addresses alone
   is not enough: `Balance` counts credits unconditionally (total balance would
   over-report forever) and `ListUnspent` attributes unknown-address outputs to
   the `default` account. Transaction records are deliberately left in place —
   sibling credits of the same tx may still be ours, and both the balance
   block-walk and pkScript resolution read through the record. Leased outputs
   and outputs spent by unmined transactions are refused with `ErrInput`;
   unknown outpoints are ignored, making the call idempotent.

2. **`waddrmgr.(*ScopedKeyManager).RemoveAccount(ns, account)`** (new): deletes
   the account row, both name/ID index entries, every derived address row, the
   address-account index entries and used-address markers, then purges the
   in-memory caches. Only `dbWatchOnlyAccountRow` accounts can be removed —
   anything else is backed by the wallet's own key material. Reserved accounts
   and the default account are refused. The last-account counter is deliberately
   **not** rolled back, so account numbers are never reused, while the removed
   account's *name* becomes available again.

3. **`wallet.(*Wallet).RemoveAccount(scope, name)`** (new): resolves the
   account, collects its addresses' output scripts, matches them against
   `TxStore.OutputsToWatch` (which includes leased and unmined-spent outputs, so
   those fail the removal cleanly instead of being missed), calls
   `RemoveCredits`, then the waddrmgr removal — all in one db transaction so the
   two stores can never end up inconsistent.

`AccountBalances` previously assumed the account-number sequence has no gaps,
which removal introduces; it now skips missing accounts.

Semantics / caveats:

- Removal is allowed with a non-zero balance. Funds are unaffected — whoever
  holds the account's keys retains control, and re-importing the same xpub
  (plus rescan) restores tracking.
- Confirmed transaction *history* involving the account remains in the store;
  only unspent credits are withdrawn.
- Chain-backend notification filters are additive, so removed addresses stay
  watched until restart; notifications for them are dropped by the existing
  `ErrAddressNotFound` path in `addRelevantTx`.
- Removing the last account of a scope leaves the scoped key manager in place
  (symmetric with `ImportAccount`, which creates it on demand).

Module mechanics: `RemoveCredits` lives in the separately-tagged `wtxmgr`
module, so this PR re-adds the local
`replace github.com/btcsuite/btcwallet/wtxmgr => ./wtxmgr` directive, following
the repo's established workflow (cf. 59d3cf64, 302643a3, dropped in c6f1caf9).
After merge: tag `wtxmgr/v1.7.0`, bump the require, drop the replace.

## Steps to Test

```sh
cd wtxmgr && go test -run TestRemoveCredits ./... && cd ..
go test -run TestRemoveAccount ./waddrmgr/...
go test -run TestRemoveAccount ./wallet/...
```

New coverage, positive and negative paths:

- `wtxmgr`: `TestRemoveCredits` — mined + unmined removal, balance accounting,
  sibling-credit isolation, idempotency, lease refusal (and removal after
  release), unmined-spend refusal.
- `waddrmgr`: `TestRemoveAccount` — full deletion of account/indexes/addresses/
  used-markers, isolation of other accounts, name reusable / number not reused,
  reserved + default + unknown account guards; `TestRemoveAccountNonWatchOnly` —
  seed-derived accounts refused and left intact.
- `wallet`: `TestRemoveAccount` — end-to-end: import → deposit (imported +
  default account in one tx) → remove → balances and `ListUnspent` only reflect
  the default account → re-import same xpub under same name → guards.

Full `wtxmgr`, `waddrmgr` and `wallet` package suites pass. End-to-end against
lnd: the `RemoveAccount` RPC integration test (import Carol's xpub into Dave's
node, fund, remove with non-zero balance, assert no trace across restart,
re-import) passes against this branch via local replace directives.

### Testing

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -64,3 +64,11 @@ require (
 // If you change this please run `make lint` to see where else it needs to be
 // updated as well.
 go 1.25.11
+
+// We use a replace directive here for our internal wtxmgr module. This ensures
+// that we can freely move between tagged releases and development commits of
+// this module without needing to constantly update the go.mod file.
+//
+// TODO: Remove once a new wtxmgr version containing RemoveCredits has been
+// tagged and pinned here.
+replace github.com/btcsuite/btcwallet/wtxmgr => ./wtxmgr
--- a/go.sum
+++ b/go.sum
@@ -30,8 +30,6 @@ github.com/btcsuite/btcwallet/wallet/txsizes v1.3.0 h1:2W9qt0edMoX8crx0Wm4Cv+eAj
 github.com/btcsuite/btcwallet/wallet/txsizes v1.3.0/go.mod h1:42aE6+LMZSSEisQAa15Xml25ncuJFfhCrkcpB5OmkZk=
 github.com/btcsuite/btcwallet/walletdb v1.6.0 h1:Yund5XbdqFxNW7+R2Sxs02bMC5fMrmORj4GN8MV55no=
 github.com/btcsuite/btcwallet/walletdb v1.6.0/go.mod h1:q9xif0Csp52GVb3l252BbHCuyiCnuEbrPWu/HAsvaYc=
-github.com/btcsuite/btcwallet/wtxmgr v1.6.0 h1:ivSSnYCD4Kb5yAMZVyBA1VMYABIFcopPEcmHCrRZXcE=
-github.com/btcsuite/btcwallet/wtxmgr v1.6.0/go.mod h1:Raor7IBIwHSIKE9Lr5o+R9rwX7sRMHU1zjxgEQgn9h8=
 github.com/btcsuite/go-socks v0.0.0-20170105172521-4720035b7bfd h1:R/opQEbFEy9JGkIguV40SvRY1uliPX8ifOvi6ICsFCw=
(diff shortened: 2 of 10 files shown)


yyforyongyu:
Thanks for the PR! We will instead implement it once the base `sql-wallet` is merged as we are doing a big upgrade of the wallet to make it sql ready.

PR #1353, opened 2026-08-31
wallet: centralize managed runtime config

yyforyongyu:
## Summary

Centralizes managed Wallet runtime configuration in one validated,
Manager-owned snapshot.

Implements roadmap Task 456.

## Changes

- Moves backend, chain, synchronization, retry, recovery, auto-lock, and
  compact-filter policy into `ManagerConfig`.
- Deep-copies mutable chain parameters at Manager, Wallet, and `Wallet.Info`
  ownership boundaries. Callers must provide canonical, fully populated chain
  parameters; malformed programmer inputs may panic.
- Narrows managed creation and loading to `CreateWalletParams` and
  `LoadWalletParams`, validating wallet identity before cache or backend work.
- Keeps the legacy KVDB public passphrase request-scoped and enforces its
  one-Wallet-per-Manager constraint without introducing a runtime alias.
- Uses Manager-normalized retry and auto-lock defaults as the single runtime
  authority.

## Scope

- Base: upstream `sql-wallet`.
- Uses the existing caller-owned `chain.Interface` and shared `db.Store`.
- Adds no Manager worker, event fan-out, chain-source factory, Store wrapper,
  KVDB alias, durable metadata, CLI, RPC, or daemon behavior.
- Keeps the upstream `task-runtime-config` branch published as the base for the
  next dependent PR.

## Intentional roadmap divergence

Under the approved best-fit implementation exception, this PR does not add the
roadmap's temporary chain-source factory or wallet Store adapter. Managed
Wallets temporarily consume the direct shared chain source; a `TODO(yy)`
records that later coordinator work must provide Manager-owned event fan-out.
The existing Store remains shared because wallet identity is already bound by
its operations.

The KVDB public passphrase remains a narrow create/load request input instead
of Manager configuration, and no KVDB alias is introduced. This keeps legacy
backend context out of Manager runtime policy while KVDB supports only one
Wallet per Manager.

The final identity-focused Create and Load request shape is adopted early;
later roadmap work still owns any context-first API transition.

### DIFF
--- a/bwtest/harness.go
+++ b/bwtest/harness.go
@@ -233,11 +233,18 @@ func (h *HarnessTest) NewWalletManager() *wallet.Manager {
 			"kvdb, sqlite, and postgres", h.dbType)
 	}
 
-	manager, err := wallet.NewManager(h.Context(), wallet.ManagerConfig{
-		Backend:     backend,
-		DataSource:  h.WalletDBSource,
-		ChainParams: h.NetParams(),
-	})
+	// The Manager owns one immutable runtime policy. Harness scenarios share
+	// their existing chain client with every Wallet assembled by that Manager.
+	managerCfg := wallet.ManagerConfig{
+		Backend:                 backend,
+		DataSource:              h.WalletDBSource,
+		ChainParams:             *h.NetParams(),
+		ChainSource:             h.ChainClient,
+		RecoveryWindow:          defaultWalletRecoveryWindow,
+		WalletSyncRetryInterval: defaultWalletSyncRetryInterval,
+	}
+
+	manager, err := wallet.NewManager(h.Context(), managerCfg)
 	require.NoError(h, err, "failed to create wallet manager")
 
 	h.mu.Lock()
--- a/bwtest/harness.go
+++ b/bwtest/harness.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"path/filepath"
 	"runtime/debug"
+	"slices"
 	"sync"
 	"testing"
 
@@ -39,9 +40,9 @@ const (
 )
 
 // walletRegistration tracks every wallet owned by one Manager. Each map value
-// is a copied reload configuration, or nil for wallets registered through
-// RegisterWallet.
-type walletRegistration map[*wallet.Wallet]*wallet.Config
+// is a copied reload request, or nil for wallets registered only for lifecycle
+// ownership.
+type walletRegistration map[*wallet.Wallet]*wallet.LoadWalletParams
 
 // HarnessTest is the integration test harness.
 type HarnessTest struct {
@@ -368,10 +369,10 @@ func (h *HarnessTest) RegisterWallet(manager *wallet.Manager,
 	h.registerWallet(manager, w, nil)
 }
 
-// registerWallet records a wallet under manager and copies reloadConfig when
-// provided.
+// registerWallet records a wallet under manager and owns a copy of optional
+// reloadParams so fixture mutations cannot alter a later reload credential.
 func (h *HarnessTest) registerWallet(manager *wallet.Manager, w *wallet.Wallet,
-	reloadConfig *wallet.Config) {
+	reloadParams *wallet.LoadWalletParams) {
 
 	h.Helper()
 
@@ -390,13 +391,16 @@ func (h *HarnessTest) registerWallet(manager *wallet.Manager, w *wallet.Wallet,
 		)
 	}
 
-	var config *wallet.Config
-	if reloadConfig != nil {
-		copiedConfig := *reloadConfig
-		config = &copiedConfig
+	var params *wallet.LoadWalletParams
+	if reloadParams != nil {
+		copiedParams := *reloadParams
+		copiedParams.PubPassphrase = slices.Clone(
+			reloadParams.PubPassphrase,
+		)
+		params = &copiedParams
 	}
 
-	registration[w] = config
+	registration[w] = params
 }
 
 // DeregisterWallet releases a wallet from harness ownership.
--- a/wallet/common_test.go
+++ b/wallet/common_test.go
@@ -127,7 +127,8 @@ func testSQLiteManager(tb testing.TB) *Manager {
 	m, err := NewManager(context.Background(), ManagerConfig{
 		Backend:     DBBackendSQLite,
 		DataSource:  filepath.Join(tb.TempDir(), "runtime.sqlite"),
-		ChainParams: &chainParams,
+		ChainParams: chainParams,
+		ChainSource: &bwmock.Chain{},
 	})
 	require.NoError(tb, err)
 	tb.Cleanup(func() { _ = m.Close() })
@@ -169,7 +170,8 @@ func testKVDBManagerAt(tb testing.TB, dbPath string) *Manager {
 	m, err := NewManager(context.Background(), ManagerConfig{
 		Backend:     DBBackendKVDB,
 		DataSource:  dbPath,
-		ChainParams: &chainParams,
+		ChainParams: chainParams,
+		ChainSource: &bwmock.Chain{},
 	})
 	require.NoError(tb, err)
 	tb.Cleanup(func() { _ = m.Close() })
--- a/wallet/common_test.go
+++ b/wallet/common_test.go
@@ -148,7 +148,12 @@ func testSQLManager(tb testing.TB, store db.Store) *Manager {
 			store:   store,
 			closeFn: func() error { return nil },
 		},
-		chainParams: &chainParams,
+		config: ManagerConfig{
+			ChainSource:             &bwmock.Chain{},
+			ChainParams:             chainParams,
+			WalletSyncRetryInterval: initialBackoff,
+			AutoLockDuration:        defaultLockDuration,
+		},
 	}
 }
 
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -235,7 +235,7 @@ func (w *Wallet) Start(startCtx context.Context) error {
 // retries and exponential backoff. It ensures the wallet attempts to stay
 // synced even if the backend connection is flaky.
 func (w *Wallet) runSyncLoop() {
-	backoff := initialBackoff
+	backoff := w.initialSyncBackoff()
 
 	for {
 		startTime := time.Now()
@@ -270,6 +270,16 @@ func (w *Wallet) runSyncLoop() {
 	}
 }
 
+// initialSyncBackoff returns the Wallet-local retry policy, preserving the
+// controller default for legacy or test Wallets assembled without Manager.
+func (w *Wallet) initialSyncBackoff() time.Duration {
+	if w.cfg.WalletSyncRetryInterval <= 0 {
+		return initialBackoff
+	}
+
+	return w.cfg.WalletSyncRetryInterval
+}
+
 // waitForBackoff handles the delay between synchronization retry attempts. It
 // resets the backoff if the previous run was stable, waits for the calculated
 // delay, and then returns the updated backoff duration for the next attempt.
@@ -280,7 +290,7 @@ func (w *Wallet) waitForBackoff(startTime time.Time, backoff time.Duration,
 	// If the syncer ran for a significant amount of time, we consider it a
 	// "stable" run and reset the backoff.
 	if time.Since(startTime) > stableRunTime {
-		backoff = initialBackoff
+		backoff = w.initialSyncBackoff()
 	}
 
 	log.Infof("Restarting sync loop in %v...", backoff)
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -486,10 +486,17 @@ func (w *Wallet) Info(ctx context.Context) (*Info, error) {
 		return nil, fmt.Errorf("decode wallet sync tip: %w", err)
 	}
 
+	// Info is an ownership boundary, so return a fresh network snapshot
+	// instead of exposing the Wallet's retained mutable configuration.
+	chainParams, err := cloneChainParams(*w.cfg.ChainParams)
+	if err != nil {
+		return nil, fmt.Errorf("copy chain parameters: %w", err)
+	}
+
 	info := &Info{
 		BirthdayBlock:    w.birthdayBlock,
 		Backend:          w.cfg.Chain.BackEnd(),
-		ChainParams:      w.cfg.ChainParams,
+		ChainParams:      &chainParams,
 		Locked:           !w.state.isUnlocked(),
 		Synced:           w.state.isSynced(),
 		SyncedTo:         syncedTo,
--- a/wallet/controller_test.go
+++ b/wallet/controller_test.go
@@ -1527,41 +1527,57 @@ func TestControllerStart_BirthdayNotSet(t *testing.T) {
 	w.wg.Wait()
 }
 
-// TestControllerUnlock_DefaultTimeout verifies default timeout usage.
-func TestControllerUnlock_DefaultTimeout(t *testing.T) {
+// TestControllerUnlockDefaultTimeout verifies an omitted request timeout is
+// replaced with the Manager-normalized Wallet policy before timer handling.
+func TestControllerUnlockDefaultTimeout(t *testing.T) {
 	t.Parallel()
 
-	// Arrange: Setup a wallet with an auto-lock duration and start the
-	// main loop.
-	w, deps := createTestWalletWithMocks(t)
-
-	w.cfg.AutoLockDuration = time.Minute
+	// Arrange: Put a Wallet in the state accepted by Unlock and configure a
+	// distinctive default. Leave its main loop stopped so the test can inspect
+	// the serialized request at the boundary immediately before handleUnlockReq
+	// uses the request timeout to reset the auto-lock timer.
+	w, _ := createTestWalletWithMocks(t)
+	autoLockDuration := time.Minute
+	w.cfg.AutoLockDuration = autoLockDuration
 
 	require.NoError(t, w.state.toStarting())
 	require.NoError(t, w.state.toStarted())
 
-	w.wg.Add(1)
+	// Act: Invoke Unlock asynchronously because it waits for the main loop's
+	// response, then intercept the exact request the main loop would handle.
+	errChan := make(chan error, 1)
+	go func() {
+		errChan <- w.Unlock(t.Context(), UnlockRequest{
+			Passphrase: []byte("pass"),
+		})
+	}()
 
-	go w.mainLoop()
+	var req unlockReq
+	select {
+	case rawReq := <-w.requestChan:
+		var ok bool
 
-	pass := []byte("pass")
-	req := UnlockRequest{Passphrase: pass}
-	deps.vault.On(
-		"Unlock", mock.Anything, pass,
-	).Return(nil).Once()
-	// Auto-lock might trigger if the test runs slowly, but it's not
-	// guaranteed.
-	deps.vault.On("Lock").Return().Maybe()
+		req, ok = rawReq.(unlockReq)
+		require.True(t, ok)
 
-	// Act: Perform Unlock with default timeout.
-	err := w.Unlock(t.Context(), req)
+	case <-time.After(time.Second):
+		t.Fatal("timeout waiting for unlock request")
+	}
 
-	// Assert: Verify success.
-	require.NoError(t, err)
+	// Assert: The serialized request carries the configured default into the
+	// timer handler. Complete the intercepted request and verify Unlock returns
+	// successfully so its caller-facing response path is not left blocked.
+	require.Equal(t, autoLockDuration, req.req.Timeout)
 
-	// Clean up.
-	w.cancel()
-	w.wg.Wait()
+	req.resp <- nil
+
+	select {
+	case err := <-errChan:
+		require.NoError(t, err)
+
+	case <-time.After(time.Second):
+		t.Fatal("timeout waiting for unlock response")
+	}
 }
 
 // TestControllerStart_DeleteExpiredFail verifies Start fails when
--- a/wallet/controller_test.go
+++ b/wallet/controller_test.go
@@ -8,9 +8,12 @@ import (
 	"testing"
 	"time"
 
+	"github.com/btcsuite/btcd/chaincfg/v2"
 	"github.com/btcsuite/btcd/chainhash/v2"
 	"github.com/btcsuite/btcd/wire/v2"
+	bwmock "github.com/btcsuite/btcwallet/bwtest/mock"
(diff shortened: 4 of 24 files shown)


wallet/manager_config.go
```diff
@@ -118,3 +120,101 @@ func (c ManagerConfig) timeout() time.Duration {
 
 	return c.Timeout
 }
+
+// cloneChainParams returns an ownership-isolated network snapshot. The copy
+// includes every nested mutable value so later caller or Info mutations cannot
+// change the network policy retained by a Manager or Wallet.
+func cloneChainParams(params *chaincfg.Params) (*chaincfg.Params, error) {
+	if params == nil {
```
yyforyongyu: unnecessary nil check, remove it

wallet/manager_config.go
```diff
@@ -118,3 +120,101 @@ func (c ManagerConfig) timeout() time.Duration {
...
+		return nil, nil
+	}
+
+	cloned := *params
+	cloned.DNSSeeds = slices.Clone(params.DNSSeeds)
+
+	if params.GenesisBlock != nil {
```
yyforyongyu: why do we need all these nil check? the wallet should panic if any of these params are missing?

wallet/manager_config.go
```diff
@@ -118,3 +120,101 @@ func (c ManagerConfig) timeout() time.Duration {
...
+
+// cloneDeployment recreates supported time boundaries without retaining their
+// mutable clock state. Unknown implementations fail closed because copying an
+// interface value would silently share caller state.
+func cloneDeployment(deployment chaincfg.ConsensusDeployment) (
+	chaincfg.ConsensusDeployment, error) {
+
+	cloned := deployment
+	switch starter := deployment.DeploymentStarter.(type) {
+	case nil:
```
yyforyongyu: why would this be nil?

wallet/manager_config.go
```diff
@@ -64,9 +71,39 @@ type ManagerConfig struct {
 	MaxConnections int
 
 	// ChainParams identifies the network every wallet in this Manager runs
-	// on. Required; the Manager copies it into each wallet's config.
+	// on. Required; the Manager retains an ownership-isolated copy.
```
yyforyongyu: if we pass `chaincfg.Params` instead of the pointer, do we still need those copy methods? think it can be simplfied?

wallet/manager_config.go
```diff
@@ -138,6 +138,32 @@ func (c ManagerConfig) validate() error {
...
+			c.SyncMethod)
+	}
+
+	if c.WalletSyncRetryInterval < 0 ||
+		c.WalletSyncRetryInterval > maxBackoff {
+
+		return fmt.Errorf("%w: WalletSyncRetryInterval must be between 0 "+
+			"and %v", ErrInvalidParam, maxBackoff)
+	}
+	if c.RecoveryWindow > MaxRecoveryWindow {
```
yyforyongyu, line 179: newline above

wallet/manager_config.go
```diff
@@ -138,6 +138,32 @@ func (c ManagerConfig) validate() error {
...
+		return fmt.Errorf("%w: RecoveryWindow must not exceed %d",
+			ErrInvalidParam, MaxRecoveryWindow)
+	}
+	if c.MaxCFilterItems < 0 {
```
yyforyongyu: newline above

yyforyongyu: also why not making `MaxCFilterItems` a unit so we don't even need this check?

wallet/manager_config_test.go
```diff
@@ -138,6 +170,110 @@ func TestManagerConfigValidate(t *testing.T) {
...
+	sqlKVDBInputs.KVDBWalletAlias = testWalletName
+	sqlKVDBInputs.KVDBPubPassphrase = []byte("public")
+
+	tests := []struct {
+		name    string
+		cfg     ManagerConfig
+		wantIs  error
+		wantMsg string
+	}{
+		{"missing factory", missingFactory, ErrMissingParam,
```
yyforyongyu: malformed, use multi-line with named struct,
```
{
   name: xxx
  ...
}, {...
```

wallet/manager_config.go
```diff
@@ -120,6 +121,11 @@ type ManagerConfig struct {
 // validate checks Manager-wide ownership and backend consistency before any
 // Store is opened. Safe duration defaults are applied only when the immutable
 // runtime snapshot is built after this validation succeeds.
+//
+// The branch count reflects the flat public configuration: each backend and
+// policy field is validated here so no second validation layer can disagree.
+//
+//nolint:cyclop
```
yyforyongyu: why this linter??

wallet/manager_config.go
```diff
@@ -156,10 +163,12 @@ func (c ManagerConfig) validate() error {
 		return fmt.Errorf("%w: WalletSyncRetryInterval must be between 0 "+
 			"and %v", ErrInvalidParam, maxBackoff)
 	}
+
 	if c.RecoveryWindow > MaxRecoveryWindow {
 		return fmt.Errorf("%w: RecoveryWindow must not exceed %d",
 			ErrInvalidParam, MaxRecoveryWindow)
 	}
+
```
yyforyongyu, line 183: this is clearly a violation of commit hygiene? why none of the reviewers or lead caught it?

wallet/manager_config.go
```diff
@@ -211,6 +221,101 @@ func (c ManagerConfig) autoLockDuration() time.Duration {
...
+// isNilChainSource rejects both a nil interface and an interface holding a
+// typed nil, preventing a latent nil receiver from entering Wallet runtime.
+func isNilChainSource(source chain.Interface) bool {
+	if source == nil {
+		return true
+	}
+
+	value := reflect.ValueOf(source)
+	// Only nil-capable kinds need explicit handling; every other concrete
```
yyforyongyu: missing newline above

wallet/manager_config.go
```diff
@@ -211,6 +221,101 @@ func (c ManagerConfig) autoLockDuration() time.Duration {
 	return c.AutoLockDuration
 }
 
+// managerRuntime is the immutable policy snapshot retained by Manager. Its
+// fields remain private so no Wallet or information API can acquire the source
+// factory, credentials, or mutable network data owned at this boundary.
+type managerRuntime struct {
```
yyforyongyu: this is just the manager config plus `chainSourceFactory`? why having this struct?

wallet/manager_test.go
```diff
@@ -317,9 +317,12 @@ func TestCreateWalletParamsPolicy(t *testing.T) {
 
 			backend := &recordingManagerBackend{}
 			m := &Manager{
-				wallets:     make(map[string]*Wallet),
-				backend:     backend,
-				chainParams: &chainParams,
+				wallets: make(map[string]*Wallet),
+				backend: backend,
+				runtime: &managerRuntime{
```
yyforyongyu: i doubt we need the `runtime` field

wallet/common_test.go
```diff
@@ -24,6 +24,12 @@ import (
 // testWalletName is the wallet name shared by Store-backed wallet tests.
 const testWalletName = "test-wallet"
 
+// newTestChainSource supplies a fresh strict chain mock to Manager tests so
+// tests that exercise chain behavior can keep expectations local.
+func newTestChainSource() *bwmock.Chain {
```
yyforyongyu: why define this shim? this is unnecessary?

wallet/manager_config_test.go
```diff
@@ -314,6 +315,7 @@ func TestManagerRetainsRuntimeSnapshot(t *testing.T) {
 	// and passphrase bytes along with distinctive scalar runtime values.
 	params, err := cloneChainParams(chainParams)
 	require.NoError(t, err)
+
```
yyforyongyu, line 217: did you check commit hygiene? where are these newlines coming from?

wallet/manager_test.go
```diff
@@ -10,11 +10,17 @@ import (
...
+	walletmock "github.com/btcsuite/btcwallet/wallet/internal/bwtest/mock"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
 	_ "github.com/btcsuite/btcwallet/walletdb/bdb"
+	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 )
 
+// secondWalletName identifies the non-primary Wallet used to prove that
+// Manager-owned runtime policy remains independent of durable wallet identity.
+const secondWalletName = "second"
```
yyforyongyu: this const is only used in the test, why keep it global?

wallet/controller.go
```diff
@@ -235,7 +235,15 @@ func (w *Wallet) Start(startCtx context.Context) error {
 // retries and exponential backoff. It ensures the wallet attempts to stay
 // synced even if the backend connection is flaky.
 func (w *Wallet) runSyncLoop() {
-	backoff := initialBackoff
+	w.runSyncLoopWithTimer(time.After)
```
yyforyongyu: why do we wanna this change? to make it testable?

wallet/manager.go
```diff
@@ -387,10 +387,8 @@ func validateInitialAccountKeys(accounts []WatchOnlyAccount) error {
 // for tracking. If the named wallet does not exist, Load returns
 // ErrWalletNotFound.
 func (m *Manager) Load(cfg Config) (*Wallet, error) {
-	// The Manager owns the network for every wallet it serves; see Create.
-	cfg.ChainParams = &m.config.ChainParams
-
-	err := cfg.validate()
+	name := m.config.loadWalletName(cfg.Name)
+	err := validateManagedWalletName(name)
```
yyforyongyu: this check is unnecessary now? this name is loaded, so why would it has an error? the name won't be persisted if the validation fails before saving it?

wallet/manager_config.go
```diff
@@ -272,6 +272,17 @@ func (c ManagerConfig) walletConfig(name string) (Config, error) {
 	}, nil
 }
 
+// loadWalletName returns the canonical durable identity before validation.
+// SQL keeps the requested name, while legacy kvdb always reopens its one
+// Wallet under the Manager-configured runtime alias.
+func (c ManagerConfig) loadWalletName(name string) string {
```
yyforyongyu: i think we should name it `patchKVDBName`? and add a TODO to remove it once kvdb is removed?

wallet/manager_kvdb.go
```diff
@@ -72,19 +74,24 @@ func (b *kvdbManagerBackend) create(ctx context.Context, cfg Config,
 	params CreateWalletParams, rootKey *hdkeychain.ExtendedKey) (
 	*walletData, error) {
 
-	if b.store != nil {
-		return nil, fmt.Errorf("%w: kvdb serves one wallet per "+
-			"database; %q is already loaded", ErrInvalidParam,
-			b.walletName)
+	empty, err := b.isEmpty()
```
yyforyongyu: I'd rather move all kvdb-related changes to its own commit so it's easier to be reviewed

bwtest/harness.go
```diff
@@ -243,7 +243,9 @@ func (h *HarnessTest) NewWalletManager() *wallet.Manager {
 		RecoveryWindow:          defaultWalletRecoveryWindow,
 		WalletSyncRetryInterval: defaultWalletSyncRetryInterval,
 	}
-	if backend == wallet.DBBackendKVDB {
+	// This branch intentionally supplies the runtime inputs required by the
```
yyforyongyu: missing newline above, also a commit hygiene violation

wallet/manager_config.go
yyforyongyu: Removed. `cloneChainParams` now accepts `chaincfg.Params` by value and no longer has the unnecessary top-level nil check.

wallet/manager_config.go
yyforyongyu: Simplified. The valid-network `GenesisBlock`, `GenesisHash`, and `PowLimit` values are now copied unconditionally. Only `BIP0034Hash` retains a guard because regression net legitimately leaves it nil.

wallet/manager_config.go
yyforyongyu: Removed. The selected deployment nil branch no longer exists; supported deployment boundaries are cloned explicitly and unsupported implementations return an error.

wallet/manager_config.go
yyforyongyu: `ManagerConfig.ChainParams` now takes a `chaincfg.Params` value. The deep clone remains necessary because that value still contains mutable pointers, slices, checkpoint hashes, and deployment interfaces.

wallet/manager_config.go
yyforyongyu: Added. The validation blocks now have the requested separator, owned by the validation commit.

wallet/manager_config.go
yyforyongyu: Addressed both points: the obsolete validation block is gone, `MaxCFilterItems` is now `uint32`, and the negative-value check is no longer needed.

wallet/manager_config_test.go
yyforyongyu: Reformatted. The validation table now uses multiline named-field entries throughout.

wallet/manager_config.go
yyforyongyu: Removed. The linter suppression is gone; validation is split into focused helpers that pass the repository lint target without suppression.

wallet/manager_config.go
yyforyongyu: You were right; this was missed initially. The stack was rebuilt so the validation spacing belongs to `wallet: validate Manager runtime policy`, and the final commit-hygiene audit reports zero violations.

wallet/manager_config.go
yyforyongyu: Addressed by removal. The reflection-based nil-source helper and this formatting location disappeared when `ChainSourceFactory` was removed.

wallet/manager_config.go
yyforyongyu: Removed. There is no duplicate runtime struct or `ChainSourceFactory`; `Manager` retains the normalized `ManagerConfig` snapshot directly.

wallet/manager_test.go
yyforyongyu: Removed. `Manager` now stores the retained snapshot in its `config ManagerConfig` field; the separate `runtime` field no longer exists.

wallet/common_test.go
yyforyongyu: Removed. Tests now allocate a fresh `&bwmock.Chain{}` directly at each fixture instead of routing through the zero-value shim.

wallet/manager_config_test.go
yyforyongyu: Fixed. The blank-only changes were folded into the commits that introduce their surrounding tests; blame now attributes this Arrange spacing to the runtime-policy assembly commit.

wallet/manager_test.go
yyforyongyu: Localized. Each affected test now declares its own `secondWalletName` Arrange value; the test-file-level constant is gone.

wallet/controller.go
yyforyongyu: Yes. The private seam lets the retry-policy test observe the configured interval without sleeping. Production still calls it with `time.After`, so runtime timing behavior and the public API are unchanged.

wallet/manager.go
yyforyongyu: Kept intentionally for SQL: `requestedName` is caller input, not a value already loaded from persistence. The boundary check rejects an empty request before cache or Store access and preserves `ErrMissingParam`; KVDB alias substitution does not make the SQL path redundant.

wallet/manager_config.go
yyforyongyu: Renamed to `patchKVDBName` and documented with a TODO to remove the compatibility patch when the deprecated KVDB backend is removed.

wallet/manager_kvdb.go
yyforyongyu: Done. KVDB alias, credential, namespace, and reopen behavior now live in the dedicated `wallet: centralize kvdb runtime inputs` commit; generic managed Load assembly remains separate.

bwtest/harness.go
yyforyongyu: Fixed. The compatibility branch has the requested separator, and its explanation plus lint annotation now originate in the commit that owns the KVDB compatibility behavior.

wallet/controller_test.go
```diff
@@ -1527,41 +1527,56 @@ func TestControllerStart_BirthdayNotSet(t *testing.T) {
...
-	// main loop.
-	w, deps := createTestWalletWithMocks(t)
-
-	w.cfg.AutoLockDuration = time.Minute
+	// Arrange: Put a Wallet in the state accepted by Unlock and configure a
+	// distinctive default. Leave its main loop stopped so the test can inspect
+	// the serialized request at the boundary immediately before handleUnlockReq
+	// uses the request timeout to reset the auto-lock timer.
+	w, _ := createTestWalletWithMocks(t)
+	w.cfg.AutoLockDuration = 37 * time.Minute
```
yyforyongyu: how did you come up with this weird number?

wallet/manager_config.go
```diff
@@ -227,6 +227,63 @@ func (c ManagerConfig) validateBackendRuntime() error {
 	return nil
 }
 
+// snapshot copies mutable inputs and normalizes defaults before a Store is
+// opened, so Manager retains one self-contained configuration value.
+func (c ManagerConfig) snapshot() (ManagerConfig, error) {
```
yyforyongyu: this seems unnecessary?

wallet/manager_config.go
```diff
@@ -284,6 +284,18 @@ func (c ManagerConfig) walletConfig(name string) (Config, error) {
 	}, nil
 }
 
+// patchKVDBName substitutes the Manager-owned alias for legacy kvdb while SQL
+// continues to use the caller's durable wallet identity.
+//
+// TODO: Remove this compatibility patch with the deprecated kvdb backend.
```
yyforyongyu: add `TODO(yy)`

wallet/manager.go
```diff
@@ -395,6 +395,10 @@ func (m *Manager) Load(cfg Config) (*Wallet, error) {
 		return nil, err
 	}
 
+	// Apply the legacy single-Wallet alias only after the caller's identity
+	// passes the common validation boundary.
+	cfg.Name = m.config.patchKVDBName(cfg.Name)
```
yyforyongyu: One thought - we can remove this, and always require a name to be passed? so check it in `validate` - for kvdb, this name won't have any effect, but for the caller they just need to remember to pass the name?

wallet/manager_config.go
```diff
@@ -64,8 +74,39 @@ type ManagerConfig struct {
...
+	// values use the safe default rather than disabling automatic locking.
+	AutoLockDuration time.Duration
+
+	// MaxCFilterItems sets the automatic compact-filter fallback threshold.
+	// Zero uses the syncer's default threshold.
+	MaxCFilterItems uint32
+
+	// KVDBWalletAlias is the runtime identity of the sole legacy kvdb Wallet.
+	// It is required for kvdb and invalid for SQL backends.
+	KVDBWalletAlias string
```
yyforyongyu: I don't think we need this - the kvdb wallet doesn't have a name.

wallet/manager_config.go
```diff
@@ -64,8 +74,39 @@ type ManagerConfig struct {
...
+
+	// KVDBPubPassphrase opens the sole legacy kvdb Wallet. Manager copies the
+	// bytes at construction; the field is required only for kvdb.
+	KVDBPubPassphrase []byte
```
yyforyongyu: this can instead be put on `Config` instead of `ManagerConfig`? why having it here?

wallet/manager_config.go
```diff
@@ -110,6 +166,158 @@ func (c ManagerConfig) validate() error {
...
+
+	c.KVDBPubPassphrase = slices.Clone(c.KVDBPubPassphrase)
+
+	return c, nil
+}
+
+// walletConfig constructs one Wallet-local policy from the Manager snapshot.
+// It invokes the retained source boundary and copies mutable values again so
+// sibling Wallets cannot consume or alter each other's runtime configuration.
+func (c ManagerConfig) walletConfig(name string) (Config, error) {
```
yyforyongyu, line 190: there is no need for it to take the `name` param if we remove the kvdb alias

wallet/controller_test.go
yyforyongyu: Replaced the arbitrary duration with the existing one-minute test policy so the value now communicates intent.

wallet/manager_config.go
yyforyongyu: Removed the unnecessary helper/assembly indirection. Runtime normalization now happens directly when NewManager captures its snapshot.

wallet/manager_config.go
yyforyongyu: Added TODO(yy) at the direct ChainSource field documenting the later Manager-owned fan-out work.

wallet/manager.go
yyforyongyu: Done. Create and Load now always require the caller-supplied name and validate it before cache or backend work, including KVDB.

wallet/manager_config.go
yyforyongyu: Removed the KVDB alias entirely. KVDB remains limited to one Wallet per Manager and does not persist or normalize an alias.

wallet/manager_config.go
yyforyongyu: Removed the passphrase from ManagerConfig. It now exists only in the narrow create/load request and is consumed by the KVDB backend without entering Manager or Wallet runtime policy.

wallet/manager_config.go
yyforyongyu: The KVDB alias is removed. walletConfig still needs the name to assemble the individual Wallet runtime identity used by Config.Name; it no longer performs any KVDB-specific normalization.

wallet/manager_config.go
```diff
@@ -64,8 +65,40 @@ type ManagerConfig struct {
 	MaxConnections int
 
 	// ChainParams identifies the network every wallet in this Manager runs
-	// on. Required; the Manager copies it into each wallet's config.
-	ChainParams *chaincfg.Params
+	// on. Required; the Manager retains an ownership-isolated value. Callers
```
yyforyongyu: we should use a newline, and use `NOTE:`, sth like
```
// on.
//
// NOTE: the Manager retains an ownership-isolated value. Callers
```

yyforyongyu: Addressed in d81c99dc06. The ownership caveat now starts a separate `NOTE:` paragraph.

GustavoStingelin (approved):
easy to follow, lgtm

PR #1354, opened 2026-08-31
wallet: route account requests during stop

yyforyongyu:
## Summary

Implements roadmap Task 458

Admit public AccountManager requests through the Wallet's existing main loop
and make terminal shutdown observable through a stable error.

## Change Description

- Distinguish initialized and terminally stopped Wallet state and return the
  standalone `ErrWalletStopped` sentinel after shutdown begins.
- Route public AccountManager methods through the existing request channel,
  execute accepted handlers concurrently, and drain them through the existing
  Wallet WaitGroup.
- Preserve the private pre-start account import path used during Wallet
  construction.
- Add lifecycle, concurrency, cancellation, error-propagation, and stopped-call
  coverage in existing test files.

## Notes

PR #1353, the former prerequisite, has merged. This draft is based directly on
`sql-wallet`.

Stacking after Task 456 was a review-order choice even though the roadmap lists
Tasks 456 and 458 as sibling dependencies of Task 459.

The intentional divergence from the current roadmap task plan is that this PR
limits request-loop migration to AccountManager. Other maintained Wallet
components and full private lifecycle teardown are not included.

### DIFF
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -424,6 +424,23 @@ func (w *Wallet) ListAccountsByName(ctx context.Context,
 	})
 }
 
+// accountResp carries one account snapshot or the lookup error so an accepted
+// handler can always finish even when its caller stops waiting.
+type accountResp struct {
+	info *AccountInfo
+	err  error
+}
+
+// getAccountReq owns the immutable lookup inputs and buffered response path
+// that cross the Wallet's request boundary.
+type getAccountReq struct {
+	reqCtx
+
+	scope waddrmgr.KeyScope
+	name  string
+	resp  chan accountResp
+}
+
 // GetAccount returns the account for a given account name and key scope.
 // The account snapshot, including the running balance, is fetched in a
 // single Store read.
@@ -435,24 +452,54 @@ func (w *Wallet) GetAccount(ctx context.Context, scope waddrmgr.KeyScope,
 		return nil, err
 	}
 
-	info, err := w.cache.GetAccount(ctx, db.GetAccountQuery{
-		WalletID: w.id,
-		Scope:    db.KeyScope(scope),
-		Name:     &name,
-	})
-	if err != nil {
-		// Preserve waddrmgr.ManagerError semantics so callers
-		// using waddrmgr.IsError(err, ...) keep working when kvdb
-		// wraps the underlying manager error via fmt.Errorf.
-		var mErr waddrmgr.ManagerError
-		if errors.As(err, &mErr) {
-			return nil, mErr
-		}
+	req := getAccountReq{
+		reqCtx: reqCtx{ctx: ctx},
+		scope:  scope,
+		name:   name,
+		resp:   make(chan accountResp, 1),
+	}
 
+	err = w.sendReq(ctx, req)
+	if err != nil {
 		return nil, err
 	}
 
-	return w.accountInfoFromStore(info)
+	resp, err := waitForReq(ctx, req.resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.info, resp.err
+}
+
+// handleGetAccount executes an accepted lookup with its caller context and
+// sends exactly one buffered response so shutdown never depends on a receiver.
+func (w *Wallet) handleGetAccount(req getAccountReq) {
+	info, err := w.cache.GetAccount(req.ctx, db.GetAccountQuery{
+		WalletID: w.id,
+		Scope:    db.KeyScope(req.scope),
+		Name:     &req.name,
+	})
+	if err != nil {
+		// Preserve waddrmgr.ManagerError semantics so callers using
+		// waddrmgr.IsError(err, ...) keep working when kvdb wraps the
+		// underlying manager error via fmt.Errorf.
+		var mErr waddrmgr.ManagerError
+		if errors.As(err, &mErr) {
+			req.resp <- accountResp{err: mErr}
+			return
+		}
+
+		req.resp <- accountResp{err: err}
+
+		return
+	}
+
+	account, err := w.accountInfoFromStore(info)
+	req.resp <- accountResp{
+		info: account,
+		err:  err,
+	}
 }
 
 // RenameAccount renames an existing account. The new name must be unique within
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -692,6 +692,19 @@ func (w *Wallet) handleRenameAccount(req renameAccountReq) {
 	req.resp <- err
 }
 
+// importAccountReq carries every import option through Wallet admission while
+// retaining the existing private implementation for pre-start Manager use.
+type importAccountReq struct {
+	reqCtx
+
+	name                 string
+	accountKey           *hdkeychain.ExtendedKey
+	masterKeyFingerprint uint32
+	addrType             waddrmgr.AddressType
+	dryRun               bool
+	resp                 chan accountResp
+}
+
 // ImportAccount imports an account from an extended public key. Private
 // extended keys are rejected. The key scope is derived from the version
 // bytes of the extended key. The account name must be unique within the
@@ -717,9 +730,47 @@ func (w *Wallet) ImportAccount(ctx context.Context,
 		return nil, err
 	}
 
-	return w.importAccountInternal(
-		ctx, name, accountKey, masterKeyFingerprint, addrType, dryRun,
+	accountKeySnapshot, err := snapshotExtendedPubKey(
+		accountKey, true, w.cfg.ChainParams,
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	req := importAccountReq{
+		reqCtx:               reqCtx{ctx: ctx},
+		name:                 name,
+		accountKey:           accountKeySnapshot,
+		masterKeyFingerprint: masterKeyFingerprint,
+		addrType:             addrType,
+		dryRun:               dryRun,
+		resp:                 make(chan accountResp, 1),
+	}
+
+	err = w.sendReq(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := waitForReq(ctx, req.resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.info, resp.err
+}
+
+// handleImportAccount invokes the existing non-admitting implementation for
+// a request that mainLoop has already accepted and handleReq will complete.
+func (w *Wallet) handleImportAccount(req importAccountReq) {
+	info, err := w.importAccountInternal(
+		req.ctx, req.name, req.accountKey, req.masterKeyFingerprint,
+		req.addrType, req.dryRun,
+	)
+	req.resp <- accountResp{
+		info: info,
+		err:  err,
+	}
 }
 
 // importAccountInternal is the internal implementation of ImportAccount,
@@ -801,6 +852,28 @@ func dbScopeAddrSchema(
 	return &converted, nil
 }
 
+// snapshotExtendedPubKey validates caller-owned key material before copying
+// it into a request. Validation prevents private key serialization, while the
+// reparse ensures an admitted handler never retains a mutable caller pointer
+// after cancellation returns.
+func snapshotExtendedPubKey(pubKey *hdkeychain.ExtendedKey,
+	isAccountKey bool, chainParams *chaincfg.Params) (
+	*hdkeychain.ExtendedKey, error) {
+
+	err := validateExtendedPubKey(pubKey, isAccountKey, chainParams)
+	if err != nil {
+		return nil, err
+	}
+
+	snapshot, err := hdkeychain.NewKeyFromString(pubKey.String())
+	if err != nil {
+		return nil, fmt.Errorf("%w: copy extended public key: %s",
+			ErrInvalidAccountKey, err.Error())
+	}
+
+	return snapshot, nil
+}
+
 // validateExtendedPubKey ensures a sane derived public key is provided.
 func validateExtendedPubKey(pubKey *hdkeychain.ExtendedKey,
 	isAccountKey bool, chainParams *chaincfg.Params) error {
@@ -815,6 +888,13 @@ func validateExtendedPubKey(pubKey *hdkeychain.ExtendedKey,
 	// Private keys are not allowed.
 	if pubKey.IsPrivate() {
 		return fmt.Errorf("%w: private keys cannot be imported",
+			ErrInvalidAccountKey)
+	}
+
+	// A zeroed or otherwise malformed key has no four-byte network version.
+	// Reject it before isPubKeyForNet decodes the version as a uint32.
+	if len(pubKey.Version()) != binary.Size(waddrmgr.HDVersion(0)) {
+		return fmt.Errorf("%w: invalid extended public key version",
 			ErrInvalidAccountKey)
 	}
 
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -353,6 +353,22 @@ func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
 	}
 }
 
+// listAccountsResp lets a list handler finish even if caller cancellation
+// causes the public method to stop receiving its result.
+type listAccountsResp struct {
+	infos []AccountInfo
+	err   error
+}
+
+// listAccountsReq carries one fully formed Store query so every public list
+// variant shares the same admitted handler without losing its filter.
+type listAccountsReq struct {
+	reqCtx
+
+	query db.ListAccountsQuery
+	resp  chan listAccountsResp
+}
+
 // ListAccounts returns every account across all key scopes with its balance.
 func (w *Wallet) ListAccounts(ctx context.Context) ([]AccountInfo, error) {
 	err := w.state.validateStarted()
@@ -360,9 +376,25 @@ func (w *Wallet) ListAccounts(ctx context.Context) ([]AccountInfo, error) {
 		return nil, err
 	}
 
-	return w.listAccountInfos(ctx, db.ListAccountsQuery{
-		WalletID: w.id,
-	})
+	req := listAccountsReq{
+		reqCtx: reqCtx{ctx: ctx},
+		query: db.ListAccountsQuery{
+			WalletID: w.id,
+		},
+		resp: make(chan listAccountsResp, 1),
+	}
+
+	err = w.sendReq(ctx, req)
+	if err != nil {
+		return nil, err
+	}
+
+	resp, err := waitForReq(ctx, req.resp)
+	if err != nil {
+		return nil, err
+	}
+
+	return resp.infos, resp.err
 }
 
 // listAccountInfos converts cache.ListAccounts snapshots into wallet-owned
@@ -392,6 +424,16 @@ func (w *Wallet) listAccountInfos(ctx context.Context,
 	return results, nil
 }
 
+// handleListAccounts performs the Store read for any admitted list variant;
+// handleReq owns the matching shutdown accounting.
+func (w *Wallet) handleListAccounts(req listAccountsReq) {
+	infos, err := w.listAccountInfos(req.ctx, req.query)
+	req.resp <- listAccountsResp{
+		infos: infos,
+		err:   err,
+	}
+}
+
 // ListAccountsByScope returns all accounts for the given key scope.
 func (w *Wallet) ListAccountsByScope(ctx context.Context,
 	scope waddrmgr.KeyScope) ([]AccountInfo, error) {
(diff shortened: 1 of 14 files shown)


wallet/controller.go
```diff
@@ -346,13 +346,10 @@ func (w *Wallet) performRuntimeSetup(startCtx context.Context) error {
...
-	err := w.state.toStopping()
-	if err != nil {
-		// If the wallet is not started, we can consider it stopped.
-		log.Warnf("Wallet already stopped: %v", err)
-		return nil
-	}
+	// Close the Wallet-owned boundary before resources are canceled, then
+	// preserve those resources until calls accepted before the cutoff finish.
+	drained := w.state.closeAndDrain()
+	<-drained
```
yyforyongyu: we should never do this without catching a timeout - what if `<-drained` blocks forever?

wallet/state.go
```diff
@@ -69,6 +96,23 @@ func (l lifecycle) String() string {
...
+	admissionOnce sync.Once
+
+	// admissionRequests carries acquire and release bookkeeping only. Public
+	// method bodies remain in their caller goroutines and run concurrently.
+	admissionRequests chan admissionRequest
+
+	// stopAdmission competes with acquire requests for the terminal cutoff.
+	// terminal publishes that cutoff to late callers and worker handoffs, while
+	// drained closes after the loop has consumed every accepted release.
+	stopAdmission chan struct{}
```
yyforyongyu: we prefer name go channels using suffix `xxxChan`

wallet/state.go
```diff
@@ -69,6 +96,23 @@ func (l lifecycle) String() string {
 //     unlocked. This dictates the ability to perform sensitive operations like
 //     signing.
 type walletState struct {
+	// admissionOnce initializes and starts the admission owner only when a
+	// call, worker, or stop first needs it. This preserves walletState's
+	// zero-value compatibility without attaching an idle goroutine to
+	// status-only states.
+	admissionOnce sync.Once
```
yyforyongyu: find a better name than `admission`

wallet/state.go
```diff
@@ -203,6 +247,128 @@ func (s *walletState) toStopped() error {
...
+// active count is local so no caller can observe or mutate it concurrently.
+func (s *walletState) runAdmissionLoop() {
+	var active uint64
+
+	for {
+		select {
+		case request := <-s.admissionRequests:
+			switch request.action {
+			case admissionAcquire:
+				active++
```
yyforyongyu: why do we need to track `active++`? we only care the last call when the stop signal is received? also name it `quitChan`?

wallet/state.go
```diff
@@ -203,6 +247,128 @@ func (s *walletState) toStopped() error {
...
+	}
+
+	select {
+	case <-s.terminal:
+		return nil, ErrWalletStopped
+
+	case s.admissionRequests <- request:
+	}
+
+	err := <-decision
```
yyforyongyu: we should never read a go channel without a timeout gate - what if the channel is blocking forever??

wallet/internal/keyvault/vault_lifecycle_test.go
```diff
@@ -26,6 +26,15 @@ var (
 	wrongPassphrase = []byte("wrong-passphrase")
 )
 
+// cleanupWalletVault registers teardown that fails the owning test if runtime
+// secrets cannot be cleared, rather than discarding the cleanup result.
+func cleanupWalletVault(t *testing.T, vault *WalletVault) {
```
yyforyongyu: why making this unnecessary change? this is just a shim that improves nothing?

wallet/controller.go
```diff
@@ -173,64 +176,50 @@ type Controller interface {
...
 	return nil
 }
 
+// Start delegates the temporary public lifecycle API to private one-shot
+// startup. Callers must serialize Start with Stop until Manager-owned lifecycle
+// removes this compatibility surface.
+//
+// This is part of the Controller interface.
+func (w *Wallet) Start(startCtx context.Context) error {
+	return w.start(startCtx)
```
yyforyongyu: i don't get it - why privatizing the old Start then reintroduce it?

wallet/controller.go
```diff
@@ -196,12 +196,12 @@ func (w *Wallet) start(startCtx context.Context) error {
...
 	// cannot begin its terminal Wait first.
-	w.wg.Add(2)
-	go w.mainLoop()
-	go func() {
-		defer w.wg.Done()
-		w.runSyncLoop()
-	}()
+	w.mainDone = make(chan struct{})
+	w.syncDone = make(chan struct{})
+	w.wg.Add(walletWorkerCount)
```
yyforyongyu: why making this change? extremely unnecessary??

wallet/controller.go
yyforyongyu: Kept this wait unbounded intentionally. Once shutdown closes the call boundary, teardown cannot safely close Wallet dependencies until every previously accepted call releases its slot. A timeout here would reintroduce the race this change prevents. Stop(ctx) still bounds the public caller wait while the one-shot shutdown continues in the background. The updated comment now states that contract.

wallet/state.go
yyforyongyu: Addressed. All channels introduced by this change now use the Chan suffix, including callReqChan, quitChan, terminalChan, drainedChan, and the worker completion channels.

wallet/state.go
yyforyongyu: Addressed. The implementation now uses call-tracking terminology throughout: callEvent, callRequest, callLoopOnce, startCallLoop, runCallLoop, and beginCall.

wallet/state.go
yyforyongyu: The active count is required because accepted calls run concurrently and can finish in any order; there is no single last request whose release proves all earlier calls finished. Shutdown closes admission and waits for the count to reach zero. The stop signal has also been renamed quitChan, and the invariant is documented next to the loop.

wallet/state.go
yyforyongyu: Kept this receive unbounded intentionally. After a request is published, the state loop owns the protocol and must return one definitive decision. Timing out locally could abandon an accepted call without its matching release and make shutdown drain forever. Cancellation is honored while publishing the request, before ownership transfers; the updated comment documents this boundary.

wallet/internal/keyvault/vault_lifecycle_test.go
yyforyongyu: Kept the shared cleanup assertion because Vault.Lock now returns an error and each cleanup must verify it. Removing the helper would duplicate the same required assertion across the existing tests without reducing production behavior.

wallet/controller.go
yyforyongyu: The private start and stop methods provide the one-shot lifecycle contract used by Manager-owned startup and shutdown. The exported methods remain thin delegates only for existing callers during the current API transition; they do not introduce a second lifecycle path.

wallet/controller.go
yyforyongyu: Addressed the unnecessary part by removing the one-use worker-count constant and the one-use sync-loop and worker-result wrappers. The two named workers remain because shutdown must join both and retain their results before closing dependencies.

wallet/state.go
```diff
@@ -113,15 +126,20 @@ func (s *walletState) String() string {
...
-// toStarting transitions the wallet state from Stopped to Starting.
+// toStarting transitions the wallet state from Initialized to Starting.
 // It initializes the synchronization and authentication states to their
-// secure defaults. It returns an error if the wallet is already started or
-// not in the Stopped state.
+// secure defaults. A Wallet that reached Stopped is terminal and cannot start
+// again.
 func (s *walletState) toStarting() error {
-	// 1. Lifecycle (System State): Atomic transition from Stopped to
+	current := lifecycle(s.lifecycle.Load())
```
yyforyongyu: this `current` is defined and used only once so why having it?

wallet/state_test.go
```diff
@@ -350,9 +391,33 @@ func TestStateValidateStarted(t *testing.T) {
...
+		// Wallet that has never run from one that has stopped.
+		state := newWalletState(nil)
+
+		// Act: Validate an operation before the first Start.
+		err := state.validateStarted()
+
+		// Assert: Pre-start access retains the original broad state
+		// error and does not claim that the Wallet has stopped.
+		require.ErrorIs(t, err, ErrStateForbidden)
+		require.NotErrorIs(t, err, ErrWalletStopped)
```
yyforyongyu: given that the error is not wrapped, there is no need for this `NotErrorIs` check?

wallet/account_manager.go
```diff
@@ -424,9 +424,35 @@ func (w *Wallet) ListAccountsByName(ctx context.Context,
...
+type accountResp struct {
+	info *AccountInfo
+	err  error
+}
+
+// accountReqContext carries caller cancellation and values through the
+// Wallet's transient account requests. Requests exist only until their
+// handlers finish, so retaining this context is bounded by the Wallet's
+// shutdown drain.
+type accountReqContext struct {
```
yyforyongyu: why having this struct? the `ctx` can just be part of `getAccountReq`?

wallet/benchmark_helpers_test.go
```diff
@@ -326,12 +327,31 @@ func setupBenchmarkWallet(tb testing.TB,
 		}
 	}
 
+	// Legacy benchmark setup lacks the maintained lifetime. Initialize it so
```
yyforyongyu, line 330: this change belongs to previous commit?

wallet/controller.go
```diff
@@ -630,6 +644,18 @@ func (w *Wallet) mainLoop() {
 	}
 }
 
+// handleReq dispatches concurrent work and completes its WaitGroup tracking.
+func (w *Wallet) handleReq(req any) {
+	defer w.wg.Done()
+
+	switch r := req.(type) {
+	case getAccountReq:
```
yyforyongyu, line 691: this belongs to next commit? for this commit, we only need to introduce a dumb `handleReq` which does nothing, only do sth like,
```go
func (w *Wallet) handleReq(req any) {
	defer w.wg.Done()

	switch r := req.(type) {
        default:
		log.Errorf("Wallet received unknown request type: %T", req)
	}
}
```

wallet/account_manager.go
```diff
@@ -353,16 +353,55 @@ func propertiesToAccountInfo(props *waddrmgr.AccountProperties,
...
+// causes the public method to stop receiving its result.
+type listAccountsResp struct {
+	infos []AccountInfo
+	err   error
+}
+
+// listAccountsReq carries one fully formed Store query so every public list
+// variant shares the same admitted handler without losing its filter.
+type listAccountsReq struct {
+	accountReqContext
```
yyforyongyu: ok maybe one pro of having this struct is such that we don't need to do `nolint` everywhere - then this should be called `reqCtx` instead of `accountReqContext`, as it will be used by all public API requests.

wallet/common_test.go
```diff
@@ -157,13 +157,27 @@ func testSQLManager(tb testing.TB, store db.Store) *Manager {
...
 	t.Helper()
 
 	require.NoError(t, w.state.toStarting())
 	require.NoError(t, w.state.toStarted())
+
+	// The request loop is part of the started Wallet contract, so keep it
+	// alive until the test finishes even though chain sync remains disabled.
+	w.wg.Add(1)
+
+	go w.mainLoop()
```
yyforyongyu, line 173: does the change of the `wallet/common_test.go` belong to this commit?

wallet/account_manager.go
```diff
@@ -744,6 +744,19 @@ func (w *Wallet) handleRenameAccount(req renameAccountReq) {
 	req.resp <- err
 }
 
+// importAccountReq carries every import option through Wallet admission while
+// retaining the existing private implementation for pre-start Manager use.
+type importAccountReq struct {
+	accountReqContext
+
+	name                 string
```
yyforyongyu, line 700: should we define the `Params` for these public APIs? like `ImportAccountParams` such that we don't need to re-define all these fields again? same for other public methods

wallet/state_test.go
yyforyongyu: Fixed. I removed the redundant negative sentinel assertions from the initialized and stopped validateStarted cases; the positive ErrorIs assertions remain.

wallet/account_manager.go
yyforyongyu: I kept one shared context carrier per the follow-up discussion so the containedctx exemption stays centralized. It is now named reqCtx and lives beside handleReq.

wallet/benchmark_helpers_test.go
yyforyongyu: I kept this in the account lookup commit. That is the first commit where GetAccount sends through the unbuffered request channel, so the benchmark fixture first needs mainLoop there; the preceding framework commit routes no public method.

wallet/controller.go
yyforyongyu: Fixed. The framework commit now adds only concurrent fallback wiring and an unknown-request-only handleReq. reqCtx, getAccountReq, handleGetAccount, and the GetAccount dispatch case now begin in the following lookup commit.

wallet/account_manager.go
yyforyongyu: Fixed. The shared carrier is now reqCtx, is defined beside handleReq in controller.go, and is embedded by each routed account request.

wallet/common_test.go
yyforyongyu: Yes, this belongs to the mutation commit and remains there. startLoadedWalletForTest first needs mainLoop when NewAccount becomes request-routed; moving it earlier would precede its first functional requirement.

wallet/account_manager.go
yyforyongyu: I kept the existing public signatures. Introducing Params types would change the AccountManager interface and every caller, which is outside this request-admission change; the private request structs remain the transport boundary for now.

wallet/account_manager_test.go
```diff
@@ -644,6 +645,222 @@ func TestGetAccountIncludesImportedPseudoAccount(t *testing.T) {
...
+	_, initializedErr := w.GetAccount(t.Context(), scope, testAccountName)
+	require.NoError(t, w.Stop(t.Context()))
+	_, stoppedErr := w.GetAccount(t.Context(), scope, testAccountName)
+
+	// Assert: Initialized retains the broad state error, terminal access
+	// gains the specific sentinel, and neither attempt reaches Store.
+	require.ErrorIs(t, initializedErr, ErrStateForbidden)
+	require.NotErrorIs(t, initializedErr, ErrWalletStopped)
+	require.ErrorIs(t, stoppedErr, ErrWalletStopped)
+	deps.store.AssertExpectations(t)
```
yyforyongyu: this is unnecessary? the setup method already performs this check in `t.Cleanup`

wallet/account_manager_test.go
```diff
@@ -644,6 +645,222 @@ func TestGetAccountIncludesImportedPseudoAccount(t *testing.T) {
...
+	close(release)
+
+	for range 2 {
+		result := <-results
+		require.NoError(t, result.err)
+		require.NotNil(t, result.info)
+	}
+
+	require.NoError(t, <-stopResult)
+	deps.store.AssertExpectations(t)
```
yyforyongyu: same here, unnecessary, remove all these checks

wallet/account_manager_test.go
yyforyongyu: Removed this duplicate assertion. `createTestWalletWithMocks` already registers `AssertExpectations(t)` through `t.Cleanup`, so the expectation is now verified once by the shared fixture.

wallet/account_manager_test.go
yyforyongyu: Removed all five task-added duplicate Store assertions, including this one. The existing fixture cleanup remains the single expectation check.

wallet/controller.go
```diff
@@ -882,31 +926,34 @@ func (w *Wallet) handleChangePassphraseReq(req changePassphraseReq) {
...
 }
 
-// waitForResp waits for the response from an operation request or handles
-// cancellation.
-func (w *Wallet) waitForResp(ctx context.Context, resp <-chan error) error {
+// waitForReq waits for a typed result after sendReq has transferred ownership
+// to mainLoop. It intentionally observes only caller cancellation: Wallet
+// shutdown must continue draining the accepted handler, whose response channel
+// is buffered so completion never depends on the caller remaining present.
+func waitForReq[T any](ctx context.Context, resp <-chan T) (T, error) {
```
yyforyongyu: rename `resp` to `respChan` for clarity, also, the changes made on `waitForReq` should be in its own commit? then another commit to introduce `handleReq`?

yyforyongyu: Addressed. `waitForReq` now names its channel parameter `respChan`. I also split the former mixed commit into `wallet: preserve accepted request results`, which owns the typed waiter and existing control-request result semantics, followed by `wallet: wire request dispatch`, which owns only the concurrent `mainLoop` fallback and `handleReq`.

wallet/controller.go
```diff
@@ -616,8 +646,11 @@ func (w *Wallet) mainLoop() {
 				w.handleChangePassphraseReq(r)
 
 			default:
-				log.Errorf("Wallet received unknown request "+
-					"type: %T", req)
+				// Non-control requests run concurrently. Register the handler
+				// before launch so Stop drains every accepted request.
```
GustavoStingelin, line 649: Could we make terminal shutdown continue independently of the `Stop` caller and have repeated calls join the same completion? The newly tracked account handlers expose a gap in the existing `Stop` implementation: canceling its caller abandons the vault lock and terminal-state transition, and another `Stop` returns `nil` while the accepted handler is still using Store.

I reproduced this on `22a9a743208bb8a302ab19ff79fd957b55275949` using the existing Wallet test fixtures under `go test -race`:

1. Create a started Wallet with `createStartedWalletWithMocks(t)`. Block the mock `Store.GetAccount` on a release channel and call the public `w.GetAccount` in a goroutine.
2. Wait until the Store mock is entered, so the request is definitely accepted.
3. Call `w.Stop(ctx)` with an already-canceled context. It begins shutdown and returns `context.Canceled`.
4. Before releasing the Store call, call `w.Stop(t.Context())` again. It returns `nil` immediately, although the accepted lookup is still blocked.
5. Release the Store call and wait for the lookup and `w.wg.Wait()`. The lookup finishes normally, but the lifecycle remains `lifecycleStopping` and `Vault.Lock` was never called. Another `Stop` still returns `nil`.

The detached goroutine only waits and closes its local `done` channel. Cleanup remains after the caller's cancellation return, and subsequent calls exit at the failed `toStopping` transition. A caller relying on successful Stop can therefore tear down dependencies before the account work has drained.

I realize those Stop branches predate this PR and the description excludes full private teardown. Still, this looks required by [Task 458](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/458-manager-operation-admission.md): it requires accepted work to finish before stop completes, repeated stop to return one terminal result, and cleanup to finish. [Task 331](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/331-manager-wallet-lifecycle.md) explicitly consumes Task 458's Wallet lifecycle mechanics. The [earlier resolved reply](https://github.com/btcsuite/btcwallet/pull/1354#discussion_r3906649088) also says shutdown continues in the background when the caller stops waiting, which the current head does not implement.

A Wallet-owned shutdown completion path that drains, locks the vault, records the terminal result, and lets each Stop caller independently cancel its wait would address both consequences.

GustavoStingelin (changes requested):

wallet/controller.go
GustavoStingelin: Will it be fixed in a follow-up?

GustavoStingelin: looks like #1373 does not solve it

yyforyongyu: very nice catch! Yeah indeed - the plan was to remove the public Start/Stop on the wallet and let it be managed by the wallet manager only, but you are right that we still have this issue if we having multiple competing sources of "quit" - will address in a new commit, i think the cleanest fix is to remove the `ctx` from `w.stop` as we don't wanna abandon an already started shutdown process internally

yyforyongyu: Addressed in 5380cabb4a43513ad5a6e8856b475d1b308fafc5 with the serialized blocking lifecycle contract selected for Wallet. Complete teardown now lives in private context-free `w.stop()`: after the terminal transition begins it cancels runtime work, blocks on the existing WaitGroup, locks the Vault, and records `lifecycleStopped` before returning. The temporary public `Stop(context.Context)` adapter retains its parameter only for compatibility and cannot abandon that teardown on cancellation. Manager owns lifecycle serialization, so concurrent Stop callers are outside this private one-shot contract; a sequential repeated Stop remains idempotent. `TestControllerStop` now passes an already-canceled context while holding a worker in drain, proves Stop cannot return early, then verifies Vault cleanup, terminal state, and sequential repetition under `go test -race`. This intentionally avoids another Wallet channel, mutex, or loop.

wallet/controller.go
```diff
@@ -357,14 +357,21 @@ func (w *Wallet) performRuntimeSetup(startCtx context.Context) error {
...
-// the shutdown is complete.
+// Stop synchronously shuts down the Wallet through its private lifecycle
+// operation. The context remains only for Controller compatibility and is not
+// consulted, because terminal teardown cannot be abandoned after it begins.
 // The Manager serializes Stop with Start, so Stop can transition an
 // Initialized Wallet directly to its terminal Stopped state.
 //
 // This is part of the Controller interface.
-func (w *Wallet) Stop(stopCtx context.Context) error {
+func (w *Wallet) Stop(_ context.Context) error {
```
yyforyongyu, line 367: Note that this method will be moved to the wallet manager per task 331

GustavoStingelin (approved):

PR #1355, opened 2026-08-31
wallet: define indeterminate commit error

yyforyongyu:
## Summary

Implements roadmap Task 467.

## Change Description

Define `wallet.ErrIndeterminateCommit` as the stable public identity for ambiguous durable-mutation outcomes and cover `errors.Is` matching through contextual wrapping.

Follow-on wallet creation, account and address, genesis, and funding changes will use this identity when an ambiguous SQL commit leaves the durable result unprovable at their public boundary. Those operation-owning changes will add their mapping or reconciliation locally and wrap this sentinel with useful context.


Abdulkbk (approved):
lgtm

PR #1356, opened 2026-09-01
wallet: assemble startup candidates

yyforyongyu:
## Summary

Add the minimal backend discovery seam needed for later aggregate Manager startup.

## Change Description

- list every durable SQL Wallet through the existing exhaustive Store iterator
- load the legacy kvdb Wallet exactly once, returning either an empty result or its sole unnamed Wallet
- keep the legacy kvdb startup public passphrase on ManagerConfig with its removal TODO, copy it only into the kvdb backend, and remove the unused field from Wallet Config
- reuse the existing context-aware durable metadata lookup without adding cancellation guards around the legacy Store open
- do not add `assembleWalletCandidates`, candidate result or cleanup helpers, or an equivalent assembler; later aggregate Manager startup will consume backend discovery directly

## Notes

- base branch: `sql-wallet`; prerequisite PR #1373 has merged into that base, so this PR's review scope contains only its four Wallet discovery commits
- intentionally narrows roadmap Task 459 by omitting its planned assembler and cleanup helpers entirely rather than deferring those helpers to later work
- leaves aggregate Manager lifecycle cancellation, cleanup, and wiring for a later task

Implements roadmap Task 459

### DIFF
--- a/wallet/manager.go
+++ b/wallet/manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
 	"sort"
 	"sync"
 	"time"
@@ -173,6 +174,11 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
 
 	cfg.ChainParams = chainParams
 
+	// Startup has no load request from which to recover a legacy public
+	// passphrase, so isolate the configured bytes before the backend retains
+	// them for aggregate startup.
+	cfg.KVDBPubPassphrase = slices.Clone(cfg.KVDBPubPassphrase)
+
 	if cfg.WalletSyncRetryInterval == 0 {
 		cfg.WalletSyncRetryInterval = initialBackoff
 	}
@@ -202,6 +208,10 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
 	if err != nil {
 		return nil, translateDatabaseIdentityError(err)
 	}
+
+	// The kvdb backend now owns the cloned startup credential. Do not retain
+	// another reference in the Manager's otherwise immutable configuration.
+	cfg.KVDBPubPassphrase = nil
 
 	return &Manager{
 		wallets: make(map[string]*Wallet),
--- a/wallet/manager_config.go
+++ b/wallet/manager_config.go
@@ -53,6 +53,7 @@ var errUnsupportedBackend = errors.New("unsupported database backend")
 //	MaxConnections                     SQL backends
 //	SignetChallengeDigest              SQL backends
 //	NoFreelistSync, Timeout            kvdb only
+//	KVDBPubPassphrase                  kvdb only
 type ManagerConfig struct {
 	// Backend selects the database implementation. Required.
 	Backend DBBackend
@@ -107,6 +108,14 @@ type ManagerConfig struct {
 	// in returned array order. It is nil off signet; its first four bytes,
 	// decoded little-endian, must equal ChainParams.Net.
 	SignetChallengeDigest []byte
+
+	// KVDBPubPassphrase opens the sole legacy Wallet during aggregate
+	// startup. It is an input to Manager rather than retained Wallet policy
+	// because no caller request exists when durable Wallets are assembled.
+	//
+	// TODO(yy): Remove KVDBPubPassphrase once legacy kvdb support is fully
+	// deprecated and removed.
+	KVDBPubPassphrase []byte
 
 	// NoFreelistSync controls bbolt freelist synchronization.
 	//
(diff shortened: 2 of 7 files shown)


wallet/manager.go
```diff
@@ -172,6 +173,10 @@ func NewManager(ctx context.Context, cfg ManagerConfig) (*Manager, error) {
 	}
 
 	cfg.ChainParams = chainParams
+	// Startup has no load request from which to recover a legacy public
```
yyforyongyu, line 177: add a newline above

wallet/manager_config.go
```diff
@@ -108,6 +109,11 @@ type ManagerConfig struct {
 	// decoded little-endian, must equal ChainParams.Net.
 	SignetChallengeDigest []byte
 
+	// KVDBPubPassphrase opens the sole legacy Wallet during aggregate
+	// startup. Manager copies the bytes at construction because no caller
+	// request exists when durable Wallets are assembled.
+	KVDBPubPassphrase []byte
```
yyforyongyu, line 118: why are we adding this back? 1354 moved this from previous design, and this should be handled by the wallet? like `CreateWalletParams`?

wallet/manager_kvdb.go
```diff
@@ -96,6 +98,32 @@ func (b *kvdbManagerBackend) create(ctx context.Context,
 	return b.open(ctx, params.Name, adapterCfg)
 }
 
+// listWallets returns the sole legacy Wallet when its namespaces exist. The
+// occupancy check avoids opening an uninitialized database. The load request
+// carries only the copied startup passphrase because kvdb has no durable Wallet
+// name and cannot contain a second identity.
+func (b *kvdbManagerBackend) listWallets(
```
yyforyongyu, line 104: given that kvdb only has one wallet, there is no need to have all these `isEmpty` and other methods? we should just load the bucket and return an error if the buckets are not initialized, and, return empty or the one wallet?

wallet/manager_sql.go
```diff
@@ -9,9 +9,14 @@ import (
 	"github.com/btcsuite/btcd/chaincfg/v2"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	"github.com/btcsuite/btcwallet/wallet/internal/db/page"
 	vault "github.com/btcsuite/btcwallet/wallet/internal/keyvault"
 )
 
+// managerWalletPageLimit bounds each Store request while listWallets follows
+// the existing cursor iterator through the complete durable Wallet set.
+const managerWalletPageLimit = 100
```
yyforyongyu: this is premature - don't build the pagination yet as it's not needed, we will hardly see a couple of wallets anyway.

wallet/manager_test.go
```diff
@@ -483,6 +484,103 @@ func (b *managerBackendMock) close() error {
 	return args.Error(0)
 }
 
+// TestSQLManagerBackendListsWallets verifies ordered complete listing and
+// all-or-nothing iterator failures.
+func TestSQLManagerBackendListsWallets(t *testing.T) {
```
yyforyongyu, line 489: i'd rather split the first two commits like this:
- sql/kvdb introduces the new listWallets, hence the tests
- manager make use of it, hence the tests

wallet/manager.go
yyforyongyu: Addressed in c12a754f. There is now a blank line between the chain-parameter snapshot and the independent startup-credential block.

wallet/manager_config.go
yyforyongyu: Kept intentionally for restart. CreateWalletParams supplies the public passphrase only while creating a KVDB wallet and is unavailable after the process restarts. The planned Manager.Start(ctx) has no per-wallet load request, while waddrmgr.Open still requires the original public passphrase for a wallet created with a custom value. ManagerConfig is therefore the restart-time input. NewManager clones it into the KVDB backend and clears the Manager-retained copy; SQL ignores it.

wallet/manager_kvdb.go
yyforyongyu: Addressed. KVDB listing now calls the existing load path exactly once. A wrapped db.ErrWalletNotFound becomes a non-nil empty result, an initialized database returns its sole Wallet, and every other error propagates. The separate isEmpty transaction was removed.

wallet/manager_sql.go
yyforyongyu: Addressed by removing the Manager-specific pagination test and any new pagination mechanism. SQL now delegates exhaustive discovery to the existing Store.IterWallets API. The private batch-size constant only supplies the positive request that IterWallets requires; the Store iterator owns traversal.

wallet/manager_test.go
yyforyongyu: The follow-up scope was narrowed, so this PR no longer adds candidate assembly or a temporary Manager caller. The rewritten history is three atomic commits: the compiling backend discovery seam, focused SQL coverage, and focused KVDB coverage. Manager use and its lifecycle tests are deferred to the aggregate Manager.Start change, as stated in the PR description.

Abdulkbk (approved):
LGTM

PR #1357, opened 2026-09-01
wallet: validate transaction input amounts

Abdulkbk:
Implements [Task 404][task] — validate Wallet transaction input amounts.

`TxCreator.CreateTransaction` and `PsbtManager.FundPsbt` both fund from
Wallet-produced input sources. The automatic and manual closures accumulated their totals with
raw addition, so a negative, over-maximum, or overflowing stored amount
wrapped the total rather than being reported, and authoring then spent
that total on trust — the sufficiency test, the fee, and the change paid
back to the wallet all rest on it.

### Changes

- A wallet-owned checked addition and result validator: every value and
  the reported total must lie in `0..MaxSatoshi`, there must be one value
  per input, and the bounded aggregate must equal the reported total. The
  aggregate bound is applied before each addition, so a long enough set of
  individually valid inputs reports the bound it exceeded rather than an
  overflow.
- The validator is applied once in `createInputSource`, the boundary both
  public wrappers cross, so automatic, manual, and selected results are all
  covered without touching the sources the deprecated wrappers build.
- Both accumulating closures now use the checked add. The automatic source
  takes a coin only once its value is known to be addable — that state
  outlives the call, so a failed add must neither record an input against a
  total that never took it, nor drop the offending coin and let a later call
  succeed without it. The manual source forms its sum up front and holds a
  failure until it is asked, offering no partial selection.

The sentinels stay unexported. A caller cannot ask for a malformed input
set — the values come from the store — so reaching one means the store
returned an amount it could not have written, and there is nothing for a
caller to match on.

### Testing

`package wallet` tests drive both public calls through mocked Store results
on both selection paths, covering a negative value, a value or total above
`MaxSatoshi`, an overflowing accumulation, and a positive case on each path.
The amount sets differ per path because the sources consume differently:
automatic selection stops once the running total reaches the target, so it
can only reach an overflow by accumulating downward, while manual selection
always sums the whole selection and reaches aggregate violations automatic
never sees.

No change-script derivation is registered in the regression cases, so a
refusal arriving late enough to allocate change fails on the unexpected
mock call. Reverting the wrapper locally confirms it: the corrupt sets then
either report insufficient funds or reach `NewDerivedAddress`.

Count and reported-total mismatches are covered against the validator
directly, since both wallet sources report the total they actually
accumulated and no Store result can drive them into those classes.

### Scope

Per the task's exclusions, `wallet/txauthor` is untouched — Task 357's
private authoring arithmetic is neither exported nor consumed here — as are
the deprecated Wallet methods and `cmd/`.

Depends on #1348, now merged; rebased onto `sql-wallet`.

[task]: https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/404-transaction-input-amounts.md

### DIFF
--- a/wallet/psbt_manager_test.go
+++ b/wallet/psbt_manager_test.go
@@ -4711,3 +4711,104 @@ func TestFinalizePsbtLocked(t *testing.T) {
 	err := w.FinalizePsbt(t.Context(), packet)
 	require.ErrorIs(t, err, ErrStateForbidden)
 }
+
+// TestFundPsbtRejectsCorruptInputAmounts verifies that a store result carrying
+// an unrepresentable or inconsistent amount is refused by FundPsbt too, on both
+// the automatic and the manual selection path, without mutating the caller's
+// packet.
+//
+// Both public wrappers prepare their sources at the same boundary, so asserting
+// it here rather than only through CreateTransaction is what pins the check to
+// that boundary instead of to one wrapper. The amount sets and the violations
+// they must be refused under are shared with
+// TestCreateTransactionRejectsCorruptInputAmounts.
+func TestFundPsbtRejectsCorruptInputAmounts(t *testing.T) {
+	t.Parallel()
+
+	// fundPacket builds a packet requesting one 99,700-sat payment and
+	// funds it. A packet carrying no inputs asks the wallet to select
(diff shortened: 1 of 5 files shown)


yyforyongyu (approved):
LGTM!

PR #1358, opened 2026-09-01
wallet+db: add account manager error boundary

GustavoStingelin:
## Change Description

Implements Task [378](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/378-account-manager-error-boundary.md) by translating public AccountManager failures to wallet-owned error identities instead of exposing `db`, `waddrmgr`, or `keyvault` sentinels.

This change:

- defines the AccountManager-specific `ErrAccountAlreadyExists`, `ErrAccountOperationUnsupported`, and `ErrAccountDerivationExhausted` identities;
- normalizes missing-account, duplicate-name, invalid-request, unsupported-operation, locked-state, and derivation-exhaustion outcomes across SQL and kvdb;
- preserves cancellation, deadline, lifecycle, and invalid-account-key identities while retaining diagnostic context without leaking backend identities;
- enforces validation, lock, and occupied-name precedence before account mutation; and
- classifies only account-name uniqueness violations in SQLite and PostgreSQL, leaving unrelated constraint failures unchanged.

The internal initial-account import path remains untranslated so this PR does not introduce an AccountManager error contract into `Manager.Create`. The broader real-backend mutation-error matrix remains owned by Task [462](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/462-account-mutation-error-itests.md).

## Steps to Test

```bash
go test ./... -count=1

go test -race -shuffle=on ./wallet ./wallet/internal/db/pg \
  ./wallet/internal/db/sqlite \
  -run 'Test(AccountManager|AccountInfo|NewAccount|GetAccount|ListAccounts|RenameAccount|ImportAccount|IsAccountNameConflict)' \
  -count=1

scripts/check-each-commit.sh origin/sql-wallet

make itest-db db=sqlite
make itest-db db=postgres
```

### Testing

Please see our [Contribution Guidelines](https://github.com/btcsuite/btcwallet/blob/master/docs/developer/contribution_guidelines.md) for further guidance.

### DIFF
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -350,6 +350,33 @@ type newAccountReq struct {
 	resp  chan accountResp
 }
 
+// requireAccountNameAvailable reports a name already taken within scope as
+// ErrAccountAlreadyExists. The lookup is read-only and skips the balance query
+// because only the name matters here. An absent account leaves the name
+// available, a missing scope included, so the first account of a scope can
+// still be created.
+func (w *Wallet) requireAccountNameAvailable(ctx context.Context,
+	scope waddrmgr.KeyScope, name string) error {
+
+	_, err := w.cache.GetAccount(ctx, db.GetAccountQuery{
+		WalletID:    w.id,
+		Scope:       db.KeyScope(scope),
+		Name:        &name,
+		SkipBalance: true,
+	})
+	if err == nil {
+		return fmt.Errorf("%w: %q in scope %d/%d",
+			ErrAccountAlreadyExists, name, scope.Purpose, scope.Coin)
+	}
+
+	translated := accountManagerErr(err)
+	if errors.Is(translated, ErrAccountNotFound) {
+		return nil
+	}
+
+	return translated
+}
+
 // NewAccount creates the next account and returns its account info. The name
 // must be unique under the key scope. In order to support automatic seed
 // restoring, new accounts may not be created when all of the previous 100
@@ -366,6 +393,22 @@ func (w *Wallet) NewAccount(ctx context.Context, scope waddrmgr.KeyScope,
 	err = ctx.Err()
 	if err != nil {
 		return nil, err
+	}
+
+	err = waddrmgr.ValidateAccountName(name)
+	if err != nil {
+		return nil, fmt.Errorf("%w: %s", ErrInvalidParam, err.Error())
+	}
+
+	// Hardened derivation needs the master HD private key, which stays sealed
+	// while the wallet is locked. A watch-only wallet holds no such key at
+	// all, so it is refused as unsupported below instead of being reported as
+	// locked.
+	if !w.IsWatchOnly() {
+		err = w.state.canSign()
+		if err != nil {
+			return nil, err
+		}
 	}
 
 	req := newAccountReq{
@@ -391,10 +434,31 @@ func (w *Wallet) NewAccount(ctx context.Context, scope waddrmgr.KeyScope,
 // handleNewAccount performs derivation and persistence only after mainLoop
 // admits the request; handleReq owns its shutdown completion.
 func (w *Wallet) handleNewAccount(req newAccountReq) {
-	deriveFn, err := w.buildAccountDeriveFn(req.ctx)
+	// An occupied name is reported ahead of the watch-only refusal and ahead
+	// of an exhausted derivation range, so the caller hears about the part of
+	// the request it can restate.
+	err := w.requireAccountNameAvailable(req.ctx, req.scope, req.name)
 	if err != nil {
 		req.resp <- accountResp{err: err}
+		return
+	}
 
+	// Refuse before any write is attempted. Hardened account derivation needs
+	// the master HD private key, which a watch-only wallet does not hold, and
+	// the backends otherwise refuse at different points: the SQL stores reach
+	// the derivation callback while a rootless legacy wallet fails its scope
+	// lookup first and would report a missing account instead.
+	if w.IsWatchOnly() {
+		req.resp <- accountResp{
+			err: accountManagerErr(errWatchOnlyAccountDerivation),
+		}
+
+		return
+	}
+
+	deriveFn, err := w.buildAccountDeriveFn(req.ctx)
+	if err != nil {
+		req.resp <- accountResp{err: accountManagerErr(err)}
 		return
 	}
 
@@ -406,17 +470,16 @@ func (w *Wallet) handleNewAccount(req newAccountReq) {
 		}, deriveFn,
 	)
 	if err != nil {
-		// Preserve the legacy waddrmgr.ManagerError contract so that
-		// callers using waddrmgr.IsError(err, ...) keep working after
-		// kvdb wraps the underlying manager error via fmt.Errorf.
-		var mErr waddrmgr.ManagerError
-		if errors.As(err, &mErr) {
-			req.resp <- accountResp{err: mErr}
+		mappedErr := accountManagerErr(err)
+		if errors.Is(err, db.ErrUnknownKeyScope) &&
+			!errors.Is(mappedErr, context.Canceled) &&
+			!errors.Is(mappedErr, context.DeadlineExceeded) {
 
-			return
+			mappedErr = fmt.Errorf("%w: %s", ErrInvalidParam,
+				err.Error())
 		}
 
-		req.resp <- accountResp{err: err}
+		req.resp <- accountResp{err: mappedErr}
 
 		return
 	}
@@ -424,7 +487,7 @@ func (w *Wallet) handleNewAccount(req newAccountReq) {
 	account, err := w.accountInfoFromStore(info)
 	req.resp <- accountResp{
 		info: account,
-		err:  err,
+		err:  accountManagerErr(err),
 	}
 }
 
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -209,6 +209,9 @@ func (w *Wallet) buildAccountDeriveFn(
 //     account is created for each of the default key scopes and CAN be renamed.
 //   - "imported": A special account that holds all individually imported keys.
 //     This account is global and CANNOT be renamed.
+//
+// Errors use wallet-owned sentinels without exposing backend error identities.
+// Context cancellation and deadlines are preserved.
 type AccountManager interface {
 	// NewAccount creates a new account for a given key scope and name. The
 	// provided name must be unique within that key scope.
@@ -242,10 +245,10 @@ type AccountManager interface {
 		oldName string, newName string) error
 
 	// ImportAccount imports an account from an extended public key.
-	// Private extended keys are rejected. The key scope is derived from
-	// the version bytes of the extended key. The account name must be
-	// unique within the derived scope. If dryRun is true, the import is
-	// validated but not persisted. SQL wallets accept this XPub-only
+	// Invalid or private keys return ErrInvalidAccountKey. The key scope is
+	// derived from the version bytes of the extended key. The account name
+	// must be unique within the derived scope. If dryRun is true, the import
+	// is validated but not persisted. SQL wallets accept this XPub-only
 	// material only when the wallet is watch-only under ADR 0012. The
 	// legacy kvdb backend retains its grandfathered mixed-mode import
 	// behavior until migration; neither path imports signing material.
@@ -904,7 +907,7 @@ type importAccountReq struct {
 // ImportAccount imports an account from an extended public key. Private
 // extended keys are rejected. The key scope is derived from the version
 // bytes of the extended key. The account name must be unique within the
-// derived scope.
+// derived scope. Invalid account keys return ErrInvalidAccountKey.
 //
 // SQL wallets accept this XPub-only material only when the wallet is
 // watch-only under ADR 0012. The legacy kvdb backend retains its grandfathered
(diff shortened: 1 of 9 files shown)


GustavoStingelin:
@saubyk pls, add this one to the board

yyforyongyu:
Task 378 is simplified, could you ask your agent to load it again? Thanks!

GustavoStingelin:
@yyforyongyu, ready for a new review

wallet/internal/db/pg/account_test.go
```diff
@@ -0,0 +1,87 @@
...
+// The unique indexes the accounts table declares. Both are unique-violation
+// sources, so the classifier has to tell them apart by name.
+const (
+	accountNameIndex   = "uidx_accounts_wallet_scope_account_name"
+	accountNumberIndex = "uidx_accounts_scope_account_number"
+)
+
+// uniqueViolation builds the driver error PostgreSQL reports when an insert
+// collides with the named unique index.
+func uniqueViolation(index string) error {
```
yyforyongyu: this method is just a shallow shim? why having this?

wallet/internal/db/pg/account_test.go
```diff
@@ -0,0 +1,87 @@
...
+		err  error
+	}{{
+		// A colliding account number is a distinct outcome the wallet
+		// reports separately, even though it shares the SQLSTATE.
+		name: "unique violation on the account-number index",
+		err:  uniqueViolation(accountNumberIndex),
+	}, {
+		// The name index appears in errors that are not collisions.
+		name: "foreign-key violation naming the account-name index",
+		err: &pgconn.PgError{
```
yyforyongyu, line 53: and why this is not wrapped like `uniqueViolation`? seems quite arbitrary?

wallet/internal/db/sqlite/account_test.go
```diff
@@ -0,0 +1,111 @@
...
+func provokeAccountConflicts(t *testing.T) accountConflicts {
+	t.Helper()
+
+	conn, err := sql.Open("sqlite", ":memory:")
+	require.NoError(t, err)
+	conn.SetMaxOpenConns(1)
+	t.Cleanup(func() { require.NoError(t, conn.Close()) })
+
+	_, err = conn.ExecContext(t.Context(), `
+		CREATE TABLE accounts (
```
yyforyongyu: why do we need to use raw sql here?

wallet/account_manager.go
```diff
@@ -27,8 +27,118 @@ import (
...
+)
+
+// accountManagerErr translates a store, legacy manager, or Vault failure into
+// the wallet-owned identity for that outcome. Only the source text is kept so
+// no internal identity crosses the boundary; unclassified errors keep only
+// their text, and caller-owned cancellation keeps its identity.
+//
+//nolint:cyclop // One explicit mapping switch is the clearest form here.
+func accountManagerErr(err error) error {
+	var identity error
```
yyforyongyu: nit: can we find a better name other than `identity`? why is this called `identity`?

wallet/account_manager.go
```diff
@@ -27,8 +27,118 @@ import (
...
+	}
+
+	return fmt.Errorf("%w: %s", identity, err.Error())
+}
+
+// isManagerErr reports whether err carries a waddrmgr.ManagerError with the
+// given code. waddrmgr.IsError is a bare type assertion, so it cannot see a
+// manager error that a store wrapped for context, which is how the legacy
+// backend returns every one of them.
+func isManagerErr(err error, code waddrmgr.ErrorCode) bool {
```
yyforyongyu: `isAddrMgrErr`? `Manager` is too generic and the caller has no idea which manger is this referring to

wallet/account_manager_test.go
```diff
@@ -644,6 +656,159 @@ func TestGetAccountIncludesImportedPseudoAccount(t *testing.T) {
...
+			// Act: read the account by name.
+			account, err := w.GetAccount(
+				t.Context(), waddrmgr.KeyScopeBIP0084,
+				testAccountName,
+			)
+
+			// Assert: only the wallet-owned identity is reported.
+			require.Nil(t, account)
+			assertAccountManagerError(t, err, tc.inject, tc.want)
+			deps.store.AssertExpectations(t)
```
yyforyongyu: these assertions are unnecessary as mocks have already been asserted in the setup during `t.Cleanup`

wallet/account_manager.go
```diff
@@ -27,8 +27,118 @@ import (
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet/internal/addresstype"
 	"github.com/btcsuite/btcwallet/wallet/internal/db"
+	"github.com/btcsuite/btcwallet/wallet/internal/db/pg"
```
yyforyongyu: similar issue: https://github.com/btcsuite/btcwallet/pull/1363#discussion_r3942696914

wallet/account_manager.go
```diff
@@ -223,6 +223,18 @@ func (w *Wallet) buildAccountDeriveFn(
 //     account is created for each of the default key scopes and CAN be renamed.
 //   - "imported": A special account that holds all individually imported keys.
 //     This account is global and CANNOT be renamed.
+//
+// # Errors
+//
+// Named failures are reported through wallet-owned sentinels —
```
yyforyongyu: I don't think we wanna have this kinda docs - they are too detailed and hard to maintain.

wallet/account_manager.go
```diff
@@ -354,6 +354,19 @@ func (w *Wallet) accountInfoFromStore(
 	}, nil
 }
 
+// validateAccountRequest performs the admission checks every public
+// AccountManager method shares. The wallet must be started, and the caller's
+// context must still be live, so a request whose caller already gave up is
+// refused before the method's own refusals and before any Store work.
+func (w *Wallet) validateAccountRequest(ctx context.Context) error {
```
yyforyongyu: Is this really a good check? Why are we mixing the internal wallet state check into a `accountRequest` check, while the latter feels more like asserting the params of a `NewAccount` request? Can we remove this new method instead

yyforyongyu (changes requested):
Initially i thought we just wanna define public errors and map them internally, guess we need a bit more than that. Will post more after synthesizing my agentic findings.

wallet/account_manager.go
```diff
@@ -27,8 +27,118 @@ import (
...
+	case errors.Is(err, db.ErrAccountNotFound),
+		errors.Is(err, db.ErrKeyScopeNotFound),
+		isManagerErr(err, waddrmgr.ErrAccountNotFound),
+		isManagerErr(err, waddrmgr.ErrScopeNotFound):
+
+		identity = ErrAccountNotFound
+
+	case pg.IsAccountNameConflict(err), sqlite.IsAccountNameConflict(err),
+		isManagerErr(err, waddrmgr.ErrDuplicateAccount),
+		isManagerErr(err, waddrmgr.ErrAlreadyExists):
```
yyforyongyu: This mapping is incorrect: `waddrmgr.ErrAlreadyExists` means the address-manager database already exists, while account-name collisions use `ErrDuplicateAccount`; please remove the unrelated case so an unexpected internal failure cannot match `ErrAccountAlreadyExists`.

wallet/account_manager.go
```diff
@@ -27,8 +27,118 @@ import (
...
+
+	// waddrmgr reports every account-name rule violation, including the
+	// reserved and empty-name cases, under ErrInvalidAccount.
+	case errors.Is(err, db.ErrMissingAccountName),
+		errors.Is(err, db.ErrMissingAccountPublicKey),
+		errors.Is(err, db.ErrMissingField),
+		errors.Is(err, db.ErrInvalidParam),
+		errors.Is(err, db.ErrReservedAccountName),
+		errors.Is(err, db.ErrInvalidAccountQuery),
+		errors.Is(err, db.ErrUnknownKeyScope),
```
yyforyongyu: `ListAccounts` has no scope argument, yet the shared mapper turns an unexpected `db.ErrUnknownKeyScope` from that method into caller-facing `ErrInvalidParam`; please translate validation errors only at exits where the public request can actually cause them and leave unexpected backend errors unnamed.

wallet/account_manager.go
```diff
@@ -27,8 +27,118 @@ import (
...
+		return err
+
+	case errors.Is(err, db.ErrAccountNotFound),
+		errors.Is(err, db.ErrKeyScopeNotFound),
+		isManagerErr(err, waddrmgr.ErrAccountNotFound),
+		isManagerErr(err, waddrmgr.ErrScopeNotFound):
+
+		identity = ErrAccountNotFound
+
+	case pg.IsAccountNameConflict(err), sqlite.IsAccountNameConflict(err),
```
yyforyongyu: This makes the wallet error boundary enumerate the PostgreSQL and SQLite implementations directly, so adding or replacing a Store requires changing wallet business logic; please carry one backend-neutral account-name-conflict classification through the shared DB error layer and match that here.

wallet/account_manager_test.go
```diff
@@ -753,34 +961,892 @@ func TestRenameAccount(t *testing.T) {
...
+		name:   "sql derivation range exhausted",
+		inject: db.ErrMaxAccountNumberReached,
+		want:   ErrAccountDerivationExhausted,
+	}, {
+		name:   "kvdb derivation range exhausted",
+		inject: wrapped(legacyErr(waddrmgr.ErrAccountNumTooHigh)),
+		want:   ErrAccountDerivationExhausted,
+	}, {
+		name:   "kvdb duplicate account",
+		inject: wrapped(legacyErr(waddrmgr.ErrDuplicateAccount)),
```
yyforyongyu, line 2257: The SQL duplicate-name path is tested only inside the backend classifiers, while these public-boundary cases inject only the legacy `ErrDuplicateAccount`; please inject a SQL name-conflict error through a public `AccountManager` method and assert `ErrAccountAlreadyExists` without leaking the raw driver identity.

wallet/account_manager.go
```diff
@@ -113,6 +223,18 @@ func (w *Wallet) buildAccountDeriveFn(
 //     account is created for each of the default key scopes and CAN be renamed.
 //   - "imported": A special account that holds all individually imported keys.
 //     This account is global and CANNOT be renamed.
+//
+// # Errors
+//
+// Named failures are reported through wallet-owned sentinels —
+// ErrAccountNotFound, ErrAccountAlreadyExists,
+// ErrAccountOperationUnsupported, ErrAccountDerivationExhausted,
+// ErrInvalidParam and ErrStateForbidden — so callers match on the wallet
```
yyforyongyu: The public sentinel list omits `ErrInvalidAccountKey`, even though `ImportAccount` returns that identity unchanged; please include it in the caller-facing contract or document it directly on the method.

wallet/account_manager_test.go
```diff
@@ -753,34 +961,892 @@ func TestRenameAccount(t *testing.T) {
...
+			reported = append(reported, identity.Error())
+		}
+	}
+
+	return reported
+}
+
+// legacyErr builds the error the legacy waddrmgr backend reports for code.
+func legacyErr(code waddrmgr.ErrorCode) error {
+	return waddrmgr.ManagerError{ErrorCode: code}
```
yyforyongyu: `legacyErr` leaves `Description` empty, so `injected.Error()` is empty and the later `ErrorContains` diagnostic-preservation assertion always passes; please give the fixture a non-empty description.

wallet/internal/db/pg/account_test.go
```diff
@@ -0,0 +1,87 @@
...
+	"testing"
+
+	"github.com/jackc/pgx/v5/pgconn"
+	"github.com/stretchr/testify/require"
+)
+
+// The unique indexes the accounts table declares. Both are unique-violation
+// sources, so the classifier has to tell them apart by name.
+const (
+	accountNameIndex   = "uidx_accounts_wallet_scope_account_name"
```
yyforyongyu: These test-local index names duplicate the migration-owned schema, and the SQLite test separately recreates the accounts table with raw SQL, so both classifier tests can pass after the real schema drifts; please derive the fixtures from the migrated schema or shared authoritative metadata.

wallet/account_manager_test.go
```diff
@@ -753,34 +961,892 @@ func TestRenameAccount(t *testing.T) {
...
+	require.NotErrorIs(t, got, injected)
+	requireNoInternalIdentity(t, got)
+}
+
+// accountManagerCall names one call per public AccountManager method. The
+// shared admission checks refuse every request before it reaches the Store, so
+// the arguments only have to be well formed.
+type accountManagerCall struct {
+	name string
+	call func(ctx context.Context, w *Wallet) error
```
yyforyongyu: This function-valued table hides all seven public subject calls behind `tc.call`, while `assertAccountManagerError` also hides the identity and non-leakage assertions; please use data-only cases with the public calls and contract assertions visible in their subtests.

wallet/account_manager.go
```diff
@@ -511,62 +718,115 @@ func (w *Wallet) ImportAccount(ctx context.Context,
...
+	}
+
+	return result, nil
+}
+
+// importRequestErr translates a rejected import request. Key-material
+// validation is already wallet-owned and keeps its identity; every other
+// rejection is a request the wallet cannot serve as written, such as version
+// bytes and an address type that together select no key scope.
+func importRequestErr(err error) error {
```
yyforyongyu: `importRequestErr` is a single-use translation shim that neither isolates a side effect nor removes duplication; please inline its small branch at the call site.

yyforyongyu (changes requested):

wallet/internal/db/pg/account_test.go
GustavoStingelin: Removed the shim; the tests now call `Store.ClassifyError` directly.

wallet/internal/db/pg/account_test.go
GustavoStingelin: Made this consistent: both positive and negative cases now exercise bare and wrapped driver errors.

wallet/internal/db/sqlite/account_test.go
GustavoStingelin: Replaced the handwritten schema and SQL with the real migrations and generated sqlc queries.

wallet/account_manager.go
GustavoStingelin: Renamed it to `mappedErr`.

wallet/account_manager.go
GustavoStingelin: Renamed it to `isAddrMgrErr`.

wallet/account_manager_test.go
GustavoStingelin: Removed the redundant `AssertExpectations` calls; fixture cleanup already verifies them.

wallet/account_manager.go
GustavoStingelin: Wrapped every commit body at 72 columns.

wallet/account_manager.go
GustavoStingelin: Reduced this to a short interface-level contract and kept method-specific details next to their methods.

wallet/account_manager.go
GustavoStingelin: Removed `validateAccountRequest`. Lifecycle/context admission is explicit in each public method, while parameter validation stays with the operation.

wallet/account_manager.go
GustavoStingelin: Removed that case. Account collisions use `ErrDuplicateAccount`; `ErrAlreadyExists` now falls through to diagnostic-only handling.

wallet/account_manager.go
GustavoStingelin: Moved request-specific validation mapping to the relevant exits. Unexpected `ListAccounts` validation errors now keep their diagnostic text without a wallet sentinel; a public-method test covers that case.

wallet/account_manager.go
GustavoStingelin: Added `SQLError.Constraint` and shared `dberr.IsAccountNameConflict`; each backend sets the account-name marker in its own classifier. `AccountManager` no longer imports SQL backends. I kept `ReasonConstraint` unchanged so retry policy and error counters are unaffected, and put this change in a separate commit.

wallet/account_manager_test.go
GustavoStingelin: `TestRenameAccountTranslatesPostgresNameConflict` now injects a classified SQL conflict through `RenameAccount` and checks that only `ErrAccountAlreadyExists` escapes, not the SQL wrapper or raw driver. The real-Store classifier assertion covers the backend side; the broader public integration matrix stays in Tasks 430/462.

wallet/account_manager.go
GustavoStingelin: Documented `ErrInvalidAccountKey` directly on `ImportAccount`, including its interface declaration.

wallet/account_manager_test.go
GustavoStingelin: Added a non-empty description to `legacyErr`, so the diagnostic-preservation assertions are no longer vacuous.

wallet/internal/db/pg/account_test.go
GustavoStingelin: SQLite fixtures now apply the real migrations and use generated queries. PostgreSQL unit tests share the production index constant, and the existing migrated-Store duplicate-name test now checks the marker on both databases. That catches schema drift without adding another integration harness.

wallet/account_manager_test.go
GustavoStingelin: Split the function-valued table into method-specific tests with data-only cases and visible calls/assertions. There is a little repetition, but each test can now be read on its own; the redundant omnibus assertion helper is gone.

wallet/account_manager.go
GustavoStingelin: Inlined it in `handleImportAccount`, retaining the `ErrInvalidAccountKey` special case.

GustavoStingelin (commented):
Rebased onto latest sql-wallet and folded the fixes into the seven-commit history. Replies below describe the changes; broader query/mutation integration coverage remains in Tasks 430/462.

wallet/internal/db/err/errors.go
```diff
@@ -252,6 +252,18 @@ func (r Reason) Valid() bool {
...
+type Constraint uint8
+
+const (
+	// ConstraintUnknown leaves an unrecognized constraint unclassified.
+	ConstraintUnknown Constraint = iota
+
+	// ConstraintAccountName identifies a duplicate account name within a
+	// wallet and key scope, not an account-number or other unique violation.
+	ConstraintAccountName
+)
```
GustavoStingelin, line 265: @yyforyongyu I was avoiding doing this before the review to keep the scope tighter, but here we go  🏎️

yyforyongyu (approved):
LGTM - think there are some improvements and simplifications we can make, but will address them in a different PR!

PR #1359, opened 2026-09-01
wallet+txauthor: preserve PSBT metadata during funding

Abdulkbk:
[Task 346 — preserve PSBT metadata during funding][task]

Rebased onto `sql-wallet` and reworked for @yyforyongyu's review: the sign and
finalize modes are gone, the whole change lives in `wallet`, sorting is
`psbt.InPlaceSort`, and the series is now ten small commits.

### Problem

`FundPsbt` rewrites the caller's own packet in place and rebuilds it from the
authored transaction, which loses or corrupts most of what the caller put in:

- The caller's packet is left **partially rewritten on any error**, with no way
  to tell how far funding got.
- **Input and output metadata is wiped**: witness scripts, taproot leaf
  scripts, merkle roots, explicit sighash types.
- **Version, locktime and per-input sequence numbers are reset** to authoring's
  defaults, silently disabling a caller's locktime and withdrawing its opt-in
  to replacement.
- The **change output is located by comparing value and script**, so a caller
  paying itself the same amount it gets back as change is told the wrong change
  index.
- Validation is shallow, runs partly after a change address has been derived,
  and **panics** on a packet carrying a nil transaction input.

### The series

Structural validation arrives first, then the funding transformation, then the
wiring:

| | commit |
|--|--|
| 1 | packet structural validation — cardinality via `psbt.VerifyInputOutputLen`, nil-entry guards, unsigned-tx check, duplicate outpoints, unclassified fields |
| 2 | input UTXO records — parent hashes to the outpoint, index in range, both views agree |
| 3 | input sighash types — only types that exist |
| 4 | the funding admission check — no signatures, no `SIGHASH_SINGLE` |
| 5 | the deep-clone primitive |
| 6 | caller input metadata by outpoint |
| 7 | the change-output swap mirrored onto output records |
| 8 | the change output tracked by identity through sorting |
| 9 | `FundPsbt` wiring |
| 10 | the funding contract, documented in the godoc and the workflow guide |

Commits 1–3 are the reusable side-effect-free core: no wallet state, no
lookups, no mutation of what they are handed. Commit 4 is the funding-specific
layer on top, and the only part wired into `FundPsbt`. Nothing here anticipates
Tasks 347 or 407 — signature-structure validation and the sign/finalize modes
are theirs to add when they need them.

### Fixed since the last review

- **Reachable panics.** A caller can assemble a `Packet` as a plain struct, so
  it can hold a nil `TxIn`, and a nil entry can also sit inside the parent
  transaction an input carries — hashing that parent to compare it against the
  outpoint reads straight through its own inputs and outputs. Both panicked
  instead of returning `ErrPacketMalformed`. Every pointer-bearing list is now
  swept first, the nested parent included (commit 1).
- **`0x80` was admitted.** `SIGHASH_DEFAULT | ANYONECANPAY` is not a valid
  sighash for any spending path — the default type is the absence of a type, so
  there is nothing for the modifier to modify — but the base masked to
  `DEFAULT` and the modifier was allowed, so it passed. Now rejected, with the
  negative case covered (commit 3).
- **Cross-family derivation.** A caller taproot derivation on a wallet BIP32
  input reached the merge, leaving a packet carrying both families, which
  `SignPsbt` rejects as `ErrAmbiguousDerivation`. Now reported as a conflict,
  both ways round (commit 6).
- **Duplicated validation authority.** The transaction-to-record cardinality
  rule now comes from `psbt.VerifyInputOutputLen`, wrapped in
  `ErrPacketMalformed`, rather than a second copy that could drift.
- **`NonWitnessUtxo` and taproot leaf hashes**, neither of which the wallet
  derives for a taproot input, are carried rather than dropped or compared
  (commit 6).
- **A stale workflow example.** The `PsbtManager` doc walked the caller from
  `FundPsbt` into signing the *bare* packet. It now carries the funded one, and
  the interface says the caller's packet is not modified. `docs/developer/psbt_workflows.md` had the same problem, and was signing a variable it never defined (commit 10).

### Fixed in this round

- **The clone was undermined by the next call.** `createTxIntent` read the
  *caller's* packet and copied each `wire.TxOut` by value, which shares its
  `PkScript`, so the funded packet came back holding the caller's script
  memory. Confirmed with a repro: writing to the funded packet changed the
  caller's. It now builds from the clone, with a regression test that scribbles
  over every returned script (commit 9).
- **A signer policy was refusing valid packets.** The gate called
  `validateDerivation`, whose only other caller is `signPsbtInput` and whose
  own comment calls it a signing policy. It caps an input at one derivation
  record, so funding refused multisig and coordinator packets — metadata Task
  346 requires preserving. The gate now validates derivation *structure*
  instead, and the merge reconciles the wallet's record as one member of the
  caller's set, keeping the rest (commits 1 and 6).
- **Finalized inputs were missed.** `psbt.isFinalized` asks whether
  `FinalScriptSig`/`FinalScriptWitness` are non-nil; the check used `len() > 0`,
  so an empty-but-present record passed and was then dropped. Now tested for
  presence, with the empty cases covered (commit 4).
- **Known records went unchecked.** The gate verified that records were
  present, never that they were usable, so a derivation carrying a key the psbt
  parser rejects could pass and come back in a packet the caller could not
  serialize. Keys are now parsed, and a key named twice is refused (commit 1).

### Test rework

Against `docs/developer/unit_testing_guidelines.md`, which I had not followed:

- **§2, data-only cases.** The `mutate func(*psbt.Packet)` closures are gone
  from all five tables. Cases whose setup differs structurally are standalone
  functions; cases that vary only by data (sighash types, which record is nil,
  which signature form) stay tabular with data-only structs.
- **§3, Arrange/Act/Assert**, labelled, matching the existing files.
- **§6, `AssertExpectations`** where a test owns the expectation. Note it is
  *not* called on `mocks.syncer`: the harness starts the syncer's `run` loop in
  the background, so asserting it is racy — it cost me an intermittent failure
  before I tracked it down.
- The **clone independence oracle** no longer uses `clonePacket` to snapshot the
  original. It compares the wire encoding, so a shallowly-copied field can no
  longer make the snapshot share storage with the original and agree with it.
- The **`NonWitnessUtxo` fixture** now builds its outpoint from
  `parent.TxHash()` with a valid index, so it is a state the validator would
  actually admit, and preservation is also exercised end to end through
  `FundPsbt` against a taproot input.
- The **malformed-packet cases are now a data-only table at the `FundPsbt`
  boundary**, each asserting that no store, chain or key-vault call was made,
  with the validator-level duplicates removed. One case did not move: a packet
  holding a nil `TxIn` cannot be serialized, so it cannot take the byte
  snapshot the table's other assertions rely on, and it stays a unit test.
- **Derivation fixtures carry keys that parse**, rather than repeated bytes,
  so they are states the gate would actually admit.
- **Shallow `// Arrange` / `// Act` / `// Assert` labels are gone** — 67 of
  them. The ones that name what is being set up or checked stay.

### Known ordering, worth a reviewer's eye

Contradiction rejection (`ErrConflictingInputMetadata`) happens in the merge,
which runs after authoring — so a rejected contradiction has already cost a
freshly derived change address. The task's "before change-allocation"
requirement applies to the structural gate, which does run first, and this is
not a regression: any decoration failure already burned an address at the same
point. Moving it earlier means pre-fetching the wallet's input facts before
authoring, which reaches into change allocation and is out of scope here.

### Behaviour changes worth flagging

1. `FundPsbt` returns a validated clone rather than the caller's pointer, and
   leaves the caller's packet untouched on success and on failure. Two existing
   tests asserted the old contract and now assert this one.
2. Packets carrying unclassified (`Unknowns`) fields are rejected rather than
   silently stripped, at global, input and output level.
3. A change output missing after sorting is now an error rather than a silent
   `-1`. It cannot happen, but an error beats reporting no change on a
   transaction that has one.

No funding lease behaviour is added. Out of scope and unchanged: amount
arithmetic, size and dust calculation, input leasing and cleanup,
watched-change delivery, signing, and the Vault API.

[task]: https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/346-psbt-funding-metadata.md


yyforyongyu (changes requested):
With the updated task, i think we can keep the implementation inside `wallet` without modifying txauthor, which gives us a minimal PR, sth like,
  - Validate, then clone the caller’s PSBT.
  - Preserve version, locktime, sequences, and input/output metadata.
  - Use the existing `AuthoredTx.ChangeIndex` to mirror the change-output swap onto `POutput` metadata.
  - Save the change output pointer, call canonical `psbt.InPlaceSort`, then find its final index by pointer identity.
  - Preserve valid caller metadata, including `NonWitnessUtxo` and taproot leaf hashes.
  - Remove `OutputOrigin`, `ChangeOutputOrigin`, custom BIP69 sorting, their tests, and deprecated API work.

  This meets Task 346 without expanding the txauthor API or duplicating sorting logic.

Abdulkbk:
> With the updated task, i think we can keep the implementation inside `wallet` without modifying txauthor, which gives us a minimal PR, sth like,
> 
> * Validate, then clone the caller’s PSBT.
> * Preserve version, locktime, sequences, and input/output metadata.
> * Use the existing `AuthoredTx.ChangeIndex` to mirror the change-output swap onto `POutput` metadata.
> * Save the change output pointer, call canonical `psbt.InPlaceSort`, then find its final index by pointer identity.
> * Preserve valid caller metadata, including `NonWitnessUtxo` and taproot leaf hashes.
> * Remove `OutputOrigin`, `ChangeOutputOrigin`, custom BIP69 sorting, their tests, and deprecated API work.
> 
> This meets Task 346 without expanding the txauthor API or duplicating sorting logic.

Alright, let me rework it.

Abdulkbk:
Narrowed the scope of the PR

wallet/psbt_packet.go
```diff
@@ -0,0 +1,455 @@
...
+const (
+	// psbtOpFund admits a packet that the wallet is about to fund. Funding
+	// rewrites the transaction, so the packet must carry no signatures at
+	// all.
+	psbtOpFund psbtOperation = iota
+
+	// psbtOpSign admits a packet that the wallet is about to add
+	// signatures to. Signatures from other parties are expected here and
+	// are preserved untouched.
+	psbtOpSign
```
yyforyongyu: These sign/finalize modes have no production callers: validatePacket is currently called only with psbtOpFund. Task 346 explicitly excludes signing and assigns that behavior to Tasks 347/407. Please keep a reusable side-effect-free structural core plus a Fund-specific admission check, and defer the sign/finalize modes and signature-structure validators to their owning tasks.

wallet/psbt_packet.go
```diff
@@ -0,0 +1,455 @@
...
+	if packet == nil || packet.UnsignedTx == nil {
+		return ErrPacketNil
+	}
+
+	tx := packet.UnsignedTx
+
+	// The per-input and per-output metadata is addressed by position, so a
+	// packet whose metadata is a different length than its transaction
+	// cannot be interpreted at all.
+	if len(packet.Inputs) != len(tx.TxIn) {
```
yyforyongyu: This duplicates the transaction-to-record cardinality invariant already owned by psbt.VerifyInputOutputLen. Please call that function and wrap its error with ErrPacketMalformed so there is only one validation authority.

wallet/psbt_packet.go
```diff
@@ -0,0 +1,455 @@
...
+
+	if len(packet.Outputs) != len(tx.TxOut) {
+		return fmt.Errorf("%w: %d outputs but %d output records",
+			ErrPacketMalformed, len(tx.TxOut), len(packet.Outputs))
+	}
+
+	// A PSBT's transaction carries its unlocking scripts in the per-input
+	// records, never inline, so anything inline means the packet is not
+	// the unsigned template it claims to be.
+	err := packet.SanityCheck()
```
yyforyongyu: A caller can construct a Packet containing nil TxIn/TxOut or nil pointer-bearing metadata records. SanityCheck dereferences TxIn entries, and clonePacket later dereferences other records, so these inputs can panic instead of returning ErrPacketMalformed. Please validate the relevant pointers before calling SanityCheck or cloning.

wallet/psbt_packet.go
```diff
@@ -0,0 +1,455 @@
...
+			"flags %#x", ErrUnsafeSighash, idx, uint32(sigHash))
+	}
+
+	// ANYONECANPAY is masked off above and so cannot appear in base, which
+	// the exhaustive check has no way to see.
+	//nolint:exhaustive
+	switch base {
+	// SigHashDefault is only meaningful for taproot inputs, but it is
+	// indistinguishable from an unset field, so it is admitted for all.
+	case txscript.SigHashDefault, txscript.SigHashAll,
```
yyforyongyu, line 385: This admits 0x80 (SIGHASH_DEFAULT | ANYONECANPAY): the base is DEFAULT and the modifier is considered allowed, but that combination is not a valid Taproot or legacy sighash. Please reject ANYONECANPAY when the base is SigHashDefault and add the negative case.

wallet/psbt_funding.go
```diff
@@ -0,0 +1,359 @@
...
+	if decorated.NonWitnessUtxo == nil {
+		decorated.NonWitnessUtxo = caller.NonWitnessUtxo
+	}
+
+	// Taproot derivations carry leaf hashes, which say which scripts the
+	// key signs for. The wallet never derives those, so where the caller
+	// and the wallet agree on the key itself, the caller's records are the
+	// ones to keep: they hold everything the wallet's do and the leaf
+	// hashes besides.
+	if len(caller.TaprootBip32Derivation) > 0 {
```
yyforyongyu, line 112: checkDerivationAgrees compares only matching derivation families. A caller Taproot derivation on a wallet BIP32 input therefore reaches this assignment and returns a packet carrying both families, which SignPsbt rejects as ambiguous; the inverse case silently loses caller metadata. Please reject cross-family caller/wallet derivations before copying either representation.

wallet/psbt_manager.go
```diff
@@ -625,70 +632,109 @@ func (w *Wallet) FundPsbt(ctx context.Context, intent *FundIntent) (
...
-		ctx, intent.Packet, authoredTx,
+	funded, changeIndex, err := w.populatePsbtPacket(
+		ctx, packet, authoredTx,
 	)
 	if err != nil {
 		return nil, 0, err
 	}
 
-	return packet, changeIndex, nil
+	return funded, changeIndex, nil
```
yyforyongyu: FundPsbt now returns a funded clone, but the public PsbtManager workflow example above still passes barePacket to SignPsbt, FinalizePsbt, and Extract. Please update those steps to use fundedPacket and document on the interface that the caller packet is not modified.

wallet/psbt_funding_test.go
```diff
@@ -0,0 +1,584 @@
...
+// transaction survives on an input the wallet records none for.
+//
+// The wallet attaches a parent only to segwit v0 inputs, so a taproot input
+// comes back from decoration without one. Dropping the caller's would lose a
+// record it is entitled to have back, and the validator has already checked
+// that the parent hashes to the outpoint being spent.
+func TestRestoreInputMetadataKeepsNonWitnessUtxo(t *testing.T) {
+	t.Parallel()
+
+	outPoint := wire.OutPoint{Index: 3}
```
yyforyongyu: This fixture cannot pass the public validator: the outpoint has a zero hash and index 3, while parent has one output and a different TxHash. As a result the test bypasses the invariant it claims was already checked. Please construct the outpoint from parent.TxHash() with a valid index and exercise preservation through FundPsbt.

wallet/psbt_clone_test.go
```diff
@@ -0,0 +1,142 @@
...
+			},
+			Bip32Path: []uint32{10},
+		}}
+
+	clone := clonePacket(packet)
+	require.Equal(t, packet, clone)
+
+	// A snapshot to compare the original against once the clone has been
+	// scribbled over.
+	before := clonePacket(packet)
```
yyforyongyu: This uses clonePacket itself to construct the independence oracle. If clonePacket shallow-copies a field, mutating clone can update packet and before through the same backing data, allowing require.Equal(before, packet) to pass. Please capture expected state independently, for example with explicit expected values or a non-mutating serialized snapshot.

wallet/psbt_packet_test.go
```diff
@@ -0,0 +1,560 @@
...
+// operation, whatever the wallet is about to do with the packet.
+func TestValidatePacketRejects(t *testing.T) {
+	t.Parallel()
+
+	tests := []struct {
+		name string
+
+		// mutate bends an otherwise valid packet out of shape. A nil
+		// mutate means the case supplies the packet itself.
+		mutate func(*psbt.Packet)
```
yyforyongyu: Repository test guidance requires table cases to remain data-only. These executable mutation closures hide setup and make case behavior harder to audit; please express the cases as declarative inputs or explicit subtests. The same pattern appears in the new FundPsbt tables.

yyforyongyu (changes requested):
My main comment is, it looks like several methods are introduced but not used?
- `validatePacket` has exactly one production caller `validateFundIntent`, always with `psbtOpFund`.
- `psbtOpSign` and `psbtOpFinalize` are referenced only by tests.
- Consequently, `validatePartialSigs`, `validateScriptSpendSigs`, and `validateSchnorrSig` are unreachable from the current production call path.

In addition we should split them into smaller commits so it's easier to review them.

wallet/psbt_packet.go
Abdulkbk: removed, plus the other unused ones.

wallet/psbt_packet.go
Abdulkbk: Done

wallet/psbt_packet.go
Abdulkbk: added `validatePacketPointers` which runs before sanitycheck or cloning

wallet/psbt_packet.go
Abdulkbk: now rejecting ANYONECANPAY when the base is SigHashDefault

wallet/psbt_funding.go
Abdulkbk: added checkDerivationFamily, rejecting the mismatch both ways before either representation is copied.

wallet/psbt_manager.go
Abdulkbk: updated

wallet/psbt_manager.go
```diff
@@ -708,6 +711,11 @@ func (w *Wallet) FundPsbt(ctx context.Context, intent *FundIntent) (
 // handleFundPsbt completes its accepted operation with the caller's packet.
 // The admitted caller waits for this result before reusing its inputs.
 func (w *Wallet) handleFundPsbt(r fundPsbtReq) {
+	// Everything from here on works on a clone. The caller's packet is
+	// never the thing being rewritten, so however this ends, and however
+	// far it got, the caller gets its own packet back exactly as it was.
+	packet := clonePacket(r.intent.Packet)
```
yyforyongyu, line 717: Cloning here is undermined by the next call: `createTxIntent(r.intent)` reads the original packet and shallow-copies each `wire.TxOut`, after which `authorTransaction` carries the same `PkScript` slices into the funded transaction. The returned packet therefore shares output-script backing memory with the caller packet. Please build the transaction intent from the clone and add a regression that mutating the returned script cannot mutate the caller packet.

wallet/psbt_packet.go
```diff
@@ -0,0 +1,439 @@
...
+
+		err := validateInputUtxos(pIn, outPoint, i)
+		if err != nil {
+			return err
+		}
+
+		// Reuse the signing path's derivation rules so that a packet
+		// admitted here cannot be rejected later for a reason this
+		// gate could have caught.
+		_, err = validateDerivation(pIn, i)
```
yyforyongyu: `validateDerivation` enforces the signer’s one-derivation policy, not structural PSBT validity. Multiple derivation records are valid metadata for multisig/coordinator packets, and Task 346 requires funding to preserve them. Funding should validate their structure, then reconcile the wallet-owned derivation as one member of the caller’s set while retaining the remaining records.

wallet/psbt_packet.go
```diff
@@ -0,0 +1,439 @@
...
+	return nil
+}
+
+// inputIsSigned reports whether an input carries any signature material, in
+// any of the forms a PSBT can hold it.
+func inputIsSigned(pIn *psbt.PInput) bool {
+	return len(pIn.PartialSigs) > 0 ||
+		len(pIn.TaprootKeySpendSig) > 0 ||
+		len(pIn.TaprootScriptSpendSig) > 0 ||
+		len(pIn.FinalScriptSig) > 0 ||
```
yyforyongyu: Final fields are defined by presence, not non-zero length: the canonical PSBT finalizer treats `FinalScriptSig != nil` or `FinalScriptWitness != nil` as finalized. A non-nil empty record passes this check and is then silently dropped during funding. Please test these two fields against `nil` and cover the empty-but-present cases.

wallet/psbt_packet.go
```diff
@@ -0,0 +1,439 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// validatePacketOutputs runs the per-output half of validatePacket.
+func validatePacketOutputs(packet *psbt.Packet) error {
+	for i := range packet.Outputs {
```
yyforyongyu, line 296: The output gate checks pointer presence and unknown fields but never validates the contents of known records. For example, a `Bip32Derivation` containing a one-byte public key passes here even though the canonical PSBT parser rejects it, so funding can return a malformed packet. Please validate known input/output record structure before dependency work.

wallet/psbt_manager_test.go
```diff
@@ -4953,3 +4925,530 @@ func TestFundPsbtRejectsCorruptInputAmounts(t *testing.T) {
...
+	// derivation and none of the caller's tags.
+	require.GreaterOrEqual(t, changeIndex, int32(0))
+	require.Len(t, funded.Outputs[changeIndex].Bip32Derivation, 1)
+	require.Nil(t, funded.Outputs[changeIndex].WitnessScript)
+}
+
+// TestFundPsbtRejectsMalformedPacket verifies that a structurally invalid
+// packet is refused before the wallet acts on it.
+//
+// The individual structural rules are covered against the validator itself.
```
yyforyongyu: The public malformed-packet test covers only duplicate outpoints, while the other admission rules are exercised only through the private validator. Please make these malformed cases a data-only `FundPsbt` table that also asserts no dependencies were called, and remove the overlapping helper-only cases instead of duplicating both layers.

wallet/psbt_packet_test.go
```diff
@@ -386,3 +386,68 @@ func TestValidatePacketUnclassifiedFields(t *testing.T) {
...
+	packet := testPacket(t)
+	packet.Inputs[0].NonWitnessUtxo = &wire.MsgTx{
+		Version: 2,
+		TxIn:    []*wire.TxIn{{}},
+		TxOut: []*wire.TxOut{{
+			Value: 1, PkScript: []byte{0x51},
+		}},
+	}
+
+	// Act.
```
yyforyongyu: nit: these AAA comments are a bit shallow, i think we can remove them if the tests are simple to follow.

docs/developer/psbt_workflows.md
```diff
@@ -62,20 +62,22 @@ flowchart LR
     ```
     
     The `fundedPacket` now contains the necessary inputs (fully decorated) and a change output.
+    `FundPsbt` does not modify the packet it was given, so the remaining steps
+    work on `fundedPacket`.
```
yyforyongyu, line 66: 👍

yyforyongyu (changes requested):
Looking good, only a few comments otherwise should be good to go!

yyforyongyu (approved):
There are still a few issues, like the size of the PR, but overall this is a pure improvement, so i think we can defer fixing these issues in a future task or PR as these are not critical but def nice to have.

### Correctness issues

  1. Incomplete structural validation — wallet/psbt_packet.go:316-357,492-497

     Valid keys are checked, but other known record contents are not. Invalid Taproot leaf hashes/control blocks/       
     internal keys and non-nil empty TaprootKeySpendSig can pass admission. Funding may return an invalid PSBT or       
     silently discard the empty signature.

     Suggested fix: After pointer guards, serialize and reparse the cloned packet using the canonical psbt
     implementation. Keep only Task 346 policy checks—duplicate outpoints, unknown fields, signatures and SIGHASH_SINGLE
     —in wallet code.

  2. Metadata is not reconciled against the actual spend — wallet/psbt_funding.go:122-138,203-264

     Caller-provided witness/Taproot fields are copied without confirming they belong to the wallet UTXO’s script type. 
     A different public key can also reuse the wallet’s fingerprint/path and be treated as an unrelated cosigner.       

     Suggested fix: Classify metadata using the authoritative wallet output and spend script. Match wallet derivations  
     by fingerprint, path, key and script participation; preserve genuine cosigner records but reject contradictions and
     incompatible fields.

  3. Funded multisig metadata cannot currently be signed by this wallet — wallet/psbt_manager.go:1154,1366

     Funding correctly preserves multiple derivations, but SignPsbt still rejects more than one derivation. The         
     documented FundPsbt → SignPsbt workflow therefore fails for a metadata-rich multisig packet.

     Suggested fix: Do not reject multiple derivations in funding. Update signing in its owning follow-up to select the 
     wallet-owned derivation while retaining the cosigner records.

  4. A rejected metadata contradiction can consume a change address — wallet/psbt_manager.go:737-825

     Wallet/caller reconciliation occurs after authoring, and authoring has already requested a change script.

     Suggested fix: Either pre-decorate and reconcile manual inputs before authoring, or address this through Task 356’s
     provisional change-allocation design.

  ### Contract and test issues

  5. Validation occurs after a syncer dependency call — wallet/psbt_manager.go:682-691

     validateSynced() queries the chain syncer before malformed-packet validation.

     Suggested fix: Call validateFundIntent before validateSynced.

  6. The public malformed-packet test remains incomplete and duplicated — wallet/psbt_manager_test.go:5322-5421         

     The table now covers five cases, but case construction is hidden behind a string/switch, many rules remain only in 
     the large private-validator suite, and the test explicitly expects the syncer call despite claiming no dependencies
     are called.

     Suggested fix: Use declarative public FundPsbt cases covering every admission rule and assert no dependencies are  
     touched. Retain private tests only for cases impossible to express safely at the public boundary.

  7. Private-helper testing is excessive

     psbt_clone_test.go, psbt_packet_test.go, and psbt_funding_test.go add about 1,600 lines before the public FundPsbt 
     tests, with overlapping coverage.

     Suggested fix: Make public contract tests authoritative and keep narrow helper tests only for deep-copy aliasing,  
     pointer-identity change tracking and similarly isolated mechanics.

  ### Maintainability and hygiene

  8. Validation duplicates the canonical PSBT parser

     This duplication caused correctness issue 1 and accounts for much of the PR’s size.

     Suggested fix: Reuse canonical serialization/parsing and keep wallet-specific policy validation local.

  9. Funding reconciliation overlaps existing merge primitives

     mergePsbtInputs cannot replace it wholesale because combining has equal authority while funding makes wallet facts 
     authoritative.

     Suggested fix: Share low-level equality/deduplication functions, but retain separate policy-specific orchestration.
     Do not turn mergePsbtInputs into a complicated mode-driven function.

  10. File organization is broader than necessary

  Three new PSBT implementation files spread one funding workflow across roughly 1,150 production lines.

  Suggested fix: Keep this in wallet; no new package is justified. Consolidate task-specific cloning and reconciliation 
  into psbt_funding.go, retaining a separate generic packet validator only if Tasks 347/407 will immediately reuse it.

Abdulkbk:
> There are still a few issues, like the size of the PR, but overall this is a pure improvement, so i think we can defer fixing these issues in a future task or PR as these are not critical but def nice to have.
> 
> ### Correctness issues
> 1. Incomplete structural validation — wallet/psbt_packet.go:316-357,492-497
>    Valid keys are checked, but other known record contents are not. Invalid Taproot leaf hashes/control blocks/
>    internal keys and non-nil empty TaprootKeySpendSig can pass admission. Funding may return an invalid PSBT or
>    silently discard the empty signature.
>    Suggested fix: After pointer guards, serialize and reparse the cloned packet using the canonical psbt
>    implementation. Keep only Task 346 policy checks—duplicate outpoints, unknown fields, signatures and SIGHASH_SINGLE
>    —in wallet code.
> 2. Metadata is not reconciled against the actual spend — wallet/psbt_funding.go:122-138,203-264
>    Caller-provided witness/Taproot fields are copied without confirming they belong to the wallet UTXO’s script type.
>    A different public key can also reuse the wallet’s fingerprint/path and be treated as an unrelated cosigner.
>    Suggested fix: Classify metadata using the authoritative wallet output and spend script. Match wallet derivations
>    by fingerprint, path, key and script participation; preserve genuine cosigner records but reject contradictions and
>    incompatible fields.
> 3. Funded multisig metadata cannot currently be signed by this wallet — wallet/psbt_manager.go:1154,1366
>    Funding correctly preserves multiple derivations, but SignPsbt still rejects more than one derivation. The
>    documented FundPsbt → SignPsbt workflow therefore fails for a metadata-rich multisig packet.
>    Suggested fix: Do not reject multiple derivations in funding. Update signing in its owning follow-up to select the
>    wallet-owned derivation while retaining the cosigner records.
> 4. A rejected metadata contradiction can consume a change address — wallet/psbt_manager.go:737-825
>    Wallet/caller reconciliation occurs after authoring, and authoring has already requested a change script.
>    Suggested fix: Either pre-decorate and reconcile manual inputs before authoring, or address this through Task 356’s
>    provisional change-allocation design.
> 
> ### Contract and test issues
> 5. Validation occurs after a syncer dependency call — wallet/psbt_manager.go:682-691
>    validateSynced() queries the chain syncer before malformed-packet validation.
>    Suggested fix: Call validateFundIntent before validateSynced.
> 6. The public malformed-packet test remains incomplete and duplicated — wallet/psbt_manager_test.go:5322-5421
>    The table now covers five cases, but case construction is hidden behind a string/switch, many rules remain only in
>    the large private-validator suite, and the test explicitly expects the syncer call despite claiming no dependencies
>    are called.
>    Suggested fix: Use declarative public FundPsbt cases covering every admission rule and assert no dependencies are
>    touched. Retain private tests only for cases impossible to express safely at the public boundary.
> 7. Private-helper testing is excessive
>    psbt_clone_test.go, psbt_packet_test.go, and psbt_funding_test.go add about 1,600 lines before the public FundPsbt
>    tests, with overlapping coverage.
>    Suggested fix: Make public contract tests authoritative and keep narrow helper tests only for deep-copy aliasing,
>    pointer-identity change tracking and similarly isolated mechanics.
> 
> ### Maintainability and hygiene
> 8. Validation duplicates the canonical PSBT parser
>    This duplication caused correctness issue 1 and accounts for much of the PR’s size.
>    Suggested fix: Reuse canonical serialization/parsing and keep wallet-specific policy validation local.
> 9. Funding reconciliation overlaps existing merge primitives
>    mergePsbtInputs cannot replace it wholesale because combining has equal authority while funding makes wallet facts
>    authoritative.
>    Suggested fix: Share low-level equality/deduplication functions, but retain separate policy-specific orchestration.
>    Do not turn mergePsbtInputs into a complicated mode-driven function.
> 10. File organization is broader than necessary
> 
> Three new PSBT implementation files spread one funding workflow across roughly 1,150 production lines.
> 
> Suggested fix: Keep this in wallet; no new package is justified. Consolidate task-specific cloning and reconciliation into psbt_funding.go, retaining a separate generic packet validator only if Tasks 347/407 will immediately reuse it.

I will create the follow-up. Will try to make it as contained as possible.

PR #1360, opened 2026-09-02
wtxmgr: keep confirmation leases past wall-clock expiry

bhandras:
## Problem

#1351 made output leases survive a shallow reorg after their spend confirms.
Before the first confirmation, the lease still used its wall-clock deadline.

If confirmation took longer than that deadline, expired-lock cleanup deleted
the row. A later confirmation could no longer start confirmation tracking. This
breaks the reorg protection needed by lightningnetwork/lnd#11125 during a fee
spike or other slow-confirmation period.

## Fix

Treat a non-zero `WithReleaseAfterSpend` depth as a separate lease lifetime:

- `depth == 0`: release at the wall-clock expiration
- `depth > 0`: ignore wall-clock expiration and release at the requested spend
  depth or by an explicit owner `UnlockOutput`

The database encoding is unchanged.

## Safety invariant

Elapsed wall-clock time cannot remove a confirmation-controlled lease. Only
spend maturity or an authenticated owner release can remove it.

An abandoned confirmation-controlled lease now requires explicit owner
release. This trades bounded availability for the requested reorg guarantee.

## Compatibility

- Existing time-only leases keep their current behavior.
- `WithReleaseAfterSpend(0)` remains time-only.
- Surviving non-zero-depth rows written by #1351 gain the corrected behavior
  immediately after upgrade.
- Older software treats these rows as time-only leases. Do not downgrade while
  a confirmation-controlled lease is active.

## Tests

- expiry before the first spend confirmation
- expired cleanup preserves a confirmation-controlled lease
- another owner remains blocked and cannot release the lease
- the correct owner can release it explicitly
- a later time-only lease does not inherit the old release depth
- late confirmation records its spend height
- rollback, expiry while disconnected, reconfirmation, and final maturity

Local checks:

- `go test ./...` and `go test -race ./...` in `wtxmgr`
- `go test ./wallet` and `go test -race ./wallet`
- `go vet` for both changed modules
- `make fmt-check`
- `make tidy-module-check`

### DIFF
--- a/wtxmgr/db.go
+++ b/wtxmgr/db.go
@@ -1341,22 +1341,21 @@ func fetchLockedOutput(ns walletdb.ReadBucket,
 }
 
 // isLockedOutput determines whether an output is locked. If it is, its assigned
-// ID is returned, along with its absolute expiration time. If the output lock
-// exists, but its expiration has been met, then the output is considered
-// unlocked.
+// ID is returned, along with its absolute expiration time. Confirmation-
+// controlled locks ignore that time and remain active until explicit release or
+// spend maturity. Other locks become inactive when their expiration is met.
 func isLockedOutput(ns walletdb.ReadBucket, op wire.OutPoint,
 	timeNow time.Time) (LockID, time.Time, bool) {
 
-	lockID, expiry, releaseAfterSpendConfs, spendHeight, exists :=
+	lockID, expiry, releaseAfterSpendConfs, _, exists :=
 		fetchLockedOutput(ns, op)
 	if !exists {
 		return LockID{}, time.Time{}, false
 	}
 
-	// Once a retained lease has observed a confirmed spend, its wall clock
-	// expiry no longer applies. A negative spend height means that the observed
-	// spend was disconnected and is waiting to confirm again.
-	if releaseAfterSpendConfs > 0 && spendHeight != 0 {
+	// A confirmation-controlled lease uses explicit release or spend maturity
+	// instead of wall-clock expiry for its complete lifetime.
+	if releaseAfterSpendConfs > 0 {
 		return lockID, expiry, true
 	}
 
--- a/wtxmgr/tx_test.go
+++ b/wtxmgr/tx_test.go
@@ -2752,17 +2752,17 @@ func TestOutputLocks(t *testing.T) {
 			},
 		},
 		{
-			// Asserts that reusing an ID after its retained lease
-			// expires does not carry the old release policy into the
-			// new lease.
-			name: "reuse lock ID after expiry",
+			// Asserts that a confirmation-controlled lease ignores its
+			// wall-clock expiry but still permits explicit owner release.
+			name: "confirmation lease ignores wall clock",
 			run: func(t *testing.T, s *Store, ns walletdb.ReadWriteBucket) {
 				t.Helper()
 
-				lockID := LockID{1}
+				lockID1 := LockID{1}
+				lockID2 := LockID{2}
 
 				expiry, err := s.LockOutput(
-					ns, lockID, confirmedOutPoint, 10*time.Minute,
+					ns, lockID1, confirmedOutPoint, 10*time.Minute,
 					WithReleaseAfterSpend(3),
 				)
 				if err != nil {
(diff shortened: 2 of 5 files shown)


bhandras:
Follow-up to #1351 for the pre-confirmation expiry edge case. CI is green.

@GustavoStingelin @yyforyongyu, could you please take a look?

Lrifton92 (commented):
Nice, focused follow-up. I traced the `releaseAfterSpendConfs > 0 && spendHeight != 0` -> `releaseAfterSpendConfs > 0` change through all five read sites (`isLockedOutput` in db.go plus `insertMinedTx`, `UnlockOutput`, `DeleteExpiredLockedOutputs`, `ListLockedOutputs`) and it's consistent. Dropping `retainedLockActive` in `insertMinedTx` is the real fix — a spend confirming after the wall-clock deadline can now start tracking instead of being cleaned up first. The reorg path is safe: a disconnected spend keeps `spendHeight == -1`, so it stays active under the new predicate exactly as before, and maturity/reset still key on `spendHeight` in `DeleteMaturedLockedOutputs`/`resetLockedOutputSpendHeights`. Collapsing the hybrid state also makes the lifetime much easier to reason about.

One thing to confirm out loud: a depth>0 lease whose spend is never confirmed (aborted funding, RBF-away, crash before `UnlockOutput`) is now retained indefinitely rather than reclaimed at wall-clock — the PR body acknowledges this, so mainly worth a note that lnd #11125 always has a release path so these can't accumulate as permanently-locked UTXOs. Test coverage looks solid: the renamed "confirmation lease ignores wall clock" case and the new pre-spend `DeleteExpiredLockedOutputs` assertion in `TestOutputLockReleaseAfterSpendAcrossReorg` both pin the new behavior. LGTM modulo that confirmation.

wtxmgr/db.go
```diff
@@ -1341,22 +1341,21 @@ func fetchLockedOutput(ns walletdb.ReadBucket,
...
 		return LockID{}, time.Time{}, false
 	}
 
-	// Once a retained lease has observed a confirmed spend, its wall clock
-	// expiry no longer applies. A negative spend height means that the observed
-	// spend was disconnected and is waiting to confirm again.
-	if releaseAfterSpendConfs > 0 && spendHeight != 0 {
+	// A confirmation-controlled lease uses explicit release or spend maturity
+	// instead of wall-clock expiry for its complete lifetime.
+	if releaseAfterSpendConfs > 0 {
```
yyforyongyu, line 1358: The active-lease policy is independently re-encoded in `insertMinedTx`, `UnlockOutput`, `DeleteExpiredLockedOutputs`, and `ListLockedOutputs`, which allows those paths to diverge from `isLockedOutput`; please make `isLockedOutput` the single owner of active-state semantics, centralize the confirmation-controlled mode predicate needed by spend tracking, and route these consumers through those canonical functions.

wtxmgr/tx_test.go
```diff
@@ -3066,6 +3068,14 @@ func TestOutputLockReleaseAfterSpendAcrossReorg(t *testing.T) {
 			return err
 		}
 
+		// The first confirmation may arrive after the wall-clock deadline.
+		// Confirmation-controlled leases must still record that spend.
+		store.clock.(*clock.TestClock).SetTime(expiry.Add(time.Second))
```
ziggie1984, line 3073: 🫡

ziggie1984 (approved):
LGTM

yyforyongyu (approved):
The agent found one improvement we could have, but given that this is fixed in the upcoming new wallet, think we can just ignore.

wtxmgr/db.go
bhandras: I agree this would make the old store easier to evolve, but I don't think it belongs in this bug fix. These paths do not consume the same state: insertMinedTx needs the raw depth and updates spend height, UnlockOutput must authorize the owner even after the output is spent, and the cleanup/list paths already iterate lock records, so routing them through isLockedOutput would add record reads or require a broader lock-state abstraction. This PR keeps the rule simple and consistent: a non-zero releaseAfterSpendConfs makes the lease time-independent, with tests covering each consumer. Given that this code is replaced by the new wallet, I'm leaving the wider refactor out.

GustavoStingelin (approved):
blindly approved

PR #1361, opened 2026-09-03
wallet: publish a transaction label limit

Abdulkbk:
Follow-up to #1347, which covered `LabelTx` with integration tests and found
that the API has no limit of its own.

Every store refuses a label past 500 and each refuses it differently, so a
caller wanting to handle the case had to match a database driver's error:

| backend | error at 501 bytes |
|---|---|
| kvdb | `wtxmgr.ErrLabelTooLong` |
| SQLite | `*sqlite.Error: CHECK constraint failed: length(tx_label) <= 500 (275)` |
| PostgreSQL | `*pgconn.PgError: value too long for type character varying(500) (SQLSTATE 22001)` |

The stores also disagree on what they count. `wtxmgr` limits bytes while both
SQL schemas limit characters, so a 200-character label of three-byte runes was
refused by kvdb and stored as 600 bytes by the other two. There was no single
limit to publish.

This adds `wallet.MaxTxLabelLength` and `wallet.ErrLabelTooLong`, checked in
`LabelTx` above the stores. The `txwriter reject oversize label` integration
case becomes an `ErrorIs` like every other rejection in that suite, and the
boundary case takes the limit from the wallet, which retires the `itest`
package's last use of `wtxmgr`.

## The limit counts bytes

A byte budget of 500 is the stricter of the two readings: any label within it is
also within a 500-character limit, so no store constraint can fire for a label
the wallet already accepted. Checking characters instead would let the wallet
accept labels kvdb then refuses, which is the problem being fixed. Those store
constraints stay where they are, now as defence in depth rather than as the
contract.

**This tightens the SQL backends.** A label over 500 bytes but under 500
characters is accepted today and will be refused. Nothing in the codebase writes
labels that large, and the deprecated `LabelTransaction` has always documented
that the call fails when the label is too long.

## Testing

- `go test ./wallet/ -run TestLabelTx` — added cases at the limit, one byte past
  it, and a multi-byte label a character limit would have let through.
- `make itest icase='txwriter'` on `db=sqlite`, `db=kvdb` and `db=postgres`, and
  on `chain=bitcoind`.
- Full itest suite shuffled on sqlite (`-shuffleseed=42`) and kvdb
  (`-shuffleseed=7`), no failures or skips.
- `golangci-lint` clean on `wallet` and `itest`.

### DIFF
--- a/itest/tx_writer_test.go
+++ b/itest/tx_writer_test.go
@@ -10,7 +10,6 @@ import (
 	"github.com/btcsuite/btcwallet/bwtest"
 	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet"
-	"github.com/btcsuite/btcwallet/wtxmgr"
 	"github.com/stretchr/testify/require"
 )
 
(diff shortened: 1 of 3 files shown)


yyforyongyu (approved):
Straightforward changes, LGTM!

PR #1362, opened 2026-09-03
wallet+kvdb: clear a transaction label with an empty one

Abdulkbk:
Follow-up to #1347, which covered `LabelTx` with integration tests and found
that an empty label means two different things depending on the backend.

**Stacked on #1361.** Its two commits are included here until it merges; the
three commits from `wtxmgr: delete a transaction label` onwards are this PR's.

## The gap

`db.UpdateTxParams.Label` documents the empty string as valid and as clearing
any prior label, and both SQL schemas default `tx_label` to `''`. SQLite and
PostgreSQL implement exactly that. The legacy kvdb store refused the same call
with `wtxmgr.ErrEmptyLabel`, because it forwarded every label to `PutTxLabel`,
which rejects an empty one — even when the transaction had no label to clear.

Measured on all three backends before the change:

| probe | sqlite / postgres | kvdb |
|---|---|---|
| `""` over an existing label | cleared | `ErrEmptyLabel`, label kept |
| `""` on an unlabeled transaction | no-op | `ErrEmptyLabel` |
| after a reload | still cleared | original label still there |

So clearing a label succeeded or failed depending on which database the wallet
was opened with, and no caller could clear one portably at all. `LabelTx` is the
only public way to remove a label, so on kvdb there was none.

## The change

kvdb routes an empty label to a new `wtxmgr.DeleteTxLabel` instead of
`PutTxLabel`. That store records a transaction with no label as a missing entry
rather than a stored empty value — `FetchTxLabel` reports a zero-length stored
label as `ErrEmptyLabel` — so clearing has to be a delete, and the labels bucket
is private to `wtxmgr`, which is why the method belongs there. `PutTxLabel` is
untouched, so the deprecated `LabelTransaction` keeps refusing an empty label
exactly as before.

`wallet/tx_writer_test.go`'s `TestLabelTxEmptyLabel` asserted the kvdb refusal as
though it were the wallet's contract, by mocking the store into returning it. It
now asserts what the API promises.

## Testing

- `go test ./wtxmgr/... -run TxLabel` — added `TestDeleteTxLabel`, covering a
  delete before any label exists, after one is written, and the read-back that
  reports no label rather than an error.
- `go test ./wallet/internal/db/kvdb/ -run TestUpdateTx` — clearing an existing
  label and clearing when there is none.
- New integration case `txwriter clear label`: clears a label, clears again when
  there is none, and proves the removal survives a reload. It runs unchanged on
  kvdb, SQLite and PostgreSQL — it could not be written before this change.
- Full itest suite shuffled on sqlite (`-shuffleseed=42`) and kvdb
  (`-shuffleseed=7`), no failures or skips; `golangci-lint` clean on `wallet`,
  `bwtest` and `itest`.

### DIFF
--- a/itest/tx_writer_test.go
+++ b/itest/tx_writer_test.go
@@ -93,6 +93,61 @@ func testLabelTxReplace(h *bwtest.HarnessTest) {
 	)
 }
 
+// testLabelTxClear verifies that an empty label removes the label a
+// transaction already had, that removing a label a transaction never had is
+// not an error, and that a cleared label stays cleared across a reload.
+//
+// The empty label is the only way to remove one through this API, so a wallet
+// that refused it would leave a caller no way to undo a label at all.
+func testLabelTxClear(h *bwtest.HarnessTest) {
+	const label = "rent for march"
+
+	w, funding := h.NewWallet(bwtest.WalletFixture{
+		AddrType: txWriterFundingType,
+		Amounts:  []btcutil.Amount{oneBTC},
+	})
+
+	txHash := funding.WalletOutpoints[0].Hash
+
+	err := w.LabelTx(h.Context(), txHash, label)
+	require.NoError(h, err, "failed to label transaction")
+
+	err = w.LabelTx(h.Context(), txHash, "")
+
+	require.NoError(h, err, "clearing a label was refused")
+
+	detail, err := w.GetTx(h.Context(), txHash)
+	require.NoError(h, err, "failed to read the cleared transaction")
+	require.Empty(h, detail.Label, "point read still reports a label")
+
+	details, err := w.ListTxns(h.Context(), unminedHeight, 0)
+	require.NoError(h, err, "failed to list transactions")
+
+	listed := make(map[chainhash.Hash]string, len(details))
+	for _, detail := range details {
+		listed[detail.Hash] = detail.Label
+	}
+
+	// Assert the transaction is listed before reading its label, so an
+	// absent entry fails here rather than reading as an empty label.
+	require.Contains(h, listed, txHash, "list read lost the transaction")
+	require.Empty(h, listed[txHash], "list read still reports a label")
+
+	// A transaction with no label is what the caller asked for and already
+	// has, so asking again is not an error.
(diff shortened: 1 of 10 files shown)


itest/tx_writer_test.go
```diff
@@ -94,21 +93,70 @@ func testLabelTxReplace(h *bwtest.HarnessTest) {
...
+
+	details, err := w.ListTxns(h.Context(), unminedHeight, 0)
+	require.NoError(h, err, "failed to list transactions")
+
+	listed := make(map[chainhash.Hash]string, len(details))
+	for _, detail := range details {
+		listed[detail.Hash] = detail.Label
+	}
+
+	require.Empty(h, listed[txHash], "list read still reports a label")
```
yyforyongyu, line 134: This uses `require.Empty` on a map lookup, which also passes when `ListTxns` omits the transaction - we can assert `require.Contains(h, listed, txHash)` before checking the empty label

yyforyongyu (commented):
Pending a rebase plus a comment, otherwise good to go!

yyforyongyu (changes requested):

yyforyongyu:
Pending another rebase

wtxmgr/tx_test.go
```diff
@@ -2433,6 +2433,86 @@ func TestTxLabel(t *testing.T) {
...
+		return walletdb.Update(db, func(tx walletdb.ReadWriteTx) error {
+			return store.DeleteTxLabel(getBucket(tx), *txid)
+		})
+	}
+
+	// Deleting before any label has been written must succeed: the caller
+	// asked for a transaction with no label and already has one.
+	err = tryDeleteLabel()
+	if err != nil {
+		t.Fatalf("expected: no error, got: %v", err)
```
yyforyongyu: so we prefer using `require` lib, like `require.NoError` here, can we fix that?

yyforyongyu (changes requested):
cool one last thing - we wanna use `require` here

yyforyongyu (approved):
LGTM!

PR #1363, opened 2026-09-04
wallet: report exact synchronization status

GustavoStingelin:
## Change Description

Implements [Task 455](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/455-exact-wallet-sync-status.md).

`Wallet.Info().Synced` is true only when live processing is ready and the
committed wallet tip exactly matches the observed chain tip by height and
hash. A wallet one block behind reports `Synced=false` without changing the
existing small-gap operation policy.

- Keep `Info.SyncedTo` equal to the committed wallet tip.
- Remove unsupported `Info.IsRecoveryMode` and `Info.RecoveryProgress`,
  together with their unused helpers.
- Retain private `syncStateRescanning` and existing scan/spend admission.
  Targeted rescans alone do not make public `Info.Synced` false, advance
  `SyncedTo`, or change another wallet's readiness.
- Read `GetBestBlock` inside the admitted `Info` handler and preserve source
  errors. Keep the existing `chainSyncer` interface unchanged, with no
  duplicate scan-admission flag or status wrapper.
- Reuse existing controller/syncer test files and fixtures, including the
  real-store two-wallet regression and existing mock-backed batch checks.

### Why keep the direct tip read?

Targeted rescans currently occupy the same sync loop that processes live
notifications. A loop-owned tip cache can therefore remain equal to
`SyncedTo` after the source has advanced, incorrectly reporting synchronized.

The direct query costs one source read per `Info` call and makes source
availability part of that call's result. When the source advances during a
rescan, `Info` reports unsynced while `SyncedTo` stays at the committed tip.
Independent tip observation or parallel live processing needs separate work
before introducing a cache.

The three commits separate recovery-field removal, exact-tip reporting, and
targeted-rescan status, with each change's related tests alongside it.

## Steps to Test

Run the wallet tests with race detection and shuffled order:

```bash
go test -race -shuffle=on -count=1 ./wallet/...
make fmt
make lint-check
```

Run Task 433's existing controller-info scenario unchanged on all three
database backends:

```bash
make itest chain=btcd db=kvdb icase="controller info"
make itest chain=btcd db=sqlite icase="controller info"
make itest chain=btcd db=postgres icase="controller info"
```

### Testing

- [ ] The PR passes all CI checks.
- [x] Tests cover exact match, height/hash mismatch, source failure, and
  delivery-not-ready cases.
- [x] Tests distinguish small-gap operation admission from exact public status.
- [x] Targeted-scan tests cover unchanged/advancing source tips, two-wallet
  independence, committed-tip preservation, and state restoration on failure.
- [x] The unchanged controller-info scenario passes locally on kvdb, SQLite,
  and PostgreSQL.

- [x] The change is scoped to one roadmap task.
- [x] Public synchronization behavior and the source-read failure are documented.
- [x] No new logging statements are introduced.

### DIFF
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -597,7 +597,7 @@ func (w *Wallet) handleInfo(r infoReq) {
 	}
 
 	state := w.sync.syncState()
-	liveReady := state == syncStateSynced
+	liveReady := state == syncStateSynced || state == syncStateRescanning
 
 	synced := liveReady &&
 		syncedTo.Height == bestHeight && syncedTo.Hash.IsEqual(bestHash)
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -71,7 +71,8 @@ type UnlockRequest struct {
 }
 
 // Info provides a comprehensive snapshot of the wallet's static configuration
-// and dynamic synchronization state.
+// and dynamic synchronization state. Producing the snapshot reads the observed
+// tip from the chain source and fails if that tip is unavailable.
 type Info struct {
 	// BirthdayBlock is the block from which the wallet started scanning.
 	BirthdayBlock waddrmgr.BlockStamp
@@ -87,7 +88,8 @@ type Info struct {
 	// Locked indicates if the wallet is currently locked.
 	Locked bool
 
-	// Synced indicates if the wallet is synced to the chain tip.
+	// Synced is true only when live delivery is ready and SyncedTo exactly
+	// matches the observed chain tip by height and hash.
 	Synced bool
 
 	// SyncedTo is the block to which the wallet is currently synced.
@@ -147,7 +149,8 @@ type Controller interface {
 	ChangePassphrase(ctx context.Context, req ChangePassphraseRequest) error
 
 	// Info returns a comprehensive snapshot of the wallet's static
-	// configuration and dynamic synchronization state.
+	// configuration and dynamic synchronization state. It returns an error
+	// if the chain source's observed tip cannot be read.
 	Info(ctx context.Context) (*Info, error)
 
 	// Start starts the background processes necessary to manage the wallet.
@@ -528,7 +531,8 @@ type rescanReq struct {
 }
 
 // Info returns a comprehensive snapshot of the wallet's static configuration
-// and dynamic synchronization state.
+// and dynamic synchronization state. It returns an error if the chain source's
+// observed tip cannot be read.
 //
 // This is part of the Controller interface.
 func (w *Wallet) Info(ctx context.Context) (*Info, error) {
@@ -585,12 +589,25 @@ func (w *Wallet) handleInfo(r infoReq) {
 		return
 	}
 
+	bestHash, bestHeight, err := w.cfg.Chain.GetBestBlock()
+	if err != nil {
+		r.respChan <- infoResp{err: fmt.Errorf("get chain tip: %w", err)}
+
+		return
+	}
+
+	state := w.sync.syncState()
+	liveReady := state == syncStateSynced
+
+	synced := liveReady &&
+		syncedTo.Height == bestHeight && syncedTo.Hash.IsEqual(bestHash)
+
 	info := &Info{
 		BirthdayBlock: w.birthdayBlock,
 		Backend:       w.cfg.Chain.BackEnd(),
 		ChainParams:   &chainParams,
 		Locked:        !w.state.isUnlocked(),
-		Synced:        w.state.isSynced(),
+		Synced:        synced,
 		SyncedTo:      syncedTo,
 	}
 
--- a/wallet/controller.go
+++ b/wallet/controller.go
@@ -92,13 +92,6 @@ type Info struct {
 
 	// SyncedTo is the block to which the wallet is currently synced.
 	SyncedTo waddrmgr.BlockStamp
-
-	// IsRecoveryMode indicates if the wallet is currently in recovery
-	// mode.
-	IsRecoveryMode bool
-
-	// RecoveryProgress is the progress of the recovery (0.0 - 1.0).
-	RecoveryProgress float64
 }
 
 // ChangePassphraseRequest contains the parameters for changing wallet
@@ -593,14 +586,12 @@ func (w *Wallet) handleInfo(r infoReq) {
 	}
 
 	info := &Info{
-		BirthdayBlock:    w.birthdayBlock,
-		Backend:          w.cfg.Chain.BackEnd(),
-		ChainParams:      &chainParams,
-		Locked:           !w.state.isUnlocked(),
-		Synced:           w.state.isSynced(),
-		SyncedTo:         syncedTo,
-		IsRecoveryMode:   w.state.isRecoveryMode(),
-		RecoveryProgress: 0,
+		BirthdayBlock: w.birthdayBlock,
+		Backend:       w.cfg.Chain.BackEnd(),
+		ChainParams:   &chainParams,
+		Locked:        !w.state.isUnlocked(),
+		Synced:        w.state.isSynced(),
+		SyncedTo:      syncedTo,
 	}
 
 	r.respChan <- infoResp{info: info}
--- a/wallet/syncer.go
+++ b/wallet/syncer.go
@@ -108,9 +108,8 @@ const (
 	// chain tip.
 	syncStateSynced
 
-	// syncStateRescanning indicates the wallet is running a historical
-	// scan for specific user-provided targets, such as accounts or
-	// addresses, without rewinding the global synchronization state.
+	// syncStateRescanning indicates a historical scan of user-provided
+	// targets without rewinding the Wallet's live synchronization tip.
 	syncStateRescanning
 )
 
--- a/wallet/syncer.go
+++ b/wallet/syncer.go
@@ -290,13 +290,6 @@ func (s *syncer) syncState() syncState {
 	return syncState(s.state.Load())
 }
 
-// isRecoveryMode returns true if the wallet is currently syncing or
-// rescanning.
-func (s *syncer) isRecoveryMode() bool {
-	status := s.syncState()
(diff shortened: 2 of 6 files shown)


wallet/syncer_test.go
```diff
@@ -35,6 +35,20 @@ import (
 // that need a recovery horizon beyond the addresses already on record.
 const testScanRecoveryWindow = 20
 
+type scanBatchRecorder struct {
+	db.Store
+
+	batches []db.ScanBatchParams
+}
+
+func (s *scanBatchRecorder) ApplyScanBatch(ctx context.Context,
```
yyforyongyu, line 44: these are missing comments? also i think we already have `walletmock.Store`, why defining this struct?

wallet/syncer.go
```diff
@@ -1827,9 +1827,9 @@ func (s *syncer) extractAddrEntries(txOuts []*wire.TxOut) []AddrEntry {
 func (s *syncer) handleScanReq(ctx context.Context,
 	req *scanReq) error {
 
-	// If the wallet is already syncing or rescanning, we can't accept a
-	// full resync request. This prevents conflicting rescan operations.
-	if s.isRecoveryMode() {
+	// A full synchronization pass owns the Wallet's live chain position, so
```
yyforyongyu, line 1830: btw forgot to mention in past PRs - but the commit msg body should be wrapped.

wallet/syncer.go
```diff
@@ -107,11 +107,6 @@ const (
 	// syncStateSynced indicates the wallet is running and synced to the
 	// chain tip.
 	syncStateSynced
-
-	// syncStateRescanning indicates the wallet is running a historical
-	// scan for specific user-provided targets, such as accounts or
-	// addresses, without rewinding the global synchronization state.
-	syncStateRescanning
```
yyforyongyu, line 114: this belongs to previous commit?

wallet/syncer.go
```diff
@@ -232,6 +234,10 @@ type chainSyncer interface {
 
 	// syncState returns the current synchronization state.
 	syncState() syncState
+
+	// liveStatus returns one snapshot of live delivery readiness and the
+	// observed chain tip.
+	liveStatus() (liveSyncStatus, error)
```
yyforyongyu, line 240: i don't think we need this? why expanding the chainSyncer interface when this method is used once and only provides an additional `GetBestBlock`? which can easily be done in `Info`

wallet/syncer.go
```diff
@@ -262,6 +262,10 @@ type syncer struct {
 	// state tracks the chain synchronization status.
 	state atomic.Uint32
 
+	// scanRequestPending tracks whether a scan request was accepted and has
+	// not completed yet.
+	scanRequestPending atomic.Bool
```
yyforyongyu, line 267: I think this breaks the chainSyncer abstraction

yyforyongyu: i think this should stay behind `chainSyncer`. `scanRequestPending` duplicates the scan state we already have. keep `syncStateRescanning`, track the block already observed by the sync loop, and return it from the existing method:

```go
type chainSyncer interface {
	run(ctx context.Context) error
	requestScan(ctx context.Context, req *scanReq) error
	syncState() (syncState, waddrmgr.BlockStamp)
}

type syncer struct {
	// ...

	state        atomic.Uint32
	currentBlock atomic.Pointer[waddrmgr.BlockStamp]

	// ...
}

func (s *syncer) syncState() (syncState, waddrmgr.BlockStamp) {
	state := syncState(s.state.Load())
	currentBlock := s.currentBlock.Load()
	if currentBlock == nil {
		return state, waddrmgr.BlockStamp{}
	}

	return state, *currentBlock
}
```

update `currentBlock` in the existing startup/catch-up and block connect/disconnect paths whenever the syncer learns the chain tip:

```go
currentBlock := &waddrmgr.BlockStamp{
	Hash:      block.Hash,
	Height:    block.Height,
	Timestamp: block.Time,
}
s.currentBlock.Store(currentBlock)
```

then `Info` only uses the existing `chainSyncer` method:

```go
state, currentBlock := w.sync.syncState()

liveReady := state == syncStateSynced ||
	state == syncStateRescanning

synced := liveReady &&
	syncedTo.Height == currentBlock.Height &&
	syncedTo.Hash.IsEqual(&currentBlock.Hash)
```

this keeps targeted scans in the existing state machine for scan and spend admission, while they no longer make public `Info.Synced` false. it also avoids `scanRequestPending`, `liveStatus`, and a `GetBestBlock` call on every `Info` request.

yyforyongyu (changes requested):
Cool think the commits can be structured better for reviews and we need to not break the `chainSyncer` interface, plus we can avoid the `GetBestBlock` RPC call in `Info` as the syncer already reads every block.

wallet/syncer_test.go
GustavoStingelin: Removed `scanBatchRecorder`. The existing `walletmock.Store` test already checks that targeted batches contain no synced blocks; the new regression uses the existing real-store fixture to check public `Info` and persisted tips.

wallet/syncer.go
GustavoStingelin: Wrapped the bodies of all three commits.

wallet/syncer.go
GustavoStingelin: Kept `syncStateRescanning` after the later discussion, so this deletion is gone. The public recovery fields and their unused helpers/tests are together in the first commit.

wallet/syncer.go
GustavoStingelin: Removed `liveStatus`. `Info` now reads `GetBestBlock` in its admitted handler and uses the existing `syncState` method; `chainSyncer` is unchanged.

wallet/syncer.go
GustavoStingelin: Kept `syncStateRescanning` and the existing scan/spend admission, and removed `scanRequestPending`. I kept the direct tip query because a targeted scan occupies the loop that consumes live notifications: a cached tip can still match `SyncedTo` after the source advances. The regression checks public `Info` during that scan. This costs one query per call, but correctly reports unsynced while live block advancement waits for the rescan. I’d defer caching until tip observation can stay fresh independently of the scan.

yyforyongyu: > a cached tip can still match SyncedTo after the source advances. 

Correct, and this is a planned task in the roadmap, which the new fan-out design, but won't bother this PR with that fow now.

wallet/controller.go
```diff
@@ -592,15 +589,26 @@ func (w *Wallet) handleInfo(r infoReq) {
 		return
 	}
 
+	bestHash, bestHeight, err := w.cfg.Chain.GetBestBlock()
```
yyforyongyu, line 592: Just FYI this is an inherent race because `GetWallet` and `GetBestBlock` read two independently changing sources. It is rare and limited to one `Info` call, and the proper fix belongs to the Manager fan-out work in Tasks 370/371/423, so I don’t think we should redesign it in this PR. What could happen:

- `GetWallet` reads the committed wallet tip at A.
- `syncState` reads `syncStateSynced`.
- The syncer commits B while the state remains `syncStateSynced` for a small live update.
- The chain disconnects or reorgs back to A before `GetBestBlock`.
- `GetBestBlock` returns A, so `Info` reports `Synced=true` using the stale wallet tip even though the committed tip is now B.
- Reading `GetBestBlock` after `syncState` narrows the window, but does not eliminate it because `syncState` is not a version or synchronization barrier.

yyforyongyu (approved):

yyforyongyu (approved):
Overall this looks mergeable. The updated PR is focused and improves the public synchronization result without introducing the two control-flow behaviors below. Both already exist in the base branch, so I think we can address them in a follow-up PR under Task 455. Please keep Task 455 open until that follow-up is merged. The separate cross-source snapshot race can remain with the planned Manager frontier and notification fan-out work.

The Task 455 follow-up should cover:

1. The sync worker can report ready while it is in retry backoff.

   - `runSyncStep` returns a non-shutdown error.
   - `syncer.run` returns it without changing `syncStateSynced`.
   - `runSyncLoop` enters backoff and is not consuming notifications.
   - If the committed tip still matches `GetBestBlock`, `Info` reports `Synced=true` even though live processing is unavailable.

   Before returning the error from `syncer.run`, move the existing state to a non-ready value:

   ```go
   s.state.Store(uint32(syncStateSyncing))
   return err
   ```

   This should only happen when the worker exits with an actual error, so active small-gap syncing keeps the current operation-admission behavior. The regression test should force a worker error at an exact tip, hold the retry backoff, and verify that `Info.Synced` is false.

2. A targeted-rescan failure currently restarts the live sync worker.

   - `scanWithTargets` returns an error and restores `syncStateSynced` through its defer.
   - `waitForEvent` returns that error through `runSyncStep` and `syncer.run`.
   - `runSyncLoop` enters backoff, so a wallet-scoped historical scan interrupts live notification processing.

   `Rescan` already returns after the request is queued, so no caller is waiting for the execution error. Keep a non-shutdown targeted-scan failure local at the scan mailbox boundary and continue the live loop:

   ```go
   case job := <-s.scanReqChan:
       err := s.handleScanReq(ctx, job)
       if err == nil || job.typ != scanTypeTargeted ||
           errors.Is(err, context.Canceled) ||
           errors.Is(err, ErrWalletShuttingDown) {

           return err
       }

       log.Warnf("Targeted rescan failed: %v", err)

       return nil
   ```

   Cancellation, shutdown, and full rewind failures should continue to propagate. The regression test should submit a failing targeted scan through the mailbox and prove the next live notification is processed without a worker restart.

The follow-up should also remove the obsolete state-only `isSynced` predicate and replace the current small-gap status test with one that exercises actual operation admission.

GustavoStingelin:
follow up: https://github.com/btcsuite/btcwallet/pull/1387

PR #1365, opened 2026-09-05
wallet+db: persist account sync policy

yyforyongyu:
## Summary

Implements roadmap Task 345

## Change Description

Persist immutable SQL account no-chain-sync policy for derived and imported
accounts in the canonical pre-production schemas, including read/write
plumbing, truthful legacy kvdb behavior, backend-neutral schema invariants,
and both boolean values through existing Store create/read tests.

## Notes

PRs #1342 and #1345 are prerequisites and are merged into the actual base
branch, `sql-wallet`. Review scope begins at commit `7fdca5ec87` after base
`0d118ce307`.

Intentional roadmap divergence: because the SQL wallet schema is
pre-production, this changes the canonical account schema directly and does
not provide an upgrade migration; existing development databases must be
recreated.

Field-specific reopen verification is also omitted. Existing wallet
integration tests own close-and-reopen lifecycle coverage, while this change
tests both stored values through normal SQL Store create/read paths to avoid
creating a per-column reopen-test category.


wallet/internal/sql/pg/migrations/000011_account_no_chain_sync.down.sql
```diff
@@ -0,0 +1,29 @@
+-- Remove the forward trigger before replacing its function so no schema
+-- object refers to the policy column when the rollback drops that column.
+DROP TRIGGER IF EXISTS trg_assert_account_identity_immutable ON accounts;
+
+CREATE OR REPLACE FUNCTION assert_account_identity_immutable()
```
yyforyongyu: is this related to the change, why do we need this?

wallet/internal/db/kvdb/accountstore_getaccount.go
```diff
@@ -305,6 +305,10 @@ func loadAccountInfo(ns walletdb.ReadBucket,
 		InternalKeyCount: props.InternalKeyCount,
 		ImportedKeyCount: props.ImportedKeyCount,
 		IsWatchOnly:      walletWatchOnly || accountIsImported,
+		// The legacy account encoding has no per-account synchronization
```
yyforyongyu, line 309: newline above

wallet/internal/db/accountstore_createderived.go
```diff
@@ -261,7 +261,7 @@ func CreateDerivedAccountWithOps(ctx context.Context,
 
 	return BuildAccountInfo(
 		accountID, &accNumber, params.Name, false, 0, 0, 0,
-		walletIsWatchOnly, row.CreatedAt, params.Scope, addrSchema,
+		walletIsWatchOnly, false, row.CreatedAt, params.Scope, addrSchema,
```
yyforyongyu: add an inline doc explaining that derived accounts always sync chain?

wallet/internal/db/data_types.go
```diff
@@ -438,6 +438,12 @@ type AccountInfo struct {
 	// may be removed in a future cleanup task.
 	IsWatchOnly bool
 
+	// NoChainSync reports the account's durable automatic synchronization
+	// policy. It is independent of wallet custody and does not change key
```
yyforyongyu: this comment is too verbose on things it should not explain, instead, explain what this variable does instead, no need to mention it's durable or independent?

wallet/internal/db/kvdb/accountstore_createderived.go
```diff
@@ -20,6 +20,14 @@ func (s *Store) CreateDerivedAccount(ctx context.Context,
 	params db.CreateDerivedAccountParams,
 	deriveFn db.AccountDerivationFunc) (*db.AccountInfo, error) {
 
+	// The legacy bucket format cannot persist this SQL account policy. Reject
+	// true before opening a transaction so the caller never receives a value
+	// that would be lost on the next read.
+	if params.NoChainSync {
```
yyforyongyu: we should just ignore it sliently rather than creating yet another error case - the kvdb is deprecated path and we don't wanna maintain it.

wallet/internal/db/kvdb/accountstore_createimported.go
```diff
@@ -18,6 +18,14 @@ import (
 func (s *Store) CreateImportedAccount(_ context.Context,
 	params db.CreateImportedAccountParams) (*db.AccountInfo, error) {
 
+	// The legacy bucket format cannot persist this SQL account policy. Reject
+	// true before validation or a transaction so a successful import can never
+	// lose the requested value when it is read back.
+	if params.NoChainSync {
```
yyforyongyu: same here re removing the check - we should remove all similar checks

wallet/internal/db/accountstore_createimported.go
```diff
@@ -94,8 +99,9 @@ type CreateImportedAccountOps interface {
 	EnsureKeyScope(ctx context.Context, walletID uint32, scope KeyScope,
 		addrSchema *ScopeAddrSchema) (int64, error)
 
-	// CreateImportedAccount inserts the imported account row for the provided
-	// insert request and returns the created account row ID.
+	// CreateImportedAccount inserts the imported account row for the
+	// provided request and returns its row ID. The adapter stores NoChainSync
```
yyforyongyu: i don't think we wanna emphasize on this new `NoChainSync` on this new method? I see the tendency here - whenever we update a new area, like a var, you seem to emphasize this area only, which makes the whole comment unbalanced towards some details that should not appear in a highlevel place.

wallet/internal/db/itest/fixtures_pg_test.go
```diff
@@ -26,6 +26,12 @@ func testBackend() dberr.Backend {
 	return dberr.BackendPostgres
 }
 
+// testQueryParam returns PostgreSQL's numbered marker so shared raw-SQL tests
+// bind values safely without duplicating their dialect-neutral assertions.
+func testQueryParam(position int) string {
```
yyforyongyu: why do we need these shallow shims?

wallet/internal/db/itest/fixtures_sqlite_test.go
```diff
@@ -27,6 +27,12 @@ func testBackend() dberr.Backend {
 	return dberr.BackendSQLite
 }
 
+// testQueryParam returns SQLite's positional marker so shared raw-SQL tests
+// bind values safely without duplicating their dialect-neutral assertions.
+func testQueryParam(_ int) string {
```
yyforyongyu: same here, too shallow?

wallet/internal/sql/pg/migrations.go
```diff
@@ -53,6 +53,22 @@ func ApplyMigrations(db *sql.DB) error {
 	return nil
 }
 
+// RollbackMigration reverts exactly the latest PostgreSQL migration so
+// migration tests can seed and reopen the immediately preceding durable schema.
+func RollbackMigration(db *sql.DB) error {
```
yyforyongyu: explain why we need this change here? the migration is unrelated? plus the latter two test commits should be merged into this commit?

wallet/internal/sql/pg/migrations/000011_account_no_chain_sync.down.sql
yyforyongyu: Removed. That function restoration was only needed by the now-removed down migration; the canonical pre-production schema no longer requires it.

wallet/internal/db/data_types.go
yyforyongyu: Fixed. The field comment now states only that automatic chain synchronization excludes the account.

wallet/internal/db/accountstore_createderived.go
yyforyongyu: The false argument is the isImported flag, not the chain-sync policy. Derived accounts must pass false there; the following params.NoChainSync argument carries the requested policy, so I left the call unchanged.

wallet/internal/db/itest/fixtures_pg_test.go
yyforyongyu: Removed. The shared invariant test now selects the backend placeholder syntax locally, following the existing test structure.

wallet/internal/db/kvdb/accountstore_createimported.go
yyforyongyu: Fixed. The imported kvdb path now ignores the field and continues its historical behavior.

wallet/internal/sql/pg/migrations.go
yyforyongyu: Removed. With the pre-production schema decision, the rollback hooks, later migration, and upgrade tests are gone. SQLite and PostgreSQL reopen coverage are combined in one durability-test commit.

wallet/internal/db/kvdb/accountstore_createderived.go
yyforyongyu: Fixed. kvdb now silently normalizes the by-value request to false so both stored and returned state reflect the legacy backend behavior, without adding a new error case.

wallet/internal/db/accountstore_createimported.go
yyforyongyu: Fixed. The high-level interface comment now only describes inserting the account row and returning its ID.

wallet/internal/db/itest/fixtures_sqlite_test.go
yyforyongyu: Removed. The shared invariant test now selects the backend placeholder syntax locally, following the existing test structure.

wallet/internal/db/kvdb/accountstore_getaccount.go
yyforyongyu: Fixed by adding the requested blank line above the explanatory field comment.

wallet/internal/db/itest/accountstore_createinvariants_test.go
```diff
@@ -196,6 +199,171 @@ func TestDerivedAccountNumberImmutable(t *testing.T) {
...
+			NoChainSync: true,
+		}, SpendableDeriveFn(),
+	)
+	require.NoError(t, err)
+	scopeID := GetKeyScopeID(t, store.Queries(), walletID, scope)
+	accountID := GetAccountID(
+		t, store.Queries(), scopeID, "immutable-policy",
+	)
+
+	omittedInsert := `INSERT INTO accounts
```
yyforyongyu: add inline docs explaining why we need to use raw sql query here

yyforyongyu: Addressed in 5ceea075. The inline comments now explain that existing Store methods cannot omit the policy column, pass NULL or an invalid SQLite boolean value, or update the immutable policy. Raw SQL is therefore required to exercise those database constraints directly, and each statement documents the specific behavior it verifies.

wallet/internal/db/itest/accountstore_createinvariants_test.go
```diff
@@ -196,6 +199,178 @@ func TestDerivedAccountNumberImmutable(t *testing.T) {
...
+	accountID := GetAccountID(
+		t, store.Queries(), scopeID, "immutable-policy",
+	)
+
+	// A raw insert can omit the policy column, unlike Store writes, and
+	// therefore verifies that the database itself supplies the false default.
+	omittedInsert := `INSERT INTO accounts
+		(wallet_id, scope_id, account_name, is_derived, public_key)
+		VALUES (?, ?, ?, FALSE, ?)`
+	// A raw insert can pass NULL or SQLite's integer 2, values that the Store's
```
yyforyongyu: newline above?

yyforyongyu: Fixed in 9ea347f4. Added the missing blank lines so each explanatory comment is visually separated from the preceding raw SQL declaration.

wallet/internal/db/itest/accountstore_createinvariants_test.go
```diff
@@ -196,6 +199,180 @@ func TestDerivedAccountNumberImmutable(t *testing.T) {
...
+	// typed bool parameter cannot represent, to exercise schema rejection.
+	explicitInsert := `INSERT INTO accounts
+		(wallet_id, scope_id, account_name, is_derived, no_chain_sync,
+		 public_key) VALUES (?, ?, ?, FALSE, ?, ?)`
+
+	// A raw update reaches the immutability trigger because the Store
+	// intentionally exposes no operation for changing this policy.
+	update := `UPDATE accounts SET no_chain_sync = ? WHERE id = ?`
+
+	if _, ok := any(store).(*pg.Store); ok {
```
GustavoStingelin: I'd keep this test shared and put the backend-specific SQL operations and driver assertions in the fixture helpers, following [insertWalletSecretRaw](https://github.com/btcsuite/btcwallet/blob/9ea347f483d6fc34cefae16cc181af00dc427f4c/wallet/internal/db/itest/fixtures_pg_test.go#L336-L357) and [requireDriverConstraintError](https://github.com/btcsuite/btcwallet/blob/9ea347f483d6fc34cefae16cc181af00dc427f4c/wallet/internal/db/itest/fixtures_pg_test.go#L50-L64), which have matching SQLite implementations.

That lets the shared test express the same default/NULL/immutability expectations without checking `*pg.Store` or selecting SQL syntax. I'd also drop the SQLite integer-boolean test, since our typed SQL queries already take a `bool`.

wallet/internal/db/itest/accountstore_createinvariants_test.go
```diff
@@ -196,6 +199,180 @@ func TestDerivedAccountNumberImmutable(t *testing.T) {
 	requireDriverConstraintError(t, err)
 }
 
+// TestAccountNoChainSyncSurvivesReopen verifies both stored policy values
+// remain observable after all state from the creating Store has been released.
+func TestAccountNoChainSyncSurvivesReopen(t *testing.T) {
```
GustavoStingelin: I don't think we need the reopen test here. I'd cover `false` and `true` in the existing Store create/read tests and keep lifecycle coverage in shared DB tests or wallet itests. Requiring reopen checks for individual fields would be a broader change in our DB testing policy, so I think the task requirement should be adjusted.

GustavoStingelin (changes requested):

wallet/internal/db/itest/accountstore_createinvariants_test.go
yyforyongyu: yeah the agent fought really hard on this - but totally agree, we should try to add minimal changes here and don't wanna open a new category of tests.

wallet/internal/db/itest/accountstore_createinvariants_test.go
yyforyongyu: cool will update

yyforyongyu: Addressed in f7e1dc1e. The shared invariant test no longer imports PostgreSQL code or selects placeholder syntax. Its raw default, NULL, and update operations now use matching build-tagged fixture helpers, and the SQLite-only integer-boolean case has been removed.

wallet/internal/db/itest/accountstore_createinvariants_test.go
yyforyongyu: Addressed in f7e1dc1e. Removed the field-specific reopen test and its added lifecycle plumbing. The existing derived and imported Store create/read integration tests now assert both false and true, while the existing wallet integration suite remains responsible for close-and-reopen lifecycle coverage.

GustavoStingelin (approved):
LGTM!

PR #1366, opened 2026-09-07
wallet: carry account chain-sync policy

yyforyongyu:
## Summary

Implements roadmap Task 392.

## Change Description

Carry account sync policy through NewAccountParams and AccountInfo. Preserve
ordinary creation and reject NoChainSync=true after admission, before secret
preparation or storage mutation, until activation is enabled.

## Notes

Depends on #1365. The base branch task-account-creation-base mirrors its head
f7e1dc1e05f8fd74c66c65af81c5527efb4e7e05; review scope starts after #1365.
The account error boundary from #1358 is now merged into sql-wallet and
included in this base. Receiving remains outside this PR.

Issue #34, opened 2014-01-14
respond to importprivkey before rescan


dajohi:
btcwallet should respond to the importprivkey command before initiating a rescan request to btcd.  this would stop the client, such as btcctl, from waiting for rescan to finish.

PR #1367, opened 2026-09-07
wallet: reject NoChainSync receiving requests

yyforyongyu:
## Summary

Implements roadmap Task 393: reject receiving requests for NoChainSync accounts.

## Change Description

Guard NewAddress before allocation using its existing account read. Guard GetUnusedAddress using policy from its existing address query, with the guarded allocation fallback for an empty result. Reuse the public unsupported-operation error and preserve internal change allocation without adding database queries.

Update both API docs and existing wallet/SQL tests. Unit, SQLite/PostgreSQL integration, ordinary receiving checks across all three backends, lint, and generation checks pass.

## Notes

Depends on #1366; review scope starts after it. Automatic scan exclusions remain in #1368. Separate key allocation belongs to Task 395, as described in [roadmap PR #34](https://github.com/yyforyongyu/btcwallet-sql-roadmap/pull/34).

### DIFF
--- a/wallet/address_manager.go
+++ b/wallet/address_manager.go
@@ -168,7 +168,8 @@ type OutputScriptInfo struct {
 // addresses and scripts.
 type AddressManager interface {
 	// NewAddress returns a new address for the given account and address
-	// type.
+	// type. NoChainSync accounts return ErrAccountOperationUnsupported
+	// because receiving requires automatic chain tracking.
 	//
 	// NOTE: This method should be used with caution. Unlike
 	// GetUnusedAddress, it does not scan for previously derived but unused
@@ -180,11 +181,10 @@ type AddressManager interface {
 		addrType waddrmgr.AddressType,
 		change bool) (address.Address, error)
 
-	// GetUnusedAddress returns the first, oldest, unused address by
-	// scanning forward from the start of the derivation path. This method
-	// is the recommended default for obtaining a new receiving address, as
-	// it prevents address reuse and avoids creating gaps in the address
-	// chain that could impact wallet recovery.
+	// GetUnusedAddress returns the oldest address not recorded as used by
+	// the wallet, or allocates one if none remains. NoChainSync accounts
+	// return ErrAccountOperationUnsupported because their address use is not
+	// tracked automatically.
 	GetUnusedAddress(ctx context.Context, accountName string,
 		addrType waddrmgr.AddressType, change bool) (
 		address.Address, error)
@@ -572,6 +572,8 @@ type derivationInfoResp struct {
 // NewAddress returns a new address for the given account and address type.
 // This method is a low-level primitive that will always derive a new, unused
 // address from the end of the address chain.
+// NoChainSync accounts return ErrAccountOperationUnsupported before allocation
+// because receiving requires automatic chain tracking.
 //
 // It returns the next external or internal address for the wallet dictated by
 // the value of the `change` parameter. If change is true, then an internal
@@ -676,15 +678,26 @@ func (w *Wallet) newAddress(ctx context.Context, accountName string,
 		return nil, fmt.Errorf("%w: %v", ErrUnknownAddrType, addrType)
 	}
 
+	// Require receiving admission during the allocation's existing account
+	// read, so excluded accounts consume no child and need no extra lookup.
 	addrInfo, err := w.store.NewDerivedAddress(
 		ctx, db.NewDerivedAddressParams{
-			WalletID:    w.id,
-			AccountName: accountName,
-			Scope:       db.KeyScope(keyScope),
-			Change:      change,
+			WalletID:         w.id,
+			AccountName:      accountName,
+			Scope:            db.KeyScope(keyScope),
+			Change:           change,
+			RequireChainSync: true,
 		},
 	)
 	if err != nil {
+		// Keep the Store identity internal while exposing the established
+		// public unsupported-operation error with the refusal diagnostic.
+		if errors.Is(err, db.ErrAccountOperationUnsupported) {
+			return nil, fmt.Errorf(
+				"%w: %s", ErrAccountOperationUnsupported, err.Error(),
+			)
+		}
+
 		return nil, err
 	}
 
@@ -705,13 +718,15 @@ func (w *Wallet) newAddress(ctx context.Context, accountName string,
 
 // GetUnusedAddress returns the first, oldest, unused address by scanning
 // forward from the start of the derivation path. The address is considered
-// "unused" if it has never appeared in a transaction. This method is the
-// recommended default for obtaining a new receiving address. It prevents
-// address reuse and avoids creating gaps in the address chain, which is
-// critical for reliable wallet recovery under standards like BIP44 that
-// enforce a gap limit of 20 unused addresses. If all previously derived
-// addresses have been used, this method will delegate to NewAddress to
-// generate a new one.
+// "unused" if the wallet has not recorded it as used. For accounts that track
+// address use, this is the recommended default for obtaining a receiving
+// address: it avoids reuse and gaps in the address chain, supporting recovery
+// under standards like BIP44 that enforce a gap limit of 20 unused addresses.
+// If all previously derived addresses have been used, this method delegates
+// to NewAddress to generate a new one.
+// NoChainSync accounts return ErrAccountOperationUnsupported because their
+// address use is not tracked automatically. Rejection does not allocate a child
+// or register a chain watch.
 //
 // TODO(yy): The current implementation of GetUnusedAddress is inefficient for
 // wallets with a large number of used addresses. It iterates from the first
@@ -803,6 +818,18 @@ func (w *Wallet) handleGetUnusedAddress(r getUnusedAddressReq) {
 	for storeAddr, err := range addresses {
 		if err != nil {
 			r.respChan <- addressResp{err: err}
+
+			return
+		}
+
+		// Read the policy from the existing account join before treating a
+		// locally unused child as a receiving address.
+		if storeAddr.NoChainSync {
+			r.respChan <- addressResp{
+				err: fmt.Errorf("%w: account %q has chain "+
+					"synchronization disabled",
+					ErrAccountOperationUnsupported, r.accountName),
+			}
 
 			return
 		}
--- a/wallet/address_manager_test.go
+++ b/wallet/address_manager_test.go
@@ -6,6 +6,7 @@ package wallet
 
 import (
 	"bytes"
+	"fmt"
 	"iter"
 	"sync"
 	"testing"
@@ -353,13 +354,16 @@ func expectStoreNewAddress(t *testing.T, w *Wallet, deps *mockWalletDeps,
 
 	t.Helper()
 
+	// Public receiving requires the Store to reject excluded accounts before
+	// allocation and forward the caller context; success requires watching.
 	deps.store.On(
-		"NewDerivedAddress", mock.Anything,
+		"NewDerivedAddress", t.Context(),
 		db.NewDerivedAddressParams{
-			WalletID:    w.id,
-			AccountName: accountName,
-			Scope:       db.KeyScope(scope),
-			Change:      change,
+			WalletID:         w.id,
+			AccountName:      accountName,
+			Scope:            db.KeyScope(scope),
+			Change:           change,
+			RequireChainSync: true,
 		},
 	).Return(addressInfoFromAddr(t, addr), nil).Once()
 	deps.chain.On("NotifyReceived", []address.Address{addr}).Return(nil).Once()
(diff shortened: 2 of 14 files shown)


wallet/address_manager.go
```diff
@@ -469,9 +472,36 @@ func (w *Wallet) addrBalances(ctx context.Context) (map[string]btcutil.Amount,
...
 }
 
+// checkReceivingAccount rejects accounts whose persisted policy cannot support
+// receiving, before callers derive addresses or consult observed address use.
+func (w *Wallet) checkReceivingAccount(ctx context.Context, accountName string,
+	keyScope waddrmgr.KeyScope) error {
+
+	// Policy is immutable account metadata; the existing account read supplies
+	// it without fetching balances or entering address allocation or scanning.
+	account, err := w.cache.GetAccount(ctx, db.GetAccountQuery{
```
yyforyongyu: should this be fixed at the sql query level instead? why using another sql query to get the account?

yyforyongyu: Updated in 2abe7d0051c5. `NewAddress` now checks the policy in the shared SQL allocation workflow using the account row it already reads, eliminating the separate wallet lookup.

`GetUnusedAddress` retains its pre-scan check because it can return an existing address without allocating. Internal transaction-change allocation keeps its existing behavior. SQLite and PostgreSQL regression tests cover receiving refusal and preserved transaction-change allocation.

wallet/internal/db/data_types.go
```diff
@@ -826,6 +826,12 @@ type AddressSecret struct {
 // NewDerivedAddressParams contains the parameters for creating a new derived
 // address.
 type NewDerivedAddressParams struct {
+	// RequireChainSync requests receiving-policy enforcement during account
+	// selection. Receiving callers set this even for internal addresses;
+	// transaction-change allocation leaves it false to retain key derivation
+	// independently of automatic chain synchronization.
+	RequireChainSync bool
```
yyforyongyu: why name it this instead of `NoChainSync`?

wallet/address_manager.go
```diff
@@ -469,9 +472,36 @@ func (w *Wallet) addrBalances(ctx context.Context) (map[string]btcutil.Amount,
...
+		Scope:       db.KeyScope(keyScope),
+		Name:        &accountName,
+		SkipBalance: true,
+	})
+	if err != nil {
+		return fmt.Errorf("get receiving account: %w", err)
+	}
+
+	if account.NoChainSync {
+		return fmt.Errorf("receiving account %q: %w", accountName,
```
yyforyongyu: we should mention the account is no chain sync?

wallet/internal/db/addressstore_newderived.go
```diff
@@ -141,6 +147,13 @@ func NewDerivedAddressWithOps(ctx context.Context,
 		return nil, fmt.Errorf("get account: %w", err)
 	}
 
+	// Receiving promises chain tracking, unlike internal transaction change.
+	// Refuse from the account row already loaded before touching counters.
+	if params.RequireChainSync && account.NoChainSync {
```
yyforyongyu, line 150: here is one thought - i actually don't think the address should care about no chain sync? as long as the derived addresses are not used during chain sync we are fine?

wallet/address_manager.go
yyforyongyu: Assessed and transferred the exact signed commit `25c5893d63fc9c74ecec635e615341dc5a2df158` into #1366 as requested. It already addresses the redundant NewAddress lookup using the account row read by SQL allocation; GetUnusedAddress retains its necessary pre-scan check. Focused public tests, full Wallet/shared DB tests, lint, and the existing SQLite/PostgreSQL receiving-refusal regression passed. #1367 now retains only its backend regression-test commit relative to #1366.

wallet/address_manager.go
```diff
@@ -185,6 +186,8 @@ type AddressManager interface {
 	// is the recommended default for obtaining a new receiving address, as
 	// it prevents address reuse and avoids creating gaps in the address
 	// chain that could impact wallet recovery.
+	// NoChainSync accounts return ErrAccountOperationUnsupported because
```
yyforyongyu: no longer accurate?

wallet/internal/db/data_types.go
yyforyongyu: The per-call RequireChainSync flag is removed. The allocation result now carries the existing NoChainSync account policy solely to skip immediate watch registration, without an extra account lookup. Published in b6ef0d4496aa50e6025bb05ab793c5d905c78765.

wallet/address_manager.go
yyforyongyu: The refusal is now removed from both address APIs, so this diagnostic is no longer needed. NoChainSync accounts can allocate an address or use the existing locally-unused selection without registering a chain watch. Published in b6ef0d4496aa50e6025bb05ab793c5d905c78765.

wallet/internal/db/addressstore_newderived.go
yyforyongyu: Adopted: address derivation no longer rejects NoChainSync accounts, and the allocation result lets NewAddress skip immediate watch registration. GetUnusedAddress now uses its existing local selection and unwatched fallback without an account-policy check. The complementary automatic scan exclusions remain in #1368. Published in b6ef0d4496aa50e6025bb05ab793c5d905c78765.

wallet/address_manager.go
yyforyongyu: Updated both the documentation and behavior: GetUnusedAddress no longer checks or rejects NoChainSync accounts. It returns the oldest address not locally recorded as used, or allocates one without registering a chain watch. The docs clarify that untracked use can cause repeated calls to return the same address after payment. Published in b6ef0d4496aa50e6025bb05ab793c5d905c78765.

wallet/internal/db/data_types.go
```diff
@@ -841,6 +846,11 @@ type NewDerivedAddressParams struct {
 	// Change indicates whether to create a change address (true) or an
 	// external address (false).
 	Change bool
+
+	// RequireChainSync rejects NoChainSync accounts before allocating a
+	// receiving address. The false default preserves internal key and change
+	// allocation; this requirement never changes the stored account policy.
+	RequireChainSync bool
```
yyforyongyu, line 849: there is no need to have this `RequireChainSync`? We can know whether this account has nochainsync or not already by looking into the account?

wallet/internal/sql/pg/queries/addresses.sql
```diff
@@ -172,6 +172,8 @@ SELECT
     da.account_id,
     acc.account_number,
     acc.account_name,
+    -- Receiving admission uses the account already joined by this query.
+    acc.no_chain_sync,
```
yyforyongyu, line 176: merge `sqlc: regenerate address policy results` into this commit

wallet/address_manager_test.go
```diff
@@ -654,11 +740,75 @@ func TestGetUnusedAddress(t *testing.T) {
...
+	// Assert: selection returns the requested internal child, and every
+	// expected scan, fallback allocation, and notification occurred.
 	require.NoError(t, err)
 	require.Equal(t, changeAddrVal.String(), unusedChangeAddr.String())
+	deps.store.AssertExpectations(t)
+	deps.chain.AssertExpectations(t)
+
+	// Both branches must reject excluded accounts whether a child is already
+	// listed or the empty iterator reaches the existing allocation fallback.
+	for _, change := range []bool{false, true} {
```
yyforyongyu: i find this test hard to follow

wallet/internal/db/itest/addressstore_test.go
```diff
@@ -1931,24 +1931,28 @@ func TestListAddressesWatchOnlyWallet(t *testing.T) {
...
 func TestNewDerivedAddress(t *testing.T) {
 	t.Parallel()
 
-	store := NewTestStore(t)
-	walletID := newWallet(t, store, "wallet-derived")
+	// Arrange: observe the existing derivation callback while using a real
+	// Store. Sequential cases can prove refusal never invokes derivation.
+	deriveCalls := 0
+	derive := mockDeriveFunc()
+	store := NewTestStoreWithDerive(t, func(ctx context.Context,
```
yyforyongyu: why having this method when we already have all the mock stores?

wallet/internal/db/data_types.go
yyforyongyu: The stored account already supplies NoChainSync through the existing read. RequireChainSync is an internal operation requirement: public NewAddress sets it, while internal transaction-change allocation leaves it false. Keeping that distinction preserves internal allocation without an extra account lookup. It is confined to wallet/internal/db and is not exposed by the public receiving APIs.

wallet/internal/sql/pg/queries/addresses.sql
yyforyongyu: Combined the SQL query change and its generated output in 7b5e6ed6be00490e6a9c638332772b6ef8fb84cd. The stack now has three commits.

wallet/address_manager_test.go
yyforyongyu: Moved these existing rejection scenarios into TestGetUnusedAddressNoChainSync in the same file. All four branch/path cases now have names, independent strict mocks, and linear Arrange/Act/Assert phases; stored-child and empty-account setups no longer share conditional mock logic. The focused tests, wallet package tests, and lint pass.

wallet/internal/db/itest/addressstore_test.go
yyforyongyu: NewTestStoreWithDerive is an existing fixture that constructs a real SQLite or PostgreSQL Store. The wrapper observes the existing derivation callback so this integration test proves refusal occurs before derivation, in addition to checking real SQL persistence and counter preservation. A mock Store would bypass those checks; an empty database after rollback alone would not prove derivation was skipped.

wallet/address_manager_test.go
```diff
@@ -532,6 +549,65 @@ func TestNewAddress(t *testing.T) {
...
+		name        string
+		accountName string
+		addrType    waddrmgr.AddressType
+	}{
+		{
+			name:        "untracked account p2wkh",
+			accountName: "key-only",
+			addrType:    waddrmgr.WitnessPubKey,
+		},
+	}
```
GustavoStingelin: I'd flatten this into a regular test since there's only one case.

wallet/address_manager_test.go
```diff
@@ -654,11 +743,125 @@ func TestGetUnusedAddress(t *testing.T) {
...
+		name   string
+		change bool
+	}{
+		{
+			name: "external",
+		},
+		{
+			name:   "internal",
+			change: true,
+		},
```
GustavoStingelin, line 839: I'd set `change: false` explicitly so both branch values are easy to compare.

```suggestion
		{
			name:   "external",
			change: false,
		},
		{
			name:   "internal",
			change: true,
		},
```

wallet/internal/db/itest/addressstore_test.go
```diff
@@ -1985,6 +2004,106 @@ func TestNewDerivedAddress(t *testing.T) {
...
+		name   string
+		change bool
+	}{
+		{
+			name: "external",
+		},
+		{
+			name:   "internal",
+			change: true,
+		},
```
GustavoStingelin, line 2038: I'd make `change: false` explicit here too, so both cases show the branch value being tested.

```suggestion
		{
			name:   "external",
			change: false,
		},
		{
			name:   "internal",
			change: true,
		},
```

wallet/internal/db/pg/addressstore_listaddresses.go
```diff
@@ -81,6 +81,9 @@ func listAddressesByAccount(ctx context.Context, q *sqlc.Queries,
 					err)
 		}
 
+		// The existing account join supplies receiving policy without a
+		// separate lookup or hiding this child from ordinary address lists.
+		item.NoChainSync = row.NoChainSync
```
GustavoStingelin: I'd map `NoChainSync` inside the `ListAddressesByAccountRow` case of `addressRowToInfo`, so callers don't have to finish mapping the returned `db.AddressInfo`. Same for SQLite.

GustavoStingelin (changes requested):

wallet/internal/db/pg/addressstore_listaddresses.go
yyforyongyu: nice catch! will update

wallet/address_manager_test.go
yyforyongyu: Flattened TestNewAddressNoChainSync into a regular test with the same assertions in 7461105813e436d87565a05add35114536b8ff0b.

wallet/address_manager_test.go
yyforyongyu: Added explicit change: false to the external case in 7461105813e436d87565a05add35114536b8ff0b.

wallet/internal/db/itest/addressstore_test.go
yyforyongyu: Added explicit change: false to the external case in 7fd795ffc0c1749717a8343613488f98edd688c1.

wallet/internal/db/pg/addressstore_listaddresses.go
yyforyongyu: Moved NoChainSync mapping into the ListAddressesByAccountRow case of addressRowToInfo for both PostgreSQL and SQLite in 7461105813e436d87565a05add35114536b8ff0b. Both dialects pass the existing allocation and address-list integration tests.

GustavoStingelin (approved):
LGTM!

PR #1368, opened 2026-09-07
wallet: exclude key-only recovery accounts

yyforyongyu:
## Summary

Exclude NoChainSync accounts from recovery scan inputs.

## Change Description

Select eligible accounts in the existing SQL account-list queries before target resolution, horizon loading, and address loading. Keep ordinary account listing inclusive and use the existing missing-account error for excluded explicit targets. Cover persisted SQL policy at zero and positive recovery windows.

## Notes

Depends on #1367; review scope starts after that PR.

### DIFF
--- a/wallet/internal/sql/pg/queries/accounts.sql
+++ b/wallet/internal/sql/pg/queries/accounts.sql
@@ -247,6 +247,7 @@ ORDER BY a.account_number NULLS LAST, a.account_name;
 
 -- name: ListAccountsByWalletScope :many
 -- Lists all accounts for a wallet and scope tuple.
+-- Scan callers opt in so ordinary account listing keeps key-only accounts.
 SELECT
     a.id,
     a.account_number,
@@ -267,13 +268,18 @@ FROM accounts AS a
 INNER JOIN key_scopes AS ks ON a.scope_id = ks.id
 INNER JOIN wallets AS w ON a.wallet_id = w.id
 WHERE
-    ks.wallet_id = $1
-    AND ks.purpose = $2
-    AND ks.coin_type = $3
+    ks.wallet_id = sqlc.arg('wallet_id')
+    AND ks.purpose = sqlc.arg('purpose')
+    AND ks.coin_type = sqlc.arg('coin_type')
+    AND (
+        sqlc.arg('chain_sync_only')::BOOLEAN = FALSE
+        OR a.no_chain_sync = FALSE
+    )
 ORDER BY a.account_number NULLS LAST, a.account_name;
 
 -- name: ListAccountsByWalletAndName :many
 -- Lists all accounts for a wallet filtered by account name.
+-- Scan callers opt in so ordinary account listing keeps key-only accounts.
 SELECT
     a.id,
     a.account_number,
@@ -293,11 +299,18 @@ SELECT
 FROM accounts AS a
 INNER JOIN key_scopes AS ks ON a.scope_id = ks.id
 INNER JOIN wallets AS w ON a.wallet_id = w.id
-WHERE ks.wallet_id = $1 AND a.account_name = $2
+WHERE
+    ks.wallet_id = sqlc.arg('wallet_id')
+    AND a.account_name = sqlc.arg('account_name')
+    AND (
+        sqlc.arg('chain_sync_only')::BOOLEAN = FALSE
+        OR a.no_chain_sync = FALSE
+    )
 ORDER BY a.account_number NULLS LAST, a.account_name;
 
 -- name: ListAccountsByWallet :many
 -- Lists all accounts for a wallet.
+-- Scan callers opt in so ordinary account listing keeps key-only accounts.
 SELECT
     a.id,
     a.account_number,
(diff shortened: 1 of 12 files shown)


wallet/syncer.go
```diff
@@ -2040,6 +2055,12 @@ func (s *syncer) storeTargetedScanHorizons(ctx context.Context,
 			return nil, err
 		}
 
+		// Horizon loading precedes watch loading and state construction,
+		// so reject forbidden discovery before either can do any work.
+		if s.cfg.RecoveryWindow > 0 && info.NoChainSync {
```
yyforyongyu: why are we checking recovery window and nochainsync together? also shouldn't nochainsync accounts be filtered before we even enter the syncing?

yyforyongyu: Addressed in 71461a4d2e727919fb9bf97c7e0d2b9b2d7944a3. `loadStoreScanData` now filters `NoChainSync` accounts before horizon and address loading, independently of `RecoveryWindow`. Both loaders reuse the admitted account snapshot, so the downstream policy checks are removed. Explicit forbidden targets are rejected during admission with `ErrNoChainSyncRecoveryTarget`.

Regression coverage includes zero and positive recovery windows, early rejection before horizon lookup, and persisted SQL policy. All 19 CI checks pass on the updated head.

wallet/syncer.go
```diff
@@ -1959,36 +1966,27 @@ func (s *syncer) scanWithTargets(ctx context.Context, req *scanReq) error {
...
-// storeScanHorizons loads account horizon data through the store. Targets are
-// the identity-aware scanTargets produced by resolveScanTargets, so each one
-// already carries the durable AccountName that lets the Store resolve an
-// imported account without colliding with the default derived account at the
+// storeScanHorizons loads horizons for accounts admitted by loadStoreScanData.
+// Targets come from resolveScanTargets with a durable AccountName, so imported
+// accounts resolve without colliding with the default derived account at the
 // masked number 0.
 func (s *syncer) storeScanHorizons(ctx context.Context,
+	accounts []db.AccountInfo,
```
yyforyongyu: why do we add another param to this method? the filtering of nochainsync accounts should happen when we prepare the `[]scanTarget`?

yyforyongyu: Addressed in the published stack at 22deadef94dad20dd57be8b692d101e4387ccdd7. Selection now happens in the existing SQL account-list queries through `ChainSyncOnly`, before target resolution, horizon loading, and address loading. The extra loader parameters and `ErrNoChainSyncRecoveryTarget` are removed; excluded explicit targets use the existing `db.ErrAccountNotFound` path. Verified persisted behavior on SQLite and PostgreSQL at zero and positive recovery windows.

wallet/syncer.go
```diff
@@ -2494,9 +2500,12 @@ func (s *syncer) resolveScanTargets(ctx context.Context,
...
 	ctx context.Context) (map[waddrmgr.AccountScope]string, error) {
 
+	// Resolve targets only from eligible SQL rows. An excluded explicit
+	// target then follows the existing missing-account error path.
 	accounts, err := s.store.ListAccounts(ctx, db.ListAccountsQuery{
-		WalletID:    s.walletID,
-		SkipBalance: true,
+		WalletID:      s.walletID,
+		SkipBalance:   true,
+		ChainSyncOnly: true,
```
GustavoStingelin: After this query admits ordinary account A, renaming A away and a `NoChainSync` account into its old name makes `storeScanHorizonAccount` load the excluded xpub. I reproduced this through two public `RenameAccount` calls on SQLite with a recovery window of 20; the excluded account's scripts entered the watch set.

I'd carry the admitted identity or snapshot into horizon loading, or recheck `NoChainSync` on the fetched row before derivation, and cover that rename sequence.

wallet/syncer.go
GustavoStingelin: The removal follows the discussion above, but [Task 452](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/91966f26f77981b30c0d9a07a977c87475ba6418/tasks/452-key-only-recovery-derivation.md) still requires `wallet.ErrNoChainSyncRecoveryTarget`, and [Task 398](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/91966f26f77981b30c0d9a07a977c87475ba6418/tasks/398-semantic-account-recovery-targets.md) expects to preserve it. Using `db.ErrAccountNotFound` makes an excluded target indistinguishable from a missing one.

I'd keep the wallet-owned policy error at target admission, or update both tasks if treating these cases identically is the intended contract.

wallet/internal/sql/pg/queries/accounts.sql
```diff
@@ -267,13 +268,18 @@ FROM accounts AS a
...
 INNER JOIN wallets AS w ON a.wallet_id = w.id
 WHERE
-    ks.wallet_id = $1
-    AND ks.purpose = $2
-    AND ks.coin_type = $3
+    ks.wallet_id = sqlc.arg('wallet_id')
+    AND ks.purpose = sqlc.arg('purpose')
+    AND ks.coin_type = sqlc.arg('coin_type')
+    AND (
+        sqlc.arg('chain_sync_only')::BOOLEAN = FALSE
```
GustavoStingelin: I'd combine `4d712798` and `ac7a1db0` so the SQL queries, generated bindings, and callers stay in sync within one commit, as we did in #1367. The recovery consumer change can remain separate.

GustavoStingelin (approved):
LGTM

PR #1369, opened 2026-09-07
wallet: create exact accounts with sync policy

yyforyongyu:
## Summary

Create next-or-exact root-derived SQL accounts, including exact accounts with automatic chain synchronization disabled. Implements roadmap Task 343 and Task 468.

## Change Description

Reuse atomic account creation and the scope cursor for sparse numbers and lower holes. Preserve public conflict and indeterminate-commit errors, sequential behavior, and unsupported kvdb exact selection. An admitted creation waits for its Store outcome so cancellation cannot hide an uncertain commit.

## Notes

Prerequisites #1366, #1367, and #1368 are merged into sql-wallet. This PR contains only the eight exact-account and activation commits, including the work formerly in #1370.

### DIFF
--- a/itest/account_manager_test.go
+++ b/itest/account_manager_test.go
@@ -104,17 +104,22 @@ func deterministicImportedAccountKeys(
 // testAccountManagerCreateAccount verifies that a new derived account's
 // returned view matches an immediate read and survives a wallet reload.
 func testAccountManagerCreateAccount(h *bwtest.HarnessTest) {
+	// Arrange: create an unlocked wallet with the harness so this ordinary
+	// sequential request exercises the same contract on every backend.
 	const accountName = "account manager created"
 
 	scope := waddrmgr.KeyScopeBIP0084
 	ctx := h.Context()
 	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
 
+	// Act: create one root-derived account through the public operation.
 	created, err := w.NewAccount(ctx, wallet.NewAccountParams{
 		Scope: scope,
 		Name:  accountName,
 	})
 
+	// Assert: compare returned key facts and the complete account view with
+	// immediate and reopened reads to prove durable sequential creation.
 	require.NoError(h, err, "failed to create derived account")
 	require.Equal(h, accountName, created.AccountName)
 	require.NotNil(
@@ -167,6 +172,56 @@ func testAccountManagerCreateAccount(h *bwtest.HarnessTest) {
 	require.Equal(h, want, durableInfo)
 }
 
+// testAccountManagerCreateExactAccount verifies an exact account's disabled
+// chain synchronization and complete account facts survive reopening.
+func testAccountManagerCreateExactAccount(h *bwtest.HarnessTest) {
+	// Exact selection is SQL-only; the sequential case covers kvdb.
+	if *dbBackend != string(wallet.DBBackendSQLite) &&
+		*dbBackend != string(wallet.DBBackendPostgres) {
+
+		h.Skip("exact account creation requires SQL")
+	}
+
+	// Arrange: a sparse, excluded account exercises exact activation; the
+	// harness owns the unlocked wallet and its cleanup.
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	number := wallet.AccountNumber(7)
+	params := wallet.NewAccountParams{
+		Scope:         waddrmgr.KeyScopeBIP0084,
+		Name:          "sparse exact",
+		AccountNumber: &number,
+		NoChainSync:   true,
+	}
+
+	// Act: create publicly through both derivation and storage.
+	created, err := w.NewAccount(h.Context(), params)
+
+	// Assert: compare exact identity and full facts after reopen,
+	// normalizing key prefixes with the existing comparison helper.
+	require.NoError(h, err)
+	require.Equal(h, number, *created.AccountNumber)
+	require.True(h, created.NoChainSync)
+	require.Zero(h, created.ExternalKeyCount)
+	require.Zero(h, created.InternalKeyCount)
+
+	// Account creation alone must leave child addresses unallocated.
+	addresses, err := w.ListAddresses(
+		h.Context(), params.Name, waddrmgr.WitnessPubKey,
+	)
+	require.NoError(h, err)
+	require.Empty(h, addresses)
+
+	want := *created
+	want.PublicKey = canonicalAccountKey(h, want.PublicKey)
+	w = h.ReloadWallet(w)
+	durable, err := w.GetAccount(
+		h.Context(), params.Scope, params.Name,
+	)
+	require.NoError(h, err)
+	durable.PublicKey = canonicalAccountKey(h, durable.PublicKey)
+	require.Equal(h, want, *durable)
+}
+
 // testAccountManagerFillAccountHole verifies that reopening after sparse
 // allocation leaves lower exact account numbers available.
 func testAccountManagerFillAccountHole(h *bwtest.HarnessTest) {
@@ -201,6 +256,48 @@ func testAccountManagerFillAccountHole(h *bwtest.HarnessTest) {
 	// Assert: earlier sparse creation did not consume the requested hole.
 	require.NoError(h, err)
 	require.Equal(h, hole, *filled.AccountNumber)
+}
+
+// testAccountManagerAdvanceAccountCursor checks durable sparse allocation.
+func testAccountManagerAdvanceAccountCursor(h *bwtest.HarnessTest) {
+	// kvdb rejects exact requests, so this SQL cursor contract is inapplicable.
+	if *dbBackend != string(wallet.DBBackendSQLite) &&
+		*dbBackend != string(wallet.DBBackendPostgres) {
+
+		h.Skip("exact account creation requires SQL")
+	}
+
+	// Arrange: persist excluded seven and ordinary two, then reopen so the
+	// next request observes only the durable cursor left by both writes.
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	number := wallet.AccountNumber(7)
+	params := wallet.NewAccountParams{
+		Scope:         waddrmgr.KeyScopeBIP0084,
+		Name:          "excluded sparse",
+		AccountNumber: &number,
+		NoChainSync:   true,
+	}
+	_, err := w.NewAccount(h.Context(), params)
+	require.NoError(h, err)
+
+	hole := wallet.AccountNumber(2)
+	params.Name = "lower hole"
+	params.AccountNumber = &hole
+	params.NoChainSync = false
+	_, err = w.NewAccount(h.Context(), params)
+	require.NoError(h, err)
+	w = h.ReloadWallet(w)
+	h.UnlockWallet(w)
+
+	params.Name = "next sequential"
+	params.AccountNumber = nil
+
+	// Act: allocate the next account without an exact selector.
+	next, err := w.NewAccount(h.Context(), params)
+
+	// Assert: the cursor advanced past seven and did not retreat for two.
+	require.NoError(h, err)
+	require.Equal(h, wallet.AccountNumber(8), *next.AccountNumber)
 }
 
 // testAccountManagerCreateAccountSequence verifies that derived account numbers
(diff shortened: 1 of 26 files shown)


wallet/internal/sql/pg/queries/key_scopes.sql
```diff
@@ -58,6 +58,17 @@ SET next_account_number = next_account_number + 1
 WHERE id = $1
 RETURNING (next_account_number - 1)::BIGINT AS account_number;
 
+-- name: AdvanceNextAccountNumber :exec
+-- Reserves the cursor past an exact account without consuming lower holes.
+-- Updating this row serializes exact and sequential allocation in the caller's
+-- account transaction, including when the cursor already exceeds the minimum.
+UPDATE key_scopes
```
yyforyongyu, line 65: merge `wallet+db: generate account cursor bindings` with this commit

wallet/internal/db/err/account_test.go
```diff
@@ -72,3 +72,72 @@ func TestIsAccountNameConflict(t *testing.T) {
...
+// unrelated uniqueness failures using real SQLite and PostgreSQL diagnostics.
+func TestIsAccountNumberConflict(t *testing.T) {
+	t.Parallel()
+
+	// Arrange: provoke SQLite's actual number constraint and pair it with
+	// PostgreSQL metadata; plain diagnostic text must not gain SQL identity.
+	conn, err := sql.Open("sqlite", ":memory:")
+	require.NoError(t, err)
+	conn.SetMaxOpenConns(1)
+	t.Cleanup(func() { require.NoError(t, conn.Close()) })
```
yyforyongyu: use multi-line

wallet/internal/db/err/account_test.go
```diff
@@ -72,3 +72,72 @@ func TestIsAccountNameConflict(t *testing.T) {
...
+	_, err = conn.ExecContext(t.Context(), `
```
yyforyongyu: add a newline above, also explain why it's necessary to use raw sql here as the current store doesn't have the necessary methods or?

wallet/internal/db/itest/accountstore_createderived_test.go
```diff
@@ -501,3 +502,98 @@ func TestCreateDerivedAccountConcurrent(t *testing.T) {
 		require.Equal(t, uint32(i), results[i])
 	}
 }
+
+// TestCreateDerivedAccountExactConcurrent verifies a sparse first account and
+// competing exact/sequential requests share one monotonic allocation cursor.
+func TestCreateDerivedAccountExactConcurrent(t *testing.T) {
```
yyforyongyu, line 505: Unnecessary test?

wallet/internal/db/itest/accountstore_createderived_test.go
```diff
@@ -597,3 +599,98 @@ func TestCreateDerivedAccountExactConcurrent(t *testing.T) {
...
+
+		if test.want != nil {
+			require.ErrorIs(t, createErr, test.want)
+		} else {
+			requireConstraintSQLError(t, createErr)
+		}
+
+		var accounts, secrets, addresses int
+
+		err = store.DB().QueryRowContext(t.Context(), `
```
yyforyongyu: add inline comments explaining why we need raw sql here? why do we even need this? can we remove it?

itest/account_manager_test.go
```diff
@@ -102,20 +102,44 @@ func deterministicImportedAccountKeys(
...
 	ctx := h.Context()
 	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
 
-	created, err := w.NewAccount(ctx, wallet.NewAccountParams{
+	params := wallet.NewAccountParams{
 		Scope: scope,
 		Name:  accountName,
-	})
+	}
+	if *dbBackend != "kvdb" {
```
yyforyongyu: add a newline above, then add inline comments explaining why this special case for kvdb

itest/account_manager_test.go
```diff
@@ -165,6 +190,27 @@ func testAccountManagerCreateAccount(h *bwtest.HarnessTest) {
 	durableInfo := *durable
 	durableInfo.PublicKey = canonicalAccountKey(h, durableInfo.PublicKey)
 	require.Equal(h, want, durableInfo)
+
+	if params.AccountNumber == nil {
+		return
```
yyforyongyu: explain this early exit in doc?

wallet/account_manager.go
```diff
@@ -420,17 +425,27 @@ func (w *Wallet) accountInfoFromStore(
...
+// ErrAccountOperationUnsupported after admission and before secret preparation.
+// Failures return no account; ErrIndeterminateCommit means persistence may
+// have succeeded.
 func (w *Wallet) NewAccount(ctx context.Context,
 	params NewAccountParams) (*AccountInfo, error) {
 
-	err := w.validateNewAccountRequest(ctx, params.Scope, params.Name)
+	// Build one request for admission and persistence so path validation uses
+	// the same scope and optional number that the Store will allocate.
+	storeParams := db.CreateDerivedAccountParams{
```
yyforyongyu: just call it `dbParams`???

wallet/account_manager.go
```diff
@@ -442,19 +457,19 @@ func (w *Wallet) NewAccount(ctx context.Context,
 			ErrAccountOperationUnsupported)
 	}
 
+	// Wallet assembly supplies addrStore only for kvdb, whose allocator is
+	// sequential. Reject exact selection before loading or decrypting the root.
+	if params.AccountNumber != nil && w.addrStore != nil {
```
yyforyongyu: why do we have this check here? isn't this leaking the kvdb context into the upper layer method?

wallet/account_manager_test.go
```diff
@@ -2091,3 +2091,82 @@ func TestExtractAddrFromPKScript(t *testing.T) {
...
+		Name:          testAccountName,
+		AccountNumber: &number,
+	})
+
+	// Assert: validation exposes only the public identity and no account,
+	// without making a secret or Store call before refusing the request.
+	require.ErrorIs(t, err, ErrInvalidParam)
+	require.NotErrorIs(t, err, db.ErrInvalidParam)
+	require.Nil(t, info)
+	deps.store.AssertExpectations(t)
```
yyforyongyu: there is no need to assert `AssertExpectations`?? all these have been handled already in the test setup via `t.Cleanup`

wallet/internal/sql/pg/queries/key_scopes.sql
yyforyongyu: Combined the SQL query and its generated bindings in one commit. The remaining commits keep tests with the behavior they introduce.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

wallet/internal/db/err/account_test.go
yyforyongyu: Expanded the cleanup closure onto separate lines.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

wallet/internal/db/err/account_test.go
yyforyongyu: Added the spacing and explanation. This error-classification package has no Store dependency; the fixture needs a genuine SQLite constraint error from the driver.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

wallet/internal/db/itest/accountstore_createderived_test.go
yyforyongyu: Keeping this one because the neighboring test races only sequential requests. This test mixes exact and sequential allocation and verifies unique numbers plus the monotonic cursor, which the sequential-only test cannot establish. I removed the duplicate standalone activation scenario instead and folded its policy/reload checks into the existing public creation test.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

wallet/internal/db/itest/accountstore_createderived_test.go
yyforyongyu: Kept physical account and account_secrets counts, with an explanation: normal account reads can hide orphan secrets after a failed transaction, and the Store has no independent row-count API. Removed the redundant addresses count; account creation writes no addresses, and the public creation test already checks the empty initial list.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

itest/account_manager_test.go
yyforyongyu: Added a blank line and a comment explaining that kvdb supports only sequential creation, so only SQL prepares sparse exact allocation.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

itest/account_manager_test.go
yyforyongyu: Documented that kvdb finishes after ordinary creation and reload; the remaining lower-hole and cursor assertions apply only to SQL.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

wallet/account_manager.go
yyforyongyu: Renamed the local value to dbParams.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

wallet/account_manager.go
yyforyongyu: The backend-specific check is needed here to refuse exact kvdb creation before loading or decrypting the root. buildAccountDeriveFn eagerly prepares secrets before the Store call, so relying on the Store refusal alone would be too late. I kept this existing assembly signal rather than adding a capability interface or changing the derivation lifecycle solely for this check.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

wallet/account_manager_test.go
yyforyongyu: Removed the four task-added duplicate AssertExpectations calls. The shared t.Cleanup already verifies these mocks directly; the strict On/Return/Once expectations remain.

Verified and published in 4ab348a85ad8628e65790053c9b6d0c0164b5ecd.

wallet/internal/db/err/errors.go
```diff
@@ -262,6 +262,10 @@ const (
 	// ConstraintAccountName identifies a duplicate account name within a
 	// wallet and key scope, not an account-number or other unique violation.
 	ConstraintAccountName
+
+	// ConstraintAccountNumber identifies an occupied derived-account number
+	// within a key scope, independently of the account name.
+	ConstraintAccountNumber
 )
```
GustavoStingelin, line 269: So, are we gonna need to merge https://github.com/btcsuite/btcwallet/pull/1375 first?

yyforyongyu: yeah and rebased!

itest/account_manager_test.go
```diff
@@ -167,6 +172,134 @@ func testAccountManagerCreateAccount(h *bwtest.HarnessTest) {
...
+
+	params.Name = "next sequential"
+	params.AccountNumber = nil
+
+	// Act: allocate the next account without an exact selector.
+	next, err := w.NewAccount(h.Context(), params)
+
+	// Assert: the cursor advanced past seven and did not retreat for two.
+	require.NoError(h, err)
+	require.Equal(h, wallet.AccountNumber(8), *next.AccountNumber)
```
GustavoStingelin, line 300: q: what's the expected recovery strategy for clients using sparse exact accounts? Since recovery starts from persisted accounts, are clients expected to retain or deterministically reconstruct the account numbers and recreate those accounts before scanning during a seed restore?

Without that information, an account gap limit could miss funded accounts beyond a large hole. Widening the search also increases candidate-address work as `accounts searched × addresses per account` (e.g. 1,000 × 1,000 = one million candidates). Should we document the expected recovery contract alongside exact account creation?

GustavoStingelin (approved):
LGTM!

I've just left a question about the recovery.

itest/account_manager_test.go
yyforyongyu: At the account level that is never promised, even for accounts before this PR, as documented on the `Rescan`, the intended behavior is,

1. Restore the wallet root from the seed.
2. Recreate desired root-derived accounts using NewAccount with their scope and exact number, or import supplied account keys.
3. Rescan the eligible accounts.

PR #1370, opened 2026-09-07
wallet: activate exact no-chain-sync accounts

yyforyongyu:
## Summary

Implements roadmap Task 468.

## Change Description

Allow NoChainSync=true for exact root-derived SQL accounts through NewAccountParams. Reuse atomic account creation to persist and return the policy without deriving receiving addresses or admitting positive-lookahead recovery. Preserve sequential and modern kvdb refusal.

Wallet tests, the nine chain/database activation combinations, existing creation/refusal cases, lint, and commit hygiene pass.

## Notes

Complete dependency stack:

- #1365 (Task 345, persisted account policy) and #1358 (Task 378, account errors) feed #1366 (Task 392, creation request).
- #1366 feeds #1369 (Task 343, exact allocation) and #1367 (Task 393, receiving rejection).
- #1365 also feeds #1368 (Task 452, recovery exclusion).
- This PR depends on #1369, #1367, and #1368.

The base branch combines those three reviewed prerequisite heads. Review scope starts after all dependencies and contains only this PR's two activation/test commits. Keep this PR draft until its prerequisite stack is ready; rebase the task-only commits as the dependencies land.

### DIFF
--- a/itest/list_on_test.go
+++ b/itest/list_on_test.go
@@ -44,6 +44,10 @@ var allTestCases = []*testCase{
 		TestFunc: testAccountManagerCreateAccount,
 	},
 	{
+		Name:     "account manager activate no chain sync account",
+		TestFunc: testAccountManagerActivateNoChainSyncAccount,
+	},
+	{
 		Name:     "account manager create account sequence",
 		TestFunc: testAccountManagerCreateAccountSequence,
 	},
--- a/wallet/account_manager.go
+++ b/wallet/account_manager.go
@@ -243,7 +243,8 @@ type NewAccountParams struct {
 	AccountNumber *AccountNumber
 
 	// NoChainSync requests exclusion from automatic chain synchronization.
-	// True is currently rejected with ErrAccountOperationUnsupported.
+	// True requires exact SQL selection; other requests return
+	// ErrAccountOperationUnsupported.
 	NoChainSync bool
 }
 
@@ -294,7 +295,8 @@ type NewAccountParams struct {
 type AccountManager interface {
 	// NewAccount creates the next or requested exact root-derived account. The
 	// provided name must be unique within that key scope. NoChainSync=true
-	// is currently rejected with ErrAccountOperationUnsupported.
+	// is supported only for exact SQL selection; other requests return
+	// ErrAccountOperationUnsupported.
 	NewAccount(ctx context.Context, params NewAccountParams) (*AccountInfo,
 		error)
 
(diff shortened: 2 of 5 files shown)


itest/list_on_test.go
```diff
@@ -43,6 +43,11 @@ var allTestCases = []*testCase{
 		Name:     "account manager create account",
 		TestFunc: testAccountManagerCreateAccount,
 	},
+	// Activation has a distinct SQL policy contract from ordinary creation.
```
yyforyongyu: remove this redundant comment

wallet/account_manager.go
```diff
@@ -450,9 +453,11 @@ func (w *Wallet) NewAccount(ctx context.Context,
 		return nil, err
 	}
 
-	// Keep exclusion unavailable until receiving and recovery honor it.
-	// Refuse here so no backend prepares secrets or mutates account state.
-	if params.NoChainSync {
+	// Exclusion is supported only by exact SQL creation. Refuse other
+	// requests before preparing secrets or changing persistent account state.
+	if params.NoChainSync &&
+		(params.AccountNumber == nil || w.addrStore != nil) {
```
yyforyongyu: why do we need to check `w.addrStore`? we should remove it and let the wallet panic! and why `AccountNumber` must be nil?

wallet/account_manager_test.go
```diff
@@ -734,34 +735,58 @@ func TestNewAccount(t *testing.T) {
...
+	tests := []struct {
+		name      string
+		addrStore waddrmgr.AddrStore
+	}{
+		{
+			name: "sql sequential",
+		},
+		{
+			name:      "kvdb sequential",
+			addrStore: legacyStore,
```
yyforyongyu: this test is unnecessary

itest/list_on_test.go
yyforyongyu: Removed the redundant registration comment in 3bc1f959672a8eaac7b58ec63dc8a78de5a753ba.

wallet/account_manager.go
yyforyongyu: Removed the redundant `w.addrStore` check from this rule; the existing backend guard still handles unsupported kvdb creation. Moved the account-number requirement into `validateNewAccountRequest` and added an inline explanation: when an account does not watch for on-chain synchronization, its account number must be specified because sequential allocation is unsupported. Updated in 4f229e472f2f59621ee1741f1bd2e830032466e4.

wallet/account_manager_test.go
yyforyongyu: Removed the duplicate kvdb sequential case, its table, and the unused legacy mock in 4f229e472f2f59621ee1741f1bd2e830032466e4. Kept the single public regression test that verifies a no-chain-sync request without an account number is rejected before secret preparation or account mutation.

yyforyongyu:
Superseded by #1369, which now includes both exact account creation and NoChainSync activation. The two activation commits were replayed without patch changes as d6330b6a78b6 and 8e6d84711137. Further review and CI follow-up continue on #1369.

PR #1371, opened 2026-09-07
wallet: create accounts in custom scopes

yyforyongyu:
## Summary

Implements roadmap Task 390: create exact root-derived accounts in custom SQL scopes.

## Change Description

Add an optional address schema to account creation, validate it before secret preparation, and reuse existing scope metadata and atomic creation. Preserve kvdb rejection and ambiguous-commit behavior.

Verified unit and race tests, SQLite/PostgreSQL database tests, all nine chain/database integration combinations, and lint.

## Notes

Depends on #1369, which consolidates exact-account creation and no-chain-sync activation. Targets `task-exact-account`; review scope begins after that dependency and includes only the four custom-scope commits.

### DIFF
--- a/wallet/account_manager_test.go
+++ b/wallet/account_manager_test.go
@@ -2065,6 +2065,10 @@ func TestNewAccountInvalidPath(t *testing.T) {
 		Scope:         waddrmgr.KeyScopeBIP0084,
 		Name:          testAccountName,
 		AccountNumber: &number,
+		AddrSchema: &waddrmgr.ScopeAddrSchema{
+			ExternalAddrType: waddrmgr.WitnessPubKey,
+			InternalAddrType: waddrmgr.WitnessPubKey,
+		},
 	})
 
 	// Assert: validation exposes only the public identity and no account,
@@ -2098,7 +2102,18 @@ func TestNewAccountExactUnsupported(t *testing.T) {
 			t.Parallel()
 
 			w, deps := createUnlockedWalletWithMocks(t)
-			scope := waddrmgr.KeyScopeBIP0084
+			scope := waddrmgr.KeyScope{
+				Purpose: 1017,
+			}
+			schema := waddrmgr.ScopeAddrSchema{
+				ExternalAddrType: waddrmgr.WitnessPubKey,
+				InternalAddrType: waddrmgr.WitnessPubKey,
+			}
+			deps.store.On("GetKeyScopeSchema", mock.Anything, w.id,
+				db.KeyScope(scope)).Return(
+				db.ScopeAddrSchema{}, db.ErrKeyScopeNotFound,
+			).Once()
+
 			number := AccountNumber(7)
 
 			expectAccountNameAvailable(deps, scope, testAccountName)
@@ -2108,6 +2123,7 @@ func TestNewAccountExactUnsupported(t *testing.T) {
 				Scope:         scope,
 				Name:          testAccountName,
 				AccountNumber: &number,
+				AddrSchema:    &schema,
 				NoChainSync:   test.noChainSync,
 			})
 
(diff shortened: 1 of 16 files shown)


wallet/account_manager_test.go
```diff
@@ -2109,13 +2109,19 @@ func TestNewAccountInvalidPath(t *testing.T) {
...
+			InternalAddrType: waddrmgr.WitnessPubKey,
+		},
 	})
 
 	// Assert: validation exposes only the public identity and no account,
 	// without making a secret or Store call before refusing the request.
 	require.ErrorIs(t, err, ErrInvalidParam)
 	require.NotErrorIs(t, err, db.ErrInvalidParam)
 	require.Nil(t, info)
+	deps.store.AssertExpectations(t)
```
yyforyongyu: there is no need for this assertions? these mockers have already been asserted during their setup via `t.Cleanup`!

wallet/account_manager_test.go
```diff
@@ -2221,3 +2241,236 @@ func TestNewAccountCustomScopeMissingFields(t *testing.T) {
...
+					InternalAddrType: test.internal,
+				},
+			})
+
+			// Assert: only the scope read occurs; no root preparation or
+			// mutation can establish the incompatible account.
+			require.ErrorIs(t, err, ErrInvalidParam)
+			require.Nil(t, info)
+			deps.store.AssertExpectations(t)
+			deps.vault.AssertExpectations(t)
```
yyforyongyu: all these `AssertExpectations` should be removed!

wallet/account_manager_test.go
yyforyongyu: Removed the duplicate Store/Vault assertions and the unused deps binding. The fixture already verifies both mocks through t.Cleanup. Fixed in e487af9d9; focused tests, race checks, and the wallet package pass.

wallet/account_manager_test.go
yyforyongyu: Removed all 12 task-added AssertExpectations calls across the six wallet tests. Existing fixture cleanup, Arrange expectations, and Once counts remain unchanged. The removals are folded into their introducing commits and pushed in e487af9d9. Tests, race checks, and lint pass.

wallet/account_manager_test.go
```diff
@@ -2155,3 +2201,345 @@ func TestNewAccountCancellationPreservesCommitError(t *testing.T) {
...
+	<-w.lifetimeCtx.Done()
+
+	// Assert: observe that Stop cannot finish before the read returns, then
+	// release it and check the admitted request's ordinary missing-input error.
+	select {
+	case err := <-stopped:
+		t.Fatalf("Stop returned before scope validation: %v", err)
+	case <-time.After(20 * time.Millisecond):
+	}
+
```
GustavoStingelin, line 2255: Could we replace the 20 ms wait with deterministic synchronization? Waiting for `lifetimeCtx.Done()` confirms shutdown started, but a scheduling delay could keep a broken `Stop` from returning within that window and let the test pass.

For example, put the mock fixture and test goroutines inside `synctest.Test`, then use `synctest.Wait()` before checking `stopped`:

```go
synctest.Test(t, func(t *testing.T) {
	// Existing fixture setup and blocked scope read go here.

	stopped := make(chan error, 1)
	go func() { stopped <- w.Stop(t.Context()) }()

	<-w.lifetimeCtx.Done()
	synctest.Wait() // Other goroutines have exited or durably blocked.

	select {
	case err := <-stopped:
		t.Fatalf("Stop returned before scope validation: %v", err)
	default:
	}

	resume()

	response := <-result
	require.ErrorIs(t, response.err, ErrInvalidParam)
	require.Nil(t, response.info)
	require.NoError(t, <-stopped)
})
```

This would check that shutdown actually blocks on the pending request rather than relying on elapsed wall-clock time.

wallet/internal/db/accountstore_createderived.go
```diff
@@ -299,3 +313,26 @@ func CreateDerivedAccountWithOps(ctx context.Context,
...
+	)
+	if err != nil {
+		return 0, ScopeAddrSchema{}, fmt.Errorf("ensure scope: %w", err)
+	}
+
+	if params.AddrSchema != nil && *params.AddrSchema != schema {
+		return 0, ScopeAddrSchema{}, fmt.Errorf(
+			"conflicting scope schema: %w", ErrInvalidParam,
+		)
+	}
```
GustavoStingelin: so, is it a test to see if EnsureScope correct created the scope? reaching this branch should be a bug

GustavoStingelin (approved):
LGTM!

PR #1373, opened 2026-09-07
wallet: route component requests through admission

yyforyongyu:
## Summary

Route maintained wallet component operations through the existing request admission and shutdown drain.

## Change Description

Keep handlers in their owning components and wait synchronously for every accepted operation, including through caller cancellation and shutdown. Reuse existing operations with their original inputs and keep nested work inside the accepted handler. Cover the shared completion contract and distinct mailbox, callback, key-delivery and partial-PSBT regressions, with repeated per-method lifecycle tests removed. Commits are grouped by complete component.

### DIFF
--- a/wallet/address_manager.go
+++ b/wallet/address_manager.go
@@ -469,6 +469,106 @@ func (w *Wallet) addrBalances(ctx context.Context) (map[string]btcutil.Amount,
 	return balances, nil
 }
 
+// newAddressReq retains value parameters until derivation and notification end.
+type newAddressReq struct {
+	reqCtx
+
+	accountName string
+	addrType    waddrmgr.AddressType
+	change      bool
+	respChan    chan addressResp
+}
+
+// getUnusedAddressReq keeps lookup and fallback derivation in one admission.
+type getUnusedAddressReq struct {
+	reqCtx
+
+	accountName string
+	addrType    waddrmgr.AddressType
+	change      bool
+	respChan    chan addressResp
+}
+
+// addressResp carries either address acquisition result to a buffered receiver.
+type addressResp struct {
+	addr address.Address
+	err  error
+}
+
+// getAddressInfoReq borrows the destination until the accepted lookup ends.
+type getAddressInfoReq struct {
+	reqCtx
+
+	addr     address.Address
+	respChan chan addressInfoResp
+}
+
+// addressInfoResp returns metadata once the accepted lookup completes.
+type addressInfoResp struct {
+	info AddressInfo
+	err  error
+}
+
+// listAddressesReq keeps balance and address iteration under one admission.
+type listAddressesReq struct {
+	reqCtx
+
+	accountName string
+	addrType    waddrmgr.AddressType
+	respChan    chan listAddressesResp
+}
+
+// listAddressesResp returns the list after balance and address iteration end.
+type listAddressesResp struct {
+	addresses []AddressProperty
+	err       error
+}
+
+// importPublicKeyReq borrows the key through persistence and registration.
+type importPublicKeyReq struct {
+	reqCtx
+
+	pubKey      *btcec.PublicKey
+	addrType    waddrmgr.AddressType
+	respErrChan chan error
+}
+
+// importTaprootScriptReq retains the supplied script through vault encryption.
+type importTaprootScriptReq struct {
+	reqCtx
+
+	tapscript waddrmgr.Tapscript
+	respChan  chan addressInfoResp
+}
+
+// scriptForOutputReq borrows script bytes until spending lookups complete.
+type scriptForOutputReq struct {
+	reqCtx
+
+	output   wire.TxOut
+	respChan chan outputScriptResp
+}
+
+// outputScriptResp returns spending metadata after dependency access ends.
+type outputScriptResp struct {
+	info OutputScriptInfo
+	err  error
+}
+
+// getDerivationInfoReq borrows its destination until path lookup completes.
+type getDerivationInfoReq struct {
+	reqCtx
+
+	addr     address.Address
+	respChan chan derivationInfoResp
+}
+
+// derivationInfoResp preserves a path result across cancellation and shutdown.
+type derivationInfoResp struct {
+	info *psbt.Bip32Derivation
+	err  error
+}
+
 // NewAddress returns a new address for the given account and address type.
 // This method is a low-level primitive that will always derive a new, unused
 // address from the end of the address chain.
@@ -532,6 +632,39 @@ func (w *Wallet) NewAddress(ctx context.Context, accountName string,
 	if err != nil {
 		return nil, err
 	}
+
+	// Admission keeps dependency access joined through concurrent Stop.
+	r := newAddressReq{
+		reqCtx:      reqCtx{ctx: ctx},
+		accountName: accountName,
+		addrType:    addrType,
+		change:      change,
+		respChan:    make(chan addressResp, 1),
+	}
+
+	err = w.sendReq(ctx, r)
+	if err != nil {
+		return nil, err
+	}
+
+	// Once admitted, wait for the result even if cancellation arrives.
+	result := <-r.respChan
+
+	return result.addr, result.err
+}
+
+// handleNewAddress delivers the result of an accepted component request.
+func (w *Wallet) handleNewAddress(r newAddressReq) {
+	// Reuse address derivation shared with the unused-address fallback.
+	addr, err := w.newAddress(
+		r.ctx, r.accountName, r.addrType, r.change,
+	)
+	r.respChan <- addressResp{addr: addr, err: err}
+}
+
+// newAddress derives and registers an address inside its accepted request.
+func (w *Wallet) newAddress(ctx context.Context, accountName string,
+	addrType waddrmgr.AddressType, change bool) (address.Address, error) {
 
 	// Addresses cannot be derived from the catch-all imported accounts.
 	if accountName == waddrmgr.ImportedAddrAccountName {
@@ -613,49 +746,90 @@ func (w *Wallet) GetUnusedAddress(ctx context.Context, accountName string,
 		return nil, err
 	}
 
-	if accountName == waddrmgr.ImportedAddrAccountName {
-		return nil, ErrImportedAccountNoAddrGen
+	// Admission keeps dependency access joined through concurrent Stop.
+	r := getUnusedAddressReq{
+		reqCtx:      reqCtx{ctx: ctx},
+		accountName: accountName,
+		addrType:    addrType,
+		change:      change,
+		respChan:    make(chan addressResp, 1),
 	}
 
-	keyScope, err := addrType.KeyScope()
-	if err != nil {
-		return nil, fmt.Errorf("%w: %v", ErrUnknownAddrType, addrType)
-	}
-
-	req, err := addressPageRequest()
+	err = w.sendReq(ctx, r)
 	if err != nil {
 		return nil, err
 	}
 
+	// Once admitted, wait for the result even if cancellation arrives.
+	result := <-r.respChan
+
+	return result.addr, result.err
+}
+
+// handleGetUnusedAddress reuses or derives an address without a second
+// admission.
+// The admitted caller waits for this result before reusing its inputs.
+func (w *Wallet) handleGetUnusedAddress(r getUnusedAddressReq) {
+	if r.accountName == waddrmgr.ImportedAddrAccountName {
+		r.respChan <- addressResp{err: ErrImportedAccountNoAddrGen}
+
+		return
+	}
+
+	keyScope, err := r.addrType.KeyScope()
+	if err != nil {
+		r.respChan <- addressResp{
+			err: fmt.Errorf("%w: %v", ErrUnknownAddrType, r.addrType),
+		}
+
+		return
+	}
+
+	req, err := addressPageRequest()
+	if err != nil {
+		r.respChan <- addressResp{err: err}
+
+		return
+	}
+
 	addresses := w.store.IterAddresses(
-		ctx, db.ListAddressesQuery{
+		r.ctx, db.ListAddressesQuery{
 			WalletID:    w.id,
-			AccountName: &accountName,
+			AccountName: &r.accountName,
 			Scope:       (*db.KeyScope)(&keyScope),
 			Page:        req,
 		},
 	)
 	for storeAddr, err := range addresses {
 		if err != nil {
-			return nil, err
+			r.respChan <- addressResp{err: err}
+
+			return
 		}
 
 		unusedAddr, ok, err := nextUnusedStoreAddress(
-			storeAddr, change, w.cfg.ChainParams,
+			storeAddr, r.change, w.cfg.ChainParams,
 		)
 		if err != nil {
-			return nil, err
+			r.respChan <- addressResp{err: err}
+
+			return
 		}
 
 		if !ok {
 			continue
 		}
 
-		return unusedAddr, nil
+		r.respChan <- addressResp{addr: unusedAddr}
+
+		return
 	}
 
 	// Otherwise, we'll generate a new one.
-	return w.NewAddress(ctx, accountName, addrType, change)
+	responseAddr, responseErr := w.newAddress(
+		r.ctx, r.accountName, r.addrType, r.change,
+	)
+	r.respChan <- addressResp{addr: responseAddr, err: responseErr}
 }
 
 // nextUnusedStoreAddress returns the unused address candidate represented by a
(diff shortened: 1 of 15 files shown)


wallet/common_test.go
```diff
@@ -163,6 +163,21 @@ func testSQLManager(tb testing.TB, store db.Store) *Manager {
 func startLoadedWalletForTest(t *testing.T, w *Wallet) {
 	t.Helper()
 
+	// Hand-built storage fixtures need the same request runtime as loaded
+	// Wallets. Preserve any existing runtime so their configured seams survive.
+	if w.lifetimeCtx == nil {
```
yyforyongyu: why do we have these unnecessary nil checks? the tests should just init them?

wallet/controller.go
```diff
@@ -682,6 +760,11 @@ func (w *Wallet) handleReq(req any) {
 	defer w.wg.Done()
 
 	switch r := req.(type) {
+	case infoReq:
+		info, err := w.info(r.ctx)
+		r.resp <- infoResp{info: info, err: err}
+	case rescanReq:
+		r.resp <- w.handleRescanReq(r)
```
yyforyongyu: can you respect the current convention? we just call `w.handleXXX` here and send the result inside that method?

wallet/controller_test.go
```diff
@@ -1434,8 +1699,7 @@ func TestControllerResync(t *testing.T) {
 		t.Parallel()
 
 		w, deps := createTestWalletWithMocks(t)
-		require.NoError(t, w.state.toStarting())
-		require.NoError(t, w.state.toStarted())
+		startLoadedWalletForTest(t, w)
```
yyforyongyu, line 1605: why do we make this change? seems unnecessary?

wallet/tx_creator.go
```diff
@@ -640,12 +641,91 @@ func normalizeAndValidateTxIntent(intent *TxIntent) error {
...
+}
+
+// authoredTxResp retains the completed transaction after caller cancellation.
+type authoredTxResp struct {
+	tx  *txauthor.AuthoredTx
+	err error
+}
+
+// snapshotInputsPolicy owns supported coin sources and retains the selector.
+func snapshotInputsPolicy(policy *InputsPolicy) *InputsPolicy {
```
yyforyongyu: why this change?

wallet/controller.go
```diff
@@ -760,7 +760,7 @@ type reqCtx struct {
 // handleReq owns completion bookkeeping for requests accepted by mainLoop. Its
 // type switch is the single routing table for concurrent public method work.
 //
-//nolint:cyclop,funlen // Enumerate all request routes in one table.
+//nolint:cyclop,gocyclo,funlen // Enumerate all request routes in one table.
```
yyforyongyu, line 769: this should be folded into `wallet: route transaction reader requests`?

wallet/common_test.go
yyforyongyu: Removed the nil guards from startLoadedWalletForTest. The four incomplete fixtures now initialize their lifetime context/cancel pair, request channel, and stopped lock timer explicitly before starting the loop. The helper only starts and joins the initialized runtime.

wallet/controller.go
yyforyongyu: Updated all 32 non-account request cases to call w.handleXXX(r), with response delivery owned by the handler in its component file. Single-use operational bodies became handlers directly; shared operations and the signing operations with deferred key zeroing remain reused. handleReq retains only dispatch and joined completion bookkeeping.

wallet/controller_test.go
yyforyongyu: The request-loop setup is required now that Resync sends through the unbuffered request channel. The previous state transitions did not start a receiver, so the test would hang at admission. Reverting this helper call reproduces a timeout in TestControllerResync/Success. I kept the loop setup; its cleanup cancels and joins the accepted work.

wallet/tx_creator.go
yyforyongyu: CreateTransaction and FundPsbt may finish accepted work after the caller cancels and returns. The snapshot owns the policy pointers and UTXO slices so the caller can safely reuse its input while that work drains. TestTxCreatorSnapshotsIntent covers this regression, including policy inputs. Callback-bearing selection strategies still retain their synchronous lifetime.

wallet/controller.go
yyforyongyu: Kept gocyclo in the signer commit: handleReq has complexity 23 at the transaction-reader commit and first exceeds gocyclo’s threshold of 30 in the signer commit (34). Adding it to the reader commit would introduce an unused exemption. I also adjusted the existing exemptions to their first required commit after shortening the dispatcher and removed the two contextcheck directives that became unused.

wallet/controller.go
```diff
@@ -568,6 +621,12 @@ func (w *Wallet) Resync(ctx context.Context, startHeight uint32) error {
 func (w *Wallet) Rescan(ctx context.Context, startHeight uint32,
 	targets []waddrmgr.AccountScope) error {
 
+	// Terminal rejection must precede argument validation for retained Wallets.
+	err := w.state.validateStarted()
```
yyforyongyu: this fix belongs to a different commit?

wallet/controller.go
```diff
@@ -510,6 +512,27 @@ func (w *Wallet) ChangePassphrase(ctx context.Context,
...
+	info *Info
+	err  error
+}
+
+// rescanReq owns the target list until the joined syncer accepts the scan.
+type rescanReq struct {
+	typ         scanType
+	startHeight uint32
+	targets     []waddrmgr.AccountScope
+	resp        chan error
```
yyforyongyu: name it `respErrChan`

wallet/controller.go
```diff
@@ -510,6 +512,27 @@ func (w *Wallet) ChangePassphrase(ctx context.Context,
 	return reqErr
 }
 
+// infoReq keeps metadata work inside the accepted handler's Wallet lifetime.
+type infoReq struct {
+	reqCtx
+
+	resp chan infoResp
```
yyforyongyu: name it `respChan` - for all data types that are go channels, they should end with `xxxChan`

wallet/address_manager.go
```diff
@@ -469,6 +470,152 @@ func (w *Wallet) addrBalances(ctx context.Context) (map[string]btcutil.Amount,
...
 }
 
+// newAddressReq retains value parameters until derivation and notification end.
+type newAddressReq struct {
+	reqCtx
+
+	accountName string
+	addrType    waddrmgr.AddressType
+	change      bool
+	resp        chan addressResp
```
yyforyongyu: same here re the name, we should rename all resp to be respChan

wallet/address_manager.go
```diff
@@ -776,7 +1074,7 @@ func (w *Wallet) ListAddresses(ctx context.Context, accountName string,
 		})
 	}
 
-	return properties, nil
+	r.resp <- listAddressesResp{addresses: properties}
```
yyforyongyu: i am concerned about this direct sending, what if this is blocking? should we add a method on the wallet, sth like `sendResp` which is similar to `w.sendReq`?

wallet/address_manager.go
```diff
@@ -469,6 +470,152 @@ func (w *Wallet) addrBalances(ctx context.Context) (map[string]btcutil.Amount,
...
+}
+
+// derivationInfoResp preserves a path result across cancellation and shutdown.
+type derivationInfoResp struct {
+	info *psbt.Bip32Derivation
+	err  error
+}
+
+// snapshotAddress copies a supported destination without changing its encoding.
+func (w *Wallet) snapshotAddress(a address.Address) (address.Address, error) {
```
yyforyongyu: keep it simple and just name it `copyAddress`

wallet/address_manager.go
```diff
@@ -469,6 +470,152 @@ func (w *Wallet) addrBalances(ctx context.Context) (map[string]btcutil.Amount,
...
+	if err != nil {
+		return nil, fmt.Errorf("pay to addr script: %w", err)
+	}
+
+	// String preserves P2PK encoding; EncodeAddress would convert it to P2PKH.
+	return address.DecodeAddress(a.String(), w.cfg.ChainParams)
+}
+
+// snapshotTapscript copies the keys and byte slices an accepted import retains.
+func snapshotTapscript(t waddrmgr.Tapscript) waddrmgr.Tapscript {
```
yyforyongyu: keep it simple and just name it `copyTapscript`

wallet/controller.go
```diff
@@ -220,6 +220,9 @@ func (w *Wallet) Start(startCtx context.Context) error {
 	// 4. Start background goroutines.
 	w.wg.Add(1)
 
+	// Requests retain their own contexts; the routed legacy tip accessor still
+	// has no context parameter. Its existing exemption follows the call path.
+	//nolint:contextcheck // SyncedTo takes no context.
```
yyforyongyu, line 224: why do we need this? this comment plus nolint can be removed?

wallet/tx_creator.go
```diff
@@ -640,12 +641,91 @@ func normalizeAndValidateTxIntent(intent *TxIntent) error {
...
+			coins.UTXOs = slices.Clone(coins.UTXOs)
+			owned.Source = &coins
+		}
+	}
+
+	return &owned
+}
+
+// snapshotTxIntent owns the data authoring may use after a canceled return.
+func snapshotTxIntent(intent *TxIntent) *TxIntent {
```
yyforyongyu: remove these two snapshot methods - i don't think they are necessary - if we wanna keep the data immutable, we should fix it in the data, like using an interface or build a copy method on them, but i don't wanna expand the PR scope by introducing these. That being said, i think the script/addr snapshots can also be removed?

wallet/tx_creator.go
```diff
@@ -654,19 +734,64 @@ func (w *Wallet) CreateTransaction(ctx context.Context, intent *TxIntent) (
...
+	r := createTransactionReq{
+		reqCtx: reqCtx{ctx: ctx},
+		intent: snapshotTxIntent(intent),
+		resp:   make(chan authoredTxResp, 1),
+	}
+
+	// A supplied selector may close over caller state. Keep its existing call
+	// lifetime through completion while dependencies still use the caller ctx.
+	waitCtx := ctx
+	if r.intent != nil {
```
yyforyongyu: what is this check about? this is a behavior change?

wallet/tx_creator.go
```diff
@@ -654,19 +734,64 @@ func (w *Wallet) CreateTransaction(ctx context.Context, intent *TxIntent) (
...
+	result, err := waitForReq(waitCtx, r.resp)
 	if err != nil {
 		return nil, err
 	}
 
-	return w.authorTransaction(
-		intent.Outputs, intent.FeeRate, inputSource, changeSource,
+	// Preserve default normalization only after a completed response; a caller
+	// that already left through cancellation must receive no later mutation.
+	if intent != nil && intent.Inputs == nil {
```
yyforyongyu: why do we need this? when will these be mutated?

wallet/tx_publisher_test.go
```diff
@@ -33,13 +36,14 @@ var (
...
-// matchCreateTxParams returns a matcher for Store CreateTx parameters.
+// matchCreateTxParams returns a matcher for Store CreateTx parameters. The
+// witness hash compares transaction contents because admission owns a copy.
 func matchCreateTxParams(walletID uint32, tx *wire.MsgTx, label string,
 	credits map[uint32]address.Address) any {
 
 	return mock.MatchedBy(func(params db.CreateTxParams) bool {
 		return params.WalletID == walletID &&
-			params.Tx == tx &&
+			params.Tx != nil && params.Tx.WitnessHash() == tx.WitnessHash() &&
```
yyforyongyu: behavior change? why this fix?

wallet/signer.go
```diff
@@ -481,6 +482,161 @@ var _ SpendDetails = (*LegacySpendDetails)(nil)
...
+	if hashes != nil {
+		owned := *hashes
+		hashes = &owned
+	}
+
+	return tx, output, hashes
+}
+
+// snapshotSpendDetails copies known scripts and reports whether it owns them.
+func snapshotSpendDetails(details SpendDetails) (SpendDetails, bool) {
```
yyforyongyu: i think we should remove all these

wallet/controller.go
yyforyongyu: Removed the additional Rescan preflight. The existing submission path already checks state; empty-target validation keeps its prior order. The stopped-entry test now supplies a nonempty target.

wallet/controller.go
yyforyongyu: Renamed the error response channel to respErrChan, including request construction, handler sends and the synchronous receive.

wallet/controller.go
yyforyongyu: Renamed typed response channels to respChan and retained task-added local channels to names ending in Chan.

wallet/address_manager.go
yyforyongyu: Applied this throughout the component requests: respChan for typed results and respErrChan for error-only results.

wallet/address_manager.go
yyforyongyu: Each response channel has capacity one and each admitted handler sends exactly one result, so the send has space even before its caller receives. The caller now waits synchronously for that result after admission. Kept the direct send; no additional sendResp mechanism is needed.

wallet/address_manager.go
yyforyongyu: Removed snapshotAddress and its callers entirely under the clarified synchronous completion contract, so no replacement copy helper is needed.

wallet/address_manager.go
yyforyongyu: Removed snapshotTapscript and the import copy. Accepted imports now finish before the public call returns, using the supplied script.

wallet/controller.go
yyforyongyu: Kept this narrow contextcheck suppression: Start now reaches the routed UTXO/transaction handlers through mainLoop, and those existing operations call legacy SyncedTo, which has no context parameter. The linter follows that path. Changing the legacy accessor would expand this PR; the current package and repository lint checks pass.

wallet/tx_creator.go
yyforyongyu: Removed both transaction/policy snapshot helpers, along with task-added address, script, signer and PSBT input copies. All admitted component calls now wait for completion, as clarified for graceful shutdown. Existing authoring and cryptographic copies remain where the underlying operations already required them.

wallet/tx_creator.go
yyforyongyu: Removed the selector-specific wait context and check. All accepted authoring now waits synchronously, including supplied selectors; dependencies still receive the original caller context.

wallet/tx_creator.go
yyforyongyu: Removed the copy-back branch. The existing normalizer installs the default Inputs policy directly on the supplied intent, and CreateTransaction waits for completion. The retained default-policy test checks that behavior.

wallet/tx_publisher_test.go
yyforyongyu: Restored the original transaction pointer matcher. Broadcast now passes the supplied transaction directly, so the witness-hash workaround is unnecessary.

wallet/signer.go
yyforyongyu: Removed the signer snapshot helpers and task-added signing-input copies. Accepted signing waits through callbacks and key cleanup, and exported private keys are delivered to callers for zeroing. Retained only the distinct callback/key-delivery regression tests.

wallet/controller_test.go
```diff
@@ -1289,6 +1288,172 @@ func TestHandleChangePassphraseReq_Errors(t *testing.T) {
...
+			require.ErrorIs(t, result.err, tc.err)
+
+			if tc.err == nil {
+				require.Equal(t, "mock", result.info.Backend)
+			} else {
+				require.Nil(t, result.info)
+			}
+
+			require.NoError(t, <-stoppedChan)
+			deps.store.AssertExpectations(t)
```
yyforyongyu: can you make sure we don't do this? the test setup already performs the mock assertion, so we don't need `AssertExpectations`?

yyforyongyu: Removed all eight duplicate AssertExpectations calls added by this PR: four in the Controller test and two in each retained PSBT regression test. createTestWalletWithMocks already verifies those exact mocks in cleanup, after the request loop is joined. Independently created mocks retain their own checks. The branch-wide audit also removed nine redundant stopped-state tests and the obsolete nil-packet case, and simplified address/key-delivery coverage without changing production code.

GustavoStingelin (approved):
LGTM!

PR #1374, opened 2026-09-07
itest: cover public derivation and ECDH contracts

Abdulkbk:
Task 441 of the [#1250](https://github.com/btcsuite/btcwallet/issues/1250)
integration-test lane.

`Signer.DerivePubKey` and `Signer.ECDH` had one smoke case between them,
asserting only that a derived key was non-nil.

## Task 441 acceptance criteria mapped to tests

| Criterion | Case |
| --- | --- |
| Representative path, locked, watch-only, invalid-input and durable-reopen coverage for both methods | paths / wallet state / watchonly / reject request / durable reopen, and the four `signer ecdh ...` cases |
| ECDH independently verified without Store or private-key reads | `signer ecdh agreement` compares against `GenerateSharedSecret(peerPriv, DerivePubKey(...))` |
| Shared cases contain no backend skip branch | every case runs unchanged on all three databases; no case reads the backend |
| SQL imported-XPub behavior covered here | `signer derive imported xpub` (see the Task 361 note below) |

## The cases

Adds ten cases to `itest/signer_test.go`, all registered in the shared matrix
with no backend skip branch. Public keys are compared against an independent
`hdkeychain` derivation from the account XPub that `GetAccount` reports, and
the shared secret against the value a locally generated peer computes from
`DerivePubKey`'s output, so neither assertion depends on a store or
private-key read.

| Case | Contract |
| --- | --- |
| `signer derive pubkey paths` | Both exported selectors resolve one account; external and internal branches; the largest unhardened branch and index; the first external child backs the first address the wallet issued |
| `signer derive pubkey wallet state` | Unstarted rejects; locked succeeds; unlocking changes nothing |
| `signer derive pubkey reject request` | Six rejected requests as a data-only table: no account identity, unknown name, unknown number, unknown key scope, hardened branch, hardened index |
| `signer derive pubkey watchonly` | A wallet with no private material still derives |
| `signer ecdh agreement` | Both sides of the exchange agree; the secret is bound to peer key and path |
| `signer ecdh wallet state` | Unstarted and locked reject, returning the zero secret |
| `signer ecdh reject account` | Unknown account number and unknown key scope |
| `signer ecdh watchonly` | The public half stays available while the secret half is refused |
| `signer derivation durable reopen` | Both survive close and reload |
| `signer derive imported xpub` | A numberless account resolves by name, refuses a numbered selector, and stays resolvable after reopen |

The imported-account key fixture moves from `itest/account_manager_test.go` to
`itest/fixtures_test.go` so no component owns a value another one uses.

### One public API addition

`wallet.ErrInvalidAccountSelector` is exported (it was `errInvalidAccountSelector`).
A selector naming neither or both account identities is a public contract of
every operation taking one, but no caller outside the package could match it
except on the error's text. `itest/AGENTS.md` requires stable error contracts
to be matched with `errors.Is`, so the alternative was to leave that rejection
uncovered. This is the only production change in the PR.

### Deferred to #1376

Task 441 also asks for coverage of `reject nil or unsupported inputs without
exposing secret material`. That is not asserted here, because the wallet did
not reject those inputs: a nil peer key faulted on the goroutine `handleReq`
spawns, after the leaf private key had been derived. #1376 rejects nil
`Signer` arguments before admission, and covers the off-curve key that returned
an all-zero secret with no error. Those assertions belong with that fix rather
than with this lane.

[t361]: https://github.com/btcsuite/btcwallet/issues/1250

### Verification

- **Databases** (chain `btcd`): kvdb, SQLite, PostgreSQL — 10/10 each.
- **Chain backends** (db `sqlite`): btcd, neutrino — 10/10 each.
- Full suite passes on SQLite and kvdb, and with `-shuffleseed` set, so no case
  depends on registration order.
- Every commit builds and passes on its own.
- Each decisive assertion was checked by breaking its expected value and
  confirming the case fails for the stated reason.
- CI is green on this head (19/19), which additionally covers the `bitcoind`
  lanes and the lint and format checks.

### DIFF
--- a/itest/account_manager_test.go
+++ b/itest/account_manager_test.go
@@ -16,19 +16,6 @@ import (
 // every backend.
 const accountManagerFundingType = waddrmgr.WitnessPubKey
 
-// importedAccountKeys holds deterministic public and private account material
-// used to exercise imported-account contracts without sharing wallet state.
-// accountPrivateKey is present only so a case can assert private keys are
-// refused.
-type importedAccountKeys struct {
-	scope                waddrmgr.KeyScope
-	addrType             waddrmgr.AddressType
-	accountKey           *hdkeychain.ExtendedKey
-	otherAccountKey      *hdkeychain.ExtendedKey
-	accountPrivateKey    *hdkeychain.ExtendedKey
-	masterKeyFingerprint uint32
-}
-
 // canonicalAccountKey returns a derived account key under the network default
 // public version so backend-specific serialized versions compare consistently.
 func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
@@ -44,61 +31,6 @@ func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
 	)
 
 	return []byte(normalized.String())
-}
-
-// deterministicImportedAccountKeys derives distinct BIP84 account XPubs from a
-// fixed root so imported-account cases are reproducible.
-func deterministicImportedAccountKeys(
-	h *bwtest.HarnessTest) importedAccountKeys {
-
-	h.Helper()
-
-	scope := waddrmgr.KeyScopeBIP0084
-	root, err := hdkeychain.NewMaster(
-		[]byte{
-			0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08,
-			0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f, 0x10,
-		},
-		h.NetParams(),
-	)
-	require.NoError(h, err, "failed to derive deterministic root key")
-
-	defer root.Zero()
-
-	purpose, err := root.Derive(hdkeychain.HardenedKeyStart + scope.Purpose)
-	require.NoError(h, err, "failed to derive BIP84 purpose key")
-
-	defer purpose.Zero()
-
-	coinType, err := purpose.Derive(
-		hdkeychain.HardenedKeyStart + h.NetParams().HDCoinType,
-	)
-	require.NoError(h, err, "failed to derive BIP84 coin type key")
-
-	defer coinType.Zero()
-
-	// Neuter hands the public key the same chain code and parent
-	// fingerprint slices the private key holds, so the two account private
-	// keys below must outlive this call. Zeroing either one rewrites the
-	// XPub being returned.
-	accountPrivateKey, err := coinType.Derive(hdkeychain.HardenedKeyStart)
-	require.NoError(h, err, "failed to derive imported account private key")
-	accountKey, err := accountPrivateKey.Neuter()
-	require.NoError(h, err, "failed to derive imported account public key")
-
-	otherPrivateKey, err := coinType.Derive(hdkeychain.HardenedKeyStart + 1)
-	require.NoError(h, err, "failed to derive second imported private key")
-	otherAccountKey, err := otherPrivateKey.Neuter()
-	require.NoError(h, err, "failed to derive second imported public key")
-
-	return importedAccountKeys{
-		scope:                scope,
-		addrType:             waddrmgr.WitnessPubKey,
-		accountKey:           accountKey,
-		otherAccountKey:      otherAccountKey,
-		accountPrivateKey:    accountPrivateKey,
-		masterKeyFingerprint: purpose.ParentFingerprint(),
-	}
 }
 
 // testAccountManagerCreateAccount verifies that a new derived account's
--- a/itest/fixtures_test.go
+++ b/itest/fixtures_test.go
@@ -7,10 +7,14 @@ import (
 	"time"
 
 	"github.com/btcsuite/btcd/btcutil/v2"
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
 	"github.com/btcsuite/btcd/chainhash/v2"
 	"github.com/btcsuite/btcd/wire/v2"
+	"github.com/btcsuite/btcwallet/bwtest"
 	"github.com/btcsuite/btcwallet/pkg/btcunit"
+	"github.com/btcsuite/btcwallet/waddrmgr"
 	"github.com/btcsuite/btcwallet/wallet/txrules"
+	"github.com/stretchr/testify/require"
 )
 
 // Fixture values shared by more than one component's integration tests. They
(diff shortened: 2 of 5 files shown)


itest/signer_test.go
```diff
@@ -7,36 +7,744 @@
...
+func testSignerECDHRejectAccount(h *bwtest.HarnessTest) {
+	// absentAccountNumber is a BIP44 account number no test wallet creates.
+	const absentAccountNumber = 4242
+
+	// Arrange: A started, unlocked wallet holding only its default account.
+	ctx := h.Context()
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	h.NewWalletAddressOfType(w, signerAddrType)
+
+	peerKey, err := btcec.NewPrivateKey()
```
GustavoStingelin: [P2] Cover nil and invalid ECDH peer inputs

I’d add separate rejection cases with a valid account path and an unlocked wallet, passing nil and an invalid non-nil peer. Task 441 explicitly requires this, but every new ECDH case uses a valid generated key. On this head, I reproduced a nil-peer process panic and `new(btcec.PublicKey)` returning a zero secret with no error. #1376 already owns the production guards; these integration cases should exercise that contract across kvdb, SQLite, and PostgreSQL and assert a stable error plus a zero result.

itest/signer_test.go
```diff
@@ -7,36 +7,744 @@
...
+	internalPath := path
+	internalPath.DerivationPath.Branch = 1
+
+	// Act: Exchange with the original peer over the account's internal branch.
+	internalSecret, err := w.ECDH(ctx, internalPath, peerKey.PubKey())
+
+	// Assert: The secret is bound to the derivation path as well, so the
+	// wallet is not answering from one fixed key.
+	require.NoError(h, err, "failed to compute the internal branch secret")
+	require.NotEqual(
```
GustavoStingelin: [P2] Verify the exact internal-branch shared secret

I’d derive the public key for `internalPath`, compute the peer-side expected secret, and compare it with `internalSecret`. This assertion only proves the result differs from the external secret, so a zero or wrong-path result passes. I confirmed that by forcing internal-branch ECDH to return `[32]byte{}, nil` through a temporary compiler overlay: the btcd/SQLite `signer ecdh agreement` case still passed. The distinctness check can remain as a secondary assertion.

itest/signer_test.go
```diff
@@ -7,36 +7,744 @@
...
+		{
+			name:   "branch maximum",
+			branch: hdkeychain.HardenedKeyStart - 1,
+			index:  0,
+		},
+		{
+			name:   "index maximum",
+			branch: 0,
+			index:  hdkeychain.HardenedKeyStart - 1,
+		},
```
GustavoStingelin: Hmmm, I guess this rule only applies to watch-only wallets? For spendable ones, we should be able to derive hardened values here too?

GustavoStingelin, line 71: Q: Hmm, so for address indexes, does the spec mean we can only have `2^31` unhardened indexes, and that everything from `2^31` through `2^32 - 1` is considered invalid BIP44?

GustavoStingelin: I always thought we could have 2^32 address index here

itest/signer_test.go
```diff
@@ -7,36 +7,744 @@
...
+	// Arrange: A started, unlocked wallet and a peer key generated here, so
+	// the peer half of the exchange is never known to the wallet.
+	ctx := h.Context()
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	h.NewWalletAddressOfType(w, signerAddrType)
+
+	peerKey, err := btcec.NewPrivateKey()
+	require.NoError(h, err, "failed to generate the peer key")
+
+	defer peerKey.Zero()
```
GustavoStingelin, line 349: nit: I guess we don't need this zeroing, as it's random test material? Think the task should not ask for this zeroing.

itest/signer_test.go
GustavoStingelin: we may also document it here

GustavoStingelin (commented):
looking good, with the missing parts defered to https://github.com/btcsuite/btcwallet/pull/1376, i think we should state it in the description

itest/signer_test.go
Abdulkbk: Added the doc

Yes, the address_index is unhardened, so 0 ... 2^31-1 is the correct range.

itest/signer_test.go
```diff
@@ -7,36 +7,739 @@
...
-func testSignerDerivePubKey(h *bwtest.HarnessTest) {
-	// Arrange: Use the harness-owned lifecycle and address fixture to create a
-	// real started wallet with its default BIP84 account. The fixture restores
-	// the wallet's locked state before the Signer request is constructed.
+// signerScope is the key scope the derivation cases operate in. Its default
+// account is materialized by deriving an address of the matching type.
+var signerScope = waddrmgr.KeyScopeBIP0084
+
+// signerAddrType is the address type whose key scope is signerScope.
+const signerAddrType = waddrmgr.WitnessPubKey
```
yyforyongyu, line 21: These declarations independently encode the same `WitnessPubKey`/`BIP0084` relationship even though `waddrmgr.AddressType.KeyScope()` is the canonical mapping, so fixture setup and the tested signer path can drift if either value changes. Suggested fix: remove the package-level `signerScope`, derive the scope from `signerAddrType.KeyScope()` in each test that needs it, and require that lookup to succeed.

yyforyongyu (changes requested):
Again very clean! Just one comment.

itest/signer_test.go
Abdulkbk: Removed the package-level signerScope. Each case now derives its scope from signerAddrType.KeyScope() and requires the lookup to succeed, so signerAddrType is the only place the relationship is stated.

GustavoStingelin (approved):
LGTM!

itest/signer_test.go
yyforyongyu: I don't think the latest change is using `signerAddrType.KeyScope() `? Did you forget to push?

yyforyongyu (changes requested):
Looks like the claimed fix was not implemented.

itest/signer_test.go
Abdulkbk: Yeah, forgot to push.

Abdulkbk:
> Looks like the claimed fix was not implemented.

Forgot to push

yyforyongyu (approved):
LGTM!

PR #1376, opened 2026-09-10
wallet: reject nil Signer arguments before admission

Abdulkbk:
Four `Signer` entry points dereference an argument without checking it, and
`ECDH` additionally returns an all-zero "shared secret" for a key that is not a
curve point. Three of the four fault on the goroutine `handleReq` spawns, where
nothing in the package recovers, so a caller mistake takes the process down
rather than returning an error.

Found while writing the Task 441 `DerivePubKey`/`ECDH` integration tests, which
could not assert the "reject invalid input without exposing secret material"
contract, because the invalid input was not rejected.

### ECDH, nil key

```
  secp256k1.(*PublicKey).AsJacobian(...)   pubkey.go:227
  secp256k1.GenerateSharedSecret(...)      ecdh.go:16
  wallet.(*Wallet).ecdh(...)               signer.go:694
  wallet.(*Wallet).handleECDH(...)         signer.go:679
  wallet.(*Wallet).handleReq(...)          controller.go:841
created by wallet.(*Wallet).mainLoop in goroutine 36
```

The fault happens **after** `derivePathPrivKey` produced the leaf private key,
and `defer privKey.Zero()` never runs because the goroutine dies.

### ECDH, off-curve key — the quieter one

A non-nil key that is not a curve point multiplies to the point at infinity,
whose X coordinate is zero:

```
err=<nil>  secret=0000000000000000000000000000000000000000000000000000000000000000
```

A caller who forgot to parse a peer key gets a deterministic, publicly
predictable 32 bytes and no error. Both sides can agree on it while holding no
key at all. That is worse than the nil case, because a fault is loud and this
is not.

### The other three, and their fields

| Method | Argument | Faults on |
| --- | --- | --- |
| `SignDigest` | `intent` | caller's goroutine, first field read |
| `ComputeUnlockingScript` | `params`, `params.Output` | handler goroutine |
| `ComputeRawSig` | `params`, `params.Details` | handler goroutine |

`Output` and `Details` are dereferenced unconditionally and documented as
required, so guarding them needs no judgement. `Tx` and `SigHashes` are left
alone, because deciding whether they are optional is a real design question.

### Identity

Every nil argument returns `ErrNilArguments`, which `SignPsbt` already returns
for structurally identical code (`wallet/psbt_manager.go:997`). The off-curve
key returns `ErrInvalidSignParam`, since it is invalid rather than absent. No
new exported identity. The `Signer` interface docs now state both contracts.

### Verification

- Every guard was removed again afterwards to confirm its test fails, and fails
  with the fault rather than a plain assertion error.
- The ECDH tests register no store expectation, so the strict mock fails the
  test if a key is derived at all — they prove "no derivation", not just "an
  error came back".
- `go test ./wallet/` passes; all four commits build and pass on their own.

Every guard fires in the exported method, before `sendReq` and before any Store
call, so none of them can behave differently per database backend. That is why
the regression lives in `wallet` unit tests rather than in `itest`: an
integration case would repeat one backend-independent assertion three times,
and would assert less than these do, since registering no store expectation
lets the strict mock prove no key was derived at all.

### DIFF
--- a/wallet/signer.go
+++ b/wallet/signer.go
@@ -72,7 +72,8 @@ type Signer interface {
 	// will be the raw 32-byte shared secret (the X-coordinate of the
 	// result point).
 	//
-	// A nil remote key returns ErrNilArguments without deriving a wallet key.
+	// A nil remote key returns ErrNilArguments, and one that is not a curve
+	// point returns ErrInvalidSignParam; neither derives a wallet key.
 	ECDH(ctx context.Context, path BIP32Path, pub *btcec.PublicKey) (
 		[32]byte, error)
 
@@ -668,6 +669,17 @@ func (w *Wallet) ECDH(ctx context.Context, path BIP32Path,
 	if pub == nil {
 		return [32]byte{}, fmt.Errorf(
 			"%w: remote public key is nil", ErrNilArguments,
+		)
+	}
+
+	// A key that is not a curve point multiplies to the point at infinity,
+	// whose X coordinate is zero. That is a valid-looking 32-byte secret both
+	// sides could agree on without either holding a private key, so it has to
+	// be an error rather than a result.
+	if !pub.IsOnCurve() {
+		return [32]byte{}, fmt.Errorf(
+			"%w: remote public key is not a curve point",
+			ErrInvalidSignParam,
 		)
 	}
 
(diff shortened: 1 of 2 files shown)


yyforyongyu (approved):
LGTM!

PR #1379, opened 2026-09-12
wallet+db: add atomic bulk address allocation

yyforyongyu:
## Summary

Implement task 344.

Add atomic bulk address allocation for SQL wallets, returning complete batches
only after chain registration succeeds.

## Change Description

- Allocate 1–100 ordered addresses by account selector in one SQL transaction,
  preserving skipped-child progress, imported ownership and commit-error semantics.
- Reuse Store and chain.Interface for allocation and registration, and consume
  the dependency's startup replay and catch-up behavior.
- Cover concurrent allocation, failure semantics, imported-XPub result metadata,
  numbered SQL selection and committed-address replay after reopen.

## Notes

Depends on #1380. Review scope starts after that PR's task-live-replay branch;
startup readiness and synchronization changes belong to the dependency.


wallet/internal/db/pg/addressstore_newderived.go
```diff
@@ -32,6 +32,67 @@ func (s *Store) NewDerivedAddress(ctx context.Context,
...
+	var (
+		addresses []db.AddressInfo
+		exhausted bool
+	)
+
+	err := s.execWrite(ctx, func(qtx *sqlc.Queries) error {
+		// Leave nil callbacks to the shared admission check. Otherwise skip
+		// scripts already owned without modifying their rows or secrets.
+		derive := s.deriveAddress
+		if derive != nil {
```
yyforyongyu, line 49: check `if derive == nil` and return early? also why would `derive` be nil here?

wallet/address_manager.go
```diff
@@ -860,6 +1008,19 @@ func (w *Wallet) handleGetUnusedAddress(r getUnusedAddressReq) {
 			continue
 		}
 
+		// A prior allocation may have committed before registration failed.
+		// Re-register stored SQL rows before exposing them through reuse.
+		if w.addrStore == nil {
```
yyforyongyu, line 1013: explain why we need this change here as it's not related to this task? also why care about `w.addrStore == nil`?

wallet/internal/db/pg/addressstore_newderived.go
yyforyongyu: An early return would be a valid alternative. Config.DeriveAddress is explicitly optional for a Store used without derivation; the wallet Manager normally supplies it. The existing guard preserves a nil callback instead of wrapping it in a non-nil function, so shared allocation preflight can reject the call before account lookup or counter work. I kept that shared error boundary and clarified the reason in both SQL adapters.

wallet/address_manager.go
yyforyongyu: This is needed for NewBulkAddresses because the SQL transaction commits before live-watch registration. If registration fails, the call returns an error but the unused addresses remain stored. GetUnusedAddress can later return one of those rows, so it must successfully register it before exposing it. The existing addrStore == nil check identifies the SQL path; kvdb rejects bulk allocation. I clarified this connection in the code comment, and the failed-registration-then-reuse regression covers it.

wallet/address_manager_test.go
```diff
@@ -1739,3 +1742,155 @@ func TestNewBulkAddressesRejectsAdmission(t *testing.T) {
...
+			// expectation, to prove the public method never retries it.
+			w, deps := createTestWalletWithMocks(t)
+			w.addrStore = nil
+			startLoadedWalletForTest(t, w)
+			deps.store.On("NewDerivedAddresses", t.Context(),
+				mock.Anything, uint32(1)).Return(nil, tc.storeErr).Once()
+
+			// Act: Use normal admission so error mapping is exercised at the
+			// public boundary rather than by directly calling the mapper.
+			batch, err := w.NewBulkAddresses(t.Context(),
```
yyforyongyu: we should use a newline here?

wallet/address_manager_test.go
```diff
@@ -1739,3 +1742,155 @@ func TestNewBulkAddressesRejectsAdmission(t *testing.T) {
...
+				NewAccountSelectorByName(waddrmgr.KeyScopeBIP0084, "batch"),
+				false, 1,
+			)
+
+			var (
+				reused   address.Address
+				reuseErr error
+			)
+			if !tc.cancel {
+				reused, reuseErr = w.GetUnusedAddress(t.Context(), "batch",
```
yyforyongyu: same here re newline

wallet/address_manager.go
```diff
@@ -1051,12 +1088,24 @@ func nextUnusedStoreAddress(storeAddr db.AddressInfo,
...
+	addr := extractAddrFromPKScript(storeAddr.ScriptPubKey, w.cfg.ChainParams)
 	if addr == nil {
 		return nil, false, fmt.Errorf("%w: from pkscript %x",
 			ErrUnableToExtractAddress, storeAddr.ScriptPubKey)
 	}
 
+	// Bulk allocation can commit before registration fails. Re-register
+	// those stored rows before reuse exposes them. A nil legacy address
+	// store identifies SQL; kvdb rejects bulk allocation.
+	if w.addrStore == nil {
```
yyforyongyu: I don't like this check, as it's implicit, also leaks this context everywhere, if we wanna disable this for kvdb, we should do it in the public method `NewBulkAddresses` - check the db backend once there, and reject kvdb.

wallet/internal/db/itest/pg_test.go
```diff
@@ -263,6 +263,19 @@ func NewTestStoreWithDerive(t *testing.T,
...
 	t.Helper()
+
+	store, _ := newReopenableTestStore(t, deriveAddress)
+
+	return store
+}
+
+// newReopenableTestStore retains the same database configuration across opens
+// so batch tests can prove progress survives a closed connection pool.
+func newReopenableTestStore(t *testing.T,
```
yyforyongyu, line 274: how are these changes related to the current `NewBulkAddresses`?

wallet/controller.go
```diff
@@ -886,6 +892,13 @@ func (w *Wallet) verifyBirthday(ctx context.Context) error {
 		return fmt.Errorf("get wallet birthday: %w", err)
 	}
 
+	// SQL replay can start Neutrino history before syncer initialization.
+	// Install the persisted birthday first so replay cannot start at genesis.
+	client, ok := w.cfg.Chain.(*chain.NeutrinoClient)
```
yyforyongyu: again we are fixing unrelated bugs here in `NewBulkAddresses`?

wallet/address_manager_test.go
yyforyongyu: Agreed. I moved the arguments below the opening parenthesis and fixed the same wrapping pattern in the bulk tests, following the repository formatting rules. The changes are folded into the commits that introduced the tests.

wallet/address_manager_test.go
yyforyongyu: Agreed. The GetUnusedAddress call now starts its arguments on the next line, with the closing parenthesis on its own line. This is formatting only.

wallet/address_manager.go
yyforyongyu: NewBulkAddresses already rejects kvdb at the public entry point, before allocating. This check has a separate purpose in GetUnusedAddress: a SQL bulk transaction can commit and then fail watch registration, leaving an unused row that reuse must register before returning. Keeping this watch SQL-only preserves existing kvdb reuse behavior; making it unconditional would add a new chain call that can fail for kvdb. I kept both the public bulk rejection and this reuse protection.

wallet/internal/db/itest/pg_test.go
yyforyongyu: This fixture lets the bulk tests close the PostgreSQL pool and reopen the same database. TestNewDerivedAddressesDurableBatches uses it to verify that concurrent batches remain disjoint and the next child follows committed progress after reopen; the invalid-child, rollback, and exhaustion tests use it for the same durability requirement. NewTestStoreWithDerive delegates to it so we can reuse the existing setup without duplicating it.

wallet/controller.go
yyforyongyu: Bulk allocation commits before watch registration, so failed delivery can leave durable addresses that a fresh startup must register. That replay now runs before syncer initialization. On Neutrino, WatchAddrsFromTip calls NotifyReceived, which can start a rescan immediately using its current startTime. Installing the persisted birthday first prevents this new replay from starting at the zero time; setting it later does not restart an already-running scan. This ordering is required by the bulk restart path.

wallet/address_manager.go
```diff
@@ -668,6 +699,119 @@ func (w *Wallet) NewAddress(ctx context.Context, accountName string,
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Copy the semantic selector into the admitted request. The Store resolves
+	// it inside the write transaction, including the receiving-policy guard.
+	params := db.NewDerivedAddressParams{
+		WalletID: w.id, Scope: db.KeyScope(selector.keyScope),
+		AccountNumber: (*uint32)(selector.accountNumber),
+		Change:        internal, RequireChainSync: true,
```
yyforyongyu: `RequireChainSync: true` should be in newline? how come you came up with this unreasonable format?!

yyforyongyu: Fixed in 4ba3ba629. Both the Store parameter literal and the request literal now put each field on its own line, including RequireChainSync. Values and behavior are unchanged; repository formatting and checks pass.

wallet/internal/db/itest/addressstore_test.go
```diff
@@ -3409,3 +3410,442 @@ func TestNewDerivedAddressMaxIndexInternal(t *testing.T) {
...
+			type result struct {
+				batch []db.AddressInfo
+				err   error
+			}
+
+			results := make(chan result, 2)
+
+			// Act: Race two whole batches, join their results, and reopen the
+			// Store before requesting an already exhausted branch again.
+			for range 2 {
```
yyforyongyu: we are not testing concurrent behavior here so this is an overkill?

wallet/internal/db/itest/pg_test.go
```diff
@@ -274,14 +287,18 @@ func NewTestStoreWithDerive(t *testing.T,
...
-	})
+		t.Cleanup(func() {
+			_ = store.Close()
+		})
 
-	return store
+		return store
+	}
+
+	return open(), open
```
yyforyongyu, line 301: if we are calling `open()` here, then why return it? we can just return `open`?

yyforyongyu: i also don't think we wanna expand test cases here, why all of sudden we start asserting reopen behavior?

wallet/internal/db/sqlite/addressstore_newderived.go
```diff
@@ -32,6 +32,69 @@ func (s *Store) NewDerivedAddress(ctx context.Context,
...
+		exhausted bool
+	)
+
+	err := s.execWrite(ctx, func(qtx *sqlc.Queries) error {
+		// Config.DeriveAddress may be nil for a Store that never derives.
+		// Preserve nil so shared preflight rejects allocation before account
+		// or counter work. Otherwise skip already-owned scripts without
+		// modifying their rows or secrets.
+		derive := s.deriveAddress
+		if derive != nil {
```
yyforyongyu: same here re check `if derive == nil`? such that we can save some indentations?

wallet/internal/db/pg/addressstore_newderived.go
yyforyongyu: Implemented the early return in 476c1e8bf. A nil derivation callback now fails before a write transaction opens, and the collision wrapper is no longer nested. Config permits nil for Stores that do not derive addresses; the shared helper guard stays unchanged, with no new exported error.

wallet/internal/db/itest/addressstore_test.go
yyforyongyu: Simplified in 71e9e0882: the exhaustion test now makes two sequential calls, with the channel, result struct and goroutines removed. It retains the existing outcome and terminal-counter assertions. Concurrency remains covered by the dedicated batch test.

wallet/internal/db/itest/pg_test.go
yyforyongyu: Returning only the opener is possible, but it adds an explicit initial open at its callers. I kept the existing convenience pair to avoid that optional refactor.

For reopen coverage, the bulk contract requires durable rows and consumed child progress across restart on both SQL dialects. I kept the dedicated durability test and terminal-progress reopen checks, and removed the repeated close/reopen steps from ordinary skipped-child and rollback tests in 71e9e0882. No new cases or fixture changes were added.

wallet/internal/db/sqlite/addressstore_newderived.go
yyforyongyu: Implemented in 476c1e8bf for both SQLite and PostgreSQL: reject a nil configured deriver before execWrite, then define the collision wrapper without the extra conditional. The shared guard is unchanged and no sentinel or helper is exported.

wallet/internal/db/itest/addressstore_test.go
```diff
@@ -3409,3 +3410,430 @@ func TestNewDerivedAddressMaxIndexInternal(t *testing.T) {
...
+	}
+
+	require.ElementsMatch(t, []uint32{0, 2}, starts)
+}
+
+// TestNewDerivedAddressesSurvivesReopen checks that both address rows and the
+// next-child counter survive closing and reopening the SQL Store.
+func TestNewDerivedAddressesSurvivesReopen(t *testing.T) {
+	t.Parallel()
+
```
GustavoStingelin: Do we really need to test reopening here?
Same as: https://github.com/btcsuite/btcwallet/pull/1365#discussion_r3980675545

wallet/internal/db/addressstore_newderived.go
```diff
@@ -162,26 +196,117 @@ func NewDerivedAddressWithOps(ctx context.Context,
...
+				PubKey:       pubKey,
+			})
+
+		case errors.Is(err, ErrAddressChildUnavailable):
+			// The counter advances even though this child is not returned.
+		default:
+			return nil, false, err
+		}
+
+		if lastIndex == hdkeychain.HardenedKeyStart-1 {
```
GustavoStingelin: Exhaustion near the end of the branch commits the counter past valid children that were already derived.

E.g. the external counter is at `2^31-5` and we ask for 10: we derive 5 valid children, stop at `2^31-1`, return `exhausted=true`, and the caller commits the counter advance with no rows. Those 5 usable addresses are gone for good, and the branch now rejects even a single `NewAddress` that would have succeeded before this call.

This also contradicts the comment above ("Exhaustion must not persist a prefix") and the interface doc ("exhaustion may commit consumed invalid indexes"). Here, valid children get consumed too.

This matters more once `NewAddress` returns an existing unused address and only falls back to creating one when none is available. A bulk refill that runs near the end of the branch would then break the fallback path. Could we roll back on exhaustion whenever a valid candidate would be dropped, and only commit when every consumed child was invalid or already owned? If committing is intended, the comments need to say so.

wallet/address_manager.go
```diff
@@ -655,6 +699,123 @@ func (w *Wallet) NewAddress(ctx context.Context, accountName string,
...
+// that children are reusable. Exhaustion may persist consumed invalid children.
+func (w *Wallet) NewBulkAddresses(ctx context.Context, selector AccountSelector,
+	internal bool, count uint32) ([]AddressInfo, error) {
+
+	if count == 0 || count > MaxBulkAddressCount {
+		return nil, fmt.Errorf("%w: count must be 1..%d",
+			ErrInvalidParam, MaxBulkAddressCount)
+	}
+
+	err := selector.validate()
```
GustavoStingelin: `NewAddress` and `GetUnusedAddress` both reject the catch-all imported account up front with `ErrImportedAccountNoAddrGen` (see `newAddress`). This path skips that check. It opens a write transaction, bumps the counter, and then fails inside derivation with the internal `errMissingAccountPubKey`. `bulkAddressErr` sends that to the default case, so callers can't match `ErrImportedAccountNoAddrGen`. Nothing is persisted because the transaction rolls back, but the error identity is wrong.

Since `NewAddress` is expected to fall back to creating addresses when there's no unused one, every creation entry point should agree on this. Could we add the same guard here for `selector.accountName`, or better, reject a non-derived account with no pubkey in the shared `derivedAddressAccount` preflight, before `NextIndex`, so the count-one and batch paths both get it?

wallet/internal/db/sqlite/addressstore_newderived.go
```diff
@@ -32,6 +32,70 @@ func (s *Store) NewDerivedAddress(ctx context.Context,
 	return info, nil
 }
 
+// NewDerivedAddresses commits batches under the counter lock, skipping owned
+// scripts in the same transaction to preserve raw-import metadata.
+func (s *Store) NewDerivedAddresses(ctx context.Context,
```
GustavoStingelin: This doesn't quite follow the ops pattern (`wallet/internal/db/AGENTS.md`). This method and `pg.NewDerivedAddresses` are line-for-line identical: the nil-callback guard, the owned-script derive wrapper, and the mapping from `exhausted` to `ErrMaxAddressIndexReached`. The only backend-specific part is `GetAddressByScriptPubKey`.

Suggestion: add a batch-specific ops interface (e.g. `NewDerivedAddressesOps` embedding `NewDerivedAddressOps` plus `AddressOwned(ctx, walletID, script) (bool, error)`) and move the skip logic and the exhausted→error mapping into `NewDerivedAddressesWithOps`. Then each backend method shrinks to the `execWrite` + helper call, like `NewDerivedAddress` above.

Smaller points:
- `errors.New("address derivation callback is nil")` redefines `db.errNilAddressDerivationFunc` in both backends. It would be better as one exported sentinel (or checked in the shared helper).
- The `NewDerivedAddressOps` doc, which lists the ordered stages, still only describes the count-one flow. It should cover the batch flow too: collect candidates first, then insert.

wallet/internal/db/sqlite/addressstore_newderived.go
```diff
@@ -32,6 +32,70 @@ func (s *Store) NewDerivedAddress(ctx context.Context,
...
+	params db.NewDerivedAddressParams, count uint32) ([]db.AddressInfo, error) {
+
+	// Derivation is optional for Stores used only for other operations.
+	// Reject a receiving allocation before opening a write transaction.
+	if s.deriveAddress == nil {
+		return nil, errors.New("address derivation callback is nil")
```
GustavoStingelin: nit: this redefines `db.errNilAddressDerivationFunc` (the same thing happens in `pg`). Could we reuse a single exported sentinel, or leave the check to the shared helper, which already rejects a nil callback?

wallet/internal/db/addressstore_newderived.go
```diff
@@ -162,26 +196,117 @@ func NewDerivedAddressWithOps(ctx context.Context,
...
-		ctx, params, account, accountNumber, ops, deriveFn,
-	)
+	return account, accountNumber, nil
+}
+
+// NewDerivedAddressesWithOps derives a complete batch within the caller's write
+// transaction. Exhausted reports consumed children to commit without rows;
+// other errors require rollback. The caller must discard
+// all results on a commit error and must never retry an ambiguous commit.
+func NewDerivedAddressesWithOps(ctx context.Context,
```
GustavoStingelin: nit: the ordered-stages doc on `NewDerivedAddressOps` only describes the single-address flow. Now that this helper shares that interface, it should also describe the batch flow: derive all candidates first (skipping unavailable children), then insert them, plus how exhaustion is handled.

wallet/internal/db/sqlite/addressstore_newderived.go
```diff
@@ -32,6 +32,70 @@ func (s *Store) NewDerivedAddress(ctx context.Context,
...
+			if err != nil || data == nil {
+				return data, err
+			}
+
+			_, err = qtx.GetAddressByScriptPubKey(
+				ctx, sqlc.GetAddressByScriptPubKeyParams{
+					WalletID:     int64(params.WalletID),
+					ScriptPubKey: data.ScriptPubKey,
+				},
+			)
```
GustavoStingelin: I think the derive callback should only do derivation. This ownership lookup is a DB concern that's hidden inside it, so it's invisible to the shared helper and gets copied into `pg` as well.

It would fit the ops pattern better as an ops method (e.g. `AddressOwned`, see the comment on `NewDerivedAddresses` above). The shared helper would then call it after derivation and mark the child as `ErrAddressChildUnavailable`, and the backend would pass `s.deriveAddress` directly.

GustavoStingelin (changes requested):

wallet/internal/db/itest/addressstore_test.go
yyforyongyu: I kept this dedicated reopen case because the bulk contract requires both the complete batch and its next-child counter to survive closing and reopening the store. It runs for both SQL backends. The ordinary skip/rollback cases no longer repeat reopen checks; this is the one focused durability regression.

wallet/internal/db/addressstore_newderived.go
yyforyongyu: Agreed that the comments were incomplete: when the terminal range cannot supply the requested batch, every attempted index is consumed, including valid candidates, and no batch rows are inserted. I clarified the public, Store and helper documentation. I kept the post-commit exhaustion result because rolling back would also make skipped invalid children reusable, contrary to the consumed-child contract. The existing mixed valid-prefix/invalid-terminal test checks the committed counter and absence of rows.

wallet/address_manager.go
yyforyongyu: Fixed by rejecting the reserved imported account name before request admission with ErrImportedAccountNoAddrGen, matching the other receiving methods. Added one case to the existing admission test; strict Store and chain mocks ensure neither dependency is called. Separately named imported-XPub accounts remain supported.

wallet/internal/db/sqlite/addressstore_newderived.go
yyforyongyu: The duplicated ownership wrapper could be factored, but I am keeping the current transaction-bound adapter query for this bounded change. In particular, exhaustion must be mapped after execWrite succeeds: returning that error from the transaction helper would roll back the progress we need to preserve. There is no demonstrated correctness failure requiring another Ops method here.

wallet/internal/db/sqlite/addressstore_newderived.go
yyforyongyu: I kept this early check so a missing callback is rejected before execWrite opens a transaction. The shared helper error is private and its identity is not part of the Store contract. Moving the check into the helper would lose that early rejection; exporting a sentinel solely to share this message would add unnecessary API surface.

wallet/internal/db/addressstore_newderived.go
yyforyongyu: Updated the Ops documentation to distinguish the single-address ordering from batch preflight, full candidate collection before insertion, and terminal exhaustion that commits attempted-index progress without inserting batch rows.

wallet/internal/db/sqlite/addressstore_newderived.go
yyforyongyu: The separation suggestion makes sense as an optional refactor. Here the configured derivation callback remains unchanged; the private SQL adapter wrapper adds the ownership query using the same transaction. Existing collision coverage verifies that raw imported ownership is preserved. I am keeping that local mechanism rather than expanding Ops for this PR without a demonstrated failure.

wallet/internal/db/sqlite/addressstore_newderived.go
yyforyongyu: Correction to my earlier reply: the ownership part is now implemented. `AddressOwned` is part of the existing `NewDerivedAddressOps`, and the shared batch candidate loop uses it to consume an already-owned child without changing its row or secret. SQLite and PostgreSQL implement it with their existing `GetAddressByScriptPubKey` query in the same write transaction. I retained the early nil-callback guard and the post-commit exhaustion mapping because moving exhaustion into the transaction callback would roll back the counter progress that must remain consumed.

wallet/internal/db/sqlite/addressstore_newderived.go
yyforyongyu: Agreed and fixed. Correction to my earlier reply: the configured `AddressDerivationFunc` is now passed directly to the shared helper and performs derivation only. The shared candidate loop checks ownership through `NewDerivedAddressOps.AddressOwned`; each SQL implementation runs the existing ownership query in the allocating transaction. The cross-dialect raw-ownership regression continues to verify that the collided imported row and secret remain unchanged.

wallet/internal/db/sqlite/addressstore_newderived.go
yyforyongyu: You are right; correction to my earlier reply. I removed the duplicate adapter-local nil checks from both SQLite and PostgreSQL. `NewDerivedAddressesWithOps` now owns this validation through the existing private sentinel and rejects the nil callback before any adapter operation or mutation, matching the count-one path. The adapters continue to pass the original callback directly now that ownership lookup is handled by `AddressOwned`; no exported sentinel, new helper, or additional test was needed. Folded into `db: wire atomic SQL address batches`.

GustavoStingelin (approved):
LGTM 🚢

PR #1380, opened 2026-09-14
wallet: restore live transaction watches

yyforyongyu:
## Summary

Implement task 402: restore and maintain full-node live transaction watches.

## Change Description

Expose the existing `NotifySpent` operation through `chain.Interface`.
Register stored addresses and outpoints during sync initialization, refresh
watches after catch-up and targeted scan commits, and keep readiness false
until registration succeeds. Register an existing unused address before
returning it.

Preserve the existing catch-up and filtered-block notification paths. Neutrino
keeps its existing address rescan; spend registration is a documented no-op.
Cover persisted and recovered watches, registration retry, cancellation, and
readiness with focused tests.


wallet/common_test.go
```diff
@@ -281,6 +281,10 @@ func createTestWalletWithMocks(t *testing.T) (*Wallet, *mockWalletDeps) {
 	mockStore := &walletmock.Store{}
 	mockTxStore := &bwmock.TxStore{}
 	mockSyncer := &mockChainSyncer{}
+	// Most fixture users exercise requests without starting a real syncer;
```
yyforyongyu: add a newline above

wallet/syncer.go
```diff
@@ -266,6 +270,142 @@ type syncer struct {
 	publisher TxPublisher
 }
 
+// liveRPCClient exposes the legacy spend operation consumed by btcd handlers.
+type liveRPCClient interface {
+	NotifySpent(points []*wire.OutPoint) error
+}
+
+// liveBitcoindClient exposes additive filters without optional mempool lookup.
+type liveBitcoindClient interface {
```
yyforyongyu: the `s.cfg.Chain` is already the `chain.Interface`? so why do we need to define these interfaces again?

wallet/syncer.go
```diff
@@ -266,6 +270,142 @@ type syncer struct {
...
+// These assertions bind the private method shapes to the existing adapters.
+var (
+	_ chainSyncer        = (*syncer)(nil)
+	_ liveRPCClient      = (*chain.RPCClient)(nil)
+	_ liveBitcoindClient = (*chain.BitcoindClient)(nil)
+)
+
+// liveOnlyChain recognizes the existing full-node adapter mechanisms; SPV
+// lacks these operations and retains its existing recovery behavior.
+func liveOnlyChain(client chain.Interface) bool {
```
yyforyongyu: we should remove it

wallet/syncer.go
```diff
@@ -266,6 +270,142 @@ type syncer struct {
...
+		return fmt.Errorf("read watch addresses: %w", err)
+	}
+
+	// A persisted obligation must finish its handoff even if its caller leaves.
+	err = s.cfg.Chain.WatchAddrsFromTip(context.WithoutCancel(ctx), addrs)
+	if err != nil {
+		return fmt.Errorf("register addresses: %w", err)
+	}
+
+	err = s.watchLiveOutpoints(ctx, addrs)
```
yyforyongyu: this is asymmetric? we call `storeScanAddresses` + `WatchAddrsFromTip` here, but wrap `storeScanUnspent` + `NotifySpent` in a new method `watchLiveOutpoints`?

wallet/syncer.go
```diff
@@ -266,6 +270,142 @@ type syncer struct {
...
+	}
+
+	if len(outpoints) == 0 {
+		return nil
+	}
+
+	// Btcd uses legacy redeeming notifications; bitcoind's additive
+	// filter avoids NotifySpent's optional mempool input lookup.
+	switch client := s.cfg.Chain.(type) {
+	case liveBitcoindClient:
```
yyforyongyu: we should not use it, just use `NotifySpent`? why care about the backend type here?

wallet/syncer.go
```diff
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"slices"
```
yyforyongyu: syncer related pkg change should be the first commit? then we make use of it in the controller in a followup commit?

chain/neutrino.go
```diff
@@ -586,6 +586,12 @@ func (s *NeutrinoClient) NotifyReceived(addrs []address.Address) error {
 	return nil
 }
 
+// NotifySpent cannot provide an outpoint-only mempool watch through SPV;
+// Neutrino's historical input tracking remains part of Rescan.
+func (s *NeutrinoClient) NotifySpent(_ []*wire.OutPoint) error {
```
yyforyongyu, line 591: if this is not supported, how can we get the onchain state of these outpoints?

wallet/address_manager.go
```diff
@@ -724,6 +724,8 @@ func (w *Wallet) newAddress(ctx context.Context, accountName string,
 // under standards like BIP44 that enforce a gap limit of 20 unused addresses.
 // If all previously derived addresses have been used, this method delegates
 // to NewAddress to generate a new one.
+// Full-node success includes live registration for children materialized
```
yyforyongyu: unnecessary doc addition, remove it

wallet/address_manager.go
```diff
@@ -778,7 +780,12 @@ func (w *Wallet) GetUnusedAddress(ctx context.Context, accountName string,
 	// Once admitted, wait for the result even if cancellation arrives.
 	result := <-r.respChan
 
-	return result.addr, result.err
+	// A failed registration must not expose the selected receiving address.
+	if result.err != nil {
```
yyforyongyu: why do we make this change? this is unrelated? plus the old return behavior is already correct?

wallet/address_manager.go
```diff
@@ -847,7 +854,15 @@ func (w *Wallet) handleGetUnusedAddress(r getUnusedAddressReq) {
 			continue
 		}
 
-		r.respChan <- addressResp{addr: unusedAddr}
+		// A scan may have just persisted this child. Join its live watch
+		// before returning, even if the scan's own refresh is pending.
+		if w.cfg.Chain.BackEnd() != neutrinoBackend {
```
yyforyongyu: i'd rather make `NotifySpent` a noop in neutrino than expose backend details in the code - all chain backend should have been abstracted via `chain.Interface`

itest/manager_test.go
```diff
@@ -273,3 +281,295 @@ func testManagerCreateWatchOnly(h *bwtest.HarnessTest) {
...
+	// Arrange: Fund one output, then discard the entire client filter along
+	// with the stopped Manager. The harness retains ownership of each resource.
+	manager := h.NewWalletManager()
+	// Join the current Manager before cleanup stops replacement sources;
+	// the harness still owns registry teardown, including failure exits.
+	defer func() {
+		require.NoError(h, manager.Stop())
+	}()
+
+	_, err := manager.Start(h.Context())
```
yyforyongyu, line 296: what are these tests? did you not check the required coding format for tests? how are we putting everything without newlines or appropriate comments???

wallet/common_test.go
yyforyongyu: Added the blank line between the mock declaration and its expectation setup in `wallet/common_test.go`.

wallet/syncer.go
yyforyongyu: Removed `liveRPCClient` and `liveBitcoindClient`. `NotifySpent` is now on `chain.Interface`, and the syncer calls that shared method directly.

wallet/syncer.go
yyforyongyu: Removed `liveOnlyChain`. The remaining Neutrino-specific checks are being assessed separately under your newer backend-abstraction comment.

wallet/syncer.go
yyforyongyu: Removed `watchLiveOutpoints`; `refreshLiveWatches` now contains both address and outpoint registration, including the existing output-eligibility filtering.

wallet/syncer.go
yyforyongyu: The syncer now calls `s.cfg.Chain.NotifySpent(outpoints)` directly. Removed the backend type switch and the separate `LoadTxFilter` path. The newer Neutrino-semantics comment remains open.

wallet/syncer.go
yyforyongyu: The syncer implementation now precedes the controller startup-wiring commit. I am also splitting the remaining mixed wallet commit by responsibility, as requested separately.

chain/neutrino.go
yyforyongyu: Confirmed outpoint state still comes from block scanning: the wallet loads persisted outputs through `storeScanUnspent`, tracks their scripts/outpoints in recovery, and checks spends in blocks. Neutrino also retains its existing rescan input watches. `NotifySpent` adds native transaction notifications on full nodes; it is now a documented no-op on Neutrino, which does not provide external mempool transaction discovery. Updated the comments to make that distinction explicit.

wallet/address_manager.go
yyforyongyu: Removed the added two-line `GetUnusedAddress` documentation. The reason for registration remains beside the handler code.

wallet/address_manager.go
yyforyongyu: Restored the original `return result.addr, result.err`. Registration failure now sends an error-only response from `handleGetUnusedAddress`, matching its other error branches. The existing failure regression still verifies that no address is returned.

wallet/address_manager.go
yyforyongyu: Changed Neutrino `NotifySpent` to a documented no-op and removed the backend check from `handleGetUnusedAddress`; it now calls `chain.Interface.NotifyReceived` uniformly, as `newAddress` already does. The syncer retains the distinct history/event guards: Neutrino address replay would otherwise start a historical rescan before its birthday is installed, and its confirmed-event path still differs. Those guards do not select a spend-registration implementation. No new backend interface or capability layer was added.

itest/manager_test.go
yyforyongyu: Reworked all three live-watch integration tests using `itest/utxo_manager_test.go` as the style reference: logical stanzas with blank lines, local explanations for source replacement and external-only spends, descriptive assertion failures, script/transaction preparation in Arrange, and separate Act/Assert phases for mining and confirmation. Kept the same three scenarios and existing harness ownership; added no helper or test case.

chain/neutrino_test.go
```diff
@@ -316,3 +316,24 @@ func TestNeutrinoClientNotifyReceivedRescan(t *testing.T) {
 		require.Equal(t, wantMsgs, gotMsgs)
 	}
 }
+
+// TestNeutrinoClientNotifySpent verifies the SPV no-op needs no chain service
+// and leaves historical input tracking to the existing rescan.
+func TestNeutrinoClientNotifySpent(t *testing.T) {
```
yyforyongyu: this test should now be removed

wallet/common_test.go
```diff
@@ -126,6 +126,9 @@ func createTestChain(tb testing.TB) *bwmock.Chain {
 	tb.Helper()
 
 	chain := &bwmock.Chain{}
+	// Lifecycle fixtures use SPV identity to leave live full-node replay to
```
yyforyongyu: newline above

wallet/syncer.go
```diff
@@ -1762,15 +1766,27 @@ func (s *syncer) processChainUpdate(ctx context.Context, update any) error {
...
+
 		matches := s.prepareTxMatches([]*wtxmgr.TxRecord{n.TxRecord})
+
 		return s.putTxNotifications(ctx, matches, n.Block)
 
 	case chain.FilteredBlockConnected:
+		// Full-node filters may be incomplete for the reported interval.
+		// Wake normal catch-up from its durable tip instead of advancing
+		// past transactions absent from this filtered notification.
+		if s.cfg.Chain.BackEnd() != neutrinoBackend {
```
yyforyongyu: i don't think this is needed

wallet/syncer_test.go
```diff
@@ -5388,9 +5407,10 @@ func TestProcessChainUpdate(t *testing.T) {
 	t.Parallel()
 
 	tests := []struct {
-		name   string
-		update interface{}
-		setup  func(*walletmock.Store, *bwmock.Chain)
+		name    string
+		backend string
```
yyforyongyu: stop expanding this - if we wanna test backend specific we should create new tests? tho i doubt because if we wanna check integration behavior that belongs to the itest?

wallet/syncer.go
```diff
@@ -270,6 +271,69 @@ type syncer struct {
 // by Neutrino, unlike full-node live transaction subscriptions.
 const neutrinoBackend = "neutrino"
 
+// Ensure the startup and scan worker use the same synchronization contract.
+var _ chainSyncer = (*syncer)(nil)
+
+// refreshLiveWatches joins persisted obligations from startup or scan commits
```
yyforyongyu: rewrite all the godocs in this PR - you should use existing language and wording from other files.

wallet/syncer.go
```diff
@@ -270,6 +271,69 @@ type syncer struct {
...
+// with native registration, without storing a second copy of the watch set.
+func (s *syncer) refreshLiveWatches(ctx context.Context) error {
+	// SPV address registration starts history; startup must first install
+	// its birthday in initChainSync before that rescan can run.
+	if s.cfg.Chain.BackEnd() == neutrinoBackend {
```
yyforyongyu: remove this check, why do we still need it?

wallet/syncer.go
```diff
@@ -2261,19 +2328,20 @@ func (s *syncer) storeImportedScanAddresses(
 func (s *syncer) storeScanAddressRelevant(isRawImported, isDefaultScope bool,
 	info db.AddressInfo) bool {
 
-	// SQL has no legacy branch restriction on persisted account addresses.
-	if s.addrStore == nil || isRawImported || isDefaultScope {
+	// SQL and full-node live replay include all persisted account addresses;
+	// only legacy SPV recovery retains the custom-scope branch restriction.
+	if s.addrStore == nil || isRawImported || isDefaultScope ||
+		s.cfg.Chain.BackEnd() != neutrinoBackend {
```
yyforyongyu: remove it

wallet/syncer.go
```diff
@@ -270,6 +271,69 @@ type syncer struct {
...
+	}
+
+	credits, err := s.storeScanUnspent(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Match owned members even when an output is bare multisig. String keeps
+	// P2PK identities distinct from P2PKH while excluding NoChainSync rows.
+	eligible := make(map[string]struct{}, len(addrs))
```
yyforyongyu: should these be done in `storeScanUnspent` instead? if not, we should create a new method like `filterEligibleOutpoints`? and i am not sure if we really need this filtering?

wallet/address_manager.go
```diff
@@ -847,6 +847,15 @@ func (w *Wallet) handleGetUnusedAddress(r getUnusedAddressReq) {
 			continue
 		}
 
+		// A scan may have just persisted this child. Join its live watch
+		// before returning, even if the scan's own refresh is pending.
+		err = w.cfg.Chain.NotifyReceived([]address.Address{unusedAddr})
```
yyforyongyu: why not using `WatchAddrsFromTip` but this method?

wallet/syncer_test.go
```diff
@@ -3223,7 +3361,11 @@ func TestStoreScanAddressesNonDefaultScope(t *testing.T) {
 			store := &walletmock.Store{}
 			addrStore := &bwmock.AddrStore{}
 			s := newSyncer(
-				Config{ChainParams: &chainParams}, nil, nil,
+				Config{
+					// KVDB branch filtering here is the SPV contract.
```
yyforyongyu: do we still need this now that we should be db and chain backend agnostic?

chain/neutrino_test.go
yyforyongyu: Removed the trivial NotifySpent no-op test. Existing Neutrino integration tests cover confirmed payments and spends through block scanning.

wallet/common_test.go
yyforyongyu: Added the logical separation after the mock declaration. The Neutrino identity used to bypass registration is gone; the generic chain-name expectation remains only for Info responses.

wallet/syncer.go
yyforyongyu: Removed the backend distinction. Filtered blocks now wake ordinary scanning on every backend, and confirmed RelevantTx notifications leave confirmation writes to that same scan. Removed the unused filtered-block writer.

wallet/syncer_test.go
yyforyongyu: Reduced this to a data-only table for ignored confirmed/block notifications. Removed backend fields, setup closures and conditional mock setup. Existing separate tests cover unmined transactions and disconnects; real delivery stays in the integration tests.

wallet/syncer.go
yyforyongyu: Reviewed the added and changed Godocs and replaced the abstract registration wording with concrete address, outpoint, Store and notification descriptions, matching nearby code.

wallet/syncer.go
yyforyongyu: Removed the wallet check. Neutrino now handles its lack of mempool notifications inside WatchAddrsFromTip as a documented no-op that honors cancellation. NotifyBlocks still starts its block notifications after birthday setup; confirmed state comes from block scanning.

wallet/syncer.go
yyforyongyu: Removed the chain-backend condition and restored the pre-PR predicate. This leaves the existing SQL/KVDB custom-scope policy unchanged while removing chain identity from address selection.

wallet/syncer.go
yyforyongyu: Removed the duplicate filtering entirely. ListOutputsToWatch already owns output eligibility and excludes NoChainSync accounts; KVDB does not support that policy. The syncer now registers every returned outpoint without another map, script parse, or helper.

wallet/address_manager.go
yyforyongyu: Changed this to WatchAddrsFromTip with context.WithoutCancel(r.ctx), preserving completion of the accepted registration. Updated the existing success/error and completion-barrier expectations.

wallet/syncer_test.go
yyforyongyu: Removed the added chain mock and comment. Kept the existing SQL/KVDB table because it tests the pre-existing custom-scope policy, which this PR now leaves unchanged.

bwtest/mock/mempool_chain.go
```diff
@@ -187,6 +188,12 @@ func (c *MempoolChain) Rescan(*chainhash.Hash, []address.Address,
 // NotifyReceived implements the chain.Interface interface.
 func (c *MempoolChain) NotifyReceived([]address.Address) error { return nil }
 
+// NotifySpent returns an unsupported error because MempoolChain does not
+// provide transaction notifications.
+func (c *MempoolChain) NotifySpent(_ []*wire.OutPoint) error {
```
yyforyongyu, line 193: why do we need to change this mempool chain? it is not used?

wallet/syncer.go
```diff
@@ -327,13 +363,23 @@ func (s *syncer) initChainSync(ctx context.Context) error {
 		return fmt.Errorf("unable to check for rollback: %w", err)
 	}
 
+	// A retry may follow a committed scan whose registration failed.
+	// Register stored addresses and outpoints even if the tip is current.
+	// TODO(yy): Avoid rereading and registering the full set on each retry
```
yyforyongyu, line 369: add a newline above `//`

wallet/syncer.go
```diff
@@ -327,13 +363,23 @@ func (s *syncer) initChainSync(ctx context.Context) error {
...
+	}
+
 	// Explicitly request connected/disconnected block notifications. Only
 	// the bitcoind and neutrino wrappers enable block notifications as a
 	// side effect of NotifyReceived; btcd's NotifyReceived (the embedded
 	// rpcclient method) registers only address notifications, so without
 	// this call a btcd-backed wallet would stop receiving live block
 	// connect/disconnect events. The call is idempotent on the wrappers
 	// that already enabled notifications, so it is safe for all backends.
+
```
yyforyongyu: why adding a newline here??

wallet/syncer_test.go
```diff
@@ -382,6 +451,7 @@ func TestInitChainSync(t *testing.T) {
 	// Arrange: Initialize a store-backed syncer and mock its dependencies
 	// for the initial synchronization sequence.
 	mockChain := &bwmock.Chain{}
+
```
yyforyongyu: stop adding unnecessary changes to this PR!

wallet/syncer_test.go
```diff
@@ -4312,13 +4426,28 @@ func TestInitChainSync_NotifyBlocksError(t *testing.T) {
 	// Arrange: Setup a store-backed syncer whose block-notification request
 	// fails.
 	mockChain := &bwmock.Chain{}
+
```
yyforyongyu: unnecessary change!!

wallet/syncer_test.go
```diff
@@ -4312,13 +4426,28 @@ func TestInitChainSync_NotifyBlocksError(t *testing.T) {
...
 	store := &walletmock.Store{}
 	s := newSyncer(
 		Config{Chain: mockChain}, nil, nil, nil,
 		store, 0,
 	)
 
 	mockChain.On("IsCurrent").Return(true).Once()
+	// Registration reads an empty watch set before enabling blocks.
```
yyforyongyu, line 4769: now it's the place to add a newline yet you missed it?

itest/manager_test.go
```diff
@@ -273,3 +279,122 @@ func testManagerCreateWatchOnly(h *bwtest.HarnessTest) {
...
+	}
+
+	// Arrange: Persist an unused address and a funded output so reopening
+	// must restore both kinds of watch from the store.
+	manager := h.NewWalletManager()
+
+	// Stop the current Manager before harness cleanup stops its replacement
+	// source, including when an assertion aborts the test.
+	defer func() {
+		require.NoError(h, manager.Stop(), "failed to stop manager")
```
yyforyongyu: can you double check whether this `Stop` has already been called inside `t.Cleanup` in `NewWalletManager`?

bwtest/mock/mempool_chain.go
yyforyongyu: Adding NotifySpent to chain.Interface requires every implementation to provide it. MempoolChain has a compile-time interface assertion and is assigned to Wallet.cfg.Chain by the publisher benchmarks in wallet/tx_publisher_benchmark_test.go. This method only preserves interface conformance; the fake does not implement spend notifications, so I kept the unsupported stub.

wallet/syncer.go
yyforyongyu: Added a comment paragraph separator above TODO(yy).

wallet/syncer.go
yyforyongyu: Removed the extra blank line so the existing comment stays attached to NotifyBlocks.

wallet/syncer_test.go
yyforyongyu: Removed the extra blank between the mock declarations.

wallet/syncer_test.go
yyforyongyu: Removed the extra blank between the mock declarations in TestInitChainSync_NotifyBlocksError.

wallet/syncer_test.go
yyforyongyu: Added the blank line before the registration setup comment.

itest/manager_test.go
yyforyongyu: Yes. NewWalletManager registers the Manager with the harness, and harness teardown calls Stop, including after a failed startup. Removed the duplicate deferred stops from all three live-watch cases. The explicit mid-test Stop/DeregisterWallet/ReleaseManager steps remain because they perform the reopen transition under test. Existing cases pass using the harness cleanup.

wallet/syncer.go
```diff
@@ -931,6 +977,13 @@ func (s *syncer) putSyncBatch(ctx context.Context, scanState *RecoveryState,
 		return err
 	}
 
+	refreshWatches := len(params.Horizons) > 0 || len(params.Transactions) > 0
+	if refreshWatches {
+		// Clear readiness before publishing the tip. Normal catch-up
+		// restores it only after watch registration succeeds.
+		s.state.Store(uint32(syncStateSyncing))
```
GustavoStingelin, line 987: Keep incomplete filtered notifications from advancing the scanned tip

This readiness guard only covers `putSyncBatch`. `processChainUpdate` still sends `FilteredBlockConnected` through `putBlockNotifications`, which advances `SyncedTo` for the next height without proving complete wallet coverage. Bitcoind emits that filtered event before `BlockConnected`, so a payment to a recovery-lookahead child outside the backend filter can be skipped.

On `57c42cdc3`, I reproduced this with real SQLite: the ordinary-block control recovered the payment; the incomplete filtered notification produced `finished=true`, `tip=1`, `state=synced`, but public `GetTx` returned `tx not found`. The relevant path remains in `8387599e`; I haven't rerun that probe on the newer head.

This is inherited code, but Task 402 explicitly requires closing this gap. I'd keep the scanner authoritative for the scanned tip and add this regression. That can retain compact-filter matching rather than downloading every block. The earlier [reply saying the filtered writer was removed](https://github.com/btcsuite/btcwallet/pull/1380#discussion_r4025858904) no longer describes this head.

wallet/syncer.go
```diff
@@ -327,6 +363,16 @@ func (s *syncer) initChainSync(ctx context.Context) error {
 		return fmt.Errorf("unable to check for rollback: %w", err)
 	}
 
+	// A retry may follow a committed scan whose registration failed.
+	// Register stored addresses and outpoints even if the tip is current.
+	//
+	// TODO(yy): Avoid rereading and registering the full set on each retry
+	// while preserving registration that failed after a committed scan.
+	err = s.refreshLiveWatches(ctx)
```
GustavoStingelin, line 372: Join watch registration before publishing Start/Create results

This runs inside the background sync worker, while `Wallet.start` immediately marks the Wallet started after launching it. Manager can therefore publish the Wallet before registration finishes or fails. On `57c42cdc3`, holding `WatchAddrsFromTip` blocked still allowed `Manager.Create` to return successfully and `Info` to respond with `Synced=false`. The startup wiring remains unchanged in `8387599e`.

Task 402 requires registration before Start/Create publishes operational Wallets. I'd give both calls a registration completion/failure barrier and test it directly, without making them wait for chain synchronization. Registration currently follows `waitUntilBackendSynced`, so simply waiting for initialization here would also wait for backend sync. Consider establishing the full-node watches before that wait or during synchronous runtime setup, while retaining rollback/catch-up correctness and Neutrino's existing behavior. Full-node registration needs a working backend connection, but not completion of chain sync.

wallet/syncer.go
```diff
@@ -941,6 +994,12 @@ func (s *syncer) putSyncBatch(ctx context.Context, scanState *RecoveryState,
 		return fmt.Errorf("apply sync scan batch: %w", err)
 	}
 
+	// A committed horizon or credit can introduce watches after startup;
+	// finish registration before catch-up reports this batch complete.
+	if refreshWatches {
+		return s.refreshLiveWatches(ctx)
```
GustavoStingelin, line 1003: Propagate lost live readiness to terminal Manager shutdown

Returning a registration error here enters `runSyncLoop`'s retry/backoff path, and `TestLiveWatchScanRetry` explicitly expects service to resume. The unused-address registration error is also only returned to its caller. Stream closure is worse: `ErrWalletShuttingDown` becomes a nil sync result, the controller logs "Chain sync loop exited normally", and the Manager continues serving a Wallet whose sync worker has exited.

The new readiness reset before worker backoff does not cover stream closure: the shutdown-error branch returns nil before storing `syncStateSyncing`. If the tip query still succeeds and matches the stored tip, `Info.Synced` can therefore remain true after the sync worker has exited.

Task 402 has required whole-Manager termination on lost live readiness since the [September 10 roadmap decision](https://github.com/yyforyongyu/btcwallet-sql-roadmap/commit/be10f50488c14b6f69d426ce4df3b3190ef24c57). There is currently no Wallet-to-Manager runtime failure path, so this needs lifecycle wiring as well as error handling. I'd add that propagation without making the failing worker wait on shutdown that joins it, and replace the retry expectation with the required two-Wallet failure/stream-closure coverage. Validation and caller-cancellation errors should remain request-local.

wallet/syncer.go
```diff
@@ -266,6 +266,42 @@ type syncer struct {
...
+// notifications. The Store selects the addresses and outputs to watch.
+func (s *syncer) refreshLiveWatches(ctx context.Context) error {
+	addrs, err := s.storeScanAddresses(ctx)
+	if err != nil {
+		return fmt.Errorf("read watch addresses: %w", err)
+	}
+
+	// Wallet shutdown must still cancel registration. A later sync
+	// initialization reads the persisted watch set again.
+	err = s.cfg.Chain.WatchAddrsFromTip(ctx, addrs)
```
GustavoStingelin, line 283: Preserve Neutrino's existing registration behavior

On a fresh Neutrino client, startup replay now calls `WatchAddrsFromTip` before `NotifyBlocks`. Through `NotifyReceived`, that starts the birthday rescan with every persisted address, where the previous `NotifyBlocks` path started it with an empty address watch set. This adds historical address matching alongside the Wallet syncer's own catch-up.

The per-batch replay then sends the full stored address set again, and repeated `GetUnusedAddress` calls re-register the same unused address. For an active Neutrino rescan, these calls reach `rescan.Update(AddAddrs(...))`. In the pinned v0.18.0, `updateFilter` appends addresses and scripts without deduplication (`rescan.go:1204–1213`), so repeated replay grows memory and filtering work.

I see the earlier feedback asking to remove the Neutrino check and keep backend details behind `chain.Interface`; I wouldn't prescribe reintroducing that check. Adding newly derived addresses to an active scan already happened before this PR. The changes at issue are seeding startup with the persisted set and repeatedly replaying it. Adapter deduplication addresses duplicate accumulation, but doesn't preserve the previous startup behavior.

Task 402 says to preserve Neutrino unchanged, consume Task 401 unchanged, and add no adapter fix or cache. I'd reconcile that scope with the abstraction feedback before choosing the implementation.

chain/interface.go
```diff
@@ -54,6 +54,12 @@ type Interface interface {
 		map[wire.OutPoint]address.Address) error
 	NotifyReceived([]address.Address) error
 
+	// NotifySpent adds outpoints to the full-node transaction watch set.
+	// Notifications may describe unmined or confirmed spends; this does not
+	// promise a complete mempool snapshot. Neutrino treats this as a no-op;
+	// its confirmed spends are discovered by block scanning.
+	NotifySpent([]*wire.OutPoint) error
```
GustavoStingelin, line 61: non-blocking: just thinking loud, its an interface segregation flaw for neutrino and light clients. We should have a better design for this in the future.

GustavoStingelin:
^ posted some AI findings, I'm still processing these comments in my mind, but I do not want to block it for too much time, so I've just submitted. 

Read those as suspicious if they don't make sense for now.

wallet/syncer.go
GustavoStingelin: Following up after checking Neutrino: `WatchAddrsFromTip → NotifyReceived` builds the rescan without a `StartBlock`, so `newRescanState` starts from `chain.BestBlock()` (v0.18.0 rescan.go:393). `StartTime` only decides whether filters are matched from the current block onward (rescan.go:489); it never moves the scan back. So startup registration scans from the tip, not from the birthday, and doesn't overlap the syncer's catch-up. master has the same rescan.go as v0.18.0.

What's left is the lack of dedup in updateFilter: every replay into an active scan appends the full set again, so memory and filter-matching work grow. That isn't a correctness issue, so I'm fine handling it in a follow-up issue.

wallet/syncer.go
GustavoStingelin: nah, think it is not really a problem for now

wallet/syncer.go
GustavoStingelin: can be a follow-up issue

wallet/syncer.go
GustavoStingelin: also follow-up

GustavoStingelin (approved):
I think these comments were raised bc the sync state machine currently is a bit complex to reason about; everyone is setting and clearing the actual sync state. I would say to defer these edge cases to a future rework of this, or even to the multiwallet notification issue.

PR #1382, opened 2026-09-14
wallet: fix end block resolution for bitcoind client

### LINKED ISSUE #1381: [bug]: GetTransactions resolves Bitcoind end block height into start
### Background

`Wallet.GetTransactions` resolves block identifiers into transaction-store heights. When `endBlock` is specified by hash and the wallet uses `BitcoindClient`, the resolved end height is incorrectly assigned to `start`.

Relevant code in `wallet/wallet.go`:

```go
case *chain.BitcoindClient:
    var err error
    start, err = client.GetBlockHeight(endBlock.hash)
```

### Your environment

* version of `btcwallet`: master
* commit: a960541f
* operating system: `<uname -a output>`
* backend: BitcoindClient

### Steps to reproduce

Call `Wallet.GetTransactions` with:

```text
startBlock = height/hash A
endBlock   = hash B
backend    = BitcoindClient
```

where `height(A) != height(B)`.

### Expected behaviour

The resolved range should be:

```text
start = height(A)
end   = height(B)
```

### Actual behaviour

The Bitcoind branch assigns the end block height to `start`, leaving
`end` at its default value of `-1`:

```text
start = height(B)
end   = -1
```

This changes the range passed to `RangeTransactions`.

A regression test and fix will be submitted with the associated pull request.

ginavalent:
## Change Description

Fix `Wallet.GetTransactions` resolving an end block hash through `BitcoindClient`.

The resolved height was assigned to `start` instead of `end`, which could overwrite the caller's requested starting boundary and leave the ending boundary at `-1`.

This change centralizes block range resolution and adds regression coverage for distinct start and end block hashes.

Fixes https://github.com/btcsuite/btcwallet/issues/1381

## Steps to Test

```bash
go test ./wallet
go vet ./wallet
```

The regression test verifies that:

```text
startBlock -> start height
endBlock   -> end height
```

including the case where the start and end heights differ.

### Testing

- [x] The change is not insubstantial.
- [x] The change obeys the Code Documentation and Commenting guidelines, and lines wrap at 80.
- [x] Commits follow the Ideal Git Commit Structure.


wallet/wallet_test.go
```diff
@@ -315,6 +315,48 @@ func TestGetTransaction(t *testing.T) {
 	}
 }
 
+// TestResolveBlockRangeBitcoindEndHash ensures hash-based start and end
+// identifiers are assigned to their respective range boundaries.
+func TestResolveBlockRangeBitcoindEndHash(t *testing.T) {
```
Lrifton92, line 320: This calls `resolveBlockRange` with a fake resolver, so it covers the helper but not the type switch in `GetTransactions`, which is where the original bug lived. I checked both directions:

- reintroducing the bug inside the helper (`start, err = resolveHash(endBlock.hash)` on the end branch) fails the test with `expected: 20, actual: 10`, good;
- deleting `resolveHash = client.GetBlockHeight` from the `*chain.BitcoindClient` case still passes, and in production that would silently return the whole history (see the comment on `resolveBlockRange`).

The structural change is what really prevents a repeat, so I would not insist on a backend-level test, but the name promises a Bitcoind check the test does not perform; `TestResolveBlockRange` would be accurate. The error case also only covers a failing start hash; since the bug was on the end side, one more call with `nil, NewBlockIdentifierFromHash(&endHash)` and a failing resolver would pin that too.

wallet/wallet.go
```diff
@@ -2561,6 +2561,40 @@ func NewBlockIdentifierFromHash(hash *chainhash.Hash) *BlockIdentifier {
...
+// the transaction store. Hashes are resolved through resolveHash, while
+// height-based identifiers are used directly.
+func resolveBlockRange(startBlock, endBlock *BlockIdentifier,
+	resolveHash func(*chainhash.Hash) (int32, error)) (int32, int32, error) {
+	var start, end int32 = 0, -1
+
+	if startBlock != nil {
+		if startBlock.hash == nil {
+			start = startBlock.height
+		} else if resolveHash != nil {
```
Lrifton92, line 2574: `else if resolveHash != nil` keeps master's behaviour for an unrecognised client type: the hash identifier is dropped and the boundary stays at 0 / -1, so the caller gets the full range with no error. That is pre-existing, not introduced here, but the helper now makes it an explicit contract, and it is exactly what turns a wiring mistake in the switch into wrong results instead of an error.

The only in-tree `chain.Interface` implementations are `RPCClient`, `BitcoindClient`, `NeutrinoClient` and `mockChainClient`, and no test calls `GetTransactions` at all, so a `default:` case returning an error in the switch looks safe; `resolveHash` would then always be non-nil when a hash is present and both nil checks could go. Fine as a follow-up if you prefer to keep this PR minimal. If the fallback stays, the doc comment ("Hashes are resolved through resolveHash") should mention the nil case.

Lrifton92 (commented):
The fix is correct, and the refactor is the right shape for it: with a single `resolveHash` feeding both boundaries, the copy-paste that caused #1381 (`start, err = client.GetBlockHeight(endBlock.hash)`, `wallet/wallet.go:2643` on master) can no longer happen per backend.

I walked every input combination against master (`a960541f`) and found no behaviour change other than the intended one:

| start / end | master | this PR |
|---|---|---|
| nil / nil | 0, -1 | 0, -1 |
| height / height (incl. end = -1) | passed through, no client needed | same |
| hash, nil chain client | `no chain server client` | same (check hoisted, still only when a hash is present) |
| hash, `RPCClient` | `GetBlockHeaderVerbose().Height`, error returned | same |
| hash, `NeutrinoClient` | `GetBlockHeight` on both sides | same |
| end hash, `BitcoindClient` | **start = h(end), end = -1** | start preserved, end = h(end) |
| unknown hash | backend error returned, start resolved first | same |
| hash, unrecognised client type | hash silently ignored | same |
| start > end | forwarded to `RangeTransactions` (iterates backwards) | same |

`go build ./...` passes and `TestResolveBlockRangeBitcoindEndHash` passes at `3b6b084`.

### Lint will fail once CI is approved

The CI run for this head is still `action_required`, so nothing has actually run yet. `make lint-check` is part of the workflow; running `golangci-lint` built from `tools/go.mod` with the repo's `.golangci.yml` reports 9 new issues on this head and 0 on master:

    wallet/wallet.go:2568       whitespace  (blank line after multi-line signature)
    wallet/wallet.go:2617       whitespace  (the blank line removed from GetTransactions)
    wallet/wallet.go:2627       whitespace  (multi-line if condition)
    wallet/wallet.go:2576       wsl_v5
    wallet/wallet.go:2588       wsl_v5
    wallet/wallet.go:2639       wsl_v5, nlreturn
    wallet/wallet.go:2654       wsl_v5
    wallet/wallet_test.go:352   perfsprint (fmt.Errorf -> errors.New)

All mechanical. Two smaller points inline, neither blocking on its own.

PR #1383, opened 2026-09-15
itest: cover unsafe signer extraction

yyforyongyu:
## Summary

Implement Task 443.

Cover the existing UnsafeSigner extraction contracts through real wallets.

## Change Description

Add six integration cases for public-key identity, missing accounts and foreign
addresses, and locked/watch-only rejection. Use direct Wallet calls, explicit
key-presence assertions, and a documented harness fixture for named sequential
accounts with returned account numbers.

All 54 focused case executions pass across kvdb, SQLite, and PostgreSQL with
btcd, bitcoind, and neutrino. Formatting and the integration linter pass.


itest/signer_test.go
```diff
@@ -737,3 +738,225 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
 		"the reopened wallet derived a different imported child",
 	)
 }
+
+// unsafeSignerPublicKey consumes an extraction result before any assertion can
+// retain or print it, including an unexpected key returned alongside an error.
+func unsafeSignerPublicKey(key *btcec.PrivateKey) *btcec.PublicKey {
```
yyforyongyu: i don't think we need this shallow method - we can just call `key.PubKey()` and `key.Zero()` directly in the tests?

itest/signer_test.go
```diff
@@ -737,3 +738,225 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	return key.PubKey()
+}
+
+// testUnsafeSignerDerivePrivKey checks that extraction matches public
+// derivation and preserves the same identity across a durable wallet reload.
+func testUnsafeSignerDerivePrivKey(h *bwtest.HarnessTest) {
+	// Arrange: Materialize the default account through the existing fixture
+	// and obtain an independent public oracle for a nonzero child index.
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	h.NewWalletAddressOfType(w, waddrmgr.WitnessPubKey)
```
yyforyongyu: why calling this when the returned addr is not used?

itest/signer_test.go
```diff
@@ -737,3 +738,225 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	path := wallet.BIP32Path{
+		KeyScope: waddrmgr.KeyScopeBIP0084,
+		DerivationPath: waddrmgr.DerivationPath{
+			InternalAccount: waddrmgr.DefaultAccountNum,
+			Branch:          params.Branch,
+			Index:           params.Index,
+		},
+	}
+
+	var signer wallet.UnsafeSigner = w
```
yyforyongyu: what is this? explain this assignment?

itest/signer_test.go
```diff
@@ -737,3 +738,225 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+// against public address metadata before and after reopening the wallet.
+func testUnsafeSignerGetPrivKeyForAddress(h *bwtest.HarnessTest) {
+	// Arrange: Use an owned witness address whose public metadata provides an
+	// oracle without calling any private-key extraction method.
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	addr := h.NewWalletAddressOfType(w, waddrmgr.WitnessPubKey)
+	info, err := w.GetAddressInfo(h.Context(), addr)
+	require.NoError(h, err)
+
+	var signer wallet.UnsafeSigner = w
```
yyforyongyu: again i don't think we need this?

itest/signer_test.go
```diff
@@ -737,3 +738,225 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	require.NotNil(h, pubKey)
+	require.Equal(
+		h, info.PubKey.SerializeCompressed(), pubKey.SerializeCompressed(),
+	)
+
+	// Arrange: Keep the address and public oracle across the harness-owned
+	// close/reopen boundary, then unlock the replacement for the same query.
+	w = h.ReloadWallet(w)
+	h.UnlockWallet(w)
+	signer = w
```
yyforyongyu: why do we need this reassignment??

itest/signer_test.go
```diff
@@ -737,3 +738,225 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	)
+}
+
+// testUnsafeSignerRejectUnknownPath checks the stable absence error for an
+// uncreated account in an otherwise materialized scope.
+func testUnsafeSignerRejectUnknownPath(h *bwtest.HarnessTest) {
+	// Arrange: Only account zero exists in this scope. Request account one
+	// while unlocked so the result exercises absence, not the signing gate.
+	w, _ := h.NewWallet(bwtest.WalletFixture{Unlocked: true})
+	h.NewWalletAddressOfType(w, waddrmgr.WitnessPubKey)
```
yyforyongyu: again the returned addr is not used??

itest/signer_test.go
yyforyongyu: Updated in 9048cacee: removed unsafeSignerPublicKey and all Zero calls. Success cases explicitly require a returned key and compare its public identity inline. Rejection cases explicitly require nil using boolean assertions so failures cannot print private material.

itest/signer_test.go
yyforyongyu: Updated in 9048cacee: the path test uses h.CreateTestAccount to create one named account unconditionally and uses the returned account number. It no longer derives an unused address or branches on default-account existence.

itest/signer_test.go
yyforyongyu: Removed all six `wallet.UnsafeSigner` aliases. The tests call the extraction methods directly on `w`; the existing production compile-time assertion already checks interface implementation.

itest/signer_test.go
yyforyongyu: Removed the alias. The test now makes one direct w.GetPrivKeyForAddress call and compares its public identity with the independent address metadata.

itest/signer_test.go
yyforyongyu: Removed the reassignment along with the alias. The test now makes one extraction call directly on w, so there is no replacement wallet to assign.

itest/signer_test.go
yyforyongyu: Updated in 9048cacee: the test uses h.CreateTestAccount to create one named account unconditionally and requests the account number immediately following the returned number. The unused-address fixture and conditional default-account setup are both removed.

itest/signer_test.go
```diff
@@ -737,3 +740,246 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+
+	hasKey := key != nil
+	require.True(h, hasKey, "expected an extracted key")
+	require.Equal(
+		h, want.SerializeCompressed(), key.PubKey().SerializeCompressed(),
+	)
+
+	// Arrange: Reopen the same persistent wallet through its lifecycle owner.
+	// The replacement starts locked, so unlock it before extraction.
+	w = h.ReloadWallet(w)
```
yyforyongyu: why reload the wallet? we are not testing restart behavior here??

itest/signer_test.go
```diff
@@ -737,3 +740,246 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	hasKey := key != nil
+	require.True(h, hasKey, "expected an extracted key")
+	require.Equal(
+		h, info.PubKey.SerializeCompressed(),
+		key.PubKey().SerializeCompressed(),
+	)
+
+	// Arrange: Keep the address and public oracle across the harness-owned
+	// close/reopen boundary, then unlock the replacement for the same query.
+	w = h.ReloadWallet(w)
```
yyforyongyu: same re reload, why do we do this?

itest/signer_test.go
yyforyongyu: Removed the reload and repeated extraction in 9048cacee. The case now creates one named account through h.CreateTestAccount, performs one DerivePrivKey extraction, and compares its public identity with the independent DerivePubKey oracle.

itest/signer_test.go
yyforyongyu: Removed the reload and repeated extraction in 9048cacee. The case now performs one GetPrivKeyForAddress extraction and compares its public identity with GetAddressInfo metadata.

itest/signer_test.go
```diff
@@ -737,3 +737,198 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	)
+	require.NoError(h, err, "failed to encode the foreign address")
+
+	// Act: Request both forms of extraction from the watch-only wallet to
+	// verify that neither entry point bypasses the signing-state gate.
+	pathKey, pathErr := w.DerivePrivKey(h.Context(), path)
+	addrKey, addrErr := w.GetPrivKeyForAddress(h.Context(), addr)
+
+	// Assert: The maintained Wallet refuses both requests with its existing
+	// state sentinel and exports no key, independently of the storage backend.
```
Lrifton92, line 920: This case reaches the same gate as `testUnsafeSignerRejectLocked`: the fixture is never unlocked, and `canSign()` only checks lifecycle and lock state (`wallet/state.go:270-281`), not watch-only mode. "Independently of the storage backend" is true only because the wallet stays locked. I checked what happens if it is unlocked: kvdb refuses to unlock a watch-only wallet ("address manager is watching-only"), while on sqlite the unlock succeeds and the two calls then return `ErrAccountNotInStore` and `ErrNoAssocPrivateKey` rather than `ErrStateForbidden`.

Asserting the locked state is fine as the common denominator, but it would help to say so explicitly, the way `testSignerECDHWatchOnly` does at lines 565-571, so a reader does not take this as coverage of a watch-only-specific refusal. The doc comment at 901-902 ("at the existing signing-state boundary") has the same ambiguity.

itest/signer_test.go
```diff
@@ -737,3 +737,198 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	// Act: Extract the private key for the same account path as the public
+	// oracle so the two results can be compared independently.
+	key, err := w.DerivePrivKey(h.Context(), path)
+
+	// Assert: Require a key through a boolean check so a failure cannot print
+	// private material, then compare its public identity with the oracle.
+	require.NoError(h, err)
+
+	hasKey := key != nil
+	require.True(h, hasKey, "expected an extracted key")
```
Lrifton92: `DerivePrivKey` and `GetPrivKeyForAddress` document that "the caller must zero the key" (`wallet/signer.go:1512-1513`, `1542-1543`). The earlier thread suggested calling `key.PubKey()` and `key.Zero()` directly; the update dropped the helper and the `Zero` calls together. A `defer key.Zero()` right after the non-nil check here and at line 806 would keep the tests consistent with the contract they cover. Test keys only, so a nit.

Lrifton92 (commented):
Test-only change; I checked each case against the implementation at 7b7a92a and ran the new cases locally.

What I verified:

- `go build ./...` and `go vet -tags="itest dev nolog" ./itest ./bwtest` are clean.
- With btcd v0.26.0 (the go.mod version), the six `unsafe_signer` cases pass on btcd/kvdb and btcd/sqlite. The full `(unsafe_signer|signer)` group passes on btcd/sqlite with `-count=5` and on neutrino/kvdb.
- `derive privkey` / `get privkey for address`: the oracles are independent of the extraction path (`DerivePubKey` walks the account xpub, `GetAddressInfo` reads stored address metadata), so the public-key comparisons are meaningful.
- `reject unknown path`: `resolveDerivedPrivKeyFromStore` maps `db.ErrAccountNotFound` to `ErrAccountNotInStore` (`wallet/signer.go:1195-1198`), which is what the case asserts. Only the account component can be absent here, so "unknown account" would describe it more precisely than "unknown path", but that is cosmetic.
- `reject foreign address`: the generator point resolves to `ErrAddressNotFound`, mapped to `ErrNoAssocPrivateKey` at `wallet/signer.go:1629-1632`.
- `reject locked`: both entry points stop at `canSign()` (`wallet/state.go:276-278`). I also confirmed with a throwaway case that the same `path` and `addr` extract successfully once the wallet is unlocked (kvdb and sqlite), so the "valid owned inputs" claim in the comment holds and the lock is the only variable.
- `CreateTestAccount`: `NewAccount` requires `canSign()` for non-watch-only wallets (`wallet/account_manager.go:541-546`) and never changes lock state, so the doc comment is accurate.

Two points inline. Neither blocks.

One coverage gap, noted for completeness rather than as a request: the owned-but-no-private-key branches of `privKeyForAddress` (the imported-xpub child at `wallet/signer.go:1611-1618` and `resolveImportedAddrPrivKey`) are not exercised. I understand they are not reachable uniformly: SQL wallets only accept xpub/pubkey imports in watch-only mode, and kvdb cannot unlock a watch-only wallet.

itest/signer_test.go
GustavoStingelin: I don't think zeroing is necessary for these generated test keys, since they don't protect real funds.

itest/signer_test.go
```diff
@@ -737,3 +737,198 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+
+	// Act: Extract the private key for the same account path as the public
+	// oracle so the two results can be compared independently.
+	key, err := w.DerivePrivKey(h.Context(), path)
+
+	// Assert: Require a key through a boolean check so a failure cannot print
+	// private material, then compare its public identity with the oracle.
+	require.NoError(h, err)
+
+	hasKey := key != nil
```
GustavoStingelin: nit: I'd use `require.NotNil(h, key, "expected an extracted key")` directly. It expresses the same assertion more clearly and only fails when the key is nil, so the failure cannot expose private-key material.

itest/signer_test.go
```diff
@@ -737,3 +737,198 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	// Act: Resolve the owned address through the wallet to compare the
+	// extracted key with the independent public address metadata.
+	key, err := w.GetPrivKeyForAddress(h.Context(), addr)
+
+	// Assert: Check presence as a boolean to keep private material out of
+	// diagnostics, then compare the key's public identity with the metadata.
+	require.NoError(h, err)
+
+	hasKey := key != nil
+	require.True(h, hasKey, "expected an extracted key")
```
GustavoStingelin: nit: Same here: `require.NotNil(h, key, "expected an extracted key")` avoids the temporary boolean and only fails for a nil key.

Same comment for all tests using this pattern. There is no important material to leak.

GustavoStingelin (changes requested):
looking good, just leaving some nits about the nil assertion.

Is it also realated to some task in the roadmap? if yes, I think it is missing a link in the PR description.

itest/signer_test.go
yyforyongyu: No code change here. This case intentionally verifies the common locked signing-state boundary for a watch-only fixture. The setup comment and IsWatchOnly assertion make that fixture and precondition explicit, so we are keeping the case focused on the maintained cross-backend contract.

itest/signer_test.go
yyforyongyu: No code change here. These are generated test keys with no funds, and the repository owner explicitly waived Zero calls for this test.

itest/signer_test.go
yyforyongyu: Updated in 1122be589: the successful extraction now uses require.NotNil directly before comparing the public identity. This assertion reports nilness without formatting private-key material.

itest/signer_test.go
yyforyongyu: Updated in 1122be589: both successful extraction sites now use require.NotNil directly. Rejection cases retain boolean nil assertions because require.Nil would format an unexpected non-nil private key on failure.

yyforyongyu:
Addressed the successful-extraction assertion feedback in 1122be589: both sites now use `require.NotNil` before public-key comparison. The rejection paths keep boolean nil checks so a failure cannot format a returned private key. I am leaving roadmap and local planning references out of the PR description per project direction.

Abdulkbk (approved):
lgtm

GustavoStingelin (approved):

PR #1384, opened 2026-09-15
itest: cover public signing contracts

yyforyongyu:
## Summary

Implement Task 442.

Prove the public Wallet signing contracts across kvdb, SQLite, and
PostgreSQL through shared integration tests.

## Change Description

Cover digest signing, standard unlocking scripts, and caller-assembled raw
signatures, including ECDSA, Schnorr, Taproot key and script paths, supported
tweaks, and public error identities. Verify locked and watch-only rejection
and signing after a persistent Wallet reopen.

Verify signatures with public keys and execute assembled spends with the
script engine. Reuse the existing harness and backend registrations.

Clarify the public BIP32 account selector and return ErrNilArguments for
missing outputs required by raw signing, preventing a Wallet panic. Preserve
legacy signing with an explicit redeem script and no output. Keep these
production corrections together in one commit with regression coverage.


itest/signer_test.go
```diff
@@ -737,3 +742,220 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
 		"the reopened wallet derived a different imported child",
 	)
 }
+
+// signerPath converts public address metadata into the full signing selector,
+// retaining the account number and child rather than assuming address index 0.
+func signerPath(derivation *wallet.AddressDerivation) wallet.BIP32Path {
```
yyforyongyu: this should be `createSignerPath`?

itest/signer_test.go
```diff
@@ -959,3 +962,191 @@ func testSignerRejectDigestIntent(h *bwtest.HarnessTest) {
 		})
 	}
 }
+
+// signerTransaction builds a deterministic unsigned spend without chain state.
+// Signing needs the previous output's script and amount, not a mined coin.
+func signerTransaction(pkScript []byte) (*wire.TxOut, *wire.MsgTx) {
```
yyforyongyu: call it `createSignerTx`

itest/signer_test.go
yyforyongyu: Renamed to `createSignerPath`, including the declaration comment and every caller. Folded into 10d3a59475; the path construction is unchanged.

itest/signer_test.go
yyforyongyu: Renamed to `createSignerTx`, including the declaration comment and every caller. Folded into b5178f00f1; the transaction fixture is unchanged.

itest/fixtures_test.go
```diff
@@ -108,20 +109,33 @@ func deterministicImportedAccountKeys(
...
+
+	defer otherPrivateKey.Zero()
+
 	otherAccountKey, err := otherPrivateKey.Neuter()
 	require.NoError(h, err, "failed to derive second imported public key")
 
+	// The second account needs only public material, so detach it before
+	// its private key is zeroed on return, including assertion-failure paths.
+	otherAccountKey, err = hdkeychain.NewKeyFromString(otherAccountKey.String())
+	require.NoError(h, err, "failed to copy second imported public key")
```
GustavoStingelin: I think we can simplify this fixture. These keys are derived from a fixed test seed, so wiping them doesn't protect user secrets or verify the signing contract.

The serialization round-trips are needed because `Neuter` shares buffers with the private key being zeroed. I'd remove the fixture zeroing/cleanup and those copies together, including the early `keys.accountPrivateKey.Zero()` in the watch-only test. Constructing the Wallet with only the XPub establishes its watch-only state; wiping the separate fixture key adds no coverage.

itest/signer_test.go
```diff
@@ -737,3 +746,965 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+			require.Equal(t, tc.wantSigScript, len(unlocking.SigScript) != 0)
+			tx.TxIn[0].Witness = unlocking.Witness
+			tx.TxIn[0].SignatureScript = unlocking.SigScript
+
+			engine, err := txscript.NewEngine(
+				pkScript, tx, 0, txscript.StandardVerifyFlags,
+				nil, hashes, prevOut.Value, fetcher,
+			)
+			require.NoError(t, err)
+			require.NoError(t, engine.Execute())
```
GustavoStingelin: I think we should assert the requested sighash mode here. The script engine reads the mode from the returned signature, so a valid spend can pass even if `params.HashType` was ignored.

On SQLite, replacing `params.HashType` with `txscript.SigHashAll` in the non-Taproot signing branches still passes all five rows, including `witness none` and `nested single anyonecanpay`. An assertion comparing the signature's final byte with `byte(tc.hashType)` catches both.

I'd keep the engine verification and add the exact-mode assertion, including Taproot's 64-byte DEFAULT / 65-byte explicit-mode encoding. The raw legacy and explicit raw Taproot cases need the same check.

itest/signer_test.go
```diff
@@ -737,3 +746,965 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+func createSignerPath(derivation *wallet.AddressDerivation) wallet.BIP32Path {
+	// Account is hardened in the BIP32 path; InternalAccount is the portable
+	// account number used to resolve its persistent signing material.
+	account := hdkeychain.HardenedKeyStart + derivation.Account
+
+	return wallet.BIP32Path{
+		KeyScope: derivation.KeyScope,
+		DerivationPath: waddrmgr.DerivationPath{
+			InternalAccount:      derivation.Account,
+			Account:              account,
```
GustavoStingelin, line 952: I think this exposes an inconsistency in the public signing contract. `waddrmgr.DerivationPath.InternalAccount` is documented and used by `ScopedKeyManager.DeriveFromKeyPath` as a legacy database lookup identity, but the Wallet signer passes it as `GetAccountSecretQuery.AccountNumber`, meaning a BIP44 account number.

[ADR 0013's Store Adapter Identity section](https://github.com/btcsuite/btcwallet/blob/3ccded935a7c0d9fc6c60dba9cd791633f0ff6c7/docs/developer/adr/0013-normalized-account-address-identity.md#store-adapter-identity) explicitly separates those identities: a derived kvdb account's internal ID may equal its account number numerically, but that does not make them interchangeable. The ADR should be the authority here; backend IDs should stay inside the adapter boundary.

I'd clarify the Wallet-specific meaning in the public `BIP32Path` documentation and have this helper follow that contract. The hardening calculation is consistent for these derived accounts; the concern is the different meaning assigned to `InternalAccount`. A comment only in the test would leave callers facing the same ambiguity.

GustavoStingelin (changes requested):

itest/signer_test.go
```diff
@@ -737,3 +746,965 @@ func testSignerDeriveImportedXPub(h *bwtest.HarnessTest) {
...
+	info, err := w.GetAddressInfo(h.Context(), addr)
+	require.NoError(h, err)
+
+	pkScript, err := txscript.PayToAddrScript(addr)
+	require.NoError(h, err)
+
+	prevOut, tx := createSignerTx(pkScript)
+	fetcher := txscript.NewCannedPrevOutputFetcher(pkScript, prevOut.Value)
+	hashes := txscript.NewTxSigHashes(tx, fetcher)
+	tests := []struct {
```
Abdulkbk, line 1606: I noticed that we don't have a row for `nil output` and that's because `ComputeRawSig` doesn't check for it to return an error. It panics when `output` is nil.

I think we can fix this here, add the `nil-check` for `output` in ComputeRawSig, and add a row testing that here.

Abdulkbk (commented):
Left a suggestion.

itest/fixtures_test.go
yyforyongyu: Restored the canonical fixture to its existing implementation and removed the watch-only test's early Zero call. The added account-key cleanup and both serialization copies are gone; the Wallet still receives only the canonical XPub. Folded into a138268b57. The imported-account and watch-only cases pass across all nine chain/database combinations.

itest/signer_test.go
yyforyongyu: Documented on BIP32Path.DerivationPath that Wallet interprets InternalAccount as the unhardened BIP44 account number, not a backend account ID (edd7a43168). The test helper now points to that public contract; its existing assignment from AddressDerivation.Account and hardening arithmetic are unchanged.

itest/signer_test.go
yyforyongyu: Confirmed the panic through a real SQLite Wallet and fixed it in edd7a43168. Missing outputs now return wrapped ErrNilArguments in the spend implementations that require them. The existing rejection table covers legacy fallback, Segwit, and both Taproot paths. Validation stays local because legacy signing with an explicit RedeemScript already works without Output; the existing P2SH unit test now verifies that signature with Output omitted. Wallet tests and the focused integration cases pass on SQLite, PostgreSQL, and kvdb across all three chain backends.

itest/signer_test.go
yyforyongyu: Also note that the legacy lookup account identity **is the account number** for a wallet-derived account, and it may only be different for imported accounts. This is also not sth we wanna spend too much time on - the legacy path will no longer be maintained, so we should not let the legacy behavior shape our current code too much.

Abdulkbk (approved):
LGTM

GustavoStingelin (approved):

PR #1385, opened 2026-09-15
wallet: classify initial account input errors

yyforyongyu:
## Summary

Give invalid initial-account creation inputs a stable `ErrWalletParams`
identity. Follow-up to #1328.

## Change Description

Wrap the existing key/network and address-type validation errors while
preserving their causes. Update the existing regression to use `ErrorIs`
instead of matching messages.

One commit, two files: 3 additions and 6 deletions. The signer-fixture repair
is now in `sql-wallet` and is no longer part of this PR.

Validation: wallet package tests, tagged integration compilation, wallet lint,
and commit/Go audits pass on the rebased revision.


Lrifton92 (commented):
Small and consistent with the rest of `createWallet`: the other parameter checks in `params.validate()` already return `ErrWalletParams`-wrapped errors, so the two pre-persistence account checks now get the same identity while keeping their cause via the double `%w` (fine with `go 1.25.11` in go.mod). I ran `TestManagerCreateRejectsInvalidInitialAccounts` at 8ebe27c and both cases pass.

One suggestion on the test: `require.ErrorIs(t, err, ErrWalletParams)` alone no longer pins *which* check rejected the input, since both cases now satisfy the same sentinel. If the wrong-network key were ever rejected by a different check, or the two causes got swapped, the test would still pass. Keeping the cause assertion next to `ErrorIs` would preserve what the old test covered:

- wrong network: `validateExtendedPubKey` already wraps `ErrInvalidAccountKey` ("expected extended public key for current network"), so `require.ErrorIs(t, err, ErrInvalidAccountKey)` works for that case;
- unsupported address type: `keyScopeFromPubKey` returns an unwrapped `fmt.Errorf("unsupported address type %v", ...)`, so keeping `wantErr` with `require.ErrorContains` is the only way to assert it.

Not blocking. LGTM otherwise.

Abdulkbk (approved):
lgtm

PR #1386, opened 2026-09-15
wallet: add durable key-only allocation

yyforyongyu:
## Summary

Implement task 395: persist the next child key and return its public key, branch, index and available wallet-root origin.

## Change Description

Return `AllocatedKey` with an optional `KeyOrigin`. Named imported-xpub accounts can allocate with a nil origin while retaining their child coordinates. Wallet-derived accounts also return scope, account number and fingerprint. Existing signing and PSBT metadata remain unchanged.

Use one Store call and native transaction. SQL single-address allocation reuses the durable batch allocator for collision skipping, invalid children and exhaustion. Kvdb preserves its native metadata and restricts numeric selectors to wallet-derived accounts. Remove the unused singular workflow in a separate cleanup commit.

## Notes

Stacked on #1379 through upstream `task-bulk-address`. Review scope begins after that dependency and contains five commits, including separate SQL singular-allocation and obsolete-workflow fixes. Account creation and broader recovery filtering remain separate prerequisites.


wallet/internal/db/itest/addressstore_test.go
```diff
@@ -3417,9 +3417,10 @@ func TestNewDerivedAddressesDurableBatches(t *testing.T) {
 	t.Parallel()
 
 	for _, tc := range []struct {
-		name   string
-		count  uint32
-		change bool
+		name        string
+		count       uint32
+		change      bool
+		noChainSync bool
```
yyforyongyu: we might as well just create a standalone test for this `noChainSync` than expanding current tests?

wallet/internal/db/data_types.go
```diff
@@ -856,6 +856,11 @@ type NewDerivedAddressParams struct {
 	// receiving address. The false default preserves internal key and change
 	// allocation; this requirement never changes the stored account policy.
 	RequireChainSync bool
+
+	// RequireNoChainSync admits only key-only accounts with a wallet-root
+	// locator. It checks persisted policy without changing it or enabling
+	// receiving behavior for callers that allocate keys.
+	RequireNoChainSync bool
```
yyforyongyu: why do we even need this? we already have `RequireChainSync`, and the addr's account defines the nochainsync?

wallet/internal/db/itest/addressstore_test.go
yyforyongyu: Addressed in ccf2a1087. Restored the existing SQL batch tests unchanged and put account-policy admission coverage in the standalone TestAllocateNextKeyRejectsAccount test in wallet/address_manager_test.go.

wallet/internal/db/data_types.go
yyforyongyu: Removed RequireNoChainSync without adding a replacement flag in ccf2a1087. AllocateNextKey now reads the account and rejects SQL accounts without NoChainSync before allocation, then allocates by the resolved account number so a rename cannot change the selected account. RequireChainSync is unchanged. Kvdb uses its existing root-derived account and allocation semantics.

wallet/address_manager.go
```diff
@@ -699,6 +716,137 @@ func (w *Wallet) NewAddress(ctx context.Context, accountName string,
...
+		SkipBalance:   true,
+	}
+
+	account, err := w.store.GetAccount(r.ctx, query)
+	if err != nil {
+		r.respChan <- addressInfoResp{err: err}
+		return
+	}
+
+	if account.IsImported || account.AccountNumber == nil ||
```
yyforyongyu: why we don't allow this method for imported account?

wallet/address_manager.go
```diff
@@ -699,6 +716,137 @@ func (w *Wallet) NewAddress(ctx context.Context, accountName string,
...
+	// transactions. Kvdb does not need the SQL batch contract for one key.
+	params := db.NewDerivedAddressParams{
+		WalletID:      w.id,
+		Scope:         query.Scope,
+		AccountNumber: account.AccountNumber,
+		Change:        r.internal,
+	}
+
+	var stored *db.AddressInfo
+	if w.addrStore != nil {
```
yyforyongyu: this is def abstraction leak?

wallet/address_manager_test.go
```diff
@@ -2008,3 +2009,403 @@ func TestNewBulkAddressesReplaysCommittedBatch(t *testing.T) {
...
+			HasDerivationPath:    true,
+			Branch:               branch,
+			Index:                19,
+			ScriptPubKey:         script,
+			PubKey:               pubKey,
+		}
+		legacy := stored
+
+		legacy.MasterKeyFingerprint = 0
+		for _, tc := range []struct {
```
yyforyongyu: simplify this test - we are using two `for`s, which is bad as it's hard to track!

wallet/address_manager.go
```diff
@@ -699,6 +716,137 @@ func (w *Wallet) NewAddress(ctx context.Context, accountName string,
...
+		default:
+			err = bulkAddressErr(result.err)
+		}
+
+		return nil, nil, err
+	}
+
+	// Legacy rows may omit a fingerprint; reuse the canonical root identity
+	// already cached for public derived account metadata.
+	if w.addrStore != nil {
```
yyforyongyu: i don't think we need this - for deprecated kvdb, we can just let it be empty?

wallet/address_manager.go
yyforyongyu: Removed the blanket imported-account check in 4bbe743da598. Imported xpubs can derive public children; the remaining restriction is missing locator metadata: the current Store exposes no wallet-derived AccountNumber for them, so the existing conversion cannot return AddressDerivation. We reject that missing-number shape before consuming a child rather than inventing a root locator. Ordinary derived accounts are accepted regardless of NoChainSync.

wallet/address_manager.go
yyforyongyu: Fixed in 933535f1e5dd and 4bbe743da598. AllocateNextKey now always calls Store.NewDerivedAddresses with count one, with no w.addrStore check. The kvdb implementation delegates that request to its existing NewDerivedAddress transaction; other counts remain unsupported. No new Store policy flag or NoChainSync restriction is added.

wallet/address_manager_test.go
yyforyongyu: Flattened TestAllocateNextKeyReturnsLocator into one explicit table in 4bbe743da598. Each case has its own Arrange/Act/Assert flow and the same Store expectations, covering branch, selector, scanning policy, and preserved fingerprint without nested case loops.

wallet/address_manager.go
yyforyongyu: Agreed. Removed the fingerprint override in 4bbe743da598 and kept deprecated kvdb allocation and lookup metadata unchanged, including zero fingerprints. The reopen regression in 7eeaa6ea6122 compares the returned derivation directly with ordinary persisted lookup.

wallet/address_manager.go
```diff
@@ -175,6 +175,14 @@ type OutputScriptInfo struct {
 // AddressManager provides an interface for generating and inspecting wallet
 // addresses and scripts.
 type AddressManager interface {
+	// AllocateNextKey commits a fresh child without automatic chain tracking
+	// and returns its public key and locator. Accounts must expose a derived
+	// account number; NoChainSync does not restrict allocation. Metadata is
+	// preserved, including possibly zero legacy kvdb fingerprints. Errors
+	// never imply an index can be reused.
```
yyforyongyu: this is too verbose, i think i wanna mention these in the docs,
- what this method does, should be simple, create and persist next key
- other details should be discarded

wallet/address_manager.go
```diff
@@ -699,6 +717,123 @@ func (w *Wallet) NewAddress(ctx context.Context, accountName string,
 	return result.addr, result.err
 }
 
+// AllocateNextKey commits a fresh external or internal child of an existing
+// account with a wallet-derived account number and returns its public key and
+// locator, regardless of NoChainSync. Imported XPub accounts currently expose
```
yyforyongyu: if it's `regardless of NoChainSync`, then there is no need to even mention that? if sth is not affecting this method, why mention that?

wallet/address_manager.go
yyforyongyu: Simplified both AllocateNextKey API comments in e3e62e2af7a3 to state that the method creates and persists the next child key and returns its public key and locator. Removed the backend, policy, and error-detail paragraphs from those summaries; necessary implementation rationale stays beside the relevant code.

wallet/address_manager.go
yyforyongyu: Removed the NoChainSync wording from both API comments in e3e62e2af7a3, with no replacement contrast sentence. Allocation eligibility still covers ordinary and excluded accounts in the real Store regressions, and the public API makes one allocation request without a receiving-policy requirement.

wallet/internal/db/data_types.go
```diff
@@ -832,6 +832,10 @@ type AddressSecret struct {
 // NewDerivedAddressParams contains the parameters for creating a new derived
 // address.
 type NewDerivedAddressParams struct {
+	// RequireAccountNumber rejects accounts without a wallet-derived number
+	// before consuming a child, when the caller needs a root key locator.
+	RequireAccountNumber bool
```
yyforyongyu: why do we need this? we can just return an error if the callsite is asking to derive from a non-derivable account?

wallet/address_manager_test.go
```diff
@@ -2238,3 +2238,109 @@ func TestAllocateNextKeyMapsStoreFailure(t *testing.T) {
...
+	}
+
+	const perBranch = 3
+
+	results := make([]allocation, 2*perBranch)
+
+	// Act: Allocate several still-unused children per branch. Each worker owns
+	// one result slot and is joined before assertions or database teardown.
+	var workers sync.WaitGroup
+	for i := range results {
```
yyforyongyu: can we use the new `synctest` lib here?

wallet/internal/db/data_types.go
yyforyongyu: The distinction is that an imported xpub account *is derivable*, but has no wallet-root account number for `AddressDerivation`. The shared `NewDerivedAddress` method must still allow ordinary receiving from that account.

`AllocateNextKey` sets this requirement so the Store returns `ErrAccountOperationUnsupported` before consuming a child, inside the same allocation transaction. Rejecting all such accounts would break imported-xpub receiving; checking the result afterward would consume a child without returning a usable locator. The existing SQL and kvdb tests verify rejection first, followed by successful ordinary allocation at child zero. I kept the flag and the single Store call.

wallet/address_manager_test.go
yyforyongyu: Done in 0f17dbcfcb43355345feb146389c1a86640e751f. The entire existing fixture, including wallet goroutines, database reopen and cleanup, now runs inside `synctest.Test`; `synctest.Wait` replaces the allocation WaitGroup. All existing durability assertions remain. The wallet package tests, focused Go 1.25.11 race tests, and wallet lint pass.

wallet/internal/db/data_types.go
yyforyongyu: Updated following our discussion: `RequireAccountNumber` is now removed. `AllocateNextKey` returns `(*AllocatedKey, error)` with the public key, branch and index, plus an optional `KeyOrigin` containing scope/account/fingerprint. Named imported-xpub accounts allocate successfully with a nil origin; child coordinates are still returned and persisted. Branch/index are not duplicated in the origin, and the existing `AddressDerivation`, signing and PSBT consumers are unchanged.

Kvdb numeric selectors still identify wallet-derived BIP44 accounts; its internal numbers for imported accounts cannot be used as those selectors. The existing imported-account test verifies that numeric refusal consumes no child and subsequent named allocation returns index zero.

Implemented in 3feeea58db7eea5e6d4ec57accd4e0566678008c, with the selector check in 4c4f4f5857cdb85a3ac5b9091c2d2b1f56a8f994. Wallet/shared-db/kvdb tests, focused race checks, SQL allocation race tests on both backends, btcd manager smoke tests and lint pass. This supersedes my earlier explanation for retaining the flag.

PR #1387, opened 2026-09-15
wallet: fix sync readiness and isolate targeted rescan failures

GustavoStingelin:
## Summary

Implements Task 455 follow-up identified in [the review of #1363](https://github.com/btcsuite/btcwallet/pull/1363#pullrequestreview-5184916627).

Two error paths could leave live synchronization behavior inconsistent with the reported readiness:

- A worker failure could leave `Info.Synced` true during retry backoff when the wallet and source tips still matched.
- An ordinary targeted-rescan failure could restart the live sync worker and unnecessarily delay notification processing.

## Changes

- Clear live readiness before returning a non-shutdown worker error, so retry backoff reports the wallet as unsynced without changing `SyncedTo`.
- Log ordinary targeted-rescan failures at the scan mailbox boundary and continue live processing without restarting the worker.
- Preserve cancellation, shutdown, full-rewind error propagation, and existing operation-admission behavior.
- Remove the unused state-only `walletState.isSynced` predicate.

Targeted rescans remain serialized with live processing. This does not introduce parallel scanning or change the public rescan protocol: execution failures after a request is queued remain visible only in logs.

## Regression coverage

- Public `Info` reports unsynced during actual worker backoff, even with matching tips.
- The same worker processes the next live block after a targeted-rescan failure and submits it to the Store.
- Successful scans, cancellation, shutdown, and full-rewind failures retain their expected behavior.
- A separate public `Rescan` test verifies that healthy small-gap admission remains available while `Info.Synced` is false.


wallet/controller_test.go
```diff
@@ -1703,6 +1704,76 @@ func TestControllerInfoExactSyncStatus(t *testing.T) {
...
+			case <-failWorker:
+			case <-w.lifetimeCtx.Done():
+			}
+		}).Return(nil, errDBMock).Once()
+
+		startLoadedWalletForTest(t, w)
+		w.wg.Go(w.runSyncLoop)
+
+		synctest.Wait()
+		deps.store.AssertNumberOfCalls(t, "ListTxns", 1)
```
yyforyongyu, line 1761: add an inline doc above explaining why we need to assert here? As normally, `.Once()` already asserts that when test ends

wallet/controller_test.go
```diff
@@ -1760,6 +1831,53 @@ func TestControllerInfoSmallGapReportsUnsynced(t *testing.T) {
...
+	require.NoError(t, err)
+	require.False(t, info.Synced)
+
+	// Act: Submit a scan while the live delivery state is still ready.
+	err = w.Rescan(t.Context(), 50, targets)
+
+	// Assert: The public call succeeds and reaches the syncer mailbox
+	// boundary despite the one-block gap reported by Info.
+	require.NoError(t, err)
+	deps.syncer.AssertExpectations(t)
```
yyforyongyu: This unnecessary check has happened multiple times? e.g., https://github.com/btcsuite/btcwallet/pull/1358#discussion_r3952880623

wallet/controller_test.go
```diff
@@ -1774,6 +1776,89 @@ func TestControllerInfoSyncWorkerBackoff(t *testing.T) {
...
+		).Times(5)
+		expectMatchingRollbackBatch(deps.store, deps.chain)
+		deps.chain.On("IsCurrent").Return(true).Once()
+		deps.chain.On("NotifyBlocks").Return(nil).Once()
+		deps.chain.On("GetBestBlock").Return(
+			&tip.Hash, int32(tip.Height), nil,
+		).Times(3)
+		deps.store.On("ListTxns", mock.Anything, db.ListTxnsQuery{
+			WalletID: w.id, UnminedOnly: true,
+		}).Return(nil, nil).Times(3)
```
yyforyongyu, line 1811: Nice! Love all these precise control of num of calls, and this is the true power of these mockers.

wallet/syncer.go
```diff
@@ -684,6 +684,8 @@ func (s *syncer) run(ctx context.Context) error {
 				return nil
 			}
 
+			s.state.Store(uint32(syncStateSyncing))
```
yyforyongyu, line 687: `syncStateSyncing` now also represents a worker that has exited and is waiting in retry backoff, but its declaration still says only catching up or rewinding; please update that comment to include the non-ready retry state.

wallet/controller_test.go
```diff
@@ -1760,6 +1916,53 @@ func TestControllerInfoSmallGapReportsUnsynced(t *testing.T) {
 	require.False(t, info.Synced)
 }
 
+// TestControllerSmallGapAllowsRescan verifies that scan admission remains
+// available while Info reports the wallet one block behind the source.
+func TestControllerSmallGapAllowsRescan(t *testing.T) {
```
yyforyongyu, line 1894: `TestControllerSmallGapAllowsRescan` repeats the same one-block-gap setup and `Info.Synced == false` assertion as `TestControllerInfoSmallGapReportsUnsynced` above before adding the admission check, so the older test has no distinct state or boundary; please remove the older test and let this one own the combined status/admission contract.

yyforyongyu (changes requested):
Nice usage of the new `synctest`! I think we can keep using it for other controller/syncer related tests. Left a few comments, most are small issues we should address.

wallet/controller_test.go
GustavoStingelin: added

wallet/syncer.go
GustavoStingelin: updated

wallet/controller_test.go
GustavoStingelin: removed

wallet/controller_test.go
GustavoStingelin: yeah, the agents and I always forget that it's already set in this mock test cleanup.

Removed some, commented others

yyforyongyu (approved):
LGTM!

PR #1388, opened 2026-09-15
wallet: remove unconfirmed transactions through TxWriter

Abdulkbk:
Implements [Task 412 — expose transaction invalidation on TxWriter](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/412-transaction-invalidation.md).

Adds `TxWriter.DeleteUnconfirmedTx(ctx, hash)` so a caller such as lnd can drop
an unconfirmed transaction it knows will never confirm, together with the
transactions that depend on it, and get back the coins that branch spent. It is
the replacement for lnd's use of the deprecated `RemoveDescendants`.

Following review the branch is now removed rather than retained as failed
history. A retained row has to be reconciled against the chain if the
transaction confirms after all; removing it leaves chain synchronization as the
single authority, free to record the transaction again from scratch.

### Store

`DeleteUnminedTx` is a new `TxStore` operation rather than a change to
`InvalidateUnminedTx`, so publish-failure cleanup keeps its current behaviour
and its retained history.

Branch discovery is the invalidation workflow's and row removal the leaf
`DeleteTx` flow's, so the shared helper only joins them: every member clears the
spend edges it claimed, drops the outputs it created, then loses its row. That
order is what the schema's `ON DELETE RESTRICT` references require. The legacy
kvdb store already removed the whole branch when asked to invalidate one, so
both workflows share a single helper there.

### Errors

`ErrTxNotFound` for a hash the wallet does not hold, and the new
`ErrTxNotUnconfirmed` for a confirmed, coinbase, or already invalid
transaction. No internal `db` sentinel escapes. The behaviour is identical on
kvdb, SQLite and PostgreSQL, so no backend gate remains.

### Commits

1. `db`: parameterize the unmined-target validator
2. `db`: add the unmined branch deletion workflow
3. `db`: admit branch deletion to the TxStore contract
4. `db`: cover unmined branch deletion
5. `wallet`: expose unconfirmed transaction removal on TxWriter
6. `itest`: cover unconfirmed transaction removal

### Tests

- Store: the shared workflow's ordering and rejections, plus real-store cases
  for a branch whose root also spends an output the wallet does not own, an
  unrelated spend that survives it, and refusals for a confirmed row and an
  unknown hash.
- kvdb: branch removal, and the refusal for a confirmed transaction.
- Wallet: the Store parameters, error translation, lifecycle, and cancellation
  before admission.
- itest: removal with a dependent transaction, a refused confirmed
  transaction, and a refused unknown hash, on every backend. A fourth case
  covers rediscovery: a removed transaction that the chain confirms anyway is
  recorded again with its block and its coins, so removal costs history rather
  than correctness.

### Follow-up

Publish-failure cleanup still retains its branch as `failed` history, so the
same reconciliation question applies there if such a transaction later confirms.
That path is untouched here and is left for its own change.


wallet/wallet.go
```diff
@@ -425,6 +425,10 @@ type Wallet struct {
 	// Callers read it through IsWatchOnly(); the read-side joins against
 	// account_secrets are obsolete now that the value is wallet-level.
 	isWatchOnly bool
+
+	// retainsInvalidTxns reports whether the store keeps invalidated txns
+	// as failed history. The legacy kvdb store deletes them instead.
+	retainsInvalidTxns bool
```
yyforyongyu: I don't think we wanna expose this backend detail just because kvdb is not supported, ideally we should just perform a db backend check in the new API method and return an error there instead, so we don't leak context.

wallet/tx_writer.go
```diff
@@ -109,3 +128,81 @@ func (w *Wallet) handleLabelTx(r labelTxReq) {
...
+// is rejected by chain sync. Invalidate only a tx that can no longer confirm.
+//
+// NOTE: This method is part of the TxWriter interface.
+func (w *Wallet) InvalidateTx(ctx context.Context, hash chainhash.Hash) error {
+	err := w.state.validateStarted()
+	if err != nil {
+		return err
+	}
+
+	if !w.retainsInvalidTxns {
```
yyforyongyu: Given that all db backends implement `Store.InvalidateUnminedTx`, i don't think we should return an error for kvdb. 

Actually, I think it's fine that kvdb deletes the tx, as it's a deprecated backend and we don't provide maintenance, so it's ok to have this divergence, where sql maintains the history but kvdb deletes it.

wallet/tx_writer_backend_test.go
```diff
@@ -0,0 +1,352 @@
+package wallet
```
yyforyongyu: I think we should use `db/itest` and `itest` to replace these tests

wallet/tx_writer.go
```diff
@@ -109,3 +128,81 @@ func (w *Wallet) handleLabelTx(r labelTxReq) {
...
+// untouched. It returns ErrTxNotFound if the wallet has no record of the tx,
+// ErrTxNotInvalidatable if the tx is confirmed, coinbase, or already invalid,
+// and errors.ErrUnsupported on kvdb wallets, whose store deletes invalid txns
+// rather than retaining them.
+//
+// The wallet cannot record an invalidated tx again, so a tx that still confirms
+// is rejected by chain sync. Invalidate only a tx that can no longer confirm.
+//
+// NOTE: This method is part of the TxWriter interface.
+func (w *Wallet) InvalidateTx(ctx context.Context, hash chainhash.Hash) error {
```
yyforyongyu: so a bit background context, the goal of this API is to replace `lnd`'s current usage of `RemoveDescendants` without deleting the tx for the sql, so it should support all db backends.

yyforyongyu (changes requested):
> On SQL an invalidated transaction cannot be recorded again. If it later...the fix belongs to CreateTx reuse policy and needs its own task.

I don't think so. The current invalidate tx design creates the issue so we should fix it here. I think there are two ways to think about the design,
- we can delete the tx in this new API, such that it mimics `RemoveDescendants`, also lets chain sync recreate the tx if it later appears onchain.
- we can keep the tx, but we need to allow the chain syncer to later move it back to `TxStatusPublished`, restore the account graph/edges.

It seems to me that option 1 is the simpler hence better design.

wallet/internal/db/txstore_deleteunminedtx_common.go
```diff
@@ -0,0 +1,123 @@
...
+	"github.com/btcsuite/btcd/chainhash/v2"
+)
+
+// UnminedTxTarget is the normalized root metadata a branch workflow needs.
+type UnminedTxTarget = InvalidateUnminedTxTarget
+
+// DeleteUnminedTxOps is the small backend adapter the shared DeleteUnminedTx
+// workflow needs. Discovery is shared with the invalidation workflow and row
+// removal with the leaf DeleteTx flow.
+type DeleteUnminedTxOps interface {
```
yyforyongyu, line 20: 👍

yyforyongyu, line 20: The target-loading, active-descendant-listing, and spend-clearing methods duplicate the discovery contract in `InvalidateUnminedTxOps`, including the invalidation-specific `LoadInvalidateTarget` name. Could we extract a neutral branch-discovery interface, such as `UnminedBranchOps` with `LoadUnminedTxTarget`, and embed it in both workflows so this shared contract has one owner?

wallet/internal/db/kvdb/txstore.go
```diff
@@ -1289,13 +1289,39 @@ func (s *Store) DeleteTx(ctx context.Context, _ db.DeleteTxParams) error {
...
+	}
+
+	return nil
+}
+
+// DeleteUnminedTx removes an unmined tx branch through the legacy wtxmgr path.
+func (s *Store) DeleteUnminedTx(_ context.Context,
+	params db.DeleteUnminedTxParams) error {
+
+	err := s.removeUnminedTxBranch(params.Txid, db.ErrDeleteRequiresUnmined)
```
yyforyongyu, line 1304: This delete path returns `ErrTxNotFound` after kvdb invalidation because the legacy row has already been removed, while SQL retains the failed row and maps the same request to `ErrTxNotUnconfirmed`. Could we define one backend-neutral outcome here (the simplest may be to treat no active unconfirmed row as `ErrTxNotFound` on every backend) and cover it with a shared test?

docs/developer/tx_invalidation_flows.md
```diff
@@ -1,3 +1,5 @@
+| `InvalidateUnminedTx` | `failed` | `failed` | Publisher-side cleanup rejects one local unmined branch. |
```
yyforyongyu, line 1: These two rows sit before the document heading, so they are detached from the event-outcome table, and the invalidation row is duplicated while `DeleteUnminedTx` is absent from the actual table in section 5. Could we remove these opening rows and add only the deletion row to the section 5 table?

yyforyongyu (approved):
Clean and easy to follow! Looking good - I think we can use a followup PR to properly fix [this case](https://github.com/yyforyongyu/btcwallet-sql-roadmap/issues/39#issuecomment-5709533571), and address the remaining issues found.

PR #1389, opened 2026-09-18
itest: add AccountManager query integration tests

GustavoStingelin:
Add the baseline public AccountManager query coverage from
[Task 430](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/430-account-query-itests.md). Related to #1250.

- Register 12 shared cases for `ListAccounts`, `ListAccountsByScope`,
  `ListAccountsByName`, and `GetAccount` across KVDB, SQLite, and PostgreSQL.
- Cover default, created, and imported accounts; exact and missing filters;
  empty inventories; and the public `ErrAccountNotFound` identity.
- Compare complete, raw `AccountInfo` values against setup/creation results
  and repeat successful queries after a full, locked wallet reopen.
- Add deterministic, test-name-derived seeds to the wallet fixture helper.
- Fix KVDB creation responses to use the same scope-specific public-key
  serialization as subsequent reads. The raw comparisons cover this mismatch.

Each case keeps one query method as its subject, with explicit account setup
and data-only filter tables. No backend-specific assertions or Store reads are
added to the query tests.

Scope note: the current shared import API cannot construct an absent
fingerprint; zero and nonzero fingerprints and absent/zero/nonzero account
numbers are represented. Optional import fingerprints and the new SQL account
shapes remain with [Task 391](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/391-path-bound-watch-only-xpub-accounts.md) and [Task 464](https://github.com/yyforyongyu/btcwallet-sql-roadmap/blob/main/tasks/464-no-chain-sync-account-query-itests.md).

### Test

```sh
for chain in btcd bitcoind neutrino; do
  for db in kvdb sqlite postgres; do
    make itest chain="$chain" db="$db" \
      icase='account manager query' nocache=1
  done
done

go test -race -count=1 ./bwtest ./wallet/internal/db/kvdb
go build ./...
golangci-lint run --timeout=5m --build-tags='itest dev nolog' \
  ./itest ./bwtest ./wallet/internal/db/kvdb
```


itest/account_manager_test.go
```diff
@@ -35,6 +35,147 @@ func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
...
+
+	control, err := w.ListAccountsByName(ctx, "query shared")
+	require.NoError(h, err)
+	require.Len(h, control, 2)
+
+	for _, account := range control {
+		require.Equal(h, "query shared", account.AccountName)
+	}
+
+	missingNames := []string{
```
yyforyongyu, line 626: Could we reduce this to two representative misses: `""`, which proves a non-nil empty-name filter is not treated as an unfiltered list, and one ordinary absent name? The case, prefix, and whitespace variants all exercise the same exact-equality path and can grow without adding distinct behavior.

itest/account_manager_test.go
```diff
@@ -35,6 +35,147 @@ func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
...
+		Scope: scope,
+		Name:  "query shared",
+	})
+	require.NoError(h, err)
+
+	control, err := w.ListAccountsByScope(ctx, scope)
+	require.NoError(h, err)
+	require.NotEmpty(h, control)
+
+	missingScopes := []waddrmgr.KeyScope{
```
yyforyongyu, line 659: `KeyScope` is composite, so I think two cases are useful: one wrong purpose with the correct coin and one correct purpose with the wrong coin, which independently prove both fields participate in matching; the zero scope `{}` can be removed as redundant.

wallet/internal/db/kvdb/accountstore_createderived.go
```diff
@@ -51,6 +51,17 @@ func (s *Store) CreateDerivedAccount(ctx context.Context,
...
+		// still in the creation transaction. Preserve the other fields
+		// returned by the shared workflow.
+		stored, err := loadAccountInfo(
+			ns, ops.scopedMgr, *built.AccountNumber, mgr.WatchOnly(),
+		)
+		if err != nil {
+			return err
+		}
+
+		built.PublicKey = stored.PublicKey
```
yyforyongyu, line 64: could you explain why we need this? how can this value differ? and is this caught by the test?

itest/account_manager_test.go
```diff
@@ -35,6 +35,145 @@ func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
...
+	h.Helper()
+
+	fundedScope, err := accountManagerFundingType.KeyScope()
+	require.NoError(h, err)
+
+	// Funding already materialized its scope. Request an address in every
+	// other scope so lazy defaults exist before capturing the fixture.
+	for _, scope := range waddrmgr.DefaultKeyScopes {
+		if scope != fundedScope {
+			h.NewWalletAddressOfType(
```
yyforyongyu, line 53: Q: I guess atm our new wallet doesn't initialize all these default key scopes like the old wallet yet right?

itest/account_manager_test.go
```diff
@@ -35,6 +35,692 @@ func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
...
+
+	return defaults
+}
+
+// testAccountManagerQueryListSpendable checks ListAccounts against
+// the explicit spendable fixture before and after a locked reopen.
+func testAccountManagerQueryListSpendable(h *bwtest.HarnessTest) {
+	// Arrange: capture default views, then create overlapping scope names.
+	// Use the setup and creation responses as the expected account facts.
+	seed := h.SeedFromTestName()
```
yyforyongyu, line 77: The spendable setup at lines 77-103 is repeated in the scope, name, and get cases, and the watch-only inventory is likewise repeated four times; this gives eight copies of the same fixture algorithm that can drift, so please extract one spendable and one watch-only fixture builder while keeping each query call and assertion visible in its case.

itest/account_manager_test.go
```diff
@@ -35,6 +35,692 @@ func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
...
+		if scope != fundedScope {
+			h.NewWalletAddressOfType(
+				w, waddrmgr.ScopeAddrMap[scope].ExternalAddrType,
+			)
+		}
+	}
+
+	// Use the setup results directly, as with named creation responses.
+	// Keying by scope lets callers spell out exact query memberships.
+	accounts, err := w.ListAccounts(h.Context())
```
yyforyongyu, line 61: `prepareQueryDefaults` obtains every default expectation from `w.ListAccounts`, but `testAccountManagerQueryListSpendable` then uses `w.ListAccounts` as the subject, so an incorrect default-account projection can pass both the pre-reopen and durable comparisons; please build the default expectations independently, for example with one `GetAccount` per canonical scope.

itest/account_manager_test.go
```diff
@@ -35,6 +35,692 @@ func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
...
+	})
+	require.NoError(h, err)
+
+	bip84Suffix, err := w.NewAccount(ctx, wallet.NewAccountParams{
+		Scope: waddrmgr.KeyScopeBIP0084,
+		Name:  "query shared suffix",
+	})
+	require.NoError(h, err)
+
+	want := []wallet.AccountInfo{
```
yyforyongyu, line 105: `prepareQueryDefaults` iterates the canonical `waddrmgr.DefaultKeyScopes`, but the expected inventories and query table then hard-code its four current members, creating a second source of truth; please derive the default rows and scope queries from `DefaultKeyScopes` and layer the named accounts onto those results.

yyforyongyu (changes requested):
Very clean and easy to follow! Left a few comments, should be good to go once addressed!

wallet/internal/db/kvdb/accountstore_createderived.go
GustavoStingelin: Creation returns the derived key with the network-default version, while KVDB reads use the scope-specific version through AccountProperties. For BIP84 on regtest, that means `tpub` versus `vpub`. The underlying key is identical, but `AccountInfo.PublicKey` differs.

Yes, the new list-spendable test catches this.

itest/account_manager_test.go
GustavoStingelin: Right, SQL wallet creation doesn’t initialize the default scopes/accounts, unlike KVDB. Here, `NewWalletAddressOfType` calls the harness’s `ensureAccount`, which creates the missing account before requesting an address. This loop ensures every default scope is present for the query tests.

itest/account_manager_test.go
GustavoStingelin: Reduced this to `""` and `"query absent"`; removed the redundant case, prefix, and whitespace variants.

itest/account_manager_test.go
GustavoStingelin: Removed the zero scope and kept the two cases that vary purpose and coin independently.

itest/account_manager_test.go
GustavoStingelin: Extracted one spendable and one watch-only fixture builder, returning named account results. Wallet setup, expected membership, query calls, assertions, and reopen checks remain visible in each test.

itest/account_manager_test.go
GustavoStingelin: The list tests now capture default expectations with one `GetAccount` per canonical scope; `prepareQueryDefaults` only materializes the defaults. The `GetAccount` test uses `ListAccounts` for its snapshot and independently checks the known fixture identities, key counts, and balances so shared projection bugs are not hidden by agreement between the two reads.

itest/account_manager_test.go
GustavoStingelin: Default expectations and scope-query loops now use `waddrmgr.DefaultKeyScopes`, with the named accounts added explicitly to the expected inventories.

itest/account_manager_test.go
```diff
@@ -35,6 +35,652 @@ func canonicalAccountKey(h *bwtest.HarnessTest, key []byte) []byte {
...
+	})
+	require.NoError(h, err)
+
+	bip44Shared, err := w.NewAccount(ctx, wallet.NewAccountParams{
+		Scope: waddrmgr.KeyScopeBIP0044,
+		Name:  "query shared",
+	})
+	require.NoError(h, err)
+
+	bip84Suffix, err := w.NewAccount(ctx, wallet.NewAccountParams{
```
yyforyongyu, line 85: we now have `h.CreateTestAccount` from the signer itest PR, can it be used here?

yyforyongyu (approved):
LGTM!

Abdulkbk (approved):
lgtm

PR #1390, opened 2026-09-18
wallet: guard missing notification tx details

cuishuang:
## Change Description

`wtxmgr.Store.UniqueTxDetails` returns `(nil, nil)` when a transaction cannot be found in the requested mined or unmined location.

Both `notifyUnminedTransaction` and `notifyMinedTransaction` currently continue processing a nil result. This can cause a nil pointer dereference through `details.Hash` or `makeTxSummary`.

This is a regression of the behavior restored by #540. The nil checks previously lived in `chainntfns.go`, but were lost when the transaction lookups were moved into `NotificationServer` in commit 61d85e4.

Return early from both notification paths when transaction details are missing. Add regression tests confirming that missing details do not panic, emit a transaction notification, or modify the notification server's coalescing state.

## Steps to Test

```bash
go test ./wallet -run \
  'TestNotify(Unmined|Mined)TransactionMissingDetails' -count=1

go test ./wallet -count=1

golangci-lint run ./wallet/...
```

### Testing


wallet/notifications_test.go
```diff
@@ -0,0 +1,78 @@
...
+			server.notifyUnminedTransaction(
+				tx, ns, chainhash.Hash{1},
+			)
+
+			return nil
+		})
+	})
+	require.NoError(t, err)
+	require.Empty(t, client)
+	require.Same(t, existing, server.currentTxNtfn)
```
Lrifton92, line 41: `notifyUnminedTransaction` never assigns `s.currentTxNtfn`, it only reads it for the sanity-check log (`notifications.go:232`), so this `require.Same` cannot fail whatever the function does. Setting `existing` does have a real effect: it makes the unguarded path dereference `details.Hash` in that `log.Errorf` rather than inside `makeTxSummary`. Running the test once with `currentTxNtfn` nil and once non-nil would cover both dereference sites, and the `Same` assertion could go.

wallet/notifications.go
```diff
@@ -223,6 +223,10 @@ func (s *NotificationServer) notifyUnminedTransaction(dbtx walletdb.ReadTx,
 		return
 	}
 
+	if details == nil {
```
Lrifton92, line 226: Nit: the "It's possible that the transaction was not found ... so we'll avoid notifying it" comment (and the TODO) still sits above the `UniqueTxDetails` call at :211, but it now describes this guard. Moving it down here, perhaps with a debug log, would keep the reason next to the early return.

Lrifton92 (approved):
Reviewed at 3feaac9.

The regression is real and the fix restores the pre-61d85e4 behaviour. Before that commit, `addRelevantTx` only called the notifier when `details != nil`. Its diff drops both checks when the `UniqueTxDetails` lookups move into `notifyUnminedTransaction` and `notifyMinedTransaction`, and nothing re-adds them there. `UniqueTxDetails` returns `(nil, nil)` whenever the raw unmined record (`wtxmgr/query.go:239-241`) or the mined record for that block (`:246-248`) is absent. On master a nil result then reaches `details.Hash` (unmined, when a mined notification is being coalesced) or `makeTxSummary`.

I checked that the tests protect the fix. With either guard removed (`wallet/notifications.go:226-228` or `:301-303`), the matching test fails on `require.NotPanics` with `runtime error: invalid memory address or nil pointer dereference`. At head, both tests pass their assertions. On Windows they still fail in `TempDir` cleanup because the wallet db is left open, but that is pre-existing: `testWallet` returns a no-op cleanup (`wallet/example_test.go:44`), and `TestCreateWatchingOnly` and `TestLabelTransaction` fail the same way on master. No added line exceeds 80 columns at tab width 8.

On reachability, since the tests use a hash that is simply absent from the store. The mined nil looks unreachable from `addRelevantTx`: `insertMinedTx` either returns `ErrDuplicateTx` (so `addRelevantTx` exits early) or writes the record in the same db transaction before the notification. The unmined nil can happen: `insertMemPoolTx` returns `nil` rather than `ErrDuplicateTx` when one of the tx's outputs is already in the unspent bucket (`wtxmgr/unconfirmed.go`, the `existsRawUnspent` loop), so `addRelevantTx` carries on and the unmined lookup finds nothing. Either way it means the store and the notification disagree. A `log.Debugf` (unmined, which the existing TODO says is expected) or a `log.Warnf` (mined, which should be impossible) before returning would keep that visible instead of silently dropping it. Non-blocking.

Two small points inline.

PR #1392, opened 2026-09-20
itest: verify allocated key results

yyforyongyu:
## Summary

Verify allocated keys and persisted address metadata through integration tests.

Implements roadmap Task 465

## Change Description

- Check external and internal key results, address listings, and metadata after reopening the wallet.
- Check concurrent allocation uniqueness and continued allocation after reopening for SQLite and PostgreSQL.

## Notes

Depends on https://github.com/btcsuite/btcwallet/pull/1386. Review scope begins after that PR; this change contains only the integration tests.

The focused SQLite/PostgreSQL matrix passes across btcd, bitcoind, and neutrino. PostgreSQL race checks pass. SQLite race checks encounter an existing global logger race, reproduced on the unchanged base.


itest/address_manager_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// expectedAllocatedAddress independently derives address metadata from the
+// fixture account and requested branch; only the chosen child index comes from
+// allocation, since BIP32 permits the allocator to skip invalid children.
+func expectedAllocatedAddress(h *bwtest.HarnessTest,
```
yyforyongyu: this should be `assertAllocatedAddr`?

itest/address_manager_test.go
```diff
@@ -0,0 +1,148 @@
...
+}
+
+// testAddressManagerAllocateKey verifies an excluded account's allocated keys
+// match its XPub and retain complete point/list visibility after reopening.
+func testAddressManagerAllocateKey(h *bwtest.HarnessTest) {
+	// Modern kvdb cannot persist NoChainSync, so it cannot supply this fixture.
+	if *dbBackend != string(wallet.DBBackendSQLite) &&
+		*dbBackend != string(wallet.DBBackendPostgres) {
+
+		h.Skip("NoChainSync key allocation requires SQL")
```
yyforyongyu, line 72: check kvdb

itest/address_manager_test.go
```diff
@@ -146,3 +148,135 @@ func testAddressManagerAllocateKey(h *bwtest.HarnessTest) {
 	require.NoError(h, err)
 	require.ElementsMatch(h, wantList, listed)
 }
+
+// testAddressManagerAllocateKeyConcurrent verifies unused custom-scope keys
+// have distinct durable locators across concurrent requests and wallet reopen.
+func testAddressManagerAllocateKeyConcurrent(h *bwtest.HarnessTest) {
```
yyforyongyu, line 153: we can use  the newer `synctest` here?

itest/address_manager_test.go
yyforyongyu: Keeping `expectedAllocatedAddress`: it derives and returns the expected metadata from the account XPub. It does not compare the allocation result; those assertions remain visible in the test cases. An `assertAllocatedAddr` name would imply a comparison that this helper does not perform.

itest/address_manager_test.go
yyforyongyu: Changed both cases to skip with `*dbBackend == string(wallet.DBBackendKVDB)`, folded into the original commits. The conversion is needed because the flag is a string and the constant has type `wallet.DBBackend`. Each reference has a statement-local staticcheck exception because the constant is deprecated. Focused SQLite/PostgreSQL tests and kvdb skip checks pass; CI is green.

itest/address_manager_test.go
yyforyongyu: Keeping the `sync.WaitGroup` here. The harness starts the wallet request loop before these callers run. `AllocateNextKey` creates a response channel in the caller, which that existing wallet goroutine sends on. Wrapping just the callers in `synctest.Test` would create a bubbled channel used from outside the bubble, violating synctest isolation. This case also uses real database and chain infrastructure. The WaitGroup joins all six callers before assertions and reload without changing the harness lifecycle. See the [synctest isolation rules](https://pkg.go.dev/testing/synctest#hdr-Isolation).

PR #1393, opened 2026-09-20
itest: cover public address batch contracts

yyforyongyu:
## Summary

Add public address-batch integration coverage on real wallet backends.

Implements roadmap Task 461

## Change Description

- Verify independently derived, ordered SQL batches, complete metadata, durable
  point/list visibility, and branch progress after reopening.
- Verify count validation and non-mutation for SQL, excluded accounts, and
  modern kvdb, including its unsupported-batch contract.
- Use the existing harness and case registry. Focused cases pass across SQLite,
  PostgreSQL, and kvdb with btcd, bitcoind, and neutrino.

## Notes

Stacked on #1379; review scope starts after that prerequisite, with
`task-bulk-address` as the base.

The base has no AddressManager integration-test file. The conventional
`itest/address_manager_test.go` keeps these cases separate from account,
manager-lifecycle, and UTXO tests without adding fixture infrastructure.


itest/address_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+
+	return want
+}
+
+// testAddressManagerAllocateBatch proves ordered SQL batches and their complete
+// public metadata survive reopening without reusing previously delivered keys.
+func testAddressManagerAllocateBatch(h *bwtest.HarnessTest) {
+	// Kvdb cannot allocate atomic batches; its refusal is a separate contract.
+	if *dbBackend != string(wallet.DBBackendSQLite) &&
```
yyforyongyu: we should do `*dbBackend == wallet.DBBackendKVDB` and skip?

itest/address_manager_test.go
```diff
@@ -0,0 +1,194 @@
...
+	"github.com/btcsuite/btcd/btcutil/v2/hdkeychain"
+	"github.com/btcsuite/btcwallet/bwtest"
+	"github.com/btcsuite/btcwallet/waddrmgr"
+	"github.com/btcsuite/btcwallet/wallet"
+	"github.com/stretchr/testify/require"
+)
+
+// batchAddressExpectations derives a fresh account's children from public
+// material before allocation, so returned addresses cannot define the oracle.
+func batchAddressExpectations(h *bwtest.HarnessTest,
```
yyforyongyu: we usually prefer linear code so it's easier to follow, however, if there are indeed flow that can be shared, we should name this clearly - so this should be `assertBatchAddrDerived`?

itest/address_manager_test.go
```diff
@@ -192,3 +192,123 @@ func testAddressManagerAllocateBatch(h *bwtest.HarnessTest) {
...
 }
+
+// testAddressManagerRejectSQLBatch proves invalid counts precede receiving
+// policy checks and every rejected request leaves public state unchanged.
+func testAddressManagerRejectSQLBatch(h *bwtest.HarnessTest) {
+	// Excluded accounts are SQL-only; kvdb refusal has its own fixture.
+	if *dbBackend != string(wallet.DBBackendSQLite) &&
+		*dbBackend != string(wallet.DBBackendPostgres) {
+
+		h.Skip("excluded account batch checks require SQL")
```
yyforyongyu, line 202: same here, check kvdb and skip instead

itest/address_manager_test.go
yyforyongyu: Changed to a direct kvdb skip using `*dbBackend == string(wallet.DBBackendKVDB)`. The conversion matches the string flag. The guard has a narrow, documented staticcheck exception because these tests intentionally exercise the supported legacy backend. CI passes.

itest/address_manager_test.go
yyforyongyu: Renamed the helper to `deriveBatchAddressExpectations` to make its role explicit. It derives independent expected metadata before allocation and is shared by the three scenarios; it does not check an actual batch. The allocation calls and result assertions remain inline in each scenario.

itest/address_manager_test.go
yyforyongyu: Changed this guard to skip kvdb directly with `*dbBackend == string(wallet.DBBackendKVDB)`, matching the allocation case. The conversion handles the string flag, and the documented staticcheck exception covers intentional use of the legacy backend. CI passes.

PR #1395, opened 2026-09-22
wallet: harden psbt funding admission

Abdulkbk:
First of a few small follow-ups to #1359, split so each one is easy to review
on its own. This is the cheapest of them: two independent fixes, no new
machinery.

### Judge the packet before consulting the syncer

`FundPsbt` asked the wallet whether it was synced before it looked at the
packet, so a caller handing over a malformed packet paid for a syncer round
trip to be told something reading the packet alone would have settled.
`validateFundIntent` takes no wallet state and performs no lookups, so it is
safe to run first.

A caller that is both unsynced and holding a bad packet now hears about the
packet. That is the more useful answer: the packet is malformed whatever the
wallet is doing.

This also makes the malformed-packet table honest. It previously registered a
`syncState` expectation because the old order needed one, which undercut its
claim to be proving that nothing gets consulted. It now registers no
dependency expectation at all, so anything the wallet touches on the way to a
rejection fails the test.

### Refuse an empty taproot key spend signature

An empty but present `TaprootKeySpendSig` passed admission and was then
dropped, because the check asked for the field's length while the field means
something by being there at all.

This is the reasoning already applied to `FinalScriptSig` and
`FinalScriptWitness` in #1359, and the key spend signature is the remaining
single-record field that was still testing length. The list-valued fields keep
testing length, where an empty list genuinely is the absence of a record.

### Still to come

Separate PRs, in rough order: delegating structural validation to the
canonical psbt parser (which should shrink the validator and its tests
considerably), reconciling caller metadata against the actual spend script,
then file and test consolidation once that shape settles.

Two findings belong to other tasks rather than here: `SignPsbt` still refuses
the multiple derivations funding now preserves, so the documented
`FundPsbt` → `SignPsbt` path fails for a multisig packet; and a rejected
metadata contradiction still consumes a change address, since reconciliation
happens after authoring.

### Testing

Both fixes are mutation-checked: restoring the old validation order fails the
malformed-packet table, and testing the key spend signature by length fails
its new case.


yyforyongyu (approved):
LGTM!

PR #1396, opened 2026-09-22
multi: delete unmined tx branches instead of retaining them

Abdulkbk:
Follow-up to #1388, covering the review comments left after approval and the
remaining items in yyforyongyu/btcwallet-sql-roadmap#39.

### What changed

- Publication cleanup deletes the branch instead of marking it `failed`. A
  publish error does not prove the tx can never confirm, and a retained
  terminal row cannot be re-recorded: if the tx does confirm later, the block
  batch fails with `ErrTxAlreadyExists` and the sync tip stops advancing.

- Branch deletion now reaches invalidated descendants. The delete flow lists
  candidates with a terminal-inclusive query, so a descendant already marked
  `failed` is removed along with its root instead of being left behind.
  Conflict handling, rollback and leaf deletion keep the existing
  status-filtered listing.

- A root the wallet already made terminal reports `ErrTxNotFound` on both SQL
  and kvdb. kvdb keeps no such row, so both backends now answer alike.

- Shared root discovery is extracted into `UnminedBranchOps`, used by both the
  invalidate and delete workflows.

### Tests

New store itests cover lease and replacement cleanup, rollback on failure, and
the invalidated-descendant case. A public itest proves a deleted branch is
re-learned after a wallet reload while the mempool still holds it.

### For review

Two points that deserve an explicit decision rather than my judgement:

1. `InvalidateUnminedTx` no longer has a production caller. Keeping it costs
   nothing, but it is test-only as of this PR.

2. Deleting rather than retaining erases wallet-authored terminal rows. That is
   a retention-policy change, not only a bug fix.


Abdulkbk:
I wonder if we should remove `InvalidateUnminedTx` altogether since it has no prod caller now

wallet/internal/sql/pg/queries/transactions.sql
```diff
@@ -321,6 +351,25 @@ WHERE
...
+-- - Matches DeleteUnminedTransactionByHash but drops the status predicate, so
+--   a row an earlier event made `replaced`, `failed` or `orphaned` is removed
+--   too. Confirmed rows stay protected by the block_height check.
+-- - Used by branch deletion only, which takes a whole branch including
+--   descendants already made terminal.
+-- - Leaf DeleteTx must keep using DeleteUnminedTransactionByHash, whose status
+--   predicate is what stops it erasing retained history.
+-- Performance:
+-- - Targets at most one row by `(wallet_id, tx_hash)`.
+DELETE FROM transactions
```
yyforyongyu: nit: we usually do `.sql` files update in an independent commit so it's easier to reason about, non-blocking and just for future reference.

wallet/internal/db/txstore_deleteunminedtx_common.go
```diff
@@ -94,6 +78,33 @@ func DeleteUnminedTxWithOps(ctx context.Context, params DeleteUnminedTxParams,
...
+	target, err := ops.LoadUnminedTxTarget(ctx, params.WalletID, params.Txid)
+	if err != nil {
+		return UnminedTxTarget{}, fmt.Errorf("load delete tx target: %w", err)
+	}
+
+	if !target.HasBlock && !target.IsCoinbase &&
+		!IsUnminedStatus(target.Status) {
+
+		return UnminedTxTarget{}, fmt.Errorf("tx %s: %w", target.TxHash,
+			ErrTxNotFound)
```
yyforyongyu: we now return `ErrTxNotFound` for a retained terminal root, but the sentinel, `TxWriter` interface and method comments in `wallet/tx_writer.go` still say an already-invalid tx returns `ErrTxNotUnconfirmed`. Can we update those three places? The not-found description should also make clear that SQL may still return the retained row from `GetTx`.

wallet/tx_publisher.go
```diff
@@ -213,11 +213,11 @@ func (w *Wallet) handleBroadcast(r broadcastReq) {
 	// First, we'll attempt to add the tx to our wallet's DB. This will
 	// allow us to track the tx's confirmation status, and also
 	// re-broadcast it upon startup. If any of the subsequent steps fail,
-	// this tx is invalidated via InvalidateUnminedTx.
+	// this tx is removed via DeleteUnminedTx.
```
yyforyongyu: can we update the Store contract in `wallet/internal/db/interface.go` too? `InvalidateUnminedTx` still lists publisher-side rejection as its use case. With this change, that belongs to `DeleteUnminedTx`, whose comment should also mention terminal descendants are removed and terminal unconfirmed roots return `ErrTxNotFound`.

wallet/internal/sql/pg/queries/transactions.sql
```diff
@@ -140,6 +140,36 @@ WHERE
     AND t.tx_status IN (0, 1)
 ORDER BY t.received_time DESC, t.id DESC;
 
+-- name: ListUnminedTransactionsWithInvalid :many
```
yyforyongyu: can we reuse `ListTransactionsWithoutBlock` here? It already has the same columns, filters and ordering, including terminal rows. Same for SQLite. Wiring the deletion adapters to that query would avoid the duplicate SQL and generated code.

wallet/tx_publisher.go
```diff
@@ -239,25 +239,25 @@ func (w *Wallet) handleBroadcast(r broadcastReq) {
...
 	}
 
-	// If the tx was rejected for any other reason, then we'll invalidate it
-	// from the tx store, as otherwise, we'll attempt to continually
-	// re-broadcast it, and the UTXO state of the wallet won't be accurate.
-	removeErr := w.invalidateUnminedTx(r.ctx, r.tx)
+	// If the tx was rejected for any other reason, then we'll remove it from
+	// the tx store, as otherwise, we'll attempt to continually re-broadcast
+	// it, and the UTXO state of the wallet won't be accurate.
+	removeErr := w.deleteUnminedTx(r.ctx, r.tx)
```
yyforyongyu: can we extend `testBroadcastRejected` in `itest/tx_publisher_test.go` to assert `GetTx` returns `wallet.ErrTxNotFound`? It currently skips checking persistence because the old behavior differed by backend, so it can still pass if cleanup leaves a terminal row behind. We can add the assertion to that case and update its comment without adding another test.

wallet/tx_writer_test.go
```diff
@@ -181,6 +181,12 @@ func TestDeleteUnconfirmedTxStoreErrors(t *testing.T) {
 				db.ErrDeleteRequiresUnmined),
 			wantErr: ErrTxNotUnconfirmed,
 		},
+		{
+			name: "terminal tx",
```
yyforyongyu: nit: this seems to repeat the `missing tx` case above. Both mock `db.ErrTxNotFound` and expect `ErrTxNotFound`; only the message text differs, so this doesn't exercise terminal-state handling. Can we drop this row and keep that coverage in the Store test? Non-blocking.

wallet/internal/db/kvdb/txstore_test.go
```diff
@@ -4308,3 +4308,41 @@ func TestDeleteUnminedTxRejectsConfirmed(t *testing.T) {
 	)
 	require.ErrorIs(t, err, db.ErrDeleteRequiresUnmined)
 }
+
+// TestDeleteUnminedTxRejectsRemovedTx verifies that removing a tx the legacy
+// store no longer holds reports the missing-tx identity, which is how kvdb
+// reaches the same outcome SQL gives for a retained terminal row.
+func TestDeleteUnminedTxRejectsRemovedTx(t *testing.T) {
```
yyforyongyu: nit: can we fold the second-delete assertion into `TestDeleteUnminedTxRemovesBranch` instead? The recorded-to-removed-to-missing transition is useful, but we can keep it covered without another full fixture for unchanged kvdb code. Non-blocking.

yyforyongyu (changes requested):
> I wonder if we should remove InvalidateUnminedTx altogether since it has no prod caller now

yeah think we can just remove it in a follow-up PR to keep this the current one focused.

yyforyongyu (approved):
LGTM

